PR #5852, opened 2022-10-20
feat(eslint-plugin): [no-array-delete] add rule

### LINKED ISSUE #4432: Rule proposal: no-array-delete
### Description

Deleting an array element can lead to unexpected (for some) behaviour since it leaves 'length' unchanged. Some might say that using delete on array elements should best be avoided and there are alternatives - splice() seems to be preferred.

Having a rule to prevent delete being used on arrays would be beneficial.

### Fail

```ts
> var myArray = [ 1, 2, 3 ];
undefined
> delete myArray[1];
true
> myArray
(3) [1, empty, 3]
> myArray[1]
undefined
> myArray.length
3
```

### Pass

```ts
> var myArray = [ 1, 2, 3 ];
(3) [1, 2, 3]
> myArray.splice(1,1)
[2]
> myArray
(2) [1, 3]
```

mahdi-farnia:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4432

## Overview

Before:
Using `delete` operator for arrays in JavaScript leaves an empty value & unchanged length property as the actual index is not deleted, Just its value replaced by empty.

```ts
declare const array: any[];
declare const i: number;

delete array[i];
```

After:

```ts
declare const array: any[];
declare const i: number;

array.splice(i, 1);
```

Closes #4432


packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,57 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+import * as ts from 'typescript';
+
+import * as util from '../util';
+
+export default util.createRule<[], 'arrayDeleteViolation'>({
```
JoshuaKGoldberg, line 6: Nit: no need for explicit type parameters here. They can be inferred:

```suggestion
export default util.createRule({
```

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,57 @@
...
+  name: 'no-array-delete',
+  meta: {
+    docs: {
+      description: 'Disallow delete operator for arrays',
+      recommended: 'error',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      arrayDeleteViolation:
```
JoshuaKGoldberg, line 15: Nit: "violation" is implied by the fact that it's a message.

```suggestion
      arrayDelete:
```

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,57 @@
...
+        'Using delete operator for arrays are forbidden. Use array.splice instead.',
```
JoshuaKGoldberg, line 16: Nit: just tweaking the phrasing to emphasize that what the user is doing is a bad idea ("dangerous"). In the past we've had user irritation about things described as "forbidden" without any indication that it _should_ be forbidden.

```suggestion
        'Using the delete operator on an array is dangerous. Use array.splice instead.',
```

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
```diff
@@ -0,0 +1,51 @@
...
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2015,
+    tsconfigRootDir: rootDir,
+    project: './tsconfig.json',
+  },
+  parser: '@typescript-eslint/parser',
+});
+
+ruleTester.run('no-array-delete', rule, {
```
JoshuaKGoldberg, line 14: Let's add a few more test cases to be really thorough:

* Computed values that are always a number: `arr[Math.random() ? 1 : 1]`
* Computed values that aren't always a number: `arr[Math.random() ? 1 : ""]`
* What if the element is a Sequence Expression: `arr[something(), 1]`
    * I think this is a bug in the fixer right now. Before fixing, the code would delete element `1`; after, it passes `something(), 1` as two function arguments.

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,57 @@
...
+    schema: [],
+    type: 'problem',
+    fixable: 'code',
+  },
+  defaultOptions: [],
+  create(context) {
+    return {
+      "UnaryExpression[operator='delete']"(
+        node: TSESTree.UnaryExpression,
+      ): void {
```
JoshuaKGoldberg, line 27: We can get even more precise with the query here:

```suggestion
      "UnaryExpression[operator='delete'] MemberExpression[computed]"(
        node: TSESTree.MemberExpressionComputedName & { parent: TSESTree.UnaryExpression },
      ): void {
```

Two main benefits:
* It gives more precision to the type system on what the node shapes are
* It succinctly filters away undesirable node shapes that this rule wouldn't want to check for, like a delete on a non-computed member expression (`delete a.b`)

That first benefit is quite nice because it'll save the rule from having to call `getChildAt`. We generally prefer not to use that API because it's imprecise and doesn't give much information to the type system on the types of the children. Instead of `expression.getChildAt(0)`, with this suggested change, you can ask for `node.object` (roughly).

Overall: instead of looking at `originalNode` and accessing children from it, try to stay within the ESTree AST -- only dipping into TypeScript-land when you're sure you'll need type information.

JoshuaKGoldberg (changes requested):
Thanks for sending this in - I'm excited to see the `no-array-delete` rule land soon! 🙌 

Requesting changes on some structural smoothing-out and more test cases, as well as splitting out the unrelated code into their own PRs.

Let me know if any of this is unclear - and thanks for working with us on it all @mahdi-farnia!

JoshuaKGoldberg:
Ping @mahdi-farnia - is this still something you have time to work on? No worries if not, just checking in - thanks!

mahdi-farnia:
> Ping @mahdi-farnia - is this still something you have time to work on? No worries if not, just checking in - thanks!

Hi, Mmm... No nothing left hear.
But it seems I forget to correct my mistake 😅

That should be done in a few minutes...👍

JoshuaKGoldberg:
@mahdi-farnia I normally wait till folks re-request review - but it looks like this might be ready for review. Do you want me to take another look? :slightl

mahdi-farnia:
> @mahdi-farnia I normally wait till folks re-request review - but it looks like this might be ready for review. Do you want me to take another look? 🙂 

Omg, I thought i should wait till you review them and not to mention you again. 🫤 

Yes, please 😄

JoshuaKGoldberg:
Well, darn, sorry to keep you waiting! I filed #6492 to make this explicit in our docs. ❤️

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      output: `
+declare const arr: any[];
+declare const i: number;
+
+arr.splice(i, 1);
+      `,
+      errors: [{ messageId: 'arrayDeleteViolation' }],
+    },
+  ],
+});
```
JoshuaKGoldberg, line 51: Some missing test cases:

* Dynamically computed member object (e.g. `delete getTarget()[0]`)
* Dynamically computed member property (e.g. `delete data[getKey()]`) 
* Targets that are a union type including array (e.g. `string | string[]`)
* Targets that are declared as `Array<something>`, not just `something[]`
* Multi-dimensional arrays (`something[][]`)
* Generics:
    * `function annoyingEdgeCase1<T extends unknown[]>(t: T) { delete t[0]; }`
    * `function annoyingEdgeCase1<T extends unknown>(t: T[]) { delete t[0]; }`

JoshuaKGoldberg (approved):
Progress! I think we're going to need a good amount of testing on this. Deletions and arrays together can be surprisingly tricky!

JoshuaKGoldberg (changes requested):
wrong button...

mahdi-farnia:
So, After a while and some laziness...
I apply your suggestions and fix some other cases, Like NumberLike types for computed `MemberExpression`. (enums)

I did every thing that i should (I think). If any thing else left, PLEASE let me know 🙏 .

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,57 @@
...
+import * as ts from 'typescript';
+
+import * as util from '../util';
+
+export default util.createRule<[], 'arrayDeleteViolation'>({
+  name: 'no-array-delete',
+  meta: {
+    docs: {
+      description: 'Disallow delete operator for arrays',
+      recommended: 'error',
```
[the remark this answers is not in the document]
JoshuaKGoldberg: @mahdi-farnia this thread is marked as resolved but the code still has the value set as `'error'`. I see the same thing with at least one other thread (`arrayDeleteViolation`). Did you push up all the changes you wanted to?

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,90 @@
...
+      "UnaryExpression[operator='delete'] > MemberExpression[computed]"(
+        node: TSESTree.MemberExpressionComputedName & {
+          parent: TSESTree.UnaryExpression;
+        },
+      ): void {
+        const parserServices = util.getParserServices(context);
+        const checker = parserServices.program.getTypeChecker();
+        const originalNode = parserServices.esTreeNodeToTSNodeMap.get(node);
+
+        const target = originalNode.getChildAt(0),
```
JoshuaKGoldberg, line 36: We have some newly enabled lint rules you'll need to fix for.

JoshuaKGoldberg (changes requested):
Still some requested changes pending?

mahdi-farnia:
Sorry for late response 
I didn't check my emails 

I'm going to check requested changes...

Is there anything else left to handle? ( I'm curious about final merging time, if it's completed )

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,90 @@
...
+  checker: ts.TypeChecker,
+): boolean {
+  for (const t of unionTypeParts(type)) {
+    if (checker.isArrayType(t)) {
+      return true;
+    }
+  }
+
+  return false;
+}
```
JoshuaKGoldberg, line 80: [Nitpicking] Can this function's body be replaced by `return unionTypeParts(type).some(checker.isArrayType)`?

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function isTypeNumberOrNumberLiteralOrNumberLikeType(type: ts.Type): boolean {
```
JoshuaKGoldberg, line 82: [Request] You can use `tsutils.isTypeFlagSet` instead of this big name.

JoshuaKGoldberg:
I think all the comments I posted in this PR are about it. 🙂

packages/eslint-plugin/src/rules/no-array-delete.ts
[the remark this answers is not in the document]
mahdi-farnia: It seems it's my mistake! (I did not mean that!)

mahdi-farnia:
Done 🙃 (for now!)

JoshuaKGoldberg (changes requested):
You've still got some lint failures annotated in this PR's files diff. Please address those and then I can re-review. Thanks!

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,92 @@
...
+        const parserServices = util.getParserServices(context);
+        const checker = parserServices.program.getTypeChecker();
+        const originalNode = parserServices.esTreeNodeToTSNodeMap.get(node);
+
+        const target = originalNode.getChildAt(0);
+        const key = originalNode.getChildAt(2);
+
+        const targetType = util.getConstrainedTypeAtLocation(checker, target);
+
+        if (!util.isTypeAnyArrayType) {
```
armano2, line 41: are you sure about this?

mahdi-farnia:
The lint problem resolved, that was due to uninstalled (i don't know why) package eslint-plugin-unicorn.
Because of that i could not figure out linting problems

packages/eslint-plugin/src/rules/no-array-delete.ts
mahdi-farnia: better solution? I'm not an expert, if you notice some better options i would like to know...

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        const keyType = util.getConstrainedTypeAtLocation(checker, key);
+
+        if (
+          !isTypeArrayTypeOrArrayInUnionOfTypes(targetType, checker) ||
+          !isTypeNumberOrNumberLiteralOrNumberLikeType(keyType)
+        ) {
+          return;
+        }
```
armano2, line 52: can we split this condition to exit early?

```suggestion
        if (!isTypeArrayTypeOrArrayInUnionOfTypes(targetType, checker)) {
          return;
        }

        const keyType = util.getConstrainedTypeAtLocation(checker, key);

        if (!isTypeNumberOrNumberLiteralOrNumberLikeType(keyType)) {
          return;
        }
```

in best case scenario this should reduce amount of type requests

mahdi-farnia: Thank you ❤️, didn't see that

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,92 @@
...
+  defaultOptions: [],
+  create(context) {
+    return {
+      "UnaryExpression[operator='delete'] > MemberExpression[computed]"(
+        node: TSESTree.MemberExpressionComputedName & {
+          parent: TSESTree.UnaryExpression;
+        },
+      ): void {
+        const parserServices = util.getParserServices(context);
+        const checker = parserServices.program.getTypeChecker();
```
armano2, line 33: can you move those 2 lines directly to `create(context) {`

packages/eslint-plugin/src/rules/no-array-delete.ts
armano2: `isTypeAnyArrayType` is a function, but you are not calling it you are just checking if its present

mahdi-farnia: HOLY CRAP, thank you. I DID NOT WRITE CODE LIKE THIS IN ANY PERIOD OF MY LIFE.

mahdi-farnia:
@armano2 thank you so much. your review saved a review and an edit by me.

mahdi-farnia:
Actually this is my first **serious** contribution to an open-source project.

If you see something strange in my code or something that experts don't do in their code, please don't blame me

packages/eslint-plugin/docs/rules/no-array-delete.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+### ✅ Correct
+
+```ts
+declare const array: unknown[];
+declare const index: number;
+
+array.splice(index, 1);
+```
```
JoshuaKGoldberg, line 32: [Docs] Many rules have a _When Not To Use It_ heading (#7257). I think these docs would benefit from that.

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,90 @@
...
+        const keyType = util.getConstrainedTypeAtLocation(checker, key);
+
+        if (!isTypeNumberOrNumberLiteralOrNumberLikeType(keyType)) {
+          return;
+        }
+
+        context.report({
+          node,
+          messageId: 'arrayDelete',
+          suggest: [
```
JoshuaKGoldberg, line 55: [Bug] This test case causes a suggested output that's a type error:

```ts
declare const arr: unknown[];
declare const i: number | string;

delete arr[i];
```

The suggestion fixer shouldn't suggest something that's a type error - so I think it shouldn't suggest anything at all in this case.

JoshuaKGoldberg (changes requested):
🔥 Looks great - very close to merge I think! I did a closer pass and think it's just a few small-ish things blocking review now. Mostly the bug around suggestions.

mahdi-farnia:
Seems this path is too long 😄

I'll apply suggested changes in the next weekend... (or as soon as I can)

mahdi-farnia:
After all, please help me resolving conflicts with `packages/eslint-plugin/src/configs/strict.ts` 🙏

JoshuaKGoldberg:
You should be able to run the command mentioned at the top of the file:

> ```js
> // For developers working in the typescript-eslint monorepo:
> // You can regenerate it using `yarn generate:configs`
> ```

My typical strategy for autogenerated files that are giving me Git merge woes is:

1. `git checkout main -- path/to/file` to undo any of my branch's changes to the file
2. Re-run `yarn` to reinstall and rebuild everything
3. Run the command to generate them: here, that's running `yarn generate:configs` from the `packages/eslint-plugin` directory

Does that help?

JoshuaKGoldberg:
👋 Hey @mahdi-farnia! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

mahdi-farnia:
> 👋 Hey @mahdi-farnia! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

I was waiting for you :) (i requested a review)

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
```diff
@@ -0,0 +1,390 @@
...
+declare const i: string;
+delete arr[i];
+    `,
+    `
+declare const multiDimesnional: Array<unknown>[][][][];
+declare const i: number;
+delete multiDimesnional[i][i][i][i][i].someProp;
+    `,
+  ],
+  invalid: [
```
JoshuaKGoldberg, line 36: A few more test cases off the top of my head:

* `ReadonlyArray<...>`
* `readonly ...[]`
* `Readonly<...[]>`
* `Readonly<Array<...>>`

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,91 @@
+import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+import { unionTypeParts } from 'tsutils';
+import * as ts from 'typescript';
+
+import * as util from '../util';
```
StyleShit, line 6: should probably be switched to named imports

StyleShit (changes requested):

mahdi-farnia:
Sorry for late response. I am busy these days. I'll apply every requested changes as soon as I touch my computer. 🙂

JoshuaKGoldberg:
@mahdi-farnia since it's been a month I'm going to go ahead and close this PR to keep our queue small. Totally understand not having time for this - thanks for all the work you did to get it started!

If you end up having time to pick it back up, feel free to send a new PR (or re-open this one if it hasn't been auto-locked).

If anybody else ends up wanting to take on these changes in the interim, please do add a co-author attribution if you use any code from this PR.

Thanks!

PR #5881, opened 2022-10-25
feat(scope-manager): ignore ECMA version

### LINKED ISSUE #2900: [scope-manager] should ignore `parserOptions.ecmaVersion`
See discussion in https://github.com/yannickcr/eslint-plugin-react/issues/2882

There are only two cases in this project that we honour `parserOptions.ecmaVersion`, and both are in `scope-manager`:
https://github.com/typescript-eslint/typescript-eslint/blob/498f397ff3898dde631f37311615b555f38a414e/packages/scope-manager/src/ScopeManager.ts#L78-L84

These usages were brought across when I forked `eslint-scope`. 

However in the context of TS scope analysis - the `ecmaVersion` makes zero sense.
TS will never treat your code as if it's ES5 or lower - it is always (syntactically and semantically) "ESNext".

Whilst other parts of the ecosystem respects them - and they should continue to - our parser and scope analyser should (like TS itself) always treat code as if it is ESNext in terms of language features.

JoshuaKGoldberg:
BREAKING CHANGE:
Removes part of the public API for the parser and scope-manager.

## PR Checklist

- [x] Addresses an existing open issue: fixes #2900

## Overview

Per the issue, ignores `ecmaVersion` in the options, and assumes it's always ESNext.


packages/scope-manager/src/ScopeManager.ts
```diff
@@ -77,13 +76,6 @@ class ScopeManager {
   public isImpliedStrict(): boolean {
     return this.#options.impliedStrict === true;
   }
-  public isStrictModeSupported(): boolean {
-    return this.#options.ecmaVersion != null && this.#options.ecmaVersion >= 5;
-  }
-
-  public isES6(): boolean {
-    return this.#options.ecmaVersion != null && this.#options.ecmaVersion >= 6;
-  }
```
JoshuaKGoldberg, line 86: Is there some unified scope management interface we're expected to adhere to? I couldn't find anything on GitHub/Google.

bradzacher: Yeah it's this:
https://eslint.org/docs/latest/developer-guide/scope-manager-interface#scopemanager-interface

bradzacher (changes requested):
we can also remove the code from higher up the chain!

start from here and work your way down!
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/parser/src/parser.ts#L108

packages/scope-manager/src/ScopeManager.ts
JoshuaKGoldberg: Hmm yeah I couldn't tell if it was *necessary* to implement that or not. I'll just add these back in.

bradzacher: They're not necessary as part of the documented API but it's probably safer to leave them in yeah. I've found that there is a surprising number of packages that explicitly depend on the private internals of the scope manager.

packages/scope-manager/src/analyze.ts
```diff
@@ -122,7 +95,7 @@ function analyze(
     jsxFragmentName:
       providedOptions?.jsxFragmentName ?? DEFAULT_OPTIONS.jsxFragmentName,
     sourceType: providedOptions?.sourceType ?? DEFAULT_OPTIONS.sourceType,
-    lib: providedOptions?.lib ?? [mapEcmaVersion(ecmaVersion)],
```
bradzacher, line 125: Hmmm. I wonder if maybe I was too over-zealous with my recommendation to cull higher?

Having this auto-resolved is a really nice property. For the most part it shouldn't cause any issues for people (it just includes a few more global variables for people using rules like `no-undef` or `no-use-before-define`), but there might be edge cases?

Hmm. WDYT? I'm torn on if it's worth the code.

bradzacher (approved):

packages/scope-manager/src/analyze.ts
JoshuaKGoldberg: Neither of those are rules we recommend folks use - and folks should generally always be in >=ES2021 or so these days. I say let's kill them for now, and use the RC period to see if anybody notices! 🔪

JoshuaKGoldberg: Oh I will mark this as a breaking change now though.

JoshuaKGoldberg:
AGH the branch change to `v6` didn't go through. 🤦

PR #5885, opened 2022-10-25
fix(typescript-estree): wrap import = declaration in an export node

### LINKED ISSUE #4130: An exported `TSImportEqualsDeclaration` should be wrapped in an `ExportNamedDeclaration` instead of having `.isExport = true`
Code from #4129

```ts
namespace Bar {
  export const bar = 123;
}

export namespace Foo {
  export import TheBar = Bar;
}
```

I didn't know that `export import` was even valid at all in TS.
Right now this will emit a `TSImportEqualsDeclaration` with `.isExported = true`.
This is wrong because it means that rules have to have special handling specifically for this case in order to understand `TheBar` is exported.

We should breaking change fix the AST here so that the AST is instead `ExportNamedDeclaration > TSImportEqualsDeclaration` to align with everything else.

JoshuaKGoldberg:
BREAKING CHANGE:
Removes AST property and changes AST structure

## PR Checklist

- [x] Addresses an existing open issue: fixes #4130

## Overview

A note from our pairing: it's fine to create as many (multiple) nodes as we like from one original node.

Co-authored-by: Brad Zacher <brad.zacher@gmail.com>


packages/ast-spec/src/declaration/TSImportEqualsDeclaration/fixtures/entity-name-many/snapshots/1-TSESTree-AST.shot
```diff
@@ -17,7 +17,6 @@ Program {
         },
       },
       importKind: "value",
-      isExport: false,
```
JoshuaKGoldberg, line 20: I have more work to do on fixing up these snapshots. TODO.

packages/ast-spec/src/declaration/TSImportEqualsDeclaration/fixtures/entity-name-many/snapshots/5-AST-Alignment-AST.shot
```diff
@@ -2,5 +2,88 @@
...
+          name: 'F',
+
+          range: [7, 8],
+          loc: {
+            start: { column: 7, line: 1 },
+            end: { column: 8, line: 1 },
+          },
+        },
+        importKind: 'value',
++       isExport: false,
```
JoshuaKGoldberg, line 24: @bradzacher turns out I wasn't done with this - there are still a few `.shot` files with `isExport: false`. `yarn jest spec.ts` in `packages/typescript-estree` shows the failures.

But, on astexplorer.net, `@babel/parser` and other Babel parsers throw an error trying to parse code like `import F = A.B.C;`

Am I missing something about how these are generated?

JoshuaKGoldberg: I'll try the changes in https://github.com/typescript-eslint/typescript-eslint/issues/6065 instead.

packages/typescript-estree/tests/ast-alignment/fixtures-to-test.ts
```diff
@@ -371,6 +371,16 @@ tester.addFixturePatternConfig('typescript/basics', {
...
     'import-type-error',
+
+    /**
+     * Babel's 'typescript' transform gives these TypeScript-like properties
+     * such as `isExport: false`, but we don't include those in our AST.
+     */
+    'import-equal-declaration',
+    'import-export-equal-declaration',
+    'import-equal-type-declaration',
+    'import-export-equal-type-declaration',
```
JoshuaKGoldberg, line 382: @bradzacher I'm not 100% that this is accurate, but hey - it got the tests to pass! 🙃

bradzacher: this is one way to do it.
the other way would be do add a transformer to mutate the AST to match babels.

either way sucks.

this is why the ast-spec explicitly snapshots the differences so we don't need to maintain any hacks

bradzacher (approved):
futureisnow_oldman

Issue #5586, opened 2022-09-01
RFC: Use package.json exports to "hide" the dist folder for packages and control our exported surface-area


bradzacher:
Currently we don't use package.json exports, so anyone can import any file in any package by directly referencing a path within the `dist` folder - i.e.
```ts
import * as TSESLint from '@typescript-eslint/utils/dist/ts-eslint';
```

This is obviously a sub-par experience for consumers due to the deep path, and is a sub-par experience for us because any file technically forms part of our "external API".

In ESLint v8 they began using exports to lock down their package, and I think we should follow their lead to some degree.

I propose that we follow one of the following options:

### 1) export all of dist, without the dist

```json
{
  "exports": {
    "./*": "./dist/*"
  },
  "typesVersions": {
    "*": {
      "*": ["dist/*"]
    }
  }
}
```

This is the least restrictive thing we could do - it would simply mean consumers needn't include `dist/` in their import path - which is nice as it hides an implementation detail.

### 2) export all, but ban specific things

```json
{
  "exports": {
    "./*": "./dist/*"
    "./some-internal/file": null
  },
  "typesVersions": {
    "*": {
      "*": ["dist/*"]
    }
  }
}
```

This is slightly more restrictive, but still very permissive.
This would allow us to build APIs that are intended to be internal only and hide them, whilst still generally allowing all imports without the `dist/`.

### 3) export specific, allowed parts of our API

```jsonc
{
  "exports": {
    "./": "./dist/index",
    "./ts-eslint": "./dist/ts-eslint/index",
    "./ast-utils": "./dist/ast-utils/index",
    // ...
  },
  "typesVersions": {
    "*": {
      "./": "./dist/index",
      "./ts-eslint": "./dist/ts-eslint/index.d.ts",
      "./ast-utils": "./dist/ast-utils/index.d.ts",
      // ...
    }
  }
}
```

This is the most restrictive thing we could do as we would ban deep imports except those that we choose. It would also allow us to remove the `dist/` from imports.

JoshuaKGoldberg:
+1, once in a while TypeScript auto-imports something from `dist/` and it's annoying.

JoshuaKGoldberg:
+1, once in a while TypeScript auto-imports something from `dist/` and it's annoying.

[...]

PR #5886, opened 2022-10-26
chore: v6.0.0

JoshuaKGoldberg:
BREAKING CHANGE: v6, hooray! 🥳 

This PR contains all of the changes for v6.0.0.
We will ~rebase~ merge this onto `main`


package.json
```diff
@@ -46,7 +46,7 @@
     "typecheck": "nx run-many --target=typecheck --all --parallel"
   },
   "engines": {
-    "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
+    "node": "^14.17.0 || >=16.0.0"
```
bmish, line 49: Using minimum Node 14.18 would enable full support for `node:` protocol imports: https://2ality.com/2021/12/node-protocol-imports.html

```suggestion
    "node": "^14.18.0 || >=16.0.0"
```

bmish: I'd also recommend dropping support for Node 17 which is end-of-life already:

```suggestion
    "node": "^14.18.0 || ^16.0.0 || >=18.0.0"
```

bmish:
Is it time to drop ESLint v6 support from peer deps? It was released in 2019: https://eslint.org/blog/2019/06/eslint-v6.0.0-released/

```
"peerDependencies": {
  "eslint": "^6.0.0 || ^7.0.0 || ^8.0.0"
},
```

bmish:
What do you think of strictly-defining the Node API with `exports` in each package?

This would follow the lead of ESLint core and other ESLint plugins so that users cannot reach in and import/depend on arbitrary files inside the package.

* https://eslint.org/docs/latest/user-guide/migrating-to-8.0.0#remove-lib
   > Beginning in v8.0.0, ESLint is strictly defining its public API. Previously, you could reach into individual files such as require("eslint/lib/rules/semi") and this is no longer allowed. There are a limited number of existing APIs that are now available through the /use-at-your-own-risk entrypoint for backwards compatibility, but these APIs are not formally supported and may break or disappear at any point in time.

https://nodejs.org/api/packages.html#main-entry-point-export

bradzacher:
> What do you think of strictly-defining the Node API with exports in each package?

We're all for it - #5586

bradzacher:
> Is it time to drop ESLint v6 support from peer deps?

Sounds good to me! Last major we dropped v5 support - so happy to do it.
Filed #5968 to track

package.json
bradzacher: sounds good to me - I updated #5917 (which was left open).

bmish:
In case you guys aren't already tracking it, I see a lot of "TODO" comments throughout the codebase mentioning breaking changes that would be good to make now.

bmish:
I also see 5 deprecated rules that could be deleted.

JoshuaKGoldberg:
@bmish thanks for pointing these out! Anything that we've marked as deprecated (should be removed in general) deserves an issue tracking it. If you see anything that doesn't yet have one, feel free to file it!

I want to eventually do a sweep but haven't had the time. 🥲

package.json
MichaelDeBoey: Not sure if we want to update other values for now as I think it would be best to keep these in line with next major ESLint version 🤔

bmish: ESlint will most likely use the same Node engine requirement that I recommended in its next major version.

bradzacher:
I rebased this on master and it caused GH to seriously bork up all of the open PRs.
So I won't do that again and will just merge master in here and we can deal with the merge commits when we rebase merge this on top of master.

bradzacher:
From the netlify logs:
> error ts-api-utils@0.0.22: The engine "node" is incompatible with this module. Expected version ">=18". Got "16.19.1"

We'll need to bump the netlify CI version!

JoshuaKGoldberg:
Hmm, that's odd because https://github.com/JoshuaKGoldberg/ts-api-utils/blob/4f34bb59ff484b59d750a4156921974ca37b4c60/package.json#L85-L87:

```json
"engines": {
	"node": ">=16.13.0"
}
```

JoshuaKGoldberg:
Ahh, we can just bump `ts-api-utils`. https://github.com/JoshuaKGoldberg/ts-api-utils/pull/55

packages/typescript-estree/src/convert.ts
```diff
@@ -1058,23 +1058,6 @@ export class Converter {
...
-            'A property assignment cannot have a question token.',
-          );
-        }
-
-        if (exclamationToken) {
-          this.#throwError(
-            exclamationToken,
-            'A property assignment cannot have an exclamation token.',
-          );
-        }
```
JoshuaKGoldberg: FYI @fisker - TypeScript 5.0 removed a bunch of these deprecated properties. That's why I removed these lines. Let me know if there's a better way to handle the removal please!

fisker: The diff is so big, I can't see what's actually removed.

As I understand, `illegalDecorators` also removed in v5, in Prettier I made [this change](https://github.com/prettier/prettier/pull/13819/files#diff-2aa2b00967877a193c98032139d7685caef7777c295e3da65d1b5bc3e1bd668fL26) to make it work.

bradzacher: Funnily enough even though TS removed the properties from the types - TS doesn't actually parser error on the bad syntax - they're still semantic errors :frowning:
https://www.typescriptlang.org/play?ts=5.1.0-dev.20230313#code/C4TwDgpgBAKlC8UDeAoKUBmB7LAuKAjAoQNwoC+KAxlgHYDOwUAHvnIkpUA

So I've re-added the errors and patched TS types to add back the removed properties in 6586d98b944a5d5e37e7b469e7e044389ed919be

JoshuaKGoldberg:
Ran into some gnarly merge conflicts. Fixed most of them but ran out of time for the rest. Will finish soon.

PR #5893, opened 2022-10-26
fix(typescript-estree): don't allow single-run unless we're in type-aware linting mode

### LINKED ISSUE #5880: Bug: Reproducible `allowAutomaticSingleRunInference` failure

### Issue Description

I believe I've reproduced #5136.

I seem to have it narrowed down to the following _very specific_ combination:

- Using `typescript-eslint` together with `eslint-plugin-import`;
- Using `allowAutomaticSingleRunInference`;
- Importing from `"@mui/joy/styles"` - it seems this import is haunted somehow.

Even weirder - it doesn't always reproduce locally, but fails every time when running on GitHub Actions:

https://github.com/aaronadamsCA/typescript-eslint-issue-5136/actions/runs/3321426085/jobs/5489056308

This was not easy to isolate, but hopefully this reproduction can help resolve the issue for good.

For now we will simply disable `allowAutomaticSingleRunInference` in our repositories to avoid this issue.

https://github.com/aaronadamsCA/typescript-eslint-issue-5136

### Repro Steps

1) clone the repo
2) `yarn install`
3) `yarn run check`

### Versions

| `@typescript-eslint/eslint-plugin`     | `5.41.0` |
| `@typescript-eslint/parser`            | `5.41.0` |
| `ESLint`                               | `8.26.0` |
| `node`                                 | `16.17.0` |

Also note Gatsby has a transitive dependency on `@typescript-eslint/eslint-plugin@npm:4.33.0`, but I don't believe this is related.

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5880
    - fixes https://github.com/import-js/eslint-plugin-import/issues/2496

## Overview

This was an absolute doozy to debug!

`eslint-plugin-import` does its own out-of-band parsing for some of its lint rules (like `import/no-cycle` or `import/namespace`). It does this by reusing the config originally passed to ESLint for the currently linted file - namely the `parser` and `parserOptions` config items.

This parsing is always forced to be in non-type-aware mode thanks to some hard-coding I added to their tooling way way back (https://github.com/import-js/eslint-plugin-import/blob/c3d14cb920bdc6d277134973d37364db22c3a8b8/utils/parse.js#L79-L84). This was necessary for performance and because they often parse files not covered by a tsconfig (like `node_modules` files).

Back to the issue at hand. When we added the single-run optimisation, we had to add a fallback state for when you do `eslint --fix` - a state where we would fall out of the optimised path to handle the changed, fixed file contents with a dynamic program. The logic is simple - it just increments a counter if we're doing single-run linting, and if that counter is `> 1`, then we use the slow path:
https://github.com/typescript-eslint/typescript-eslint/blob/2ee81df5a365d82ef4b3dfc124d4ec39c7bcb725/packages/typescript-estree/src/parser.ts#L185-L203
This slow path uses a (slightly dodgy) "isolated" program for the file that's light-weight, slightly incorrect, but relatively fast to create (faster than rebuilding the programs for the tsconfig from the ground up!). Using a "dodgy" program is okay because the fixed file's contents won't ever influence the types of any other file in the workspace (by definition fixers must be safe and not break the build!).

So what is the bug?
Due to `eslint-plugin-import`'s out-of-band parsing it is possible for a file to be parsed multiple times with the second time being the eslint parse. This is an issue because unfortunately `inferSingleRun` didn't consider the case where we aren't doing type-aware linting for a file. So the bug would be:
1) lint file A
1) parse file A
    1) create a single-run program for the tsconfig
1) `eslint-plugin-import` sees A depends on B -> out-of-band parse B
    1) `parseAndGenerateServicesCalls[B] = 1`
1) lint file B
1) parse file B
    1) `parseAndGenerateServicesCalls[B] = 2`
    1) `parseSettings.singleRun === true && parseAndGenerateServicesCalls[B] > 1`, so create an isolated program for that file and use that instead of the true program for the file from (2,i)

The fix is simple - ensure we can never enter single run mode unless we're doing type-aware linting!

I don't know a good way to test this, but I confirmed that this fixes the problem by patching this change into the repro repos provided in both issues and it fixed the bug!

### DIFF
--- a/packages/typescript-estree/src/parseSettings/inferSingleRun.ts
+++ b/packages/typescript-estree/src/parseSettings/inferSingleRun.ts
@@ -15,6 +15,16 @@ import type { TSESTreeOptions } from '../parser-options';
  * @returns Whether this is part of a single run, rather than a long-running process.
  */
 export function inferSingleRun(options: TSESTreeOptions | undefined): boolean {
+  if (
+    // single-run implies type-aware linting - no projects means we can't be in single-run mode
+    options?.project == null ||
+    // programs passed via options means the user should be managing the programs, so we shouldn't
+    // be creating our own single-run programs accidentally
+    options?.programs != null
+  ) {
+    return false;
+  }
+
   // Allow users to explicitly inform us of their intent to perform a single run (or not) with TSESTREE_SINGLE_RUN
   if (process.env.TSESTREE_SINGLE_RUN === 'false') {
     return false;


ljharb:
Sometimes the tiny fixes that take the longest time to figure out :-) thanks for diving into this one!

packages/typescript-estree/src/parseSettings/inferSingleRun.ts
```diff
@@ -15,6 +15,16 @@ import type { TSESTreeOptions } from '../parser-options';
  * @returns Whether this is part of a single run, rather than a long-running process.
  */
 export function inferSingleRun(options: TSESTreeOptions | undefined): boolean {
+  if (
+    // single-run implies type-aware linting - no projects means we can't be in single-run mode
+    options?.project == null ||
```
JoshuaKGoldberg, line 20: This I'm confused about. I thought single-run linting _can_ have a lack of project if users don't provide a `project` in their ESLint config?

Looking through the stack:
1. `parseForESLint(code, options?)`: can be called with just `code`?
2. `parseAndGenerateServices(code, options)`: passes `parserOptions` to...
3. `createParseSettings(code, options)`: passes `options` to `inferSingleRun`

bradzacher:
@JoshuaKGoldberg putting it as a top-level comment cos there's some good context I want to be visible

---

In the past for all type-aware linting we created a watch host and a builder program. This api let use calculate types and efficiently respond to changes in an incremental fashion (I.e. TS can recalculate just the changes to the types instead of recreating the entire program from scratch).

In the "one and done" CLI run usecase you don't need to respond to changes because each file is linted exactly once and file contents are (assumed to be) constant.

The problem is that the incremental API has a lot of overhead involved even if there are no changes made.

Single run mode is a "fast path" which relies on the aforementioned assumption that a CLI run is constant - it uses the non-incremental APIs to calculate types. This means we can calculate and fetch types much faster just for CLI runs.

With all that being said - if we aren't doing type-aware linting then none of the above is necessary! We can essentially just create a `ts.SourceFile` to parse the text, convert the AST, then ditch the SourceFile. It's near-zero cost to do that, unlike creating a program.

Unfortunately this bug meant that if a user used type-aware linting, then the non-type-aware `eslint-plugin-import` parse would change the program we used to get types to a bad one, and thus cause lint rules to misbehave.

So the fix is to make sure that the single run code that makes type-aware linting fast only ever applies to type-aware linting like it should!

bradzacher:
> Where do the projects come from when someone runs ESLint in single-run mode without specifying one?

The user is parsing in type-aware mode (with projects) for the main linting, but `eslint-plugin-import` is parsing sans-projects (it does `delete parserOptions.project`).

The issue specifically is that if you parse a file sans-project, we shouldn't increment our counter for that file because the counter tracks how many times a file was parsed in single run mode (which we cannot be in without a project).

Right now the counter is incremented because the parser gets into a pseudo-single-run state and some of the single-run code paths fire because all they look for is the flag.

So this fix makes sure we can't get into that pseudo-state by forcing the flag off if we don't have a project.

If you grep the code for `.singleRun` you'll see the two places we look at the flag without looking for projects:
1) the counter bump - https://github.com/typescript-eslint/typescript-eslint/blob/2ee81df5a365d82ef4b3dfc124d4ec39c7bcb725/packages/typescript-estree/src/parser.ts#L193
2) the isolated program switch - https://github.com/typescript-eslint/typescript-eslint/blob/2ee81df5a365d82ef4b3dfc124d4ec39c7bcb725/packages/typescript-estree/src/parser.ts#L199

JoshuaKGoldberg (approved):
OK I took a deeper look at this locally and I think I understand. Thanks for the thorough explanations @bradzacher! 

The combination of your last message -especially the reference to the counter, how `singleRun` is used- and debugging through locally clears it up a bunch.

PR #5573, opened 2022-08-30
test: add regression testing against old TS and ESLint versions in the CI

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1752
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

#5571 shows that we really need to be testing against old TS versions.
This PR adds testing to the CI so that we can at least catch it.

PR #5896, opened 2022-10-27
feat(utils): add `RuleTester` API for top-level dependency constraints

bradzacher:
## Overview

As I continued working on #5573, I realised that there are some plugin rules that won't ever work on old versions of TS.
For example `consistent-type-imports` hard requires at least TS 4.1 as that was the first version where the required syntax was added.

This PR adds a new API to the rule tester that allows you to list a set of dependency constraints for all tests in the rule tester constructor. If the constraints don't pass then absolutely no tests are run.

### DIFF
--- a/packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
+++ b/packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
@@ -146,7 +146,7 @@ class RuleTester extends BaseRuleTester.RuleTester {
         };
         (this.staticThis.describe as DescribeWithSkip).skip(name, () => {
           this.staticThis.it(
-            'All test skipped due to unsatisfied constructor dependency constraints',
+            'All tests skipped due to unsatisfied constructor dependency constraints',
             () => {},
           );
         });
@@ -154,7 +154,7 @@ class RuleTester extends BaseRuleTester.RuleTester {
         // otherwise just declare an empty test
         this.staticThis.describe(name, () => {
           this.staticThis.it(
-            'All test skipped due to unsatisfied constructor dependency constraints',
+            'All tests skipped due to unsatisfied constructor dependency constraints',
             () => {
               // some frameworks error if there are no assertions
               assert.equal(true, true);
--- a/packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
+++ b/packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
@@ -50,6 +50,14 @@ interface RunTests<
 
 type AfterAll = (fn: () => void) => void;
 
+function isDescribeWithSkip(
+  value: unknown,
+): value is RuleTesterTestFrameworkFunction & {
+  skip: RuleTesterTestFrameworkFunction;
+} {
+  return 'skip' in describe && typeof describe.skip === 'function';
+}
+
 class RuleTester extends BaseRuleTester.RuleTester {
   readonly #baseOptions: RuleTesterConfig;
 
@@ -134,17 +142,10 @@ class RuleTester extends BaseRuleTester.RuleTester {
         this.#baseOptions.dependencyConstraints,
       )
     ) {
-      type DescribeWithMaybeSkip = RuleTesterTestFrameworkFunction & {
-        skip?: RuleTesterTestFrameworkFunction;
-      };
-      const describe: DescribeWithMaybeSkip = this.staticThis.describe;
-      if ('skip' in describe && typeof describe.skip === 'function') {
+      if (isDescribeWithSkip(this.staticThis.describe)) {
         // for frameworks like mocha or jest that have a "skip" version of their function
         // we can provide a nice skipped test!
-        type DescribeWithSkip = RuleTesterTestFrameworkFunction & {
-          skip: RuleTesterTestFrameworkFunction;
-        };
-        (this.staticThis.describe as DescribeWithSkip).skip(name, () => {
+        this.staticThis.describe.skip(name, () => {
           this.staticThis.it(
             'All tests skipped due to unsatisfied constructor dependency constraints',
             () => {},
--- a/packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
+++ b/packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
@@ -55,7 +55,12 @@ function isDescribeWithSkip(
 ): value is RuleTesterTestFrameworkFunction & {
   skip: RuleTesterTestFrameworkFunction;
 } {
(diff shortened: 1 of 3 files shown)


packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
```diff
@@ -1,28 +1,42 @@
...
 const TS_ESLINT_PARSER = '@typescript-eslint/parser';
 const ERROR_MESSAGE = `Do not set the parser at the test level unless you want to use a parser other than ${TS_ESLINT_PARSER}`;
 
-type RuleTesterConfig = Omit<TSESLint.RuleTesterConfig, 'parser'> & {
+type RuleTesterConfig = Omit<BaseRuleTester.RuleTesterConfig, 'parser'> & {
   parser: typeof TS_ESLINT_PARSER;
+  /**
+   * Constraints that must pass in the current environment for any tests to run
+   */
+  dependencyConstraints?: DependencyConstraint;
```
JoshuaKGoldberg, line 23: We're now repeating this comment & property three times. How about extracting it into an interface named something like `WithDependencyConstraints`?

packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
```diff
@@ -107,9 +125,48 @@ class RuleTester extends TSESLint.RuleTester {
...
+      const describe: DescribeWithMaybeSkip = this.staticThis.describe;
+      if ('skip' in describe && typeof describe.skip === 'function') {
+        // for frameworks like mocha or jest that have a "skip" version of their function
+        // we can provide a nice skipped test!
+        type DescribeWithSkip = RuleTesterTestFrameworkFunction & {
+          skip: RuleTesterTestFrameworkFunction;
+        };
+        (this.staticThis.describe as DescribeWithSkip).skip(name, () => {
+          this.staticThis.it(
+            'All test skipped due to unsatisfied constructor dependency constraints',
```
JoshuaKGoldberg, line 149:
```suggestion
            'All tests skipped due to unsatisfied constructor dependency constraints',
```

packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
```diff
@@ -107,9 +125,48 @@ class RuleTester extends TSESLint.RuleTester {
...
+            () => {},
+          );
+        });
+      } else {
+        // otherwise just declare an empty test
+        this.staticThis.describe(name, () => {
+          this.staticThis.it(
+            'All test skipped due to unsatisfied constructor dependency constraints',
```
JoshuaKGoldberg, line 157:
```suggestion
            'All tests skipped due to unsatisfied constructor dependency constraints',
```

packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
```diff
@@ -107,9 +125,48 @@ class RuleTester extends TSESLint.RuleTester {
...
+      this.#baseOptions.dependencyConstraints &&
+      !satisfiesAllDependencyConstraints(
+        this.#baseOptions.dependencyConstraints,
+      )
+    ) {
+      type DescribeWithMaybeSkip = RuleTesterTestFrameworkFunction & {
+        skip?: RuleTesterTestFrameworkFunction;
+      };
+      const describe: DescribeWithMaybeSkip = this.staticThis.describe;
+      if ('skip' in describe && typeof describe.skip === 'function') {
```
JoshuaKGoldberg, line 141: Non-blocking suggestion: maybe extract this into a type predicate to make it more clear? Then I suspect you can remove the `as DescribeWithSkip` later on.

packages/utils/src/ts-eslint/RuleTester.ts
```diff
@@ -169,23 +169,26 @@ declare class RuleTesterBase {
    * @param text a string describing the rule
```
JoshuaKGoldberg, line 169: Existing code, not a blocker: this JSDoc was already out of date. I have a pending issue on my own personal template repo for finding a good set of lint rules that make sure JSDocs/TSDocs don't get out of sync: https://github.com/JoshuaKGoldberg/template-typescript-node-package/issues/17#issuecomment-1286232744. I can file an issue here to add it in once I do.

JoshuaKGoldberg (approved):
Awesome, looks good to me! Just a few non-blocking suggestions, along with an error message typo fix.

packages/utils/src/ts-eslint/RuleTester.ts
bradzacher: *technically* the jsdoc is correct, it just applies to the `RuleTesterTestFrameworkFunction` type, not the property itself.

so should defs move the jsdoc onto the type itself

packages/utils/tests/eslint-utils/rule-tester/RuleTester.test.ts
```diff
@@ -715,5 +715,69 @@ describe('RuleTester', () => {
...
+              code: 'valid',
+              errors: [],
+            },
+          ],
+          valid: [
+            {
+              code: 'valid',
+            },
+          ],
+        });
```
bradzacher, line 779: oops - forgot the assertion

packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
bradzacher: This comment is slightly different in that it uses `any tests` instead of `the test`.

I did want to avoid *another* interface in this file cos there's a lot of stuff here.
but happy to do it if you feel strongly!

packages/utils/tests/eslint-utils/rule-tester/RuleTester.test.ts
```diff
@@ -715,5 +715,69 @@ describe('RuleTester', () => {
...
+            "my-rule",
+            [Function],
+          ]
+        `);
+
+        // trigger the describe block
+        mockedDescribe.mock.lastCall?.[1]();
+        expect(mockedIt.mock.lastCall).toMatchInlineSnapshot(`
+          [
+            "All test skipped due to unsatisfied constructor dependency constraints",
```
bradzacher, line 753:
```suggestion
            "All tests skipped due to unsatisfied constructor dependency constraints",
```

packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
JoshuaKGoldberg: No strong feelings 😄

JoshuaKGoldberg (approved):

PR #5904, opened 2022-10-30
docs(eslint-plugin): [consistent-type-imports] make a note about `parserOptions.emitDecoratorMetadata`

### LINKED ISSUE #5903: Bug: [consistent-type-imports] False positive when using DI and decorators (tsyringe)

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.4&sourceType=module&code=GYVwdgxgLglg9mABDMArAptAhgIwDboA8AKgHwAUAlAFyLlRYBOA5ulLcZYgLymIBucGABNEAbwBQiRIzYhGSUJFgI6DFmw40BQ0ZOnSpiAL4BuCcYkSA9NcQABKAGcAtDGZg4siTAC2ABy8ocUQAISYTRGBGOF9EAHIAOmscJnjzCXsUDGx8dCoJCDwsJydEADE4OHEjCAQnKEYQaC9yf0YYfiwodERUxlpwxi4xS2MgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6MgeyeUuX0Ra0ipWpQC2xDtHyo0kRNGhTI4MAF8QaoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3QAacDUjp4yPABF+sTAEM8KgLLo8qgCbrVFMKIlTa6AB6IsyeOiK7USuGo2ZhJzGKkBfEL5AA

### Repro Code

```typescript
function injectable<T>(): (target: T) => void {
  return function (target: T): void {
    
  };
}

import { Bar } from './bar'; // Error: All imports in the declaration are only used as types.

@injectable()
class Foo {
  constructor(private bar: Bar) {}
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/consistent-type-imports": "error"
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true,
    "emitDecoratorMetadata": true,
  }
}
```

### Expected Result

I expected that #2751 (#2559) would've fixed this and the rule wouldn't have assumed that it is only used as a type, but turns out it only catches decorators from within the constructor:

```ts
import { Bar } from './bar'; // ok

class Foo {
  constructor(@deco(() => Bar) private bar: Bar) {}
}
```

```ts
import { Bar } from './bar'; // Expected ok, but got Error: All imports in the declaration are only used as types.

@injectable()
class Foo {
  constructor(private bar: Bar) {}
}
```

### Actual Result

The rule assumes `Bar` is used as a type, causing the DI library (in the case [tsyringe](https://github.com/Microsoft/tsyringe)) to be unable to inject the dependencies (`Bar`) at runtime. This is pretty similar to the issue with Nest.js (#2559)

### Additional Info

_No response_

### Versions

| `@typescript-eslint/eslint-plugin`     | `5.41.0` |
| `@typescript-eslint/parser`            | `playground default` |
| `node`                                 | `playground default` |

bradzacher:
## PR Checklist

- [X] Addresses an existing open issue: fixes #5903

## Overview

Makes a note about the required config in the rule's docs so that it's clearer for users.

### DIFF
--- a/packages/eslint-plugin/docs/rules/consistent-type-imports.md
+++ b/packages/eslint-plugin/docs/rules/consistent-type-imports.md
@@ -52,7 +52,7 @@ const x: import('Bar') = 1;
 
 The `emitDecoratorMetadata` compiler option changes the code the TypeScript emits. In short - it causes TypeScript to create references to value imports when they are used in a type-only location. If you are using `emitDecoratorMetadata` then our tooling will require additional information in order for the rule to work correctly.
 
-If you are using [type-aware linting](../../../../docs/linting/TYPED_LINTING.md), then you just need to ensure that the `tsconfig.json` you've configured for `parserOptions.project` has `emitDecoratorMetadata` turned on. Otherwise you can explicitly tell our tooling to analyse your code as if the compiler option was turned on [by setting `parserOptions.emitDecoratorMetadata` to `true`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/parser/README.md#parseroptionsemitdecoratormetadata).
+If you are using [type-aware linting](../../../../docs/linting/TYPED_LINTING), then you just need to ensure that the `tsconfig.json` you've configured for `parserOptions.project` has `emitDecoratorMetadata` turned on. Otherwise you can explicitly tell our tooling to analyze your code as if the compiler option was turned on [by setting `parserOptions.emitDecoratorMetadata` to `true`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/parser/README.md#parseroptionsemitdecoratormetadata).
 
 ## When Not To Use It
 
--- a/packages/eslint-plugin/docs/rules/consistent-type-imports.md
+++ b/packages/eslint-plugin/docs/rules/consistent-type-imports.md
@@ -52,7 +52,7 @@ const x: import('Bar') = 1;
 
 The `emitDecoratorMetadata` compiler option changes the code the TypeScript emits. In short - it causes TypeScript to create references to value imports when they are used in a type-only location. If you are using `emitDecoratorMetadata` then our tooling will require additional information in order for the rule to work correctly.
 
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (dismissed):
💯

packages/eslint-plugin/docs/rules/consistent-type-imports.md
```diff
@@ -48,6 +48,12 @@ type T = import('Foo').Foo;
 const x: import('Bar') = 1;
 ```
 
+## Usage with `emitDecoratorMetadata`
+
+The `emitDecoratorMetadata` compiler option changes the code the TypeScript emits. In short - it causes TypeScript to create references to value imports when they are used in a type-only location. If you are using `emitDecoratorMetadata` then our tooling will require additional information in order for the rule to work correctly.
+
+If you are using [type-aware linting](./docs/linting/TYPED_LINTING.md), then you just need to ensure that the `tsconfig.json` you've configured for `parserOptions.project` has `emitDecoratorMetadata` turned on. Otherwise you can explicitly tell our tooling to analyze your code as if the compiler option was turned on [by setting `parserOptions.emitDecoratorMetadata` to `true`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/parser/README.md#parseroptionsemitdecoratormetadata).
```
bradzacher, line 55: @JoshuaKGoldberg / @Josh-Cena do you knows know how I can do this link to the type-aware linting docs?

I tried `../../../../docs/linting/TYPED_LINTING.md`, but docusaurus failed that build (https://app.netlify.com/sites/typescript-eslint/deploys/635e718d23d0fd00090b2e67#L152)

Am I able to do this link or do I need to hard-link it to the full URL?

JoshuaKGoldberg: I don't know - I would personally use the hard-link full URL and then ask for help.

Josh-Cena: The two Markdown files are rendered by different plugins, so you can't link them with file paths. Josh is right—you need to use a URL link here.

PR #5906, opened 2022-10-31
fix(ast-spec): add TSQualifiedName to TypeNode union

bradzacher:
## Overview

Found this was missing whilst doing some work - oops!


JoshuaKGoldberg (approved):
🚀

PR #5594, opened 2022-09-03
feat(eslint-plugin): [prefer-optional-chain] support suggesting `!foo || !foo.bar` as a valid match for the rule

omril1:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #5245
-   [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
-   [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

* Ignores the issue when `TSNonNullExpression` is mixed in the properties (was reverted in #5564 / #5553)

PR #5919, opened 2022-11-02
fix(eslint-plugin): [prefer-optional-chain] fixer produces wrong logic

### LINKED ISSUE #1438: [prefer-optional-chain] fixer produces wrong logic
**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/prefer-optional-chain": ["error"]
  }
}
```

```TS
declare const a: { b: string | null } | null;
a !== null && a.b !== null && a.b.length > 0;
/*
^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer using an optional chain expression instead, as it's more concise and easier to read.
*/
```

**Expected Result**

```TS
a?.b != null && a.b.length > 0;
```

**Actual Result**

```TS
a?.b !== null && a.b.length > 0;
/*
                 ^ Object is possibly 'null'.
*/
```

**Additional Info**

**Versions**

| `@typescript-eslint/eslint-plugin` | `2.15.0` |
| `@typescript-eslint/parser`        | `2.15.0` |
| `TypeScript`                       | `3.7.3` |
| `ESLint`                           | `6.1.0` |
| `node`                             | `10.16.0` |
| `npm`                              | `6.9.0` |

sviat9440:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1438


packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -221,10 +221,19 @@ export default util.createRule({
 
         if (expressionCount > 1) {
           if (previous.right.type === AST_NODE_TYPES.BinaryExpression) {
+            let operator = previous.right.operator;
+            if (
+              previous.right.operator === '!==' &&
+              previous.right.right.type === AST_NODE_TYPES.Literal &&
+              previous.right.right.raw === 'null'
```
JoshuaKGoldberg, line 228: Just marking this todo for later...

```suggestion
              // TODO(#4820): Use the type checker to know whether this is `null`
              previous.right.right.type === AST_NODE_TYPES.Literal &&
              previous.right.right.raw === 'null'
```

JoshuaKGoldberg (approved):
A nice targeted change - thanks @sviat9440! 🔥

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
```diff
@@ -133,6 +133,22 @@ const baseCases = [
...
+    output: 'foo?.bar != null',
+  },
+  {
+    code: 'foo != null && foo.bar !== null',
+    output: 'foo?.bar != null',
+  },
+  {
+    code: 'foo !== null && foo.bar != null',
+    output: 'foo?.bar != null',
+  },
```
JoshuaKGoldberg, line 151: 🤔 now that #5594 has been merged into this branch, these cases are no longer reporting any errors. What do you see as the right behavior @omril1 @sviat9440?

omril1: Let me think about it.
My first thought was that we could filter them out, but it might be better to split the base cases into groups

sviat9440: Added selectors

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
```diff
@@ -11,149 +11,227 @@ const ruleTester = new RuleTester({
...
-    code: 'foo.bar && foo.bar.baz && foo.bar.baz && foo.bar.baz.buzz',
-    output: 'foo.bar?.baz?.buzz',
-  },
-  // chained members with element access
-  {
-    code: 'foo && foo[bar] && foo[bar].baz && foo[bar].baz.buzz',
-    output: 'foo?.[bar]?.baz?.buzz',
-  },
-  {
+// eslint-disable-next-line @typescript-eslint/no-namespace
```
sviat9440, line 14: @JoshuaKGoldberg I wanted to wrap the cases nicely in namespace, but the rule `@typescript-eslint/no-namespace` does not allow this. I tried to change to `module', but another rule `@typescript-eslint/prefer-namespace-keyword` throws an error.

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
omril1: IMO something like this can be more readable:

```typescript
const baseCases = [/* ... */];
const nonStrictEqualCases = [/* ... */];
const negatedOrCases = baseCases.map(c => ({
  ...c,
  code: c.code.replace(/foo/g, '!foo').replace(/&&/g, '||'),
  errors: [
    {
      ...c.errors[0],
      suggestions: [
        {
          ...c.errors[0].suggestions![0],
          output: `!${c.errors[0].suggestions![0].output}`,
        },
      ],
    },
  ],
}));
const allBasesCases = [...baseCases, ...nonStrictEqualCases];

/* 
  ...later on
	invalid: [
		...allBasesCases.map(...),
		...allBasesCases.map(...),
		...negatedOrCases,
  ]
  */
```

With variable names subject to change

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
bradzacher: I would just drop the namespace - I don't think it adds much value.
if you want to isolate things, then I would move the test into a subfolder and isolate them in a separate module entirely. Eg

- `tests/rules/prefer-optional-chain/base-cases.ts`
- `tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts`

sviat9440: fixed

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
sviat9440: I think that in the future additional conditions may appear, and this will increase the number of variables by 2^n.

bradzacher (approved):
it's worth mentioning that a lot of people don't like inexact equal for `null`. It's a stylistic thing 🤷 

Given this is a suggestion fixer - it doesn't matter that we're fixing to `!=` because the fixer is entirely optional and user-driven.

But it's definitely a potential future piece of work to add a second suggestion fix to use `!== null && !== undefined` instead.
OFC it's also a lot harder to do that fix because you need to pay attention to other logic and precedence

bradzacher:
oops - if you can resolve the conflicts, then we can look at merging this in.

PR #5965, opened 2022-11-10
feat(eslint-plugin): [member-ordering] add a required option for required vs. optional member ordering

### LINKED ISSUE #5596: Enhancement: [member-ordering] Port requiredFirst from eslint-plugin-typescript-sort-keys

https://typescript-eslint.io/rules/member-ordering/

### Description

[eslint-plugin-typescript-sort-keys](https://github.com/infctr/eslint-plugin-typescript-sort-keys) supports a `requiredFirst` option that enforces required members come before optional. Can we add that to `@typescript-eslint/member-ordering`, to bring it up to parity?

Porting discussion from https://github.com/infctr/eslint-plugin-typescript-sort-keys/issues/50#issuecomment-1108934841

### Fail

```typescript
interface U {
  d: T
  c?: T
  b?: T
  a: T
}
```

### Pass

```typescript
interface U {
  a: T
  d: T
  b?: T
  c?: T
}
```

### Additional Info

Examples taken from https://github.com/infctr/eslint-plugin-typescript-sort-keys/blob/master/docs/rules/interface.md.

asdf93074:
## PR Checklist

- [X] Addresses an existing open issue: fixes #5596

## Overview

Added a new property required to SortedOrderConfig.
Added a corresponding error message id for when a rule fails because of required being set to 'first' or 'last'.

If required is 'first', then before any sorting/grouping configurations, we ensure that the members array has all its required items first and all optional items afterwards. Afterwards, we perform the remaining sorting/grouping checks on the required subset and the optional subset.


packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -46,6 +49,7 @@ type Order = AlphabeticalOrder | 'as-written';
 interface SortedOrderConfig {
   memberTypes?: MemberType[] | 'never';
   order: Order;
+  requiredFirst?: boolean;
```
JoshuaKGoldberg, line 52: Sorry, I realize this is adding on more complexity after we already marked the issue as accepting PRs, but... what if people want to add required _last_? What do you think about changing this to `required?: 'first' | 'last'`?

asdf93074: Yeah, I think its a good idea. I don't mind tacking it on in this PR.

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -707,34 +820,111 @@ export default util.createRule<Options, MessageIds>({
...
+        const lastOptionalMemberIndex = findLastIndex<Member>(
+          members,
+          isMemberOptional,
+        );
+
+        if (firstRequiredMemberIndex != -1) {
+          const requiredMembers: Member[] = members.slice(
+            firstRequiredMemberIndex,
+          );
+          memberSets.push(requiredMembers);
```
JoshuaKGoldberg, line 911: [Nitpick] Why bother making the variable if it's only used once, on a next line?

```suggestion
          memberSets.push(members.slice(firstRequiredMemberIndex));
```

Here and with the other three times this is done.

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -707,34 +820,111 @@ export default util.createRule<Options, MessageIds>({
...
 
-      const hasAlphaSort = !!(order && order !== 'as-written');
+      if (!checkRequiredOrder(members, required)) {
+        return;
+      }
 
-      // Check order
-      if (Array.isArray(memberTypes)) {
-        const grouped = checkGroupSort(members, memberTypes, supportsModifiers);
+      if (required === 'first') {
```
JoshuaKGoldberg, line 869: The purpose of each of these two big blocks (`required === 'first'`, `required === 'last'`) is to separate into 1-2 groups of optional vs. required members, right? If so, these can be simplified to just get the first index where two members don't have the same optionality:

```tsx
const switchIndex = members.findIndex(
  (member, index) =>
    index && isMemberOptional(member) !== isMemberOptional(members[index - 1])
);
```

_(I don't love `switchIndex` as a name, but haven't spent the time to think of a better one :smile:)_

I think you can simplify a lot of code by using this technique:
* Checking whether `switchIndex === -1` means the code can avoid creating a `memberSets` variable or intermediate array variables. It can instead directly call `checkOrder` on array slices.
* If `switchIndex > 0` and `< members.length`, you can compare the member immediately after it to the member immediately before. That effectively replaces `checkRequiredOrder`'s logic.

JoshuaKGoldberg, line 869: The purpose of each of these two big blocks (`required === 'first'`, `required === 'last'`) is to separate into 1-2 groups of optional vs. required members, right? If so, these can be simplified to just get the first index where two members don't have the same optionality:

```tsx
const switchIndex = members.findIndex(
  (member, index) =>
    index && isMemberOptional(member) !== isMemberOptional(members[index - 1])
);
```

_(I don't love `switchIndex` as a name, but haven't spent the time to think of a better one :smile:)_

I think you can simplify a lot of code by using this technique:
* Checking whether `switchIndex === -1` means the code can avoid creating a `memberSets` variable or intermediate array variables. It can instead directly call `checkOrder` on array slices.
* If `switchIndex > 0` and `< members.length`, you can compare the member immediately after it to the member immediately before. That effectively replaces `checkRequiredOrder`'s logic.

JoshuaKGoldberg, line 869: The purpose of each of these two big blocks (`required === 'first'`, `required === 'last'`) is to separate into 1-2 groups of optional vs. required members, right? If so, these can be simplified to just get the first index where two members don't have the same optionality:

```tsx
const switchIndex = members.findIndex(
  (member, index) =>
    index && isMemberOptional(member) !== isMemberOptional(members[index - 1])
);
```

_(I don't love `switchIndex` as a name, but haven't spent the time to think of a better one :smile:)_

I think you can simplify a lot of code by using this technique:
* Checking whether `switchIndex === -1` means the code can avoid creating a `memberSets` variable or intermediate array variables. It can instead directly call `checkOrder` on array slices.
* If `switchIndex > 0` and `< members.length`, you can compare the member immediately after it to the member immediately before. That effectively replaces `checkRequiredOrder`'s logic.

JoshuaKGoldberg (changes requested):
Thanks for sending this in @asdf93074! I'm excited to get this feature in. 🎉 

There's a lot of changed code in this PR and the `member-ordering` rule is already a lot of lines. We should be able to trim the changes down a bit to be easier to read & review. I can take a deeper look once the refactors are done. Let me know if anything's unclear or not right!

packages/eslint-plugin/src/rules/member-ordering.ts
asdf93074: Right, I see that this fits for when we want to know whether we should split the array and also where we should split it from.

> If switchIndex > 0 and < members.length, you can compare the member immediately after it to the member immediately before. That effectively replaces checkRequiredOrder's logic.

I probably don't get this one. Could you explain it a bit more? Did you mean to say that we should start from `switchIndex` and check every index after it to ensure that it has the same optionality as `member[switchIndex`] itself?

JoshuaKGoldberg: Exactly that, yes!

packages/eslint-plugin/src/rules/member-ordering.ts
asdf93074: I wanted verbosity over shorter code hence, the explicit variable creation.

packages/eslint-plugin/src/rules/member-ordering.ts
asdf93074: I made the changes. It really simplifies the splitting part of the code because `switchIndex` represents this information so well. However, for the `checkRequiredOrder` part, I feel that the previous approach was better simply because it was easier to understand at a glance and also easier to reason about for me. We also needed a check here to figure out if everything before the `switchIndex` was in the correct position too (all optional if required is last and vice-versa for required being first).

Also, switchIndex as a name seems appropriate to me. We could maybe make it more verbose e.g. optionalitySwitchIndex but within context, it still seems fine given the small body of the function itself.

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -376,6 +384,26 @@ function getMemberName(
...
+  switch (node.type) {
+    case AST_NODE_TYPES.TSPropertySignature:
+    case AST_NODE_TYPES.TSMethodSignature:
+    case AST_NODE_TYPES.TSAbstractPropertyDefinition:
+    case AST_NODE_TYPES.PropertyDefinition:
+    case AST_NODE_TYPES.TSAbstractMethodDefinition:
+    case AST_NODE_TYPES.MethodDefinition:
+      return !!node.optional;
+  }
+  return false;
```
JoshuaKGoldberg, line 404: Indeed, codecov is right: there are some node types not tested...

```ts
type:
  | TSESTree.AST_NODE_TYPES.StaticBlock
  | TSESTree.AST_NODE_TYPES.TSCallSignatureDeclaration
  | TSESTree.AST_NODE_TYPES.TSConstructSignatureDeclaration
  | TSESTree.AST_NODE_TYPES.TSIndexSignature
```

Could you add tests that include members of these types please?

JoshuaKGoldberg (changes requested):
Thanks for making the changes @asdf93074! I actually personally find this a good deal easier to read, but maybe that's just my personal preference. If another maintainer wants to weigh in that'd be good too 😄 - I'm nervous about stomping over perfectly good code. (and to be fair, yours did work quite well before too!)

Just requesting changes on a bit more testing. Otherwise this looks good to me!

packages/eslint-plugin/src/rules/member-ordering.ts
asdf93074: Going off the assumption that we want all of these types to be considered as required too, how should the error message be changed for these cases? For example:
```js
      code: `
class X {
  a?: string;
  static {}
  b?: string;
}
      `,
      options: [
        {
          default: {
            memberTypes: 'never',
            order: 'as-written',
            required: 'first',
          },
        },
      ],
```
The error thrown here is `Member a should be declared after all required members.` because it considers the static block to be required. Would a user here be aware of this assumption as well?

JoshuaKGoldberg: Hmm, that's a great point. My gut says _yes_ they should be -- and if they're not, then the rule's error existing would inform them.

The rule can already be configured to e.g. [put static fields before instance fields](https://typescript-eslint.io/rules/member-ordering/#static-fields-before-instance-fields). So I feel comfortable allowing this option in as-is.

What do you think?

asdf93074: Yep, sounds good to me. We might also want the other messages to be a bit more explicit now since they can be ran within the context of required/optional members.

JoshuaKGoldberg (approved):
Fantastic, thanks for sending this in and iterating on it with me @asdf93074! 🙌

JoshuaKGoldberg:
I'll leave this open for a bit in case another maintainer wants to poke at it or suggest an alternative API name (which we can apply ourselves if you don't want to keep receiving extra work because of our indecision 😅). But I've set a reminder to merge it this weekend if nobody requests changes, so it'll be available in our next stable version.

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -99,6 +103,10 @@ const objectConfig = (memberTypes: MemberType[]): JSONSchema.JSONSchema4 => ({
         'natural-case-insensitive',
       ],
     },
+    required: {
```
bradzacher, line 106: my preference would be to invert the name (and the behaviour too I guess?).
in my experience you think about members as being a "member" or an "optional member", not a "required member" or an "optional member".

🤷 not going to block on it, just adding my 2c

JoshuaKGoldberg: How about `optionality?: 'optional-first' | 'required-first'`? Just spitballing...

JoshuaKGoldberg: @asdf93074 @bradzacher I'm leaning a bit harder towards `optionality`. Let me know if you'd prefer something else - otherwise I'll work on Monday morning before the release.

asdf93074: `firstMembers?: 'required' | 'optional'`

In seriousness though, I'm okay with the above suggestion. Although, `optionality` and `required` do not communicate the full purpose of this configuration (**order** of members based on **optionality**) alone. Maybe `optionalityOrder` or `memberOptionalityOrder` would complete it?

JoshuaKGoldberg: I like `optionalityOrder` a lot!

JoshuaKGoldberg: +1 to your reasoning, I like `optionalityOrder` a lot! Will go with it now.

I also forgot to check that `member-ordering.md` be updated to reflect the new option - so I'll do that now too. Thanks!

JoshuaKGoldberg (approved):
🙌

packages/eslint-plugin/src/rules/member-ordering.ts
asdf93074: > I also forgot to check that member-ordering.md be updated to reflect the new option - so I'll do that now too.

Ah, I was waiting for the core stuff to be approved before I added that in, but thank you for doing it.

PR #5971, opened 2022-11-11
chore: drop support for node v14.17, v17

### LINKED ISSUE #5917: Drop support for NodeJS v12
### Suggestion

https://github.com/nodejs/Release/blob/main/schedule.json#L66

NodeJS v12 was EOL 2022-04-30.
We also stopped testing on Node v12 in #5512 last month - so we don't even have signal if we have support!

bmish:
BREAKING CHANGE:
Drops support for end-of-life node v14.17 and v17

Fixes #5917.

As per: https://github.com/typescript-eslint/typescript-eslint/issues/5917#issuecomment-1311082959

Using minimum Node 14.18 enables full support for `node:` protocol imports: https://2ality.com/2021/12/node-protocol-imports.html

Node 17 is [end-of-life](https://github.com/nodejs/Release) already.


JoshuaKGoldberg (approved):
💯, thanks!

Love to see those versions ticking up 😄

PR #5972, opened 2022-11-11
feat: drop support for ESLint v6

### LINKED ISSUE #5968: Drop support for ESLint v6
### Suggestion

Just like we did in our v5 release (https://github.com/typescript-eslint/typescript-eslint/issues/2465).

Most users should have have migrated off of v6 by this point, and I think a year back is far enough for us to support.
Supporting old versions restricts which APIs we can use, and makes it harder to build extension rules as we have to take care to fill in missing messageIds ourselves.

ESLint v8.0.0 was released 10 October 2021 - over 1 year ago.
ESLint v7.0.0 was released 09 May 2020 - over 2 years ago.
ESLint v6.0.0 was released 21 June 2019 - over 3 years ago.

bmish:
BREAKING CHANGE:
Drops support for ESLint v6.

Fixes #5968.

Breaking change targeting v6 (#5886).

Also changed the ESLint v6 integration test to test ESLint v7 instead.


packages/type-utils/package.json
```diff
@@ -49,7 +49,7 @@
     "typescript": "*"
   },
   "peerDependencies": {
-    "eslint": "*"
```
bmish, line 52: Seems like we might as well include the constraint here too.

JoshuaKGoldberg (approved):
Looks great, thanks!

PR #5981, opened 2022-11-13
feat: raise tsconfig target to ES2021

### LINKED ISSUE #5980: Repo: Raise tsconfig target
### Suggestion

Consider whether we should raise tsconfig [target](https://www.typescriptlang.org/tsconfig#target) / [lib](https://www.typescriptlang.org/tsconfig#lib) from ES2017 to something higher. 

https://github.com/typescript-eslint/typescript-eslint/blob/91d71bcf6b2f822116d25b281e289fe1d5005401/tsconfig.base.json#L16-L17

We're already planning to increase requirements for v6 including:

* https://github.com/typescript-eslint/typescript-eslint/pull/5972 (released 2019)
* https://github.com/typescript-eslint/typescript-eslint/issues/5917 (released 2019, EOL 2022)
* https://github.com/typescript-eslint/typescript-eslint/pull/5915 (released Feb 2021)

This all seems to indicate people would have a relatively modern environment to run TypeScript linting in.

The newest possible we could use is `ES2021`, is that too new or is there any reason we couldn't use that?

This may be a breaking change that could be included with v6 (https://github.com/typescript-eslint/typescript-eslint/pull/5886).

bmish:
BREAKING CHANGE:
Updates the internal TSConfig target to ES2021.

Fixes #5980.

I didn't see any other production files to update this in. All the other `tsconfig.json` files present are test fixtures. Let me know if we should update those too.

Potential breaking change targeting v6 (#5886).

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -13,8 +13,8 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es2017",
-    "lib": ["es2017"],
+    "target": "ES2021",
+    "lib": ["ES2021"],
     "paths": {}
   }
 }


JoshuaKGoldberg (approved):
Nice, glad this was all that was needed. Thanks for sending this in!

PR #5983, opened 2022-11-13
fix(eslint-plugin): [class-literal-property-style] correct handling of override cases

### LINKED ISSUE #5961: Bug: [class-literal-property-style] A literal getter with a setter does not pass lint, and incorrectly auto-fix

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.4&sourceType=module&code=MYGwhgzhAEAqCmEAu0DeBYAUNaF4oDcwQBXeACiJAC5ckAnASwDsBzASjSxxwHpfoAOmHdoAXyyjW+aFTLl2tZEzZdsPevhL1m0AORJESPQG5REzGKA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6M+AQ2WVsf0TRO8WsWgB7UtCK1kRJOiiCJ0SODABfEBqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3QAacDUjoAHoizJ46IngCGqACL9YmNXl3Cwo9FIC+IM0A

### Repro Code

```typescript
class Test {
  set value(val: string) {
    // ...
  }

  get value(): string {
    return 'test';
  }
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/class-literal-property-style": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true,
  }
}
```

### Expected Result

Should be no errors because the setter exists

### Actual Result

Error: `Literals should be exposed using readonly fields.`

Auto-fix: 

```
class Test {
  set value(val: string) {
    // ...
  }

  readonly value = 'test';
}
```

Then TypeScript error:
``` Duplicate identifier 'value'```

### Additional Info

_No response_

### Versions

| package | version |
| -- | -- |
| `@typescript-eslint/eslint-plugin` | `5.42.1` |
| `@typescript-eslint/parser` | `5.42.1` |
| `TypeScript` | `4.8.4` |
| `ESLint` | `8.15.0` |
| `node` | `web` |

### PR DESCRIPTION
**Pull request by sviat9440**
[prefer-optional-chain] A literal getter with a setter does not pass lint, and incorrectly auto-fix (#5961)

[prefer-optional-chain] The overridden literal getter does not pass lint and incorrectly auto-fix (#5962)

## PR Checklist

- [x] Addresses an existing open issue: fixes #5961, #5962

## Overview

1. Ignore accessors with computed names (like `get [a + b](): string {}`, because we cannot be sure that the computed names are not overrides. 
```
/// a.ts

const a = 'a';
const b = 'b';

class A {
  get [a + b](): string {
    return ...;
  }
}

/// b.ts
import { A } from './a';

const a = 'c';
const b = 'd';

class B extends A {
  get [a + b](): string {
    return ...;
  }
}
```
In this case, visually, the accessors keys are no different, and can be taken as an override. But in fact, this is a completely different key.
So, the supported nodes are now: `Identifier`, `PrivateIdentifier`, `Literal`

2. The rule now uses type checker, and if `parserOptions.project` is not defined, rule will cause an error. (*BREAKING CHANGE!*)
3. If the literal getter overrides the getter of the parent class. the rule will not cause an error, because there is no way to fix it.
4. If the literal getter has a setter in a pair, the rule will not cause an error, because there is no way to fix it.

### CODE REVIEW

**Comment by sviat9440** [comment]
2. For backward compatibility, I can add a try catch when creating a ParserServices, and not apply logic to find an override if `parserOptions.project` is not defined. If needed...

**Comment by bradzacher** [comment]
> The rule now uses type checker, and if parserOptions.project is not defined, rule will cause an error

Strongly disagree with this change.
Overriding a getter with another getter that specifically and intentionally returns a constant value is a really, really, really rare usecase. TBH I've never actually seen code do it.
The rule is well over 2 years old at this point and you're the first report of the problem - another point in the column of really rare.

It's not worth locking the rule into type information and restricting many people from using it just to prevent a really, really rare false-positive.

As per the comments on #5962 - the change we would accept is making the fixer a suggestion fixer so that the rule cannot automatically fix and break code.

---

>  For backward compatibility, I can add a try catch when creating a ParserServices, and not apply logic to find an override if parserOptions.project is not defined. If needed...

Making rules conditionally dependent on type information is not something we really like doing in this project. Most users struggle with understanding config as it is, and adding extra forks to rules dependent on certain configs that are separate to the rule's configs further complicates things.

**Review by JoshuaKGoldberg — CHANGES_REQUESTED** [review]
Blocked on #5961 passing triage.

**Comment by JoshuaKGoldberg** [comment]
Ping @sviat9440 - is this something you have the time to work on? No worries if not, just checking in - thanks!

**Comment by sviat9440** [comment]
@JoshuaKGoldberg I'll take a look this weekend

File: packages/eslint-plugin/src/rules/class-literal-property-style.ts [hunk 1]
```diff
@@ -110,21 +97,30 @@ export default util.createRule<Options, MessageIds>({
...
+
+function getPropertyName(key: TSESTree.PropertyName): unknown {
+  switch (key.type) {
+    case AST_NODE_TYPES.PrivateIdentifier:
+    case AST_NODE_TYPES.Identifier:
+      return key.name;
+    case AST_NODE_TYPES.Literal:
+      return key.raw;
+    default:
+      return undefined;
```
JoshuaKGoldberg, line 115: 🤔 this is never hit by tests. But the TypeScript types believe it can be hit. One of the two following cases are probably true:

* The types are wrong, and should be improved so that this `default: return undefined;` can be removed
* The unit tests are missing some cases that should be added

JoshuaKGoldberg (approved):
LGTM, thanks! I'll defer to @bradzacher since Brad was active in the comments, and this is a tricky edge case kind of scenario.

JoshuaKGoldberg (changes requested):
Whoops, meant to request changes on the test coverage dipping...

sviat9440:
@JoshuaKGoldberg ready to go

JoshuaKGoldberg:
👋 ping @sviat9440, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

JoshuaKGoldberg:
Closing this PR as it's been stale for ~6 weeks without activity. Feel free to reopen @sviat9440 if you have time - but no worries if not! If anybody wants to drive it forward, please do post your own PR - and if you use this as a start, consider adding `Co-authored-by: @sviat9440` at the end of your PR description. Thanks! 😊

PR #5986, opened 2022-11-14
docs(eslint-plugin): [no-shadow] add FAQ about enum members

bradzacher:
## Overview

High time I documented this - I kept on forgetting to add it.

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-shadow.md
+++ b/packages/eslint-plugin/docs/rules/no-shadow.md
@@ -81,7 +81,7 @@ type Func = (test: string) => typeof test;
 
 ## FAQ
 
-### Why does the rule reports on enum members that share the same name as a variable in a parent scope?
+### Why does the rule report on enum members that share the same name as a variable in a parent scope?
 
 Reporting on this case isn't a bug - it is completely intentional and correct reporting! The rule reports due to a relatively unknown feature of enums - enum members create a variable within the enum scope so that they can be referenced within the enum without a qualifier.
 
--- a/packages/eslint-plugin/docs/rules/no-shadow.md
+++ b/packages/eslint-plugin/docs/rules/no-shadow.md
@@ -98,4 +98,4 @@ console.log(Test.B);
 // what should be logged?
 ```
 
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (approved):
✨

packages/eslint-plugin/docs/rules/no-shadow.md
```diff
@@ -78,3 +78,24 @@ Examples of **correct** code with `{ ignoreFunctionTypeParameterNameValueShadow:
...
+enum Test {
+  A = 1,
+  B = A,
+}
+
+console.log(Test.B);
+// what should be logged?
+```
+
+Naively looking at the above code it might look like the log should output `2`, because the outer variable `A`'s value is `2` - however the code instead outputs `1`, which is the value of `Test.A`. This is because the unqualified code `B = A` is equivalent to the fully-qualified code `B = Test.A`. Due to this behaviour, the enum member has **shadowed** the outer variable declaration.
```
JoshuaKGoldberg, line 101: > _behaviour_

> ```plaintext
> ... Unknown word (behaviour)
> ```

_chuckles in American_ 🇺🇸

packages/eslint-plugin/docs/rules/no-shadow.md
```diff
@@ -78,3 +78,24 @@ Examples of **correct** code with `{ ignoreFunctionTypeParameterNameValueShadow:
 const test = 1;
 type Func = (test: string) => typeof test;
 ```
+
+## FAQ
+
+### Why does the rule reports on enum members that share the same name as a variable in a parent scope?
+
+Reporting on this case isn't a bug - it is completely intentional and correct reporting! The rule reports due to relatively unknown feature of enums - enum members create a variable within the enum scope so that they can be referenced within the enum without a qualifier.
```
Josh-Cena, line 86:
```suggestion
Reporting on this case isn't a bug - it is completely intentional and correct reporting! The rule reports due to a relatively unknown feature of enums - enum members create a variable within the enum scope so that they can be referenced within the enum without a qualifier.
```

packages/eslint-plugin/docs/rules/no-shadow.md
Josh-Cena, line 101:
```suggestion
Naively looking at the above code, it might look like the log should output `2`, because the outer variable `A`'s value is `2` - however, the code instead outputs `1`, which is the value of `Test.A`. This is because the unqualified code `B = A` is equivalent to the fully-qualified code `B = Test.A`. Due to this behavior, the enum member has **shadowed** the outer variable declaration.
```

Josh-Cena (commented):
Little suggestions

PR #5998, opened 2022-11-16
feat(eslint-plugin): [adjacent-overload-signatures] check BlockStatement nodes

lsdsjy:
## PR Checklist

## Overview

Overloaded function declarations can also appear in BlockStatements and should be checked.

### DIFF
--- a/packages/eslint-plugin/tests/rules/adjacent-overload-signatures.test.ts
+++ b/packages/eslint-plugin/tests/rules/adjacent-overload-signatures.test.ts
@@ -252,6 +252,21 @@ class Test {
   '#private'(arg: number): void {}
 }
     `,
+    // block statement
+    `
+function wrap() {
+  function foo(s: string);
+  function foo(n: number);
+  function foo(sn: string | number) {}
+}
+    `,
+    `
+if (true) {
+  function foo(s: string);
+  function foo(n: number);
+  function foo(sn: string | number) {}
+}
+    `,
   ],
   invalid: [
     {
@@ -259,8 +274,27 @@ class Test {
 function wrap() {
   function foo(s: string);
   function foo(n: number);
-  export type bar = number;
+  type bar = number;
   function foo(sn: string | number) {}
+}
+      `,
+      errors: [
+        {
+          messageId: 'adjacentSignature',
+          data: { name: 'foo' },
+          line: 6,
+          column: 3,
+        },
+      ],
+    },
+    {
+      code: `
+if (true) {
+  function foo(s: string);
+  function foo(n: number);
+  let a = 1;
+  function foo(sn: string | number) {}
+  foo(a);
 }
       `,
       errors: [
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/tests/rules/adjacent-overload-signatures.test.ts
```diff
@@ -256,6 +256,24 @@ class Test {
   invalid: [
     {
       code: `
+function wrap() {
+  function foo(s: string);
+  function foo(n: number);
+  export type bar = number;
```
JoshuaKGoldberg, line 262: This specific `export` line isn't valid syntax, since it's inside a function:
https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAdwE4EMAOAKAlIgbwChFFRJYEy45sBnALkTqlRjAHNcBuEs8aPCTAa2MEzAgAtgCMApqh585AD0xxUURFACemOYhnpUiALyJJshb1LlBVEbTrjmrdh0QAfC9PmLCAL5EQUA

```suggestion
```

packages/eslint-plugin/tests/rules/adjacent-overload-signatures.test.ts
```diff
@@ -256,6 +256,24 @@ class Test {
...
+      `,
+      errors: [
+        {
+          messageId: 'adjacentSignature',
+          data: { name: 'foo' },
+          line: 6,
+          column: 3,
+        },
+      ],
+    },
```
JoshuaKGoldberg, line 274: Let's add some more testing so we can be more confident about this change:
* Some test cases in `valid` 
* At least one other kind of block statement - to make sure it's not just `function` blocks that work

JoshuaKGoldberg (changes requested):
Thanks for sending this in @lsdsjy! Per our [Contributing guide](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md#pull-requests) we normally ask that PRs address a previously-opened issue, to help us triage requests & make sure they're valid before sending.

In this case I think it's fine to not because the bug is pretty clear. So just for next time please 😄 

Requesting changes on what I mentioned inline.

packages/eslint-plugin/tests/rules/adjacent-overload-signatures.test.ts
lsdsjy: You're right! Since it's an invalid case, I'll only delete the `export` keyword.

packages/eslint-plugin/tests/rules/adjacent-overload-signatures.test.ts
lsdsjy: I've added two valid test cases and another invalid case. However, other than `if` statements and function declarations, I can't think of some more common scenes where people would define functions... What do you think?

JoshuaKGoldberg: Yeah I think it's great as-is. In theory someone _could_ put them in other places like `while` loops, but IMO we don't need to exhaustively check every single possible block statement type. Just a few to make sure the rule works across a few different types.

JoshuaKGoldberg (approved):
Great, thanks for this @lsdsjy! 🙌

PR #5916, opened 2022-11-02
tests: add dependency constraints to eslint-plugin tests

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: ref #1752
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

Doing this ahead of #5573 to improve reviewability.
This PR marks all of our dependency requirements for tests back to TSv4.0.8 - I didn't go back further because #5915 bumps the minimum past that anyways

PR #5999, opened 2022-11-16
fix(typescript-estree): don't consider a cached program unless it's specified in the current `parserOptions.project` config

bradzacher:
## PR Checklist

## Overview

Discovered this while working on #5916

Previously we would consider any cached program as fair game during parsing.
Most of the time this isn't a problem because users always have the same config for an entire lint run.

However if the user has a different value for `parserOptions.project` around the place, then we would behave incorrectly - we would iterate through our cached programs and return a matching program for a file even if its tsconfig wasn't included in the current `parserOptions.project`! This is really bad because it means we'll lint a file with probably incorrect types!!!

I discovered this in the above PR because I defined two separate test runs with different tsconfigs in the same file.
Because our rule tester uses the exact same filename, the parser checked the cache, found the old program matched, and returned it - meaning we never used the new tsconfig (causing the test to fail!)

The fix is simply to enforce that we only consider a program from the cache if and only if its tsconfig exists in the current `parserOptions.project` set.

### DIFF
--- a/packages/eslint-plugin/tests/rules/non-nullable-type-assertion-style.test.ts
+++ b/packages/eslint-plugin/tests/rules/non-nullable-type-assertion-style.test.ts
@@ -1,14 +1,11 @@
-import path from 'path';
-
 import rule from '../../src/rules/non-nullable-type-assertion-style';
-import { RuleTester } from '../RuleTester';
+import { getFixturesRootDir, RuleTester } from '../RuleTester';
 
-const rootDir = path.resolve(__dirname, '../fixtures/');
 const ruleTester = new RuleTester({
   parserOptions: {
     sourceType: 'module',
-    tsconfigRootDir: rootDir,
-    project: './tsconfig.noUncheckedIndexedAccess.json',
+    tsconfigRootDir: getFixturesRootDir(),
+    project: './tsconfig.json',
   },
   parser: '@typescript-eslint/parser',
 });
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/tests/rules/non-nullable-type-assertion-style.test.ts
```diff
@@ -1,14 +1,11 @@
-import path from 'path';
```
bradzacher, line 1: the changes here were extracted from #5916 as an easy test for this behaviour.
could add a regression test against this, but I didn't bother because ultimately it's such a rare case.

packages/typescript-estree/src/create-program/createWatchProgram.ts
```diff
@@ -159,11 +159,21 @@ function getProgramsForProjects(parseSettings: ParseSettings): ts.Program[] {
...
    * before we go into the process of attempting to find and update every program
    * see if we know of a program that contains this file
    */
   for (const [tsconfigPath, existingWatch] of knownWatchProgramMap.entries()) {
+    if (!currentProjectsFromSettings.has(tsconfigPath)) {
+      // the current parser run doesn't specify this tsconfig in parserOptions.project
+      // so we don't want to consider it for caching purposes.
+      //
+      // if we did consider it we might return a program for a project
+      // that wasn't specified in the current parser run (which is obv bad!).
```
JoshuaKGoldberg, line 174: To you, maybe 😛 

```suggestion
      // that wasn't specified in the current parser run (which is bad!).
```

JoshuaKGoldberg (approved):
Nice find!

PR #6028, opened 2022-11-17
fix(eslint-plugin): [prefer-optional-chain] fix `ThisExpression` and `PrivateIdentifier` errors

### LINKED ISSUE #6024: Bug: [prefer-optional-chain] Unexpected member property type: ThisExpression

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.4&sourceType=module&code=MYGwhgzhAECC0G8CwAoa0DEYDcr0YCNcU9oBbAUwBcALAewBMAKASkVPQEsAzaJgQlqcIAOizQAPhOiCawsQTbI06VQCdqAVzUA7YqugBfUsZSmgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uii9o-aJHBgAviHVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3QAacDUjoAHoizJ46IngCGqACL9YmNXl3Cwo9FIC+IM0A

### Repro Code

```typescript
class A {
  #a;
  #b;

  method() {
    if (!this.#a || !this.#b) {
      return;
    }
  }
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/prefer-optional-chain": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true,
  }
}
```

### Expected Result

No errors

### Actual Result

Unexpected member property type: ThisExpression

### Additional Info

This is a regression of 5.43.0. In version 5.42.1 all works correctly.
I suppose it was a consequence of these changes: #5594 @omril1

### Versions

| package | version |
| -- | -- |
| `@typescript-eslint/eslint-plugin` | `5.43.0` |
| `@typescript-eslint/parser` | `5.43.0` |
| `TypeScript` | `4.8.4` |
| `ESLint` | `8.15.0` |
| `node` | `web` |

omril1:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6024

## Overview

This only prevents the plugin from crashing, I tried to support the syntax as well but got into a weird behavior from the fixer reverting the fix


packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -425,9 +425,12 @@ export default util.createRule({
 
           /* istanbul ignore next */
           default:
-            throw new Error(
-              `Unexpected member property type: ${node.object.type}`,
-            );
+            if (![AST_NODE_TYPES.Identifier, AST_NODE_TYPES.ThisExpression].includes(node.object.type)) {
```
JoshuaKGoldberg, line 428: A few lines above, there's a comment about `ALLOWED_NON_COMPUTED_PROP_TYPES`. But this change still works despite not adding to that list. I'll defer to @bradzacher as the person who wrote that comment. 🤔

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
```diff
@@ -194,6 +194,13 @@ ruleTester.run('prefer-optional-chain', rule, {
     'match && match$1 !== undefined;',
     'foo !== null && foo !== undefined;',
     "x['y'] !== undefined && x['y'] !== null;",
+    // private properties
+    'this.#a && this.#b;',
+    '!this.#a || !this.#b;',
+    'a.#foo && a.#foo.bar;',
+    '!a.#foo || !a.#foo.bar;',
+    'a.#foo?.bar;',
+    '!a.#foo?.bar;',
```
Josh-Cena, line 203: More test cases to consider:

```js
foo().#a
a.b.#a
new A().#b
```

I have no idea why this rule is so strict about what expressions it can lint, but you can in fact use private properties on all expressions, just that some may have nonsensical behaviors (e.g. `(typeof a).#a`). It shouldn't crash the linter anyway.

omril1: Can I fix the rule to not intentionally throw? this seems irrational to me.
It's better to not show a suggestion than to just break and do nothing

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
```diff
@@ -194,6 +194,17 @@ ruleTester.run('prefer-optional-chain', rule, {
...
+    'foo().#a;',
+    'a.b.#a;',
+    'new A().#b;',
+    '(await a).#b;',
```
Josh-Cena, line 207: You need to make them actually trigger the visitor, i.e.

```suggestion
    '!foo().#a || a;',
    '!a.b.#a || a;',
    '!new A().#b || a;',
    '!(await a).#b || a;',
```

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
bradzacher: So essentially the issue is that #5594 refactored the logic, meaning that `getText` gets called in new ways for a variety of new nodes.

This means that it gets called for more cases than originally intended.
Previously the call to `getText` was always guarded behind `isValidChainTarget` which (I believe) would have enforced the expected `ALLOWED_NON_COMPUTED_PROP_TYPES` constraint.

This is why the code specifically throws an exception here - because based on the prior logic it was not physically possible for any other case to ever get there!

Considering the code now has an unguarded call to `getText`, this switch/case needs to be updated to handle all possible cases!

JoshuaKGoldberg:
Something funky is up with the Git history 🤔

bradzacher:
it looks like you might have rebased weirdly or something? it's borked the PR contents!

could you please update your branch so it just contains your commits - thanks!
(feel free to force-push to fix this)

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -368,6 +368,15 @@ export default util.createRule({
...
     function getMemberExpressionText(node: TSESTree.MemberExpression): string {
+      if (
+        [
+          AST_NODE_TYPES.TSAsExpression,
+          AST_NODE_TYPES.AwaitExpression,
+          AST_NODE_TYPES.NewExpression,
+        ].includes(node.object.type)
+      ) {
+        return '';
+      }
```
omril1, line 379: This feels a bit like a hack since typescript doesn't seem to think they are assignable to `MemberExpression`
`if (node.object.type === AST_NODE_TYPES.TSAsExpression) {}` shows a typescript error, but with Array#includes it does not

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
omril1: I would love to change this rule to not intentionally throw, or at least catch the error.
The rule logic now has many combinations that could be missed and better be gracefully ignored.

@Josh-Cena @JoshuaKGoldberg @bradzacher WDYT?

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
JoshuaKGoldberg: It is a bit of a hack 😄 but the root cause is a bug in our typings, I think. A node's member _can_ be an `await` or `new`. [[typescript-eslint playground](https://typescript-eslint.io/play/#ts=4.9.3&sourceType=module&showAST=es&code=MYewdgzgLgBApgDwIYFsAOAbOMC8MDeAvjEhCWAJ4DcAUDQBRIDuSAlrIqpnAJQB0aAE4g0tRi3YwwcJvGTos-ISNo1EaEINj4YhKkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkeAhpgObp4VTolHoAebANOBqQSeTMmh4AcgFdUqAMIALdNADWZSmJnohAXxB6gA)]


```ts
const example = {} as any;

(await example).prop;
(await new example).prop;
```

For now, you can use a `// @ts-expect-error` and link to #6239

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
JoshuaKGoldberg: Yeah as Brad said in https://github.com/typescript-eslint/typescript-eslint/pull/6028#discussion_r1026051986, I think the rule should no longer throw. Throwing is useful for us finding missing combinations but is inconvenient for users.

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -435,9 +444,19 @@ export default util.createRule({
...
+                AST_NODE_TYPES.Identifier,
+                AST_NODE_TYPES.ThisExpression,
+                AST_NODE_TYPES.CallExpression,
+                AST_NODE_TYPES.MemberExpression,
+              ].includes(node.object.type)
+            ) {
+              throw new Error(
+                `Unexpected member property type: ${node.object.type}`,
+              );
+            }
```
JoshuaKGoldberg, line 458: Per https://github.com/typescript-eslint/typescript-eslint/pull/6028#discussion_r1051485218:

```suggestion
```

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -435,9 +444,19 @@ export default util.createRule({
 
           /* istanbul ignore next */
```
JoshuaKGoldberg, line 445: Clearly, we had some missing coverage 😄 

```suggestion
```

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
```diff
@@ -194,6 +194,18 @@ ruleTester.run('prefer-optional-chain', rule, {
     'match && match$1 !== undefined;',
     'foo !== null && foo !== undefined;',
     "x['y'] !== undefined && x['y'] !== null;",
+    // private properties
+    'this.#a && this.#b;',
```
JoshuaKGoldberg, line 198: This should still report a complaint:

```ts
class Violation {
  #property = 0;

  test() {
    const self = Math.random() ? this : undefined;

    self && self#.property;
  }
}
```

I suspect the issue is returning `''` instead of the actual node text above.

JoshuaKGoldberg (changes requested):
Progress! ✨

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
omril1: Oh, so I misunderstood his comment

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
JoshuaKGoldberg: bump on this 😇

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
JoshuaKGoldberg: bump on this 😇

JoshuaKGoldberg (changes requested):
Still a few comments please 😇

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
omril1: The highlighted line has two different properties and the example seems to be covered by this, am I missing something?
https://github.com/typescript-eslint/typescript-eslint/blob/0afc516667c81d5b6f9c8ea09f7d6415bccc987d/packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts#L141-L144

JoshuaKGoldberg: Ah sorry you're right, I missed that they're the same property. But this actually brings up that there's a bug in this PR for private identifiers: https://github.com/microsoft/TypeScript/issues/42734 blocks them from being used.

https://deploy-preview-6028--typescript-eslint.netlify.app/play#ts=4.9.3&sourceType=module&code=MYGwhgzhAEBqCWB7cAXJA7aBvAUNaAxAA4BOiRApiSgJ7QC80ADANw57QoUQoAUAlNg75gidD2gQKIAGYNoAWTAoAFgDoSYdABNEAWwHQA-JxXwYALmgBXHRRnx0FbW2GTpMo2uJlK1Gmz4AL44IUA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uii9o-aJHBgAviHVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

```ts
class Violation {
  #property = 0;

  test() {
    const self = Math.random() ? this : undefined;

    self && self.#property;
  }
}
```

omril1: So do I revert it for now? What do you suggest?

JoshuaKGoldberg: Well, it shouldn't throw an error - that's definitely a bug. But it also shouldn't complain. The rule should effectively ignore this case for now.

omril1: Kept the private properties in computed nodes

packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts
bradzacher: I don't know if throwing is specifically a bad thing - it's a good way to explicitly tell the user "hey the rule hasn't worked right here", rather than just failing silently and probably doing a false-negative.

The big problem is that there are a lot of unhandled cases now due to the previous refactor. As I mentioned in the comment Josh linked:

> Previously the call to getText was always guarded behind isValidChainTarget which (I believe) would have enforced the expected ALLOWED_NON_COMPUTED_PROP_TYPES constraint.
> ...
> Considering the code now has an unguarded call to getText, this switch/case needs to be updated to handle all possible cases!

bradzacher (dismissed):

bradzacher:
fix the merge conflicts and we can get this in!
thanks for your work!

omril1:
@bradzacher Thanks, fixed the conflict

PR #6038, opened 2022-11-19
chore: switched repo lint to use nx run-many

### LINKED ISSUE #4226: Split up linting internally
Following up on https://github.com/typescript-eslint/typescript-eslint/pull/4197: our linting is running out of memory in CI and without that change would likely eventually run out of memory on developer computers too. The short term fix to up the `--max-old-space-size` for Node is a bandaid fix.

Short of the dedicated CLI discussed in #4183, https://github.com/typescript-eslint/typescript-eslint/pull/4197#issuecomment-975015559 / https://github.com/typescript-eslint/typescript-eslint/pull/4197#issuecomment-977288696 would be a better strategy:

```
"lint": "nx run-many --target=lint --all",
```

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4226

## Overview

Uses `nx run-many` as described in #4226.

I tested this locally and verified:

- Reruns of packages do take from the Nx cache
- Modifying the contents of a source file invalidates the cache for that package and its dependents

Note that a single uncached run is now ~34% slower in the new system. However, the much more common use case of a user only modifying the `eslint-plugin` package is ~49% faster:

| Lint Type                         | Hyperfine Measurement | (old - new) / old x 100% |
| --------------------------------- | --------------------- | ------------------------ |
| Original (baseline)[^1]           | 39.509 s ± 3.353 s    | _(baseline)_             |
| New, completely uncached[^2]      | 53.210 s ± 6.738 s    | 34% worse                |
| New, uncached `eslint-plugin`[^3] | 20.134 s ± 0.538 s    | 49% better               |
| New, completely cached[^4]        | 478.3 ms ± 8.2 ms     | 98% better               |

[^1]: `hyperfine "cross-env NODE_OPTIONS=\"--max-old-space-size=16384\" eslint ."`
[^2]: `hyperfine "nx run-many --target=lint --all --parallel --skip-nx-cache"`
[^3]: `hyperfine "nx run-many --target=lint --all --parallel" --prepare "echo '/* ... */' >> packages/eslint-plugin/src/rules/no-floating-promises.ts || true"`
[^4]: `hyperfine "nx run-many --target=lint --all --parallel"`


packages/eslint-plugin-tslint/package.json
```diff
@@ -33,7 +33,7 @@
     "clean": "tsc -b tsconfig.build.json --clean",
     "postclean": "rimraf dist && rimraf coverage",
     "format": "prettier --write \"./**/*.{ts,mts,cts,tsx,js,mjs,cjs,jsx,json,md,css}\" --ignore-path ../../.prettierignore",
-    "lint": "eslint . --ignore-path='../../.eslintignore'",
+    "lint": "eslint .",
```
JoshuaKGoldberg, line 36: If I revert this change, we see that it looks like the root-level `.eslintignore` isn't getting properly resolved to this package's paths:

```plaintext
$ yarn lint
yarn run v1.22.19
$ eslint . --ignore-path='../../.eslintignore'

/Users/josh/repos/typescript-eslint/packages/eslint-plugin-tslint/tests/fixture-project/1.ts
  1:1  error    Unexpected var, use let or const instead                                        no-var
  1:5  warning  'foo' is assigned a value but never used. Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars

...

/Users/josh/repos/typescript-eslint/packages/eslint-plugin-tslint/tests/test-tslint-rules-directory/alwaysFailRule.js
  0:0  error  Parsing error: "parserOptions.programs" has been provided for @typescript-eslint/parser.
The file was not found in any of the provided program instance(s): packages/eslint-plugin-tslint/tests/test-tslint-rules-directory/alwaysFailRule.js

✖ 14 problems (11 errors, 3 warnings)
  6 errors and 0 warnings potentially fixable with the `--fix` option.
```

Am I missing something obvious?

bradzacher: Eslint doesn't recursively look up the tree I believe?

We should really switch off of the eslint ignore file in favour of config ignores IMO so we don't need to worry about this sort of thing

https://eslint.org/docs/latest/user-guide/configuring/ignoring-code#ignorepatterns-in-config-files

JamesHenry: Using Nx's built in executor assists with this, and it may well be more efficient than invoking the eslint CLI in a child process like this (i.e. would help with the slowdown @JoshuaKGoldberg mentioned in the uncached case)

I will put up a commit which applies this

JamesHenry (dismissed):
Pending switch over to Nx eslint executor (I'll apply that myself tomorrow)

packages/ast-spec/package.json
```diff
@@ -34,7 +34,7 @@
     "postclean": "rimraf dist && rimraf _ts3.4 && rimraf .rollup.cache && rimraf coverage",
     "clean-fixtures": "rimraf -g \"./src/**/fixtures/**/snapshots\"",
     "format": "prettier --write \"./**/*.{ts,mts,cts,tsx,js,mjs,cjs,jsx,json,md,css}\" --ignore-path ../../.prettierignore",
-    "lint": "eslint . --ignore-path='../../.eslintignore'",
+    "lint": "cd ../../ && nx lint @typescript-eslint/ast-spec",
```
JamesHenry, line 37: I kept these npm script aliases as @bradzacher had previously requested back when I applied this same change to scope-manager.

Personally I run everything from the root _or_ let the Nx Console VSCode extension give me a clickable target runner in the `project.json` of the relevant project:

(You can see it here between lines 5 and 6)

image

JoshuaKGoldberg: Nifty, thanks. Is there a way around the `cd ../..`, though? It's kind of annoying to be shoved back into the root directory. Sometimes I do want to have my shell specifically in a package dir.

JamesHenry: I _think_ there are still some rough edges with running nx from a nested location, but having said that I tried the example of running:

```sh
nx lint @typescript-eslint/ast-spec
```

Within `./packages/ast-spec` and it worked just fine...

It's possible in simple cases like this there are no issues and/or my knowledge is out of date.

I've pushed up an additional commit which removes the cd from all of them 🤞

JamesHenry (approved):

PR #6041, opened 2022-11-19
chore: removed eslint@6 fallbacks for rule messages, and a snapshot

JoshuaKGoldberg:
## Overview

I stumbled onto some more references to ESLint v6 in the codebase. I hadn't known about them. 🔪 

cc @bmish from #5968 -> #5972


bradzacher (approved):

bmish:
Good catch.

PR #6045, opened 2022-11-19
fix(eslint-plugin): [no-useless-constructor] use type checking to detect visibility changes

### LINKED ISSUE #948: [no-useless-constructor] Skipping constructors with explicit member accessibility
**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/no-useless-constructor": "error",
  }
}
```

```TS
public constructor(a: int) {
   super(a)
}
```

**Expected Result**
Error useless constructor
**Actual Result**
No error
**Additional Info**
Working fine without the `public` in the decleration

**Versions**

| `@typescript-eslint/eslint-plugin` | `2.1.0` |
| `@typescript-eslint/parser`               | `2.1.0` |
| `TypeScript`                              | `3.5.3` |
| `ESLint`                                  | `6.3.0` |
| `node`                                    | `8.11.3` |
| `npm`                                     | `6.11.2` |

JoshuaKGoldberg:
BREAKING CHANGE:
Adds use of type checking to the `no-useless-constructor` rule.

## PR Checklist

- [x] Addresses an existing open issue: fixes #948

## Overview

Uses the type checker to determine whether a derived constructor changes visibility compared to its base constructor.


packages/eslint-plugin/src/rules/no-useless-constructor.ts
```diff
@@ -65,15 +84,48 @@ export default util.createRule<Options, MessageIds>({
...
+      const checker = parserServices.program.getTypeChecker();
+
+      const superClassNode = parserServices.esTreeNodeToTSNodeMap.get(
+        parent.superClass,
+      );
+      const parentType = checker.getTypeAtLocation(superClassNode);
+      const parentSymbol = parentType.getSymbol();
+      const parentConstructor = parentSymbol?.members?.get(
+        '__constructor' as ts.__String,
+      );
```
JoshuaKGoldberg, line 115: Err. Is there a better way to get the constructor? I'm blanking...

bradzacher:
My biggest question here is "do we really want to do this?"
This forces the rule to be type-aware, which greatly restricts the people that can use it, which would force a lot of users back to the base rule.

I think most codebases are fine without this check - they just want the rule to not error for parameter properties or private constructors.

IDK if this issue is wide-spread enough to make the restriction worth it. Based on the few repots and low interaction with issues - it doesn't seem like it is?

JoshuaKGoldberg:
Yeah, you're right. The lack of activity from more than 3-4 people -only 2 👍s- is good evidence that this isn't a particularly necessary change.

+1 to what you pointed out in the issue that this is a very odd thing to do. OOP with extended classes messing with blank constructors is such a rare case in most modern JS codebases.

I'll close this PR & issue for now, but if someone has a real strong reason why we should reopen, please feel free to comment back and/or fork the rule in your own codebase!

cowwoc:
Use-case:

```
abstract class Parent { protected constructor() {} }
class Child extends Parent { public constructor() { super() } }
```

Users shouldn't be able to invoke the `Parent` constructor, but invoking the `Child` constructor is fair game. Does this help?

PR #6054, opened 2022-11-21
chore: remove unnecessary project names from nx commands

JamesHenry:
## Overview

The Nx CLI doest the "right thing" when it is run within a package's directory - it infers the relevant project and looks up the target script name based on that.

Therefore we can remove any `cd`ing and explicit referencing of project names we currently do within package npm script aliases


JoshuaKGoldberg (approved):
💯 !

PR #6066, opened 2022-11-23
feat: remove partial type-information program

### LINKED ISSUE #5508: Enhancement: Remove 2 year old backwards compatibility check from getParserServices

### Relevant Package

utils

### Description

The `getParserServices` function contains this comment:

```ts
// backwards compatibility check
// old versions of the parser would not return any parserServices unless parserOptions.project was set
```

Additionally, later on in that file:

```ts
const hasFullTypeInformation =
  context.parserServices.hasFullTypeInformation ??
  /* backwards compatible */ true;
```

Does that mean we can get rid of some of the code in that file?

### Fail

```typescript
n/a
```

### Pass

```typescript
n/a
```

### Additional Info

_No response_

bradzacher:
BREAKING CHANGE:
[utils] sets `program = null` for `getParserServices(context, true)`.
[typescript-estree] disallows `errorOnTypeScriptSyntacticAndSemanticIssues` if `project` not provided

## PR Checklist

- [x] Addresses an existing open issue: fixes #5508

## Overview

A few years ago I experimented with leveraging TS's unused variable diagnostics to power a lint rule. As this diagnostic is "single-file", I added the ability for the parser to generate a program in "isolated" mode (i.e. no `parserOptions.project` required and types/diagnostics only produced for the single file) as well as the ability for plugins to interact with this program.

However this approach had some major drawbacks:
1) calculating diagnostics in TS is slow - so the lint rule was slow as there's no way to tell TS to calculate specific diagnostics.
2) the isolated types are completely useless because in most modules so many types are seeded from imports.

We never bothered going back to cleanup this work (until now).

This PR:
1) updates the parser so that it only works with `ts.SourceFile`s if we're not doing type-aware linting.
    - this should have the added bonus of speeding the parser up for non-type-aware runs as we're not going through the heavy process of creating a program!
2) updates the `getParserServices` util as appropriate
    - Note: I left the ability to consume parser services without type info because the node maps can be useful in some situations, or sometimes it's useful to consume compiler options if provided.
3) removes the `semantic-diagnostics-enabled` test because it's a test that we don't get any value out of - it's just emitting TS errors which might change from version to version
4) only allows `errorOnTypeScriptSyntacticAndSemanticIssues` if we have full type info.
    - This is because TS doesn't compute a lot of diagnostics without a program!


packages/typescript-estree/tests/lib/test-utils/expectToHaveParserServices.ts
```diff
@@ -0,0 +1,12 @@
...
+  ParserServicesWithTypeInformation,
+} from '../../../src';
+
+export function expectToHaveParserServices(
+  services: ParserServices | null | undefined,
+): asserts services is ParserServicesWithTypeInformation {
+  expect(services?.program).toBeDefined();
+  expect(services?.esTreeNodeToTSNodeMap).toBeDefined();
+  expect(services?.tsNodeToESTreeNodeMap).toBeDefined();
+}
```
JoshuaKGoldberg, line 12: This overlaps heavily with `packages/parser/tests/tools/test-utils.ts` > `export function testServices`. Also the same file in `packages/type-utils`. Opportunity to deduplicate? Merge all into `type-utils`?

packages/typescript-estree/src/parser.ts
```diff
@@ -235,12 +270,21 @@ function parseAndGenerateServices<T extends TSESTreeOptions = TSESTreeOptions>(
...
+            tsNodeToESTreeNodeMap: astMaps.tsNodeToESTreeNodeMap,
+          }
+        : {
+            program: null,
+            // we still return the node maps because
+            // (a) they don't require type info and
+            // (b) they can be useful when using some of TS's internal non-type-aware AST utils
+            esTreeNodeToTSNodeMap: astMaps.esTreeNodeToTSNodeMap,
+            tsNodeToESTreeNodeMap: astMaps.tsNodeToESTreeNodeMap,
+          },
```
JoshuaKGoldberg, line 287: Nit: simplifying a bit...

```suggestion
    services: {
      program,
      // we always return the node maps even if there's no program because
      // (a) they don't require type info and
      // (b) they can be useful when using some of TS's internal non-type-aware AST utils
      esTreeNodeToTSNodeMap: astMaps.esTreeNodeToTSNodeMap,
      tsNodeToESTreeNodeMap: astMaps.tsNodeToESTreeNodeMap,
    },
```

packages/typescript-estree/src/parser-options.ts
```diff
@@ -182,12 +182,21 @@ export interface ParserWeakMapESTreeToTSNode<
...
   tsNodeToESTreeNodeMap: ParserWeakMap<TSNode | TSToken, TSESTree.Node>;
-  hasFullTypeInformation: boolean;
 }
+export interface ParserServicesWithTypeInformation
+  extends ParserServicesNodeMaps {
+  program: ts.Program;
+}
+export interface ParserServicesWithoutTypeInformation
+  extends ParserServicesNodeMaps {
+  program: null;
```
JoshuaKGoldberg, line 195: Seems odd to me that we'd have to explicitly add `program: null`. How about `?: never`?

```suggestion
  program?: never;
```

Here and in `shared.ts` > `interface ASTAndNoProgram`.

packages/typescript-estree/src/create-program/createSourceFile.ts
```diff
@@ -25,4 +26,11 @@ function createSourceFile(parseSettings: ParseSettings): ts.SourceFile {
       );
 }
 
-export { createSourceFile };
+function createNoProgram(parseSettings: ParseSettings): ASTAndNoProgram {
+  return {
+    ast: createSourceFile(parseSettings),
+    program: null,
+  };
+}
```
JoshuaKGoldberg, line 34: Nit: I might just skip adding this altogether and have the consumer put `{ ast: createSourceFile(parseSettings) }`. That's a stylistic concern though 🤷 no hard preference.

JoshuaKGoldberg (changes requested):
🔥 Mostly looks fantastic - just requesting changes on the `// backwards compatibility check` comment.

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
bradzacher: was me playing with things to make it work in my environment. will revert.

packages/typescript-estree/src/parser.ts
```diff
@@ -184,6 +200,25 @@ function parseAndGenerateServices<T extends TSESTreeOptions = TSESTreeOptions>(
   const shouldProvideParserServices =
     parseSettings.programs != null || parseSettings.projects?.length > 0;
 
+  if (typeof options !== 'undefined') {
+    if (
+      typeof options.errorOnTypeScriptSyntacticAndSemanticIssues ===
+        'boolean' &&
+      options.errorOnTypeScriptSyntacticAndSemanticIssues
```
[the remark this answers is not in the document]
bradzacher: copy pasta!
But it was there originally because the configs from consumers are untyped - so we don't want to make assumptions.

packages/eslint-plugin/src/rules/consistent-type-exports.ts
```diff
@@ -199,29 +221,6 @@ export default util.createRule<Options, MessageIds>({
...
-  const node = parserServices.esTreeNodeToTSNodeMap.get(specifier.exported);
-  const symbol = checker.getSymbolAtLocation(node);
-  const aliasedSymbol = checker.getAliasedSymbol(symbol!);
-
-  if (!aliasedSymbol || aliasedSymbol.escapedName === 'unknown') {
-    return undefined;
-  }
-
-  return !(aliasedSymbol.flags & SymbolFlags.Value);
-}
```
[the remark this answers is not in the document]
bradzacher: not really, but it saves a type import.
it saves passing the parserServices as an argument which saves us importing the type (which was renamed to `ParserServicesWithTypeInformation`)

packages/utils/src/eslint-utils/getParserServices.ts
```diff
@@ -1,39 +1,47 @@
...
   TOptions extends readonly unknown[],
+  TAllowWithoutFullTypeInformation extends boolean = false,
 >(
   context: Readonly<TSESLint.RuleContext<TMessageIds, TOptions>>,
-  allowWithoutFullTypeInformation = false,
-): ParserServices {
+  allowWithoutFullTypeInformation: TAllowWithoutFullTypeInformation = false as TAllowWithoutFullTypeInformation,
+): GetParserServicesResult<TAllowWithoutFullTypeInformation> {
   // backwards compatibility check
   // old versions of the parser would not return any parserServices unless parserOptions.project was set
```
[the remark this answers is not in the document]
bradzacher: So what I was addressing in this PR specifically was `parserServices.hasFullTypeInformation`.
I guess technically i'm not addressing the issue you filed.

I think we should leave this comment in because it allows us to provide a nice error message in case the user misconfigured something.

I'll update the comment to something better.

packages/utils/src/eslint-utils/getParserServices.ts
```diff
@@ -1,39 +1,47 @@
...
-  if (!hasFullTypeInformation && !allowWithoutFullTypeInformation) {
+  if (
+    context.parserServices.program == null &&
+    !allowWithoutFullTypeInformation
+  ) {
     throw new Error(ERROR_MESSAGE);
   }
 
+  // @ts-expect-error - this is safe and correct, TS just doesn't like the assignment to the conditional type
   return context.parserServices;
```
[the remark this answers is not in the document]
bradzacher: It actually needs to be `context.parserServices as GetParserServicesResult<TAllowWithoutFullTypeInformation>`

TS doesn't like the assignment that only matches one branch of the conditional type.

packages/typescript-estree/tests/lib/test-utils/expectToHaveParserServices.ts
bradzacher: TBH it feels smelly to export a util from our general-purpose utils package specifically so that we can delete 5 lines from one of our internal tests.
ESP when nobody externally would ever use the util in their tests.

packages/typescript-estree/src/parser-options.ts
bradzacher: IMO it's nice to have an explicit and known value for it. "it's always there and it's either the program or it's null" is explicit and easy to reason about.

It's the same thing I'm trying to do with the AST - the properties always exist and they have explicit empty states.

packages/typescript-estree/src/create-program/createSourceFile.ts
bradzacher: I did it just because it creates a nice symmetry/consistency in our `getProgramAndAST` function.

https://github.com/typescript-eslint/typescript-eslint/pull/6066/files#diff-a8d8fbbee764653ef3a8c2803fff21f3f825a8ad18e8a62cddbe6729555772caR79

```ts
function getProgramAndAST(
  parseSettings: ParseSettings,
  shouldProvideParserServices: boolean,
): ASTAndProgram {
  if (parseSettings.programs) {
    const fromProvidedPrograms = useProvidedPrograms(
      parseSettings.programs,
      parseSettings,
    );
    if (fromProvidedPrograms) {
      return fromProvidedPrograms;
    }
  }

  if (shouldProvideParserServices) {
    const fromProjectProgram = createProjectProgram(parseSettings);
    if (fromProjectProgram) {
      return fromProjectProgram;
    }

    // eslint-disable-next-line deprecation/deprecation -- will be cleaned up with the next major
    if (parseSettings.DEPRECATED__createDefaultProgram) {
      // eslint-disable-next-line deprecation/deprecation -- will be cleaned up with the next major
      const fromDefaultProgram = createDefaultProgram(parseSettings);
      if (fromDefaultProgram) {
        return fromDefaultProgram;
      }
    }

    return createIsolatedProgram(parseSettings);
  }

  // no need to waste time creating a program as the caller didn't want parser services
  // so we can save time and just create a lonesome source file
  return createNoProgram(parseSettings);
}
```

It's just a small style thing rather than the last line being `return { ast: createSourceFile(), program: null }`

JoshuaKGoldberg (approved):
No remaining blockers, and a lot of goodness in this PR. Let's get this merged!! 🙌

JoshuaKGoldberg:
Btw @bradzacher the `imports all available rule modules` unit test failure is fixed in the `v6` branch.

bradzacher:
For reference I just did a really quick perf test of the before-and-after for this.

```bash
$ TSESTREE_SINGLE_RUN=true node --cpu-prof ./node_modules/.bin/eslint --no-eslintrc --parser="@typescript-eslint/parser" --ext=".ts,.tsx" .
```
Loading the profiles into https://www.speedscope.app/ I can see that `getProgramAndAST` is about 100ms faster on our codebase.

It doesn't seem like much, but our codebase is relatively small - ~2.6k files for ~207k SLOC in total. For a larger code base I'd expect more of an impact, though it'll likely still be pretty small overall.

For reference the CPU profiles: [PROFILES.zip](https://github.com/typescript-eslint/typescript-eslint/files/10479796/PROFILES.zip)

Issue #5932, opened 2022-11-06
Docs: Maintenance > new TypeScript versions


JoshuaKGoldberg:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/docs/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

We don't support TypeScript nightlies, beta releases, or release candidates (RCs) - but do generally use the RCs as the sign to start working on support for the upcoming new version. Let's document this.

A few references:

* #5227
* #5914
* #5915
  * https://github.com/typescript-eslint/typescript-eslint/pull/5915#discussion_r1012292363

### Affected URL(s)

https://typescript-eslint.io/docs/maintenance/typescript-versions, maybe?

JoshuaKGoldberg:
Copying from https://github.com/typescript-eslint/typescript-eslint/issues/6057#issuecomment-1323015653: whenever there's a new syntax, we should update the scope manager.

PR #6076, opened 2022-11-24
fix(typescript-estree): stub out `ts.SatisfiesExpression` on old TS versions

bradzacher:
## PR Checklist

## Overview

Another thing we need to add to the "new syntax" contributor guide.
This union is part of the API surface for the package, so we need to ensure we don't break the build on old TS versions by "stubbing" out the new AST node types.

Also added version comments so it's easy to tell when we can remove them as we bump the minimum TS version


JoshuaKGoldberg (approved):
TIL, thanks! Added to https://github.com/typescript-eslint/typescript-eslint/issues/5932 too.

PR #5018, opened 2022-05-19
fix(eslint-plugin): [ban-types] Suggest using `object` to mean "any object"

RyanCavanaugh:
People vexed by
```ts
declare let j: unknown;
let m = { ...(j as {}) };
```
being banned by `ban-types`'s default of disallowing `{}` have resorted to using

```ts
let m = { ...(j as never) };
```
which TS 4.7 flags as an error, for good reason. The alternatives suggested by this rule are all inappropriate for the situation:
```ts
// Doesn't fix anything
let m2 = { ...(j as unknown) };

// Adds undesirable index signature to result type
let m3 = { ...(j as Record<string, never>) };

// Adds undesirable index signature to result type
let m4 = { ...(j as Record<string, unknown>) };
```
The "correct" fix is
```ts
let m = { ...(j as object) };
```

Now that `object` has been shipping in TS for several years, this seems like a safe suggestion in lieu of `Record`'s messy side effects

## PR Checklist

-   [ ] Addresses an existing open issue: fixes #000
-   [ ] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
-   [ ] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

PR #6079, opened 2022-11-24
fix(eslint-plugin): suggest object, not Record<string, never>

### LINKED ISSUE #5949: Website: Add IDs to the homepage's <h2>s

### Suggested Changes

The homepage of the website has a few sub-headings on it:

* _What are ESLint and TypeScript, and how do they compare?_
* _Why does this project exist?_
* _Financial Contributors_

...but because we've styled these as custom React components instead of using typical Docusaurus pages, they aren't assigned `id`s automatically. That means you can't provide a `#` hash link to them natively, like `https://typescript-eslint.io#financial-contributors`.

Proposal: let's add an explicit `id` attribute to each heading?

```diff
- <h2>Financial Contributors</h2>
+ <h2 id="financial-contributors">Financial Contributors</h2>
```

Example source in code: https://github.com/typescript-eslint/typescript-eslint/blob/1cc35d3af0b001667e01d6a6c453bc7a8e627416/packages/website/src/pages/index.tsx#L164

### Affected URL(s)

https://typescript-eslint.io

kmin-jeong:
## PR Checklist

- [ ] Addresses an existing open issue: fixes #5949
- [ ] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

### DIFF
--- a/packages/eslint-plugin/src/rules/ban-types.ts
+++ b/packages/eslint-plugin/src/rules/ban-types.ts
@@ -86,14 +86,14 @@ const defaultTypes: Types = {
   Object: {
     message: [
       'The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.',
-      '- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.',
+      '- If you want a type meaning "any object", you probably want `object` instead.',
       '- If you want a type meaning "any value", you probably want `unknown` instead.',
     ].join('\n'),
   },
   '{}': {
     message: [
       '`{}` actually means "any non-nullish value".',
-      '- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.',
+      '- If you want a type meaning "any object", you probably want `object` instead.',
       '- If you want a type meaning "any value", you probably want `unknown` instead.',
       '- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.',
     ].join('\n'),


Josh-Cena:
Are we un-suggesting `Record<string, never>` or `Record<string, unknown>`? To me the former is more problematic, but this PR is not changing the right thing.

kmin-jeong:
@Josh-Cena @JoshuaKGoldberg  I changed the part of the message according to the proposal in the issue and sent a pr draft. I saw the issue and understood that the object was used instead of 'Record<string, unknown>'. Did I get it wrong? I haven't worked on the code and the test code yet.

bradzacher:
This is the correct fix (it's just a re-creation of #5018).

kmin-jeong:
@bradzacher @Josh-Cena @JoshuaKGoldberg 
I fixed the part in the record. I add the short test code, and converted the draft into a pr. If I understand something wrong or there is something wrong with the code, I will correct it.

bradzacher (changes requested):
looks like you've introduced some type errors

kmin-jeong:
@bradzacher I add type Types and type checkrequiredTypes to fix lint errors.

packages/eslint-plugin/src/rules/ban-types.ts
```diff
@@ -3,16 +3,7 @@ import { AST_NODE_TYPES } from '@typescript-eslint/utils';
...
-type Types = Record<
-  string,
-  | null
-  | false
-  | string
-  | {
-      message: string;
-      fixWith?: string;
-    }
->;
```
bradzacher, line 15: I'm not sure I understand - how come you changed this type here? It doesn't seem like this was required for your PR?

kmin-jeong: I understood that changing record <string, never> to object. So I changed it to types and used it.

kmin-jeong: I modified The type types back to record<>.

kmin-jeong:
After looking at this issue and the discussion mentioned in the issue, I added the test code and requested the pr. 
I don't know the range where I shouldn't use type Record <string, never>. if I just need to modify the message  to 'use object instead', or if I need to modify the code in the  type Record <string, never> and let the message 'use object instead' come out. What do you think of @JoshuaKGoldberg who posted this issue and @bradzacher who posted a review? If you think it's unnecessary pr, I'll close it.

bradzacher:
@kmin-jeong the fix we're looking for here is to update the message so that it suggests using `object`.
You're first commit (f679c9950b36ed0e947d3855986bfb899b72a44e) was the correct fix!

kmin-jeong:
@bradzacher I vaguely thought I should fix the code, too. I modified ban-types.ts again like Commit (https://github.com/typescript-eslint/typescript-eslint/commit/f679c9950b36ed0e947d3855986bfb899b72a44e).

JoshuaKGoldberg:
Psst - you don't have to keep @ tagging us. We'll see the updates come in and reply when we can. 🙂

packages/eslint-plugin/tests/rules/ban-types.test.ts
```diff
@@ -388,6 +388,39 @@ let b: Foo<NS.Good>;
...
+          data: {
+            name: 'Record<string,never>',
+            customMessage: ' Use object instead.',
+          },
+          line: 3,
+          column: 21,
+        },
+      ],
+    },
+
```
JoshuaKGoldberg, line 423: There's no need for this test. It isn't testing anything new - this file already tests that providing `types` options are respected.

```suggestion

```

JoshuaKGoldberg (approved):
The code changes look good, thanks! I'll take care of the test revert.

I also filled out the `Overview` in the PR.

JoshuaKGoldberg (approved):

Josh-Cena:
Shouldn't the PR title say "instead of `Record<string, unknown>` then?

JoshuaKGoldberg:
Ah thanks, good catch on the title - updated.

And yeah they're not _quite_ the same but `object` is close enough for most cases.

PR #6080, opened 2022-11-24
test(website): add Playwright test for Playground linting

### LINKED ISSUE #5579: Repo: add e2e test(s) for the website playground
### Suggestion

Reference: #5572

#5551 introduced a breakage into the website playground.
We don't have checks for this, so we didn't discover this until a few days after it had been merged and released.

The playground is the only part of the website that can easily and silently break due to dependency changes or changes in our packages because all the code is lazy-loaded.
It's too cumbersome to expect every relevant change to be manually tested against the website - so we should have automated testing to ensure that the page works correctly.

The MVP for this would be a test that ensures the playground can load without issue. Given the architecture this will likely need to be some form of webdriver (or whatever the best framework is) based e2e browser test, but we'd be more than open to some other solution if there's a better way to do it (even if it means some re-architecting of the tooling).

Full coverage would include a test that is an e2e browser test that enters code and config into the playground and asserts that the errors are shown.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5579

## Overview

Adds a playground test that:

1. Types passing code & enables a rule with no issues
2. Changes the code to fail the rule
3. Autofixes the code to pass again

Also fixes an accessibility oddity with labels in the visual dropdown for rules. There's no space between rule name and description currently:

```diff
await page
  .getByLabel(
-    '@typescript-eslint/array-typeRequire consistently using either `T[]` or `Array<T>` for arrays',
+    '@typescript-eslint/array-type Require consistently using either `T[]` or `Array<T>` for arrays',
  )
  .check();

```

### DIFF
--- a/packages/website/playwright.config.ts
+++ b/packages/website/playwright.config.ts
@@ -4,6 +4,7 @@ import { devices } from '@playwright/test';
 const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
   fullyParallel: true,
+
   reporter: 'html',
   retries: 0,
   testDir: './tests',
@@ -22,6 +23,7 @@ const config: PlaywrightTestConfig = {
   webServer: {
     command: 'yarn start',
     port: 3000,
+    reuseExistingServer: !process.env.CI,
   },
   workers: process.env.CI ? 1 : undefined,
 };
--- a/packages/website/playwright.config.ts
+++ b/packages/website/playwright.config.ts
@@ -4,7 +4,6 @@ import { devices } from '@playwright/test';
 const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
   fullyParallel: true,
-
   reporter: 'html',
   retries: 0,
   testDir: './tests',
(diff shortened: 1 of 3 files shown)


packages/website/playwright.config.ts
```diff
@@ -4,6 +4,7 @@ import { devices } from '@playwright/test';
 const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
   fullyParallel: true,
+
```
JoshuaKGoldberg, line 7:
```suggestion
```

packages/website/playwright.config.ts
```diff
@@ -22,6 +23,7 @@ const config: PlaywrightTestConfig = {
   webServer: {
     command: 'yarn start',
     port: 3000,
+    reuseExistingServer: !process.env.CI,
```
JoshuaKGoldberg, line 26: Not strictly necessary for this PR, but really nice when working locally 😄.

https://playwright.dev/docs/test-advanced#launching-a-development-web-server-during-the-tests

bradzacher (approved):
genius-gif

PR #5762, opened 2022-10-07
feat(typescript-estree): clarify docs and error for program project without matching TSConfig

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5753
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

This PR does three things:

* It changes the error message from `@typescript-eslint/typescript-estree` -as called by `@typescript-eslint/parser` when `parserOptions.program` is specified but no TSConfig includes the file being linted. Now, the error message is much more clear on what's going on, and what some suggested remediations & further docs reading are
* It clarifies & expands the initial context mentioned in the docs
* It moves those docs from Typed Linting FAQs to the general troubleshooting page
    * I'm not 100% on this, and might be changing my opinion from several months ago... but my current hunch is that it'd be good to have all FAQs in one place

Issue #2094, opened 2020-05-25
Support for Project References


bradzacher:
Currently, we do not respect project references - the TS compiler APIs we use do not respect them, and we try to avoid doing tsconfig parsing directly ourselves.

This means that when two projects depended on one-another via project references,  we create a program for project `A`, causing TS to load the `.ts` files for project `A`, and `.d.ts` files for project `B`. Then we create a program for project `B`, causing TS to load the `.ts` files for project `B`.

This obviously wastes time and memory on duplicated work (the `.d.ts` files) (#1192), and makes it harder to work on and lint two separate but referenced projects simultaneously, as you have to rebuild one to lint the other (#1973).

We currently do not have a good solution for this right now because the way projects work in TS isn't exposed to us in the APIs that we use. We'll have to build something - though what that is exactly we don't have a clear picture of. 

Note that the way we work is different to TS. TS will build and check one project at a time when you run the CLI, whereas we will "build" every single project required to lint your codebase - meaning we are subject to different constrains compared to TS.
Likely there would need to be work done within TS itself so that it can deduplicate and manage things for us.

There are really two ways we can go about this:
1) treat project references as implicit and automatic `parserOptions.project` entries.
2) work with TS to deduplicate effort by sharing work when project references exist.

bajtos:
I came here via eslint/typescript-eslint#1192 from strongloop/loopback-next#5590, my primary motivation is to fix out-of-memory errors when linting a sizeable monorepo.

As the first step, I created a small monorepo where I can run and debug typescript-eslint, see https://github.com/bajtos/learn-a. To use a dev version of typescript-eslint, I manually created a symlink from my local clone of typescript-eslint to my playground:

```
cd learn-a/node_modules/@typescript-eslint
ln -s ~/src/typescript-eslint/packages/typescript-estree .
```

Then I modified [`DEFAULT_COMPILER_OPTIONS`](https://github.com/typescript-eslint/typescript-eslint/blob/30fafb09422b3aca881f4785d89b0536092d4952/packages/typescript-estree/src/create-program/shared.ts#L13-L24) and added `useSourceOfProjectReferenceRedirect: true` as recommended by @bradzacher in the issue description.

Finally, I added more logging to `getProgramsForProjects` to understand what kind of data is returned by `program.getResolvedProjectReferences()`.

[...]

PR #6084, opened 2022-11-25
feat(typescript-estree): allow specifying project: true

### LINKED ISSUE #101: Feature request: support looking up tsconfig.json relative to linted file
Currently, the `eslint-plugin-tslint` throws if any of:

```ts
      /**
       * The user needs to have configured "project" in their parserOptions
       * for @typescript-eslint/parser
       */
      if (!parserServices || !parserServices.program) {
        throw new Error(
          `You must provide a value for the "parserOptions.project" property for @typescript-eslint/parser`
        );
      }
```

or `lintFile` are missing. If `lintFile` is missing it'll just lint nothing instead of throwing, unless there are inline configs.

---

`parserOptions.project` is completely undocumented in `@typescript-eslint/parser`, and at any rate that would need to be autodetectable for this use case to work. Currently, services are just silently not generated if it's missing:

```ts
  const shouldProvideParserServices =
    shouldGenerateServices && extra.projects && extra.projects.length > 0;
```

This is despite there being a `getASTAndDefaultProject`, which looks like it could do the right thing. It's just unreachable code unless you already gave a project to begin with.

On further reading, it looks like the `tsconfig.json` can be relative, but it's by default calculated relative to `process.cwd()` and not to the current `eslintrc` being processed, or an ancestor lookup relative to the current file. The option to change that (`tsconfigRootDir`) is also undocumented and still can't do what I want.

---

Secondly, it should be possible to make the `tslint.json` also be relative to either the current `tsconfig.json`, the current `eslintrc`, or to also do a ancestor lookup from the current file.

---

I think that for most people's use cases the `process.cwd()` or `path.resolve(__dirname, "tslint.json")` approach could work, but in my use case I'm trying to make a single shareable eslint config for a monorepo and simply reuse it with `"eslintConfig": { "extends": ["@scope/eslint-config"] }`. This works for me for everything _except_ these use cases. Note that `path.resolve(__dirname, "tslint.json")` would still be wrong if a specific project has a different `tslint.json` than the one the shared eslint-config has.

Even if I leave it to run per project, I still have some specific folders that have a different `tsconfig.json` nested inside an outer `tsconfig.json` that belongs to the same project. e.g. a service worker would have `worker` instead of `dom` in its `lib` config. I can still probably deal with this through refactoring, but it sounds like something that could/should be supported; or at least vscode can support it.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #101

## Overview

Allows users to specify `parserOptions.project: true` to indicate each file should be linted with the closest `tsconfig.json` to that file.

- [x] Initial draft working for "one-and-done" use case of 
- [x] Get it to also work for watch mode, where `tsconfig.json` files might be moved between lints
    - Work... ish. We won't know if someone moves a `tsconfig.json` to re-lint a currently visible file in an IDE. But as soon as the file is changed, we'll know to re-check its relative `tsconfig.json` anyway.
- [x] Test it
- [x] Document it

~Throwing up a draft now to be very visible about progress.~ ✅


packages/typescript-estree/src/CachingMap.ts
```diff
@@ -0,0 +1,23 @@
+export class CachingMap<Key, Value> {
```
JoshuaKGoldberg, line 1: I've seen this pattern at least twice now: once in the `getProjectConfigFiles` in this PR, and once in `parser.ts` > `existingPrograms`. Are there other places in the repo that could use it? I did a search for `.get(` and `.set(` (which found `existingPrograms`) but didn't see any others.

packages/typescript-estree/src/create-program/createWatchProgram.ts
```diff
@@ -548,4 +548,4 @@ function maybeInvalidateProgram(
   return null;
 }
 
-export { clearWatchCaches, createWatchProgram, getProgramsForProjects };
+export { clearWatchCaches, getProgramsForProjects };
```
JoshuaKGoldberg, line 551: Not strictly necessary for this PR, but something I noticed - `createWatchProgram` isn't used anywhere else.

bradzacher:
> Get it to also work for watch mode, where tsconfig.json files might be moved between lints

pleasekillme4

packages/typescript-estree/src/CachingMap.ts
JoshuaKGoldberg: Meh, ended up not using this.

packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  log('Looking for tsconfig.json at or above file: %s', parseSettings.filePath);
+  let directory = path.dirname(parseSettings.filePath);
+
+  do {
+    const tsconfigPath = path.join(directory, 'tsconfig.json');
+    log('Checking tsconfig.json path: %s', tsconfigPath);
+
+    if (existsSyncCached(tsconfigPath)) {
+      return [tsconfigPath];
```
JoshuaKGoldberg, line 68: In theory, this could also check whether the `tsconfig.json` includes `parseSettings.filePath` as described in https://github.com/typescript-eslint/typescript-eslint/issues/101#issuecomment-1327052709, and skip if it doesn't. But what if the user _wanted_ a particular `tsconfig.json` to include the file, and instead got some differently configured higher-in-the-project `tsconfig.json`? 😱 

In other words we have two cases to handle:

* 🍎 Closest `tsconfig.json` is matched but does not include the file
* 🍌 Closest `tsconfig.json` matches but does not include the file

I think we should stick with this simpler logic and not parse the `tsconfig.json` because the error message for 🍌 is quite informative as of #5762.

packages/eslint-plugin-tslint/src/rules/config.ts
```diff
@@ -98,18 +98,13 @@ export default createRule<Options, MessageIds>({
...
     const sourceCode = context.getSourceCode().text;
     const parserServices = ESLintUtils.getParserServices(context);
-
-    /**
-     * The TSLint rules configuration passed in by the user
-     */
-    const [
-      { rules: tslintRules, rulesDirectory: tslintRulesDirectory, lintFile },
-    ] = context.options;
-
```
JoshuaKGoldberg, line 112: Same fix as #6089 🤷

docs/linting/TYPED_LINTING.md
```diff
@@ -12,8 +12,7 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   // Added lines start
   parserOptions: {
-    tsconfigRootDir: __dirname,
```
JoshuaKGoldberg, line 15: I've never seen `tsconfigRootDir` be useful in userland. I know there are cases where there are, but it feels like something we don't need to expend space for in the first-run / getting started docs?

packages/typescript-estree/src/create-program/createWatchProgram.ts
JoshuaKGoldberg: #6088

packages/typescript-estree/tests/lib/getProjectConfigFiles.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      mockExistsSync.mockReturnValue(false);
+
+      expect(() =>
+        getProjectConfigFiles({ ...parseSettings, tsconfigRootDir: '/' }, true),
+      ).toThrowErrorMatchingInlineSnapshot(
+        `"project was set to \`true\` but couldn't find any tsconfig.json relative to './repos/repo/packages/package/file.ts' within '/'."`,
+      );
+    });
+  });
+});
```
JoshuaKGoldberg, line 101: @bradzacher do you think I should add tests somewhere else too? Maybe `parse.test.ts`? I'm unsure on what best practice is here - those feel like integration / smokescreen tests?

docs/linting/TYPED_LINTING.md
JoshuaKGoldberg: Heh, a timely Twitter thread: https://twitter.com/JoshuaKGoldberg/status/1597272609275908097 @shawnmcknight

shawnmcknight: @JoshuaKGoldberg Thanks for tagging me in here and replying to the twitter thread.  This clarified things for me and I was able to get things working.

So in our environment where there's a workspace package for the eslint configuration, I was able to get things working with `parserOptions: { project: './tsconfig.eslint.json' },` and not specifying the `tsconfigRootDir`.

Out of the box, that worked perfectly with my turborepo lint runs which set the `cwd` to the workspace's directory.  I ran into a few issues with the vscode eslint extension and lint-staged so I thought I'd share here in the event it helps with someone else or making docs.  For vscode eslint extension, I found that I _must_ be using vscode workspaces (and have the workspace open) which causes the extension to set the `cwd` to the workspace's directory.  Additionally for `lint-staged`, I had to set up a `lint-staged` configuration file in each workspace, rather than the monorepo root, which enabled `lint-staged` to change the `cwd` to the workspace's directory for the files within that workspace. (Note: This was the proper way to enable `lint-staged` in a monorepo anyway, so it was just cleaning up an improper configuration).  After all this all tooling works -- turborepo lint runs, vscode eslint extension, and lint-staged.

JoshuaKGoldberg: Fantastic, thanks for the explanation @shawnmcknight! Filed a docs issue here: #6124

packages/typescript-estree/tests/lib/getProjectConfigFiles.test.ts
bradzacher: really that's the only way we can test this logic - with some form of physical disk structure.
you could test by mocking `fs.` but IDK doing it like that seems like you're skipping out on the most interesting part of the codebase (the actual interaction with the disk).

unit testing this logic just feels like we'll do something wrong and miss some weird filesystem behaviour and then ship bad code.

JoshuaKGoldberg: Ok, added a few tests. I didn't want to add anything testing timing at this high a level, but it does test resolving a sibling project, resolving a parent project, and throwing an error for not finding one.

docs/linting/Typed_Linting.md
```diff
@@ -28,14 +27,38 @@ module.exports = {
...
+  },
+  // ...
+};
+```
+
+See [the `@typescript-eslint/parser` docs for more details](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/parser/README.md#parseroptionsproject).
+
+:::note
+If your project is a multi-package monorepo, see [our docs on configuring a monorepo](./typed-linting/Monorepos.md).
+:::
```
JoshuaKGoldberg, line 60: Btw @bradzacher I just rediscovered https://github.com/typescript-eslint/typescript-eslint/issues/2094#issuecomment-661977512 - kind of cool to see the same idea pop up again:

> I was thinking that we can add a new variant of `parserOptions.project` - the value `true`.
>
> When `parserOptions.project === true`, we shall attempt to automatically find the `tsconfig.json` for the given file.

> if `parserOptions.project === true` - perform the new codepath
> ...

Do you think this PR conflicts with that? Or can we just shunt the described _new codepath_ behind `EXPERIMENTAL_useSourceOfProjectReferenceRedirect`?

packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  tsconfigMatchCache.set(filePath, {
+    exists,
+    timestamp: now,
+  });
+
+  return exists;
+}
+
+export function getProjectConfigFiles(
```
bradzacher, line 51: do we also want to cache the match for a file:tsconfig pair to save re-traversing?
I don't think it's a terrible thing for us to assume that tsconfigs are static - it's actually what we really do right now anyways if you have a statically defined set of paths.

what is the potential issue? we will keep using a project after you've moved/deleted the tsconfig?

Well here's a funny story - we do that right now. We don't ever delete projects from our cache - which means that after you've deleted your tsconfig, we'll still be using the program we created from it.

There is one way for us to handle the deletion - it's by crashing (haha!)
https://github.com/typescript-eslint/typescript-eslint/blob/b2b8ce2bebb1327ff08ce3d98c7521cd1bdab74c/packages/typescript-estree/src/create-program/getWatchProgramsForProjects.ts#L408-L409
As part of our "maybe invalidation" subroutine we call `fs.statSync` unguarded - if the tsconfig has been moved/deleted, then that'll throw and crash our parser.

This "maybe invalidation" path only occurs if we look for a file and don't find it in any of the `parserOptions.projects`. (i.e. algorithm is (1) check each project, (2) if not found - invalidate a project then check it again).

So ultimately the DevX of "if you move a tsconfig restart your IDE" isn't a terrible thing considering it's a rarer case and it's what users have to do right now.
The perf wins would be worth it IMO.

bradzacher: also worth noting that based on the above simplification - another caching point that we could make use of is directory-level caching.
if we parse `src/foo/bar/baz.ts` and find `src/tsconfig.json`, then we know that if we parse any of `src/foo/bar/bam.ts`, `src/foo/bam.ts` or `src/bam.ts`, then the closes tsconfig is always going to be `src/tsconfig.json`!!!

So you can cache not only the file:tsconfig pairs, but also the folder:tsconfig pairs.
I.e. i'd expect a pair of caches that look like:
```ts
const fileResolutonCache = new Map([
  ['src/foo/bar/baz.ts', 'src/tsconfig.json'],
  // ...
]);
const folderResolutionCache = new Map([
  ['src/foo/bar/', 'src/tsconfig.json'],
  ['src/foo/', 'src/tsconfig.json'],
  ['src/', 'src/tsconfig.json'],
]);
```

again the wins here would be huge because it means that once we find the tsconfig for a package, we likely only need to do an `O(1)` lookup to find the tsconfig instead of doing an `O(folder depth)` - which should be much faster given that it shaves off `O(folder depth)` file existence checks!

bradzacher (commented):
some thoughts i've had whilst thinking about this problem:
- do we want an exclusion list to allow people to blocklist specific tsconfigs from matching?
    - my gut says no because it's not how TS works right now so it shouldn't be something they want.
- do we want a way to mix-and-match solutions? for example "I want the tooling to include this specific tsconfig but also auto-resolve where possible"
     - eg `parserOptions.project = [true, './tsconfig.eslint.json']` would mean "try resolving the tsconfig automatically first, and if that doesn't work then use `./tsconfig.eslint.json`
     - `parserOptions.project = ['./tsconfig.eslint.json', true]` would be the opposite (try this config first, then try auto resolution)
     - this might be a cool behaviour but I don't have a solid usecase behind it beyond "it's cool", so probably just put a pin in it as a future potential enhancement.

bradzacher (approved):
RC for better caching, but it does overall look sound so far!

8cjsl9j3n2y61

bradzacher (changes requested):
YES I REALSIE THAT MY LAST REVIEW I CLICKED APPROVE WOOPSIE

packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts
JoshuaKGoldberg: Ah this is great! Much better, agreed 😄 thanks!

docs/linting/typed-linting/Monorepos.md
```diff
@@ -76,7 +75,6 @@ module.exports = {
   ],
   parser: '@typescript-eslint/parser',
   parserOptions: {
-    tsconfigRootDir: __dirname,
```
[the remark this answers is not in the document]
JoshuaKGoldberg: I'm in general a little -1 on including it because for a couple of reasons:

* It's extra config complexity that users will question
* Answering those questions takes up time from our already-more-than-ideal-length setup experience

Since the vast majority of projects do include a `tsconfig.json`, I think having auto-detection for it means it's not so useful anymore. In other words: I think that the misconfiguration case is now going to be so rare that it's not worth the extra config complexity.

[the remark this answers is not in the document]
bradzacher: the thing that I worry about is if you do something like this in our repo:
```bash
$ cd tools
$ yarn eslint ./postinstall.ts
```

Then:
- `yarn` will resolve `eslint` to the right place
- `eslint` will resolve the file fine and will find the eslint config fine
- `node` will execute with `cwd = $PROJECT_ROOT/tools`
- your code will throw because it'll reach `cwd` without finding a `tsconfig`

The error will be somewhat informative (`` project was set to `true` but couldn't find any tsconfig.json relative to '$PROJECT_ROOT/tools/postinstall.ts' within '$PROJECT_ROOT/tools' ``), but this would definitely be a confusing experience for users, IMO.

But if they had specified `tsconfigRootDir` then the lint run would have "just worked" as we could have understood their project layout correctly.

It's the reason we added this option in the first place - it's very, very common for people to run their lint run from a directory that isn't the project root, which would ofc break things as we'd attempt to resolve `./tsconfig.json` based on CWD.

On the CLI sure it's easy to work around this - instead of `yarn eslint`, run `(cd .. && yarn eslint`) to set the CWD correctly. But there's another case that's not possible to work around - if someone has their editor root set to a subfolder of the workspace.

IMO the extra line of config is worth it to make things "just work" and get ahead of potential user issues.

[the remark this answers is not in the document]
JoshuaKGoldberg: Got it - this is good info, thanks. I'll add them back in.

packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts
JoshuaKGoldberg: I pinged a couple of folks, we'll see if anybody picks up. In the meantime I'd vote to merge this into `main`, and make a breaking change in `v6` if we really need. I'm leaning more strongly towards _explicitly break if misconfigured_ in general.

packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts
```diff
@@ -0,0 +1,63 @@
...
+ * cache file checks for rapid subsequent calls to fs.existsSync. See discussion
+ * in https://github.com/typescript-eslint/typescript-eslint/issues/101.
+ */
+export function getProjectConfigFiles(
+  parseSettings: Pick<ParseSettings, 'filePath' | 'tsconfigRootDir'>,
+  project: string | string[] | true | undefined,
+): string | string[] | undefined {
+  if (project !== true) {
+    return project;
+  }
```
bradzacher, line 33: do we want to just normalise this function to return `string[] | undefined`?

```suggestion
): string | string[] | undefined {
  if (project !== true) {
    return Array.isArray(project) ? project : [project];
  }
```

packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts
```diff
@@ -0,0 +1,63 @@
+import debug from 'debug';
+import * as fs from 'fs';
+import * as path from 'path';
+
+import type { ParseSettings } from '.';
+
+const log = debug('typescript-eslint:typescript-estree:getProjectConfigFiles');
+
+const tsconfigMatchCache = new Map<string, string | undefined>();
```
bradzacher, line 9: nit global const variable naming

```suggestion
const TSCONFIG_MATCH_CACHE = new Map<string, string | undefined>();
```

packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  log('Looking for tsconfig.json at or above file: %s', parseSettings.filePath);
+  let directory = path.dirname(parseSettings.filePath);
+  const checkedDirectories = [directory];
+
+  do {
+    log('Checking tsconfig.json path: %s', directory);
+    const tsconfigPath = path.join(directory, 'tsconfig.json');
+    const cached =
+      tsconfigMatchCache.get(directory) ??
```
bradzacher, line 43: the only thing we need to be weary of is that this cache can get out-of-sync with the real state on disk for "persistent" runs.

we should provide a cache lifetime and nuke the cache at the start of this function if it's expired.

[`eslint-plugin-import` uses a custom Map class](https://github.com/import-js/eslint-plugin-import/blob/766af5ff6c2e97cbe2a2a211729f7c129da3005d/utils/ModuleCache.js#L6) which doesn't return cache entries older than the configured lifetime.
[They have a default cache lifetime of 30s](https://github.com/import-js/eslint-plugin-import/blob/766af5ff6c2e97cbe2a2a211729f7c129da3005d/utils/ModuleCache.js#L35-L37), so maybe we could use that as well? Could be a parser option if we wanted to allow people to reduce or increase it if they want.

(I got carried away and wrote up what I mean sorry for stealing the problem)
```ts
const LIFETIME_SECONDS: number | 'Infinity' = 30;

class TSConfigCache {
  #map = new Map<
    string,
    Readonly<{
      value: string | undefined,
      lastSeen: readonly [number, number],
    }>
  >();
  #reverseMap = new Map<string, string[]>();

  set(key, value): void {
    this.map.set(key, { value, lastSeen: process.hrtime() });
    const reverseEntry = this.reverseMap.get(key) ?? [];
    reverseEntry.push(value);
    this.reverseMap.set(key, reverseEntry);
  }

  get(key): string | undefined {
    const entry = this.map.get(key);
    if (entry) {
      if (LIFETIME_SECONDS === 'Infinity' || process.hrtime(entry.lastSeen)[0] < LIFETIME_SECONDS) {
        // cache hit woo!
        return entry.value;
      } else {
        // key has expired - clean it up to free up memory
        const reverseEntry = this.reverseMap.get(entry.value);
        if (reverseEntry) {
          for (const expiredKey of reverseEntry) {
            this.map.delete(expiredKey);
          }
          this.reverseMap.delete(entry.value);
        }
      }
    }
    // no hit :'(
    return undefined;
  }
}
```

Also worth noting - we could use this same cache class to cache our glob lookups!
Currently we don't cache glob lookups and so they're a point of slowness.

Also worth noting - if we're in single-run mode we can actually set `LIFETIME_SECONDS = 'Infinity'` because we treat the entire disk as constant when we're single-running! Which should save us a lot of time - esp for cases of `**` globs.

bradzacher (approved):
we're like 99% of the way there - just need to handle cache lifetime and we gucci

vito_good

docs/architecture/Parser.mdx
```diff
@@ -42,7 +42,7 @@ interface ParserOptions {
   lib?: string[];
   moduleResolver?: string;
   program?: import('typescript').Program;
-  project?: string | string[];
+  project?: string | string[] | true;
```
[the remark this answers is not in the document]
JoshuaKGoldberg: My first instinct was a -1 because the string form is simpler to read and I think clearer for newcomers... but now that the vast majority of projects is using `project: true` I think you're right. This makes sense.

docs/architecture/Parser.mdx
```diff
@@ -187,6 +190,10 @@ This option allows you to provide a path to your project's `tsconfig.json`. **Th
   project: ['./packages/**/tsconfig.json', './separate-package/tsconfig.json'];
   ```
 
+- If `true`, each source file's parse will find the nearest `tsconfig.json` file to that source file.
+
+  - This is done by checking that source file's directory for a `tsconfig.json`, then the parent's directory for a `tsconfig.json`, and so on - until one is found or the current working directory is passed.
```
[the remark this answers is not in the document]
JoshuaKGoldberg: 👌 beautiful

docs/architecture/Parser.mdx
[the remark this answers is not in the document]
JoshuaKGoldberg: Actually, hmm. Now I'm thinking again. Will ruminate.

packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts
bradzacher: For reference - I implemented the cache classes in my pr here https://github.com/typescript-eslint/typescript-eslint/pull/6367

docs/architecture/Parser.mdx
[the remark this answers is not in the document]
JoshuaKGoldberg: Ok yes I'm back in camp `string | string[] | true`:
* Single string options are easier to read & simpler to configure with. Which is an optimization I think we should strive for - folks are sensitive to config complexity!
* It'd be a breaking change to _remove_ the option for `string`, and I'd love to get this into `main` before v6
Thoughts @bradzacher?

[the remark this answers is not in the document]
bradzacher: We could do the removal separately, for sure.
The reason I think it's a good change is that unless you're in a really small repo - you're going to have 2 or more tsconfigs.
I think the majority of users probably use the array form - so offering both seems like overhead for us which increases complexity.

JoshuaKGoldberg:
goddamnit

packages/typescript-estree/src/parseSettings/getProjectConfigFiles.ts
JoshuaKGoldberg: Nice, I reused the class there. Thanks!

packages/typescript-estree/src/parseSettings/createParseSettings.ts
```diff
@@ -58,8 +64,14 @@ export function createParseSettings(
...
     range: options.range === true,
-    singleRun: inferSingleRun(options),
+    singleRun,
     tokens: options.tokens === true ? [] : null,
+    tsconfigMatchCache: new ExpiringCache(
+      singleRun
+        ? 'Infinity'
+        : options.cacheLifetime?.glob ??
+          DEFAULT_TSCONFIG_CACHE_DURATION_SECONDS,
+    ),
```
bradzacher, line 74: the parse settings are re-created every single time a new file is parsed - which means this cache will be discarded after a file is parsed.

it needs to be a module-level variable somewhere so we preserve it.

bradzacher (approved):
Overall this LGTM - so stamped

Before you land it - could you do a perf test just so we can update the docs with messaging around perf if we need to?

To test the perf I'd just use the `time` tool against main then against your branch:
- `git checkout main && yarn && time ./node_modules/.bin/eslint .`
- `git checkout parser-options-project-true && yarn && time ./node_modules/.bin/eslint .`

You'll see a line like this:
```
./node_modules/.bin/eslint .  62.97s user 4.09s system 167% cpu 40.009 total
```
The first number is the one to compare.
I'd do a couple runs of each just to smooth any outlier slowness.

I'd expect this to be a little slower due to the lookup required for every file - but I think that with the caching you've built in it shouldn't be MUCH slower. For a big, deeply nested project they might find it has more of an impact.

If the difference is minuscule then I don't think we need to worry about any messaging, but if it's different enough one way or the other it's probably worth mentioning!

But this is why it'd be worth a test - just to see if it is actually different enough.

---

If you wanted you could go hog-ham and investigate the proper perf trace before and after:
https://github.com/typescript-eslint/typescript-eslint/discussions/6366#discussioncomment-4756329

The thing that would be interesting to see would be the cost of the `createParseSettings` before and after as that's where your changes are.
Would be interesting to see the proper breakdown!

JoshuaKGoldberg:
- `main`: ./node_modules/.bin/eslint .  61.71s user 3.54s system 166% cpu 39.264 total
- `parser-options-project-true`: ./node_modules/.bin/eslint .  61.56s user 3.66s system 167% cpu 38.829 total

👌

Issue #2694, opened 2020-10-21
Rule proposal: disallow mixed string/numeric enums


osdiab:
Related to #2693 , in my project we only use either purely numeric or purely string enums; i want to discourage my teammates to make enums that have both, since it makes it trickier to get a definite set of enum values from the enum object.

```
// bad
enum Foo {
  A = "A",
  B,
  C = 5
}

// OK
enum Bar {
  A = "A",
  B = "B"
}

enum Baz {
  A,
  B,
  C = 5
}
```

brianconnoly:
Also it would be nice to have option to entirely disallow number initializers

PR #6091, opened 2022-11-25
feat(eslint-plugin): [no-unsafe-enum-assignment] add rule

### LINKED ISSUE #308: Rule proposal: Disallow number literals where enums are expected
Do not allow number literals to be used instead of enums. I could think of three places that this can happen:

* Function arguments 
* Binary expressions
* Assignment expressions

```ts
enum Foo {
    ONE,
    TWO
}

function getFoo(f: Foo) {

    if (f === Foo.ONE) { // Good

    }

    if (f === 1) { // Bad

    }

    let ff: Foo;

    ff = Foo.TWO // Good
    ff = 1 // Bad

}

getFoo(Foo.ONE) // Good
getFoo(1) // Bad
```

### Rule name
I need help with naming it according to conventions 

### Options 
I don't think this rule needs any options. I don't see why anyone wants to use number literals for enums.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #308

## Overview

Adds a new `no-unsafe-enum-assignment` rules that prevents assigning literals to enum values.

> `no-unsafe-enum-comparison` was split out to #6107.

Co-authored-by: @Zamiell


packages/eslint-plugin/docs/rules/strict-enums.md
```diff
@@ -0,0 +1,149 @@
+# `strict-enums`
+
+Disallows the usage of unsafe enum patterns.
+
+## Rule Details
+
+Horrifyingly, the TypeScript compiler will allow you to set any number to a variable containing a number enum, like this:
```
JoshuaKGoldberg: > Horrifyingly

We try not to express such poignant emotions in our docs. Even if we may sometimes feel this way on the inside 😄 

```suggestion
The TypeScript compiler will allow you to set any number to a variable containing a number enum, like this:
```

packages/eslint-plugin/docs/rules/strict-enums.md
```diff
@@ -0,0 +1,149 @@
...
+The goal of this rule is to make enums work like they do in other languages. One of the main benefits of enums is that they allow you to write code that is future-safe, because enums are supposed to be resilient to reorganization. If you arbitrarily change the values of an enum (or change the ordering of an enum with computed values), the idea is that nothing in your code-base should break.
+
+## Banned Patterns
+
+This rule bans:
+
+1. Enum incrementing/decrementing - `incorrectIncrement`
+1. Mismatched enum declarations/assignments - `mismatchedAssignment`
+1. Mismatched enum comparisons - `mismatchedComparison`
+1. Mismatched enum function arguments - `mismatchedFunctionArgument`
```
JoshuaKGoldberg: I'd take https://github.com/typescript-eslint/typescript-eslint/pull/4864/files#r922968222 one step further and move the more full explanation from the bottom of the file up to here. That's good info down there - and users don't see message IDs.

```suggestion
```

packages/eslint-plugin/docs/rules/strict-enums.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Options
+
+No options are provided.
+
+## Attributes
+
+- [ ] ✅ Recommended
+- [ ] 🔧 Fixable
+- [x] 💭 Requires type information
```
JoshuaKGoldberg: These are now done automatically 🤗 

```suggestion
```

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -84,7 +84,7 @@ export default util.createRule<Options, MessageId>({
         const checker = parserServices.program.getTypeChecker();
         const tsNode = parserServices.esTreeNodeToTSNodeMap.get(node);
         const type = util.getConstrainedTypeAtLocation(checker, tsNode);
-        if (!tsutils.isTypeFlagSet(type, ts.TypeFlags.VoidLike)) {
+        if (!util.isTypeFlagSet(type, ts.TypeFlags.VoidLike)) {
```
JoshuaKGoldberg: There's no real need to change all the existing `tsutils.isTypeFlagSet` locations to the new helper, is there?

packages/type-utils/src/typeFlagUtils.ts
```diff
@@ -1,31 +1,63 @@
...
+  let flags = 0;
   for (const t of unionTypeParts(type)) {
     flags |= t.flags;
   }
   return flags;
 }
 
+function isFlagSet(flags: number, flag: number): boolean {
+  return (flags & flag) !== 0;
+}
```
JoshuaKGoldberg, line 19: This is the same as what `tsutils` does on the inside: https://github.com/ajafff/tsutils/blob/03b4df15d14702f9c7a128ac3fae77171778d613/util/util.ts#L74-L76

```ts
function isFlagSet(obj: {flags: number}, flag: number) {
    return (obj.flags & flag) !== 0;
}
```

There's no need to prefer our own, I think?

packages/type-utils/src/typeFlagUtils.ts
```diff
@@ -1,31 +1,63 @@
...
+/**
+ * Similar to the `isTypeFlagSet` function, but does not decompose unions.
+ * Instead performs a single bit flag check.
+ */
+export function isTypeFlagSetSimple(
+  type: ts.Type,
+  flagsToCheck: number | ts.TypeFlags,
+): boolean {
+  return isFlagSet(type.flags, flagsToCheck);
 }
```
JoshuaKGoldberg: * I'm not a fan of words like "simple" in code. Most things aren't simple if you don't know them yet. Calling them that can be intimidating to people who read it and think they should know something they don't.
* This is the same as `tsutils.isTypeFlagSet`. We don't need it.

```suggestion
```

packages/type-utils/src/getEnumNames.ts
```diff
@@ -0,0 +1,14 @@
...
+ * number-based enum, so those have to be filtered out.)
+ *
+ * This function will work properly for both number and string enums.
+ */
+export function getEnumNames<T extends string>(
+  enumObject: Record<T, unknown>,
+): T[] {
+  const keys = Object.keys(enumObject);
+  return keys.filter(x => isNaN(parseInt(x, 10))) as T[];
+}
```
JoshuaKGoldberg: (I removed the `getEnumValues` function because it was only used in internal-only, now-removed-by-me code)

packages/eslint-plugin/src/rules/strict-enums.ts
```diff
@@ -0,0 +1,804 @@
...
+  | 'mismatchedComparison'
+  | 'mismatchedFunctionArgument';
+
+export default util.createRule<Options, MessageIds>({
+  name: 'strict-enums',
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Disallow the usage of unsafe enum code patterns',
+      recommended: false,
```
JoshuaKGoldberg, line 53: We can do this now! 🎉 

```suggestion
      recommended: 'strict',
```

packages/eslint-plugin/src/rules/strict-enums.ts
```diff
@@ -0,0 +1,804 @@
...
+          'The "getEnumKind" function is not meant to be used on unions. Use the "getEnumKinds" function instead.',
+        );
+      }
+
+      if (type.isIntersection()) {
+        throw new Error(
+          'The "getEnumKind" function is not meant to be used on intersections.',
+        );
+      }
+
```
JoshuaKGoldberg, line 122: There's no need for these developer-facing checks. If the function is used improperly, a unit test should break.

```suggestion
```

packages/eslint-plugin/src/rules/strict-enums.ts
```diff
@@ -0,0 +1,804 @@
...
+        return EnumKind.HAS_STRING_VALUES;
+      }
+
+      if (isNumberLiteral && !isStringLiteral) {
+        return EnumKind.HAS_NUMBER_VALUES;
+      }
+
+      throw new Error(
+        'Failed to derive the type of enum, since it did not have string values or number values.',
+      );
```
JoshuaKGoldberg, line 146: 🤔 What about a mixed enum?

```ts
enum Testing {
  first = 1,
  second = '2'  
}
```

packages/eslint-plugin/src/rules/strict-enums.ts
```diff
@@ -0,0 +1,804 @@
...
+        return `${i}st`;
+      }
+      if (j == 2 && k != 12) {
+        return `${i}nd`;
+      }
+      if (j == 3 && k != 13) {
+        return `${i}rd`;
+      }
+      return `${i}th`;
+    }
```
JoshuaKGoldberg, line 233: Ordinality logic didn't yet exist in the codebase because we don't need it 😄. We can just say `Argument {{ ordinal }}`. 

```suggestion
```

packages/eslint-plugin/src/rules/strict-enums.ts
```diff
@@ -0,0 +1,804 @@
...
+      return getTypeFromTSNode(tsNode);
+    }
+
+    function getTypeFromTSNode(tsNode: ts.Node): ts.Type {
+      return typeChecker.getTypeAtLocation(tsNode);
+    }
+
+    function getTypeName(type: ts.Type): string {
+      return util.getTypeName(typeChecker, type);
+    }
```
JoshuaKGoldberg: There's really no need for these two wrapper functions. They don't change anything.

```suggestion
```

JoshuaKGoldberg (commented):
(putting notes as I go through)

packages/type-utils/src/typeFlagUtils.ts
JoshuaKGoldberg: Aww, `tsutils` doesn't export `isFlagSet`...

packages/type-utils/src/typeFlagUtils.ts
```diff
@@ -1,31 +1,42 @@
...
+const ANY_OR_UNKNOWN = ts.TypeFlags.Any | ts.TypeFlags.Unknown;
+
 /**
- * Gets all of the type flags in a type, iterating through unions automatically
+ * Gets all of the type flags in a type, iterating through unions automatically.
  */
-export function getTypeFlags(type: ts.Type): ts.TypeFlags {
-  let flags: ts.TypeFlags = 0;
+export function getTypeFlags(type: ts.Type): number | ts.TypeFlags {
+  let flags = 0;
```
JoshuaKGoldberg, line 10: What's actually happening is we're assuming that `flags`, despite starting off as an invalid value for `ts.TypeFlags`, will eventually become one. I think a type assertion is correct here.

```suggestion
export function getTypeFlags(type: ts.Type): ts.TypeFlags {
  let flags = 0 as ts.TypeFlags;
```

packages/scope-manager/src/referencer/Reference.ts
```diff
@@ -66,7 +66,7 @@ class Reference {
   /**
    * In some cases, a reference may be a type, value or both a type and value reference.
    */
-  readonly #referenceType: ReferenceTypeFlag;
+  readonly #referenceType: number | ReferenceTypeFlag;
```
JoshuaKGoldberg, line 69: I'm going to try reverting these, see what happens...

packages/eslint-plugin/docs/rules/strict-enums.md
```diff
@@ -0,0 +1,122 @@
...
+enum Fruit {
+  Apple,
+  Banana,
+}
+
+let fruit = Fruit.Apple;
+fruit = 999; // No error
+```
+
+This rule flags the following unsafe enum practices:
```
JoshuaKGoldberg, line 19: @bradzacher @Zamiell it just struck me that there's a parallel between these _unsafe_ practices and our `no-unsafe-*` rules around `any`! What do you think about splitting this rule up into more granular rules? Starting name suggestions:

* `no-unsafe-enum-value-changes`: enum incrementing/decrementing / `incorrectIncrement`
* `no-unsafe-enum-assignments`: mismatched enum declarations/assignments / `mismatchedAssignment`
* `no-unsafe-enum-comparisons`: mismatched enum comparisons / `mismatchedComparison`
* `no-unsafe-enum-parameters`: mismatched enum function arguments / `mismatchedFunctionArgument`

Two thoughts in support of this:
* We had a hard time naming this rule because it's so general. That on its own is an indication it might want to be split up.
* Each of these behaviors might want to be configured granularly. It'd be easier to do that when they're in individual rules.

I'm happy to do the splitup work as part of getting this PR in.

packages/eslint-plugin/tests/rules/strict-enums/strict-enums-assignment.test.ts
```diff
@@ -0,0 +1,298 @@
...
+import {
+  fruit2EnumDefinition,
+  fruitEnumDefinition,
+  strictEnumsRuleTester,
+} from './strict-enums';
+
+const valid: ValidTestCase<unknown[]>[] = [];
+const invalid: InvalidTestCase<MessageIds, unknown[]>[] = [];
+
+valid.push({
```
JoshuaKGoldberg, line 17: I don't think I've ever seen lint rule test files `.push` to `invalid` and/or `valid` before!

Zamiell:
Hey Josh,

> What do you think about splitting this rule up into more granular rules?

I'm fine with splitting `no-unsafe-enum-value-changes` into a separate rule (that errors on any increment/modification operator). However, I think the others should be in one rule because they are essentially doing the same check (e.g. `number !== enum`).

Furthermore:
Originally, when I wrote this, I did a
- `isAssigningNonEnumValueToEnumVariable` function for variable assignment, which is a "right to left" operation (unidirectional)
- `typesHaveLooseEnumUsage` function for comparison, which is a both "left to right" and "right to left" operation (bidirectional)
- `checkCallExpression` function for functions, which is a "right to left" operational (unidirectional)

I think that this rule can be rewritten to be more DRY by getting rid of these three functions, and instead making a function with the following signature:

```ts
function isEnumMismatch(
  leftType: ts.Type,
  rightType: ts.Type,
  // If false, will only check for a mismatch from right to left. If true, will check for a mismatch from both directions
  bidirectional: boolean,
): boolean {}
```

Zamiell:
Josh,

Why did you change some function JSDoc to leading line comments? I think this is objectively worse because It hides the information on mouseover in VSCode (at either the function declaration or elsewhere in the file when the function is invoked).

JoshuaKGoldberg:
> rewritten to be more DRY

I'll take a look, thanks!

> change some function JSDoc to leading line comments

🙈 because I was very tired when coding haha. +1 to JSDoc on functions. I'd meant to just change over inline code blocks to `//`s.

Zamiell:
Also, here are some additional important tests that address some bugs I found in the wild:

comparison:

```ts
valid.push({
  name: "Comparing a number literal to a type that contains a number literal",
  code: `${fruitEnumDefinition}
declare const fruitOrInt: Fruit \| -1;
fruitOrInt === -1;
  `,
});
```

function:

```ts
valid.push({
  name: "Using a number literal in a function that takes a number literal",
  code: `${fruitEnumDefinition}
function useFruit(fruit: Fruit | -1) {}
useFruit(-1);
        `,
});

valid.push({
  name: "Using a function with this void and 2 arguments",
  code: `${fruitEnumDefinition}
function foo(this: void, arg1: Fruit, arg2: number) {}
foo(Fruit.Apple, 123);
  `,
});

valid.push({
  name: "Using a function with this void and 3 arguments",
  code: `${fruitEnumDefinition}
function foo(this: void, arg1: Fruit, arg2: number, arg3: number) {}
foo(Fruit.Apple, 123, 456);
  `,
});
```

Zamiell:
RE: Refactoring the tests
The problem with removing the names is that you lose the ability to run specific tests, which I found to be essential when debugging specific parts of the rule.

JoshuaKGoldberg:
@Zamiell can we move these conversations to threads on the code themselves please? I find it hard to keep track of them as free-floating PR comments. Waiting eagerly for the day when GitHub adds threads to issues and PRs...

packages/eslint-plugin/tests/rules/strict-enums/strict-enums-assignment.test.ts
```diff
@@ -0,0 +1,235 @@
+import rule from '../../../src/rules/strict-enums';
+import {
+  fruit2EnumDefinition,
+  fruitEnumDefinition,
+  strictEnumsRuleTester,
+} from './shared';
+
+strictEnumsRuleTester.run('strict-enums-assignment', rule, {
```
JoshuaKGoldberg, line 8: https://github.com/typescript-eslint/typescript-eslint/pull/6091#issuecomment-1327759179:

> RE: Refactoring the tests
> The problem with removing the names is that you lose the ability to run specific tests, which I found to be essential when debugging specific parts of the rule + printing out variables.

You can use `only: true`, no? https://eslint.org/docs/latest/developer-guide/unit-tests#running-individual-tests

We generally use that as it gets to be a real chore to make unique names per test.

Zamiell: Oh, I see.

packages/eslint-plugin/docs/rules/strict-enums.md
JoshuaKGoldberg: From https://github.com/typescript-eslint/typescript-eslint/pull/6091#issuecomment-1327548775:

> Hey Josh,
> 
> > What do you think about splitting this rule up into more granular rules?
> 
> I'm fine with splitting `no-unsafe-enum-value-changes` into a separate rule (that errors on any increment/modification operator). However, I think the others should be in one rule because they are essentially doing the same check (e.g. `number !== enum`).

I've played around this a bit more and think I understand better. It feels like there are two kinds of operations in play:

* `no-unsafe-enum-assignments`: e.g. `fruit++`
* `no-unsafe-enum-comparisons`: e.g. `fruit === vegetable`

🤔 now I'm less sure about splitting into two rules...

Zamiell:
Another idea for the rule is that is should prohibit mixing number and string enums together inside the same enums. TypeScript technically allows this, but it seems insane for anyone to want to do this in practice.

JoshuaKGoldberg:
#2694

JoshuaKGoldberg:
#2694 ✨

packages/eslint-plugin/docs/rules/strict-enums.md
JoshuaKGoldberg: I slept on it and am now more certain on it being two rules 😄.

packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts
```diff
@@ -0,0 +1,336 @@
...
+      return tsutils
+        .unionTypeParts(type)
+        .some(subType => util.isTypeFlagSet(subType, ts.TypeFlags.EnumLiteral));
+    }
+
+    function getTypeFromNode(node: TSESTree.Node): ts.Type {
+      return typeChecker.getTypeAtLocation(
+        parserServices.esTreeNodeToTSNodeMap.get(node),
+      );
+    }
```
bradzacher, line 45: you call this function for the left and right of each of the assignment cases.
this is going to be pretty expensive across an entire codebase because you're going to be getting a number of "useless" types.

Put another way - there's a number of cases that we can completely ignore purely based on the AST analysis.
For example: `const x = {}` We know we can ignore the `{}` because it's never going to possibly be an enum!

We should use a switch/case here to bail out some cases to improve perf by avoiding type calculation.

```ts
switch (node.type) {
  case AST_NODE_TYPES.ArrayExpression:
  case AST_NODE_TYPES.ArrowFunctionExpression:
  case AST_NODE_TYPES.ClassExpression:
  case AST_NODE_TYPES.ObjectExpression:
  case AST_NODE_TYEPS.FunctionExpression:
  // add any extras we know for sure could never be an enum
   return null;

  default:
   return /* get type */;
}
```

TBH thinking about it you could invert the problem and only check specific cases instead? Like:
- `Identifier`
- `ChainExpression`
- `MemberExpression`
- `CallExpression`
- `ConditionalExpression`
- `Literal`
- `YieldExpression`
- `AwaitExpression`
- `TSAsExpression`
- `TSSatisfiesExpression`
- `AssignmentExpression` (`const x: Enum = y = 1` this code makes me sick)
- `LogicalExpression`
- `BinaryExpression`
- `ImportExpression`
- `TemplateLiteral`
- etc

But there's a lot of cases to check and these cases are probably going to grow over time. So probably more maintainable to include a blocklist instead of an allowlist.

packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts
```diff
@@ -0,0 +1,336 @@
...
+      const providedTypeArguments = typeChecker.getTypeArguments(providedType);
+      const recipientTypeArguments =
+        typeChecker.getTypeArguments(recipientType);
+
+      for (
+        let i = 0;
+        i <
+        Math.min(recipientTypeArguments.length, providedTypeArguments.length);
+        i += 1
+      ) {
```
bradzacher, line 121: this formatting makes me physically ill, prettier

packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /**
+     * Checks whether a provided node mismatches
+     *
+     * @param provided
+     * @param recipient
+     */
+    function compareProvidedNode(
+      provided: TSESTree.Node,
```
bradzacher, line 263: nit - in other parts of the codebase I have used "sender" and "receiver".
IDK if we want to standardise or not.
I was never happy about the names tho.

packages/eslint-plugin/src/rules/no-use-before-define.ts
```diff
@@ -64,7 +64,7 @@ function isOuterEnum(
   reference: TSESLint.Scope.Reference,
 ): boolean {
   return (
-    variable.defs[0].type == DefinitionType.TSEnumName &&
+    variable.defs[0].type === DefinitionType.TSEnumName &&
```
bradzacher, line 67: we really should turn on the `eqeqeq` rule

packages/type-utils/src/typeFlagUtils.ts
```diff
@@ -1,20 +1,27 @@
...
 
+const ANY_OR_UNKNOWN = ts.TypeFlags.Any | ts.TypeFlags.Unknown;
+
 /**
- * Gets all of the type flags in a type, iterating through unions automatically
+ * Gets all of the type flags in a type, iterating through unions automatically.
  */
 export function getTypeFlags(type: ts.Type): ts.TypeFlags {
-  let flags: ts.TypeFlags = 0;
+  let flags = 0 as ts.TypeFlags;
```
bradzacher, line 10: I wonder if it would be worth adding an option to allowlist an enum name pattern as a knowingly unsafe flags enum the rule should ignore?

bradzacher (changes requested):

bradzacher:
ilikeit

Forgot to add a comment with my RC - looking good so far!
RC for performance improvement

packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts
Zamiell: whenever prettier does this it means you are doing too much on a single line.
the math.min should be on a separate line with a good variable name

packages/type-utils/src/typeFlagUtils.ts
Zamiell: isn't `flags: ts.TypeFlags = 0` since it type checks that 0 can fit inside fo the `ts.TypeFlags` type? whereas the `as` assertion disables the type checker completely

bradzacher: nope - TS doesn't do ***any*** validation of number enums. It's why they're so *completely* unsafe.

`0` isn't even a valid `ts.TypeFlags` value!
https://github.com/microsoft/TypeScript/blob/ff919e3b76c096547a4a33d7632102382eb025cd/src/compiler/types.ts#L5999-L6089

packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts
HolgerJeromin: Wow, both semicolons of the for loop looks like a normal line ending semicolons. :-)

Caching that `Math.min` would be even faster because both length are constant AFAIU.

packages/type-utils/src/typeFlagUtils.ts
HolgerJeromin: This should be possible with the new rule. So I am unsure if an allowlist is needed.
```ts
let flags: ts.TypeFlags | 0 = 0;

File: packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts (hunk 19 above)
Comment (bradzacher) [reply packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts:121]: That's a good point. The loop condition is evaluated each iteration.

Granted the loop should only be a couple of iterations, but regardless should probs extract it outwards.

File: packages/type-utils/src/typeFlagUtils.ts (hunk 22 above)
Comment (Zamiell) [reply packages/type-utils/src/typeFlagUtils.ts:10]: > TS doesn't do any validation of number enums.

ok, i know that in general, TS doesn't do any validation of number enums.
but *in general* the pattern I describe is safer typescript code, and if number enums get future improvements, then the pattern I describe would be better than using `as`.

furthermore, if 0 isn't a valid `ts.TypeFlags` value, then it should probably be expressed as Holger says in order to demonstrate that within the code

File: packages/eslint-plugin/src/rules/no-use-before-define.ts (hunk 21 above)
Comment (JoshuaKGoldberg) [reply packages/eslint-plugin/src/rules/no-use-before-define.ts:67]: Yeah.

1. #6227
2. #6228

File: packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts (hunk 20 above)
Comment (JoshuaKGoldberg) [reply packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts:263]: Yeah I'm dissatisfied overall - and wonder if there's a language/industry-general term... will change here for standardization.

Comment (JoshuaKGoldberg) [reply packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts:263]: Yeah I'm dissatisfied overall - and wonder if there's a language/industry-general term... TypeScript source often calls one _"target"_ but that feels more abstract to me. 🤷 

If you don't mind, I do like _"provided"_ and _"recipient"_ a bit more.

File: packages/eslint-plugin/typings/typescript.d.ts [hunk 23]
```diff
@@ -4,6 +4,8 @@ declare module 'typescript' {
   interface TypeChecker {
     // internal TS APIs
 
+    getContextualTypeForArgumentAtIndex(node: Node, argIndex: number): Type;
```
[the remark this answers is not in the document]
JoshuaKGoldberg: I see it as far back as 2.X.

```shell
npm i typescript@2
```

```js
ts = require("typescript")
program = ts.createProgram([], {});
checker = program.getTypeChecker();
checker.getContextualTypeForArgumentAtIndex.toString()

'function (node, argIndex) {\n' +
  '                node = ts.getParseTreeNode(node, ts.isCallLikeExpression);\n' +
  '                return node && getContextualTypeForArgumentAtIndex(node, argIndex);\n' +
  '            }'
```

packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts
JoshuaKGoldberg: A few unhappy counterpoints:

```ts
enum Fruit { Apple }

const fruits: Fruit[] = [0, 1];

function receiveFruits(received: Fruit[]) {}
receiveFruits([0, 1]);

const getFruit: () => fruit = () => 0;
```

I did add `[id.typeAnnotation]` to `VariableDeclarator[init]` so we don't check specifically the case of `const x = {}`. But I don't think there's much else we can do? 😕

package.json
```diff
@@ -105,8 +105,9 @@
     "pretty-format": "^29.0.3",
     "rimraf": "^3.0.2",
     "tmp": "^0.2.1",
-    "ts-node": "^10.7.0",
+    "ts-node": "10.7.0",
     "tslint": "^6.1.3",
+    "tsx": "^3.12.1",
```
JoshuaKGoldberg, line 110: You can ignore `package.json` changes around `ts-node` and `tsx`. See #6246.

packages/eslint-plugin/src/rules/no-unsafe-enum-assignment.ts
bradzacher: I guess this is more complicated than I would have liked 😅.

Even class expressions can be fucky:

```ts
interface Foo<T extends Fruit> {
  prop: T;
}

export const clazz: Foo<Fruit> = class WTF {
    static prop: 1 = 1;
}
```

This is going to be a very expensive rule 😢

bradzacher:
Additional cases that this rule should handle:

- class properties
  ```ts
  class F {
    prop: Fruit = 1;
  }
  ```
   - There's also cases like this:
     ```ts
     interface F {prop: Fruit}
     class Foo implements F {
       prop = 1;
     }
     ```
- object properties
  ```ts
  const x = { prop: 1 } as { prop: Fruit ))
  ```
- assertions in general?
  ```ts
  const x = 1 as Fruit;
  const y = <Fruit>1;
  ```
- JSX attributes
  ```ts
  type Props = { prop: Fruit };
  declare function Component(props: Props): JSX.Element;
  (<Component prop={1} />);
  ```
- return statements
  ```ts
  function x(): Fruit {
    return 1;
  }
  ```
- computed member expression properties
  ```ts
  declare const foo: { [k in Fruit]: string };
  foo[0];
  foo?.[0];

  declare const bar: { [Fruit.Apple]: string };
  bar[0];
  bar?.[0];
  ```

Edge cases to think about:
```ts
declare function fn(arg: Fruit): void;
fn(...[1]);

const x = [1, Fruit.Apple, 2] as const;
const y: readonly Fruit[] = x;

interface Foo<T extends Fruit> { prop: T; }
export const clazz: Foo<Fruit> = class WTF {
    static prop = 1;
}
```

packages/type-utils/src/typeFlagUtils.ts
JoshuaKGoldberg: I'd say let's keep this as a followup if people request it. It can use the format from #6017.

JoshuaKGoldberg:
Yeah, there are a _lot_ of cases to handle here. Basically any identifier and literal has to be checked - including array literals and parameters to generic functions!

I don't have the time to work on this rule given other, more important features to work on. So I'm going to close this PR. If anybody else wants to pick it up, please do!

Note that the code will likely need to be refactored a good bit to support all the new cases.

bradzacher:
I wonder @JoshuaKGoldberg - do you think this should be built directly into our `no-unsafe-*` rules, rather than being their own rule.

The reason I originally split up the rules into a set was to isolate the logic for checking assignments in each case to reduce the overall maintenance complexity.
So those rules already contain much of the logic required by this rule in terms of getting contextual types and checking generics.

They don't currently do some of the checks I've listed (like checking when `{prop: any}` is assigned to `{prop: string}`), but at least there would be parity.

WDYT? Do we refactor to reuse logic or build into those rules?

JoshuaKGoldberg:
Hmmmmm, my gut reaction is that I'd rather extract those things out into shared utilities. +1 to reusing the logic for sure. Enums really aren't the same use case as `any` safety. I also had to reimplement a bunch of them in [TypeStat](https://github.com/JoshuaKGoldberg/TypeStat) and that was a huge pain.

PR #6096, opened 2022-11-26
feat(eslint-plugin): [strict-boolean-expressions] add allow nullable enum to strict boolean expressions

### LINKED ISSUE #5614: Enhancement: [@typescript-eslint/strict-boolean-expressions] allowNullableEnum option

https://typescript-eslint.io/rules/strict-boolean-expressions/

### Description

I propose that a new option be added to the `@typescript-eslint/strict-boolean-expressions` rule: `allowNullableEnum`, which would behave similarly to the `allowNullableNumber` option, but for Enums which contain a zero as one of their possible values.  It could default to true in order to remain backwards-compatible, but allow being set false in order to catch potential issues where one of the valid values of the Enum is falsy.

### Fail

```typescript
enum ExampleEnum {
  This = 0,
  That = 1,
}

const rand = Math.random();

let theEnum: ExampleEnum | null = null;
if (rand < .3) {
  theEnum = ExampleEnum.This;
} else if (rand < .7) {
  theEnum = ExampleEnum.That;
}

// No warning thrown
if (theEnum) {
  console.log('That!');
}
```

### Pass

```typescript
enum ExampleEnum {
  This = 0,
  That = 1,
}

const rand = Math.random();

let theEnum: ExampleEnum | null = null;
if (rand < .3) {
  theEnum = ExampleEnum.This;
} else if (rand < .7) {
  theEnum = ExampleEnum.That;
}

if (theEnum === ExampleEnum.That) {
  console.log('That!');
}
```

### Additional Info

_No response_

kazizi55:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5614 

## Overview

Adds a new rule `allowNullableEnum`, in order to catch potential issues where one of the valid values of the Enum is falsy.


packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -814,6 +853,17 @@ export default util.createRule<Options, MessageId>({
         }
       }
 
+      const enums = types.filter(type =>
+        tsutils.isTypeFlagSet(
+          type,
+          ts.TypeFlags.Enum | ts.TypeFlags.EnumLike | ts.TypeFlags.EnumLiteral,
```
JoshuaKGoldberg, line 859: I tried removing the other type flags locally and all tests still passed. Either they can be removed safely or there's some missing cases to be tested.

```suggestion
          ts.TypeFlags.EnumLike,
```

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -814,6 +853,17 @@ export default util.createRule<Options, MessageId>({
...
+        ),
+      );
+
+      if (enums.length) {
```
JoshuaKGoldberg, line 863: `enums` is only used once, to check its `.length`. That means we can bail out of the loop early as soon as a match is found. A slight performance improvement:

```suggestion
      if (enums.some(enum => ...)) {
```

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -718,6 +725,37 @@ export default util.createRule<Options, MessageId>({
         return;
       }
 
+      // nullable enum
+      if (is('nullish', 'number', 'enum')) {
+        if (!options.allowNullableEnum) {
+          if (isLogicalNegationExpression(node.parent!)) {
+            // if (!nullableEnum)
```
JoshuaKGoldberg, line 732: Are these leftover from somewhere?

```suggestion
```

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
```diff
@@ -965,6 +984,56 @@ if (y) {
...
+        const rand = Math.random();
+        let theEnum: ExampleEnum | null = null;
+        if (rand < 0.3) {
+          theEnum = ExampleEnum.This;
+        }
+        if (theEnum) {
+        }
+        if (!theEnum) {
+        }
+      `,
```
JoshuaKGoldberg, line 1004: We generally ask that each test case tests a single area of logic. That makes it easier to debug them in isolation, and if one facet of logic fails, we can see exactly which case/test(s) it impacts. Could you please split both the `valid` and `invalid` cases up into the two error cases they're each meant to test?

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
```diff
@@ -965,6 +984,56 @@ if (y) {
...
+      code: `
+        enum ExampleEnum {
+          This = 0,
+          That = 1,
+        }
+        const rand = Math.random();
+        let theEnum: ExampleEnum | null = null;
+        if (rand < 0.3) {
+          theEnum = ExampleEnum.This;
+        }
```
JoshuaKGoldberg, line 999: Nit: this can be simplified a bit 😎 

```suggestion
        const theEnum = Math.random () < 0.3 ? ExampleEnum.This : null;
```

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -13,6 +13,7 @@ export type Options = [
     allowNullableBoolean?: boolean;
     allowNullableString?: boolean;
     allowNullableNumber?: boolean;
+    allowNullableEnum?: boolean;
```
JoshuaKGoldberg, line 16: We'll want to also document this on `strict-boolean-expressions.md`. The new option is good!

JoshuaKGoldberg (changes requested):
Looks great so far, thanks! Requesting changes on simplifying the code a bit, splitting up tests, and adding docs. ✨

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kazizi55: Added the option to the doc!
https://github.com/typescript-eslint/typescript-eslint/pull/6096/commits/2b71e6cb9b454ef937b443841070d717777b3857

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
kazizi55: splitted into two error cases!
https://github.com/typescript-eslint/typescript-eslint/pull/6096/commits/fae50f4ae3ce53bf0bf46ac57c395c3bc52ab7ff

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kazizi55: No, these are intentional leftover just like those in other parts.
https://github.com/kazizi55/typescript-eslint/blob/add-allowNullableEnum-to-strict-boolean-expressions/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L620

https://github.com/kazizi55/typescript-eslint/blob/add-allowNullableEnum-to-strict-boolean-expressions/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L701

Shoud I remove the leftover?

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kazizi55: Thank you for suggestion!
Actually I didn't know the performance difference between `.length` and `.some`.
Replaced!
https://github.com/typescript-eslint/typescript-eslint/pull/6096#:~:text=perf%3A%20use%20some%20instead%20of%20length

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
JoshuaKGoldberg: Yeah, I'd say remove it. We're not big on leaving old commented out code - it tends to get out of date over time and make the file harder to read through.

JoshuaKGoldberg (changes requested):
Looking excellent! ✨ 

Just the two small cleanups about comments and simplifying the types filtering. I'm happy to do those soon if you don't want to keep adjusting for my pestering 😄 - up to you.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -814,6 +853,14 @@ export default util.createRule<Options, MessageId>({
         }
       }
 
+      const enums = types.filter(type =>
+        tsutils.isTypeFlagSet(type, ts.TypeFlags.EnumLike),
+      );
+
+      if (enums.some(e => e)) {
```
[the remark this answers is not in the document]
kazizi55: Thank you for your detailed explanation!
Since the change you suggested looks different from what you explained (because of using `.filter`),  I fixed the code as you explained below.
https://github.com/typescript-eslint/typescript-eslint/pull/6096/commits/be467bab791930839be36c0f24826f080332686b

packages/eslint-plugin/docs/rules/strict-boolean-expressions.md
```diff
@@ -124,6 +124,12 @@ Allows `number | null | undefined` in a boolean context.
 This is unsafe because nullable numbers can be either a falsy number or nullish.
 Set this to `true` if you don't mind implicitly treating zero or NaN the same as a nullish value.
 
+### `allowNullableEnum`
+
+Allows `number | null | undefined` in a boolean context.
```
JoshuaKGoldberg, line 129: > `number`

It could be confusing for someone reading this how the code snippet is the same as `allowNullableNumber`. How about...

```suggestion
Allows `enum | null | undefined` in a boolean context.
```

packages/eslint-plugin/docs/rules/strict-boolean-expressions.md
```diff
@@ -124,6 +124,12 @@ Allows `number | null | undefined` in a boolean context.
...
+This is unsafe because nullable enums can be either a falsy number or nullish.
+Set this to `true` if you don't mind implicitly treating the enum, whose value is zero, the same as a nullish value.
```
JoshuaKGoldberg, line 131: Nit: a bit of streamlining

```suggestion
Set this to `true` if you don't mind implicitly treating an enum whose value is zero the same as a nullish value.
```

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -718,6 +725,36 @@ export default util.createRule<Options, MessageId>({
...
+              messageId: 'conditionErrorNullableEnum',
+              fix: util.getWrappingFixer({
+                sourceCode,
+                node: node.parent,
+                innerNode: node,
+                wrap: code => `${code} == null`,
+              }),
+            });
+          } else {
+            // if (nullableEnum)
```
JoshuaKGoldberg, line 743: Nit:

```suggestion
```

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
```diff
@@ -965,6 +998,67 @@ if (y) {
...
+          messageId: 'conditionErrorNullableEnum',
+          endLine: 7,
+          endColumn: 21,
+        },
+      ],
+      output: `
+        enum ExampleEnum {
+          This = 0,
+          That = 1,
+        }
```
JoshuaKGoldberg, line 1056: Ok great thanks for splitting up the tests! Now that I can read through them a bit more easily, I'm noticing a lack of coverage:
* Enums with implicit values (`enum ExampleEnum { This, That }`)
* Enums with string values (`enum ExampleEnum { This = '', That = "a" }`)
* Mixed enums (`enum ExampleEnum { This = '', That = 0 }`)

Of particular interest is how this deals with string-valued enums with falsy values.

JoshuaKGoldberg (changes requested):
Progress! 🚀

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
kazizi55: @JoshuaKGoldberg 
Thank you!
I added some tests and fixed the code!
https://github.com/typescript-eslint/typescript-eslint/pull/6096/commits/02ec998cf9ab24216852023d4bf9000a1a2e5b92

https://github.com/typescript-eslint/typescript-eslint/pull/6096/commits/39863e08cc9e5df7a3605933622ba453f699a5c3

bradzacher:
Need to address the lint failure

JoshuaKGoldberg (approved):
Looks great, thanks for iterating on it with us! 🙌 

The lint failure was a single test case so I just went ahead and fixed it. 🚢

PR #3891, opened 2021-09-16
feat(eslint-plugin): `prefer-consistent-enums` rule

brianconnoly:
Added new rule — `prefer-consistent-enums`

Enum members can have members of different types
Enums with string initializers can be used as object to iterate over them using `Object.keys` / `Object.values`

```ts
enum Status {
  Open = 'open',
  Closed = 'closed',
}

Object.values(Status); // ['open','closed']
```

But if member with number initializer will be added (or without one) — results of `Object.keys` / `Object.values` will have additional auto generated items

```ts
enum Status {
  Pending,
  Open = 'open',
  Closed = 'closed',
}

Object.values(Status); //  ["Pending", 0, "open", "closed"]
```

```ts
enum Status {
  Pending = 123,
  Open = 'open',
  Closed = 'closed',
}

Object.values(Status); //  ["123", "Pending", "open", "closed"]
```

This can lead to bugs, so this rule can prevent them by requiring enums to have consistent member types

PR #6102, opened 2022-11-27
feat(eslint-plugin): [no-mixed-enums] add rule

### LINKED ISSUE #2694: Rule proposal: disallow mixed string/numeric enums

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2694

## Overview

Adds a `no-mixed-enums` rule that flags the first member of an enum that mismatches its type compared to previous members. Uses the type checker for this because enum members can sometimes have dynamically computed values:

```ts
enum Annoying {
  Gotcha = Math.random()
}
```

Doesn't ban specifically numeric and/or string member values. Both are extremely common patterns and I don't think there's enough user demand to justify that.

Borrows some documentation from #3891. Thanks 😄 

Co-authored-by: @brianconnoly


bradzacher (changes requested):
> because enum members can sometimes have dynamically computed values

Note that number enum members are the only ones that are allow to have "complex" computed values:
```ts
declare function str(): string;
declare function num(): number;

declare const StRiNgCoNsTrUcToR: { new(): string };
declare const NuMbErCoNsTrUcToR: { new(): number };

enum Foo {
    A = str(),          // invalid
    B = `${'a'}`,       // invalid
    C = str.name,       // invalid
    D = 'a' + 'b',      // valid
    E = new String(''), // invalid
    F = new StRiNgCoNsTrUcToR(), // invalid

    M = num(),          // valid
    N = Number.EPSILON, // valid
    O = 1 + 1,          // valid
    P = new Number(1),  // invalid
    Q = new NuMbErCoNsTrUcToR(), // valid because it's dumb
}
```
[play](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwGcMYAKASgC5DitUBzAbgChRJYEV1s95VkBbclT78ARiBjMW4aHHhg8ReAGUMAJSwA5OgGEcmggBUYAVTCGcaqgG9eIAO5DqMWnXgBfZq1kIFqJZrIALKiAKIwegbGZhZW8LaoDk4i4jAeUiAi8ABiODjxTPBF8ACC8AC8zuQANMV1xQD0DfC0AG5QEFjAhcUAQhXwAAYAJNYA5FBj7oO19fBNLajtnd11OgNEMAB0qFD8ILP1C20dXT1FACIDE2PwANTwY6Jjh43Ny2d1oQOJ9io09BIYzGZFqxyWp1WxWyPwc-w02kiRlM5ksNXmzROKyY53gQR+AnRcze8A+UKKmgGgTEEi2oQACsoAJIAGQA8powe9Ibi2QMAIz3eD815zBZk3H02F-ampEj80FFcESuoARWl8ECIXCSOiqLU6PFkPg4jAUGQBAQWAwYwI8GAAlETHcQA)

This means we don't actually need the typechecker here.
We can simply follow the following logic:
1) if no initialiser, number
1) if initialiser
    1) if string literal or template literal, string
    1) if number literal, number
    1) if binary expression,
        1) if all operands are string literals or template literals, string
        1) else number
    1) else number

There is just one single caveat to this that I can see - enums built from enums:
```ts
enum Foo {
    A = 1,
    B = 'a',
}

enum Bar {
    A = Foo.A,
    B = Foo.B,
}
```
This is valid TS and creates a mixed enum. But IDK if we *really* care for this case? Could be a documented edge case that we wait for people to comment on? IDK.

<details>
<summary>TS even optimises `Bar` at transpile time so that it directly uses the literal values...!</summary>

```ts
var Foo;
(function (Foo) {
    Foo[Foo["A"] = 1] = "A";
    Foo["B"] = "a";
})(Foo || (Foo = {}));
var Bar;
(function (Bar) {
    Bar[Bar["A"] = 1] = "A";
    Bar["B"] = "a";
})(Bar || (Bar = {}));
```

</details>

JoshuaKGoldberg:
> enums built from enums

Oh, I've definitely seen this before. 😬 Thinking I'll try to handle it...

bradzacher (commented):
okay - doing some research on the declaration merging case - I think you can only merge enums in the following cases:
1) if both declarations are in the same file:
    ```ts
    enum Foo { A = 1 }
    enum Foo { B = 2 }
    Foo.A;
    Foo.B;
    ```
2) ambiently via module augmentation
    ```ts
    import {AST_NODE_TYPES} from '@typescript-eslint/types';
    
    declare module '@typescript-eslint/types' {
      enum AST_NODE_TYPES {
        WUT = 'WUT',
      }
    }
    ```
3) via namespace declaration merging
    ```ts
    namespace Test {
      export enum Bar { A = 1 }
    }
    namespace Test {
      export enum Bar { B = 2 }
    }
    Test.Bar.A;
    Test.Bar.B;
    ```

[playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwIIGUAqB9AcgeQCICiGaAmgAoEoC+cAZlBCHAOQACMAnmAKYDOAxlGBgYAWj4AbYADsYAek49ezANwAoNQBNu-CQEMo3OCAiaArhKNtFfQcLGSZ8m8sRq4cbtLNNUmXITE5JRuHh4A6gCqaHAAvCxRaMwANO5wVGoZan7Y+ESkFCgAdInqGmpePnAAYhAQiHBIcXAAjOkV3ky19QhwAELNAEzt3UVI6qN9ZWrSeiB8YHr8Rmh88Ahp3AAekLCenf0GDU3xbRkZs-O8i8twq7zrmzvQ8JVMfUe9A-HD52r3MCKHygY3UAKBBiB6iAA)

With that in mind - we can probably rework this rule to be faster.

WDYT of the following logic:

```ts
TSEnumDeclaration(node) {
  let members = node.members;
  if (
    isWithinTSModuleDecl(node) ||
    enumNameHasTwoDeclarationsInScope(node)
  ) {
    members = members.concat(getMergedMembers(node));
  }
  
  let desiredType: 'string' | 'number';
  for (const member of members) {
    const currentType: 'string' | 'number' = 
      getTypeFromAST(member) ?? getTypeFromTypeInfo(member);

    desiredType ??= currentType;
    if (desiredType !== currentType) {
      report(member);
    }
  }
}
```

This way we stay fast for cases that we know should be fast (which should be the vast, vast majority of code and cases), and we only dip into the slow type info for the weird edge-cases.

WDYT?

JoshuaKGoldberg:
Aha you're totally right, I love it - thanks for pressing on this! +1, will do.

JoshuaKGoldberg:
Aha you're totally right - thanks for pressing on this! +1. Will do.

It's a pity we can't quite make it to untyped - but yes I do like that this is non-type-checked mostly.

packages/eslint-plugin/src/rules/no-mixed-enums.ts
```diff
@@ -0,0 +1,222 @@
...
+      // namespace MyNamespace {
+      //   export enum MyEnum { B }
+      // }
+      if (
+        node.parent!.type === AST_NODE_TYPES.ExportNamedDeclaration &&
+        node.parent!.parent!.type === AST_NODE_TYPES.TSModuleBlock
+      ) {
+        // TODO: We don't need to dip into the TypeScript type checker here!
+        // Merged namespaces must all exist in the same file.
+        // We could instead compare this file's nodes to find the merges.
```
JoshuaKGoldberg, line 163: @bradzacher I poked around locally at writing this and got it to quite a few dozen lines of code to figure out nested namespaces in the same file [[example nested namespaces playground](https://typescript-eslint.io/play/#ts=4.9.3&sourceType=module&showAST=es&code=HYQwtgpgzgDiDGEAEB5ArgFwgJyQbwCgkkIAPGAe2wyVElgWQElhgd8jiTyqaJg0YJABVoNPEgCCSALxIA5JPlIAvpzVqCdaHESpM7QsTKVqtcDsZIWbXEa4neJAUNFRxSAEKyFn5WuINAgJ0LGwAOhsccLcMcMkAbhCDCKiI2PDPBKA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA)]. I suspect this is a pretty rare case and probably fine to treat as a followup performance improvement issue. WDYT?

packages/eslint-plugin/src/rules/no-mixed-enums.ts
```diff
@@ -0,0 +1,222 @@
...
+          return typeFromImported;
+        }
+      }
+
+      // Case: Multiple enum declarations in the same file
+      // enum MyEnum { A }
+      // enum MyEnum { B }
+      if (previousSibling) {
+        return getMemberType(previousSibling.members[0]);
+      }
```
bradzacher, line 148: we should use scope analysis here, similar to the fragment I shared below:

```ts
const enumDefinitions = context.getScope()?.upper?.set.get(enumName)?.defs;
if (enumDefinitions?.length === 1) {
  // there is exactly one definition so we're safe
  return orWhatever;
}
```

packages/eslint-plugin/src/rules/no-mixed-enums.ts
bradzacher: Actually - ignore me. This won't work because you can infinitely nest namespaces like:
```ts
namespace Foo {
  namespace Bar {
    namespace Baz {
      export enum FFS { }
    }
  }
}
namespace Foo {
  namespace Bar {
    namespace Baz {
      export enum FFS { }
    }
  }
}
```

So  I think this simple check for just "if child of a namespace, assume might merge" is good enough.

packages/eslint-plugin/src/rules/no-mixed-enums.ts
JoshuaKGoldberg: Sorry @bradzacher I don't follow - are you suggesting I replace this `if` statement, or...?

The `scope.upper?.set.get(name)?.defs` is used inside `collectNodeDefinitions` for collecting `found.previousSibling.`

bradzacher: oh shit sorry - i didn't do a full read of the change and just looked locally.
i saw "previous sibling" and thought it meant "one fetched from the AST".
all good then

packages/eslint-plugin/docs/rules/no-mixed-enums.md
```diff
@@ -0,0 +1,85 @@
...
+}
+```
+
+### ✅ Correct (Explicit Numbers)
+
+```ts
+enum Status {
+  Unknown = 0,
+  Closed = 1,
+  Open = 2
```
bradzacher, line 33:
```suggestion
  Open = 2,
}
```

packages/eslint-plugin/docs/rules/no-mixed-enums.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+### ✅ Correct (Implicit Numbers)
+
+```ts
+enum Status {
+  Unknown,
+  Closed,
+  Open
```
bradzacher, line 42:
```suggestion
  Open,
}
```

packages/eslint-plugin/docs/rules/no-mixed-enums.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+### ✅ Correct (Strings)
+
+```ts
+enum Status {
+  Unknown = 'unknown',
+  Closed = 'closed',
+  Open = 'open',
```
bradzacher, line 51:
```suggestion
  Open = 'open',
}
```

packages/eslint-plugin/src/rules/no-mixed-enums.ts
```diff
@@ -0,0 +1,222 @@
+import type { Scope } from '@typescript-eslint/scope-manager';
+import { DefinitionType } from '@typescript-eslint/scope-manager';
+import type { TSESTree } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+import * as tsutils from 'tsutils';
+import * as ts from 'typescript';
+
+import * as util from '../util';
+
+type AllowedType = ts.TypeFlags.Number | ts.TypeFlags.String;
```
bradzacher, line 10: Should we switch to our own enum instead of (ab)using TS's enum?
It looks like you never return the checker value directly and instead manually specify the return type - so there's probs not much need to reuse their enum.

packages/eslint-plugin/src/rules/no-mixed-enums.ts
```diff
@@ -0,0 +1,222 @@
...
+          }
+        });
+
+        scope = scope.upper;
+      }
+
+      return found;
+    }
+
+    function getAllowedTypeForNode(node: ts.Node): AllowedType | undefined {
```
bradzacher, line 70:
```suggestion
    function getAllowedTypeForNode(node: ts.Node): AllowedType {
```

nit - never returns `undefined`

packages/eslint-plugin/src/rules/no-mixed-enums.ts
```diff
@@ -0,0 +1,222 @@
...
+      }
+
+      return getAllowedTypeForNode(valueDeclaration.members[0]);
+    }
+
+    function getMemberType(
+      member: TSESTree.TSEnumMember,
+    ): AllowedType | ts.TypeFlags.Unknown | undefined {
+      if (!member.initializer) {
+        return undefined;
```
bradzacher, line 102: why `undefined` here instead of just `Number`?

bradzacher (dismissed):
overall LGTM!

packages/eslint-plugin/src/rules/no-mixed-enums.ts
JoshuaKGoldberg: Huh, nice spot, I guess it doesn't matter - all the tests pass either way.

PR #6105, opened 2022-11-27
feat(eslint-plugin): [explicit-function-return-type] add allowFunctionsWithoutTypeParameters option

### LINKED ISSUE #730: [explicit-function-return-type] Only apply to generic functions
The template seems to be for bug reports but I have a feature request.

I would like a way to configure `explicit-function-return-type` so that it only forces generic functions to have an explicit return type.

```ts
//Error, generic functions must have explicit return type
function foo<T extends SomeType> (t : T) {
    return /*snip some code using `t`*/;
}

//OK! Not generic, so no need for explicit return type
function blah (t : SomeType) {
    return /*snip some code using `t`*/;
}
```

### Rationale

In general, `tsc`'s type inference is very fast.

However, when it comes to complicated generic types, type inference can fail and/or infer... unfavorable types. This is especially true of generic function return types.

I personally just had a check time performance issue that was fixed by adding an explicit return type annotation to one generic function.

It cut down check time by **five minutes**

-----

This missing one line of code caused me weeks of suffering =(

I do not wish for others and future me to trip over this again =(

In general, I tend to always add explicit retun types to all my generic functions. The **one** time I forgot, it messed me up =(

-----

### Letting `tsc` infer return type

image

### Explicitly setting return type

image

The difference is staggering

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #730

## Overview

Adds a `allowFunctionsWithoutTypeParameters` option to ignore non-generic functions.

The issue is quite old but I think it makes a lot of sense to add this in. I don't often use this rule, but when I do think about it, I often mostly only want it for complex, performance-degrading types.

### DIFF
--- a/packages/eslint-plugin/src/rules/explicit-function-return-type.ts
+++ b/packages/eslint-plugin/src/rules/explicit-function-return-type.ts
@@ -15,6 +15,7 @@ type Options = [
     allowHigherOrderFunctions?: boolean;
     allowDirectConstAssertionInArrowFunctions?: boolean;
     allowConciseArrowFunctionExpressionsStartingWithVoid?: boolean;
+    allowFunctionsWithoutTypeParameters?: boolean;
     allowedNames?: string[];
   },
 ];
@@ -61,6 +62,11 @@ export default util.createRule<Options, MessageIds>({
               'Whether to ignore arrow functions immediately returning a `as const` value.',
             type: 'boolean',
           },
+          allowFunctionsWithoutTypeParameters: {
+            description:
+              "Whether to ignore functions that don't have generic type parameters.",
+            type: 'boolean',
+          },
           allowedNames: {
             description:
               'An array of function/method names that will not have their arguments or return values checked.',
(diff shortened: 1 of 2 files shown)


bradzacher (approved):
cr_thumbsup

eranhirsch:
As an avid typescript eslint fan, and someone who reads the release notes for each version, then read the linked PRs for each change, i'm missing documentation for the new flag, and it's implicit default value (i'm guessing false by reading the code). These sort of things save time when considering changing my lint config based on an update... (also, I like `explicit-function-return-type`...)

bradzacher:
A lot of our docs have auto-generated bits n bobs now! So the rule page has the type and default value generated on the website, which also includes a comment with the description of the option

https://typescript-eslint.io/rules/explicit-function-return-type/#examples

But I agree - we forgot about properly documenting this - an example would be good to illustrate this what it does!

eranhirsch:
I can't see the default value in this page, nor the option show up in the sidebar of available options. The only place the new option shows up when I search for it (via the browser's search) on that page is the options documentation itself (e.g. 1 search result on the page)

JoshuaKGoldberg:
@eranhirsch you're right, sorry about that. Filed #6465. I look forward to seeing your pull request adding in the docs. 😄

PR #6118, opened 2022-11-28
fix(eslint-plugin): [sort-type-constituents, sort-type-union-intersection-members] handle some required parentheses cases in the fixer

### LINKED ISSUE #6117: Bug: [sort-type-union-intersection-members] Fix for constituents with leading pipes or ampersands produces invalid output

### Playground Link

https://typescript-eslint.io/play/#ts=4.9.3&sourceType=module&code=C4TwDgpgBAKlC8UAUAfKBvAvgSigMg0wCgg&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZAe2jqNNtiaUOTWs3yJoyRGXxCRAW0TyARhNQYATODABfEDqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
type T = (| {}) & {}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/sort-type-union-intersection-members": 2
  },
};
```

### tsconfig

_No response_

### Expected Result

`eslint --fix` should result in valid TypeScript output:

```ts
type T = {} & (| {})
```

### Actual Result

`eslint --fix` results in this invalid code:

```ts
type T = {} & | {}
```

### Additional Info

_No response_

### Versions

| `@typescript-eslint/eslint-plugin`     | `5.44.0` |
| `@typescript-eslint/parser`            | `5.44.0` |
| `TypeScript`                           | `4.9.3` |
| `ESLint`                               | `8.28.0` |
| `node`                                 | `18.12.1` |

cherryblossom000:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6117

## Overview

`requiresParentheses` now returns `true` for union types starting with `|` and intersection types starting with `&`.

### DIFF
--- a/packages/eslint-plugin/src/rules/sort-type-constituents.ts
+++ b/packages/eslint-plugin/src/rules/sort-type-constituents.ts
@@ -220,7 +220,9 @@ export default util.createRule<Options, MessageIds>({
           const fix: TSESLint.ReportFixFunction = fixer => {
             const sorted = expectedOrder
               .map(t =>
-                typeNodeRequiresParentheses(t.node, t.text)
+                typeNodeRequiresParentheses(t.node, t.text) ||
+                (node.type === AST_NODE_TYPES.TSIntersectionType &&
+                  t.node.type === AST_NODE_TYPES.TSUnionType)
                   ? `(${t.text})`
                   : t.text,
               )
--- a/packages/eslint-plugin/src/rules/sort-type-constituents.ts
+++ b/packages/eslint-plugin/src/rules/sort-type-constituents.ts
@@ -96,10 +96,18 @@ function getGroup(node: TSESTree.TypeNode): Group {
   }
 }
 
-function requiresParentheses(node: TSESTree.TypeNode): boolean {
+interface Constituent {
+  group: number;
+  node: TSESTree.TypeNode;
+  text: string;
+}
+
+function requiresParentheses({ node, text }: Constituent): boolean {
   return (
     node.type === AST_NODE_TYPES.TSFunctionType ||
-    node.type === AST_NODE_TYPES.TSConstructorType
+    node.type === AST_NODE_TYPES.TSConstructorType ||
+    (node.type === AST_NODE_TYPES.TSUnionType && text.startsWith('|')) ||
+    (node.type === AST_NODE_TYPES.TSIntersectionType && text.startsWith('&'))
   );
 }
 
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/src/rules/sort-type-union-intersection-members.ts
```diff
@@ -96,10 +96,18 @@ function getGroup(node: TSESTree.TypeNode): Group {
   }
 }
 
-function requiresParentheses(node: TSESTree.TypeNode): boolean {
+interface Constituent {
+  group: number;
+  node: TSESTree.TypeNode;
+  text: string;
+}
```
JoshuaKGoldberg, line 103: 🤔 this code is getting more complex _(which is good, it's working better!)_ and is roughly duplicated across two files. Could you extract it to a shared helper please? Maybe, `typeNodeRequiresParentheses`? `packages/eslint-plugin/src/util/` would be a good place to put it.

packages/eslint-plugin/src/rules/sort-type-constituents.ts
```diff
@@ -183,7 +191,7 @@ export default util.createRule<Options, MessageIds>({
     function checkSorting(
       node: TSESTree.TSIntersectionType | TSESTree.TSUnionType,
     ): void {
-      const sourceOrder = node.types.map(type => {
+      const sourceOrder = node.types.map((type): Constituent => {
```
JoshuaKGoldberg, line 194: Nit: any particular reason to add the return type explicitly, here and in the other file?

JoshuaKGoldberg (changes requested):
Code looks great, thanks! Just requesting changes on deduplicating it a bit. ✨

packages/eslint-plugin/src/rules/sort-type-constituents.ts
cherryblossom000: I wanted to make it clear what the `Constituent` was and where it came from. I ended up getting rid of `Constituent` when I extracted `requiresParentheses` into a utility function.

cherryblossom000:
The force push is because I rebased with main due to a merge conflict.

cherryblossom000:
I also realised another bug with this rule:

```ts
type T = (B | C) & A
```

gets fixed to

```ts
type T = C | A & B
```

This is because `(B | C) & A` gets incorrectly fixed to `A & B | C`, which is then fixed to `C | A & B`.

bradzacher (commented):
I wonder if we want to leverage TS's utilities here for completeness?
https://github.com/microsoft/TypeScript/blob/1b75edcec60415dfa77be2e04aeafce0efa2222e/src/compiler/factory/parenthesizerRules.ts

We probably can't consume them directly (cos I don't think the util existed in old versions of TS) - but we could probably fork it? Maybe something to do in a separate PR?

bradzacher (approved):
Thanks for the fix for both rules!

flextape

I'm going to land this as is so we can get the fix out ASAP.
If we want to strengthen the logic further we can follow up in a separate PR.

PR #6160, opened 2022-12-03
chore: use short form for nx project names

JamesHenry:
For example, allows us to write:

```sh
nx lint utils
```

instead of 

```sh
nx lint @typescript-eslint/utils
```

The names needed to match the package.json files a longgg time ago when I first set things up, but that's no longer the case so this would be a nice little QOL improvement


JoshuaKGoldberg (approved):
Yes! This was already starting to get irksome after only a couple of days of using `nx` commands. Love it. 🔥

PR #6161, opened 2022-12-03
feat(eslint-plugin): [restrict-plus-operands] add allow* options

### LINKED ISSUE #6110: Enhancement: [restrict-plus-operands] Bring in options from restrict-template-expressions

https://typescript-eslint.io/rules/restrict-plus-operands

### Description

`restrict-plus-operands` and `restrict-template-expressions` don't tackle the exact same JavaScript behaviors (`+` operands and template expressions don't concatenate/`toString()`/`toValue()` quite the same). But they both generally tackle the specific common case of code trying to join a string with a non-string:

```ts
'' + 0;
'' + {};
// _roughly_ equivalent to:
`${0}`;
`${{}}`;
```

As a user, I've been inconvenienced by `restrict-plus-operands` complaining on concatenating two primitive values: e.g. `number` and `string`. Sometimes I want to be able to join primitives together in my logs!

Proposal: can we add the following options that already exist in `restrict-template-expressions`?

* ~[`allowAny`](https://typescript-eslint.io/rules/restrict-template-expressions/#allowany)~: already exists!
* [`allowNumber`](https://typescript-eslint.io/rules/restrict-template-expressions/#allownumber)
* [`allowBoolean`](https://typescript-eslint.io/rules/restrict-template-expressions/#allowboolean)
* [`allowNullish`](https://typescript-eslint.io/rules/restrict-template-expressions/#allownullish)
* [`allowRegExp`](https://typescript-eslint.io/rules/restrict-template-expressions/#allowregexp)

### Fail

```typescript
'' + {}
```

### Pass

```typescript
'' + 0
```

### Additional Info

I think we've seen a bigger demand for them in `restrict-template-expressions` because of https://eslint.org/docs/latest/rules/prefer-template. Linted projects don't concatenate strings with `+` as much as they used to.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6110

## Overview

Brings in the `allowBoolean`, `allowNullish`, and `allowRegExp` options from `@typescript-eslint/restrict-template-expressions`. Also adds an `allowNumberWithString` option equivalent to the `allowNumber` option, as a more-specific-to-this-rule name.

Internally streamlines the rule a bit by having comparisons directly refer to `ts.TypeFlags` checks whenever possible.

### DIFF
--- a/packages/eslint-plugin/docs/rules/restrict-plus-operands.md
+++ b/packages/eslint-plugin/docs/rules/restrict-plus-operands.md
@@ -18,48 +18,18 @@ This rule reports when a `+` operation combines two values of different types, o
 ### ❌ Incorrect
 
 ```ts
-var foo = '5.5' + 5;
-var foo = 1n + 1;
+let foo = '5.5' + 5;
+let foo = 1n + 1;
 ```
 
 ### ✅ Correct
 
 ```ts
-var foo = parseInt('5.5', 10) + 10;
-var foo = 1n + 1n;
+let foo = parseInt('5.5', 10) + 10;
+let foo = 1n + 1n;
 ```
 
 ## Options
-
-### `checkCompoundAssignments`
-
-Examples of code for this rule with `{ checkCompoundAssignments: true }`:
-
-<!--tabs-->
-
-#### ❌ Incorrect
-
-```ts
-/*eslint @typescript-eslint/restrict-plus-operands: ["error", { "checkCompoundAssignments": true }]*/
-
-let foo: string | undefined;
-foo += 'some data';
-
-let bar: string = '';
-bar += 0;
-```
-
-#### ✅ Correct
-
-```ts
-/*eslint @typescript-eslint/restrict-plus-operands: ["error", { "checkCompoundAssignments": true }]*/
-
-let foo: number = 0;
-foo += 1;
-
-let bar = '';
-bar += 'test';
-```
 
 ### `allowAny`
 
@@ -70,18 +40,123 @@ Examples of code for this rule with `{ allowAny: true }`:
 #### ❌ Incorrect
 
 ```ts
-var fn = (a: any, b: boolean) => a + b;
-var fn = (a: any, b: []) => a + b;
-var fn = (a: any, b: {}) => a + b;
+let fn = (a: number, b: []) => a + b;
+let fn = (a: string, b: []) => a + b;
 ```
 
 #### ✅ Correct
 
 ```ts
-var fn = (a: any, b: any) => a + b;
-var fn = (a: any, b: string) => a + b;
-var fn = (a: any, b: bigint) => a + b;
-var fn = (a: any, b: number) => a + b;
+let fn = (a: number, b: any) => a + b;
+let fn = (a: string, b: any) => a + b;
+```
+
+### `allowBoolean`
+
+Examples of code for this rule with `{ allowBoolean: true }`:
+
+<!--tabs-->
+
+#### ❌ Incorrect
+
+```ts
+let fn = (a: number, b: unknown) => a + b;
+let fn = (a: string, b: unknown) => a + b;
+```
+
+#### ✅ Correct
+
+```ts
+let fn = (a: number, b: boolean) => a + b;
+let fn = (a: string, b: boolean) => a + b;
+```
+
+### `allowNullish`
+
+Examples of code for this rule with `{ allowNullish: true }`:
+
+<!--tabs-->
+
+#### ❌ Incorrect
+
+```ts
+let fn = (a: number, b: unknown) => a + b;
+let fn = (a: number, b: never) => a + b;
+let fn = (a: string, b: unknown) => a + b;
+let fn = (a: string, b: never) => a + b;
+```
+
+#### ✅ Correct
+
+```ts
+let fn = (a: number, b: undefined) => a + b;
+let fn = (a: number, b: null) => a + b;
+let fn = (a: string, b: undefined) => a + b;
+let fn = (a: string, b: null) => a + b;
+```
+
+### `allowNumberAndString`
+
+Examples of code for this rule with `{ allowNumberAndString: true }`:
+
+<!--tabs-->
+
+#### ❌ Incorrect
+
+```ts
+let fn = (a: number, b: unknown) => a + b;
+let fn = (a: number, b: never) => a + b;
+```
+
+#### ✅ Correct
+
+```ts
+let fn = (a: number, b: string) => a + b;
+let fn = (a: number, b: number | string) => a + b;
+```
+
+### `allowRegExp`
+
+Examples of code for this rule with `{ allowRegExp: true }`:
+
+<!--tabs-->
+
+#### ❌ Incorrect
+
+```ts
+let fn = (a: number, b: RegExp) => a + b;
+```
+
+#### ✅ Correct
+
+```ts
+let fn = (a: string, b: RegExp) => a + b;
+```
+
+### `checkCompoundAssignments`
+
+Examples of code for this rule with `{ checkCompoundAssignments: true }`:
+
+<!--tabs-->
+
+#### ❌ Incorrect
+
+```ts
+let foo: string | undefined;
+foo += 'some data';
+
+let bar: string = '';
+bar += 0;
+```
+
+#### ✅ Correct
+
+```ts
+let foo: number = 0;
+foo += 1;
+
+let bar = '';
+bar += 'test';
 ```
 
 ## When Not To Use It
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/restrict-plus-operands.ts
```diff
@@ -27,166 +28,204 @@ export default util.createRule<Options, MessageIds>({
...
-          return 'string';
-        }
-
-        if (types.some(value => value === 'number')) {
-          return 'number';
-        }
-
-        if (types.some(value => value === 'bigint')) {
-          return 'bigint';
-        }
```
JoshuaKGoldberg, line 105: These three inner `if (types.some(...` cases weren't covered by code coverage. Most of the intersections in the previous test suite reduced to `never`. I got rid of most of them.

packages/eslint-plugin/src/rules/restrict-plus-operands.ts
```diff
@@ -1,20 +1,21 @@
...
 import * as ts from 'typescript';
 
 import * as util from '../util';
 
 type Options = [
   {
-    checkCompoundAssignments?: boolean;
     allowAny?: boolean;
+    allowBoolean?: boolean;
+    allowNullish?: boolean;
```
bradzacher, line 11: [The nullish case will create TS errors for numbers, and is only valid in strings](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3gDwFzxQFcBbAIxBngB8DCIIBuAWAChXt4BqeARmZY7ds-QfADkY-q1CRYCZOiwBPfETIVq8QilAAzAJYoQwKSyVde-c9yVWLE06NusgA).
In the string case it also means you'll be concatenating with `'null' | 'undefined'` - which doesn't seem desirable in the vast majority of code.

I don't think that this option really makes sense in the context of the rule.

packages/eslint-plugin/src/rules/restrict-plus-operands.ts
```diff
@@ -1,20 +1,21 @@
...
+import * as tsutils from 'tsutils';
 import * as ts from 'typescript';
 
 import * as util from '../util';
 
 type Options = [
   {
-    checkCompoundAssignments?: boolean;
     allowAny?: boolean;
+    allowBoolean?: boolean;
```
bradzacher, line 10: Similar to the nullish case [the boolean case will create TS errors for numbers, and is only valid when adding with strings](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3gDwFzwCMkkIQoUBuAWAChbt4BqeARmpoee3c-gHIofdkA).
In the string case it also means you'll be concatenating with `'true' | 'false'` - which doesn't seem desirable in the vast majority of code.

I don't think that this option really makes sense in the context of the rule.

packages/eslint-plugin/src/rules/restrict-plus-operands.ts
```diff
@@ -1,20 +1,21 @@
...
+    allowNullish?: boolean;
+    allowNumberAndString?: boolean;
+    allowRegExp?: boolean;
```
bradzacher, line 13: I don't think this is a good option to add - what are the usecases for summing a regex and another value?
The only real case I could think of would be to do something like `/a/ + 'u'` - but ultimately that case is pretty flaky given that you can accomplish the same thing in a safer way with `new RegExp(/a/, 'u')`.

I don't think that this option really makes sense in the context of the rule.

packages/eslint-plugin/src/rules/restrict-plus-operands.ts
```diff
@@ -1,20 +1,21 @@
...
 
 import * as util from '../util';
 
 type Options = [
   {
-    checkCompoundAssignments?: boolean;
     allowAny?: boolean;
+    allowBoolean?: boolean;
+    allowNullish?: boolean;
+    allowNumberAndString?: boolean;
```
bradzacher, line 12: To me at least - this option seems like it circumvents the original intent of the rule - which is to enforce the two sides of the operator are the same type because TS doesn't enforce this.

TBH I don't know what the rule would check any more if you turned this option on - I think all of the remaining cases really would be TS errors.

bradzacher (dismissed):

JoshuaKGoldberg:
Drafting this pending discussion in #6110

packages/eslint-plugin/docs/rules/restrict-plus-operands.md
```diff
@@ -18,70 +18,145 @@ This rule reports when a `+` operation combines two values of different types, o
...
-var foo = parseInt('5.5', 10) + 10;
-var foo = 1n + 1n;
+let foo = parseInt('5.5', 10) + 10;
+let foo = 1n + 1n;
 ```
 
 ## Options
 
-### `checkCompoundAssignments`
+### `allowAny`
```
JoshuaKGoldberg, line 34: Per https://github.com/typescript-eslint/typescript-eslint/issues/6110#issuecomment-1481954671 - these new areas of options should have very explicit caveats & suggestions not to use them.

packages/eslint-plugin/docs/rules/restrict-plus-operands.md
```diff
@@ -18,70 +18,159 @@ This rule reports when a `+` operation combines two values of different types, o
...
+For example, instead of enabling the `allow*` options to concatenate strings and other values, use APIs akin to `console.log` that allow passing multiple values:
+
+```diff
+declare const value: boolean;
+
+- console.log("My value is: " + value);
++ console.log("My value is:", value;
+```
+
+:::
```
JoshuaKGoldberg, line 46: @bradzacher does this look like enough cautioning? Think there's anything else we should do to yell at people not to use these things?

packages/eslint-plugin/docs/rules/restrict-plus-operands.md
```diff
@@ -18,70 +18,159 @@ This rule reports when a `+` operation combines two values of different types, o
...
 ```
 
 ## Options
 
-### `checkCompoundAssignments`
+:::caution
+We generally recommend against using these options, as they limit which varieties of incorrect `+` usage can be checked.
 
-Examples of code for this rule with `{ checkCompoundAssignments: true }`:
+For example, instead of enabling the `allow*` options to concatenate strings and other values, use APIs akin to `console.log` that allow passing multiple values:
```
bradzacher, line 37: an alternative, more fleshed-out version with examples for consideration

```md
:::caution
We generally recommend against using these options, as they limit which varieties of incorrect `+` usage can be checked. This in turn severely limits the validation that the rule can do to ensure that resulting strings and numbers are correct.

Safer alternatives to using the `allow*` options include:
- Using `.toFixed()` to coerce numbers to well-formed string representations:
    ```ts
    const number = 1.123456789;
    const result = 'The number is ' + number.toFixed(2);
    // result === 'The number is 1.12'
    ```
- Calling `.toString()` on other types to mark explicit and intentional string coercion:
    ```ts
    const arg = '11';
    const regex = /[0-9]/;
    const result =
        'The result of ' + regex.toString() + '.test("' + arg + '") is ' + regex.test(arg).toString();
    // result === 'The result of /[0-9]/.test("11") is true'
    ```
- Using variadic forms of logging APIs to avoid needing to `+` values.
    ```ts
    // Remove this line
    console.log('The result is ' + true);
    // Add this line
    console.log('The result is', true);
    ```
```

<details>
<summary>Rendered</summary>

:::caution
We generally recommend against using these options, as they limit which varieties of incorrect `+` usage can be checked. This in turn severely limits the validation that the rule can do to ensure that resulting strings and numbers are correct.

Safer alternatives to using the `allow*` options include:
- Using `.toFixed()` to coerce numbers to well-formed string representations:
    ```ts
    const number = 1.123456789;
    const result = 'The number is ' + number.toFixed(2);
    // result === 'The number is 1.12'
    ```
- Calling `.toString()` on other types to mark explicit and intentional string coercion:
    ```ts
    const arg = '11';
    const regex = /[0-9]/;
    const result = 'The result of ' + regex.toString() + '.test("' + arg + '") is ' + regex.test(arg).toString();
    // result === 'The result of /[0-9]/.test("11") is true'
    ```
- Using variadic forms of logging APIs to avoid needing to `+` values.
    ```ts
    // Remove this line
    console.log('The result is ' + true);
    // Add this line
    console.log('The result is', true);
    ```

</details>

JoshuaKGoldberg: Love it! My only change suggestion is moving the `console.log` one to be first because it's the most common easy workaround I see in the wild.

PR #6162, opened 2022-12-03
chore: upgrade to yarn 3

### LINKED ISSUE #6150: Repo: Can't develop because of yarn1
### Suggestion

There are some issues with yarn1 that will never be fixed, such as not being able to run any commands with yarn1 when there are unicode characters in the windows username.
This pretty much means that these users (like me) can't build, debug and develop at all.

Migrating to yarn3 seems like a good approach, it is actively maintained.

JamesHenry:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6150 

## Overview

Migrates the codebase from yarn v1 to yarn v3, using the `node_modules` linker - **not** PnP - to ensure maximum consistency before and after the migration.

I confirmed that there were no issues with lerna publishing by publishing a canary version to verdaccio locally (this will not negatively impact the canary release post-merge):

image


packages/ast-spec/project.json
```diff
@@ -10,7 +10,7 @@
         "cwd": "packages/ast-spec",
         "commands": ["yarn build"]
       },
-      "outputs": ["packages/ast-spec/dist/**/*.ts"]
+      "outputs": ["{projectRoot}/dist/**/*.ts"]
```
JamesHenry, line 13: Unrelated to yarn 3

packages/shared-fixtures/project.json
```diff
@@ -1,6 +1,6 @@
 {
+  "name": "shared-fixtures",
```
JamesHenry, line 2: Unrelated to yarn 3

packages/website-eslint/project.json
```diff
@@ -1,6 +1,6 @@
 {
+  "name": "website-eslint",
```
JamesHenry, line 2: Unrelated to yarn 3

JamesHenry:
@bradzacher @JoshuaKGoldberg please have a play with this, it all looks good to me locally and CI is passing.

Once you are happy and approve the PR please **do not merge**, so that I can coordinate the update of the automation jobs

packages/ast-spec/project.json
JoshuaKGoldberg: Random thought: one could write Nx-specific lint rules for `project.json` files. There could be a lint rule that suggests replacing hardcoded paths like `"packages/ast-spec` with `"${projectRoot}"`.

packages/eslint-plugin-internal/package.json
```diff
@@ -18,5 +18,9 @@
     "@typescript-eslint/type-utils": "5.45.0",
     "@typescript-eslint/utils": "5.45.0",
     "prettier": "*"
+  },
+  "devDependencies": {
+    "jest": "*",
+    "rimraf": "*"
```
JoshuaKGoldberg, line 24: Kind of nice to have these more explicit now. +1

tools/postinstall.ts
```diff
@@ -1,9 +1,9 @@
 import * as execa from 'execa';
 
 /**
- * In certain circumstances we want to skip the below the steps and it may not always
- * be possible to use --ignore-scripts (e.g. if another tool is what is invoking the
- * install command, such as when nx migrate runs). We therefore use and env var for this.
+ * In certain circumstances we might want to skip the below the steps when another
+ * tool is invoking the install command (such as when nx migrate runs).
+ * We therefore use an env var for this.
  */
```
JoshuaKGoldberg, line 7: I ran `yarn dedupe` and it generated `21 insertions(+), 564 deletions(-)`. Maybe we can add `yarn dedupe` here too?

JoshuaKGoldberg (approved):
Looks great, thanks! 🔥

.gitignore
```diff
@@ -79,3 +79,12 @@ _ts3.4
...
 packages/types/src/generated/**/*.ts
+
+# Yarn 3
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
```
JoshuaKGoldberg, line 90: Can I suggest sending these in as a small PR now, before this one gets merged? That way we don't keep having 1k+ files left over when switching between this branch and others.

packages/ast-spec/project.json
JamesHenry: So this is actually the result of running `nx repair`, as special form of migration that can be run any time to correct any legacy configuration issues or things which are _definitely_ better as far as the tooling can figure out. So I think it's more a case of figuring out a good way to periodically run `nx repair`. I think my existing automated workflow for version migrations makes the most sense for that and I'll add it

.gitignore
JamesHenry: Done

tools/postinstall.ts
JamesHenry: We can't do it in the postinstall I don't think otherwise we end up in a loop when there are changes resulting from the dedupe.

I have pushed a dedicated commit with an alternative approach to ensuring that the lock file is maximally deduped over time: https://github.com/typescript-eslint/typescript-eslint/pull/6162/commits/18657c0f61619f1ef408ccd4c1ae095d55a239e1

packages/ast-spec/project.json
JamesHenry: Added https://github.com/typescript-eslint/typescript-eslint/pull/6162/commits/b549f4e233cc76b21e34368765441b3e41c68337

tools/postinstall.ts
```diff
@@ -1,9 +1,9 @@
...
 
 if (process.env.SKIP_POSTINSTALL) {
```
bradzacher, line 9: side note that we could convert this to use zx and it would be a little better than maintaining our own exec version

https://github.com/google/zx

tools/postinstall.ts
bradzacher: we can probably setup an action that runs weekly or something to generate a PR?
actually can we add it to the renovate config so it does it automatically? most of our dep upgrades come via dependabot anyways

bradzacher (dismissed):
25a

once you fix the CI I'm happy with this.
now i'll need to get used to the new console output from yarn sadge

tools/postinstall.ts
JamesHenry: > we can probably setup an action that runs weekly or something to generate a PR?

Hmm yeah that's probably going to be a lot less disruptive to contributors. Will be annoying to put together a PR and have it fail if you didn't prune recently... I'll revert that change and we can do a prune workflow in a follow up

> most of our dep upgrades come via dependabot anyways

@bradzacher isn't it the opposite? We switched from dependabot to renovate, right?

bradzacher: yes, that's what I meant! I totally said both on purpose to cover my bases

JamesHenry: (PS I've removed the call to `yarn dedupe` from ci.yml in this PR now so that we don't cause issues for users unnecessarily per the above comment)

bradzacher (approved):

tools/postinstall.ts
bradzacher: See how forward thinking i am? I think so far ahead I forget how far ahead I think.

bigbraintime

bradzacher (approved):

liuxingbaoyu:
Thanks for all this!

jest.config.base.js
```diff
@@ -18,7 +20,9 @@ module.exports = {
     'json',
     'node',
   ],
-  setupFilesAfterEnv: ['console-fail-test/setup.js'],
+  setupFilesAfterEnv: [
+    join(__dirname, 'node_modules/console-fail-test/setup.js'),
```
fisker, line 24: `<rootDir>/node_modules/console-fail-test/setup.js` should work.

bradzacher: I think the issue with using `<rootDir>` is that it's dependent where the `jest.config.js` is.
If you `cd` into one of the packages, each package folder has its own `jest.config.js`, so that means that in a package `<rootDir>` will mean `<root>/packages/foo`, not just `<root>` - i.e. the path will not exist.

By using `_dirname` it means this path is always relative to this root config file.

package.json
```diff
@@ -60,7 +60,6 @@
     "@babel/eslint-parser": "^7.22.7",
     "@babel/parser": "^7.22.7",
     "@babel/types": "^7.22.5",
-    "@nrwl/nx-cloud": "16.1.1",
```
JamesHenry, line 63: This was outdated, the package is `nx-cloud` now which we already have specified

package.json
JamesHenry, line 1: Yarn automatically reordered these deps (apart from the `@nrwl/nx-cloud` one, see note)

packages/website-eslint/package.json
```diff
@@ -31,13 +31,15 @@
     "@typescript-eslint/eslint-plugin": "6.2.0",
     "@typescript-eslint/parser": "6.2.0",
     "@typescript-eslint/scope-manager": "6.2.0",
-    "@typescript-eslint/types": "6.0.0",
```
JamesHenry, line 34: These packages had been duplicated with the `dependencies` and were not being updated, causing issues. Just needed to remove them

packages/type-utils/package.json
```diff
@@ -52,7 +52,11 @@
   },
   "devDependencies": {
     "@typescript-eslint/parser": "6.2.0",
-    "ajv": "^8.12.0",
```
JamesHenry, line 55: I had to downlevel this to match the rest of the workspace, otherwise `patch-package` hard errors on it

packages/utils/package.json
```diff
@@ -78,6 +78,10 @@
   },
   "devDependencies": {
     "@typescript-eslint/parser": "6.2.0",
+    "downlevel-dts": "*",
+    "jest": "29.6.1",
```
JamesHenry, line 82: Having to match all these `jest` versions is pretty cumbersome, but required to make the naked `jest` calls in npm scripts work with modern yarn.

The way they recommend handling this is different and involves sharing scripts between workspaces, or using `run -T` in front of the script contents.

Somewhat related - we could avoid the need for this entirely if we always let Nx run the tasks through its own executors and then just switch the script contents to be `nx test` like we did here: https://github.com/typescript-eslint/typescript-eslint/blob/ed8c6247c0454f84b126f5fbe16fe1c40f6605f3/packages/scope-manager/package.json#L43 - it would resolve `jest` for us, and we wouldn't need to bend over backwards to make yarn happy

packages/eslint-plugin/package.json
```diff
@@ -76,15 +76,19 @@
     "@types/prettier": "*",
     "@typescript-eslint/rule-schema-to-typescript-types": "6.2.0",
     "@typescript-eslint/rule-tester": "6.2.0",
-    "ajv": "^6.12.6",
+    "ajv": "^6.10.0",
```
JamesHenry, line 79: Changed to match other references

.github/workflows/nx-migrate.yml
```diff
@@ -72,6 +72,9 @@ jobs:
             rm migrations.json
           fi
 
+          # Run the special nx repair command to ensure config matches latest and greatest
+          npx nx repair
```
JoshuaKGoldberg, line 76: wat

https://github.com/nrwl/nx/issues/18385

JoshuaKGoldberg: (not a blocker, I just am curious what this actually does)

netlify.toml
```diff
@@ -5,7 +5,9 @@
   command = "NX_VERBOSE_LOGGING=true yarn patch-package && yarn nx build website"
 [build.environment]
   NETLIFY_USE_YARN = "true"
-  YARN_FLAGS = "--ignore-scripts"
+  # TODO: adjust these once https://github.com/netlify/build-image/issues/612 is resolved
```
JoshuaKGoldberg, line 8: That awkward moment when the upstream repo is archived without resolving the issue... 🤷

packages/utils/package.json
JoshuaKGoldberg: That sounds good to me! IIRC you'd mentioned this as a potential followup a while back. I'm definitely in favor.

JoshuaKGoldberg (changes requested):
This looks great! I read through and think I understand the strategies and upstream-blocked-todo-workarounds. Also I tried it out locally and -with the caveat I commented in `package.json`- it seems to work great on my end. Nice! 👏 

Only requesting changes because I think we'll need to update the local development docs.

JamesHenry:
@JoshuaKGoldberg Brought it back up to date and updated the local development guide. Hopefully we can merge now!

JoshuaKGoldberg (approved):
Tried it out locally again and it all works great! No need for me to manually run `.cjs` files from the repo. Thanks! Awesome stuff @JamesHenry 🚀

PR #6165, opened 2022-12-04
chore(eslint-plugin): valid typescript error code in eslint-recommended

nix6839:
## Overview

Fix incorrect TS error codes in comments.
image


JoshuaKGoldberg (approved):
Heh, yeah, these codes aren't guaranteed to be the same across TS versions. Thanks!

Issue #5845, opened 2022-10-18
Docs: Explain situation around no-unsafe-* rules triggering on file changes in editors


JoshuaKGoldberg:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/docs/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

A lot of users have reported issues like this one: https://twitter.com/TkDodo/status/1582335847017242625

> Every time I change something in my [@trpcio](https://twitter.com/trpcio/) router, [@tseslint](https://twitter.com/tseslint/) starts to complain on the usage sides in my editor ([@intellijidea](https://twitter.com/intellijidea/)):
> 
> "Unsafe member access on an `any` value.(no-unsafe-member-access)"
> 
> Disabling and re-enabling eslint solves this. Any ideas what that could be?

I've personally felt this in VS Code a lot.

We should document this in our FAQs for when it happens to users:
* That the current / bandaid solution is to restart your ESLint server (along with how)
* What the structural underpinnings are (with links to pending issues)

### Affected URL(s)

https://typescript-eslint.io/docs/linting/troubleshooting

bradzacher:
Here is a dump - I don't have time to create a PR or clean this up right now.
I couldn't figure out the best order to put everything in so - todo.

Some of this is probably TMI or irrelevant to the question, but whatever - I dumped a bunch.
It's not everything, but it's a good start (hopefully).

---

ESLint is designed to be a stateless, single-file linter. It and the ecosystem of "API consumers" (tools that build on top of their API - IDEs, CLI tools, etc) assume this to be true and optimise based on the assumption. For most parsers (`@babel/eslint-parser`, `vue-eslint-parser`, etc) this holds true - they parse a file and forget about it, and for our parser (`@typescript-eslint/parser`) in non-type aware mode this also holds true. However when instructed to use type information, our parser now breaks both assumptions - it now stores stateful, cross-file information.

Type-aware linting, unfortunately, doesn't fit too well into the ESLint model as it's currently designed - so we've had to implement a number of workarounds to make it fit - we've fit a square peg into a round hole by cutting the edges of the hole. This, as you can imagine, means there are a number of edge-cases where things can get funky.

[...]

PR #6172, opened 2022-12-06
feat(typescript-estree): add experimental mode for type-aware linting that uses a language service instead of a builder 

bradzacher:
## Overview

This is just an experiment to help address memory issues (#1192), maybe performance, and maybe out-of-sync types (#5845).

I was looking into some code around the place and noticed that TS exposes the concept of a "document registry" - which is a shared cache that can be reused across certain TS data structures to deduplicate memory usage (and I assume improve performance by deduplicating work).

This PR implements a new parser strategy which uses TS's "Language Service" tooling, which in turn leverages the document registry. "persistent parse" tests pass - which at least proves that it works in some manner of speaking.

One interesting thing to note here is under the hood the language serivce doesn't use a builder program. I believe the idea is that the document registry is supposed to forego the performance implications of that? I don't know exactly - it will require more testing. Though it's worth mentioning that this means this could replace our current "single-run" codepaths because it doesn't use a builder program.

TODO:
- figure out how to roll-back "dirty" states
- memory pressure testing
- runtime performance testing

### DIFF
--- a/packages/typescript-estree/src/create-program/createProjectProgram.ts
+++ b/packages/typescript-estree/src/create-program/createProjectProgram.ts
@@ -4,8 +4,7 @@ import * as ts from 'typescript';
 
 import { firstDefined } from '../node-utils';
 import type { ParseSettings } from '../parseSettings';
-import { getWatchProgramsForProjects } from './getWatchProgramsForProjects';
-import type { ASTAndProgram } from './shared';
+import type { ASTAndProgram, CanonicalPath } from './shared';
 import { getAstFromProgram } from './shared';
 
 const log = debug('typescript-eslint:typescript-estree:createProjectProgram');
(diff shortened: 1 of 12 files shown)


bradzacher:
something weird is happening on the CI - dunno why it's OOMing.
cannot reproduce locally so that's fun.

bradzacher:
Cc @jakebailey I know you had an inclination about helping us improve the perf of the tooling.

I was hoping you could help me by providing some TS insider insight ☺

Do you think this approach will help out a lot compared to the existing builder program approach?
From what I can tell by poking around and reading it'll at least save us a bunch of memory - though not too sure about runtime cost/savings yet.

Would it be worth us building our own layer on top of the document registry instead of reusing the language service?

jakebailey:
Honestly, I have no clue; @sheetalkamat or someone may have better insight.

JoshuaKGoldberg:
Closing this old draft for housekeeping since there are merge conflicts and it's taking up space in the open PRs list. Nothing bad will happen in my housekeeping if this is re-opened. Don't mind me. 😄

JoshuaKGoldberg:
@bradzacher do you still want to work on this now that `EXPERIMENTAL_useProjectService` is in (#6754)?

bradzacher:
Let's put a pin in this - if we find that the project service works for all cases then there's no need for us to worry about improving the manual config case, I think.
We may want to re-investigate this if there are strong usecases that it doesn't support - but I'm optimistic we won't have to.

Issue #6163, opened 2022-12-03
Repo: Build break in local development on main


JoshuaKGoldberg:
### Suggestion

🙃 

```plaintext
    ✖  nx run parser:build
$ tsc -b tsconfig.build.json
       src/parser.ts(108,5): error TS2322: Type '{ ecmaVersion: number | undefined; globalReturn: boolean | undefined; jsxPragma: string | null | undefined; jsxFragmentName: string | null | undefined; lib: Lib[] | undefined; sourceType: "script" | "module"; }' is not assignable to type 'AnalyzeOptions'.
         Object literal may only specify known properties, and 'ecmaVersion' does not exist in type 'AnalyzeOptions'.
```

```plaintext
josh ~/repos/typescript-eslint/packages/parser $ yarn build   
yarn run v1.22.19
$ tsc -b tsconfig.build.json
src/parser.ts:108:5 - error TS2322: Type '{ ecmaVersion: number | undefined; globalReturn: boolean | undefined; jsxPragma: string | null | undefined; jsxFragmentName: string | null | undefined; lib: Lib[] | undefined; sourceType: "script" | "module"; }' is not assignable to type 'AnalyzeOptions'.
  Object literal may only specify known properties, and 'ecmaVersion' does not exist in type 'AnalyzeOptions'.

108     ecmaVersion: options.ecmaVersion === 'latest' ? 1e8 : options.ecmaVersion,
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

JoshuaKGoldberg:
I think the issue is `scope-manager`. `packages/scope-manager/dist/analyze.d.ts` is missing `ecmaVersion for me. Running `yarn clean` and then `yarn build` within the `scope-manager` package directory fixes it.

JoshuaKGoldberg:
@JamesHenry I'm stumped here 😞 could you help me out please? 

I don't see why the outputs aren't caching properly, given that the `build` command knows it should consider the `dist/` directory as output?

https://github.com/typescript-eslint/typescript-eslint/blob/891a77eb49c3e2f520f647e4816f30ee6a6e960d/packages/scope-manager/project.json#L9

PR #6178, opened 2022-12-07
chore: drafted ESLint parsing context

JoshuaKGoldberg:
Reference of how we might be able to use a _"parsing context"_ if given one from ESLint.

Specifically, `inferSingleRun` becomes much smarter - we can _know_ (instead of guess) whether to enable single-run mode if given that parsing context.

### DIFF
--- a/packages/visitor-keys/src/visitor-keys.ts
+++ b/packages/visitor-keys/src/visitor-keys.ts
@@ -150,7 +150,7 @@ const SharedVisitorKeys = (() => {
 })();
 
 const additionalKeys: AdditionalKeys = {
-  AccessorProperty: SharedVisitorKeys.PropertyDefinition,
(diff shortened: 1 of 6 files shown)


packages/visitor-keys/src/visitor-keys.ts
```diff
@@ -177,7 +177,7 @@ const additionalKeys: AdditionalKeys = {
   RestElement: ['decorators', 'argument', 'typeAnnotation'],
   StaticBlock: ['body'],
   TaggedTemplateExpression: ['tag', 'typeParameters', 'quasi'],
-  TSAbstractAccessorProperty: SharedVisitorKeys.AbstractPropertyDefinition,
+  // TSAbstractAccessorProperty: SharedVisitorKeys.AbstractPropertyDefinition,
```
JoshuaKGoldberg, line 180: (just commented out to avoid some build shenanigans - #6163)

PR #6180, opened 2022-12-08
docs: fixed typo  "foo.property"

TopherLee513:
Change "foo.property" to "example.property"


TopherLee513:
To clarify...
Current Demo Code :
```markdown
interface Example {
  property?: string;
}
declare const example: Example;
const includesBaz = foo.property!.includes('baz');
```
Either 'example' should be 'foo' or 'foo' should be 'example'. Correct?

bradzacher (approved):
thanks for this!

PR #6195, opened 2022-12-09
feat(eslint-plugin): [block-spacing] extending base rule for TS related blocks

### LINKED ISSUE #5990: Base rule extension: block-spacing Extending base rule for TS related blocks

### Link to the base rule

https://eslint.org/docs/latest/rules/block-spacing

### Description

I believe the purpose of the base rule, applies for TS too

### Fail

```typescript
interface Shape {x: number}
interface Shape { x: number}
interface Shape {x: number }
interface Shape {x: number; y: number}
```

### Pass

```typescript
interface Shape { x: number }
interface Shape { x: number; y: number }
```

### Additional Info

Actually, having a rule that rejects single line `interface` declarations could work too

omril1:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5990


packages/eslint-plugin/src/rules/block-spacing.ts
```diff
@@ -0,0 +1,166 @@
...
+     */
+    function getOpenBrace(
+      node: TSESTree.TSEnumDeclaration,
+    ): TSESTree.PunctuatorToken {
+      // guaranteed for enums
+      // This is the only change made here from the base rule
+      return sourceCode.getFirstToken(node, {
+        filter: token =>
+          token.type === AST_TOKEN_TYPES.Punctuator && token.value == '{',
+      }) as TSESTree.PunctuatorToken;
```
omril1, line 46: The only significant change from the base rule, it worked flawlessly for interface and object type literal.
Added a filter for punctuator because enums don't have a body node in the AST

https://github.com/eslint/eslint/blob/3ae0574fc78c4a2b406625e4792cb2859cb9bcb1/lib/rules/block-spacing.js#L60

packages/eslint-plugin/src/rules/block-spacing.ts
```diff
@@ -0,0 +1,166 @@
...
+    return {
+      ...baseRules,
+
+      // This code worked "out of the box" for interface and type literal
+      // Enums were very close to match as well, the only reason they it not is that was that enums don't have a body node in the parser
+      // So the opening brace punctuator starts in the middle of the node
+      // `getFirstToken` in the base rule did not filter for the first opening brace punctuator
+      TSInterfaceBody: baseRules.BlockStatement as never,
+      TSTypeLiteral: baseRules.BlockStatement as never,
+      TSEnumDeclaration: checkSpacingInsideBraces,
```
omril1, line 163: Kept the base rule for interface and type literal, though changing to `checkSpacingInsideBraces` will work as well

packages/eslint-plugin/src/rules/block-spacing.ts
```diff
@@ -0,0 +1,166 @@
...
+      return (
+        !util.isTokenOnSameLine(left, right) ||
+        sourceCode.isSpaceBetween!(left, right) === always
+      );
+    }
+
+    /**
+     * Checks and reports invalid spacing style inside braces.
+     * @param {ASTNode} node A BlockStatement/StaticBlock/SwitchStatement node to check.
+     * @returns
```
JoshuaKGoldberg, line 70: Nit: we have types in TypeScript. `ASTNode` is imprecise and the `@returns` is implied.

```suggestion
     * @param node A BlockStatement/StaticBlock/SwitchStatement node to check.
```

Seems like these were copy & pasted from https://github.com/eslint/eslint/blob/ba74253e8bd63e9e163bbee0540031be77e39253/lib/rules/block-spacing.js#L86? I don't see other base rule files giving explicit attribution, so I guess this should be fine as-is.

packages/eslint-plugin/src/rules/block-spacing.ts
```diff
@@ -0,0 +1,166 @@
...
+        closeBrace.value !== '}' ||
+        firstToken === closeBrace
+      ) {
+        return;
+      }
+
+      // Skip line comments for option never
+      if (!always && firstToken.type === AST_TOKEN_TYPES.Line) {
+        return;
+      }
```
JoshuaKGoldberg, line 97: Indeed. If this isn't covered by tests, then either:

* We need more tests
* This is unreachable and can be removed.

packages/eslint-plugin/src/rules/block-spacing.ts
```diff
@@ -0,0 +1,166 @@
...
+
+      // Check.
```
JoshuaKGoldberg, line 99: Nit: the comment is a little redundant, I think.

```suggestion
```

packages/eslint-plugin/tests/rules/block-spacing.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  parser: '@typescript-eslint/parser',
+});
+
+type InvalidBlockSpacingTestCase = InvalidTestCase<
+  'missing' | 'extra',
+  ['always' | 'never']
+>;
+
+const options = ['always', 'never'] as const;
+const typeDeclaration = [
```
JoshuaKGoldberg, line 17: Nit: other arrays are named as the plural. Let's keep it consistent for this one.

```suggestion
const typeDeclarations = [
```

JoshuaKGoldberg (changes requested):
Mostly looks great to me! Just some touchups around untested code, and a few nitpicks. Thanks for sending this in & working on it! ✨

packages/eslint-plugin/src/rules/block-spacing.ts
omril1: I saw other rules that copied from the base rule, so I assumed it's ok, do you suggest only removing the types from the jsdoc?

JoshuaKGoldberg: Yeah there's no need for the jsdoc.

omril1:
Sorry for the force push, something in vscode messed up my commits again

packages/eslint-plugin/src/rules/block-spacing.ts
omril1: Added more tests, and now this is covered

JoshuaKGoldberg (dismissed):
Looks great to me, thanks for sending this in!

Because this is an extension of a base rule, I'd feel more comfortable getting a review from another maintainer. Just to be safe 😄 .

packages/eslint-plugin/src/rules/block-spacing.ts
JoshuaKGoldberg: Ok just following up from internal discussions! Still no need to take action here. Just posting as a reference...

Some of our extension rules include code that's very similar to the code from the base ESLint rule. Sometimes there are entire lines or even functions that look the same.

[ESLint is licensed under the MIT License](https://github.com/eslint/eslint/blob/b401cde47d44746ff91b8feced3fb3a4e32c0e12/LICENSE). MIT stipulates _(emphasis mine)_:

> Permission is hereby granted, free of charge, to any person obtaining a copy
of this software and associated documentation files (the "Software"), to deal
in the Software without restriction, including without limitation the rights
to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
copies of the Software, and to permit persons to whom the Software is
furnished to do so, subject to the following conditions:
> 
> The above copyright notice and this permission notice shall be included in
all **copies or substantial portions of the Software**.

We're not legal experts -let alone copyright lawyers- but we believe we're not in violation of this because:

* Our code does not copy substantial portions. At most, the _ideas_ of the original code are allowed in as reference, but we use our own code style & standards in the implementation.
* We link to the original ESLint source quite often in our site & packages, including in the individual docs pages for rules.

_If anybody knows my statements to be wrong and/or that we need to do something else, please let us know!_

bradzacher (approved):

PR #6211, opened 2022-12-14
feat(eslint-plugin): add `key-spacing` rule extension for interface & type declarations

### LINKED ISSUE #940: [key-spacing] key-spacing of interfaces and other TS-specific key-value pairs
**Repro**

```JSON
{
  "rules": {
    "key-spacing": ["error", { "align": "value" }]
  }
}
```

```TS
interface Foo {
  foo: string;
  barBazQuux: string;
}
```

**Expected Result**

Should be highlighted as an error and auto fixed into:

```TS
interface Foo {
  foo:        string;
  barBazQuux: string;
}
```

**Actual Result**

ESLint ignores interfaces.

**Additional Info**

Time for `@typescript-eslint/key-spacing`?

**Versions**

| `@typescript-eslint/eslint-plugin` | `2.0.0` |
| `@typescript-eslint/parser`        | `2.0.0` |
| `TypeScript`                       | `3.6.2` |
| `ESLint`                           | `5.16.0` |
| `node`                             | `12.9.0` |
| `yarn`                              | `1.17.3` |

coyotte508:
## PR Checklist

- [x] Addresses an existing open issue: fixes #940

## Overview

Make `key-spacing` work with interfaces & type declarations. ~🚧 **Work in progress** 🚧.~

Happy to take any and all feedback & help. ~Do I need to support all options & sub-options from the base rule? :sweat:~ _edit by @JoshuaKGoldberg: summarizing discussion, yes_ 😄 

## References

eslint rule: https://eslint.org/docs/latest/rules/key-spacing
eslint rule source: https://github.com/eslint/eslint/blob/main/lib/rules/key-spacing.js#L491

## Todo

- [x] add more tests
- [x] add autofix
- [x] add doc


JoshuaKGoldberg:
> Do I need to support all options & sub-options from the base rule? 😓

Yup! 😁

coyotte508:
The base implementation is done - I had to hack around a bit for index signatures: `{ [x: number]: string}`, I found no way to get the position of the `]` other than to check the text value of the whole node.

Key alignment is done for interface / types / classes properties & index signatures.

I still need to do the autofixer & the docs. And probably test more.

JoshuaKGoldberg:
Don't mind me - just marking as draft and unsubscribing so I don't get notified on pushes. Happy to review once it's ready! Just undraft. 🙂

coyotte508:
The PR is ready -  has been ready since yesterday, even though I marked it ready a bit too early.

The commit about the edge case I thought of last night, and the perf commit is purely optional. Sorry about this, I'll stop committing unless I find an error :)

julien-c:
Hi @JoshuaKGoldberg 👋 

Happy to report that @huggingface is now sponsoring `typescript-eslint` on opencollective 🎉 

We've been running this PR's code on our codebase for ~1 month so it should be quite solid already. Would love to have it land on main! 🙏

JoshuaKGoldberg:
Ah amazing - thanks for the sponsorship @julien-c! ❤️

We've been a little swamped lately with December holidays & the new year - but I'm hopeful to be able to review this PR very soon, _likely_ within a 2-3 business days. Thanks!

packages/eslint-plugin/tests/rules/key-spacing.test.ts
```diff
@@ -0,0 +1,428 @@
...
+      options: [
+        {
+          singleLine: { beforeColon: false, afterColon: true },
+          multiLine: { beforeColon: true, afterColon: true },
+        },
+      ],
+      errors: [{ messageId: 'extraKey' }, { messageId: 'extraKey' }],
+    },
+  ],
+});
```
JoshuaKGoldberg, line 428: Missing test coverage: wild and wacky types 😄. There should be test cases exercising...

* Mapped types
* Nested types (`outer: { middle: { inner: { /* ... */ } /* ... */ } }`)
* ...maybe other wild and wacky things, if you can think of them?

packages/eslint-plugin/src/rules/key-spacing.ts
```diff
@@ -0,0 +1,404 @@
...
+          }
+
+          prevNode = node;
+        }
+
+        unalignedElements = unalignedElements.concat(
+          alignGroups
+            .filter(group => group.length === 1)
+            .flatMap(group => group),
+        );
```
JoshuaKGoldberg, line 383: [Non-blocking] Just throwing it out there, you could do this to save a line...

```suggestion
        unalignedElements = unalignedElements.concat(
          ...alignGroups.filter(group => group.length === 1),
        );
```

If you have even the _slightest_ preference for the way it is now, feel free to ignore me. Just thought I'd mention the slightly smaller alternative! 😄

packages/eslint-plugin/tests/rules/key-spacing.test.ts
```diff
@@ -0,0 +1,428 @@
...
+    {
+      code: 'class X {\n  a:   number;\n\n  abc     : string\n};',
+      output: 'class X {\n  a: number;\n\n  abc: string\n};',
+      options: [{ align: 'value' }],
+      errors: [{ messageId: 'extraValue' }, { messageId: 'extraKey' }],
+    },
+    {
+      code: 'interface X {\n  a:   number;\n  // Some comment\n\n  // interrupted in the middle\n  abc: string\n};',
+      output:
+        'interface X {\n  a: number;\n  // Some comment\n\n  // interrupted in the middle\n  abc: string\n};',
```
JoshuaKGoldberg, line 251: Missing test case: multiline `/*` comments?

JoshuaKGoldberg (changes requested):
OK! Thanks for sending this in & waiting @coyotte508, and sorry again for the delay. I think the code looks pretty reasonable but for some small cleanup/refactor requests I've left inline. Mainly requesting changes here for more thorough testing of edge cases and different options. Hopefully they don't make the code too much more lots-of-different-cases-y?

I hope that the painstaking annoying edge cases you're working through in this PR have at least helped emphasize [why we strongly recommend nobody use `key-spacing` or other formatting rules`](https://typescript-eslint.io/linting/troubleshooting/formatting). 😄

coyotte508:
Thanks for the review! Marking the PR as draft, will progressively fix & mark as ready when it's ready!

packages/eslint-plugin/src/rules/key-spacing.ts
```diff
@@ -0,0 +1,404 @@
...
+        'key' in node ? node.key : node.parameters.at(-1)!,
+      ).loc.end;
+    }
+
+    function checkBeforeColon(
+      node: KeyTypeNode,
+      nBeforeColon: number,
+      mode: 'strict' | 'minimum',
+    ): void {
+      const colon = node.typeAnnotation!.loc.start.column;
```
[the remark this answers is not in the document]
coyotte508: We could use [`type-fest`](https://github.com/sindresorhus/type-fest) and `SetRequired`, it's a dependency that's already in the dependency tree of the project, and it's only typing utilities, no JS code generated. 

Ok either way - let me know what you think is better.

packages/eslint-plugin/tests/rules/key-spacing.test.ts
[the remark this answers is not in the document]
coyotte508: Will add `grapheme-splitter` as a dependency. `Intl.Segmenter` is builtin but only works for Node 16+

coyotte508: There are already tests for mapped types (eg checks for `IndexSignature`), and nested types although only 1-deep.

Will try to make a wacky example though :)

packages/eslint-plugin/src/rules/key-spacing.ts
```diff
@@ -0,0 +1,404 @@
...
+      const colonToken = sourceCode.getTokenAfter(node, util.isColonToken)!;
+
+      return sourceCode.getTokenBefore(colonToken)!;
+    }
+
+    /**
+     * Relevant nodes to our rule
+     *
+     * node.typeAnnotation will always be defined, but no way to enforce that and keep
+     * the type compatible with TSEstree.Node (except maybe TS 4.9' "satisfies" keyword)
```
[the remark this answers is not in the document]
coyotte508: I added the test case in [`ae33b9b` (#6211)](https://github.com/typescript-eslint/typescript-eslint/pull/6211/commits/ae33b9b67c6a2a689c2bc3881ab2fc18b12c15fc):

```ts
let x: {
  a: number;
  abc: string;
}
```

Is that an `ImplicitAnyInterface`? Otherwise, can you give me a code sample for what you mean?

packages/eslint-plugin/src/rules/key-spacing.ts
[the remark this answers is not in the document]
coyotte508: Sorry - it's coming back to me now

> no way to enforce that and keep the type compatible with TSEstree.Node (except maybe TS 4.9' "satisfies" keyword)

I had tried to make it required in the `KeyTypeNode` type, but I encountered a lot of issues. I think making a property required on a union is messing up with typescript.

I'll refacto a bit to make it more readable but I don't think I can avoid it

coyotte508:
> Mainly requesting changes here for more thorough testing of edge cases and different options. Hopefully they don't make the code too much more lots-of-different-cases-y?

Just [a few lines to handle unicode characters](https://github.com/typescript-eslint/typescript-eslint/pull/6211/commits/301731b3e645532be07255fad48c538eef9d0f4c) + a module that can be replaced by `Intl.Segmenter` when Node 14 stops being supported.

No if/else in added in the rule's logic :)
 
> I hope that the painstaking annoying edge cases you're working through in this PR have at least helped emphasize [why we strongly recommend nobody use `key-spacing` or other formatting rules`](https://typescript-eslint.io/linting/troubleshooting/formatting). smile

Can't say I disagree, if it was up to me I'd use prettier and call it a day :shushing_face: 

Still, this rule grows on you :shrug: 

---

I couldn't really address the `MakeRequired` comment, due to Typescript limitations I think. I marked the other comments as resolved, not sure it's the etiquette for this project or if the initial commenter should be the one to, feel free to reopen them in any case!

packages/eslint-plugin/src/rules/key-spacing.ts
```diff
@@ -1,9 +1,24 @@
...
+}
+function getStringLength(value: string): number {
+  if (isASCII(value)) {
+    return value.length;
+  }
+
+  splitter ??= new GraphemeSplitter();
+
+  return splitter.countGraphemes(value);
+}
```
coyotte508, line 20: Used https://github.com/typescript-eslint/typescript-eslint/pull/5704/files#diff-a37eec5bd9168ca642bc7a0d6656843f5eff07c54b0fc32f8df0eb420c85ff62R6

JoshuaKGoldberg:
> not sure it's the etiquette for this project or if the initial commenter should be the one to

Tangent: I miss working in a review tool that differentiated between _mark as resolved_ (code author's responsibility) and _close_ (reviewer's responsibility). Alas. +1 to this being a not-clear thing on GitHub!

packages/eslint-plugin/src/rules/key-spacing.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Huh, yeah, `MakeRequired` isn't _quite_ it. I don't really have the time or energy to dig into which nuances of optionality are applicable here 😂 so I just pushed the much simpler 2092487b. It'd be nice to make a more standard helper but that's just a small cleanup for eventually/later.

JoshuaKGoldberg (dismissed):
Looks great, thanks!

I do want to poke @bradzacher to make sure I'm not missing some dependency shenanigan with `grapheme-splitter` (same as #5704). But otherwise looks wonderful, thanks!

packages/eslint-plugin/src/rules/key-spacing.ts
```diff
@@ -68,18 +68,18 @@ export default util.createRule<Options, MessageIds>({
...
-     */
     type KeyTypeNode =
       | TSESTree.TSIndexSignature
       | TSESTree.TSPropertySignature
       | TSESTree.PropertyDefinition;
 
-    function isKeyTypeNode(node: TSESTree.Node): node is KeyTypeNode {
+    type KeyTypeNodeWithTypeAnnotation = KeyTypeNode & {
+      typeAnnotation: TSESTree.TSTypeAnnotation;
+    };
```
coyotte508, line 78: Nice!

JoshuaKGoldberg (approved):
:shipit:

packages/eslint-plugin/package.json
```diff
@@ -50,6 +50,7 @@
     "debug": "^4.3.4",
     "grapheme-splitter": "^1.0.4",
     "ignore": "^5.2.0",
+    "grapheme-splitter": "^1.0.4",
```
bradzacher, line 53: looks like a bad merge - added the entry twice!

JoshuaKGoldberg: Eek.

JoshuaKGoldberg: https://github.com/typescript-eslint/typescript-eslint/commit/99c091e0e3cf38a28aa5e1d1d2b03b602d51aa0a#r98752909

PR #6212, opened 2022-12-14
chore: run website tests against netlify deployment

JamesHenry:
Rather than running an extra dev-server on localhost within the Github agent, which has proven to be flaky, this runs the website tests against the actual Netlify deployment which we are doing anyway as part of our workflow.

In my initial implementation I had a very convoluted way of achieving it via Netlify build hooks and a webhook handler which triggered an additional workflow.

It has proven to be much simpler by leveraging the Netlify API instead (which was already done by an existing Github Action from the community)

### DIFF
--- a/packages/website/playwright.config.ts
+++ b/packages/website/playwright.config.ts
@@ -8,7 +8,7 @@ const config: PlaywrightTestConfig = {
   retries: 0,
   testDir: './tests',
   use: {
-    baseURL: process.env.PLAYWRIGHT_TEST_BASE_URL || 'http://localhost:3000',
+    baseURL: process.env.PLAYWRIGHT_TEST_BASE_URL ?? 'http://localhost:3000',
     trace: 'on-first-retry',
   },
   projects: [


JamesHenry:
Oh Netlify still doesn't dispatch `deployment_status` events even many years after it has been available and has been requested. That's very surprising.

We can potentially leverage an HTTP trigger for this though

packages/website/tests/index.spec.ts
```diff
@@ -15,8 +15,6 @@ test.describe('Website', () => {
       }
     });
     await page.goto('/');
-    expect(errorMessages).toStrictEqual([
-      "[error] Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot",
```
JamesHenry, line 19: This error message was only applicable to the local dev-server and is not shown to our users on production builds. A nice extra benefit of this approach is that our testing is more representative of our end UX

JamesHenry:
The remaining website test failure is consistent with the same failure on the latest main run, so it is not related to the changes

.github/workflows/ci.yml
```diff
@@ -179,26 +179,38 @@ jobs:
       contents: read # to fetch code (actions/checkout)
 
     name: Website tests
+    # We technically do not need to wait for build within the pipeline any more because the build we care about is happening within Netlify, however,
+    # it is highly likely that if the CI one fails, the Netlify one will as well, so in order to not waste unncessary Github Actions minutes/resources,
+    # we do still keep this requirement here.
```
JoshuaKGoldberg, line 184: _chuckles in #5573_

JoshuaKGoldberg (approved):
⚡ yess!!!

PR #6228, opened 2022-12-16
chore: enable eqeqeq internally

### LINKED ISSUE #6227: Repo: Enable ESLint's eqeqeq internally
### Suggestion

This comes up once in a while. I was surprised to see it's not enabled in any ruleset already.

https://eslint.org/docs/latest/rules/eqeqeq

> ```js
> ["error", "always", {"null": "ignore"}]
> ```

Example of where this has come up: https://github.com/typescript-eslint/typescript-eslint/pull/6091#discussion_r1050381489

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6227

## Overview

Enables the rule, allowing `== null` and `!= null` per our existing code style.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -110,7 +110,7 @@ module.exports = {
       'error',
       'always',
       {
-        null: 'ignore',
+        null: 'never',
       },
     ],
     'no-mixed-operators': 'error',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -106,6 +106,13 @@ module.exports = {
     //
 
     curly: ['error', 'all'],
+    eqeqeq: [
+      'error',
+      'always',
+      {
+        null: 'ignore',
+      },
+    ],
     'no-mixed-operators': 'error',
     'no-console': 'error',
     'no-process-exit': 'error',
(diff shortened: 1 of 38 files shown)


.eslintrc.js
```diff
@@ -106,6 +106,13 @@ module.exports = {
     //
 
     curly: ['error', 'all'],
+    eqeqeq: [
+      'error',
+      'always',
+      {
+        null: 'ignore',
```
bradzacher, line 113: Instead of ignore should we should enforce one way or the other.

I've mentioned that I personally think that using non-strict equality is best and never ever write the undefined variable. I don't think there are any cases in our codebase where we would need or want to treat undefined and null separately.

HolgerJeromin: Personally I find it much more easy to have never "==" in my code base. Nothing to think about. 
But its your code base :-)

bradzacher (approved):
if you wanted to go one step further and clean up our now out-of-style `typeof x === 'undefined'` checks as part of this - that'd be gr8

https://cs.github.com/typescript-eslint/typescript-eslint?q=%2Ftypeof+.%2B%3F+%5B%21%3D%5D%3D%3D+%27undefined%27%2F

aw_yiss_refactorin

PR #6240, opened 2022-12-17
feat(eslint-plugin): [prefer-optional-chain] use type checking for strict falsiness

### LINKED ISSUE #4820: [eslint-plugin] Breaking: enable type information for prefer-optional-chain
## Overview

Following https://github.com/typescript-eslint/typescript-eslint/pull/4706#discussion_r846543436: ~neither rule~ `@typescript-eslint/prefer-optional-chain` does not require type information right now but would be able to avoid many false positives with it. Both @bradzacher and I have felt that pain repeatedly.

For the 6.0 release, it'd be great have type information added as a breaking change! That would enable adding the ~rules~ rule to ~`recommended-requiring-type-checking`~ `stylistic-type-checked`.

JoshuaKGoldberg:
BREAKING CHANGE:
Adds type checking to `prefer-optional-chain`

## PR Checklist

- [x] Addresses an existing open issue: fixes #4820

## Overview

Uses the type checker to skip reporting a complaint when the original expression includes a falsy value that isn't `undefined`.

Also adds a `looseFalsiness` flag to allow the old, faster behavior.


packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -38,19 +39,34 @@ export default util.createRule({
       description:
         'Enforce using concise optional chain expressions instead of chained logical ands, negated logical ors, or empty objects',
       recommended: 'strict',
+      requiresTypeChecking: true,
```
omril1, line 42: In retrospect, this feels like one of those times that an enum is better than a boolean.
It doesn't require type-checking all the time, right?

JoshuaKGoldberg: Most rules that claim to require type checking only do so some of the time. For example, `no-for-in-array` only comes into play if you have `for...in` loops -- which most files don't have most of the time.

JoshuaKGoldberg:
Lots of merge conflicts here, will need to fix up... eventually.

packages/type-utils/src/isTypeIntrinsic.ts
```diff
@@ -0,0 +1,12 @@
...
+export interface IntrinsicType extends ts.Type {
+  intrinsicName: string;
+}
+
+export function isTypeIntrinsic(
+  type: ts.Type,
+  intrinsicName: string,
+): type is IntrinsicType {
+  return type.intrinsicName === intrinsicName;
+}
```
JoshuaKGoldberg, line 12: @RebeccaStevens maybe a good candidate for `ts-api-utils`, thoughts?

packages/type-utils/src/isTypeIntrinsic.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface IntrinsicType extends ts.Type {
+  intrinsicName: string;
+}
+
+export function isTypeIntrinsic(
+  type: ts.Type,
+  intrinsicName: string,
+): type is IntrinsicType {
+  return type.intrinsicName === intrinsicName;
```
JoshuaKGoldberg, line 11: This actually is covered by tests, just in the `eslint-plugin` package. I wish codecov could merge monorepo reports like this (maybe it can?).

JoshuaKGoldberg: #6701

JoshuaKGoldberg:
Closing in favor of #6397. Woop!

Issue #6116, opened 2022-11-28
Repo: Source maps don't connect for ast-spec fixtures.test.ts


JoshuaKGoldberg:
### Suggestion

From https://github.com/typescript-eslint/typescript-eslint/pull/5926#discussion_r1033057043:

> I'm confused why codecov thinks this isn't covered by tests. I can debug it locally to see the `TSAbstractAccessorProperty` and `AccessorProperty` cases being hit. Though, sourcemaps seem to not be working - maybe that's the issue?
>
> <img width="1248" alt="Screenshot of debugging in VS Code, on a breakpoint in the `const result` variable below. A watch view shows type to be 'TSAbstractAccessorProperty'." src="https://user-images.githubusercontent.com/3335181/204174343-a9e55a77-c184-4876-85a3-13796b91e57d.png">

Steps to reproduce:

1. Add a breakpoint in `packages/typescript-estree/src/convert.ts`
1. Open a JavaScript Debug Terminal
1. `cd packages/ast-spec`
1. `yarn jest fixtures`

The breakpoint will switch to `convert.js`.

StyleShit:
So, it's not exclusive to ast-spec, it's a wider issue with all `*.ts` breakpoints in workspace dependencies -- e.g., adding a breakpoint in package A and running tests in package B (that depends on A), will only break in `*.js` files since we don't generate source maps:

https://github.com/typescript-eslint/typescript-eslint/blob/f853ef9c6eb1ccdb80dcaaa2e8ca100cddcfa0ac/tsconfig.base.json#L26

We used to have them until #10361, not sure why it was changed
Adding `sourceMap: true` back solves the issue

Opened #12677 as draft. Not sure what are the consequences of this other than a slightly bigger bundle

PR #6272, opened 2022-12-24
fix(typescript-estree): account for namespace nesting in AST conversion

### LINKED ISSUE #4966: Bug: incorrect AST produced for nested namespaces
This was forgotten when we worked around #2573

For `namespace foo { namespace bar {} }` we produce the following AST:
`TSModuleDeclaration[id = foo] > TSModuleBlock > TSModuleDeclaration[id = bar] > TSModuleBlock`

For `namespace foo.bar {}` produces the following AST:
`TSModuleDeclaration[id = foo] > TSModuleDeclaration[id = bar] > TSModuleBlock`

The AST we produce is driven by the AST TS produces - they emit the nested module declaration, so our code does as well.
Ultimately however is pretty incorrect and does not correctly represent the code as written.

Instead we should emit an `EntityName`, thus allowing `TSQualifiedName` in the `.id`:
`TSModuleDeclaration[id = foo.bar] > TSModuleBlock`

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4966

## Overview

TypeScript produces a nested body AST for Identifier names like `abc.def`. This unravels their body blocks and accumulates the name if necessary.


packages/typescript-estree/src/convert.ts
```diff
@@ -735,6 +731,41 @@ export class Converter {
...
+        node.body as ts.ModuleBlock | undefined,
+        this.convertChild(node.name),
+      ];
+    }
+
+    let name = node.name.text;
+
+    while (node.body && ts.isModuleDeclaration(node.body)) {
+      node = node.body;
+      name = `${name}.${node.name.text}`;
```
JoshuaKGoldberg, line 756: #6116 🤷

packages/typescript-estree/src/convert.ts
```diff
@@ -254,12 +254,8 @@ export class Converter {
...
   ): T {
     const result = data;
-    if (!result.range) {
-      result.range = getRange(node, this.ast);
-    }
-    if (!result.loc) {
-      result.loc = getLocFor(result.range[0], result.range[1], this.ast);
-    }
+    result.range ??= getRange(node, this.ast);
+    result.loc ??= getLocFor(result.range[0], result.range[1], this.ast);
```
JoshuaKGoldberg, line 258: Couldn't resist while I was here 😄 

Feels like this should be a lint rule enabled always...

bradzacher (changes requested):
I do wish we could use `Identifier` for the dotted name - unfortunately `Identifier.name` needs to be a valid JS identifier - dots aren't allowed!

Instead we can convert this to a `TSQualifiedName`

bradzacher:
Now that #6443 has landed in main, we'll need to merge v6 then update this PR accordingly.

bradzacher (approved):
all_green

PR #6273, opened 2022-12-24
feat(typescript-estree): remove false from optional AST boolean properties

### LINKED ISSUE #5020: AST standardisation: remove `undefined`/optional from node properties unless it really makes sense
Across the AST we have a lot of cases where we've made properties optional.
I think there were various reasons for why it was done this way (not that I can find the comments any more).

From the perspective of an AST consumer, `undefined` can create some very weird states. Two examples of this:

1) There are a number of `boolean` properties that are optional. On the surface this makes it look like the property with 3 meaningful states (`true`, `false`, or `undefined`), even though in reality it only ever actually represents 2 states.
Usually in these cases the parser emits `true` or `undefined` (and never `false`), and in some rarer cases it emits `true`, and then uses `false` or `undefined` interchangeably depending on some internal, obfuscated criteria.
2) There are a few cases where an array property is optional. In these cases usually the parser will never emit an empty array, and will instead emit `undefined`.
    - From a memory-usage perspective, this could be justified (an empty array is really a heavy "empty" marker). But if that is our goal then these cases should be typed so that it is clear that there are never 0 elements (eg by defining `[Foo, ...Foo[]]`)

We should standardise the AST to remove `undefined`/optional from all places, unless there is a very good reason for it to be optional.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5020

## Overview

Switches most `?: boolean` and `: boolean` properties to `?: true` in `ast-spec`. Updates `convert.ts` accordingly to only assign those properties if they're `true`.

Does not update snapshots that need manual touchups, pending resolution of whether to use this or #6274.


JoshuaKGoldberg:
Going with #6274 per https://github.com/typescript-eslint/typescript-eslint/issues/5020#issuecomment-1397770273.

PR #6274, opened 2022-12-24
feat(typescript-estree): remove optionality from AST boolean properties

### LINKED ISSUE #5020: AST standardisation: remove `undefined`/optional from node properties unless it really makes sense

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5020

## Overview

Switches `?: boolean` properties to `: boolean` in `ast-spec`. Updates `convert.ts` accordingly to always assign those properties even if they're not `true`.

~Does not update snapshots that need manual touchups, pending resolution of whether to use this or #6273.~


packages/ast-spec/src/parameter/ArrayPattern/spec.ts
```diff
@@ -7,7 +7,7 @@ import type { DestructuringPattern } from '../../unions/DestructuringPattern';
 export interface ArrayPattern extends BaseNode {
   type: AST_NODE_TYPES.ArrayPattern;
   elements: (DestructuringPattern | null)[];
-  typeAnnotation?: TSTypeAnnotation;
-  optional?: boolean;
-  decorators?: Decorator[];
+  typeAnnotation: TSTypeAnnotation | undefined;
+  optional: boolean;
+  decorators: Decorator[];
```
bradzacher, line 12: Just because I was looking - in TS 5.0 decorators on function parameters are a semantic error under the new decorators parsing

[playground](https://www.typescriptlang.org/play?experimentalDecorators=false&emitDecoratorMetadata=false&ts=5.0.0-dev.20230131#code/CYUwxgNghgTiAEYD2A7AzgF3qZAueUKAngNwBQZAZgK4pgYCWq8lSSAFGfPAAI5LwA2lAC6+QZhgMUAcxEAaLr37wARvAC88AIyLufcAIDeYAL74T+SdJmm9yw-AB0L4IoCU8I-FMVIUNDR4AGEvJWR0DBhqeiQYTm59FWExIWtZBSUkx3UtXSyHZC8zCzArKJs7AoMigAcpADcoDARgcqlZe2yilycQe09vX18yIA)

But until `experimentalDecorators` is removed, we'll need to keep this.

packages/ast-spec/src/parameter/RestElement/spec.ts
```diff
@@ -8,8 +8,8 @@ import type { AssignmentPattern } from '../AssignmentPattern/spec';
...
   type: AST_NODE_TYPES.RestElement;
   argument: DestructuringPattern;
-  typeAnnotation?: TSTypeAnnotation;
-  optional?: boolean;
-  value?: AssignmentPattern;
-  decorators?: Decorator[];
+  typeAnnotation: TSTypeAnnotation | undefined;
+  optional: boolean;
+  value: AssignmentPattern | undefined;
+  decorators: Decorator[];
```
bradzacher, line 14: this is the only decorator usage that is always a semantic error (with or without 5.0 changes).
But probs best to just leave it in until we remove it from all the params

packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts
```diff
@@ -3,9 +3,9 @@ import { AST_NODE_TYPES } from '@typescript-eslint/utils';
 
 import * as util from '../util';
 
-type MakeRequired<Base, Key extends keyof Base> = Omit<Base, Key> &
-  Required<Pick<Base, Key>>;
-
+type MakeRequired<Base, Key extends keyof Base> = Omit<Base, Key> & {
+  [K in Key]-?: NonNullable<Base[Key]>;
+};
```
bradzacher, line 8: OOC - do you still need the `-?` here?

packages/scope-manager/src/referencer/ClassVisitor.ts
```diff
@@ -103,12 +103,18 @@ class ClassVisitor extends Visitor {
     node: TSESTree.Parameter,
     withDecorators: boolean,
   ): void {
-    if ('typeAnnotation' in node) {
```
bradzacher, line 106: duck_typing

bradzacher (approved):
matrix_deletion

This LGTM - generate all the snapshots and we can get this in.
There's too many to review by hand, but the changes themselves are all looking good!

packages/ast-spec/src/parameter/ArrayPattern/spec.ts
JoshuaKGoldberg: > until `experimentalDecorators` is removed

Heh. Maybe in TS 50.0.

packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts
JoshuaKGoldberg: Yeah, without it we get errors such as:

```ts
source.getTokensAfter(node)[0].value !== ',' &&
//                    ~~~~
```

```plaintext
Argument of type 'TypeParameterWithConstraint' is not assignable to parameter of type 'Node | Token'.
  Type 'TypeParameterWithConstraint' is not assignable to type 'TSTypeParameter'.
    Property 'constraint' is optional in type 'TypeParameterWithConstraint' but required in type 'TSTypeParameter'.
```

JoshuaKGoldberg:
Some new test failures after merge. I'm going to take a break then look at them... this is a mildly frustrating PR to work with 🙃

PR #6299, opened 2023-01-04
Type merging with empty interfaces

shramee:
## PR Checklist

- [ ] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

```ts
import 'module-a';
declare module "module-a" {
 	interface Person extends Mammal {} // Now interface Person has properties of both Person and Mammal
}
```

#### Elaborate example,

```ts
// region module-a
interface Person {
	name: string,
};
// endregion module-a


// region module-b
interface Mammal {
	age: number,
	sex: string // intentionally a string
};

interface Person extends Mammal {} // Now Person has properties of Mammal too
// endregion module-b


let person: Person = {
	age: 25,
	name: 'Jen',
	sex: 'curious',
};
```

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-empty-interface.md
+++ b/packages/eslint-plugin/docs/rules/no-empty-interface.md
@@ -68,3 +68,4 @@ This rule accepts a single object option with the following default configuratio
 ## When Not To Use It
 
 If you don't care about having empty/meaningless interfaces, then you will not need this rule.
+Or if you want to use declaration merging to merge new properties into an existing interface.


bradzacher (commented):
TBH I don't think this adds anything meaningful to the docs.

If you're intentionally declaring an empty interface on purpose so that you can use declaration merging to add properties later, then your usecase is covered by the first line "if you don't care about empty interfaces".

Additionally declaration merging like that is a really rare usecase, so I don't think that explicitly calling it out is going to be relevant to most users. This is supported by the fact that the rule docs are some 5+ years old and nobody has asked questions or requested additional clarification here.

Thanks for contributing!

shramee:
The thing is when doing it for merging interfaces the devs do care about it.
Since it is pretty much the only way to do it when overriding a module with `declare module`.
The statement their seems to make it sound like it's stupid to turn it off. Making it harder to justify it to the community (for other community driven projects).

shramee:
@bradzacher Your opinion on this? https://github.com/typescript-eslint/typescript-eslint/pull/6299#issuecomment-1370713597

PR #6310, opened 2023-01-08
chore: fix repo linting

JamesHenry:
When the renovate PR went in which upgraded the major version of `eslint-plugin-jest` some issues were introduced.

- `valid-expect` is triggered by some more unusual, advanced assertion we have, so we can simply ignore it
- `no-jest-import` rule was removed by the plugin, so we remove it from our config

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -214,7 +214,6 @@ module.exports = {
         'jest/no-alias-methods': 'error',
         'jest/no-identical-title': 'error',
         'jest/no-jasmine-globals': 'error',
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
Weird how failures were able to be introduced. https://github.com/typescript-eslint/typescript-eslint/pull/6233 had a clean build. Maybe a caching issue?

JamesHenry:
@JoshuaKGoldberg I agree, I'll raise this with the Nx team

JoshuaKGoldberg (approved):
👍

PR #6311, opened 2023-01-08
chore(website): reorganized sidebar and header nav a bit

JoshuaKGoldberg:
Apropos of nothing, I feel slightly positive about this change:

* _Docs_ encompasses all of docs, not just the _Getting Started_ area
* I think people generally want to know package _Architecture_ earlier than & more often than _Contributing_

Thoughts?


bradzacher (approved):
I was just thinking this today!
It felt a bit weird going via "Getting Started" to get to the maintenance docs.

bradzacher:
fix conflict them :shipit: sir

PR #6317, opened 2023-01-09
chore(website): link every rule doc to the playground with default config

### LINKED ISSUE #6085: Enhancement(rule docs): add a quick-link that allows you to open a rule in the playground
### Suggestion

One thing I find myself doing a lot is testing specific rules out in the playground.
Right now I have to manually open the playground and type in the config.

It would be great if we could do the following:
- Add a link to open the playground with the default config for the rule (eg `"@typescript-eslint/rule-name": "error"`)
- Add a link to each examples to allow you to open them in the playground with the right config.

I think this would vastly lower the barrier to understanding and allow users to quickly verify behaviour. Eg "the docs say this, but does it cover this?" *click link on example and enter new code to play around*.

mhnaeem:
## PR Checklist

- [X] Addresses an existing open issue: fixes #6085

## Overview

This change includes a link in every rule doc to the playground, clicking on the link opens the playground with default config for the rule already loaded.

The plugin responsible for generating/modifying the rules doc was edited to add a single link referencing to the playground with the default config hashed.


packages/website/plugins/generated-rule-docs.ts
```diff
@@ -193,35 +194,79 @@ export const generatedRuleDocs: Plugin = () => {
         type: 'paragraph',
       } as mdast.Paragraph);
 
+      // need this, so we don't have comments inside the JSON objects
+      const getEslintrcStr = (withComment: boolean): string => {
```
JoshuaKGoldberg, line 198: [Nit] A couple of small changes:

* I'm not a fan of abbreviations like "Str", as they can be hard to read - especially for ESL (English Second Language) folks.
* A `/** ... */` JSDoc would be more standard to get intellisense matching up the comment to the function

```suggestion
      /**
       * @param withComment Whether to include a full comment note.
       * @remarks `withComment` can't be usedd inside a JSON object.
       */
      const getEslintrcStr = (withComment: boolean): string => {
```

How do you feel about this?

packages/website/plugins/generated-rule-docs.ts
```diff
@@ -193,35 +194,79 @@ export const generatedRuleDocs: Plugin = () => {
...
+        children: [
+          {
+            type: 'link',
+            url: `/play#eslintrc=${convertToPlaygroundHash(
+              getEslintrcStr(false),
+            )}`,
+            children: [
+              {
+                type: 'text',
+                value: 'Click here to try this rule in the playground',
```
JoshuaKGoldberg, line 228: Nit: not everybody "clicks". Some tap, or press enter, or use other things. We can just remove that bit of the text.

```suggestion
                value: 'Try this rule in the playground ↗',
```

packages/website/plugins/generated-rule-docs.ts
```diff
@@ -193,35 +194,79 @@ export const generatedRuleDocs: Plugin = () => {
...
-  "rules": {
-    // Note: you must disable the base rule as it can report incorrect errors
-    "${extendsBaseRuleName}": "off",
-    "@typescript-eslint/${file.stem}": "${optionLevel}"
-  }
-};`,
+        value: `module.exports = ${getEslintrcStr(true)};`,
       } as mdast.Code);
+
+      root.children.splice(howToUseH2Index + 2, 0, {
```
JoshuaKGoldberg, line 218: You know, I think the link would be just a _little_ better if it was given `float: right`. That way it doesn't take up any extra vertical space:

Screenshot of a right-aligned 'Click here to try this rule in the playground' just under its code block

The best way to do this would be to add a new component to `packages/website/src/theme/MDXComponents/index.tsx`. You can see how the `<rule-attributes>` component is used in code for a reference.

Is this something you have interest & time for in this PR? No worries if not (or you think it's better this way!), we can always file a followup issue.

JoshuaKGoldberg (dismissed):
Looks great, thanks for sending this in @mhnaeem! ✨ Just left a few comments for your review. Let us know what you think & we can get this merged.

mhnaeem:
@JoshuaKGoldberg Thank you for your input, I have made all the requested changes. Let me know if something else could be improved.

packages/website/src/theme/MDXComponents/TryInPlayground.tsx
```diff
@@ -0,0 +1,21 @@
...
+}): React.ReactNode {
+  return (
+    <div>
+      <a
+        href={`/play#eslintrc=${eslintrcHash}`}
+        className={styles.tryInPlaygroundLink}
+      >
+        Try this rule in the playground ↗
+      </a>
+      <br />
```
JoshuaKGoldberg, line 18: There's no need for a `<br />` or container `<div>` that I can see. I like it not taking up any extra vertical space a bit more.

```suggestion
```

packages/website/src/theme/MDXComponents/TryInPlayground.tsx
```diff
@@ -0,0 +1,21 @@
...
+export function TryInPlayground({
+  eslintrcHash,
+}: {
+  eslintrcHash: string;
+}): React.ReactNode {
+  return (
+    <div>
+      <a
+        href={`/play#eslintrc=${eslintrcHash}`}
+        className={styles.tryInPlaygroundLink}
```
JoshuaKGoldberg, line 14: I'm remembering now that folks generally associate the ↗ with opening in a new tab.

```suggestion
        className={styles.tryInPlaygroundLink}
        target="_blank"
```

JoshuaKGoldberg (changes requested):
Looks great, thanks for sending this in @mhnaeem! I'll apply a couple of touchups now and ship. 🔥

JoshuaKGoldberg (approved):

bradzacher:
goodjobpotter

PR #6318, opened 2023-01-10
docs: document import/extensions slowness

bradzacher:
This was a surprising result that I found at work today.
I found that `import/extensions` took ~30s for a total codebase run - which was 1/3 of the overall lint time!!!

I replaced it with the listed `no-restricted-syntax` config which in comparison runs in less than 1ms over the entire codebase, which is a YUGE win.

This hadn't been documented before because when I wrote this performance section I didn't realise that this lint rule could be slow.

I tried to clearly spell out the fact that the alternative has false positives and is not a 1:1 replacement. Now-a-days I've found that there's very few npm packages that still include `.js` in their name, and false-positives like `foo.js.ts` should really not happen - so in practice it'll be 1:1, but still tried to make it clear there could be false positives.


docs/linting/Troubleshooting.md
```diff
@@ -311,6 +311,69 @@ The following rules do not have equivalent checks in TypeScript, so we recommend
...
+There are two uses for this rule:
+
+1. to enforce file extensions are always used,
+2. to enforce file extensions are never used.
+
+##### Enforcing extensions are used
+
+If you want to enforce file extensions are always used and you're **NOT** using `moduleResolution` `node16` or `nodenext`, then there's not really a good alternative for you, and you should continue using the `import/extensions` lint rule.
+
+If you want to enforce file extensions are always used and you **ARE** using `moduleResolution` `node16` or `nodenext`, then you don't need to use the lint rule at all because TypeScript will automatically enforce that you include extensions (TS 2834 & 2835)!
```
JoshuaKGoldberg, line 325: Users might not know this refers to specific error codes. And those error codes sometimes change version-to-version. Maybe let's just remove these?

```suggestion
If you want to enforce file extensions are always used and you **ARE** using `moduleResolution` `node16` or `nodenext`, then you don't need to use the lint rule at all because TypeScript will automatically enforce that you include extensions!
```

docs/linting/Troubleshooting.md
```diff
@@ -311,6 +311,69 @@ The following rules do not have equivalent checks in TypeScript, so we recommend
...
 - `import/no-deprecated`
 
+#### `import/extensions`
+
+There are two uses for this rule:
+
+1. to enforce file extensions are always used,
+2. to enforce file extensions are never used.
+
+##### Enforcing extensions are used
```
JoshuaKGoldberg, line 321: Note that the font size for `#####`s is pretty small:

Screenshot of headings with small sizes

Maybe a good time to bump that up a bit?

JoshuaKGoldberg (approved):
Fun! 🧠

docs/linting/Troubleshooting.md
bradzacher: How did you see this? Did you check it out locally?
When I tried the netlify deploy link above it doesn't include my changes :(

JoshuaKGoldberg: Oh I tried this with `yarn start` locally. Weird that the deploy doesn't have it.

bradzacher:
@JoshuaKGoldberg - I just made the performance setion its own page entirely.
I think it's a meaty enough chunk that it deserves it.

docs/linting/troubleshooting/FORMATTING.md
```diff
@@ -16,13 +27,18 @@ Linters often take seconds or more to run because they apply many logical rules
 
 ### Problems with Using Linters as Formatters
 
-Linters apply much more work than formatters -- often including potentially multiple rounds of rule fixers.
-That generally makes them run orders of magnitude slower.
+Linters are designed to run in a parse, check, report, fix cycle. This means that there is a lot of intermediate work that needs to be done before a linter can fix a formatting issue with your code.
+
+Additionally linters typically run each rule isolated from one another. This has several problems with it like:
```
JoshuaKGoldberg, line 32: Nit:

```suggestion
Additionally linters typically run each rule isolated from one another. This has several problems with it such as:
```

docs/linting/troubleshooting/Performance.md
```diff
@@ -0,0 +1,150 @@
...
+
+## The `indent` / `@typescript-eslint/indent` rules
+
+This rule helps ensure your codebase follows a consistent indentation pattern.
+However this involves a _lot_ of computations across every single token in a file.
+Across a large codebase, these can add up, and severely impact performance.
+
+We recommend not using this rule, and instead using a tool like [`prettier`](https://www.npmjs.com/package/prettier) to enforce a standardized formatting.
+
+See our [documentation on formatting](./Formatting.md) for more information.
```
JoshuaKGoldberg, line 56: The file is `FORMATTING.md` but the link here is `Formatting.md`. I'd prefer renaming the file to `Formatting.md`.

JoshuaKGoldberg (approved):
Agreed, this is a good move. Thanks!

I'll send a separate PR standardizing file names to `Upper_Snake_Case.mdx`.

Edit: #6325

docs/linting/troubleshooting/Performance.md
bradzacher: ugh damn case insensitive filesystems.
i did rename it but git didn't pick it up.
i'll have to do an explicit `git mv`

liedlgruber (commented):
docs/linting/Troubleshooting.md

liedlgruber:
> docs/linting/Troubleshooting.md

https://github.com/typescript-eslint/typescript-eslint/pull/6318#pullrequestreview-1248739612

liedlgruber (commented):
docs/linting/troubleshootingpackages/website/sidebars/sidebar.base.js/Performance.md

liedlgruber (commented):
gh pr checkout 6318

PR #6330, opened 2023-01-12
feat(eslint-plugin): [promise-function-async] check for promises in implicit return types

### LINKED ISSUE #6329: Enhancement: [promise-function-async] flag non-`async` functions returning unions with non-`Promise` types

https://typescript-eslint.io/rules/promise-function-async/

### Description

I am finding cases in my code where it accidentally is returning undefined or some other guard type (null, etc) from a function that we expected to be async, but the rule promise-function-async is not flagging those as I expected it to (allowing us to catch those mistakes).

Please add an option to `promise-function-async` that will allow it to flag any function that returns promises at all, either as the main return type, or as a union containing promises.

### Fail

```typescript
function getSomething(id?: string) {
    if (!id) {
        return undefined
    }

    return db.getByID(id)
}
```

### Pass

```typescript
async function getSomething(id?: string) {
    if (!id) {
        return undefined
    }

    return db.getByID(id)
}
```

### Additional Info

There is an issue for the opposite of my request, #1005. I’d like that feature back (probably behind an option, seeing as it was designated as a bug before).

ericbf:
When we return a union containing a promise from a function implicitly, it's often a mistake. This commit makes it so if the return type is explicit, any `Promise` in the return type (whether it's part of a union or not) will flag the function. If it is intentional, make the return type explicit.

Fixes #6329

## PR Checklist

- [X] Addresses an existing open issue: fixes #6329

## Overview

This implements the heuristic that @bradzacher suggested. If the return type is explicit, it allows unions with non-`Promise` types, but if it is implicit, it flags those.

### DIFF
--- a/packages/eslint-plugin/docs/rules/promise-function-async.md
+++ b/packages/eslint-plugin/docs/rules/promise-function-async.md
@@ -46,6 +46,7 @@ async function functionReturnsPromise() {
   return Promise.resolve('value');
 }
 
+// An explicit return type that is not Promise means this function cannot be made async, so it is ignored by the rule
 function functionReturnsUnionWithPromiseExplicitly(
   p: boolean,
 ): string | Promise<string> {
--- a/packages/eslint-plugin/docs/rules/promise-function-async.md
+++ b/packages/eslint-plugin/docs/rules/promise-function-async.md
@@ -11,9 +11,11 @@ Ensures that each function is only capable of:
 - returning a rejected promise, or
 - throwing an Error object.
 
-In contrast, non-`async` `Promise` - returning functions are technically capable of either.
+In contrast, non-`async`, `Promise`-returning functions are technically capable of either.
 Code that handles the results of those functions will often need to handle both cases, which can get complex.
 This rule's practice removes a requirement for creating code to handle both cases.
+
+> When functions return unions of `Promise` and non-`Promise` types implicitly, it is usually a mistake—this rule flags those cases. If it is intentional, make the return type explicitly to allow the rule to pass.
 
 ## Examples
 
@@ -29,6 +31,10 @@ const arrowFunctionReturnsPromise = () => Promise.resolve('value');
 function functionReturnsPromise() {
   return Promise.resolve('value');
 }
+
+function functionReturnsUnionWithPromiseImplicitly(p: boolean) {
+  return p ? 'value' : Promise.resolve('value');
+}
 ```
 
 ### ✅ Correct
@@ -38,5 +44,15 @@ const arrowFunctionReturnsPromise = async () => Promise.resolve('value');
 
 async function functionReturnsPromise() {
   return Promise.resolve('value');
+}
+
+function functionReturnsUnionWithPromiseExplicitly(
+  p: boolean,
+): string | Promise<string> {
+  return p ? 'value' : Promise.resolve('value');
+}
+
+async function functionReturnsUnionWithPromiseImplicitly(p: boolean) {
+  return p ? 'value' : Promise.resolve('value');
 }
 ```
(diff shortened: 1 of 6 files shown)


packages/type-utils/src/containsTypesByName.ts
```diff
@@ -5,13 +5,16 @@ import { isTypeFlagSet } from './typeFlagUtils';
...
 /**
  * @param type Type being checked by name.
+ * @param allowAny Whether to consider `any` and `unknown` to match.
  * @param allowedNames Symbol names checking on the type.
- * @returns Whether the type is, extends, or contains all of the allowed names.
+ * @param mustMatchAll Whether all parts have to match, as opposed to any parts matching.
+ * @returns Whether the type is, extends, or contains the allowed names (or all matches the allowed names, if mustMatchAll is true).
  */
-export function containsAllTypesByName(
+export function containsTypesByName(
```
bradzacher, line 13: `@typescript-eslint/type-utils` is a public package so:
- this rename is a breaking change
- adding a new non-optional param is a breaking change

we'll need to:
- either keep the function named as is, or add an alias to maintain backwards compat
- set a default value for the new param (probably `true`)

packages/eslint-plugin/docs/rules/promise-function-async.md
```diff
@@ -39,4 +45,14 @@ const arrowFunctionReturnsPromise = async () => Promise.resolve('value');
 async function functionReturnsPromise() {
   return Promise.resolve('value');
 }
+
+function functionReturnsUnionWithPromiseExplicitly(
```
bradzacher, line 49: I think this bears repeating in the context of an example

```suggestion
// an explicit return type that is not Promise means this function cannot be made async, and is thus ignored by the rule
function functionReturnsUnionWithPromiseExplicitly(
```

packages/eslint-plugin/src/rules/promise-function-async.ts
```diff
@@ -111,10 +111,12 @@ export default util.createRule<Options, MessageIds>({
       const returnType = checker.getReturnTypeOfSignature(signatures[0]);
 
       if (
-        !util.containsAllTypesByName(
+        !util.containsTypesByName(
           returnType,
           allowAny!,
           allAllowedPromiseNames,
+          // Only if a return type is explicitly declared must it match all
+          Boolean(node.returnType),
```
bradzacher, line 119: Nit we prefer this style in the codebase, rather than the implicit coercion

```suggestion
          node.returnType != null,
```

bradzacher (changes requested):
overall the changes look good - but we'll need to ensure we're not breaking the API

packages/type-utils/src/containsTypesByName.ts
ericbf: What if I add `containsAllTypesByName` back, making it call the new, renamed version passing in `true` to the 4th param. That will maintain the same behavior and keep it backwards compatible.

packages/eslint-plugin/docs/rules/promise-function-async.md
ericbf: Sounds great

packages/eslint-plugin/src/rules/promise-function-async.ts
ericbf: Sounds great

packages/type-utils/src/containsTypesByName.ts
ericbf: I switched the behavior around so the fourth param is now `matchAnyInstead`, with default false to maintain the same behavior.

bradzacher (approved):
thanks for this!

Issue #6033, opened 2022-11-19
Repo: Automatically remove 'awaiting response' label when PR review is requested


JoshuaKGoldberg:
### Suggestion

I've noticed that the following will happen:

1. A PR is sent or updated
2. A review requests changes and the `awaiting response` label is added
3. The author addresses changes and re-requests review
4. Some time passes
5. Another review comes in

During _4. Some time passes_, the `awaiting response` label is still on the PR - even though it's waiting for us to review it!

Proposal: let's add a repo automation that removes the `awaiting response` label from a PR when a review is requested on it? I think requesting a review generally implies a code response has been pushed.

PR #6349, opened 2023-01-18
feat(eslint-plugin): [member-ordering] add support for grouping readonly fields

### LINKED ISSUE #1063: [member-ordering] separating readonly members is not supported
Hi, I want to separate `readonly` members from others. For example;
```TS
export class Class
{
    public static readonly A_CONSTANT = 56;
    public static readonly STRING_CONSTANT = "string";

    public static CLASS_MEMBER = "can be modified";

    public readonly anInstanceMember: string;

    public modifiableInstanceMember: string;
}
```

I couldn't find an option for `readonly` members in the documentation (https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-ordering.md)

**Repro**
N/A

```JSON
{
  "rules": {
    '@typescript-eslint/member-ordering': 'error',
  }
}
```
**Versions**

| `@typescript-eslint/eslint-plugin` | `2.3.2` |
| `@typescript-eslint/parser`        | `2.3.2` |
| `node`                             | `8.14.1` |
| `npm`                              | `6.11.3` |

IronGeek:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1063

## Overview

This PR introduces new member attribute/type/Kind: `readonly-field`, a superset of `field`, which denotes all fields with readonly modifiers. This makes it possible to group/ sort fields based on their readonly modifiers and optionally with their Accessibility, and or Scope too.

Like other member types, `readonly-field` can also be joined with accessibility and scope attributes to create more specific type; the following is the complete combination of `readonly-field` with Scope and Accessibility introduced in this PR:
```json
"readonly-field"
"public-readonly-field"
"public-decorated-readonly-field"
"decorated-readonly-field"
"static-readonly-field"
"public-static-readonly-field"
"instance-readonly-field"
"public-instance-readonly-field"
"abstract-readonly-field"
"public-abstract-readonly-field"
"protected-readonly-field"
"protected-decorated-readonly-field"
"protected-static-readonly-field"
"protected-instance-readonly-field"
"protected-abstract-readonly-field"
"private-readonly-field"
"private-decorated-readonly-field"
"private-static-readonly-field"
"private-instance-readonly-field"
"#private-readonly-field"
"#private-static-readonly-field"
"#private-instance-readonly-field"
```

For example, with this configuration the following code would be treated as valid:
```json
{
  "rules": {
    "@typescript-eslint/member-ordering": [
      "error", { 
        "default": [
            "decorated-readonly-field",
            ["public-readonly-field", "readonly-field"]
            "protected-readonly-field",
            "private-readonly-field",
            "abstract-readonly-field",
            "field"
        ] 
      }
    ]
  }
}
```
```ts
abstract class Foo {
  @Dec public readonly DA: string;
  @Dec protected readonly DB: string;
  @Dec private readonly DC: string;

  public static readonly SA: string;
  public readonly IA: string;
  static readonly #SD: string;
  readonly #ID: string;

  protected static readonly SB: string;
  protected readonly IB: string;

  private static readonly SC: string;
  private readonly IC: string;

  public abstract readonly AA: string;
  protected abstract readonly AB: string;

  public RA: string;
  RB: string;
  private RC: string;
  #RD: string;
}
```


packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -329,6 +334,8 @@ const functionExpressions = [
  * @param node the node to be evaluated.
  */
 function getNodeType(node: Member): MemberKind | null {
+  const readonly = 'readonly' in node && node.readonly === true;
```
JoshuaKGoldberg, line 337: There's no need to check this value for all the fields. This can be moved inline to only the node types that need it. As a benefit, `'readonly' in node` won't need to be checked for them.

Example:

```diff
-   const readonly = 'readonly' in node && node.readonly === true;
...
    case AST_NODE_TYPES.TSAbstractPropertyDefinition:
-      return readonly ? 'readonly-field' : 'field';
+      return node.readonly ? 'readonly-field' : 'field';
```

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -514,27 +523,48 @@ function getRank(
...
+        type === 'field' ||
         type === 'method' ||
         type === 'get' ||
         type === 'set')
     ) {
       memberGroups.push(`${accessibility}-decorated-${type}`);
       memberGroups.push(`decorated-${type}`);
+
+      if (type === 'readonly-field') {
+        // if its readonly-field, push also its superset type
```
JoshuaKGoldberg, line 536: I think these are self-evident from the code. Is there a particular reason you wanted them?

```suggestion
```

(If we do keep them in, nit: it's `it's` not `its`. They're possessive, not plural.)

packages/eslint-plugin/docs/rules/member-ordering.md
```diff
@@ -1290,6 +1312,21 @@ The third grouping option is to ignore both scope and accessibility.
...
 ```
 
+### Member Group Types (Readonly Fields)
+
+It is possible to group fields by their `readonly` modifiers.
+
+```jsonc
+[
+  // Index signature
+  // No grouping for index signature.
```
JoshuaKGoldberg, line 1322: Index signatures can be `readonly`, though. And you can even have a `: number` & `: string` in the same type:

```ts
interface Example {
    [i: number]: number;
    readonly [i: string]: number;
}
```

Cases of having one or both of those be mutable or not should be in the unit tests.

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -340,13 +347,15 @@ function getNodeType(node: Member): MemberKind | null {
     case AST_NODE_TYPES.TSConstructSignatureDeclaration:
       return 'constructor';
     case AST_NODE_TYPES.TSAbstractPropertyDefinition:
-      return 'field';
+      return readonly ? 'readonly-field' : 'field';
```
JoshuaKGoldberg, line 350: When I change this back to just `return 'field'`, all the tests still pass. Looks like there's a gap in coverage here.

(spotted in [Codecov's coverage report](https://app.codecov.io/gh/typescript-eslint/typescript-eslint/pull/6349) that this line is only partially covered)

JoshuaKGoldberg (changes requested):
A good start - thanks for getting this PR up! 🙌 

Requesting changes on more test coverage & handling index signatures.

Aside: this rule gets more complicated and hard to deal with in every PR. But I don't see a way around it 😞. This comment isn't actionable and you can ignore my sadness. I'm just ... musing.

packages/eslint-plugin/src/rules/member-ordering.ts
IronGeek: Noted.

packages/eslint-plugin/src/rules/member-ordering.ts
IronGeek: Actually the comment is added there to draw attention to the implementation, since I'm not really sure if it's the right way to do it. 

I'm glad the code already speak for itself, I will remove the comments in the update.

packages/eslint-plugin/src/rules/member-ordering.ts
IronGeek: I'll add more tests to capture the cases for readonly `TSAbstractPropertyDefinition` and possibly for `TSPropertySignature` too  (it's also reported partially covered in [Codecov's coverage report](https://app.codecov.io/gh/typescript-eslint/typescript-eslint/pull/6349))

packages/eslint-plugin/docs/rules/member-ordering.md
IronGeek: I actually miss this one. I didn't realize that index signatures could be readonly too. I will try to update the code and add more test cases for this.

JoshuaKGoldberg (approved):
Whew, looks good to me! Thanks @IronGeek! 🙌 

In theory one might want something like `readonly-method`, for cases like `{ readonly A: () => void }`. But that can be a feature add if requested after this PR. The rule is (and already was) complex enough  as it is.

JoshuaKGoldberg:
Btw, sorry for the month-long wait time. This fell off my radar. After the v6 release I'm hopeful we can work on #6033 as a process improvement so it doesn't happen again.

PR #6364, opened 2023-01-22
chore(website): used lighter primary color for table of contents active links in dark mode

JoshuaKGoldberg:
## Overview

Quick little bug I noticed working in dark mode. This makes the ToC color match the link color for general body text and the left side nav.

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
Before screenshot of middling purple text on dark gray background
</td>
<td>
After screenshot of light purple text on dark gray background
</td>
</tr>
</tbody>
</table>


Josh-Cena:
I don't think overriding the active color is the best fix. Note that in dark mode, we already use a lighter color as the primary color—hence why the sidebar and content links are accessible without CSS overrides. I would need to look into why it doesn't work for the TOC.

Josh-Cena:
OK scratch that: we actually do override the link color in dark mode...

https://github.com/typescript-eslint/typescript-eslint/blob/d5622445eabe1a0d14816c40eab4b9750ee949b7/packages/website/src/css/custom.css#L47

Should I suggest lightening the primary color instead? We seem to be using this light blue in more places than the actual primary color.

bradzacher (approved):

PR #6365, opened 2023-01-22
test: verify that there are no errors on the website

### LINKED ISSUE #6354: Repo: End-to-end test that no console errors are logged on the homepage
### Suggestion

Coming over from #6350: we had an issue with our React code caused by us using invalid HTML. Specifically, React was logging this:

```plaintext
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
```

We should end-to-end test to make sure this doesn't happen on the homepage again. Right now we just test that the homepage has no accessibility issues detectable by [aXe](https://www.deque.com/axe/):

https://github.com/typescript-eslint/typescript-eslint/blob/9ffdb056f77af235031a0d7bef88c67207a12795/packages/website/tests/index.spec.ts#L4-L7

[Playwright is able to listen to console messages](https://playwright.dev/docs/api/class-consolemessage). I propose we make a second end-to-end test that checks to make sure no `console.error` calls happened.

It wouldn't surprise me if a library already exists for Playwright somewhere that does this...

Jefiozie:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6354

## Overview

I've made some small changes to test if there are any errors on the website by listening to the `console`. Added a second test with the opposite for verification. Happy to improve some more if needed. 

@JoshuaKGoldberg  I've made the choice not to search for a library as the tests are pretty simple and have no "magic" from the library. 

Update: hmm I'm not sure what I'm doing wrong here. On my machine I don't see any issues. Am I'm missing a step if I just run the playwrih test with NX?


packages/website/tests/index.spec.ts
```diff
@@ -1,7 +1,30 @@
...
+
+  test('should have no errors', async ({ page }) => {
+    const errorMessages = [] as string[];
+    page.on('console', msg =>
+      msg.type() === 'error'
+        ? errorMessages.push(`[${msg.type()}] ${msg.text()}`)
+        : null,
+    );
+    await page.goto('/');
+    expect(errorMessages.length).toBe(0);
```
JoshuaKGoldberg, line 18: This will make things easier to debug:

```suggestion
    expect(errorMessages).toBe([);
```

packages/website/tests/index.spec.ts
```diff
@@ -1,7 +1,30 @@
...
+  test('should have errors', async ({ page }) => {
+    const errorMessages = [] as string[];
+    page.on('console', msg =>
+      msg.type() === 'error'
+        ? errorMessages.push(`[${msg.type()}] ${msg.text()}`)
+        : null,
+    );
+    await page.evaluate(() => console.error('error'));
+    expect(errorMessages.length).toBe(1);
+  });
```
JoshuaKGoldberg, line 29: This might be useful for development, but not necessary for actually testing the site.

```suggestion
```

(I'll end up making a library and/or finding one after this, I hope)

packages/website/tests/index.spec.ts
```diff
@@ -1,7 +1,30 @@
...
+test.describe('Website', () => {
+  test('Axe', async ({ page }) => {
+    await page.goto('/');
+    await new AxeBuilder({ page }).analyze();
+  });
+
+  test('should have no errors', async ({ page }) => {
+    const errorMessages = [] as string[];
+    page.on('console', msg =>
+      msg.type() === 'error'
```
JoshuaKGoldberg, line 13: > `'error'`

I think we'd want to look at warnings too.

JoshuaKGoldberg (changes requested):
A good start! Hope this helps with the debugging, but let me know if not. 🙂

packages/website/tests/index.spec.ts
```diff
@@ -1,7 +1,20 @@
...
-  await page.goto('/');
-  await new AxeBuilder({ page }).analyze();
+test.describe('Website', () => {
+  test('Axe', async ({ page }) => {
+    await page.goto('/');
+    await new AxeBuilder({ page }).analyze();
+  });
+
+  test('should have no errors', async ({ page }) => {
+    const errorMessages = [] as string[];
```
JoshuaKGoldberg, line 11:
```suggestion
    const errorMessages: string[] = [];
```

We should really make a lint rule for this. I see people doing it a lot. #6374

packages/website/tests/index.spec.ts
```diff
@@ -1,7 +1,20 @@
...
+    page.on('console', msg =>
+      ['error', 'warning'].includes(msg.type())
+        ? errorMessages.push(`[${msg.type()}] ${msg.text()}`)
+        : null,
+    );
```
JoshuaKGoldberg, line 16: Just nitpicking style here, we generally try to avoid using ternary conditions this way. It's kind of confusing to read (they normally are used to create some result value).

```suggestion
    page.on('console', msg => {
      if (['error', 'warning'].includes(msg.type())) {
        errorMessages.push(`[${msg.type()}] ${msg.text()}`);
      }
    });
```

JoshuaKGoldberg:
Ooh I do get that error locally from `npx nx website start` (also when running it as aliased under `yarn start`)!

Screenshot of Chrome devtools error: 'react-dom.development.js?3c4a:86 Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot'

So this PR will have to account for that message. At least until https://github.com/facebook/docusaurus/issues/7264 comes in.

JoshuaKGoldberg (changes requested):
Requesting changes to get a passing build - but hey, nice to know the new test is working! 😄

packages/website/tests/index.spec.ts
Jefiozie: This would be a great lint rule, maybe I will give it a try after this PR.

Jefiozie:
So now the test are running for the website check. I'm not sure why the other tests are failing. Is this a known issue in the workflows?  If I need to change anything just let me know.

Thanks for the feedback @JoshuaKGoldberg

JoshuaKGoldberg:
Huh, I hadn't seen those failures before. Must be a transient issue in GitHub Actions...

JoshuaKGoldberg (approved):
💯, thanks for iterating on this @Jefiozie! Great stuff. 😄

Issue #5872, opened 2022-10-24
Repo: Standardize debug namespaces per-file


JoshuaKGoldberg:
### Suggestion

Splitting out of https://github.com/typescript-eslint/typescript-eslint/pull/5834#discussion_r1002821063:

> in an ideal world we'd actually do a unique substring per file to help differentiate where logs are coming from.
> 
> eg `typescript-eslint:typescript-estree:inferSingleRun`, `typescript-eslint:typescript-estree:create-program:useProvidedPrograms`, etc
> 
> this also allows you to easily filter the log lines to specific usecases with env vars like `DEBUG=typescript-eslint:typescript-estree:create-program:*`
> 
> also helpful because `debug` also colours each string differently IIRC so it's easy to see different sources when parsing logs.
> 
> I wonder if we can automate creation of the string via like
> 
> ```ts
> const PACKAGE_ROOT = "resolve from disk somehow";
> 
> function getLogger(filename: string) {
>   const relative = path.relative(PACKAGE_ROOT, filename);
>   const parsed = path.parse(relative);
>   return debug(`typescript-eslint:${parsed.dir.replace(/\//g, ':')}:${parsed.name}`);
> }
> 
> // then call this via
> const log = getLogger(__filename);
> ```

It'd be nifty if we could use a lint rule that enforces such a thing.

bradzacher:
> It'd be nifty if we could use a lint rule that enforces such a thing.

Sadly [`eslint-plugin-debug`](https://www.npmjs.com/package/eslint-plugin-debug) doesn't exist :(

We could create an internal rule for us to enforce the style, for sure!

JoshuaKGoldberg:
I tried out the `getLogger(__filename)` strategy in #10598. 

* Advantage: it's very convenient to no longer have to manually massage those strings!
* Disadvantage: the strings are no longer searchable

I think the lint rule will be the way to go. We can make provide auto-fixes so we don't have to manually type strings out.

PR #6367, opened 2023-01-23
feat(typescript-estree): cache project glob resolution

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: https://github.com/typescript-eslint/typescript-eslint/discussions/6366

## Overview

Whilst reviewing https://github.com/typescript-eslint/typescript-eslint/pull/6084 I realised we could save some time in the parser by caching the `parserOptions.project` glob resolution results. So I spent a bit of time throwing this together.

I tested this locally and it saved a little bit of time (~1 second) because our glob is so small (we use `packages/*/tsconfig.json` - which only matches 15 folders.
Though it's still eye-opening to see that the tiny globbing for our repo adds up to an entire second.

I double checked these results by taking a CPU profile before and after:
```bash
$ TSESTREE_SINGLE_RUN=true node --cpu-prof ./node_modules/.bin/eslint .
```
Then I inspected the profiles on https://speedscope.app
For reference - the profiles: [PROFILES.zip](https://github.com/typescript-eslint/typescript-eslint/files/10479581/PROFILES.zip)

Before change:
image

After change:
image

I would expect that this change should show a large perf improvement in the following cases:
- single-star glob that matches many directories (eg `packages/*/tsconfig.json` on something the size of [babel's packages folder](https://github.com/babel/babel/tree/main/packages) or bigger)
- lots of globs matching all over the place (eg `['foo/*/tsconfig.json', 'bar/*/tsconfig.json', 'baz/*/tsconfig.json', ...]`)
- really any case with double star globs


packages/typescript-estree/src/parseSettings/resolveProjectList.ts
```diff
@@ -0,0 +1,142 @@
...
+  getCanonicalFileName,
+} from '../create-program/shared';
+import type { TSESTreeOptions } from '../parser-options';
+import {
+  DEFAULT_TSCONFIG_CACHE_DURATION_SECONDS,
+  ExpiringCache,
+} from './ExpiringCache';
+
+const log = debug(
+  'typescript-eslint:typescript-estree:parser:parseSettings:resovleProjectGlobs',
```
JoshuaKGoldberg, line 18:
```suggestion
  'typescript-eslint:typescript-estree:parser:parseSettings:resolveProjectList',
```

#5872 heh

packages/typescript-estree/src/parseSettings/resolveProjectList.ts
```diff
@@ -0,0 +1,142 @@
...
+    singleRun: boolean;
+    tsconfigRootDir: string;
+  }>,
+): readonly CanonicalPath[] {
+  const sanitizedProjects: string[] = [];
+
+  // Normalize and sanitize the project paths
+  if (typeof options.project === 'string') {
+    sanitizedProjects.push(options.project);
+  } else if (Array.isArray(options.project)) {
```
JoshuaKGoldberg, line 41: Maybe a good place to add a `TODO(#....): log a warning if a project is not a string`, and file an issue?

packages/typescript-estree/src/parseSettings/resolveProjectList.ts
```diff
@@ -0,0 +1,142 @@
...
+    // prefix with a ! for not match glob
+    .map(folder => (folder.startsWith('!') ? folder : `!${folder}`));
+
+  const cacheKey = getHash({
+    project: sanitizedProjects,
+    projectFolderIgnoreList,
+    tsconfigRootDir: options.tsconfigRootDir,
+  });
+  if (RESOLUTION_CACHE == null) {
+    RESOLUTION_CACHE = new ExpiringCache(
```
JoshuaKGoldberg, line 71: Nit:

```suggestion
  if (RESOLUTION_CACHE ??= new ExpiringCache(
```

packages/typescript-estree/src/parseSettings/ExpiringCache.ts
```diff
@@ -0,0 +1,111 @@
...
+  clear(): void {
+    this.map.clear();
+  }
+}
+
+/**
+ * A map with key-level expiration where it is known that there will be multiple
+ * keys with the same value.
+ */
+export class ExpiringCacheWithReverseMap<TKey, TValue> extends ExpiringCache<
```
JoshuaKGoldberg, line 70: This isn't used anywhere. Dead code, to be removed?

packages/typescript-estree/src/parseSettings/ExpiringCache.ts
```diff
@@ -0,0 +1,111 @@
+export const DEFAULT_TSCONFIG_CACHE_DURATION_SECONDS = 30;
+
+/**
+ * A map with key-level expiration.
+ */
+export class ExpiringCache<TKey, TValue> {
+  readonly #cacheDurationSeconds: number | 'Infinity';
```
JoshuaKGoldberg, line 7: This map has to check `this.#cacheDurationSeconds` every time there's a `.get` or `.set`. How about switching to two different maps, with a factory-ish function between them?

```ts
interface CacheMap { ... }

// number case
class ExpiringCacheMap implements CacheMap { ... }

function getCacheMap(duration: number | 'Infinity') {
  return duration === 'Infinity' ? new Map() : new ExpiringCacheMap(duration);
}
```

JoshuaKGoldberg, line 7: > `number | 'Infinity'`

Nit: maybe extract to a shared type, since this gets reused a bunch?

JoshuaKGoldberg (changes requested):
Love to see the perf improvements!

packages/typescript-estree/src/parseSettings/ExpiringCache.ts
bradzacher: I added this ahead of your PR (#6084) just because.
Can remove it for you to copy pasta over there if you'd like!

packages/typescript-estree/src/parseSettings/ExpiringCache.ts
```diff
@@ -0,0 +1,111 @@
+export const DEFAULT_TSCONFIG_CACHE_DURATION_SECONDS = 30;
```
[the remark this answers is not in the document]
bradzacher: I think you might thinking in terms of the same glob being resolved sequentially and this cache short-circuiting that case.
That's not what this cache is for (though it will help with that case).

When we resolve the "parser settings", we resolve the globs to specific tsconfig paths on disk.
Currently every single time ESLint parses a file we do this glob resolution step.
For the persistent case this is "okay" because we treat the disk as unstable - i.e. we assume the globs could match something different (even though 99.999% of the time they don't).
For the single-run case this is BAD because we assume the disk is immutable - so we waste so much time re-resolving globs.

Setting this time too low means we'll resolve globs unnecessarily and not save much time. For example imagine you're typing in your IDE, then pause for 1s to think, then continue typing - that 1s was long enough to invalidate the glob resolution cache, so your next keystroke will be linted and we'll attempt to re-resolve the globs.

So why are we using 30s? I picked it because [eslint-plugin-import uses 30s for their cache](https://github.com/import-js/eslint-plugin-import/blob/766af5ff6c2e97cbe2a2a211729f7c129da3005d/utils/ModuleCache.js#L35-L37). But also it probably strikes a good balance between reactivity and performance in the persistent IDE usecase (and other non-single-run-but-are-actually-single-run cases).

packages/typescript-estree/src/parseSettings/ExpiringCache.ts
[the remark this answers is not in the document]
bradzacher: I chose seconds here just because there's really no need for millisecond precision.
You're never going to want less than 1s cache duration, and nor are you going to want 30.5s cache duration.

Also it aligns well with the `process.hrtime` API - where the first value returned from the method is the number of seconds (and the second is the number of nanoseconds).

packages/typescript-estree/src/parseSettings/ExpiringCache.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Got it, makes sense. 30s it is - thanks for the explanation!

packages/typescript-estree/src/parseSettings/ExpiringCache.ts
JoshuaKGoldberg: Heh, serves you right for trying to be helpful to me 😄. Good to know - I can copy & paste it from the early commit here. But yeah let's remove for now.

packages/typescript-estree/src/parseSettings/resolveProjectList.ts
bradzacher: I split it out like this because it means I can have the separate caching code path run only if the cache isn't new.

No point looking up the cache if the we know for sure the cache is empty!

packages/typescript-estree/src/parseSettings/ExpiringCache.ts
bradzacher: That's a nice idea. But I think i'll leave it as is - just saves the layer of indirection at the top-level and keeps the store a consistent type.
If we allow a dynamic duration in future it saves us undoing this change.

JoshuaKGoldberg (dismissed):
🔥 :shipit: 

_(once build fix)_

PR #6370, opened 2023-01-23
chore: add workflow to automatically run prettier when renovate updates it

JamesHenry:
## PR Checklist

## Overview

If a prettier update causes formatting to change (which it usually does) then right now a maintainer has to manually update the relevant renovate PR by running `yarn format` locally and committing the results.

After this PR goes in, that `yarn format` and commit step will be done automatically on the PR without any maintainer involvement

### DIFF
--- a/.github/workflows/prettier-update.yml
+++ b/.github/workflows/prettier-update.yml
@@ -1,0 +1,46 @@
+# This workflow is used to augment the capabilities of the renovate GitHub app by running a full
+# `prettier --write` when renovate opens a PR to change the version of prettier.
+
+name: Prettier Update
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'package.json'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  maybe_prettier_update:
+    # Only run if it was the renovate bot that triggered the workflow (otherwise we'll create a loop)
+    if: contains('["renovate[bot]"]', github.actor) == true
+    name: Run prettier formatting if required
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Check if prettier was changed as part of the latest commit on the PR
+        id: prettier-package-check
+        run: |
+          git diff HEAD~1 -G"prettier" --exit-code package.json && echo "prettier unchanged" || echo "::set-output name=was-changed::true"
+
+      - name: Run prettier formatting if prettier was changed and commit the results
+        if: ${{ steps.prettier-package-check.outputs.was-changed == 'true' }}
+        env:
+          # We cannot use secrets.GITHUB_TOKEN for this because it is not permitted to kick off subsequent actions worfklow runs, so we use a PAT instead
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (changes requested):
It would have been nice to get this natively in Renovate with [`postUpgradeTasks`](https://docs.renovatebot.com/configuration-options/#postupgradetasks) but it says they're only supported in self-hosted instances 😞...

```json
{
  "postUpgradeTasks": {
    "commands": ["yarn format"],
    "fileFilters": ["**/*"]
  }
}
```

But, this feels like something a lot of repos have to solve in more situations. Surely folks need this whenever they update Prettier? So it's something I think we should tackle more holistically.

What about a post-commit hook to run `yarn format` when the root `package.json` changes? [[their closest example, ignoring filename changes](https://github.com/okonet/lint-staged#example-run-tsc-on-changes-to-typescript-files-but-do-not-pass-any-filename-arguments)]

JamesHenry:
@JoshuaKGoldberg would a local hook be triggered/included in renovate's commit though?

JoshuaKGoldberg:
Hmm, good question. I don't actually know. Never mind me 😄

.github/workflows/prettier-update.yml
```diff
@@ -0,0 +1,46 @@
...
+          git diff HEAD~1 -G"prettier" --exit-code package.json && echo "prettier unchanged" || echo "::set-output name=was-changed::true"
+
+      - name: Run prettier formatting if prettier was changed and commit the results
+        if: ${{ steps.prettier-package-check.outputs.was-changed == 'true' }}
+        env:
+          # We cannot use secrets.GITHUB_TOKEN for this because it is not permitted to kick off subsequent actions worfklow runs, so we use a PAT instead
+          GITHUB_TOKEN: ${{ secrets.JAMES_HENRY_GITHUB_TOKEN }}
+        run: |
+          yarn --ignore-scripts
+          yarn format
```
JoshuaKGoldberg, line 37: For a bit faster install, maybe?

```suggestion
          npx prettier --write .
```

JoshuaKGoldberg (approved):
:shipit: Thanks!

.github/workflows/prettier-update.yml
bradzacher: will `npx` respect the `package.json` version of prettier?

JoshuaKGoldberg: It goes with the latest by default, which should be what the PR is upgrading to anyway.

.github/workflows/prettier-update.yml
```diff
@@ -0,0 +1,46 @@
...
+
+          # Commit all the changes to the PR (see note on not being able to use secrets.GITHUB_TOKEN for this)
+          git config --global user.email "james@henry.sc"
+          git config --global user.name "JamesHenry"
+          git remote set-url origin https://x-access-token:$GITHUB_TOKEN@github.com/$GITHUB_REPOSITORY.git
+
+          git add --all
+          git commit -m "chore: update formatting after prettier upgrade"
```
fisker, line 45: Will this fail if nothing changed.

JamesHenry: Good catch, thanks @fisker

PR #6371, opened 2023-01-24
feat(typescript-estree): use a jump table instead of switch/case for conversion logic

### LINKED ISSUE #6149: Repo: investigate switching core TS->ESTree conversion logic from a "switch/case" to a "jump table" for performance improvements
### Suggestion

https://artemis.sh/2022/08/07/emulating-calculators-fast-in-js.html

I read this a while ago but didn't think about it's possible performance application in our parser.

Our [core conversion logic](https://github.com/typescript-eslint/typescript-eslint/blob/ce2e5dad036087baa9098ad8818a752dec3f6163/packages/typescript-estree/src/convert.ts#L806-L2986) is one massive switch/case (currently 159 cases).

According to [the stackoverflow comment](https://stackoverflow.com/questions/18830626/should-i-use-big-switch-statements-in-javascript-without-performance-problems#comment27798374_18830724) (which has references to the v8 source code) - if you've got more than 128 cases then the switch might not get properly optimised and thus acts slower than hand-written if/else.

One workaround is to use a "jump table" instead of a switch. Jump tables rely on the case value instead being the key of an object - meaning the lookup is instead always as performant as a property lookup (which is usually O(1)).

We should investigate if switching to a "jump table" improves the perf of our parser.

For reference the conversion would be something like this:
```ts
switch (node.kind) {
  case SyntaxKind.SourceFile: // ...
  case SyntaxKind.Block: // ...
  // ...
}

// to

const JUMP_TABLE = {
  [SyntaxKind.SourceFile]: (node: ts.SourceFile) => { /* ... */ },
  [SyntaxKind.Block]: (node: ts.Block) => { /* ... */ },
  // ...
};

JUMP_TABLE[node.kind](node);
```

For this task we'll want to:
1) instrument `typescript-estree` to measure the performance baseline with the current code.
2) make the conversion to a jump table.
3) measure the performance after the change.
4) if (3) < (1), then merge the change.

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6149

## Overview

This PR converts the parser's internal conversion logic from a switch to a jump table.
I tried two different versions of this:
1) direct conversion of the switch to a jump table in the class.
2) conversion of the entire logic from a class to a function, in addition to the jump table conversion

I found the following results:
- Original cumulative AST conversion across our codebase was ~765ms.
- With the changes from (1), conversion time barely moved to ~750ms (~1% reduction).
- With the changes from (2), conversion time moved a bit more to ~685ms (~10% reduction).

I was hoping for a bit more than that, but ultimately that's not a that bad. In the grand scheme of things, however, this is only a ~0.2% reduction in the total lint time for our repo. This is because on our codebase the lint rules themselves take ~70% of the overall lint time - because we use type-aware rules.

TODO:
- [ ] finish the class -> functional conversion and fix tests/types.
- [ ] try this on a larger repo that doesn't use type-aware rules to see how well it translates to runtime reduction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,9 @@ jspm_packages/
 # Editor-specific metadata folders
 .vs
 
+# nodejs cpu profiles
+*.cpuprofile
+
 .DS_Store
 .idea
 dist
--- a/packages/typescript-estree/src/ast-converter.ts
+++ b/packages/typescript-estree/src/ast-converter.ts
@@ -1,7 +1,7 @@
 import type { SourceFile } from 'typescript';
 
 import type { ASTMaps } from './convert';
-import { Converter, convertError } from './convert';
+import { converter, convertError } from './convert';
 import { convertComments } from './convert-comments';
 import { convertTokens } from './node-utils';
 import type { ParseSettings } from './parseSettings';
@@ -25,12 +25,10 @@ export function astConverter(
   /**
    * Recursively convert the TypeScript AST into an ESTree-compatible AST
    */
-  const instance = new Converter(ast, {
+  const { program: estree, ...astMaps } = converter(ast, {
     errorOnUnknownASTType: parseSettings.errorOnUnknownASTType || false,
     shouldPreserveNodeMaps,
   });
-
-  const estree = instance.convertProgram();
 
   /**
    * Optionally remove range and loc if specified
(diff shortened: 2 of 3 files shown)


bradzacher:
I spent some time testing and retesting this on the large codebase at canva (~48k files) where they don't use type-aware linting.
Total saved time with this change was less than 300ms (25.22s -> 24.98s).
Total time for a lint run is a little over 6 minutes (362s), so conversion time is ~6% of the lint run. Time saved with this change was 0.06%.

It really is just insignificant - not worth landing this, closing.

PR #6376, opened 2023-01-25
feat(eslint-plugin): [naming-convention] improve performance by removing unnecessary selectors

bradzacher:
Part of my performance investigations on large codebases.
Right now we always register every single node selector in naming-convention, even if the user's config will never match that selector.

This is bad because eslint registers, parses and evaluates these selectors. All totalled these selectors combined pretty-well add up to every single identifier in the codebase!
Which is a lot of processing that's not required!

It's a simple change that can save 10s of seconds on codebases that don't use the `default` config selector.


packages/eslint-plugin/src/rules/naming-convention.ts
```diff
@@ -395,222 +417,244 @@ export default util.createRule<Options, MessageIds>({
...
 
       // #endregion typeParameter
     };
+
+    return Object.fromEntries(
+      Object.entries(selectors)
+        .map(([selector, { validator, handler }]) => {
+          if (validator == null) {
+            return null;
+          }
```
JoshuaKGoldberg, line 648: Precisely 15,213 test cases, and none with a nullable `validator`!

```suggestion
```

In `packages/eslint-plugin/src/rules/naming-convention-utils/types.ts`:

```diff
type ParsedOptions = Record<SelectorsString, null | ValidatorFunction>;
type ParsedOptions = Record<SelectorsString, ValidatorFunction>;
```

JoshuaKGoldberg (dismissed):
Love it!! ❤️‍🔥

JoshuaKGoldberg (approved):

JoshuaKGoldberg (dismissed):

JoshuaKGoldberg (approved):

PR #6394, opened 2023-01-31
feat(eslint-plugin): [no-import-type-side-effects] add rule to warn against runtime side effects with `verbatimModuleSyntax`

### LINKED ISSUE #6382: Rule proposal: warn against import declarations with no value specifiers and only inline type specifiers

### Description

Spinning off of https://github.com/import-js/eslint-plugin-import/issues/2676

The new [`verbatimModuleSyntax`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#verbatimmodulesyntax) flag will not elide an import declaration if the only specifiers are inline type specifiers:

```ts
import {type A, type B} from 'mod';

// transpiled to

import {} from 'mod';
// which is the same as
import 'mod';
```

This is obviously a problem - code that looks like it is pure type-only code can have runtime side-effects!

After some discussion, we've come to a decision that there needs to exist a rule that warns against this code. The behaviour doesn't really fit into any existing lint rule, even as an option. Additionaly as it's so hyper-targeted on TS behaviour, it doesn't make sense to add this rule to `eslint-plugin-import`.

Thus we should add the rule to our plugin.

At the same time I think we should add documentation about how to use the lint rules together. Copying from the other issue:

1) add a new lint rule which bans imports that have no value specifiers, and only type specifiers
2) anyone who wishes to use inline type specifiers by default and NOT use the TS flag, use `import/consistent-type-specifier-style` together with [`import/no-duplicates`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md#inline-type-imports) with the `{"prefer-inline": true}` config.
3) anyone who wishes to use inline type specifiers by default AND use the TS flag, just use [`import/no-duplicates`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md#inline-type-imports) with the `{"prefer-inline": true}` config in conjunction with the new rule from (1)
4) anyone who wishes to use top-level type specifiers by default (with or without the TS flag), use `import/consistent-type-specifier-style` together with [`import/no-duplicates`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md#inline-type-imports) with the `{"prefer-inline": false}` config. 

### Fail Cases

```typescript
import {type A, type B} from 'mod';
```

### Pass Cases

```typescript
import 'mod';
import * as mod from 'mod';
import type * as mod from 'mod';
import Foo from 'mod';
import type Foo from 'mod';
import {Bar} from 'mod';
import type {Bar} from 'mod';
import Foo, {Bar} from 'mod';
import Foo, {type Bar} from 'mod';
import {Bar, Baz} from 'mod';
import {type Bar, Baz} from 'mod';
```

### Additional Info

https://github.com/typescript-eslint/typescript-eslint/issues/6380

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6382

## Overview

Pretty straight-forward rule to implement with a fixer.


packages/eslint-plugin/docs/rules/no-type-imports-with-verbatim-side-effects.md
```diff
@@ -0,0 +1,74 @@
+---
+description: 'Enforce the use of top-level import type qualifier when an import only has specifiers with inline type qualifiers.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-type-imports-with-verbatim-side-effects** for documentation.
+
+The [`--verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) compiler option causes TypeScript to do simple and predictable transpilation on import declarations, namely it completely removes import declarations with a top-level `type` qualifier, and it removes any import specifiers with an inline `type` qualifier.
```
JoshuaKGoldberg, line 9: [Nit] Smaller sentences for more understandability (my editors would be pleased to see this):

```suggestion
The [`--verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) compiler option causes TypeScript to do simple and predictable transpilation on import declarations.
Namely, it completely removes import declarations with a top-level `type` qualifier, and it removes any import specifiers with an inline `type` qualifier.
```

packages/eslint-plugin/docs/rules/no-type-imports-with-verbatim-side-effects.md
```diff
@@ -0,0 +1,74 @@
...
+import { type A, type B } from 'mod';
+
+// is transpiled to
+
+import {} from 'mod';
+// which is the same as
+import 'mod';
+```
+
+For some rare cases, this may be desirable - but for many cases you will not want to leave behind an unnecessary side effect import.
```
JoshuaKGoldberg, line 23: [Nit] Making it more clear what this is referring to:

```suggestion
For the rare case of needing to import for side effects, this may be desirable - but for most cases you will not want to leave behind an unnecessary side effect import.
```

packages/eslint-plugin/src/configs/all.ts
```diff
@@ -55,7 +55,6 @@ export = {
     'no-dupe-class-members': 'off',
     '@typescript-eslint/no-dupe-class-members': 'error',
     '@typescript-eslint/no-duplicate-enum-values': 'error',
-    'no-duplicate-imports': 'off',
```
JoshuaKGoldberg, line 58: Intentional?

packages/eslint-plugin/src/rules/no-type-imports-with-verbatim-side-effects.ts
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        const specifiers: TSESTree.ImportSpecifier[] = [];
+        for (const specifier of node.specifiers) {
+          if (specifier.type !== AST_NODE_TYPES.ImportSpecifier) {
+            return;
+          }
+          if (specifier.importKind !== 'type') {
+            return;
+          }
```
JoshuaKGoldberg, line 41: [Nit] Not blocking at all, just a thought to save a couple of lines...

```suggestion
          if (
            specifier.type !== AST_NODE_TYPES.ImportSpecifier ||
            specifier.importKind !== 'type'
          ) {
```

packages/eslint-plugin/src/rules/no-type-imports-with-verbatim-side-effects.ts
```diff
@@ -0,0 +1,78 @@
...
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    const sourceCode = context.getSourceCode();
+    return {
+      ImportDeclaration(node): void {
+        if (node.importKind === 'type') {
+          return;
+        }
```
JoshuaKGoldberg, line 32: [Nit] If you don't mind the more complicated query? Not blocking at all.

```suggestion
      'ImportDeclaration[importKind!="type"]'(
        node: TSESTree.ImportDeclaration,
      ): void {
```

packages/eslint-plugin/src/rules/no-type-imports-with-verbatim-side-effects.ts
```diff
@@ -0,0 +1,78 @@
+import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+
+import * as util from '../util';
+
+type Options = [];
+type MessageIds = 'useTopLevelQualifier';
+
+export default util.createRule<Options, MessageIds>({
+  name: 'no-type-imports-with-verbatim-side-effects',
```
JoshuaKGoldberg, line 10: This rule name is a mouthful. What do you think about this?

```suggestion
  name: 'no-type-imports-with-side-effects',
```

I think this is still accurate to what the rule does, but with a little less length of name?

JoshuaKGoldberg (approved):
Great! Nothing blocking from me, only suggestions (and most of them teeny nitpicks). ✨

packages/eslint-plugin/src/rules/no-type-imports-with-verbatim-side-effects.ts
bradzacher: The reason I added `verbatim` is because this rule really only applies with `verbatimModuleSyntax: true`.
I didn't want people to get confused and think that it might apply in other setups.

But yeah the rule name was long. it kind of was a placeholder until I thought of something better then it just "stuck"

packages/eslint-plugin/src/rules/no-type-imports-with-verbatim-side-effects.ts
bradzacher: You could probably fandangle something by chaining `:has` and `:not` or something - but yeah I don't think it's super worth it.

In the beginning I used to love loading a bunch of stuff into the selectors. Now a days I think there's definitely a trade-off with selectors because it's a DSL and they're not the simplest thing to understand.

packages/eslint-plugin/src/configs/all.ts
bradzacher: this happened when I ran `yarn generate-configs`
it was removed because the rule has been deprecated.
IDK why it was missed.

bradzacher:
cc @DanielRosenwasser / @andrewbranch
This rule will be released next Monday.

andrewbranch:
Awesome, thank you!

ambiguous48:
Hi. Just wondering if the following behavior in this new rule is intentional or working as intended. Currently, it's flagging all [side-effect imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_a_module_for_its_side_effects_only). Meanwhile, the docs describe the rule as:

> Enforce the use of top-level import type qualifier when an import **only has specifiers with inline type** qualifiers

It doesn't say anything about imports that have neither `top-level` nor `inline` type imports, such as `side-effect imports`.

If this is intentional, and this rule does act on `side-effect imports`, should we mention that in the docs? Else, would it make sense for this rule to have an option to not act on `side-effect imports`? Or is explicitly ignoring this rule for every instance the recommended approach?

Repro: https://stackblitz.com/edit/node-t1xolf?file=README.md

JoshuaKGoldberg:
@ambiguous48 please [file an issue using one of our issue templates](https://github.com/typescript-eslint/typescript-eslint/issues/new/choose) (I'm guessing [🐛 Report a Bug With a Rule](https://github.com/typescript-eslint/typescript-eslint/issues/new?assignees=&labels=bug%2Cpackage%3A+eslint-plugin%2Ctriage&template=01-bug-report-plugin.yaml&title=Bug%3A+%5Brule+name+here%5D+%3Cshort+description+of+the+issue%3E)). That'll help us address it more quickly, spotlight any missing info, and help us track which version(s) bugs & fixes appear in. Thanks!

JoshuaKGoldberg:
Ah, someone already did! #6431

Which is another point in favor of using the templates. They have useful links around searching for duplicate issues.

bradzacher:
Yup - as per our contributing guide, comments on closed PRs are not the way to ask questions or report bugs.

https://typescript-eslint.io/contributing/issues#commenting

As Josh pointed out - this bug has already been reported.

PR #6397, opened 2023-01-31
feat(eslint-plugin): [prefer-optional-chain] handle cases where the first operands are unrelated to the rest of the chain and add type info

### LINKED ISSUE #1461: [prefer-optional-chain] Only checks the first occurrence in conditional
**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/prefer-optional-chain": "warn"
  }
}
```

```TS
foo && foo.a && bar && bar.a
```

**Expected Result**

```ts
foo?.a && bar?.a
```

**Actual Result**

```ts
foo?.a && bar && bar.a
```

**Additional Info**

**Versions**

| `@typescript-eslint/eslint-plugin` | `2.16.0` |
| `@typescript-eslint/parser`        | `2.16.0` |
| `TypeScript`                       | `3.7.4` |
| `ESLint`                           | `6.8.0` |
| `node`                             | `13.5.0` |
| `npm`                              | `6.13.4` |

### LINKED ISSUE #4820: [eslint-plugin] Breaking: enable type information for prefer-optional-chain

### LINKED ISSUE #5697: Enhancement: [prefer-optional-chain] Suggest optional chain when the start of the expression is already optionally chained

https://typescript-eslint.io/rules/prefer-optional-chain/

### Description

I propose that the `prefer-optional-chain` should suggest combining logical expressions that start with an optional chain

Discovered while making test cases in #5594 

### Fail

```typescript
foo?.bar && foo.bar?.() && foo.bar?.().baz
```

### Pass

```typescript
foo?.bar?.()?.baz
```

### Additional Info

_No response_

### LINKED ISSUE #6332: Enhancement: [`prefer-optional-chain`] handle cases where the first logical(s) are unrelated to the proceeding logicals

https://typescript-eslint.io/rules/prefer-optional-chain

### Description

here is totally safe to refactor as I mention in `Pass`

### Fail

```typescript
if (!baz || !foo || !foo.bar) {
  
}
```

### Pass

```typescript
if (!baz || !foo?.bar) {
  
}
```

### Additional Info

https://typescript-eslint.io/play/#ts=4.9.3&sourceType=module&code=JYMwBAFAhARghgLzAH2WKID2mVo9gOngCcBKMAbwCgwwqBfIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uki9o-aJHBgAviHVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6332, fixes #1461, fixes #4820, fixes #5697
- Supersedes existing PR: closes #6240

## Overview

This is a big refactor of the lint rule so that it can be MUCH more robust and handle more cases.

The new algorithm is based off of flattening the logical expression to an array and then iterating over it. This new algorithm allows us to handle cases like this:
```ts
if (unrelated && foo && foo.bar) {...}
```

The new algorithm also includes a beefy recursive node comparison function which clearly compares two nodes semantically, instead of comparing them loosely based on source code text. This should lead to fewer weird cases, and also allows us to handle cases like this:
```ts
if (foo.bar && foo?.bar.baz) {...}
if (foo?.bar && foo.bar.baz) {...}
```

The TL;DR of the algorithm:
1) Convert the logical chain to a list of logical operators in order
   `foo && foo.bar && foo.bar.baz --> [foo, foo.bar, foo.bar.baz]`
2) Classify each operand based on whether or not it's in a form that does a "nullish comparison" that can be converted to optional chain.
   `foo != null && foo.bar !== undefined && foo.bar.baz --> [foo<NotEqualNullOrUndefined>, foo.bar<NotStrictEqualUndefined>, foo.bar.baz<Boolean>]`
3) Further analyse the chain looking for pairs like `foo !== null && foo !== undefined` and group those as one operand
4) Analyse the chain looking for contiguous sub-chains of "subset comparisons" (eg `foo.bar` is a subset of `foo.bar.baz.bam`, but not of `foo.baz`)
5) When a sub-chain ends, report on it.

TODO:
- [x] #5697 -> handle `foo?.bar && foo.bar`
- [x] add generic logic for getting the "compared name" text in a standardised format (`getComparedName`)
- [x] add logic for analysing `&&` and `||` conditionals separately (`analyzeAndChain` and `analyzeOrChain` respectively)
- [x] delete old selectors
- [x] ensure all existing tests pass
- [x] add tests for new cases
- [x] [??????????? **_add back the fixer_** ???????????](https://github.com/typescript-eslint/typescript-eslint/pull/6397#issuecomment-1510849368)


bradzacher:
This is now "check complete" - all existing tests are passing as are the new tests.
Need to rebuild the fixer logic though.

This is going to be a big chunk of the logic to figure out the correct places to put an optional chain 🤔

bradzacher:
@typescript-eslint/triage-team - this is functionally complete in terms of raw logic.
However it is missing the fixer logic because:
1) frankly it's not a small piece of work to re-implement it.
2) it's the bit that made the rule unsafe to begin with (`x != null && x.y` returns `false | T`, but `x?.y` returns `T | undefined`)

In some future state I can definitely see us including:
- an autofixer if we detect it's safe (eg if it doesn't change the types), or
- a suggestion fixer in other cases

But I don't know if it's worth "holding up" this change to re-implement. Happy to do it if we think it's necessary, if not then we can review and land this as is.

JoshuaKGoldberg:
😞 I do dislike removing functionality. Maybe we could put this out as an experimental rule witout deleting the existing one?

bradzacher:
> Maybe we could put this out as an experimental rule without deleting the existing one?

That would create a worse state, IMO - as we would have two nearly identical rules that are mostly differentiated by having a fixer or not. What would be the value prop for a user to switch to test the rule? We'd need a breaking change to remove the experimental rule name.

Also "experimental" implies instability or that it may not be adopted - but the rule is perfectly stable (passes all the existing tests, and more) and **will** be adopted - it just doesn't have the previously available (unsafe) fixer.

JoshuaKGoldberg:
Yeah that's a good point.

Heh, I'm torn here. Expecting previously (perhaps slightly incorrect) satisfied users to be annoyed by the lack of autofixer.

Given that the autofixer today is unsafe, I think it's reasonable for us to say _"we've rewritten and removed it for safety - welcoming community PRs to add it back"_.

JoshuaKGoldberg:
...maybe we could expose the old one under a name like "deprecated"?

bradzacher:
> I've never even once saw a case of false positive fix of the type object | false.

Also applies to `0 | object` and `'' | object`.
It's rare, for sure, but we've had a few users report it - which means it's not as rare as you might hope in the wild.

The bigger issue is the conversion of return types - `x != null && x.y` returns `false | T`, whereas `x?.y` returns `undefined | T`. In a lot of cases this distinction isn't going to matter - eg in boolean locations like `if` or logical expressions, or in expression statements. But there are places where this can matter and the autofix will cause errors. Again we've had some reports of this!

Honestly I'd be happy with having a flag like `allowPotentiallyUnsafeFixesThatModifyTheReturnTypeIKnowWhatImDoing` which enables people to opt-in to an unsafe fixer - the issue is that it's just a bigger chunk of work to build the fixer.

bradzacher:
okay I have a fixer that's passing for most cases
It was a decent chunk of logic, but not terrible to do.
TL;DR it's like
```
let current = flattenChain(chain[0]);
const parts = [current];

for (i = 1:chain.length) {
  const next = flattenChain(chain[i]);
  parts.push(...next.slice(current.length));
}

fixer.replaceText(chain, stringify(parts));
```

The idea itself is pretty simple, the logic is a little gnarly - but it's working for the most part.

`70 failed, 508 passed, 578 total`
Converted back to a draft until I've got all of them passing.

.cspell.json
```diff
@@ -81,6 +81,7 @@
     "esquery",
     "esrecurse",
     "estree",
+    "falsey",
```
Josh-Cena, line 84: The word should be "falsy"

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -1,669 +1,1521 @@
...
+    // cases so that they don't fall into this generic handling
+    default:
+      return compareByVisiting(nodeA, nodeB);
+  }
+}
+const COMPARE_NODES_CACHE = new WeakMap<
+  TSESTree.Node,
+  WeakMap<TSESTree.Node, NodeComparisonResult>
+>();
+function compareNodes(
```
JoshuaKGoldberg, line 773: FWIW I found it a good bit easier to parse through this file when I moved `compareNodes` and `CompareNodesArgument` to a new file (along with the things only used by `compareNodes`).

bradzacher: Yeah it'd definitely be a good idea to modularise this file so that it is a bit more broken up and isn't as massive.

bradzacher:
Fun edge case I just discovered:
```ts
declare const foo: { bar: number } | null;
foo === null || foo.bar;
foo !== null && !foo.bar;
```

In the first case `foo.bar` is seen by my code as a "truthy boolean" check - which isn't a valid operand in an "OR" chain.
Similarly in the second case `!foo.bar` is seen as a "falsy boolean" check - which isn't a valid operand in an "AND" chain.

Which means that my code will ignore it and try not to include it in the chain to report and fix.

Will need to add special case logic to handle this as it's obviously intended to be chained together.

---

This also opens up another question about how this rule functions (that we can probably punt on for now, but worth mentioning).
Currently the rule will ignore operands with "invalid" checks. For example:
```ts
foo && foo.bar && foo.bar.baz === 1
// rule will currently fix to
foo?.bar && foo.bar.baz === 1
```

Should the rule inspect through these "invalid" checks? Or should it ignore them entirely?
I think there's a subset of them that it should include
- `!==`/`===`/`!=`/`==` comparing against anything
- not `>`/`>=`/`<`/`<=` because TS will hard error if one of the operands is nullish.
- not `typeof` because `foo && typeof foo.bar` will return `undefined | typeof foo.bar` whereas `typeof foo?.bar` will return `typeof foo.bar | 'undefined'` - which is VERY different

JoshuaKGoldberg:
> Should the rule inspect through these "invalid" checks? Or should it ignore them entirely?

...hmm. My instinct is to ignore "invalid" checks as an odd edge case. If users complain we can add it as a bugfix later on - but if the code is roughly not valid, I'd hesitate to feel confident in making changes.

bradzacher:
Note: I hate the tests with a passion because it's impossible for anyone to understand them as written 😢

Snapshot output would make it much simpler to review as you can see the expected reports and fixes properly rather than doing it opaquely.

The test coverage should show how much is covered overall - which should be all of the code!

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
```diff
@@ -0,0 +1,371 @@
...
+  return types.every(t => util.isTypeFlagSet(t, allowedFlags));
+}
+
+export function gatherLogicalOperands(
+  node: TSESTree.LogicalExpression,
+  parserServices: ParserServicesWithTypeInformation,
+  options: PreferOptionalChainOptions,
+): {
+  operands: Operand[];
+  seenLogicals: Set<TSESTree.LogicalExpression>;
```
JoshuaKGoldberg, line 128: [Nitpick] Very nitty nit: maybe rename this to `newlySeenLogicals` so it doesn't need to be renamed in the caller's destructuring?

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
```diff
@@ -0,0 +1,371 @@
...
+                comparedName: comparedExpression,
+                comparisonType: operand.operator.startsWith('!')
+                  ? NullishComparisonType.NotEqualNullOrUndefined
+                  : NullishComparisonType.EqualNullOrUndefined,
+                isYoda,
+                node: operand,
+              });
+              continue;
+            }
+            // x == something :(
```
JoshuaKGoldberg, line 198: [Praise] For some reason, this one made me chuckle 😄

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
```diff
@@ -0,0 +1,371 @@
...
+  if (options.checkBoolean === true) {
+    allowedFlags |= ts.TypeFlags.BooleanLike;
+  }
+  if (options.checkBigInt === true) {
+    allowedFlags |= ts.TypeFlags.BigIntLike;
+  }
+  return types.every(t => util.isTypeFlagSet(t, allowedFlags));
+}
+
+export function gatherLogicalOperands(
```
JoshuaKGoldberg, line 122: [Praise] Separating this function out helped a lot with readability for me, nice.

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
```diff
@@ -0,0 +1,552 @@
...
+        loc: {
+          start: subChain[0].node.loc.start,
+          end: subChain[subChain.length - 1].node.loc.end,
+        },
+        ...getFixer(sourceCode, parserServices, operator, options, subChain),
+      });
+    }
+
+    // we've reached the end of a chain of logical expressions
+    // we know the validated
```
JoshuaKGoldberg, line 501: [Question] Incomplete sentence?

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
```diff
@@ -0,0 +1,552 @@
...
+
+  for (let i = 0; i < chain.length; i += 1) {
+    const lastOperand = subChain[subChain.length - 1] as
+      | ValidOperand
+      | undefined;
+    const operand = chain[i];
+
+    const validatedOperands = analyzeOperand(parserServices, operand, i, chain);
+    if (!validatedOperands) {
+      // TODO - check if the name is a superset/equal - if it is, then it was
```
JoshuaKGoldberg, line 513: > TODO

Is there a tracking issue? _(here and elsewhere)_

JoshuaKGoldberg (approved):
This is really impressive code. Seriously awesome PR @bradzacher!

I've gone over it a few times and think I generally understand it. Particular ups on the approach of separating out the collection and analysis bits.

I say let's get this in for v6!

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
bradzacher: tracking issues are for organised people.
Filed #7170

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1,1274 +1,2134 @@
...
+      'foo && fooBar.baz;',
+      'match && match$1 !== undefined;',
+      "typeof foo === 'number' && foo.toFixed();",
+      "foo === 'undefined' && foo.length;",
+      'foo == bar && foo.bar == null;',
+      'foo === 1 && foo.toFixed();',
+      // call arguments are considered
+      'foo.bar(a) && foo.bar(a, b).baz;',
+      // type parameters are considered
+      'foo.bar<a>() && foo.bar<a, b>().baz;',
```
omril1, line 723: `type parameters` were renamed to type arguments in this context, right?

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1,1274 +1,2134 @@
...
+        code: 'foo() && foo()(bar);',
+        output: 'foo()?.(bar);',
+        errors: [
+          {
+            messageId: 'preferOptionalChain',
+            suggestions: null,
+          },
+        ],
+      },
+      // type parameters are considered
```
omril1, line 1066: Same

Issue #5552, opened 2022-08-26
Repo: drop dependency on tsutils


bradzacher:
### Suggestion

[tsutils](https://github.com/ajafff/tsutils) hasn't had a release (or even a commit) in well over a year (march 2021).

It seems like it's a dead package at this point, sadly - there are a number of issues going unanswered and PRs unreviewed. I guess that the owner has moved on to other things, and unfortunately they were the single point of failure.

This hasn't been a problem thus far, but with TS4.8 one of the types is now wrong and required a local patch to work around (#5551).
There are also some known issues caused by the package, which we cannot address without upstream patches (#5458, #3405).

I see three paths forward:
1) fork the package and release it under our namespace.
1) attempt to take ownership of the package and updated it.
1) bring the relevant utils internal into our `type-utils` package and drop the dependency.

(1) really conflicts with our `type-utils` package, so it seems immediately dead in the water.
(2) could work, but I personally don't have time to maintain it as a separate project - so we'd be relying on someone else for it (which doesn't really change our predicament - just shifts it). Also it would rely on the current owner being responsive or would rely on npm support passing ownership (which takes time).

IMO just dropping the dependency entirely and bringing the relevant utils into our `type-utils` package is the best course of action as it allows us to control the code that we so heavily rely on, which makes it easier for us to maintain everything. Given that we publish them as a package, the open-source community still benefits from our maintained versions.

JoshuaKGoldberg:
I'd actually be interested in taking on tsutils - it's a pretty useful package and I use it in TypeStat. I've tried emailing the maintainer before but haven't heard from them in a bit. Regardless of what we decide here, I'll try to get that conversation started...

bradzacher:
If you can get yourself added as a maintainer with full privileges so you needn't depend on the current owner, then that would work great! I'm confident you'll be sticking around for the forseeable future.

---

In case we do want to drop it as a dep, these are the functions that we use from the package:

- getWellKnownSymbolPropertyOfType
- hasModifier
- isAssignmentKind
- isBinaryExpression
- isBooleanLiteralType
- isCallExpression
- isCompilerOptionEnabled
- isConditionalType
- isIdentifier
- isIntersectionType
- isJsxExpression
- isModifierFlagSet
- isNewExpression
- isObjectFlagSet
- isObjectType
- isParameterDeclaration
- isPropertyAssignment
- isPropertyDeclaration
- isPropertyReadonlyInType
[...]

PR #6404, opened 2023-01-31
feat(typescript-estree): add type checker wrapper APIs to ParserServicesWithTypeInformation

### LINKED ISSUE #6244: Make a "shortcut" API around TypeScript for node types
### Discussed in https://github.com/typescript-eslint/typescript-eslint/discussions/6013

<div type='discussions-op-text'>

<sup>Originally posted by **JoshuaKGoldberg** November  7, 2022</sup>
### Suggestion

We write code like this very frequently:

```ts
const parserServices = util.getParserServices(context);
const checker = parserServices.program.getTypeChecker();
const tsNode = parserServices.esTreeNodeToTSNodeMap.get(node);
const type = checker.getTypeAtLocation(node);
```

It's a little burdensome.

@bradzacher and I were chatting in a pairing about how it might be convenient to make a wrapper around this, for convenience. Vague first draft:

```ts
declare const typeUtils: {
  [K in keyof ts.TypeChecker]:
    (services: ParserServices, node: TSESTree.Node) => ReturnType<ts.TypeChecker[K]>;
};

const type = typeUtils.getTypeAtLocation(parserServices, node);
```

We'd probably want to export this as a publicly available & documented API, so it's not confusing to consumers why our source code doesn't look like examples.

One downside would be that we're obfuscating how to use TypeScript APIs. I hate to add to complexity of understanding these already hard-to-understand things...</div>

---

We ended up with https://github.com/typescript-eslint/typescript-eslint/discussions/6013#discussioncomment-4168807: putting the types right on the parser services.

> I would say we could probably go one step further and just offer all of this off of the parser serivces!
> 
> For example:
> 
> ```ts
> type ParserServices = {
>   getTypeAtLocation(node: TSESTree.Node): ReturnType<ts.TypeChecker['getTypeAtLocation']>;
> };
> 
> const parserServices = util.getParserServices(context);
> 
> parserServices.getTypeAtLocation(node);
> ```
> 
> That would make it even more straight-forward for people to use the tooling!

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6244

## Overview

Adds to `ParserServicesWithTypeInformation` the two most commonly used TypeScript type checking APIs: `getSymbolAtLocation` and `getTypeAtLocation`. Then converts all places in code that I could find that used either after immediately retrieving the TS node for an ES node.

Also applies a few standardizations that helped me find-and-replace:

* `const services = util.getParserServices(context)` - not a variable named `parserServices`, or variables from `{ ... }`
* `const checker = services.program.getTypeChecker()` - not a variable named `typeChecker` or `types`, and not an intermediate `program` variable

Updates the _Custom Rules_ docs page to reflect this.


packages/typescript-estree/src/createParserServices.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  const checker = program.getTypeChecker();
+
+  return {
+    program,
+    ...astMaps,
+    getSymbolAtLocation: node =>
+      checker.getSymbolAtLocation(astMaps.esTreeNodeToTSNodeMap.get(node)),
+    getTypeAtLocation: node =>
+      checker.getTypeAtLocation(astMaps.esTreeNodeToTSNodeMap.get(node)),
```
bradzacher, line 25: do we want to do a weakmap memo on these?
Eg
```
function memoize<TNode, TRet>(fn: (node: TNode) => TRet): (node: TNode) => TRet {
  const cache = new WeakMap<TNode, TRet>();

  return (node) => {
    const cachedRet = cache.get(node);
    if (cachedRet) {
      return cachedRet;
    }

    const newRet = fn(node);
    cachedRet = cache.set(node, newRet);
    return newRet;
  };
}
```

That would allow us to enforce that if two rules request the exact same node then we'll 100% use a fast path.

Doing a quick look at the implementation of `getTypeAtLocation` - it doesn't do any fast caching, so it does do a bunch of work.

IDK if this sort of thing would actually save any real time though - would be worth a test. Maybe we can file an issue to look at this later?

packages/type-utils/src/getConstrainedTypeAtLocation.ts
```diff
@@ -1,14 +1,20 @@
+import type {
+  ParserServicesWithTypeInformation,
+  TSESTree,
+} from '@typescript-eslint/typescript-estree';
 import type * as ts from 'typescript';
 
 /**
  * Resolves the given node's type. Will resolve to the type's generic constraint, if it has one.
  */
 export function getConstrainedTypeAtLocation(
```
bradzacher, line 10: I wonder if this is common enough to add to our new wrapper?

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -19,19 +19,19 @@ export default util.createRule({
...
+        }
+
+        const originalNode = services.esTreeNodeToTSNodeMap.get(node);
 
-        if (
-          !util.isTypeAnyType(type) &&
-          !util.isTypeUnknownType(type) &&
-          !tsutils.isThenableType(checker, originalNode.expression, type)
-        ) {
+        if (!tsutils.isThenableType(checker, originalNode.expression, type)) {
```
bradzacher, line 34: TODO - kill tsutils dependency #5552

bradzacher (approved):
I'm happy with this! Noice

packages/eslint-plugin/src/rules/await-thenable.ts
JoshuaKGoldberg: 🥲

packages/type-utils/src/getConstrainedTypeAtLocation.ts
JoshuaKGoldberg: I think for now I'd rather steer clear of augmenting with our own APIs. I like sticking to 1:1 wrappers for simplicity's sake & having fewer API shapes to remember.

Maybe we'll eventually add caching for type functions such as `getBaseConstraintOfType`?

packages/typescript-estree/src/createParserServices.ts
JoshuaKGoldberg: Yeah I'll file an issue and do some perf measuring. Exciting!

JoshuaKGoldberg: #6425

PR #6411, opened 2023-02-02
feat(eslint-plugin): [prefer-at] create rule

### LINKED ISSUE #6401: Rule proposal: [prefer-at] Rule to prefer array.at(-1) over array[array.length - 1]

### Description

#6211 shows a use of `array.at(-1)`, as a cleaner equivalent to `array[array.length - 1]`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

Proposal: let's add a typed lint rule that, when it sees something like `array[array.length - 1]` on an Array-typed value, proposes fixing to `array.at(-1)`?

### Fail Cases

```typescript

array[array.length - 1];
```

### Pass Cases

```typescript

array.at(-1);

// [] indices are reasonable
array[0];

// Don't look through functions
const getArray = () => array;
getArray()[getArray().length - 1];
```

### Additional Info

Reference for an existing similar rule: https://typescript-eslint.io/rules/prefer-includes.

I don't think we should enable this in _recommended_ or _strict_ configs until we only support Node versions that include `.at`. Per #6396 this wouldn't make sense for users who compile to Node 14 and don't include a polyfill.

sviat9440:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6401

## Progress

- [x] Add type checks for `Array`, `String`, `Uint8Array` ...
- [x] Add `ignoreFunctions` option (default to `false`)
- [x] Add support for -2, -3 ... -i
- [x] Clear tests
- [x] Add tests for `this` reference
- [x] Add tests for functions
- [x] Add more examples for documentation
- [x] Fix coverage
- [x] Add tests for custom `String`, `Array` ...


packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,77 @@
...
+            node.property.type !== AST_NODE_TYPES.BinaryExpression ||
+            node.property.operator !== '-' ||
+            node.property.right.type !== AST_NODE_TYPES.Literal ||
+            node.property.right.value !== 1
+          ) {
+            return;
+          }
+          const objectName = getName(node.object);
+          const propertyLeftName = getName(node.property.left);
+          if (`${objectName}.length` === propertyLeftName) {
```
coyotte508, line 58: Correct me if I'm wrong - but we should check that `node.object` is of type `Array`, `String`, `Uint8Array`...?

Otherwise, this from the associated issue:

>  [x] My proposal specifically checks TypeScript syntax, or it proposes a check that requires type information to be accurate.

It's not true, it could be a regular eslint rule.

sviat9440:
[ ] Add type checks for `Array`, `String`, `Uint8Array` ...
[ ] Add test cases from associated issue
[ ] Add `ignoreFunctions` option (default to `true`)
[ ] Add support for -2, -3 ... -i

packages/eslint-plugin/tests/rules/prefer-at.test.ts
```diff
@@ -0,0 +1,475 @@
...
+
+interface CodeGenerator {
+  valid: string;
+
+  invalid: InvalidCodeWithMetadata;
+
+  name: string;
+}
+
+class VariableDeclaration implements CodeGenerator {
```
JoshuaKGoldberg, line 119: [Request] This code to generate tests is nifty, but really hard for me to read through. I'm starting to feel more and more negative about autogenerated tests over time - it's hard to identify what all the tests are, and the code tends to be moderately unreadable.

Oftentimes, code that's using generation like this for tests is creating a lot of redundant cases. I.e. if there are cases `A`, `B`, `C`, and `D`, they create `A * B * C * D`. But there often don't need to be so many tests for all possible permutations! Just individual tests for each of them (`A + B + C + D`).

Could you please change this file to have string literals for the tests? If there's a reason that's not practical we can discuss, but my guess for this small a rule we don't need so much testing complication.

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,219 @@
...
+        case AST_NODE_TYPES.PrivateIdentifier:
+          return `#${node.name}`;
+        case AST_NODE_TYPES.Identifier:
+          return node.name;
+        case AST_NODE_TYPES.Literal:
+          return node.raw;
+        case AST_NODE_TYPES.ThisExpression:
+          return 'this';
+        case AST_NODE_TYPES.CallExpression:
+          return `${getFullName(node.callee)}()`;
```
JoshuaKGoldberg, line 77: [Question] Just making sure: is there a unit test for the case of a call expression with arguments?

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    function isExpectedExpressionRight(
+      node: TSESTree.BinaryExpression,
+    ): boolean {
+      const type = getTypeAtLocation(node.right);
+      return (
+        // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+        type?.isNumberLiteral() || type?.getFlags() === ts.TypeFlags.Number
+      );
```
JoshuaKGoldberg, line 176: [Request] We can bypass the lint rule shenanigan with a helpful utility:

```suggestion
      return tsutils.isTypeFlagSet(type, ts.TypeFlags.NumberLike);
```

Anywhere that checks type flags with a `===` should probably use a `tsutils` helper instead. We've seen odd edge cases with number literals vs. numbers before.

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,219 @@
...
+        node.object.type === AST_NODE_TYPES.CallExpression ||
+        (node.object.type === AST_NODE_TYPES.MemberExpression &&
+          hasCallExpression(node.object))
+      );
+    }
+
+    function getTypeAtLocation(node: TSESTree.Node): ts.Type | undefined {
+      const originalNode = parserServices.esTreeNodeToTSNodeMap.get(node);
+      if (!originalNode) {
+        return undefined;
```
JoshuaKGoldberg, line 99: [Request] Indeed, the type system says `originalNode` is never `undefined`, and this line is never hit by tests. It can be removed. This function returns `ts.Type`.

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,219 @@
...
+          const directory = path.join(sourceFile.fileName, '../');
+          return directory.endsWith(path.join('node_modules/typescript/lib/'));
+        }) ?? false
+      );
+    }
+
+    function isExpectedExpressionLeft(
+      node: TSESTree.BinaryExpression,
+    ): boolean {
+      if (!isExpectedObject(node.left) || getName(node.left) !== 'length') {
```
JoshuaKGoldberg, line 162: [Question] Similar lint rules to this one use the `getStaticValue` from `eslint-utils`. Have you tried using it in this file? I'm wondering if it'll help do some of the logical checks that are implemented manually right now.

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,219 @@
...
+      ) {
+        return false;
+      }
+      const type = getTypeAtLocation(node.object);
+      if (type && !isSupportedObject(type)) {
+        return false;
+      }
+      const atMember = type?.getProperty('at');
+      return (
+        atMember?.declarations?.every(declaration => {
```
JoshuaKGoldberg, line 151: > `.every`

So if a project adds its own `.at` method augmentation in their own file, this will return false? That's a pity because `@types/node` does that very thing.

[Request] I'm also betting this rule doesn't need to check that the `at` is built-in. Other similar rules such as `prefer-includes` don't. See how that one and other rules that use `getStaticValue` operate.

packages/eslint-plugin/docs/rules/prefer-at.md
```diff
@@ -0,0 +1,101 @@
+---
+description: 'Enforce the use of `array.at(-1)` instead of `array[array.length - 1]`.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/prefer-at** for documentation.
+
+There are two ways to get the last item of the array:
```
JoshuaKGoldberg, line 9: [Proofreading]

```suggestion
There are two common ways to get the last item of an array:
```

packages/eslint-plugin/docs/rules/prefer-at.md
```diff
@@ -0,0 +1,101 @@
...
+- `Uint8Array`
+- `Uint8ClampedArray`
+- `Int16Array`
+- `Uint16Array`
+- `Int32Array`
+- `Float32Array`
+- `Uint32Array`
+- `Float64Array`
+- `BigInt64Array`
+- `BigUint64Array`
```
JoshuaKGoldberg, line 30: [Proofreading] Folks probably don't care about the exact full list of these integer arrays, just that they're all supported.

```suggestion
`arr.at(-1)` is a cleaner equivalent to `arr[arr.length - 1]`.

This rule suggests code use `arr.at(-1)`, as it's cleaner than `arr[arr.length - 1]`.
It supports built-in array and string types, including integer arrays such as `Int8Array`.
```

packages/eslint-plugin/docs/rules/prefer-at.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Options
+
+```ts
+interface Options {
+  ignoreFunctions?: boolean;
+}
+```
```
JoshuaKGoldberg, line 38: [Request] No need to explicitly spell out `interface Options`, it's auto-generated. See http://localhost:3000/rules/prefer-at locally (you can start the web server with `yarn start` from the monorepo root).

It would be good to have a text explanation of what `ignoreFunctions` does though.

packages/eslint-plugin/docs/rules/prefer-at.md
```diff
@@ -0,0 +1,101 @@
...
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/prefer-at** for documentation.
+
+There are two ways to get the last item of the array:
+
+- `arr[arr.length - 1]`: getting an item by index calculated relative to the length of the array.
+- `arr.at(-1)`: getting an item by the `at` method. [Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at)
```
JoshuaKGoldberg, line 12: [Proofreading] Informative link titles makes it easier to navigate through the page with a screenreader. And this is a bit less text to read:

```suggestion
- `arr.at(-1)`: getting an item by the [`at` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at).
```

packages/eslint-plugin/docs/rules/prefer-at.md
```diff
@@ -0,0 +1,101 @@
...
+let b = getArr()[getArr().length - 1];
+```
+
+<!--/tabs-->
+
+## When Not To Use It
+
+If you support a browser that does not match
+[this table](https://caniuse.com/mdn-javascript_builtins_array_at)
+or use `node.js < 16.6.0` and don't include the polyfill
```
JoshuaKGoldberg, line 101:
```suggestion
If you support a runtime environment that doesn't support [ES2022's `.at`](https://caniuse.com/mdn-javascript_builtins_array_at) and don't include a polyfill, you can ignore this rule.
```

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,219 @@
...
+        oneOf: [
+          {
+            type: 'object',
+            properties: {
+              ignoreFunctions: {
+                type: 'boolean',
+              },
+            },
+            additionalProperties: false,
+            required: ['ignoreFunctions'],
```
JoshuaKGoldberg, line 34: [Bug] This isn't required. 

```suggestion
```

JoshuaKGoldberg (changes requested):
A nice start, thank you for sending it in!

JoshuaKGoldberg:
👋 Hey @sviat9440! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

sviat9440:
Hey, I'll do it next weekend

sviat9440:
@JoshuaKGoldberg fixed

packages/eslint-plugin/src/rules/prefer-at.ts
JoshuaKGoldberg: Bump for this unresolved thread @sviat9440 ?

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,177 @@
...
+          return;
+        }
+        context.report({
+          messageId: 'preferAt',
+          data: {
+            name: objectName,
+          },
+          node,
+          fix: fixer =>
+            fixer.replaceText(node, `${objectName}.at(-${rightName})`),
```
JoshuaKGoldberg, line 172: [Question] Similar lint rules such as `prefer-regexp-exec` and `prefer-string-starts-ends-with` use a util called `getWrappingFixer`. Have you considered using it here? There are some edge cases in their tests I think we'd want for this one.

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,177 @@
...
+    const sourceCode = context.getSourceCode();
+
+    function getName(node: TSESTree.Node): string | undefined {
+      switch (node.type) {
+        case AST_NODE_TYPES.Identifier:
+          return node.name;
+        case AST_NODE_TYPES.MemberExpression:
+          return getName(node.property);
+        default:
+          return undefined;
```
JoshuaKGoldberg, line 56: [Testing] Still missing some coverage here. See https://app.codecov.io/gh/typescript-eslint/typescript-eslint/pull/6411/blob/packages/eslint-plugin/src/rules/prefer-at.ts for three lines that are yellow (partial miss) or red (full miss).

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,177 @@
...
+      checkObjectName('BigInt64Array'),
+      checkObjectName('BigUint64Array'),
+      // eslint-disable-next-line @typescript-eslint/internal/prefer-ast-types-enum
+      checkObjectName('String'),
+      checkObjectType(ts.TypeFlags.String),
+    ];
+
+    function isSupportedObject(type: ts.Type): boolean {
+      return supporterObjects.some(check => check(type));
+    }
```
JoshuaKGoldberg, line 108: [Refactor] This `supporterObjects` schema took me a while to read through. If I understand it right, it roughly:

1. Checks if the type's symbol is one of the supported names
2. Checks if the object has `ts.TypeFlags.String` in its flags

If that's roughly right, two change suggestions: 

* I think it'd be a lot simpler to move those names into a `Set<string | undefined>` and use `.has`
* We normally use `tsutils.isTypeFlagSet` for checking type flags (as you've already done elsewhere in this file) - why not use it here?

```js
supportedObjectNames.has(type.getSymbol()?.name) || tsutils.isTypeFlagSet(type, ts.TypeFlags.String);
```

Thoughts?

JoshuaKGoldberg (changes requested):
Thanks for the updates! There's still a thread from before - waiting on hearing back on that one before going deeper :).

sviat9440:
@JoshuaKGoldberg waiting for your reply

JoshuaKGoldberg:
@sviat9440 still waiting for your response on https://github.com/typescript-eslint/typescript-eslint/pull/6411/files#r1174752364 and https://github.com/typescript-eslint/typescript-eslint/pull/6411/files#r1174752796

sviat9440:
@JoshuaKGoldberg Are you kidding? I asked you a question in this thread. I will close the PR if there is no response from you.

JoshuaKGoldberg:
Aha! I see what the issue is - you posted the comments as _Pending_, in a review that hasn't been posted. So I hadn't seen them yet. Love it.

screenshot of the same thread but with only my comments visible


I'll take a look soon. Thanks for posting the screenshot! If there are other comments it'd be great if you could submit the review so I can see them.

sviat9440 (commented):
Oh, I have 6 pending comments... Interestingly, some comments are published immediately, and some are not. Strange. I will keep in mind in the future.

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,177 @@
...
+      return isExpectedExpressionRight(node) && isExpectedExpressionLeft(node);
+    }
+
+    return {
+      'MemberExpression[property.type="BinaryExpression"][property.operator="-"]'(
+        node: TSESTree.MemberExpressionComputedName & {
+          property: TSESTree.BinaryExpression & { operator: '-' };
+        },
+      ): void {
+        if (!isExpectedExpression(node.property)) {
```
JoshuaKGoldberg, line 156: [Naming] I'm finding hard to read the code with the naming convention of `isExpected*`. It's a little vague - what does _'expected'_ mean? Until you fully understand the code's intent that can be hard to determine it.

What do you think about renaming them to more explicitly say what they do? E.g. instead of `isExpectedExpression`, something like `isArrayLengthBasedLookup`?

packages/eslint-plugin/src/rules/prefer-at.ts
```diff
@@ -0,0 +1,177 @@
...
+        if (objectName !== memberName) {
+          return;
+        }
+        context.report({
+          messageId: 'preferAt',
+          data: {
+            name: objectName,
+          },
+          node,
+          fix: fixer =>
```
JoshuaKGoldberg, line 171: [Bug] Just remembered, `.at`'s signature returns `T | undefined`. That's a change from the original type of `T` from `[]` lookups. We generally put anything that can break the build into _suggestions_ fixers rather than direct _fix_ fixers. So this should be a _suggestion_ fixer.

packages/eslint-plugin/src/rules/prefer-at.ts
JoshuaKGoldberg: Behaviorally, it should help catch cases like this:

* `str[str['length'] - 1]`
* `const key = 'length'; str[str[key] - 1]`

For code, it would eliminate the need to manually write a `function getName`. Which is something a lot of our rules end up wanting. So we'd really rather than write explicit _"what is the string name of this property?"_ logic repeatedly without a real need.

https://eslint-community.github.io/eslint-utils/api/ast-utils.html#getstaticvalue

JoshuaKGoldberg: (also, for anybody seeing just this thread and wondering why we're bumping at each other 😂 the comments were pending)

JoshuaKGoldberg (changes requested):
👍 progress!

JoshuaKGoldberg:
👋 Hey @sviat9440! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

PR #6432, opened 2023-02-07
fix(eslint-plugin): [no-unnecessary-condition] account for optional chaining on potentially void values

### LINKED ISSUE #6261: Bug: [no-unnecessary-condition] Does not account for optional chaining on potentially-`void` values

### Playground Link

https://typescript-eslint.io/play/#ts=4.9.4&sourceType=module&code=DYUwLgBAbghgTgSxgI1BAvBARCgxliGAZwiLEQDsBzCAH2gHsEATAbgFgAoL2RFUAPwA6ANoAGALqsgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK+SpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A

### Repro Code

```typescript
let variable = "abc" as string | void;

variable?.[0];
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true
  }
}
```

### Expected Result

No problems

### Actual Result

The third line reports an error - `Unnecessary optional chain on a non-nullish value.`

### Additional Info

Related to #5766. That PR did not extend to the optional chaining part of this rule.
This is caused by https://github.com/typescript-eslint/typescript-eslint/blob/ccd45d4a998946b7be1161f8c8216bc458e50b4e/packages/type-utils/src/predicates.ts#L28 not checking for `void`, only `null` and `undefined`.

cparros:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6261

## Overview

This PR adds a check for `void` types in the [no-unnecessary-condition] rule. It also adds a valid test case to its valid tests array.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -126,8 +126,7 @@ export default createRule<Options, MessageId>({
       noOverlapBooleanExpression:
         'Unnecessary conditional, the types have no overlap.',
       never: 'Unnecessary conditional, value is `never`.',
-      neverOptionalChain:
-        'Unnecessary optional chain on a non-nullish value. Please ensure you are not declaring a variable with the `void` type outside of a return type or generic type argument. Use `undefined` instead.',
+      neverOptionalChain: 'Unnecessary optional chain on a non-nullish value.',
       noStrictNullCheck:
         'This rule requires the `strictNullChecks` compiler option to be turned on to function correctly.',
     },
@@ -575,10 +574,15 @@ export default createRule<Options, MessageId>({
         node.type === AST_NODE_TYPES.MemberExpression
           ? !isNullableOriginFromPrev(node)
           : true;
+      const unionParts = unionTypeParts(type);
+      const possiblyVoid = unionParts.some(
+        part => part.flags & ts.TypeFlags.Void,
+      );
       return (
         isTypeAnyType(type) ||
         isTypeUnknownType(type) ||
-        (isNullableType(type, { allowUndefined: true }) && isOwnNullable)
+        (isNullableType(type, { allowUndefined: true }) && isOwnNullable) ||
+        (possiblyVoid && isOwnNullable)
       );
     }
 
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -126,7 +126,8 @@ export default createRule<Options, MessageId>({
       noOverlapBooleanExpression:
         'Unnecessary conditional, the types have no overlap.',
       never: 'Unnecessary conditional, value is `never`.',
-      neverOptionalChain: 'Unnecessary optional chain on a non-nullish value.',
+      neverOptionalChain:
+        'Unnecessary optional chain on a non-nullish value. Please ensure you are not declaring a variable with the `void` type outside of a return type or generic type argument. Use `undefined` instead.',
       noStrictNullCheck:
         'This rule requires the `strictNullChecks` compiler option to be turned on to function correctly.',
     },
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
```diff
@@ -545,6 +545,10 @@ type OptionalFoo = Foo | undefined;
 declare const foo: OptionalFoo;
 foo?.[1]?.length;
     `,
+    `
+let variable = 'abc' as string | void;
+variable?.[0];
+    `,
```
JoshuaKGoldberg, line 551: I looked around the test file and found a few similar tests. Example: `let foo: undefined | { bar: true };`. Could you move this near them?

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -574,10 +574,15 @@ export default createRule<Options, MessageId>({
...
+      const unionParts = unionTypeParts(type);
+      const possiblyVoid = unionParts.some(
+        part => part.flags & ts.TypeFlags.Void,
+      );
       return (
         isTypeAnyType(type) ||
         isTypeUnknownType(type) ||
-        (isNullableType(type, { allowUndefined: true }) && isOwnNullable)
+        (isNullableType(type, { allowUndefined: true }) && isOwnNullable) ||
+        (possiblyVoid && isOwnNullable)
```
JoshuaKGoldberg, line 585: The `&& isOwnNullable` is duplicated here. Let's simplify:

```ts
(isOwnNullable && (isNullableType(type, { allowUndefined: true }) || possiblyVoid))
```

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -574,10 +574,15 @@ export default createRule<Options, MessageId>({
         node.type === AST_NODE_TYPES.MemberExpression
           ? !isNullableOriginFromPrev(node)
           : true;
+      const unionParts = unionTypeParts(type);
+      const possiblyVoid = unionParts.some(
+        part => part.flags & ts.TypeFlags.Void,
+      );
       return (
         isTypeAnyType(type) ||
         isTypeUnknownType(type) ||
```
JoshuaKGoldberg, line 583: Interesting: `isTypeAnyType(type) and `isTypeUnknownType(type)` both are working fine even if the type is a union. Both of those functions narrow down to checking `isTypeFlagSet(type, ts.TypeFlags.Something)`.

I tried switching from the `possiblyVoid` variable to checking `isTypeFlagSet(type, ts.TypeFlags.Void` - and it seems to work! So I think my hunch about using `unionTypeParts` was wrong 😄.

Tip: TypeScript's flags enums such as `ts.TypeFlags` work with checks like `isTypeFlagSet(type, ts.TypeFlags.One | ts.TypeFlags.Two)`. So you should be able to replace `isTypeAnyType(type) || isTypeUnknownType(type)` with a single `isTypeFlagSet` check.

JoshuaKGoldberg (dismissed):
A good start - glad it's only a little bit of code to fix the bug! Let's get it to _even less_. 🚀

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
cparros: Moved!

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
cparros: Updated

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
cparros: This is great to know! Done.

bradzacher (approved):
this is looking good to me!
thanks for fixing this!

seemsreasonable

PR #6436, opened 2023-02-07
chore: migrated `shared-fixtures` to `ast-spec`

### LINKED ISSUE #6065: Repo: move `shared-fixtures` tests to `ast-spec` fixtures
### Suggestion

In #3258 we added a new testing framework for our AST parsing tests.
We did this for a few reasons:
1) it allows us to colocate tests alongside the AST definitions - making the ast-spec self documenting of sorts
    - i.e. "this AST node is produced by this code"
    - as an aside - in some future state we'd love to generate documentation on our website using the fixtures - making it truly self documenting!
2) it allows us to automatically diff against the babel AST to clearly document differences
    - in the `shared-fixtures` tests we rely on sets of [manual patches](https://github.com/typescript-eslint/typescript-eslint/blob/44027db379e6e074a16cda2755ef554c2b0a4c5a/packages/typescript-estree/tests/ast-alignment/utils.ts) to the AST to get our "alignment" tests to pass (which is obviously pretty hacky and hard to maintain)
    - in the `shared-fixtures` tests we also rely on a [set of ignores](https://github.com/typescript-eslint/typescript-eslint/blob/44027db379e6e074a16cda2755ef554c2b0a4c5a/packages/typescript-estree/tests/ast-alignment/fixtures-to-test.ts) to document and skip fixtures we know don't align.
        - This isn't great because there's no checks to enforce an ignored fixture fails like it should (meaning it's easy to forget to un-ignore a fixture when it should pass)
        - This also doesn't document what's actually different - we usually comment with an explanation, but that doesn't document the actual AST difference!

With a new framework comes a migration! We need to create fixtures for all of our AST nodes to test all the many combinations!
There's already a number of examples present in the the `ast-spec` package (for example all of the [`declaration`](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/ast-spec/src/declaration) nodes are done), but to walk through the process:

First, setup the fixture folder:
1) open any AST node folder (any folder with a `spec.ts` eg `packages/ast-spec/src/declaration/ClassDeclaration`).
2) create a `fixtures` folder

Next, add "success" fixture cases:
1) within your `fixtures` folder, create a folder with a short kebab-cased name that describes the case you're covering
    - For example if you're testing the AST for a class declaration that implements one interface you might name it `implements-one`. Each fixture should only cover one hyper-targeted, specific case to keep things easy to debug and review.
2) within your new folder, create a file called `fixture.ts` (eg `packages/ast-spec/src/declaration/ClassDeclaration/implements-one/fixture.ts`) and fill this file with your test case.
     - As mentioned, the code should be hyper-targeted! Put another way - don't try to make your code "semantically" correct (i.e. it doesn't have to pass a type check!) - it just has to be syntactically correct.
       For example - in our "Class declaration`implements-one` test case, we shouldn't add a matching interface declaration!
3) run the test to generate the snapshots: `cd packages/ast-spec && yarn jest /fixtures`
     - This will generate a `snapshots` subfolder next to your `fixture.ts` that contains a number of snapshots generated from your test. At the time of writing that's:
         1) `1-TSESTree-AST.shot` - the AST output from parsing your fixture (parsed using our `typescript-estree` parser)
         2) `2-TSESTree-Tokens.shot` - the AST tokens parsed out of your fixture (parsed using our `typescript-estree` parser)
         3) `3-Babel-AST.shot` - same as (1) but parsed using babel
         4) `4-Babel-Tokens.shot` - same as (2) but parsed using babel
         5) `5-AST-Alignment-AST.shot` - the diff of (1) and (3)
         6) `6-AST-Alignment-Tokens.shot` - the diff of (2) and (4)
      - If one of the parser throws an error on your fixture then your fixture should instead be an error fixture (see below).
      - If there is a difference in the AST, you'll notice your fixture is auto-added to the snapshot [`fixtures-with-differences-errors.shot`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/tests/fixtures-with-differences-errors.shot). This is intentional - it's essentially an automatically generated register of the places where we our parser and babel's don't align!

To add "error" fixture cases:
1) within your `fixtures` folder, create a folder called `_error_`.
2) Follow the above "success" guide (create a kebab-case folder within `_error_` and then a `fixture.ts` within that folder, then run jest).
3) The snapshots generated will be a little different - they're there to document the thrown errors instead of the AST output! So you'll see:
    1) `1-TSESTree-Error.shot` - the error caught during the parse or `"NO ERROR"` if no error was thrown (parsed using our `typescript-estree` parser)
    2) `2-Babel-Error.shot` - same as (1) but parsed with babel
    3) `3-Alignment-Error.shot` - a string describing the error state:
       - `"Babel errored but TSESTree didn't"`
       - `"TSESTree errored but Babel didn't"`
       - `"Both errored"`

When creating fixtures it's important to stress that each fixture should be as hyper-focused as you can make it! The more code there is, the more AST gets generated and the harder it is for a human to parse the AST output to find the relevant bits of the AST. More code also means more surface area for changes to impact which means that there's more and more snapshots that unnecessarily need updating when we make parser changes (which just creates noise!).

***How to decide what to create a fixture for?*** The best way it to look at the spec for the AST node and try to create a fixture that targets each case for each property. For example `ClassDeclaration` has a boolean property `abstract: boolean` - so we want to ensure we have at least one fixture which generates a `true` value for the property, and one that generates a `false` value for the property. Again - fixtures are cheap and easy to create, so if in doubt, just add another fixture to cover a case you're thinking about.

Make sure you consider both good and bad cases! Even though our parser doesn't currently throw many syntax errors (#1852), it's good practice to also add bad code fixtures so that we can clearly document the expected current behaviour of our parser and how it aligns with babel!

Finally - don't endeavour to cover every single case in a single PR (unless you really, really want to) - feel free to just create one PR that covers one or a small number of nodes. The beauty of this framework is that we don't need to do everything at once!

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6065

## Overview

Migrates all the old `shared-fixtures` fixtures into the new `ast-spec` fixtures. Within that they're given the same folder name as their old fixture path.

It would be better long-term to split these fixtures up as described in #6065:

> Each fixture should only cover one hyper-targeted, specific case to keep things easy to debug and review.

...but there are ~375 of them, and I'd rather get them into `ast-spec` sooner. This way they'll be easier to split up later! If this PR is approved I can file a followup issue & find-and-replace the `// TODO` comments in fixtures to refer to it.

Most of the work in this PR was done in two steps:

1. `327aa0c29`: Ran `yarn tsx tools/migrate-fixtures.ts` to create `unsorted-fixtures` from the old `shared-fixtures`
2. `bec1fe800`: Deleted `shared-fixtures` altogether

<details>
<summary><code>tools/migrate-fixtures.ts</code></summary>

```ts
import * as cp from "node:child_process";
import { existsSync, promises as fs } from "node:fs";
import { tmpdir } from "node:os";
import path from "node:path";
import { promisify } from "node:util";

import glob from "glob";

const exec = promisify(cp.exec);

const newFixturesDir = "packages/ast-spec/src/unsorted-fixtures";

const oldFixtures = glob.sync("packages/shared-fixtures/fixtures/**/*.ts");

async function main(): Promise<void> {
  // Make sure the new fixtures directory exists, and has no prior contents
  if (existsSync(newFixturesDir)) {
    await fs.rm(newFixturesDir, { recursive: true });
  }

  await fs.mkdir(newFixturesDir, {
    recursive: true,
  });

  // Copy each of the old fixtures into the new fixtures directory
  for (const fixture of oldFixtures) {
    const [oldFileName, group] = fixture.split("/").reverse();
    const fixtureName = oldFileName.split(".")[0];
    const fixtureDirectory = path.join(
      newFixturesDir,
      group,
      "fixtures",
      fixtureName
    );

    await fs.mkdir(fixtureDirectory, { recursive: true });
    await fs.writeFile(
      path.join(fixtureDirectory, "fixture.ts"),
      await fs.readFile(fixture)
    );
  }

  // Run Jest on ast-spec to find new fixtures with syntax errors
  const tempFile = path.join(tmpdir(), "temp.json");
  try {
    await exec(`jest /fixtures --json > ${tempFile}`, {
      cwd: "packages/ast-spec",
    });
  } catch {
    /* expected */
  }
  const jestOutput = JSON.parse((await fs.readFile(tempFile)).toString()) as {
    testResults: {
      assertionResults: { ancestorTitles: string[]; status: string }[];
    }[];
  };
  await fs.rm(tempFile);

  // Filter & deduplicate messages from those failing tests
  const failingFixtures = Array.from(
    new Set(
      jestOutput.testResults[0]
        .filter(
          (assertion) =>
            assertion.ancestorTitles.includes("unsorted-fixtures") &&
            assertion.status !== "passed"
        )
        .map((assertion) => assertion.ancestorTitles.slice(2).join("/"))
    )
  );

  // For each failing fixture, move it to an _error_ directory
  for (const failingFixture of failingFixtures) {
    const failingFixtureDirectory = path.join(
      newFixturesDir,
      failingFixture.replace("/", "/fixtures/")
    );

    const newErrorFixtureDirectory = path.join(
      failingFixtureDirectory.replace("/fixtures/", "/fixtures/_error_/")
    );
    await fs.mkdir(newErrorFixtureDirectory, {
      recursive: true,
    });
    await fs.rename(
      path.join(failingFixtureDirectory, "fixture.ts"),
      path.join(newErrorFixtureDirectory, "fixture.ts")
    );
    await fs.rm(path.join(failingFixtureDirectory, "snapshots"), {
      recursive: true,
    });
  }

  // Re-run Jest to update snapshots a second time
  await exec(`jest /fixtures`, {
    cwd: "packages/ast-spec",
  });
}

main()
  .then(() => console.log("Done."))
  .catch((error) => {
    throw error;
  });
```

</details>


bradzacher:
Let's do this:

1) move this out of the `src` folder to a new folder called like `legacy-fixtures`
2) remove the todo comments and add info to the readme about how to add fixtures and what ***not*** to do.

Eg
> The `legacy-fixtures` folder contains all of the old fixtures from our previous testing framework. No new fixtures should be added to this folder, and instead we should be attempting to migrate and remove cases wherever possible.
>
> insert steps #6065?

After that I think we're in a good spot to land this?

bradzacher:
We'll also want to make sure these fixtures are automatically re-tested using the test tooling, so you'll need to update [the test to include them](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/tests/fixtures.test.ts)

JoshuaKGoldberg:
Hmm, would you be upset if I moved `legacy-fixtures` to within `src/`? That way I don't have to keep these logic changes for two test directories...

bradzacher:
ah yeah I guess the test isn't built to be relative to anything other than the `src` folder.
feel free to drop it in there then.

bradzacher (approved):
yas

PR #5915, opened 2022-11-02
feat: bump minimum supported TS version to 4.2.4

bradzacher:
BREAKING CHANGE:
Bumps the minimum supported range and removes handling for old versions

## PR Checklist

- [x] Addresses an existing open issue: fixes #5914, fixes #2521, closes #5367
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

DT hasn't bumped to 4.2 yet, but it's only just ticked to november, so I'd expect them to do it soon - likely before we do the major release.

This PR:
- updates docs to clarify our version support windows
- bumps the ranges around the place
- removes no-longer-necessary checks for now outdated versions

PR #6458, opened 2023-02-13
feat: add package.json exports for public packages

### LINKED ISSUE #6050: Use package.json exports to "hide" the dist folder for packages and control our exported surface-area
### Discussed in https://github.com/typescript-eslint/typescript-eslint/discussions/6015

<div type='discussions-op-text'>

<sup>Originally posted by **bradzacher** September  1, 2022</sup>
Currently we don't use package.json exports, so anyone can import any file in any package by directly referencing a path within the `dist` folder - i.e.
```ts
import * as TSESLint from '@typescript-eslint/utils/dist/ts-eslint';
```

This is obviously a sub-par experience for consumers due to the deep path, and is a sub-par experience for us because any file technically forms part of our "external API".

In ESLint v8 they began using exports to lock down their package, and I think we should follow their lead to some degree.

I propose that we follow one of the following options:

### 1) export all of dist, without the dist

```json
{
  "exports": {
    "./*": "./dist/*"
  },
  "typesVersions": {
    "*": {
      "*": ["dist/*"]
    }
  }
}
```

This is the least restrictive thing we could do - it would simply mean consumers needn't include `dist/` in their import path - which is nice as it hides an implementation detail.

### 2) export all, but ban specific things

```json
{
  "exports": {
    "./*": "./dist/*"
    "./some-internal/file": null
  },
  "typesVersions": {
    "*": {
      "*": ["dist/*"]
    }
  }
}
```

This is slightly more restrictive, but still very permissive.
This would allow us to build APIs that are intended to be internal only and hide them, whilst still generally allowing all imports without the `dist/`.

### 3) export specific, allowed parts of our API

```jsonc
{
  "exports": {
    "./": "./dist/index",
    "./ts-eslint": "./dist/ts-eslint/index",
    "./ast-utils": "./dist/ast-utils/index",
    // ...
  },
  "typesVersions": {
    "*": {
      "./": "./dist/index",
      "./ts-eslint": "./dist/ts-eslint/index.d.ts",
      "./ast-utils": "./dist/ast-utils/index.d.ts",
      // ...
    }
  }
}
```

This is the most restrictive thing we could do as we would ban deep imports except those that we choose. It would also allow us to remove the `dist/` from imports.</div>

### Conclusion

We'd like to go with option 3. Marking as a breaking change and accepting PRs! 🚀

bradzacher:
BREAKING CHANGE:
Restricts API surface by strictly defining package.json exports for each package

## PR Checklist

- [x] Addresses an existing open issue: fixes #6050

## Overview

Configures option 3 from #6050 - where we completely lock down our API so people cannot access our `dist` folder any more.

Currently blocked on https://github.com/JoshuaKGoldberg/ts-api-utils/issues/54


bradzacher:
Hmm.. I can't reproduce the typecheck failure locally... There's no reason it shouldn't work considering it's not the ESM version of the package.

Maybe there's an issue with the CI getting the wrong version? Idk why it would be different to local considering it's covered by the yarn.lock

packages/website-eslint/types/index.d.ts
```diff
@@ -1,13 +0,0 @@
-import type { TSESLint } from '@typescript-eslint/utils';
-
-import { analyze } from '@typescript-eslint/scope-manager/dist/analyze';
-import { astConverter } from '@typescript-eslint/typescript-estree/dist/ast-converter';
-import { getScriptKind } from '@typescript-eslint/typescript-estree/dist/create-program/getScriptKind';
-
-export interface LintUtils {
```
armano2, line 7: are you sure that this is correct?

bradzacher: Yes - I moved this to the `website` package.
It wasn't ever actually used in this package, nor did it represent anything that was returned/exported from `website-eslint`.

This also means it's properly covered by typechecks.

packages/eslint-plugin/typings/eslint-rules.d.ts
```diff
@@ -195,7 +195,7 @@ declare module 'eslint/lib/rules/key-spacing' {
 
 declare module 'eslint/lib/rules/keyword-spacing' {
   import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
-  import type { RuleFunction } from '@typescript-eslint/utils/dist/ts-eslint';
+  import type { RuleFunction } from '@typescript-eslint/utils/ts-eslint';
```
JoshuaKGoldberg, line 198: Feels like the `RuleFunction` type is probably a little useful broadly. How about we export it normally?

```suggestion
  import type { RuleFunction } from '@typescript-eslint/utils';
```

packages/ast-spec/tests/util/parsers/typescript-estree-import.ts
```diff
@@ -14,5 +14,4 @@
  */
 
 // We need to ignore this lint error regarding it being missing from the package.json, see above.
```
JoshuaKGoldberg, line 16:
```suggestion
```

packages/eslint-plugin-tslint/package.json
```diff
@@ -49,5 +57,9 @@
   "devDependencies": {
     "@types/lodash": "*",
     "@typescript-eslint/parser": "5.51.0"
+  },
+  "funding": {
+    "type": "opencollective",
+    "url": "https://opencollective.com/typescript-eslint"
```
JoshuaKGoldberg, line 63: 🤑

packages/eslint-plugin-tslint/tsconfig.build.json
```diff
@@ -1,6 +1,9 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    // specifically disable declarations for the plugin
```
JoshuaKGoldberg, line 4: The code clearly does that - a more interesting comment IMO would be _why_. Which is a question I have - do we not want folks programmatically accessing anything from `eslint-plugin-tslint`?

packages/eslint-plugin-tslint/tsconfig.json
```diff
@@ -5,6 +5,10 @@
     "rootDir": "."
   },
   "include": ["src", "tests"],
-  "exclude": ["tests/test-project", "tests/test-tslint-rules-directory"],
+  "exclude": [
+    "tests/fixture-project",
```
JoshuaKGoldberg, line 9: Why was this not needed before?

packages/experimental-utils/package.json
```diff
@@ -1,54 +0,0 @@
-{
```
JoshuaKGoldberg, line 1: Might make sense to split the removal of `experimental-utils` out into its own PR, and base this PR on top of that one?

packages/scope-manager/package.json
```diff
@@ -2,20 +2,36 @@
...
     "README.md",
     "LICENSE"
   ],
+  "type": "commonjs",
+  "typesVersions": {
+    "<3.8": {
+      ".": [
+        "_ts3.4/dist/index.d.ts"
+      ]
+    }
```
JoshuaKGoldberg, line 17: TypeScript 3.8 came out 3 years ago. I'd vote to just get rid of <3.8 compat while we're in the area & doing breaking changes.

patches/ts-api-utils+0.0.21.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/ts-api-utils/package.json b/node_modules/ts-api-utils/package.json
+index 41ee37c..e4279d1 100644
+--- a/node_modules/ts-api-utils/package.json
++++ b/node_modules/ts-api-utils/package.json
+@@ -8,7 +8,7 @@
+ 	},
+ 	"license": "MIT",
+ 	"author": "Josh Goldberg <npm@joshuakgoldberg.com>",
+-	"type": "module",
++	"type": "commonjs",
```
JoshuaKGoldberg, line 10: Should be fixed in https://github.com/JoshuaKGoldberg/ts-api-utils/pull/37 -> `ts-api-utils@0.0.22`.

packages/website/src/pages/index.tsx
```diff
@@ -161,7 +161,7 @@ function Home(): JSX.Element {
             key={idx}
             className={clsx(
               styles.features,
-              idx % 2 == 1 ? styles.lightBackground : '',
+              idx % 2 === 1 ? styles.lightBackground : '',
```
JoshuaKGoldberg, line 164: Would you be open to splitting out the ESLint changes into their own PR, to make it more clear what's changing for them? This PR is pretty big.

JoshuaKGoldberg (approved):
Glorious, I love it. 👏 

Left some suggestions inline but they're all nice-to-haves IMO.

packages/website/src/pages/index.tsx
bradzacher: Yeah I was just hacking on things to make it work then I noticed these files weren't being linted and fixed that.

I'll fix it in master - there are more files that need to be fixed anyways!

packages/eslint-plugin/typings/eslint-rules.d.ts
bradzacher: I don't love exporting it from multiple places just because it then gets confusing across a codebase.
"Did you mean to auto-import `RuleFunction` from `@typescript-eslint/utils` or `@typescript-eslint/utils/ts-eslint`?" Then you have half the codebase using one way and another half using the other way depending on who clicked what at the time.

Even worse you have 4 ways of consuming it:
```ts
import { TSESLint, RuleFunction } from '@typescript-eslint/utils';
import * as TSESLint2 from '@typescript-eslint/utils/ts-eslint';
import { RuleFunction as RuleFunction2 } from '@typescript-eslint/utils/ts-eslint';

// TSESLint.RuleFunction === TSESLint2.RuleFunction === RuleFunction === RuleFunction2
```

I think having the root namespaced export for a nice "single import declaration" case is good for those that want it (which is how we have already done things in the past), and having a single way to import names directly is good as well (which other codebases did a lot - via the `/dist/` path 😢).

packages/eslint-plugin-tslint/tsconfig.build.json
bradzacher: The why specifically is commented here:
https://github.com/typescript-eslint/typescript-eslint/pull/6458/files#diff-e5831196ebbae7f2aa0010959e8dad69be30d30a0b26e691d2cf0bc332230f14R1

I'll move the comment to a more useful spot.

> We purposely don't generate types for our plugin because:
> 1) there's no real reason that anyone should do a typed import of our rules,
> 2) it would require us to change our code so there aren't as many inferred types

You can try it yourself and see the errors - for TS to generate declarations we need to annotate all of our `export default createRule(...)` like `export default createRule(...) satisfies RuleModule<TMessageIds, TOptions>`. In a nutshell - TS knows that to generate the `.d.ts` file it'll need that `RuleModule` type - but TS doesn't like to insert an import on your behalf, so it errors asking you to annotate it so that the `.d.ts` generation remains just an elision process.

packages/eslint-plugin-tslint/tsconfig.json
bradzacher: I think it is because there used to be a TSConfig in the tests folder which technically hid the files from this config?
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin-tslint/tests/tsconfig.json

I deleted this tsconfig because it's not how we do things, so the files got included, so now they need to be excluded.

packages/scope-manager/package.json
bradzacher: #5915 - our minimum is now 4.2 - so yeah we can bump this up.
I'll do that separately

packages/website-eslint/typings/eslint.d.ts
```diff
@@ -0,0 +1,7 @@
+// if we don't declare this, then TS in VSCode will use its "helpfully fetched"
+// types and report errors in the IDE
```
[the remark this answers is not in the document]
bradzacher: VSCode has a helpful feature where it'll automatically fetch types for you for your JS node module imports.
So it means that in the IDE you'll open the file and VSCode will tell TS it can look in its cache for module types for the `@types/eslint` package. It finds it, uses it, then will show errors because our `RuleModule` doesn't match the `@types/eslint` `RuleModule`.

But this behaviour only happens in the IDE, not the CLI - because the CLI can't use VSCode's cache, ofc, so it just uses `any` for the import. So adding this type tells TS that it cannot use VSCode's cache ever - stubbing it out permanently.

I hate it fo sho. I wanted to use the types from `@typescript-eslint/utils/ts-eslint`, but this package is supposed to be as light-weight as possible.

In a nutshell this package is how we bundle up our project for use on the web. So it's a bit weird and hacky - mocks out some modules and such. I'll raise a PR against main to clean some stuff up.

packages/website-eslint/src/mock/typescript.js
```diff
@@ -1 +1,2 @@
+// eslint-disable-next-line no-undef
```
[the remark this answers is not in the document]
bradzacher: yeah this was me being lazy lol

packages/website-eslint/src/mock/semver.js
```diff
@@ -1,7 +1,8 @@
-import satisfies from 'semver/functions/satisfies';
 import major from 'semver/functions/major';
+import satisfies from 'semver/functions/satisfies';
 
 // just in case someone adds a import * as semver usage
 export { satisfies, major };
 
+// eslint-disable-next-line import/no-default-export
```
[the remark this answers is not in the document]
bradzacher: I was trying that but for whatever reason it wasn't working...
I'll try again when I reapply it against master

patches/ts-api-utils+0.0.21.patch
bradzacher: I guess the issue now is that TS won't let us import this file because it sees `type: 'module'` and our project is `type: 'commonjs'`

packages/typescript-estree/package.json
```diff
@@ -2,14 +2,41 @@
...
+    "<3.8": {
+      ".": [
+        "_ts3.4/dist/index.d.ts"
+      ]
+    }
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "require": "./dist/index.js"
```
fisker, line 22: Have you consider ship this package in dual mode?

patches/ts-api-utils+0.0.21.patch
bradzacher: Yup - we can't import the package without this patch.
TS looks at `type: module` and treats it as an exclusively esm package, erroring:
```
src/convert-comments.ts:1:24 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("ts-api-utils")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts' or create a local package.json file with `{ "type": "module" }`.
       
1 import * as tools from 'ts-api-utils';
                          ~~~~~~~~~~~~~~
```

So we'll need to keep this patch (in perpetuity?)
Maybe this needs to be changed in the package so it doesn't list `type`?

JoshuaKGoldberg: Hmm, I suppose we could take an issue. But I do wonder what the actual blockers to us being ESM are.

bradzacher: > I do wonder what the actual blockers to typescript-eslint going ESM are.

In order for CJS to load ESM, it must use an `import()` expression.

ESLlint is CJS and does not use `import()` to load plugins - it uses `require` (via [`createRequire`](https://nodejs.org/api/module.html#modulecreaterequirefilename))
https://github.com/eslint/eslintrc/blob/main/lib/config-array-factory.js#L1086
https://github.com/eslint/eslintrc/blob/main/lib/config-array-factory.js#L984

So we need to export a CJS API from our packages. Now we could get around that and use ESM with a CJS facade in front of an ESM package and leverage `import()` expressions to load the ESM "backend", however our plugin and parser need to have fully synchronous APIs to adhere to the ESLint requirements.

Because the plugin needs to be CJS, all our lint utilities (`utils`, `type-utils`, `scope-manager`, etc) need to be CJS.
Because the parser needs to be CJS, all our parser utilities (`visitor-keys`, `typescript-estree`, etc) need to be CJS.

So for now we're firmly stuck as a project shipping CJS :(

PR #6468, opened 2023-02-15
feat: remove `experimental-utils`

bradzacher:
BREAKING CHANGE:
Removes `experimental-utils` - we will no longer update this package and it will be forever frozen at v5.x

## Overview

We deprecated this package quite a while ago when we (finally) switched to the non-experimental name.
As of v6 we'll no longer update this package.


JoshuaKGoldberg (approved):
🔪

PR #6470, opened 2023-02-15
feat(eslint-plugin): add config that disables type-aware linting

### LINKED ISSUE #6215: Configs: add a config that turns off all type-aware linting

### Description

Our current guidance on how to setup an ESLint config is to turn everything on in the root of the config and be done with it. This advice includes turning on type-aware linting on all files.

However for a lot of codebases this can create a crappy experience because there are a lot of files NOT included in a tsconfig that ESLint will run over. This in turn causes the dreaded "The file must be included in at least one of the projects provided" error.

There's not really a good way around this right now, sadly. [Our advice](https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file) is pretty long winded and not the easiest thing to put into practice, especially if you don't actually want to lint the file with type-info. Possible workarounds include:
1) `.eslintignore` the files.
2) create a `tsconfig.eslint.json` that includes the files.
3) manually turn off type-aware rules AND `parserOptions.project` on the files.
4) rework your config so that you only enable type-aware linting on specific files using `overrides` config.

All of those options suck for a user because:
1) these files are often files that you want to do at least basic lint checks on
2) these files are often `.js` files with poor-to-no types or are config files where most type-aware rules aren't going to catch anything useful (if anything at all)
3) it's difficult to know exactly what type-aware rules you have turned on due to config composition
4) it's hard to rework your entire config without accidentally missing files that should be type-aware linted.

I realised that we're in a good position to help people automate this. We know what rules need type-info - so why not generate a config for it? That way the solution to the problem can simply be "use overrides to disable type-aware linting by extending this config", i.e.
```js
module.exports = {
  // ...
  overrides: [
    // disable type-aware linting for some files
    {
      files: ['.eslintrc.js'/*, ... */],
      extends: ['plugin:@typescript-eslint/disable-type-aware-linting'],
    },
  ],
}
```

This would be a substantially better UX and is a pretty clean solution!

### Impacted Configurations

_No response_

### Additional Info

_No response_

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6215

## Overview

Builds on top of the new config tooling to generate a config that disables type-aware linting.

### DIFF
--- a/docs/linting/Configurations.mdx
+++ b/docs/linting/Configurations.mdx
@@ -179,7 +179,7 @@ See [`configs/base.ts`](https://github.com/typescript-eslint/typescript-eslint/b
 ### `disable-type-checked`
 
 A utility ruleset that will disable type-aware linting and all type-aware rules available in our project.
-This config makes it easy to have your base config concerned with type-aware linting, and then conditionally use [overrides](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-based-on-glob-patterns) to disable type-aware linting on specific subsets of your codebase.
+This config is useful if you'd like to have your base config concerned with type-aware linting, and then conditionally use [overrides](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-based-on-glob-patterns) to disable type-aware linting on specific subsets of your codebase.
 
 See [`configs/disable-type-checked.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/disable-type-checked.ts) for the exact contents of this config.
 
--- a/docs/linting/Configurations.mdx
+++ b/docs/linting/Configurations.mdx
@@ -15,14 +15,14 @@ title: Configurations
 
 If your project does not enable [typed linting](./Typed_Linting.mdx), we suggest enabling the [`recommended`](#recommended) and [`stylistic`](#stylistic) configurations to start:
 
-```json
-{
-  "extends": [
-    "eslint:recommended",
-    "plugin:@typescript-eslint/recommended",
-    "plugin:@typescript-eslint/stylistic"
-  ]
-}
+```js title=".eslintrc.js"
+module.exports = {
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:@typescript-eslint/stylistic',
+  ],
+};
 ```
 
 > If a majority of developers working on your project are comfortable with TypeScript and typescript-eslint, consider replacing `recommended` with `strict`.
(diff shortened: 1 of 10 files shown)


docs/linting/Configurations.mdx
```diff
@@ -171,26 +170,51 @@ Many rules conflict with each other and/or are intended to be configured per-pro
...
+We don't recommend using this directly; instead, extend from an earlier recommended rule.
 
 This config is automatically included if you use any of the recommended configurations.
 
 See [`configs/base.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/base.ts) for the exact contents of this config.
 
+### `disable-type-checked`
+
+A utility ruleset that will disable type-aware linting and all type-aware rules available in our project.
+This config makes it easy to have your base config concerned with type-aware linting, and then conditionally use [overrides](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-based-on-glob-patterns) to disable type-aware linting on specific subsets of your codebase.
```
JoshuaKGoldberg, line 182: Nit: _"easy"_ as a term can be offputting to developers who aren't finding the concepts easy. Which is a lot of folks when it comes to linting configs & the like IME.

```suggestion
This config is useful if you'd like have your base config concerned with type-aware linting, and then conditionally use [overrides](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-based-on-glob-patterns) to disable type-aware linting on specific subsets of your codebase.
```

packages/types/src/parser-options.ts
```diff
@@ -51,8 +51,8 @@ interface ParserOptions {
   extraFileExtensions?: string[];
   filePath?: string;
   loc?: boolean;
-  program?: Program;
-  project?: string | string[] | true;
+  program?: Program | null;
+  project?: string | string[] | true | null;
```
JoshuaKGoldberg, line 55: Is this needed for this PR?

It's also weird to me IMO to have `null` be an option only for some of these.

JoshuaKGoldberg (changes requested):
Looks like some unrelated `null` types got added in? Otherwise looks great!

packages/types/src/parser-options.ts
bradzacher: I added null because it's an explicit value to set these to off that won't get weirdly merged by ESLint's cascading config. It also means the configs can be json-serialized without losing any data (`undefined` tells a json stringifier to omit the property).

It just means we can be sure these configs are robust.

I also considered using `[]`, but IIRC ESLint's cascading config logic will merge arrays together. Which means that `[]` is the same as not specifying a value at all.

JoshuaKGoldberg (approved):
Got it - this is great then. Yay! 🙌

PR #6473, opened 2023-02-15
chore: cleanup repo lint + typechecking

bradzacher:
I noticed a lot of missing coverage with our linting and typechecking setup.
- The top-level test and tools directories weren't being linted/typechecked
- `.js` files weren't being linted or formatted around the place
- Some of our `.d.ts` files weren't being linted or formatted

In some cases files were just part of ignore files when they shouldn't've been. 
In other cases they weren't properly covered by the nx command config.

This PR fixes all of that with a small reorganisation of the codebase and addition of some config files. It also upgrades prettier.


.prettierignore
```diff
@@ -1,22 +1,16 @@
...
 **/.vs
-packages/eslint-plugin-tslint/tests/test-tslint-rules-directory/alwaysFailRule.js
-packages/eslint-plugin/src/configs/*.json
 CONTRIBUTORS.md
 packages/ast-spec/src/*/*/fixtures/_error_/*/fixture.ts
 
-# prettier doesn't yet support auto-accessors
-packages/ast-spec/src/element/AccessorProperty/fixtures
+# prettier errors on this case because it's semantically valid
+packages/ast-spec/src/element/AccessorProperty/fixtures/modifier-abstract-with-value/fixture.ts
```
JoshuaKGoldberg, line 10: Maybe we should turn off Prettier for all fixtures?

packages/website/src/vendor/typescript-vfs.d.ts
```diff
@@ -1,21 +1,26 @@
-declare type System = import('typescript').System;
-declare type CompilerOptions = import('typescript').CompilerOptions;
-declare type CustomTransformers = import('typescript').CustomTransformers;
-declare type LanguageServiceHost = import('typescript').LanguageServiceHost;
-declare type CompilerHost = import('typescript').CompilerHost;
-declare type SourceFile = import('typescript').SourceFile;
-declare type TS = typeof import('typescript');
+/* eslint-disable import/no-duplicates -- bug -> rule suggests merging to `import * as ns, {...}` which is invalid TS syntax */
```
JoshuaKGoldberg, line 1: Is this bug filed anywhere on GitHub?

package.json
```diff
@@ -30,19 +30,19 @@
...
+    "generate-sponsors": "tsx ./packages/repo-tools/generate-sponsors.ts",
+    "generate-website-dts": "tsx ./packages/repo-tools/generate-website-dts.ts",
     "generate-lib": "nx generate-lib @typescript-eslint/scope-manager",
-    "lint-fix": "eslint . --fix",
+    "lint-fix": "yarn lint --fix",
     "lint-markdown-fix": "yarn lint-markdown --fix",
     "lint-markdown": "markdownlint \"**/*.md\" --config=.markdownlint.json --ignore-path=.markdownlintignore",
     "lint": "nx run-many --target=lint --parallel",
-    "postinstall": "yarn tsx ./tools/postinstall.ts",
+    "postinstall": "tsx ./packages/repo-tools/postinstall.ts",
```
JoshuaKGoldberg, line 41:
```suggestion
    "postinstall": "tsx ./packages/repo-tools/src/postinstall.ts",
```

packages/integration-tests/jest.config.js
```diff
@@ -5,10 +5,13 @@ require('ts-node').register({
   transpileOnly: true,
   files: ['./pack-packages.ts'],
 });
-const { tseslintPackages } = require('./pack-packages');
+const { tseslintPackages } = require('./tests/pack-packages');
```
JoshuaKGoldberg, line 8:
```suggestion
const { tseslintPackages } = require('./tools/pack-packages');
```

JoshuaKGoldberg: And then tests are failing:

```plaintext
 FAIL  tests/markdown.test.ts
  ● markdown › should lint successfully
    thrown: Array [
      [Error: ENOENT: no such file or directory, lstat '/Users/josh/repos/typescript-eslint/packages/integration-tests/tools/fixtures/markdown'],
```

Some issues with relative directory naming

.eslintrc.js
```diff
@@ -14,6 +16,9 @@ module.exports = {
     es6: true,
     node: true,
   },
+  globals: {
+    globalThis: true,
```
JoshuaKGoldberg, line 20:
```suggestion
    es2020: true,
    node: true,
```

JoshuaKGoldberg (changes requested):
Lots of changes here that look good - but the tests are borked?

packages/website/src/vendor/sandbox.d.ts
```diff
@@ -1,10 +1,10 @@
-import { TypeScriptWorker } from './tsWorker';
```
armano2, line 1: this file is downloaded from ts, we should not modify it.

.prettierignore
bradzacher: I'd rather we keep everything formatted and special case the things that are broken, personally.

packages/website/src/vendor/sandbox.d.ts
bradzacher: The changes I made:
- switch `any` for `unknown`
- replace inline type import with type-only import declarations
- deduplicate code
- apply formatter

Functionally the code is the ecactly same - it just adheres to our code style now.

armano2: you should update tools/generate-website-dts.ts, to apply those changes than :)

bradzacher: I didn't realise they were generated!

JoshuaKGoldberg (approved):
🔥

PR #6485, opened 2023-02-17
chore: collapse generated snapshots in PR view by default

Josh-Cena:
## PR Checklist

## Overview

Per discussion with @JoshuaKGoldberg. This makes snapshots collapsed just like `yarn.lock` changes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,4 @@
 
 # force github to treat out custom jest snapshot extension as normal jest snapshots
 *.shot linguist-language=Jest-Snapshot
+*.shot linguist-generated


bradzacher:
Won't this remove the syntax highlighting in GH because you're losing the language attribute?

I don't know if they should be seen as generated - aside from codebase reactors, we'll really want to be reviewing all of our snapshots as code because they represent the AST output.

I do have an idea for adding snapshot testing to the rule tester as well, so that would be more snapshots to be reviewed.

Josh-Cena:
> Won't this remove the syntax highlighting in GH because you're losing the language attribute?

No it won't. `linguist-language` and `linguist-generated` are two orthogonal attributes. You can refer to https://github.com/github/linguist/blob/master/docs/overrides.md

> I don't know if they should be seen as generated

You can still view the diff, just that they are collapsed by default. This prevents diffs like https://github.com/typescript-eslint/typescript-eslint/pull/6274 from being unviewable.

PR #6487, opened 2023-02-18
feat(eslint-plugin): [prefer-nullish-coalescing] add `ignorePrimitives` option

### LINKED ISSUE #4906: [prefer-nullish-coalescing] request for preferring "||" when string value

**Repro**

In cases with strings, we may want to prefer that empty string `""` is equivalent to nullish, and show a fallback in that case. There currently is no option to support customizing string behavior.
* https://typescript-eslint.io/rules/prefer-nullish-coalescing#options

Example:
```ts
let emailAddress: string | null = getEmailAddress();
const link = "<a href='/'>" + (emailAddress || "---") + "</a>";
```

At a minimum, it'd be good to have a `ignoreStrings` option.

### LINKED ISSUE #7055: Bug: [prefer-nullish-coalescing] False positive to `nullable boolean`

### Playground Link

https://typescript-eslint.io/play/#ts=5.0.4&sourceType=module&code=CYUwxgNghgTiAEYD2A7AzgF0VNICMAMklMAJYoDmAXPAEZJIQhQrwA+8AriqAGbkhg7eCk4QIAbgCwAKFCRYCZOixgcIAExES5anQZMW0mbOWZ4cNGKwBebLkLEyldhzW4tT3cYD0P+AH2mgF2vFAQuIFRrkEasn5R7vgh8BgwnAjRgWxu6nEyCYFJGincfAJC2bkesvH+gQAqABakaPCoCK3wEJxgANakEACe8ADuSDADLuSpLW1JdYnqeClpGVmBAPybsbUF9QEAgvQAbggdbYrwAPIA0gA0XGi68AAGOa97hQEAQpxYGCaCBQIAAHlgOvAumpOM9pmgrCA2qNSIC3ttPvslg4UmEIpksttdljGnN4GgmkgxEJaAgQCdwpwoBgEBgkG9iq8xkDWLCXii0e82K9HrR-lCMAByS4QUZQIZtV5JPBcgVNdGbTHfeAACRAKDASiaLAoCCFXLZGq5tDgUD6bUBCAgSAopDAjzQ7MdFjECApVIgQgActcGhYQAAHCYYIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6KL2jjokcGAC+ILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
declare const case1Loading: boolean | undefined | null;
declare const case2Loading: boolean;

const result = case1Loading || case2Loading;
//    case2  = false        || case2
//    case1  = true         || case2
//    case2  = undefined    || case2

//    This one is luckily working in this case
//    case1  = true         ?? case2

//    Above ones are OK, using `||`

//    But the next one is causing issues with `??`
//    case1  = false        ?? case2
//    This should be evaluate to `case2` when using with `||`, but it's always `case1` with `??`
//    Hence change `||` to `??` breaks the logic, so the rule should NOT report
```

### ESLint Config

```javascript
{
  "rules": {
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

The rule should not report against `nullable boolean`, since it will likely break the logic

### Actual Result

Prefer using nullish coalescing operator (`??`) instead of a logical or (`||`), as it is a safer operator. 11:29 - 11:31
> Fix to nullish coalescing operator (`??`).

### Additional Info

Another solution is to add an `ignoreNullableBoolean` option, since it will likely breaks the logicbreak

omril1:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4906, fixes #7055

## Overview

Add `ignorePrimitives` option that can be set individually for `string`, `boolean`, `number` and `bigint`


packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md
```diff
@@ -129,6 +129,14 @@ a ?? (b && c) ?? d;
 a ?? (b && c && d);
 ```
 
+### `ignorePrimitives`
```
JoshuaKGoldberg, line 132: There's a [`lint:markdown` complaint](https://github.com/typescript-eslint/typescript-eslint/actions/runs/4210554347/jobs/7308326641) blocking the PR build.

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md
```diff
@@ -129,6 +129,14 @@ a ?? (b && c) ?? d;
...
+If you would like to ignore certain primitive types that can be falsy then you may pass an object containing a boolean value for each primitive:
+
+- Option entries in the object:
```
JoshuaKGoldberg, line 135: Nit: this line is implied, I think.

```suggestion
```

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -1,6 +1,7 @@
 import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
 import { AST_NODE_TYPES, AST_TOKEN_TYPES } from '@typescript-eslint/utils';
 import * as tsutils from 'tsutils';
+import { isTypeFlagSet, isUnionOrIntersectionType } from 'tsutils';
```
JoshuaKGoldberg, line 4: Nit: we already `import * as tsutils` in this file, any particular reason to add a second import form?

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
```diff
@@ -20,6 +20,7 @@ const ruleTester = new RuleTester({
 
 const types = ['string', 'number', 'boolean', 'object'];
 const nullishTypes = ['null', 'undefined', 'null | undefined'];
+const ignorablePrimitiveTypes = ['string', 'number', 'boolean'] as const;
```
JoshuaKGoldberg, line 23: [Nitpick] doesn't look like this is needed?

```suggestion
const ignorablePrimitiveTypes = ['string', 'number', 'boolean'];
```

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md
```diff
@@ -129,6 +129,14 @@ a ?? (b && c) ?? d;
...
+  - `string: true`, ignores `null` or `undefined` unions with `string` (default: false).
+  - `number: true`, ignores `null` or `undefined` unions with `number` (default: false).
+  - `boolean: true`, ignores `null` or `undefined` unions with `boolean` (default: false).
+
 **_NOTE:_** Errors for this specific case will be presented as suggestions (see below), instead of fixes. This is because it is not always safe to automatically convert `||` to `??` within a mixed logical expression, as we cannot tell the intended precedence of the operator. Note that by design, `??` requires parentheses when used with `&&` or `||` in the same expression.
```
JoshuaKGoldberg, line 140: The `**_NOTE:_**` is associated with the previous section. Did you mean to put the new `###` section _after_ this note? (that's where it should be)

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -279,6 +299,27 @@ export default util.createRule<Options, MessageIds>({
           return;
         }
 
+        if (isUnionOrIntersectionType(type)) {
```
JoshuaKGoldberg, line 302: [Request] Tests still pass without this check. Is there any reason to include it? At this point I'd expect the type to be a union or intersection type, as `util.isNullableType(type, { allowUndefined: true })` was truthy.

```suggestion
```

[Tip] The type system may be upset after this about the type not being a `ts.UnionOrIntersectionType`. That's ok. I think an `as` assertion would be fine, or a type predicate just for this file.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -11,6 +12,11 @@ export type Options = [
     ignoreTernaryTests?: boolean;
     ignoreMixedLogicalExpressions?: boolean;
     allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing?: boolean;
+    ignorePrimitives?: {
+      string?: boolean;
+      boolean?: boolean;
+      number?: boolean;
```
JoshuaKGoldberg, line 18: Nit: could you alphabetize these in all the places they're declared? It's easier to read through docs when the ordering is predictable.

```suggestion
      boolean?: boolean;
      number?: boolean;
      string?: boolean;
```

This also applies to where to place the `ignorePrimitives` option. Feels like it should be after `ignoreMixedLogicalExpression`. _(I realize the options already aren't alphabetical, heh - that's an issue that can be a separate followup)_

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -279,6 +299,27 @@ export default util.createRule<Options, MessageIds>({
...
+          }
+          if (
+            ignorePrimitives?.boolean &&
+            type.types.some(t => isTypeFlagSet(t, ts.TypeFlags.BooleanLiteral))
+          ) {
+            return;
+          }
+          if (
+            ignorePrimitives?.number &&
+            type.types.some(t => isTypeFlagSet(t, ts.TypeFlags.Number))
```
JoshuaKGoldberg, line 317: [Request] There are three `.some `loops here. You can reduce the amount of loops and lines of code a teeny bit by using `|` on all the ignorable type flags.

```ts
const ignorableFlags [
  ignorePrimitives.boolean && ts.TypeFlags.BooleanLiteral,
  ignorePrimitives.number && ts.TypeFlags.Number,
  ignorePrimitives.string && ts.TypeFlags.String,
]
  .filter(isNotUndefined)
  .reduce((previous, flag) => previous | flag, 0);
```

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
```diff
@@ -751,5 +759,27 @@ declare const c: ${type};
...
+              .map(t => [t, true]),
+          ),
+        },
+      ],
+      errors: [
+        {
+          messageId: 'preferNullishOverOr',
+        },
+      ],
+    })),
```
JoshuaKGoldberg, line 783: Some missing test cases to fill in:

* Literal types (`true`, `1`, etc.)
* Template literals (<code>\`hello${string}\`</code>)
* Bigints (`1n`)
* Combinations of the types (`boolean | string | ...`), etc.

JoshuaKGoldberg (changes requested):
A good start, thanks for sending this in! ❤️ 

Requesting changes on more test cases, docs cleanups, and some code shenanigans.

JoshuaKGoldberg:
👋 Hey @omril1! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

omril1:
@JoshuaKGoldberg I'm still on it, I have some more work that I left in the middle

xsjcTony:
Any updates here? 😄

omril1:
> Any updates here? 😄

Mmm, yeah I left it out for a while, got a bit of a burnout at work 🥲.
I'll get back to it maybe in the weekend.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
omril1: It's there so typescript will know it's an intersection (`type is ts.UnionOrIntersectionType` type guard)

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md
omril1: I thought the note was in general for the whole rule so it was placed at the end, moved it back up

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -279,6 +298,33 @@ export default util.createRule<Options, MessageIds>({
...
+            return;
+          }
+          if (
+            ignorePrimitives!.boolean &&
+            type.types.some(t =>
+              tsutils.isTypeFlagSet(t, ts.TypeFlags.BooleanLiteral),
+            )
+          ) {
+            return;
+          }
```
JoshuaKGoldberg, line 315: [Cleanup] These are handled by the `type.types.some` later on. Removing them doesn't cause any test failures. Unneeded code?

```suggestion
```

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
JoshuaKGoldberg: Yes, it tells TypeScript that it's an intersection - but _we_ know it's always an intersection. _We_ know the type is a union or intersection type because `util.isNullableType` returned true a few lines earlier. This _runtime_ code is only providing utility for the _type system_. That's a form of dead/unnecessary code.

I'm suggesting writing something like `(type as ts.UnionOrIntersectionType).types`.

JoshuaKGoldberg (dismissed):
Looking great! Just a couple more code cleanups then we're good to merge I think. 🚀

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
omril1: Fixed both, thanks!

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md
```diff
@@ -131,6 +131,14 @@ a ?? (b && c && d);
 
 **_NOTE:_** Errors for this specific case will be presented as suggestions (see below), instead of fixes. This is because it is not always safe to automatically convert `||` to `??` within a mixed logical expression, as we cannot tell the intended precedence of the operator. Note that by design, `??` requires parentheses when used with `&&` or `||` in the same expression.
 
+### `ignorePrimitives`
+
+If you would like to ignore certain primitive types that can be falsy then you may pass an object containing a boolean value for each primitive:
+
+- `string: true`, ignores `null` or `undefined` unions with `string` (default: false).
+- `number: true`, ignores `null` or `undefined` unions with `number` (default: false).
```
omril1, line 139: @JoshuaKGoldberg Should I add an option for bigint? I realized that `!0n === true` only now

JoshuaKGoldberg: If you have the time, sure! I don't think very many people are working with bigints, but it's a nice nice-to-have.

omril1: Done

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md
```diff
@@ -131,6 +131,15 @@ a ?? (b && c && d);
...
+- `bigint: true`, ignores `null` or `undefined` unions with `bigint` (default: false).
+- `boolean: true`, ignores `null` or `undefined` unions with `boolean` (default: false).
```
omril1, line 141: @JoshuaKGoldberg Do markdown options need to be sorted alphabetically like the code, by how common they are or some other rule?

JoshuaKGoldberg: I personally prefer alphabetical, but there's no hard rule we enforce.

omril1:
The website tests keep failing in CI and I can't re-run them.
@JoshuaKGoldberg Is there anything I can do to move this forward?

JoshuaKGoldberg:
Ah don't worry about the website CI tests, that's a separate thing.

JoshuaKGoldberg:
FWIW you don't have to keep merging `main` if you don't want. We're a little focused on v6 right now so the issue & PR review queue is taking longer than we'd normally want.

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md
```diff
@@ -131,6 +131,15 @@ a ?? (b && c && d);
 
 **_NOTE:_** Errors for this specific case will be presented as suggestions (see below), instead of fixes. This is because it is not always safe to automatically convert `||` to `??` within a mixed logical expression, as we cannot tell the intended precedence of the operator. Note that by design, `??` requires parentheses when used with `&&` or `||` in the same expression.
 
+### `ignorePrimitives`
```
JoshuaKGoldberg, line 134: [Docs] Could you add some examples please? We generally try to -at least for newly added rule options- so users can see exactly how the code looks.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -7,10 +7,16 @@ import * as util from '../util';
...
     ignoreConditionalTests?: boolean;
-    ignoreTernaryTests?: boolean;
     ignoreMixedLogicalExpressions?: boolean;
-    allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing?: boolean;
+    ignorePrimitives?: {
+      bigint?: boolean;
+      boolean?: boolean;
+      number?: boolean;
+      string?: boolean;
+    };
```
JoshuaKGoldberg, line 18: [Non-actionable] In theory we could allow this to be a `boolean` to configure all of the primitives at once. I think it's fine to leave that as a follow-up to this PR.

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
```diff
@@ -751,5 +759,129 @@ declare const c: ${type};
...
+      },
+      {
+        ignoreablePrimitive: ['number', 'bigint'],
+        literalPrimitive: '0 | 1 | 0n | 1n',
+      },
+      {
+        ignoreablePrimitive: ['boolean'],
+        literalPrimitive: 'true | false | null',
+      },
+    ].map<TSESLint.InvalidTestCase<MessageIds, Options>>(
```
JoshuaKGoldberg, line 825: > `.map`

[Testing] I can't find a tracking right now, but I'm slowly getting more and more sour on generating tests this way. I'm finding it very difficult to read through them and the generation code obfuscates what the test actually does. Could you just have manually written error cases for these please?

JoshuaKGoldberg (changes requested):
Code looks great! Just requesting changes on a docs example & clearer tests now. Thanks! 🚀

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
omril1: I understand what you mean here, I'm all for flatting them, I thought it was how it works here based on a small sample of tests I've seen

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
omril1: Good idea, leaving it for now.

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md
omril1: I tried to look at the style for examples with options, this rule seems unique in how the examples are presented but decided to keep it the same since I've noticed bradzacher did the same

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
omril1: Do I need to open an issue for it or can I reference the same one?

JoshuaKGoldberg: We should have a new issue for tracking. It gets confusing having multiple PRs target the same issue.

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md
JoshuaKGoldberg: Heh yeah we'll eventually clean all these up... eventually. Thanks!

JoshuaKGoldberg (approved):
💯 looks great, thanks for all the touchups! Sorry to keep you waiting 😅 this really is a great PR.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
JoshuaKGoldberg: #7180 ✨

PR #6488, opened 2023-02-18
test: allow to execute tests on windows

armano2:
Partially fix running tests on windows machines, with this change most tests do pass with exception to

https://github.com/typescript-eslint/typescript-eslint/blob/98dec37c407717a0ff1002bf483113d15bbb4958/packages/utils/tests/eslint-utils/rule-tester/RuleTester.test.ts#L146-L188

filename is not correctly normalized in this test, this can be fixed after updating jest to 29.4 with EqualityTesters - https://jestjs.io/docs/expect#expectaddequalitytesterstesters

### DIFF
--- a/packages/shared-fixtures/fixtures/typescript/basics/type-import-type.src.ts
+++ b/packages/shared-fixtures/fixtures/typescript/basics/type-import-type.src.ts
@@ -1,2 +1,0 @@
-type A = typeof import('A');
-type B = import('B').X<Y>;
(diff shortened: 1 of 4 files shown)


packages/shared-fixtures/fixtures/typescript/basics/type-import-type.src.ts
```diff
@@ -1,2 +0,0 @@
-type A = typeof import('A');
-type B = import('B').X<Y>;
```
armano2, line 2: shared-fixtures has been removed in v6 and test is already correctly migrated, this is most leftover from merge conflict

JoshuaKGoldberg (approved):
As a former Windows user, I very much appreciate you sending this in 😄  thanks!

Filed #6490 too.

PR #6496, opened 2023-02-20
fix(website): correct crash when switching to dark/light mode in playground

### LINKED ISSUE #6335: Bug: Playground crashes when switching between light and dark mode

### Relevant Package

website

### Playground Link

https://typescript-eslint.io/play#ts=4.9.3&sourceType=module&code=C4TwDgpgBAglC8UDeUC+BYAUFgJhAxgDYCGATtAG5lQAeAXFEA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
Visit provided playground link to see the issue.
```

### ESLint Config

```javascript
N/A (see playground link)
```

### tsconfig

```jsonc
N/A (see playground link)
```

### Expected Result

The playground should not crash when switching from light to dark mode, or vice-versa. Playground state should be preserved when switching.

### Actual Result

The playground state may disappear and reappear after around 5 seconds of waiting. Then the page crashes.

### Additional Info

Here's a video showing the issue:

https://user-images.githubusercontent.com/567041/212404261-c2f11517-4502-4311-833d-a1b063742c96.mp4

Note that in this video I don't paste in any code -- the page loads up blank, then when I hit the dark mode button, after several seconds the code appears. Then it crashes.

The video shows the page crashing after I refresh it following a previous crash, but this also occurs for me in a fresh incognito session, ie. without any service workers, cookies, localstorage or other persistent state set.

### Versions

- Firefox 108
- Chrome 108

armano2:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6335

### DIFF
--- a/packages/website/src/components/editor/useSandboxServices.ts
+++ b/packages/website/src/components/editor/useSandboxServices.ts
@@ -145,7 +145,10 @@ export const useSandboxServices = (
         model.dispose();
       }
     };
-  }, [props.ts, colorMode, props.jsx, onLoaded]);
+    // colorMode can't be reactive here, this is going to cause crash while switching dark/light mode
+    // updating of colorMode is handled in LoadedEditor
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props.ts, props.jsx, onLoaded]);
 
   return services;
 };
--- a/packages/website/src/components/editor/useSandboxServices.ts
+++ b/packages/website/src/components/editor/useSandboxServices.ts
@@ -147,8 +147,9 @@ export const useSandboxServices = (
     };
     // colorMode can't be reactive here, this is going to cause crash while switching dark/light mode
     // updating of colorMode is handled in LoadedEditor
+    // jsx can't be reactive too, because it's used only on first load
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [props.ts, props.jsx, onLoaded]);
+  }, [props.ts, onLoaded]);
 
   return services;
 };
--- a/packages/website/src/components/editor/useSandboxServices.ts
+++ b/packages/website/src/components/editor/useSandboxServices.ts
@@ -145,9 +145,8 @@ export const useSandboxServices = (
         model.dispose();
       }
     };
-    // colorMode can't be reactive here, this is going to cause crash while switching dark/light mode
-    // updating of colorMode is handled in LoadedEditor
-    // jsx can't be reactive too, because it's used only on first load
+    // colorMode and jsx can't be reactive here because we don't want to force a recreation
+    // updating of colorMode and jsx is handled in LoadedEditor
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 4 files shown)


packages/website/src/components/editor/useSandboxServices.ts
```diff
@@ -145,7 +145,10 @@ export const useSandboxServices = (
         model.dispose();
       }
     };
-  }, [props.ts, colorMode, props.jsx, onLoaded]);
+    // colorMode can't be reactive here, this is going to cause crash while switching dark/light mode
+    // updating of colorMode is handled in LoadedEditor
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props.ts, props.jsx, onLoaded]);
```
armano2, line 151: we have to make sure that this value is non reactive,

@JoshuaKGoldberg if you don't like eslint disable we can save this value to state and never mutate it

```
const [darkMode] = setState(() => colorMode)
```

packages/website/src/components/editor/useSandboxServices.ts
```diff
@@ -145,7 +145,10 @@ export const useSandboxServices = (
         model.dispose();
       }
     };
-  }, [props.ts, colorMode, props.jsx, onLoaded]);
+    // colorMode can't be reactive here, this is going to cause crash while switching dark/light mode
```
JoshuaKGoldberg, line 148: Could you elaborate here on why?

JoshuaKGoldberg (approved):
💯 thanks! This seems cleaner to me.

packages/website/src/components/editor/useSandboxServices.ts
armano2: not sure what else you want me to write here, i pushed small fix for jsx, (chaning jsx option was causing crash to)

JoshuaKGoldberg: I mean, _why_ would providing `colorMode` cause a crash?

armano2: hmm, how to put it in proper sentence...

> in this context `useEffect` is used as constructor/destructor, constructor uses singletor to generate code and destructor is triggered always causing code to not be correctly initialized.
> This forces external lib to recreate editor when we want to update theme or one of options.

i'm bad at explaining stuff xd

JoshuaKGoldberg: Ah I think I got it! How about...

```suggestion
    // colorMode can't be reactive here because we don't want to force a recreation
    // of the editor whenever options or theme changes
```

JoshuaKGoldberg (dismissed):
🔥

bradzacher (approved):

PR #6516, opened 2023-02-23
chore(website): correct issues with scrollbars in playground

### LINKED ISSUE #6108: Website: Horizontal AST scrolls also irreversibly scroll the code editor
### Suggestion

1. Setup [[playground demo](https://typescript-eslint.io/play/#ts=4.9.3&jsx=true&sourceType=module&showAST=ts&code=MYewdgziA2CmB00QHMAUAiAhgI2AE1gDNkALASwCsBraAWzBAAcBHAJwgBcBXAN3QEoA3ACggA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA)]
    * Have a line of code such as `console.log("abcdefghijklmnop");` in code editor
    * Enable an AST viewer 
    * Shrink your browser width so that the AST viewer needs to horizontally scroll to see everything
1. Click within the code
1. Click to the right of the code

The code editor will be scrolled horizontally, with no way to take it back:

https://user-images.githubusercontent.com/3335181/204148118-81087f0d-751c-4550-a531-94ff289678bb.mov

cc @jimsu2012, re #5667 - not sure if caused by that one, but seems at least related & in the same code area.

armano2:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6108

## Overview

image

### DIFF
--- a/packages/website/src/components/EditorTabs.tsx
+++ b/packages/website/src/components/EditorTabs.tsx
@@ -2,20 +2,21 @@ import EditIcon from '@site/src/icons/edit.svg';
 import React from 'react';
 
 import styles from './Playground.module.css';
+import type { TabType } from './types';
 
-export interface FileTabsProps<T> {
-  readonly tabs: { value: T; label: string }[];
-  readonly activeTab: T;
-  readonly change: (tab: T) => void;
-  readonly showModal?: () => void;
+export interface FileTabsProps {
+  readonly tabs: TabType[];
+  readonly activeTab: TabType;
+  readonly change: (tab: TabType) => void;
+  readonly showModal: () => void;
 }
 
-export default function EditorTabs<T>({
+export default function EditorTabs({
   tabs,
   activeTab,
   change,
   showModal,
-}: FileTabsProps<T>): JSX.Element {
+}: FileTabsProps): JSX.Element {
   return (
     <div className={styles.tabContainer}>
       <div role="tablist">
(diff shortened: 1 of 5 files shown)


packages/website/src/components/Playground.module.css
```diff
@@ -43,6 +43,7 @@
   height: calc(100% - var(--ifm-navbar-height));
   top: var(--ifm-navbar-height);
   z-index: calc(var(--ifm-z-index-fixed) - 1);
+  background: var(--playground-main-color);
```
JoshuaKGoldberg, line 46: Is this needed? The only change I can see is in dark mode, and adding this background in makes the tabs less distinct from the background.

```suggestion
```

packages/website/src/components/Playground.module.css
```diff
@@ -53,7 +54,8 @@
 }
 
 .tabCode {
-  height: calc(100%);
+  height: calc(100% - 32px);
+  background: var(--code-editor-bg);
```
JoshuaKGoldberg, line 58: Same here - what does this do?

```suggestion
```

packages/website/src/components/EditorTabs.tsx
```diff
@@ -25,17 +24,17 @@ export default function EditorTabs({
...
-              aria-selected={activeTab === item}
-              disabled={activeTab === item}
-              onClick={(): void => change(item)}
+              key={item.label}
+              aria-selected={activeTab === item.value}
+              disabled={activeTab === item.value}
+              onClick={(): void => change(item.value)}
             >
-              {item}
+              {item.label}
```
JoshuaKGoldberg, line 32: These code changes don't look like they actually fix the issue? Is there a benefit to them?

JoshuaKGoldberg (changes requested):
Code looks great for fixing the overflow/width issues, yay!

Requesting changes because it looks like some other changes got mixed in?

packages/website/src/components/Playground.module.css
armano2: yes, if there is delay while updating layout of editor, we can see 2 distinct backgrounds,
i'm going to update tabs to have correct bg

armano2: i'm going to recheck if it looks ok without background changes :)

armano2: ok i reverted color changes, but those colors where afecting only 

image image

and bacgkground that is under editor while we are loading it or resizing

image
image

JoshuaKGoldberg (approved):
Got it - this looks great, thanks!~

PR #5384, opened 2022-07-26
fix: rename typeParameters to typeArguments where needed

JoshuaKGoldberg:
BREAKING CHANGE:
Adds duplicate properties on the AST and deprecates their old names

## PR Checklist

-   [x] Addresses an existing open issue: fixes #146
-   [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
-   [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

Renames `typeParameters` to `typeArguments` in locations that are indeed providing type _arguments_.

To recap the terminology:

* An _argument_ is something you provide to a recipient, such as a type provided explicitly to a call expression
* A _parameter_ is how the recipient receives what you provide, such as a function declaration's generic type parameter

Marked as a breaking change even though the old, now-`@deprecated` property is still there.

PR #6525, opened 2023-02-24
feat(typescript-estree): warn on deprecated AST property accesses

### LINKED ISSUE #6469: Enhancement: Add test-only console warnings to deprecated AST properties

### Relevant Package

typescript-estree

### Description

From @bradzacher's suggestion in https://github.com/typescript-eslint/typescript-eslint/pull/5384#issuecomment-1416962804, which renamed many properties from `typeParameters` to `typeArguments`:

> Thinking about this in terms of the broader ecosystem and the impacts - I wonder if we want to give consumers one major version to migrate? A hard cut over might be a bit difficult for the ecosystem to come handle as it would involve every single plugin updating their code (which could leave users in an impossible to upgrade state for some time).
> 
> How about this: this version we add the new property and we mark the old property as `@deprecated`, then in v7 we remove the old property entirely.
> 
> If we wanted we could also do something like define a getter that throws when in a test environment to help flag plugins that they should migrate.

#5384 kept the old properties as `@deprecated` but didn't add a getter. 

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6469

## Overview

Follows up on #5384 by wrapping the `@deprecated` properties with a getter that, when a new `suppressDeprecatedPropertyWarnings` option isn't enabled, `console.warn`s about the properties.

`suppressDeprecatedPropertyWarnings` defaults to `process.env.NODE_ENV !== "test"` ([false in Jest](https://jestjs.io/docs/environment-variables) and similar). Meaning, it'll only log in tests by default.

This also has the benefit of making the new property [_not_ `enumerable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#enumerable), which cleans up the AST snapshots and playground representation of the AST to not have the seemingly duplicate properties. 🧹


docs/linting/Troubleshooting.mdx
```diff
@@ -246,6 +246,32 @@ See [this issue comment](https://github.com/typescript-eslint/typescript-eslint/
...
+
+The [typescript-eslint v6 release post](https://deploy-preview-6515--typescript-eslint.netlify.app/blog/announcing-typescript-eslint-v6-beta) has more information on typescript-eslint v6.
+
 ## My linting feels really slow
 
 If you think you're having issues with performance, see our [Performance Troubleshooting documentation](./troubleshooting/Performance.md).
+
+```
+
+```
```
bradzacher, line 277: ¿que?

bradzacher (approved):
this is looking good to me! should be really helpful to help people migrate.

I wonder if we should only log during testing scenarios..? It might spam users otherwise. Hmmm

JoshuaKGoldberg:
> only log during testing scenarios

Right now that's the default:

https://github.com/typescript-eslint/typescript-eslint/blob/0c7af5262634d131859eaae3b4d1c26c33bbb06a/packages/typescript-estree/src/parseSettings/createParseSettings.ts#L76-L78

bradzacher:
@JoshuaKGoldberg rebase this on v6 and fix the build and you're good to land!

fisker:
@JoshuaKGoldberg This change breaks Prettier, we do an AST transform after parse, which will override some of those properties. I decide not to use the new properties because babel still use those properties, I was thinking update print logic after babel adopt those changes. Do you think maybe we should add a setter? [CI Log](https://github.com/prettier/prettier/actions/runs/4445538911/jobs/7804778641?pr=14429#step:4:60)

fisker:
Or maybe when `suppressDeprecatedPropertyWarnings: false`, not to use `Object.defineProperty`?

PR #6540, opened 2023-02-27
fix(typescript-estree): do not call export of clearProgramCache

sandersn:
Calling it exports `undefined` instead of making it an alias of `clearCaches`, which I believe was the intent.

On DefinitelyTyped, [I worked around the problem](https://github.com/microsoft/DefinitelyTyped-tools/pull/620); clearProgramCaches isn't needed now anyway.

### DIFF
--- a/packages/typescript-estree/src/clear-caches.ts
+++ b/packages/typescript-estree/src/clear-caches.ts
@@ -18,4 +18,4 @@ export function clearCaches(): void {
 }
 
 // TODO - delete this in next major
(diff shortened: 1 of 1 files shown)


bradzacher (commented):
🤦‍♂️ Thanks for this.

Note you should switch DT to use the other API (`clearCaches`) which will nuke all our caches instead of just one of them.

bradzacher (approved):

PR #6542, opened 2023-02-28
feat(typescript-estree): lazily compute loc

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: #6366

## Overview

This PR changes our AST conversion to lazily compute the `.loc` for all nodes, comments, and tokens.

TS does not compute node location during the parsing sequence. This means that in order to calculate the location, we need to ask TS to convert the range to a location - which involves doing a binary search of the line:range mapping.

Ideally TS would calculate and store these as it goes - but understandably they don't because TS doesn't need the information for the most part!

## Perf Result

This change reduces time spent in `getLineAndCharacterFor` in our codebase from 231ms to 68ms, however the time spent in `astConverter` was mostly a neutral change - suggesting that the code of just declaring the getters and setters cost the same as the time saved.

I'm really surprised that accessors are so expensive to define. Really sad that this is a dead end. I'm putting this up and closing it immediately just so there's a clear log of things for people to see.

### DIFF
--- a/packages/typescript-estree/src/convert-comments.ts
+++ b/packages/typescript-estree/src/convert-comments.ts
@@ -26,7 +26,6 @@ export function convertComments(
           ? AST_TOKEN_TYPES.Line
           : AST_TOKEN_TYPES.Block;
       const range: TSESTree.Range = [comment.pos, comment.end];
-      const loc = getLocFor(range[0], range[1], ast);
 
       // both comments start with 2 characters - /* or //
       const textStart = range[0] + 2;
@@ -40,7 +39,7 @@ export function convertComments(
         type,
         value: code.slice(textStart, textStart + textEnd),
         range,
-        loc,
+        loc: getLocFor(range[0], range[1], ast),
       });
     },
     ast,
--- a/packages/typescript-estree/src/convert.ts
+++ b/packages/typescript-estree/src/convert.ts
@@ -3,7 +3,7 @@
 import * as ts from 'typescript';
 
 import { getDecorators, getModifiers } from './getModifiers';
-import type { TSError } from './node-utils';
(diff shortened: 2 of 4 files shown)


armano2:
wdyt about changing that entire loc is a getter? that should save some time on declaring so many getters, we could make it, I don't think we have to worry about setter

bradzacher:
@armano2 I actually went down that route first before moving down one layer.
It was unfortunately about the same performance (it actually trended slightly negative, adding a dozen or so MS in some runs).

I think that `Object.defineProperty` is even slower than the accessor syntax because of the additional checks it has to do over just defining an object with accessors.

> I don't think we have to worry about setter

We do need a setter, sadly, as there's a few places during the conversion that we change the default location for a node (for example when separating out `export` to its own node).

armano2:
when separating out export to its own node)., we actually don't need a setter for our uses,

we can rely on `range[0]` and `range[1]` from current object and with that loc is going to react to all changes

bradzacher:
Sorry, yes, what I meant is that we need a setter for the lower-level approach this PR currently represents.

For a higher-level approach we can remove the setter. My first attempt (not captured in a commit) was using `Object.defineProperty` in exactly this way - delete the `.loc` assignments and rely solely on the `.range`.
However that doesn't change the performance appreciable way. There's not any additional cost to defining a property with `Object.defineProperty` to be both a getter and a setter - either way you're defining one property.

`Object.defineProperty` is just slower than defining accessors directly on an object.

I thought about maybe spreading objects to instead use the syntax, but this ofc incurs the cost of spreading - which won't be great at scale.

I thought about declaring a few base classes for all our nodes like TS does so that we can have a getter declared on the prototype (i.e. declared exactly once) - and that may be a viable option! I just did not explore it yet as it's quite a large change.

Issue #6508, opened 2023-02-21
Repo: consistency issue with probablyup/wait-for-netlify-action on main runs


JamesHenry:
This affects website tests on main runs.

Reported here: https://github.com/probablyup/wait-for-netlify-action/issues/5

JoshuaKGoldberg:
Example logs from https://github.com/typescript-eslint/typescript-eslint/pull/6554:

```
...
URL https://6400d50730c8b300086cd150--typescript-eslint.netlify.app unavailable, retrying...
URL https://6400d50730c8b300086cd150--typescript-eslint.netlify.app unavailable, retrying...
Error: Timeout reached: Unable to connect to https://6400d50730c8b300086cd150--typescript-eslint.netlify.app
```

JamesHenry:
I have finally gotten around to resolving this by creating our own lightweight action within the repo:
https://github.com/typescript-eslint/typescript-eslint/blob/main/.github/actions/wait-for-netlify/index.js

Here are the tests passing on the latest `main` commit after that action went in in #7083:
https://github.com/typescript-eslint/typescript-eslint/actions/runs/5169175265/jobs/9311306701

NOTE: In the time that these tests have been disabled it seems that all the Playground ones stopped working, so I needed to skip them to proceed:
https://github.com/typescript-eslint/typescript-eslint/pull/7083/files#diff-2a06cb80e80a23b70ff5ed6cccd175ae80e657efbc6736d910d9f0e705fb7925R5

Hopefully somebody how is more familiar with the Playground can take a look

PR #6554, opened 2023-03-01
feat(eslint-plugin): add allowNever support to restrict-template-expressions

### LINKED ISSUE #5325: Enhancement: [restrict-template-expressions] allow expressions with `never` type

https://typescript-eslint.io/rules/restrict-template-expressions#allownullish

### Description

The restrict-template-expressions rule currently complains if the template expression has type `never`. This occurs, for example, if you have code like this:

```typescript
// more variants may be added to Foo in the future
type Foo =
  | {type: "a", value: number};

function checkFoosAreMatching(foo1: Foo, foo2: Foo) {
  if (foo1.type !== foo2.type) {
    // since Foo currently only has one variant, this code is never run,
    // and `foo1.type` has type `never`.
    throw new Error(`expected ${foo1.type}, found ${foo2.type}`);
                                                  // ^ Invalid type "never" of template literal expression.
  }
}
```

I don't see the benefit of the lint warning here, which complains that `foo1.type` and `foo2.type` have type `never` and says, `Invalid type "never" of template literal expression.`

I am suggesting changing the lint rule's behaviour to allow template expressions to have the type `never`. This could be done as a new option called `allowNever`, although I personally don't see the point in giving the option to disallow `never` in template expressions. Why should this lint rule be complaining about dead code?

### Fail

```typescript
const foo: "a" | null;
const bar: "a";

if (foo !== bar) {
    // `foo` has type `null`, which is not allowed
    // with the default options for this rule
    throw new Error(`${foo} and ${bar} are not equal);
}
```

### Pass

```typescript
const foo: "a";
const bar: "a";

if (foo !== bar) {
    // `foo` and `bar` have type `never`, which should be allowed
    // by this rule
    throw new Error(`${foo} and ${bar} are not equal);
}
```

### Additional Info

This same logic should also apply to similar rules like `restrict-plus-operands` (no need to complain that an operand has type `never`), although I haven't checked to see how that rule behaves for `never` operands.

domdomegg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5325

## Overview

Adds `allowNever` support, which makes doing things like exhaustive switching easier. Uses the existing utils in the same way as `allowAny` does, and have added relevant and realistic tests.


domdomegg:
Hey @bradzacher - I think this PR is good to go but the 'Website tests' build is failing due to Netlify not being ready. I don't think I've changed anything meaningful here, and it seems also to be failing on main. I'm wary that the query listed in the contributing docs filters out this PR because of the failing status, and wanted to check this could still get reviewed.

(apologies if you're the wrong person to ping - couldn't find a doc on process for these kinds of questions in the contributing docs)

packages/eslint-plugin/tests/rules/restrict-template-expressions.test.ts
```diff
@@ -227,6 +227,44 @@ ruleTester.run('restrict-template-expressions', rule, {
...
+        
+        function checkFoosAreMatching(foo1: Foo, foo2: Foo) {
+          if (foo1.type !== foo2.type) {
+            // since Foo currently only has one variant, this code is never run, and \`foo1.type\` has type \`never\`.
+            throw new Error(\`expected \${foo1.type}, found \${foo2.type}\`);
+            // ^ Invalid type "never" of template literal expression.
+          }
+        }
+      `,
+    },
```
JoshuaKGoldberg, line 267: These test cases for the default case are nice and thorough. Though they all roughly boil down to concatenating something of `never` type. Which means they can be replaced with the simpler:

```ts
declare const value: never;

const stringy = `${value}`;
```

If you strongly feel a desire to have more thorough creations of `never` I won't block on it. 😄

JoshuaKGoldberg, line 267: I don't see any existing tests for the `never` type _without_ the `allowNever` option. Unless I'm missing them (am I?), could you add one in? We should make sure the new logic only applies when the option is enabled.

JoshuaKGoldberg (changes requested):
Looks nice and clean, thanks for sending this in @domdomegg! Just blocking on adding more tests (or pointing me to where I've missed them). 🚀

JoshuaKGoldberg:
> 'Website tests' build is failing due to Netlify not being ready.

Blurgh, yeah, this is unfortunate. #6508. Don't worry about it - it's out of your control and unrelated to this PR.

> (apologies if you're the wrong person to ping - couldn't find a doc on process for these kinds of questions in the contributing docs)

Looks like we need to clarify that! Thanks for mentioning @domdomegg 😄. Just filed #6584:

> Today, the right step is to just post a comment. You don't need to tag anybody - we'll see it regardless of the tag (the joy and pain of GitHub notifications 🥲). One of us will drop by and help when we can.

packages/eslint-plugin/tests/rules/restrict-template-expressions.test.ts
domdomegg: Good idea. Have added a check that the rule still fires when allowNever is disabled.

domdomegg: That makes sense! I've added a more simple test as suggested.

I think as extra security, and to provide clarity on what the allowNever option is trying to achieve I'll leave the existing tests.

JoshuaKGoldberg: Awesome, thanks! Whenever you feel this is ready for re-review, you can re-request my review.

Btw we can ignore the failing website tests. #6508 😢

JoshuaKGoldberg (approved):
Super, thanks @domdomegg! ✨

JoshuaKGoldberg (changes requested):
Err, once the unit test failure is resolved, that is*

https://github.com/typescript-eslint/typescript-eslint/actions/runs/4420331029/jobs/7749934631?pr=6554

```plaintext
Summary of all failing tests
 FAIL  tests/rules/restrict-template-expressions.test.ts
  ● restrict-template-expressions › invalid › 
        declare const value: never;
        const stringy = `${value}`;
      
    assert.strictEqual(received, expected)
    Expected value to strictly be equal to:
      26
    Received:
      28
    
    Message:
      Error column should be 26
```

JoshuaKGoldberg (approved):
:shipit: awesome!

PR #6555, opened 2023-03-02
test(ast-spec): snapshot codeframe of error

fisker:
## PR Checklist

## Overview

During upgrade parser to v6 in Prettier, I found some error location should be improved, but before that, let's make the error visible?

### DIFF
--- a/packages/ast-spec/package.json
+++ b/packages/ast-spec/package.json
@@ -46,6 +46,7 @@
     "@babel/core": "*",
     "@babel/eslint-parser": "*",
     "@babel/parser": "*",
+    "@babel/code-frame": "*",
     "@microsoft/api-extractor": "^7.23.2",
     "@types/babel__core": "*",
     "glob": "*",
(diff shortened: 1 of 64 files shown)


bradzacher (commented):
this is a great idea!
I wonder if we should be capturing an end as well as part of the `TSError`?

That way we could throw errors on ranges instead of just on single characters.
Looks like code-frame also supports an end range, so we would even be able to snapshot it.

WDYT?

fisker:
> I wonder if we should be capturing an end as well as part of the TSError?

Prettier throws errors include a start and an end, see [this diff](https://github.com/prettier/prettier/actions/runs/4301173716/jobs/7498092190#step:5:33).
I hope to merge this one first, and change the errors to add more information later.

fisker:
@bradzacher Don't know how to fix [this](https://github.com/typescript-eslint/typescript-eslint/actions/runs/4310317455/jobs/7518637062#step:4:38), can you help?

bradzacher (approved):

bradzacher:
oops - i didn't see the error before merging 🤦

bradzacher:
Should be fixed by f98afd2eaddab2a74f22f737d89aa89147c7fc9b

PR #6556, opened 2023-03-02
feat: improve error location

fisker:
## PR Checklist


packages/typescript-estree/src/node-utils.ts
```diff
@@ -611,21 +635,39 @@ export class TSError extends Error {
...
+    return this.location.start.offset;
+  }
+
+  get lineNumber(): number {
+    return this.location.start.line;
+  }
+
+  get column(): number {
+    return this.location.start.column;
+  }
```
fisker, line 649: Should those kept?

packages/typescript-estree/src/convert.ts
```diff
@@ -2981,13 +2985,17 @@ export class Converter {
...
   }
 
-  #throwErrorIfDeprecatedPropertyExists<Node extends ts.Node>(
-    node: Node,
-    property: unknown,
-    message: string,
-  ): void {
-    if (property) {
-      throw createError(this.ast, node.pos, message);
+  #throwError(node: ts.Node | number, message: string): void {
```
armano2, line 2988:
```suggestion
  #throwError(
    node: ts.Node | number,
    message: string,
  ): asserts node is never {
```

with that you cna remove all type casts that you added eg,

```diff
- property: this.convertJSXIdentifier(node.name),
+ property: this.convertJSXIdentifier(node.name as ts.Identifier),
```

packages/typescript-estree/src/node-utils.ts
fisker: Is this the answer? https://github.com/eslint/eslint/blob/b09a512107249a4eb19ef5a37b0bd672266eafdb/lib/linter/linter.js#LL853C32-L853C32

armano2: that's correct, we have to keep `lineNumber` and `column` as they are used in eslint
as for `index` if i remember correctly, this prop was needed in older version of eslint

packages/typescript-estree/src/convert.ts
fisker: Thanks!

armano2 (approved):

packages/typescript-estree/src/convert.ts
```diff
@@ -2981,13 +2985,17 @@ export class Converter {
...
   }
 
-  #throwErrorIfDeprecatedPropertyExists<Node extends ts.Node>(
-    node: Node,
-    property: unknown,
-    message: string,
-  ): void {
-    if (property) {
-      throw createError(this.ast, node.pos, message);
+  #throwError(node: ts.Node | number, message: string): asserts node is never {
```
armano2, line 2988:
```suggestion
  #throwError(node: ts.ReadonlyTextRange | number, message: string): asserts node is never {
```

and 

```
      start = node.start;
      end = node.end;
```

packages/typescript-estree/src/convert.ts
```diff
@@ -3033,8 +3033,8 @@ export class Converter {
   #checkIllegalDecorators(node: ts.Node): void {
     if (nodeHasIllegalDecorators(node)) {
       this.#throwUnlessAllowInvalidAST(
-        // eslint-disable-next-line deprecation/deprecation
-        node.decorators?.[0] ?? node,
+        // @ts-expect-error -- this is safe as it's guarded
+        node.illegalDecorators?.[0] ?? node,
```
armano2: good catch, i'v tried to fix all conflicts and there was a lot, but it seem i missed this one

PR #6568, opened 2023-03-04
feat(website): preserve RulesTable filters state in searchParams

### LINKED ISSUE #6171: Website: Sync Supported Rules filter state in the URL

### Suggested Changes

I want to share a link to https://typescript-eslint.io/rules/#supported-rules that prepopulates the table filter to just rules that require type information. But there's no way to do that 😢. It'd be great if the component could sync state with the URL, so a link like `https://typescript-eslint.io/rules/#supported-rules?supported-rules=requires-type-information` can make the page filter the table to just those.

### Affected URL(s)

https://typescript-eslint.io/rules

hasparus:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6171

## Overview

Howdy! 👋 

I added syncing rules table filters state to URL search params.

This is a draft because:

- [ ] I'll describe why it turned out to be so much code. Simple implementation with `useLayoutEffect` and `useState` initialiser isn't enough
- [ ] and I'll add a Playwright test case


hasparus:
It seems that the tests are failing, because they wait for the deploy preview and they timeout before it deploys.

image

JoshuaKGoldberg:
> It seems that the tests are failing, because they wait for the deploy preview and they timeout before it deploys.

Yeah 😞. #6508

packages/website/tests/rules.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+import { expect, test } from '@playwright/test';
+
+test.describe('Rules Page', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/rules');
+  });
+
+  test('Accessibility', async ({ page }) => {
+    await new AxeBuilder({ page }).analyze();
+  });
```
JoshuaKGoldberg, line 11: 😍 thanks for adding this in!

_(I'll continue reviewing soon, just wanted to express appreciation sooner)_

packages/website/src/components/RulesTable/index.tsx
```diff
@@ -122,78 +123,57 @@ export default function RulesTable({
 }: {
   extensionRules?: boolean;
 }): JSX.Element {
+  const [filters, changeFilter] = useRulesFilters(
+    extensionRules ? 'extension-rules' : 'supported-rules',
+  );
```
JoshuaKGoldberg, line 128: [Cleanup] I'm not a big fan of hardcoding IDs on boolean logic. It tends to break over time, and makes it harder to understand where the IDs come from. How about giving the `RulesTable` component a mandatory `id` prop that's then passed directly to `useRulesFilters`?

JoshuaKGoldberg (approved):
Yup, this looks great. Thanks for sending it in @hasparus!

Just one suggestion on refactoring `RulesTable`<>`useRulesFilters` a bit. What do you think?

I'm down to merge as-is if you don't like the suggestion, or make it myself if you don't have time.

packages/website/src/components/RulesTable/index.tsx
hasparus: I agree about hardcoding is no boolean logic, but I feel that 

```<RulesTable id="extension-rules">```

might be a bit confusing given that the heading above receives `id` attribute from the markdown pipeline.

image

How about a prop `ruleset: "supported-rules" | "extension-rules"` that would subsume the boolean flag **and** be usable as a search param key?

Comment (JoshuaKGoldberg) [reply packages/website/src/components/RulesTable/index.tsx]: Yeah I like that a lot!

Comment (hasparus) [reply packages/website/src/components/RulesTable/index.tsx]: I added it in https://github.com/typescript-eslint/typescript-eslint/pull/6568/commits/642031fe921885c740f552bc741865bffc53686a, but it seems there are indirect codecov changes that are failing. Do you think it's connected to my change?

**Comment by hasparus** [comment]
@JoshuaKGoldberg The deploy has finished in 3 minutes, but the status check's been hanging for 19 hours. Is there a way we can stop it or restart it? This seems like a bug in the Netlify integration.

image

File: packages/website/src/components/RulesTable/index.tsx [hunk 3]
```diff
@@ -224,3 +203,98 @@ export default function RulesTable({
...
+  typeInformation: 'neutral',
+};
+
+function useRulesFilters(
+  paramsKey: string,
+): [FiltersState, (category: FilterCategory, mode: FilterMode) => void] {
+  const [state, setState] = useState(neutralFiltersState);
+
+  useIsomorphicLayoutEffect(() => {
+    const search = new URLSearchParams(window.location.search);
```
Josh-Cena, line 228: You can replace `window.location` with `useLocation()` so you get SSR safety for free

Josh-Cena (commented):
The implementation looks mostly good to me, but as some extra reading, you may want to try `useSyncExternalStore` ([interesting read](https://thisweekinreact.com/articles/useSyncExternalStore-the-underrated-react-api)) ^^

packages/website/src/components/RulesTable/index.tsx
hasparus: I expected `useLocation()` to trigger re-renders whenever the location changes, as the docs would suggest: https://reactrouter.com/en/main/hooks/use-location

So using `location` from `useLocation` here wouldn't be the same logic, right?
And I'm in a `useLayoutEffect` which does not run on the server, so accessing `window.location` works.

Could you tell me more what do you mean by "SSR safety for free"?

Notice that I had rewritten this logic to `react-router` specific hooks, what I described in **"Alternative solution"** in the original post. It turned out to be buggy and harder to reason about than the less elegant, naive solution with `setState` and `useLayoutEffect`, so I reverted that change — https://github.com/typescript-eslint/typescript-eslint/pull/6568/commits/786557fb52e5a18664f8291b853182a8cfd22a38.

> The implementation looks mostly good to me, but as some extra reading, you may want to try useSyncExternalStore ([interesting read](https://thisweekinreact.com/articles/useSyncExternalStore-the-underrated-react-api)) ^^

Nice article! Do you think I should rewrite this PR to use `useHistorySelector` presented in the post?

JoshuaKGoldberg:
(I'll defer to @Josh-Cena - who is more knowledgeable than me on Docusaurus best practices anyway!)

packages/website/src/components/RulesTable/index.tsx
Josh-Cena: I'm okay either way! I don't have strong opinions on this, and TBH my React is getting a bit rusty these days.

hasparus: > I mean you don't have to use useIsomorphicLayoutEffect and read from a client-only API inside. useLocation is available both server-side and client-side, so it's more robust.

Well, yes, that's true, but then we shouldn't use `useState`, because a change to the location will already re-render the component. Would you agree?

Would you prefer if I went back to the state from https://github.com/typescript-eslint/typescript-eslint/pull/6568/commits/7c2e81a4f61fcaf7cbb2559ada50ef0b2d3e9b2b? and write end-to-end tests to ensure it works?

> I'm okay either way!

I won't be able to contribute any code here this week, so while I agree the solution would be more elegant with `useHistorySelector`, this PR would need to wait.

packages/website/src/components/RulesTable/index.tsx
```diff
@@ -224,3 +204,97 @@ export default function RulesTable({
...
+
+const selectSearch: HistorySelector<string> = history =>
+  history.location.search;
+const getServerSnapshot = () => '';
+
+function useRulesFilters(
+  paramsKey: string,
+): [FiltersState, (category: FilterCategory, mode: FilterMode) => void] {
+  const history = useHistory();
+  const search = useHistorySelector(selectSearch, getServerSnapshot);
```
hasparus, line 231: @Josh-Cena I updated the implementation to use `useSyncExternalStore`.

packages/website/src/components/RulesTable/index.tsx
JoshuaKGoldberg: We can ignore codecov for these website PRs. We don't have great unit test coverage on it anyway. 🤷

JoshuaKGoldberg:
> Who should I bother when the Netlify GitHub integration acts out?

Oh sorry I missed this - it's been flaky for a while. We can ignore it. #6508 😞

packages/website/src/components/lib/useHistorySelector.ts
```diff
@@ -0,0 +1,17 @@
...
+  selector: HistorySelector<T>,
+  getServerSnapshot: () => T,
+) {
+  const history = useHistory();
+  return useSyncExternalStore(
+    history.listen,
+    () => selector(history),
+    getServerSnapshot,
+  );
+}
```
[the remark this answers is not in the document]
hasparus: What do you think about `src/components/hooks`? I feel like it belongs with `useHashState`, `useResizeObserver` and `useFocus`.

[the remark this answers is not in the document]
hasparus: or... should there be just one `hooks` directory, @armano2?
Not the topic of this PR but I guess I could move all 7 of those hooks to one dir.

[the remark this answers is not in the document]
armano2: > Not the topic of this PR but I guess I could move all 7 of those hooks to one dir.

thats what i did in #6656, website componetns are a little disorganized, its hard to find stuff out

armano2 (approved):
this looks great, if you fix those linting errors i think we are ready to merge

hasparus:
Thanks for the review and bearing with me, guys 🙏

Skn0tt:
> @Skn0tt sorry for pinging you here

Sorry for getting back so late, I was on PTO the past weeks. Not sure what happened here, but it looks like things are alright now? Let me know if not, and i'll see how I can help :)

PR #6573, opened 2023-03-06
chore: remove unnecessary argument when calling `node.getStart()`

fisker:
## PR Checklist

### DIFF
--- a/packages/typescript-estree/src/convert.ts
+++ b/packages/typescript-estree/src/convert.ts
@@ -167,7 +167,7 @@ export class Converter {
         ? findNextToken(nextModifier, this.ast, this.ast)
         : findNextToken(exportKeyword, this.ast, this.ast);
 
-      result.range[0] = varToken!.getStart(this.ast);
+      result.range[0] = varToken!.getStart();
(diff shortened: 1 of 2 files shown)


armano2:
there is actually a perf reason for passing root node as first argument,
if root is not this function is going to try to find root node in loop by checking parents

> getStart: https://github.com/microsoft/TypeScript/blob/main/src/services/services.ts#L364-L367
>
> getTokenPosOfNode: https://github.com/microsoft/TypeScript/blob/main/src/compiler/utilities.ts#L1140
>
> getSourceFileOfNode: https://github.com/microsoft/TypeScript/blob/main/src/compiler/utilities.ts#L860-L870
```ts
export function getSourceFileOfNode(node: Node | undefined): SourceFile | undefined {
    while (node && node.kind !== SyntaxKind.SourceFile) {
        node = node.parent;
    }
    return node as SourceFile;
}
```

fisker:
So, should we always pass ast?

bradzacher:
I'm going to close this for now as we'll probably want to go another route with this.
Instead of removing them, we want them for perf reasons, so we should add a lint rule if needs be.

PR #6591, opened 2023-03-09
chore: simplify `noFormat`

fisker:
## PR Checklist

### DIFF
--- a/packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
+++ b/packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
@@ -306,12 +306,8 @@ class RuleTester extends BaseRuleTester.RuleTester {
  * Simple no-op tag to mark code samples as "should not format with prettier"
  *   for the internal/plugin-test-formatting lint rule
  */
-function noFormat(strings: TemplateStringsArray, ...keys: string[]): string {
-  const lastIndex = strings.length - 1;
-  return (
-    strings.slice(0, lastIndex).reduce((p, s, i) => p + s + keys[i], '') +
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (approved):
Ha, TIL - nice. Thanks!

packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
```diff
@@ -306,12 +306,8 @@ class RuleTester extends BaseRuleTester.RuleTester {
...
  *   for the internal/plugin-test-formatting lint rule
  */
-function noFormat(strings: TemplateStringsArray, ...keys: string[]): string {
-  const lastIndex = strings.length - 1;
-  return (
-    strings.slice(0, lastIndex).reduce((p, s, i) => p + s + keys[i], '') +
-    strings[lastIndex]
-  );
+function noFormat(raw: TemplateStringsArray, ...keys: string[]): string {
+  return String.raw({ raw }, ...keys);
```
Josh-Cena, line 310: This won't work if there's any escape sequence in the string. See  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw#building_an_identity_tag. I'm not sure if this is acceptable, considering we may have escape sequences in the test cases (e.g. within string literals)

fisker: This function mean to use cooked value,not raw.

Josh-Cena: Ah, sorry I misread.

PR #6602, opened 2023-03-11
Revert "feat: remove `experimental-utils`"

JoshuaKGoldberg:
Reverts typescript-eslint/typescript-eslint#6468

Discussed privately - letting the changes in #6179 live for a little while.

### DIFF
--- a/.github/workflows/semantic-pr-titles.yml
+++ b/.github/workflows/semantic-pr-titles.yml
@@ -30,6 +30,7 @@ jobs:
             eslint-plugin
             eslint-plugin-internal
             eslint-plugin-tslint
+            experimental-utils
             parser
             scope-manager
             type-utils
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -258,6 +258,10 @@
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen",
       "skipFiles": [
+        "${workspaceFolder}/packages/experimental-utils/src/index.ts",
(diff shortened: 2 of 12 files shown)


bradzacher:
just to revisit this and have a chat in the open.

`utils` was officially released in 5.10.0 (18/1/2022 - 419 days ago) at which point we officially (but silently) deprecated `experimental-utils`.

#6179 (which causes a console log) on usage will create noise for consumers that have `experimental-utils` as a transient dependency, which isn't a *great* UX.

The two options we have are:
- explicitly `npm deprecate experimental-utils` and hard-lock it as a <v6 package
- release `experimental-utils` with the console logs and v6, and create some noise for the user (to hopefully help drive migrations), and then officially drop it with v7.

_Personally_ I think the former is better because (a) the deprecation message is install-time-only, so it won't spam users' lint runs and (b) hard-locking to v5 means packages can't be lazy - they **_have_** to migrate to keep updated.

I'd love to hear your thoughts again Josh, maybe @typescript-eslint/triage-team has some thoughts as well?

JoshuaKGoldberg:
FWIW this isn't a strong feeling on my end - I'm happy going along with either.

Part of why I was excited about going with #6179's console log is that @kball got to contribute, and I didn't want that contribution to end up going nowhere. But if (a) is the better case instead, that works too.

JoshuaKGoldberg:
(closing for now since it seems this PR isn't particularly popular 😄)

PR #6609, opened 2023-03-13
feat: remove moduleResolver API

### LINKED ISSUE #3798: Running with custom module resolver is 3x slower, as resolution is invalidated by TS compiler every time

bradzacher:
BREAKING CHANGE:
Removes `parserOptions.moduleResolver`.

## PR Checklist

- [x] Addresses an existing open issue: fixes #3798

## Overview

As we discussed in the thread - we're removing this API.
We don't have any evidence that it's widely used and there is a pretty significant maintenance burden to keep it.


JoshuaKGoldberg (approved):
🔪 🔪 🔪

PR #6612, opened 2023-03-13
chore: bump allowed version ranges to include TS 5.0

### LINKED ISSUE #6380: TypeScript 5.0 Support

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6380

## Overview

We should now have full support, so bump the ranges. TS5.0 should release just after our next release so bumping ahead of their release just to make it smoother for users.


JoshuaKGoldberg (dismissed):
🔥

PR #6616, opened 2023-03-13
chore: automatically cleanup caches when the relevant PR closes

bradzacher:
## Overview

Following this doc: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries
Adds a workflow that will run when a PR closes and cleanup the cache entries for that PR automatically.

This should auto-prune our cache storage to help keep it smaller and prevent possibility of accidental sharing


JoshuaKGoldberg (approved):
🆒

PR #6656, opened 2023-03-16
chore(website): [playground] add support for extends, ts types and more

### LINKED ISSUE #6037: Website: [Playground] Add "copy as JSON" option to AST Viewer

### Suggested Changes

The AST Viewer part of the playground can have a lot of objects in it:

Screenshot of a console.log('Hello, world!') ESTree AST on the playground in the viewer

It'd be nice to add a little _Copy as JSON_ button to the bottom of the AST viewer that copies its contents as a string to the user's clipboard.

### Affected URL(s)

armano2:
## PR Checklist

- [x] Addresses an existing open issue: #000

### New features

TODO: explain what changed

- Improved mobile experience
- Preview type information #6843
  image


packages/website/package.json
```diff
@@ -23,18 +23,20 @@
     "@mdx-js/react": "1.6.22",
     "@typescript-eslint/parser": "5.55.0",
     "@typescript-eslint/website-eslint": "5.55.0",
+    "@typescript/vfs": "^1.4.0",
+    "lz-string": "^1.5.0",
     "clsx": "^1.1.1",
     "eslint": "*",
     "json-schema": "^0.4.0",
     "json-schema-to-typescript": "^12.0.0",
+    "@monaco-editor/react": "^4.3.0",
```
armano2, line 32: TOOD: remove this

armano2:
@JoshuaKGoldberg when you find some time can you do preliminary review on this?
git generated way to many file changes :)

packages/website/src/icons/check.svg
armano2, line 1: @Josh-Cena do you think that we could expose those icons in docusaurus, if so i can make a PR for this there

packages/website/src/icons/copy.svg
```diff
@@ -5,5 +5,5 @@
   height="18"
   fill="currentColor"
 >
-  <path d="M3 3v14h2V5h12V3H3zm4 4v14h14V7H7z"/>
+  <path d="M19,21H8V7H19M19,5H8A2,2 0 0,0 6,7V21A2,2 0 0,0 8,23H19A2,2 0 0,0 21,21V7A2,2 0 0,0 19,5M16,1H4A2,2 0 0,0 2,3V17H4V3H16V1Z" />
```
armano2, line 8: @Josh-Cena do you think that we could expose those icons in docusaurus, if so i can make a PR for this there

packages/website-eslint/build.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    if (output.errors.length > 0) {
+      throw new Error('error occurred');
+    }
+  }
+}
+
+console.time('building eslint for web');
+
+buildPackage('index', './src/index.js')
```
armano2, line 132: instead of using rollup to bundle packages, i opted into using esbuild and generally cleaned up previous mess, 

as we are no longer mutating code without producing source maps, i noticed significat improvement to debugging in browser

packages/website-eslint/mocks/lru-cache.js
```diff
@@ -0,0 +1,14 @@
...
+  get(name) {
+    return this.cache.get(name);
+  }
+  set(name, value) {
+    this.cache.set(name, value);
+    return value;
+  }
+}
+
+module.exports = LruCache;
```
armano2, line 14: `lru-cache` SADLY dons't workwell in browser, but code can be relatively simplified if we don't care about long persistent runs (all the calls are static)

with that change i was able to stop changing how `semver` behaves and just import it as a whoe library (i noticed that previously not everything was working)

packages/website-eslint/mocks/path.js
```diff
@@ -206,11 +206,11 @@ export function basename(path, ext) {
   return f;
 }
 
-export function extname(path) {
+function extname(path) {
   return splitPath(path)[3];
 }
 
-export default {
+module.exports = {
```
armano2, line 213: we want to use mocks as cjs instead of modules, to improve build size as almsot all pakcages that we are importing are cjs there is no longer need to convert between formats back and forward

(few have a browser build witch we ar taking advantage of)

packages/website-eslint/build.ts
```diff
@@ -0,0 +1,139 @@
...
+    supported: {},
+    define: {
+      'process.env.NODE_ENV': '"production"',
+      'process.env.NODE_DEBUG': 'false',
+      'process.env.IGNORE_TEST_WIN32': 'true',
+      'process.env.DEBUG': 'false',
+      'process.emitWarning': 'console.warn',
+      'process.platform': '"browser"',
+      'process.env.TIMING': 'undefined',
+      'define.amd': 'false',
```
armano2, line 57: esbuild sadly don't support amd and its not able to process it, this line creates short circut of all amd wrapped packages (browser version)

packages/website-eslint/src/index.js
```diff
@@ -0,0 +1,27 @@
+define(['exports', 'vs/language/typescript/tsWorker'], function (e, _t) {
```
armano2, line 1: this is needed to expose package as amd

JoshuaKGoldberg:
I will... try 🥺 😅

packages/website/src/components/playground/config.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// export const esTsVersions: string[] = ['latest', ...versions.eslintPlugin];
+
+export const detailTabs = [
+  { value: false as const, label: 'Errors' },
+  { value: 'es' as const, label: 'ESTree' },
+  { value: 'ts' as const, label: 'TypeScript' },
+  { value: 'scope' as const, label: 'Scope' },
+  { value: 'types' as const, label: 'Types' },
```
JoshuaKGoldberg, line 34: Is there a reason to have both _TypeScript_ and _Types_? Could the two be combined into one view that has both TS AST nodes _and_ the type & symbol info for each?

packages/website/src/components/linter/bridge.ts
```diff
@@ -0,0 +1,95 @@
...
+  files.set(`/.eslintrc`, config.eslintrc);
+  files.set(`/tsconfig.json`, config.tsconfig);
+  files.set(`/file.${config.fileType}`, config.code);
+
+  const fileWatcherCallbacks = new Map<string, Set<ts.FileWatcherCallback>>();
+  const directoryWatcherCallback = new Set<ts.DirectoryWatcherCallback>();
+
+  const system = createSystem(files) as PlaygroundSystem;
+
+  system.watchFile = (path, callback, pollingInterval): ts.FileWatcher => {
```
JoshuaKGoldberg, line 33: Is setting `watchFile` or other properties on a TS system documented anywhere? I'm poking through https://github.com/microsoft/TypeScript-Website/tree/v2/packages/typescript-vfs and don't see anything.

packages/website/src/components/config/ConfigEslint.tsx
```diff
@@ -1,90 +1,60 @@
...
+        system.writeFile(
+          '/.eslintrc',
+          toJson({ ...(configObject ?? {}), rules: newRules }),
+        );
       }
+      onCloseProps(false);
     },
-    [onCloseProps, configObject],
+    [onCloseProps, configObject, system],
   );
```
JoshuaKGoldberg, line 58: This `onClose` pattern (here, `ConfigTypeScript.tsx`) means the configs never get "flushed" into the system & factored into linting / type checking until the tab is closed, right?

If so: can I suggest having them come into play in (perhaps debounced) real time? Eventually it'd be nice to have the playground be able to visualize multiple points of interest at a time. That could be showing all three files side-by-side, and/or showing both the ESLint and TS outputs at the same time.

I think doing so would mean the `if (isOpen)`s in this & the other `Config*.tsx` components would be unnecessary. The components would all effectively be "open" in their interpretation at all times, potentially actively writing to the system.

packages/website/src/components/config/ConfigTypeScript.tsx
```diff
@@ -1,74 +1,62 @@
...
+function ConfigTypeScript({
+  onClose: onCloseProps,
+  isOpen,
+  system,
+}: ConfigTypeScriptProps): JSX.Element {
   const [tsConfigOptions, updateOptions] = useState<ConfigOptionsType[]>([]);
   const [configObject, updateConfigObject] = useState<TSConfig>();
 
   useEffect(() => {
     if (isOpen) {
```
JoshuaKGoldberg, line 41: _(continuing my comment below on `onClose`)_ Actually, are these `if (isOpen) {` checks still necessary? I can't repro any bugs locally by removing this one.

packages/website/src/components/playground/PlaygroundRoot.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const [config, setConfig] = useHashState(defaultConfig);
+
+  const [system] = useState<PlaygroundSystem>(() => createFileSystem(config));
+  const [activeFile, setFileName] = useState(`file.${config.fileType}`);
+  const [editorFile, setEditorFile] = useState(`file.${config.fileType}`);
+
+  const [errors, setErrors] = useState<ErrorGroup[]>([]);
+  const [astModel, setAstModel] = useState<UpdateModel>();
+  const [esQueryFilter, setEsQueryFilter] = useState<ESQuery.Selector>();
+  const [showModal, setShowModal] = useState<string | false>(false);
```
JoshuaKGoldberg, line 35: Tangentially related to this PR, but I do wish the modal didn't take over the full playground. It'd be nice to limit it to just the editor tab it's relevant for. No need to block the options or AST/errors panes...

packages/website/src/components/playground/OptionsSelector.tsx
```diff
@@ -0,0 +1,147 @@
...
+        <ActionLabel name="Copy Markdown" onClick={copyMarkdownToClipboard}>
+          <Tooltip open={copyMarkdown} text="Copied">
+            <CopyIcon width="13.5" height="13.5" />
+          </Tooltip>
+        </ActionLabel>
+        <ActionLabel name="Report as Issue" onClick={openIssue}>
+          <IconExternalLink width="13.5" height="13.5" />
+        </ActionLabel>
+      </Expander>
+      {windowSize !== 'mobile' && (
```
JoshuaKGoldberg, line 109: Restricting some information on mobile viewports is an accessibility violation - some viewers have to zoom in a bunch and/or have unexpected viewport resolutions. And having dev tools open side-by-side often makes the playground snap to mobile...

```suggestion
```

packages/website/src/components/playground/OptionsSelector.tsx
```diff
@@ -0,0 +1,147 @@
...
+          </InputLabel>
+          <InputLabel name="monaco instance">
+            <code>window.monaco</code>
+          </InputLabel>
+          <InputLabel name="typescript">
+            <code>window.ts</code>
+          </InputLabel>
+          <InputLabel name="esquery">
+            <code>window.esquery</code>
+          </InputLabel>
```
JoshuaKGoldberg, line 122: Nit: it's been confusing my brain that these have `cursor: pointer` but don't do anything on click. Can that be removed? 😄

packages/website/src/components/layout/EditorTabs.tsx
```diff
@@ -0,0 +1,53 @@
...
+            key={item.label}
+            aria-selected={active === item.value}
+            disabled={active === item.value}
+            onClick={(): void => change(item.value)}
+          >
+            {item.label}
+          </button>
+        ))}
+      </div>
+      {(active === '.eslintrc' || active === 'tsconfig.json') && showModal && (
```
JoshuaKGoldberg, line 40: Is there ever a case where this check would be necessary? I tried removing it locally and couldn't repro any issues.

```suggestion
      {showModal && (
```

packages/website/src/components/editor/LoadedEditor.tsx
```diff
@@ -1,387 +1,159 @@
...
+  const onEditorDidMount = (
+    editor: Monaco.editor.IStandaloneCodeEditor,
+  ): void => {
+    editorRef.current = editor;
+
+    addLibFiles(system, monaco)
+      .then(() => {
+        window.esquery = utils.esquery;
+        // @ts-expect-error: TODO: remove me, this is only used for debugging
+        window.system = system;
```
JoshuaKGoldberg, line 119: I kinda like exposing these as globals, as a nice little dev helper for folks... 😄

JoshuaKGoldberg:
cc @jakebailey @orta - if either or both of you have time, this PR might be of interest to you. See in particular the _`packages/website/src/components/linter/bridge.ts`_ file for how we're using `@typescript/vfs`.

packages/website-eslint/types/index.d.ts
```diff
@@ -1,15 +1,16 @@
...
-import { astConverter } from '@typescript-eslint/typescript-estree/dist/ast-converter';
-import { getScriptKind } from '@typescript-eslint/typescript-estree/dist/create-program/getScriptKind';
+import type { analyze } from '@typescript-eslint/scope-manager/dist/analyze';
+import type { astConverter } from '@typescript-eslint/typescript-estree/dist/ast-converter';
 import esquery from 'esquery';
+import type { configs, rules } from '@typescript-eslint/eslint-plugin';
 
-export interface LintUtils {
-  createLinter: () => TSESLint.Linter;
+export interface EslintUtilsModule {
```
JoshuaKGoldberg, line 8: While we're here, I'm not sold on "utils" as being the part of the name. This is a pretty important chunk - and there are already multiple things called "utils" or similar. Not sure what an alternate name would be though... "bundled"? "packaged"? 🤔 🤷

JoshuaKGoldberg (commented):
Yeah @armano2 this is fantastic, really great stuff! It works smoothly and the new features (types, scopes, file type, ...) are awesome.

My biggest point of feedback would be that if you can split up the changes at all that would be really helpful for reviewing. It's a big one 😄. But if not I think it's a direct step in many positive directions from the site today, and it'd be also great to work from what's here.

LMK when it's ready for a full review! 🙌

packages/website/src/components/linter/bridge.ts
armano2: they are part of `ts.System`, they are not required to be implemented and mostly used by watch program, 

image

packages/website/src/components/playground/OptionsSelector.tsx
armano2: this block will be removed before we merge it, as i left it to ease debugging during review of changes, in mobile it make no sense to actually display those

packages/website/src/components/playground/OptionsSelector.tsx
armano2: this is only useful during dev, and shoudl be removed before merge

packages/website/src/components/config/ConfigTypeScript.tsx
armano2: yes, we don't want to update stuff in modal, modal is rendered on page all the time, and if we remvoe it, config will be updated on first render

packages/website/src/components/playground/PlaygroundRoot.tsx
armano2: agree, but this is oo scope for this change

packages/website/src/components/layout/EditorTabs.tsx
armano2: this condition hides Visual editor button when you are in `code` tab, 
image

bradzacher:
would it be possible to rebase this against the `v6` branch?
There are a bunch of changes we've got in v6 to the website tools - so we don't want to accidentally get into a bad state and break things if we merge this into `main` before `v6`.

packages/website/src/components/linter/bridge.ts
JoshuaKGoldberg: I guess I was just hoping for first party docs explaining this - @jakebailey's been looking at documenting more of the TS APIs.

packages/website/src/components/layout/EditorTabs.tsx
JoshuaKGoldberg: Ahhhh got it!

I'd suggest making this a prop the component takes in, like `hideVisualEditor` or `showVisualEditor`, which is set in the tabs config. Then this component doesn't have to know about the specifics of that quirk.

armano2:
@bradzacher changes rebased

packages/website/src/components/config/ConfigEslint.tsx
armano2: that was my general idea, to make it more universal and just use schema to generate visual editor, i will try to improve this in feature

JoshuaKGoldberg:
👋 ping @armano2, do you still have time / plans to work on this?

JoshuaKGoldberg:
Closing out this old draft PR to keep the queue small. 🧹

PR #6670, opened 2023-03-18
fix(typescript-estree): allow writing to deprecated node properties

### LINKED ISSUE #6668: typescript-estree: allow setting deprecated node properties
## Overview

Following https://github.com/typescript-eslint/typescript-eslint/pull/6525#issuecomment-1473392658: turns out Prettier has logic that reassigns deprecated node properties. We'll need to allow that. Otherwise, https://github.com/prettier/prettier/actions/runs/4445538911/jobs/7804778641?pr=14429#step:4:60:

```plaintext
  TypeError: Cannot set property typeParameters of #<Object> which has only a getter
      at visitNode (file:///home/runner/work/prettier/prettier/src/language-js/parse/postprocess/visit-node.js:20:19)
```

Thanks @fisker for the comments 😄

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6668

## Overview

In the node's `Object.defineProperty` _accessor descriptor_, it's now given a setter that sets it to a _data descriptor_. See terminology on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty.

### DIFF
--- a/packages/typescript-estree/src/convert.ts
+++ b/packages/typescript-estree/src/convert.ts
@@ -3327,9 +3327,9 @@ export class Converter {
         ? (): Properties[typeof valueKey] => node[valueKey]
         : (): Properties[typeof valueKey] => {
             if (!warned) {
-              // eslint-disable-next-line no-console
-              console.warn(
+              process.emitWarning(
                 `The '${aliasKey}' property is deprecated on ${node.type} nodes. Use '${valueKey}' instead. See https://typescript-eslint.io/linting/troubleshooting#the-key-property-is-deprecated-on-type-nodes-use-key-instead-warnings.`,
+                'DeprecationWarning',
               );
               warned = true;
             }
--- a/packages/typescript-estree/src/convert.ts
+++ b/packages/typescript-estree/src/convert.ts
@@ -3319,23 +3319,33 @@ export class Converter {
     aliasKey: AliasKey,
     valueKey: ValueKey,
   ): Properties & Record<AliasKey, Properties[ValueKey]> {
-    let errored = false;
+    if (this.options.suppressDeprecatedPropertyWarnings) {
+      (node as any)[aliasKey] = node[valueKey];
+      return node as Properties & Record<AliasKey, Properties[ValueKey]>;
+    }
+
+    let warned = false;
 
     Object.defineProperty(node, aliasKey, {
-      get: this.options.suppressDeprecatedPropertyWarnings
-        ? (): Properties[typeof valueKey] => node[valueKey]
-        : (): Properties[typeof valueKey] => {
-            if (!this.options.suppressDeprecatedPropertyWarnings) {
-              if (!errored) {
-                // eslint-disable-next-line no-console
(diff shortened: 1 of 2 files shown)


packages/typescript-estree/src/convert.ts
```diff
@@ -3319,23 +3319,33 @@ export class Converter {
...
+
+        return node[valueKey];
+      },
+      set(value): void {
+        Object.defineProperty(node, aliasKey, {
+          enumerable: true,
+          writable: true,
+          value,
+        });
+      },
```
JoshuaKGoldberg, line 3348: @fisker - does this set of changes work for you?

fisker: This should work for us.

JoshuaKGoldberg: Ah, hmm, setting `node[aliasKey] = node[valueKey]` does again make the node properties show up when enumerating - including in AST viewers and our snapshots. @fisker would it work to still always `Object.defineProperty`, with the additions of `configurable: true` and the setter? d3633bb

fisker: It should still work. 👍

fisker (approved):

JoshuaKGoldberg:
(converting to draft as I have to poke at the website build - I probably messed up something in conversion with `this` scope...)

packages/typescript-estree/src/convert.ts
JoshuaKGoldberg: Super, thanks!

packages/typescript-estree/src/convert.ts
```diff
@@ -3319,23 +3319,30 @@ export class Converter {
...
-              if (!errored) {
-                // eslint-disable-next-line no-console
-                console.warn(
-                  `The '${aliasKey}' property is deprecated on ${node.type} nodes. Use '${valueKey}' instead. See https://typescript-eslint.io/linting/troubleshooting#the-key-property-is-deprecated-on-type-nodes-use-key-instead-warnings.`,
-                );
-                errored = true;
-              }
+            if (!warned) {
+              // eslint-disable-next-line no-console
+              console.warn(
```
bradzacher, line 3331: should we use nodejs's infra here?
[`process.emitWarning`](https://nodejs.org/api/process.html#processemitwarningwarning-options)

eg
```ts
process.emitWarning(
  `The '${aliasKey}' property is deprecated on ${node.type} nodes. Use '${valueKey}' instead. See https://typescript-eslint.io/linting/troubleshooting#the-key-property-is-deprecated-on-type-nodes-use-key-instead-warnings.`
	'DeprecationWarning',
)
```

bradzacher (approved):

packages/typescript-estree/src/convert.ts
JoshuaKGoldberg: TIL, thanks!

PR #6723, opened 2023-03-21
fix: fix illegal decorator check

fisker:
## PR Checklist

## Overview

Copied from prettier https://github.com/prettier/prettier/blob/ddf3b43c33e2e98f6413b5232ad623876d96738e/src/language-js/parse/postprocess/typescript.js#L46


packages/typescript-estree/src/node-utils.ts
```diff
@@ -766,3 +766,41 @@ export function getContainingFunction(
...
+  firstAccessor?: ts.Node;
+  secondAccessor?: ts.Node;
+} {
+  // eslint-disable-next-line  @typescript-eslint/no-unsafe-return
+  return (
+    // @ts-expect-error -- internal api
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    ts.getAllAccessorDeclarations?.(accessor.parent.members ?? [], accessor) ??
+    {}
+  );
```
fisker, line 805: What should we do with all these internal apis?

packages/typescript-estree/src/convert.ts
```diff
@@ -3107,14 +3108,38 @@ export class Converter {
...
     if (nodeHasIllegalDecorators(node)) {
       this.#throwError(
         node.illegalDecorators[0],
         'Decorators are not valid here.',
       );
     }
 
-    for (const modifier of getModifiers(node) ?? []) {
+    // @ts-expect-error -- this is safe as it's guarded
+    const modifiers: ts.ModifierLike[] = node.modifiers;
```
bradzacher, line 3120: instead of using `.modifiers` can we use [`getModifiers`](https://github.com/typescript-eslint/typescript-eslint/blob/aa537e55137d854068bea7c9a1520c2763ea8da3/packages/typescript-estree/src/getModifiers.ts#L7-L31) here?

packages/typescript-estree/src/node-utils.ts
```diff
@@ -766,3 +766,18 @@ export function getContainingFunction(
...
   return ts.findAncestor(node.parent, ts.isFunctionLike);
 }
+
+// `ts.nodeCanBeDecorated`
+export function nodeCanBeDecorated(node: ts.Node): boolean {
+  return [true, false].some(
+    useLegacyDecorators =>
+      // @ts-expect-error -- internal api
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      ts.nodeCanBeDecorated?.(
```
bradzacher, line 776: instead of making this reference - can we add it to the TS types [in our ts augmentation](https://github.com/typescript-eslint/typescript-eslint/blob/3db653276b97d79f9b0c93b48f4fd0b62d58f694/packages/typescript-estree/typings/typescript.d.ts)

bradzacher, line 776: we can't use this function, or at least not without version checks.
in versions before 5.0 this function did not take `boolean` as the first argument - so by calling it like this we will cause a crash.

Given it's an internal API - I think we might want to consider rewriting it for ourselves? The logic will be pretty static given the spec doesn't change much if at all.

packages/typescript-estree/src/convert.ts
fisker: `getModifiers` only returns `ts.Modifier`, but we need `ts.Decorator`s, unless we can `getDecorators` again.

fisker: I know.

> unless we can getDecorators again.


But I checked both modifiers and decorators here, and `modifiers + decorators` is `node.modifiers`

bradzacher: I'm not quite sure I understand - your code below does "for each modifier, only action the decorators"

So couldn't the code be
```ts
for (const decorator of getDecorators(node)) {
  // your code
}

for (const modifier of getModifiers(node)) {
  // old code
}
```

packages/typescript-estree/src/node-utils.ts
bradzacher: `hasAmbientModifier` was only added in TS 4.3, and currently on v6 we are supporting 4.2
However based on the 2y lifecycle of versions that DefinitelyTyped supports (which we want to roughly align with) we will actually now be able to bump the minimum version to 4.3.

We can queue a separate PR for that then update this PR to use `ts.hasAmbientModifier`

packages/typescript-estree/src/convert.ts
fisker: Ideally, it should work, but I still can't get why you want 4 loops when you can do it in 1 loop.

Anyway, I tried, but turns out `getModifiers() + getDecorators()` not equals to `node.modifiers`.
[`ts.canHaveModifiers()`](https://github.com/typescript-eslint/typescript-eslint/blob/4bc75716681f2118c752df9360a1c5e97409ea93/packages/typescript-estree/src/getModifiers.ts#L16) and [`ts.canHaveDecorators()`](https://github.com/typescript-eslint/typescript-eslint/blob/4bc75716681f2118c752df9360a1c5e97409ea93/packages/typescript-estree/src/getModifiers.ts#L42) are stopping me to access modifiers on some node.

Should we
1. Remove those check
2. Add option to skip check

?

packages/typescript-estree/src/node-utils.ts
fisker: `ts.hasAmbientModifier` seems not available https://github.com/typescript-eslint/typescript-eslint/actions/runs/4739172024/jobs/8413765640?pr=6723#step:4:98

bradzacher: it's still a private, internal API!

packages/typescript-estree/src/convert.ts
bradzacher: The problem here is that they purposely removed `node.modifiers` from the types in TSv5 - so we definitely don't want to be accessing it directly if we can avoid it as it's clear that their intention is for it to be a truly private property.

I guess we could add an option to skip the checks?

packages/typescript-estree/src/node-utils.ts
fisker: Let's use a fake one for now?  https://github.com/typescript-eslint/typescript-eslint/pull/6723/commits/a84fa78fd3fa020ba32ae221025058ef0c128d11

fisker:
I don't know how `ts.hasAmbientModifier` works, and can't access `ts.hasAmbientModifier` directly, so I suggest use `() => false` for now. https://github.com/typescript-eslint/typescript-eslint/pull/6723#discussion_r1171029628

packages/typescript-estree/src/node-utils.ts
JoshuaKGoldberg: If we're going with a fake `() => false` one then I just removed it altogether. Seems like everything is working well without it? I posted a link in https://github.com/microsoft/TypeScript/issues/52727#issuecomment-1595567869.

JoshuaKGoldberg (approved):
Swell, thanks @fisker! I'll merge this now. 🚀 Thanks for the informative discussions!

PR #6724, opened 2023-03-21
fix(utils): accept null as default option in `applyDefault`

### LINKED ISSUE #6439: Bug: null as default option is failing in applyDefault

### Relevant Package

utils

### Playground Link

_No response_

### Repro Code

```typescript
/**
 * @file Rule to ensure that filenames match the exports of the file
 * @author Stefan Lau
 * @see https://github.com/selaux/eslint-plugin-filenames/blob/32fc70dd7572211d1e5b97e06ec7a005c77fe8d4/lib/rules/match-exported.js
 */

import path from 'node:path';
import { camelCase, kebabCase, upperFirst, snakeCase } from 'lodash';
import {
  createRule,
  isIgnoredFilename,
  getExportedName,
  isIndexFile,
  parseFilename,
} from '../utilities';

const transformMap = {
  camel: camelCase,
  kebab: kebabCase,
  pascal(name) {
    return upperFirst(camelCase(name));
  },
  snake: snakeCase,
};

const getStringToCheckAgainstExport = (parsed, replacePattern?: RegExp) => {
  const directoryArray = parsed.dir.split(path.sep);
  const lastDirectory = directoryArray[directoryArray.length - 1];

  if (isIndexFile(parsed)) {
    return lastDirectory;
  }

  return replacePattern ? parsed.name.replace(replacePattern, '') : parsed.name;
};

const getTransformsFromOptions = (option) => {
  const usedTransforms = option?.push ? option : [option];

  return usedTransforms.map((name) => {
    return transformMap[name];
  });
};

const transform = (exportedName, transforms) => {
  return transforms.map((format) => {
    return format ? format(exportedName) : exportedName;
  });
};

const anyMatch = (expectedExport, transformedNames) => {
  return transformedNames.includes(expectedExport);
};

const getWhatToMatchMessage = (transforms) => {
  if (transforms.length === 1 && !transforms[0]) {
    return 'the exported name';
  }

  if (transforms.length > 1) {
    return 'any of the exported and transformed names';
  }

  return 'the exported and transformed name';
};

type Options = [string[] | string | null, string | null, boolean];

type MessageIds = 'indexFile' | 'regularFile';

export default createRule<Options, MessageIds>({
  create: (context, options) => {
    return {
      Program(node) {
        const transforms = getTransformsFromOptions(options[0]);
        const replacePattern = options[1]
          ? new RegExp(options[1], 'u')
          : undefined;
        const filename = context.getFilename();
        const absoluteFilename = path.resolve(filename);
        const parsed = parseFilename(absoluteFilename);
        const shouldIgnore = isIgnoredFilename(filename);
        const exportedName = getExportedName(node, options[2]);
        const isExporting = Boolean(exportedName);
        const expectedExport = getStringToCheckAgainstExport(
          parsed,
          replacePattern,
        );
        const transformedNames = transform(exportedName, transforms);
        const everythingIsIndex =
          exportedName === 'index' && parsed.name === 'index';
        const matchesExported =
          anyMatch(expectedExport, transformedNames) || everythingIsIndex;

        if (shouldIgnore || !isExporting || matchesExported) {
          return;
        }

        context.report({
          data: {
            expectedExport,
            exportName: transformedNames.join("', '"),
            extension: parsed.ext,
            name: parsed.base,
            whatToMatch: getWhatToMatchMessage(transforms),
          },
          messageId: isIndexFile(parsed) ? 'indexFile' : 'regularFile',
          node,
        });
      },
    };
  },
  defaultOptions: [null, null, false],
  meta: {
    docs: {
      description:
        'Match the file name against the default exported value in the module.',
      recommended: 'warn',
    },
    messages: {
      indexFile:
        "The directory '{{expectedExport}}' must be named '{{exportName}}', after the exported value of its index file.",
      regularFile:
        "Filename '{{expectedExport}}' must match {{whatToMatch}} '{{exportName}}'.",
    },
    schema: [
      {
        oneOf: [
          {
            items: {
              type: 'string',
            },
            type: 'array',
          },
          {
            type: 'string',
          },
          {
            type: 'null',
          },
        ],
      },
      {
        oneOf: [
          {
            type: 'string',
          },
          {
            type: 'null',
          },
        ],
      },
      {
        type: 'boolean',
      },
    ],
    type: 'suggestion',
  },
  name: 'filename-match-exported',
});
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

When invoked with options `[null, null, true],`, this rule is failing with a rather generic error:

```
TypeError: Error while loading rule 'filename-match-exported': Cannot convert undefined or null to object
Occurred while linting /some/dir/foo.js
      at Function.keys (<anonymous>)
      at deepMerge (node_modules/.pnpm/@typescript-eslint+utils@5.51.0_4vsywjlpuriuw3tl5oq6zy5a64/node_modules/@typescript-eslint/utils/src/eslint-utils/deepMerge.ts:26:31)
...
```

After digging into the stack trace, I found that the error originates in `applyDefault` function, and the actual error is:

```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at deepMerge (/Users/gajus/Documents/dev/gajus/eslint-plugin-canonical/node_modules/.pnpm/@typescript-eslint+utils@5.51.0_4vsywjlpuriuw3tl5oq6zy5a64/node_modules/@typescript-eslint/utils/src/eslint-utils/deepMerge.ts:26:31)
```

### Actual Result

Expected user options to override defaults.

### Additional Info

_No response_

### Versions

N/A

Solo-steven:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6439 

## Overview

fix issue #6439


bradzacher (changes requested):
I guess the question here is whether or not an explicit `null` should be ignored like this and overridden by the default, or if it should be seen as a value that shouldn't be overridden.

I think an explicit `null` should not be overridden, personally.

Solo-steven:
@bradzacher  Thanks for your reply, I think another solution is adding one more conditional to `isObjectNotArray` like
```ts
function isObjectNotArray<T extends ObjectLike>(
  obj: unknown | unknown[],
): obj is T {
  return typeof obj === 'object' && obj != null && !Array.isArray(obj);
}
```
this solution wouldn't overwrite the null value .

But I want to know why we shouldn't overwrite the value, is it cause more errors?

bradzacher:
> But I want to know why we shouldn't overwrite the value, is it cause more errors?

If a user does a config like `{ foo: null }`, and the default is `{ foo: { bar: '' } }`, then we should not override the explicit `null` with `{ bar: '' }`.
The same can be said for a top-level `null`. If the user configures a rule like `['error', null]`, then we should not override that explicit `null` with whatever the default is.

Instead I think that we should only override `undefined`, as we currently do for non-top-level values.

packages/utils/src/eslint-utils/applyDefault.ts
```diff
@@ -25,7 +25,6 @@ function applyDefault<TUser extends readonly unknown[], TDefault extends TUser>(
   (options as unknown[]).forEach((opt: unknown, i: number) => {
     if (userOptions[i] !== undefined) {
       const userOpt = userOptions[i];
-
```
JoshuaKGoldberg, line 28: [Nitpick] Unrelated change, let's take it out. No need to increase the diff size.

JoshuaKGoldberg (approved):
Nice, thanks for this!

JoshuaKGoldberg:
👋 @bradzacher I think this got dropped from the queue because I did a silly. Do you want to review this one?

It looks good to me, but just making sure.

JoshuaKGoldberg (dismissed):
💯 Thanks! And, sorry for the long wait!

bradzacher (approved):

PR #6728, opened 2023-03-21
chore(website): [playground] add copy as json and simplify ast viewer

### LINKED ISSUE #6037: Website: [Playground] Add "copy as JSON" option to AST Viewer

armano2:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6037

## Overview

migrate ast viewer changes from #6656

this change is required to implement type viewer

notable changes:
- improved accessibility
- code is simplified and prepared to display other json like structures
- show additional flags (languageVersion, languageVariant, scriptKind, transformFlags)
  image
- iterable elements are now properly displayed
  image
- copy as json is now available
  image


packages/website/src/hooks/useClipboard.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { useDebouncedToggle } from './useDebouncedToggle';
+
+export type useClipboardResult = [copied: boolean, copy: () => void];
+
+export function useClipboard(code: () => string): useClipboardResult {
+  const [copied, setCopied] = useDebouncedToggle<boolean>(false);
+
+  const copy = useCallback(() => {
+    void navigator.clipboard.writeText(code()).then(() => {
```
JoshuaKGoldberg, line 11: [Edge case] Aha, this is why I don't like `void` for `no-floating-promises` - it masks real issues!

Sometimes browsers fail to copy to the clipboard. Most of those times are from users not giving them permission; once in a while it's an OS issue. Either way it'd be good to give them an indication that it failed.

I think it's fine to file this as a followup issue - but it'd be nice to have in this PR.

packages/website/src/components/inputs/Tooltip.tsx
```diff
@@ -4,18 +4,20 @@ import React from 'react';
 import styles from './Tooltip.module.css';
 
 export interface TooltipProps {
-  readonly children: JSX.Element | (JSX.Element | false)[];
+  readonly children: React.ReactNode;
   readonly text: string;
   readonly position?: 'left' | 'right';
   readonly open?: boolean;
   readonly hover?: boolean;
+  readonly clasName?: string;
```
JoshuaKGoldberg, line 12: [Typo]

```suggestion
  readonly className?: string;
```

Aside: I wonder how many KB the internet would be smaller by if this was how it was spelled...

packages/website/src/hooks/useDebouncedToggle.ts
```diff
@@ -1,6 +1,6 @@
 import { useCallback, useRef, useState } from 'react';
 
-export default function useDebouncedToggle<T>(
+export function useDebouncedToggle<T>(
   value: T,
   timeout = 1000,
```
JoshuaKGoldberg, line 5: [Accessibility] Ah, I missed this before, sorry - it's an accessibility issue to have parts of the page that exist for only a limited time. If someone isn't able to read the _Copied_ message in time, such as from a reading disability or missing it on the page, they could miss what it said and not know whether the copy was successful.

Since this is existing code I'm still happy to ✔️ and file a followup issue - just raising an FYI.

FWIW one remediation I've liked in the past is to make the copy button a ✔️ until the value that would be copied is changed. The _Copied_ text can then disappear as long as there's _some_ indication of the (success) status. 

https://www.w3.org/WAI/WCAG21/Understanding/enough-time

packages/website/src/components/inputs/CopyButton.module.css
```diff
@@ -0,0 +1,62 @@
...
+.copyButtonTooltip::before {
+  margin-right: 7px;
+}
+
+.copyButtonTooltip::after {
+  min-height: 25px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  right: 23px !important;
```
JoshuaKGoldberg, line 57: This doesn't seem to do anything on my machine? Also, avoiding `!important` would be nice.

```suggestion
```

Btw, at this point I feel like maybe it'd make sense for us to go with a UI library for the tooltip & such, instead of rolling it outselves? There are accessibility considerations that I don't fully understand. Maybe a good followup?

packages/website/src/components/ast/Elements.tsx
```diff
@@ -1,109 +1,167 @@
...
+      return {
+        type: type,
+        group: 'simple',
+        tooltip: getTooltipLabel(type, value, propName, parentNodeType),
+      };
+    }
+  }, [value, propName, getTypeName, getTooltipLabel, parentNodeType]);
+
+  useEffect(() => {
+    const shouldOpen = !!selectedPath && selectedPath.startsWith(level);
```
JoshuaKGoldberg, line 101: [Nitpick]

```suggestion
    const shouldOpen = !!selectedPath?.startsWith(level);
```

packages/website/src/hooks/useClipboard.ts
```diff
@@ -0,0 +1,17 @@
+import { useCallback } from 'react';
+
+import { useDebouncedToggle } from './useDebouncedToggle';
+
+export type useClipboardResult = [copied: boolean, copy: () => void];
+
+export function useClipboard(code: () => string): useClipboardResult {
+  const [copied, setCopied] = useDebouncedToggle<boolean>(false);
```
JoshuaKGoldberg, line 8: [Nitpick]

```suggestion
  const [copied, setCopied] = useDebouncedToggle(false);
```

packages/website/src/components/editor/LoadedEditor.tsx
```diff
@@ -159,7 +159,9 @@ export const LoadedEditor: React.FC<LoadedEditorProps> = ({
       onEsASTChange(webLinter.storedAST);
       onTsASTChange(webLinter.storedTsAST);
       onScopeChange(webLinter.storedScope);
-      onSelect(sandboxInstance.editor.getPosition());
+
+      const position = sandboxInstance.editor.getPosition();
+      onSelect(position ? tabs.code.getOffsetAt(position) : undefined);
```
JoshuaKGoldberg, line 164: [Question] What's the difference between calling `onSelect` with the position vs. the offset position? I can't figure it out locally.

JoshuaKGoldberg (changes requested):
I started reviewing but then couldn't get the demo to work locally - is this just me? 😞 

Screenshot of the playground's AST pane showing only 'undefined'

packages/website/src/components/ast/tsUtils.ts
```diff
@@ -0,0 +1,83 @@
...
+
+declare global {
+  interface Window {
+    tsEnum: TsParsedEnums;
+  }
+}
+
+export function getTsEnum(type: keyof TsParsedEnums): Record<number, string> {
+  if (!window.tsEnum) {
+    window.tsEnum = {
```
JoshuaKGoldberg, line 39: [Nitpick]

```suggestion
  window.tsEnum ??= {
```

packages/website/src/hooks/useClipboard.ts
armano2: i don't think we have a way to display and error's like that, unless we add some sort of toast, i can catch and `console.error` but there will be no difference

packages/website/src/hooks/useDebouncedToggle.ts
armano2: this is oo, scope for this change, we should do follow-up to improve accessibility

packages/website/src/components/inputs/CopyButton.module.css
armano2: image
image

going with proper UI library will be nice, but this is oo scope for this change

packages/website/src/components/editor/LoadedEditor.tsx
armano2: this was done to simplify logic used to highlight elements `getPosition` returns class that contains lint/column and functions to mutate/validate those values

conversion between this structure and what we need previously was happening during comparison for every node, (this was actually more complex but lets leave at that), 

this changes that we are always sending `number` as cursor position within text and receiving [number, number] as position of selected node

packages/website/src/components/ast/tsUtils.ts
armano2: i actually don't like this being exposed in window, i'm going to change that

packages/website/src/components/inputs/CopyButton.module.css
armano2: i refactored this code, that we should no longer need it

armano2:
> I started reviewing but then couldn't get the demo to work locally - is this just me? 😞

not sure about what issue you have during local dev, but there was actual issue with undefined, i missed one thing while moving this change to standalone PR

armano2:
i did some profiling on update tree, and I refactored code a little to be a little faster, i also solved issue with re-renders when you change tabs

JoshuaKGoldberg (changes requested):
Looks wonderful, thanks! I'm real stoked about this one in particular for the AST copying 😄 

Requesting changes on a few small touchups - but nothing I'm particularly passionate about.

packages/website/src/components/Playground.tsx
```diff
@@ -126,7 +115,7 @@ function Playground(): JSX.Element {
             />
           </div>
           <ConditionalSplitPane
-            render={enableSplitPanes}
+            render={windowSize !== 'mobile'}
```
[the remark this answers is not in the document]
armano2: sure, but this is oo scope for this change, idea is to actually make split screen work on mobile devices in feature :)

packages/website/src/components/ast/tsUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+    TokenFlags: extractEnum(window.ts.TokenFlags),
+    ModifierFlags: extractEnum(window.ts.ModifierFlags),
+    ObjectFlags: extractEnum(window.ts.ObjectFlags),
+    SymbolFlags: extractEnum(window.ts.SymbolFlags),
+    FlowFlags: extractEnum(window.ts.FlowFlags),
+    TypeFlags: extractEnum(window.ts.TypeFlags),
+    ScriptKind: extractEnum(window.ts.ScriptKind),
+    ScriptTarget: extractEnum(window.ts.ScriptTarget),
+    LanguageVariant: extractEnum(window.ts.LanguageVariant),
+    // @ts-expect-error: non public API
```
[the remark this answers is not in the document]
armano2: this is intentionally a public API, and normally ts type dosn't expose properties that have this type.

packages/website/src/components/ast/tsUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+    TransformFlags: extractEnum(window.ts.TransformFlags),
+  };
+  return tsEnumCache[type];
+}
+
+export function tsEnumValue(
```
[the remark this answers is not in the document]
armano2: what do you think about:

```
tsEnumValue -> tsEnumToString
expandFlags -> tsEnumFlagToString
```

packages/website/src/components/Playground.tsx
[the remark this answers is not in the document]
JoshuaKGoldberg: Is it? I'm just suggesting moving the prop into the component, not changing it's behavior

packages/website/src/components/ast/tsUtils.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Yes! +1

packages/website/src/components/ast/tsUtils.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Yeah I'm thinking if it's a API that we care about & would want to show, it'd make sense to ask for it to be turned public.

packages/website/src/components/ast/PropertyValue.tsx
```diff
@@ -1,34 +1,92 @@
...
 
 export interface PropertyValueProps {
-  readonly value: ASTViewerModelMap;
+  readonly value: unknown;
+}
+
+export interface SimpleModel {
+  readonly value: string;
+  readonly className: string;
+  shortValue?: string;
```
JoshuaKGoldberg, line 14: [Question] Is this one not able to be `readonly`?

JoshuaKGoldberg (approved):
Yay!

packages/website/src/components/ast/PropertyValue.tsx
armano2: correct, 

its muted in

```
  const model = useMemo(() => {
    const val = getSimpleModel(value);
    if (val.value.length > 250) {
      val.shortValue = val.value.substring(0, 200);
    }
    return val;
  }, [value]);
```

armano2: i applied this suggestion :)

PR #6732, opened 2023-03-22
feat: support parse hashbang

### LINKED ISSUE #6500: Enhancement: Support Hashbang parse in `typescript-estree`

### Relevant Package

typescript-estree

### Description

I understand that ESLint core handles Hashbang automatically, but it's kind of missing feature when use `typescript-estree` alone.

Should we add support for it? It's in the spec after all.

### Fail

```typescript
N/A
```

### Pass

```typescript
N/A
```

### Additional Info

```js
> require('@typescript-eslint/typescript-estree').parse('#!/usr/bin/env node', {comment: true})
{ type: 'Program', body: [], sourceType: 'script', comments: [] } // <- missing comment
```

fisker:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6500


bradzacher (commented):
So the issue with adding support like this is that we now have an odd dichotomy.

ESLint doesn't let a parser parse hashbang comments - it replaces `#!` with `//` and then updates the first comment from `type: Line` to `type: Shebang`.

By adding parsing support and a `type: Hashbang` comment we're actually moving away from what ESLint currently does and introducing some weirdness - the types will say `HashBangComment`, but that's not a physically possible state and instead you'll get `ShebangComment`.

I wonder if we should wait to move on this properly until ESLint honours parser support for it?

JoshuaKGoldberg:
@bradzacher just following up on this, what should we do?

> I wonder if we should wait to move on this properly until ESLint honours parser support for it?

Is there a tracking issue we should mark this as blocked on?

bradzacher:
I assume eslint v9 would probably have this? I don't know if there is a tracking issue.

We'll need conditional support at some point. Doing it with types is... Difficult though.

JoshuaKGoldberg:

bradzacher:
So yeah for starters this needs to be behind an option because ESLint doesn't currently support the hashbang node.

So the issue with that is getting the types right.
For example all consumers of `typescript-estree` should get the hashbang comment if and only if they turn the option on - so the types also need to reflect this - which involves conditional types and is ugly (we already have some conditional types for options and it's sucky).

Then we also need to make sure we hide the type from the `utils` package so that ESLint consumers don't see it ever - we wouldn't want people accidentally using it and writing useless code.

So for now - I think we put this on hold and wait until ESLint releases support. Once ESLint releases support we can turn it on by default and only turn it off on older ESLint versions. That way we're okay with it being in the types everywhere and don't need to worry about the conditional side of things.

JoshuaKGoldberg:
I'm going to put this as a draft so it doesn't keep coming up on my review queue. Please yell at me if that's disruptive to you. 🙂

JoshuaKGoldberg:
This has been sitting for a while. So I'm going to close it out to keep our queue small, and mark the linked issue as blocked on ESLint.

PR #6735, opened 2023-03-22
chore(website): [playground] add tabs to ast viewer and update design

armano2:
## PR Checklist

## Overview

migrate tabs changes from #6656

notable changes:
- right panel uses tabs instead of dropdown in options
- design has been updated to use styles from button/button-group

## design

old:

image
image

new:

image
image


packages/website/src/components/layout/EditorTabs.tsx
```diff
@@ -0,0 +1,55 @@
...
+import React from 'react';
+
+import styles from './EditorTabs.module.css';
+
+export interface EditorTabsProps<T extends string | boolean> {
+  readonly tabs: ({ value: T; label: string } | T)[];
+  readonly active: T;
+  readonly showVisualEditor?: boolean;
+  readonly change: (name: T) => void;
+  readonly showModal?: (name: T) => void;
```
JoshuaKGoldberg, line 12: Out of curiosity, is there any benefit to marking props as `readonly`? It feels like extra effort - maybe we should use a lint rule against modifying props instead?

packages/website/src/components/config.ts
```diff
@@ -0,0 +1,6 @@
+export const detailTabs = [
+  { value: false as const, label: 'Errors' },
```
JoshuaKGoldberg, line 2: [Nitpick] [Style] Maybe extract out a type for each element and make this whole variable an array of that?

```suggestion
export const detailTabs: DetailTab[] = [
  { value: false, label: 'Errors' },
```

(more posting out of curiosity for your style preferences than anything else, definitely nowhere near a blocker)

JoshuaKGoldberg (approved):
Awesome! 🔥

packages/website/src/components/layout/EditorTabs.tsx
armano2: > maybe we should use a lint rule against modifying props instead?

that's what readonly is for, adding rule that covers buildin behavior is bad

there is actually a rule to mark props as readonly (for flow) https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/prefer-read-only-props.md

packages/website/src/components/config.ts
armano2: generally if type is never going to be reused I prefer to use `as const` in this context, and if type is or may be reused I'm creating types

PR #6738, opened 2023-03-22
chore(website): [playground] improve design of error viewer

armano2:
## PR Checklist

## Overview

migrate error viewer changes from #6656

notable changes:
- alerts have consistent small spacing between them
- text is printed as monospace font
- text is wrapped

image image

### DIFF
--- a/packages/website/src/components/ErrorsViewer.tsx
+++ b/packages/website/src/components/ErrorsViewer.tsx
@@ -5,6 +5,8 @@ import type Monaco from 'monaco-editor';
 import React, { useEffect, useState } from 'react';
 
 import styles from './ErrorsViewer.module.css';
+import type { AlertBlockProps } from './layout/AlertBlock';
+import AlertBlock from './layout/AlertBlock';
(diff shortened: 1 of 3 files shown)


packages/website/src/components/ErrorsViewer.tsx
```diff
@@ -95,7 +83,7 @@ function ErrorBlock({
         <div>
           {item.suggestions.map((fixer, index) => (
             <div
-              key={index}
+              key={String(index)}
```
JoshuaKGoldberg, line 86: [Nitpick] Keys can be `number`s, any reason to make this change? (here and the other one)

```suggestion
              key={index}
```

JoshuaKGoldberg (approved):
I like it!

packages/website/src/components/ErrorsViewer.tsx
armano2: hmm, did something changed in react 18?

old doc recommended to use string indexes https://legacy.reactjs.org/docs/lists-and-keys.html#keys

PR #6742, opened 2023-03-22
feat(eslint-plugin): [no-unnecessary-condition] add `allowAssertTypePredicateParameterConditions` option

### LINKED ISSUE #2752: [no-unnecessary-condition] Add option to ignore expressions as an arguments of certain functions.
We working on converting `graphql` and `express-graphql` package to TS and want to use `no-unnecessary-condition` since we had this lint rule in Flow.
The problem is that both libraries are used in a lot of JS projects so we need to check arguments of public API functions for popular mistakes.
So we created `devCheck` function for this purpose:
```ts
function devAssert(condition: unknown, message: string): asserts condition {
  const booleanCondition = Boolean(condition);
  if (!booleanCondition) {
    throw new Error(message);
  }
}
```

Problem is that we get this error:
```
  190:13  error  Unnecessary conditional, the types have no overlap  @typescript-eslint/no-unnecessary-condition
```

for this code

```ts
export function graphqlHTTP(options: Options): Middleware {
  devAssert(options != null, 'GraphQL middleware requires options.');
```

It would be ideal if we can pass `devAssert` as an option for the `no-unnecessary-condition` rule and expressions inside arguments would be ignored.

kozlovvski:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2752

## Overview

The new option makes it possible to turn the rule off for parameters in type predicates that use `asserts` keyword.

Example of code that doesn't trigger the rule with this option enabled:

```ts
function assert(condition: unknown): asserts condition {
  // do something with condition
}
const obj = {};
assert(obj != null);
```

**How it works**

For `BinaryExpression` nodes, there is a new condition which will check the parent type (provided that the new `allowAssertTypePredicateParameterConditions` option is turned on.

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md
@@ -83,6 +83,18 @@ for (; true; ) {}
 do {} while (true);
 ```
 
+### `allowAssertTypePredicateParameterConditions`
+
+Example of correct code for `{ allowAssertTypePredicateParameterConditions: true }`:
+
+```ts
+function assert(condition: unknown): asserts condition {
+  // do something with condition
+}
+const obj = {};
+assert(obj != null);
+```
+
 ### `allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing`
 
 If this is set to `false`, then the rule will error on every file whose `tsconfig.json` does _not_ have the `strictNullChecks` compiler option (or `strict`) set to `true`.
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -24,6 +24,7 @@ import {
   nullThrows,
   NullThrowsReasons,
 } from '../util';
+import { isFunctionArgument } from '../util/explicitReturnTypeUtils';
 
 // Truthiness utilities
 // #region
@@ -64,6 +65,7 @@ const isLiteral = (type: ts.Type): boolean =>
 export type Options = [
   {
     allowConstantLoopConditions?: boolean;
+    allowAssertTypePredicateParameterConditions?: boolean;
     allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing?: boolean;
   },
 ];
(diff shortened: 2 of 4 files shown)


packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
```diff
@@ -573,6 +573,23 @@ function get<Obj, Key extends keyof Obj>(obj: Obj, key: Key) {
...
+    {
+      code: `
+        function assert(condition: unknown): asserts condition {
+          // do something with condition
+        }
+        const obj = {};
+        assert(obj != null);
+      `,
+      options: [{ allowAssertTypePredicateParameterConditions: true }],
+    },
```
JoshuaKGoldberg, line 592: Requesting one more `valid` test case: `assert(obj)`. That kind of assertion comes up once in a while.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -294,6 +303,32 @@ export default createRule<Options, MessageId>({
...
+        !valueDeclaration ||
+        !ts.isFunctionDeclaration(valueDeclaration) ||
+        !valueDeclaration.type
+      ) {
+        return false;
+      }
+
+      return (
+        ts.isTypePredicateNode(valueDeclaration.type) &&
+        valueDeclaration.type.assertsModifier != null
```
JoshuaKGoldberg, line 328: If I remove this second check the unit tests still pass. Is there a case where `assertsModifier` is ever not defined? 

* If there is, then let's add a test for it
* If not, let's remove this line (and maybe file an issue somewhere about the odd type?)

FWIW I can't figure out how to make it not-defined...

packages/eslint-plugin/docs/rules/no-unnecessary-condition.md
```diff
@@ -83,6 +83,18 @@ for (; true; ) {}
 do {} while (true);
 ```
 
+### `allowAssertTypePredicateParameterConditions`
```
JoshuaKGoldberg, line 86: A lot of developers don't know what type predicates are. Could you add a line or two explaining them please?

JoshuaKGoldberg (changes requested):
Looks great for the most part! Just requesting a few small touchups. Thanks! 🚀

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
kozlovvski: > Is there a case where assertsModifier is ever not defined 

Honestly I don't remember at this point – I will look into this and write some detailed explanation, okay?

packages/eslint-plugin/docs/rules/no-unnecessary-condition.md
kozlovvski: Yeah, definitely!

kozlovvski:
Hey, I am closing this for now – I have trouble finding time recently and I don't want to keep a stale PR open.

PR #6749, opened 2023-03-23
feat(eslint-plugin): [consistent-type-assertions] add `arrayLiteralTypeAssertions` option

### LINKED ISSUE #6374: Enhancement: [consistent-type-assertions] Prefer type annotation over assertion for arrays

### Description

Code should not use an `as` type assertion when a `:` type annotation would do. In some cases (see examples later) they act the same, but `as` assertions are generally something that should be avoided whenever possible. `as` removes some type safety from TypeScript. Not ideal.

### Fail Cases

```typescript
const errorMessages = [] as string[];
```

### Pass Cases

```typescript
const errorMessages: string[] = [];
```

### Additional Info

This should also trigger for arrays that have contents in them.

```diff
- const errorMessages = ['a'] as string[];
+ const errorMessages: string[] = ['a'];
```

Alas, knowing whether the assertion actually was doing anything important is blocked on my favorite TypeScript issue https://github.com/microsoft/TypeScript/issues/9879. But since folks really shouldn't be using an assertion for a variable with an initial value, I'm ok with this edge case bug being in the rule:

```diff
// The + is now an error... 😈 deal with it!
- const errorMessages2 = [Math.random() ? 'a' : 'b'] as 'a'[];
+ const errorMessages3: 'a'[] = [Math.random() ? 'a' : 'b'];
```

danvk:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6374

## Overview

This PR adds an `arrayLiteralTypeAssertions` option to the `consistent-type-assertions` rule as described in the linked issue. This uses the same auto-fixer as the `objectLiteralTypeAssertions` option.

I didn't include an equivalent of the `"allow-as-parameter"` option but I can do so if there's a need.

If I may step on my soapbox for a moment:

1. The `objectLiteralTypeAssertions` option (and this new one) are an awkward fit for the `consistent-type-assertions`, which you would expect to just enforce consistency between `x as T` and `<T>x`. Both options would make more sense in a distinct rule that bans type assertions (rather than enforcing a consistent style).
2. I'm not sure why you'd ever want to enable `objectLiteralTypeAssertions` but not `arrayLiteralTypeAssertions`, or vice versa. It also feels odd that we're singling out these two types but allowing constructs with other types such as `const x = 'foo' as string;`

### DIFF
--- a/packages/eslint-plugin/src/rules/consistent-type-assertions.ts
+++ b/packages/eslint-plugin/src/rules/consistent-type-assertions.ts
@@ -9,12 +9,14 @@ type MessageIds =
   | 'angle-bracket'
   | 'never'
   | 'unexpectedObjectTypeAssertion'
+  | 'unexpectedArrayTypeAssertion'
   | 'replaceObjectTypeAssertionWithAnnotation'
   | 'replaceObjectTypeAssertionWithSatisfies';
 type OptUnion =
   | {
       assertionStyle: 'as' | 'angle-bracket';
       objectLiteralTypeAssertions?: 'allow' | 'allow-as-parameter' | 'never';
+      arrayLiteralTypeAssertions?: 'allow' | 'never';
     }
   | {
       assertionStyle: 'never';
@@ -36,6 +38,7 @@ export default util.createRule<Options, MessageIds>({
       'angle-bracket': "Use '<{{cast}}>' instead of 'as {{cast}}'.",
       never: 'Do not use any type assertions.',
       unexpectedObjectTypeAssertion: 'Always prefer const x: T = { ... }.',
+      unexpectedArrayTypeAssertion: 'Always prefer const x: T[] = [ ... ].',
       replaceObjectTypeAssertionWithAnnotation:
         'Use const x: {{cast}} = { ... } instead.',
       replaceObjectTypeAssertionWithSatisfies:
@@ -63,6 +66,9 @@ export default util.createRule<Options, MessageIds>({
               objectLiteralTypeAssertions: {
                 enum: ['allow', 'allow-as-parameter', 'never'],
               },
+              arrayLiteralTypeAssertions: {
+                enum: ['allow', 'never'],
+              },
             },
             additionalProperties: false,
             required: ['assertionStyle'],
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/consistent-type-assertions.ts
```diff
@@ -9,12 +9,16 @@ type MessageIds =
...
   | 'replaceObjectTypeAssertionWithAnnotation'
-  | 'replaceObjectTypeAssertionWithSatisfies';
+  | 'replaceObjectTypeAssertionWithSatisfies'
+  | 'replaceArrayTypeAssertionWithAnnotation'
+  | 'replaceArrayTypeAssertionWithSatisfies';
 type OptUnion =
   | {
       assertionStyle: 'as' | 'angle-bracket';
       objectLiteralTypeAssertions?: 'allow' | 'allow-as-parameter' | 'never';
+      arrayLiteralTypeAssertions?: 'allow' | 'never';
```
JoshuaKGoldberg, line 21: > I didn't include an equivalent of the `"allow-as-parameter"` option but I can do so if there's a need.

Yeah it feels weird to have one without the other... Could you please?

JoshuaKGoldberg (changes requested):
Sorry this took so long to get to @danvk - the implementation looks very clean and the tests are thorough. Thanks for sending it in!

Just requesting changes on the missing `"allow-as-parameter"` equivalent.

JoshuaKGoldberg:
> If I may step on my soapbox for a moment

You most certainly may 😄 I agree with your points. It does seem odd to be so specific, especially since it means the exception doesn't cover `'foo' as string` & co. Could I ask you to file a new issue with your proposed changes? It might end up being a breaking change, so we might have to wait until v7 to get it in.

JoshuaKGoldberg:
👋 Hey @danvk! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

JoshuaKGoldberg:
Closing this PR as it's been stale for a few months without activity. Feel free to reopen @danvk if you have time - but no worries if not! If anybody wants to drive it forward, please do post your own PR - and if you use this as a start, consider adding `Co-authored-by: @danvk` at the end of your PR description. Thanks! 😊

Issue #5857, opened 2022-10-20
Refactor: Deprecate createDefaultProgram/shouldCreateDefaultProgram


JoshuaKGoldberg:
## Overview

Back in the day, if a user didn't provide a `parserOptions.project`, `@typescript-eslint/parser` would create one behind-the-scenes for them. This was convenient but came at severe performance cost: the parser would recompute a bunch of information _for each file added_.

We wish to remove the `createDefaultProgram` option. But, there may be some users who still rely on it. See #3627.

For now, our plan will be:

1. In the upcoming major version, rename it to `deprecated__createDefaultProgram`
2. Remove it in the _next_ major version

JoshuaKGoldberg:
#5890 is merged into #5886; closing this issue now accordingly. It'll be released as a part of the v6 major release.

PR #6754, opened 2023-03-24
feat(typescript-estree): add EXPERIMENTAL_useProjectService option to use TypeScript project service

### LINKED ISSUE #6575: Enhancement: Try using tsserverlibrary's ProjectService to handle type-aware mode

### Relevant Package

typescript-estree

### Description

This is a follow-up to #6172 and the TypeScript / ESLint meeting we had last week.

Currently, type-aware mode uses `ts.Program` to do its work. There are some problems with this approach:

- This mode is not references aware, leading to the need to either put every file into one program, or manually specify all programs (https://typescript-eslint.io/linting/typed-linting/monorepos).
- With multiple programs in play, lots of work can be duplicated as projects can reference each other.
- This project has to manage much of the lifecycle of Programs, and do a lot of patching to the program to make things work (e.g. not emit, make sure not to exit, etc). This seems to lead to known memory leaks and other problems.
- Probably more I'm forgetting. I don't think I need list more 😄 

#6172 takes the approach of using the LanguageService to handle this stuff. This is a good step, and one taken by projects like ts-morph. However, a single LanguageService itself will only handle one project at a time, so is only a partial fix.

My idea is to go further, and use `tsserverlibrary`'s `ProjectService` instead. This is "the thing" that tsserver uses to load projects and respond to language server requests for the editor. The idea is that you can create a LanguageService, then act like a user does, opening files, making requests, closing files, and so on. This means:

- Project references are handled "for free".
- Users do not need to specify _any_ tsconfig, just like how you don't specify anything when opening VS Code.
- The lifecycle of programs and related language serivces is fully managed, hopefully fixing leaks.
- Work _shouldn't_ be duplicated, or at least not duplicated any more than anyone already experiences using TypeScript in their editors.

The general approach is as follows:

- Import `tsserverlibrary` instead of `typescript`.
- Create a `ProjectService`. Like everything else, it takes in a host, but nothing surprising.
- To look at a file, call `openClientFile` to mark the file as open (potentially providing contents), and then call `getScriptInfo` on its path.
- Drill down from `ScriptInfo` like `getScriptInfo(filename).getDefaultProject().getLanguageService(/*ensureSynchronized*/ true)` and so on until you get to Program / TypeChecker whatever else is needed.
- Eventually, close the file.

There are some problems:

- `tsserverlibrary` is an entirely different module than `typescript`. Will this break downstream rules which import `typescript` directly? Is that a thing people do? Do we have to move the API? *(uh oh)*
- In CLI mode, this all seems fine, as the service does all of the work. But in the editor, is this too heavy given `tsserver` is already running? maybe it's not a problem becuase it's no worse than now.
- If people are running one of the "parallel eslint" wrappers, will this cause resource usage problems? Or, again, is this no worse than the current state of things?

### Fail

```typescript
N/A
```

### Pass

```typescript
N/A
```

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6575

## Overview

Creates an experimental ready-to-try-only-if-you-dare prototype implementation of #6575: using TypeScript's "project service" / server to create programs instead of our existing manual project creation. Doing so brings the _"create program(s) for source files"_ part of typed linting towards how editors behave:

* Using the nearest `tsconfig.json` for each file (similar to `parserOptions.project: true`)
* Using a default ("inferred") `tsconfig.json` if one is not found applicable to the file
    * This would overcome #6826 with no additional user configuration, hooray!

```js
// .eslintrc.js
module.exports = {
  extends: [
    'eslint:recommended',
    'plugin:@typescript-eslint/recommended-type-checked',
  ],
  plugins: ['@typescript-eslint'],
  parserOptions: {
    EXPERIMENTAL_useProjectService: true,
  },
  root: true,
};
```

### Technical Details

In essence, this PR:

* Adds an `EXPERIMENTAL_useProjectService` option to typescript-estree
    * Setting `process.env.TYPESCRIPT_ESLINT_EXPERIMENTAL_TSSERVER` can also enable the option
* Adjusts `createParserSettings` to call to a new `createProjectService` function lazily when the experimental option is enabled (thereby setting `parseSettings.EXPERIMENTAL_projectService`)
* Adjusts `getProgramAndAST` to use that project service if it exists, with a new `useProgramFromProjectService`
* Massages internal tests to always use absolute (or failing that, relative to project root) paths for files if possible, as the project service expects that _(is this phrasing accurate?)_
* Skips tests that aren't applicable anymore when the project service is in use

See https://github.com/JoshuaKGoldberg/repros/tree/cbe9d9c9b19c09b085a0be2024221b2d05753cc8 for a reproduction case showing this working.

### Complete Project Linting

```shell
for branchName in v6 create-project-service; do
    git checkout $branchName
    yarn
    hyperfine "yarn lint --skip-nx-cache" > $branchName.txt
done 

hyperfine "TYPESCRIPT_ESLINT_EXPERIMENTAL_TSSERVER=true yarn lint --skip-nx-cache" > enabled.txt
```

<table>
  <thead>
    <tr>
      <th><code>v6</code></th>
      <th>#6754 baseline</th>
      <th>#6754 enabled</th>
      <th>#6754 Δ</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td><code>62.976 s ±  0.479 s</code></td>
      <td><code>65.383 s ±  0.226 s</code></td>
      <td><code>31.169 s ±  0.401 s</code></td>
      <td>-52% 🟢</td>
    </tr>
  </tbody>
</table>

### Incremental Project Linting

```shell
for glob in "eslint-plugin/src/rules/*.ts" "typescript-estree/src/create-program/*.ts" "website/**/*.tsx"; do
    for setting in false true; do
        hyperfine "TYPESCRIPT_ESLINT_EXPERIMENTAL_TSSERVER=$setting npx eslint --config .eslintrc.js packages/$glob"
    done
done
```

<table>
  <thead>
    <tr>
      <th>Glob</th>
      <td>Baseline</td>
      <td>Enabled</td>
      <td>Δ</td>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th><code>eslint-plugin/src/rules/*.ts</code></th>
      <td><code>8.552 s ± 0.100 s</code></td>
      <td><code>9.525 s ± 0.084 s</code></td>
      <td>+11% ❌</td>
    </tr>
    <tr>
      <th><code>packages/typescript-estree/src/create-program/*.ts</code></th>
      <td><code>7.162 s ± 0.079 s</code></td>
      <td><code>2.153 s ± 0.007 s</code></td>
      <td>-70% 🟢</td>
    </tr>
    <tr>
      <th><code>packages/website/**/*.tsx</code></th>
      <td><code>9.955 s ± 0.080 s</code></td>
      <td><code>3.714 s ± 0.029 s</code></td>
      <td>-63% 🟢</td>
    </tr>
  </tbody>
</table>


packages/typescript-estree/src/parser.ts
```diff
@@ -57,8 +59,43 @@ function getProgramAndAST(
...
+    // TODO "Eventually, close the file"
+    // what does this actually do?
+    //
+
+    if (program) {
+      const projectServiceProgram = createProjectProgram(parseSettings, [
+        // LOL TYPES
+        // This'll hoepfully be resolved once they fully move to modules
+        // right now it's two separate files that get dumped out
+        program as ts.Program,
```
JoshuaKGoldberg, line 86: > `tsserverlibrary` is an entirely different module than `typescript`.

Yeah, `ts.Program is not assignable to ts.Program` here. Perhaps this change should be blocked on TypeScript fully moving to ESM and having one version of each class/enum/etc.?

packages/typescript-estree/src/parser.ts
```diff
@@ -284,3 +321,16 @@ export {
   clearProgramCache,
   clearParseAndGenerateServicesCalls,
 };
+
+// next steps
+// 0. clean up comments
+// 1. ping jake with questions
```
JoshuaKGoldberg, line 327: (I'm doing these in reverse, heh)

packages/typescript-estree/src/parser.ts
jakebailey: The problem is that you basically need to be using tsserverlibrary the whole time, because it really is a differnet bundle.

As much as I want ESM to be "the thing", I don't know if it helps you all given you can't have ESM for eslint and you can't use async either?

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -0,0 +1,70 @@
+import * as ts from 'typescript';
+import * as tsserver from 'typescript/lib/tsserverlibrary';
```
jakebailey, line 2: Can tell you at a glance that this worries me; you you effectively always need to use exactly one of these imports.

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -0,0 +1,70 @@
...
+  const watchDirectory = (
+    path: string,
+    callback: ts.DirectoryWatcherCallback,
+  ) => {
+    // todo (or just ... stub out?)
+    return { close() {} };
+  };
+
+  const system = {
+    ...ts.sys,
```
jakebailey, line 29: Shocked this works; this is from an entirely different module. I guess it's an interface anyway so it's safeish.

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -0,0 +1,70 @@
...
+      loggingEnabled: () => false,
+      perftrc() {},
+      info() {},
+      startGroup() {},
+      endGroup() {},
+      msg() {},
+      getLogFileName: () => undefined,
+    },
+    session: undefined,
+  });
```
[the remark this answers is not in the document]
jakebailey: I feel like some of these could be optional, though I think we may instead want to acutally provide ATA because that's a thing that is useful for people writing javascript. Otherwise, type-using rules will not actually actually get types.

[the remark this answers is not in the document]
jakebailey: As for defaults, I think what you've written is acutally fine.

packages/typescript-estree/src/parser.ts
```diff
@@ -284,3 +321,16 @@ export {
...
+//   * what does closing a file do
+// 2. create versions of test suite tailored to this new fancy form
+// 3. testing the bujeezus out of it
+
+// aside:
+// * performance testing - need to work on that
```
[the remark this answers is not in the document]
jakebailey: That feels super wrong; where is all that time going? It surely doesn't take 30 seconds to get results out of tsserver in VS Code, right?

packages/typescript-estree/src/parser.ts
```diff
@@ -57,8 +59,43 @@ function getProgramAndAST(
...
+    // todo: maybe we'll slap this into the parseSEttings as a property?
+    // const projectService = createProjectServiceProgram(/* parseSettings */);
+    const program = parseSettings.projectService
+      .getScriptInfo(parseSettings.filePath)!
+      .getDefaultProject()
+      .getLanguageService(/*ensureSynchronized*/ true)
+      .getProgram();
+
+    // TODO "Eventually, close the file"
+    // what does this actually do?
```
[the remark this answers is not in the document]
jakebailey: Truthfully I don't know; if I answered it'd be because I just read the source and figured it out 😅 

@sheetalkamat very likely knows more.

packages/typescript-estree/src/parser.ts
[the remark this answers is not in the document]
armano2: there is actually few problems with this POC, on every file we are recreating projectService

```ts
    // todo: sometimes
    projectService: createProjectService(),
```

from what i understund here, we should make it persistant

packages/typescript-estree/src/parser.ts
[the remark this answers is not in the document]
sheetalkamat: Closing a file will make the project for open file as something that can be removed from memory. It isnt removed right away but on next file open if the project is not needed by new open file.

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  const projectService = new tsserver.server.ProjectService({
+    host: system,
+
+    // todo: look into these
+    cancellationToken: { isCancellationRequested: () => false },
+    useSingleInferredProject: false, // ???
+    useInferredProjectPerProjectRoot: false, // ???
+    typingsInstaller: {
```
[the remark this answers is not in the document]
JoshuaKGoldberg: 👍 https://github.com/microsoft/TypeScript/issues/53803

packages/typescript-estree/src/create-program/createProjectService.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: > provide ATA because that's a thing that is useful for people writing javascript

AH, we've been bit in the past by trying to do things for users. Even ignoring performance penalties (e.g. #5857's explanation), folks find it confusing for types to be different in different tools.

packages/typescript-estree/src/parser.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Ok, this makes me think it's ok that we're likely not able to explicitly close files then. Thanks!

packages/typescript-estree/src/create-program/createProjectService.ts
JoshuaKGoldberg: Heh, it'd only been used in this file for `...ts.sys` and some types. Switched to always `typescript/lib/tsserverlibrary`. But in other files we're still `from 'typescript'`. Even if were were to completely switch over all our files, community plugins would still want to be able to import from `typescript`. 🤔

packages/typescript-estree/src/parser.ts
JoshuaKGoldberg: Hmm, that might take a while: https://github.com/eslint/eslint/issues/15475

packages/typescript-estree/src/parser.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Ah, indeed. Now it's down to 3-4 seconds for `no-floating-promises` on its own. I'll post a performance comparison in the OP.

packages/typescript-estree/src/create-program/createProjectService.ts
jakebailey: It's tough, and I'm not sure what the answer is without a big shift.

If you were doing a plugin model like TS, you'd provide the TS API via a property on `context`, that way the version of TS that is being used by any given rule is exactly the one that is being used by eslint, which _sounds_ correct already, and would let you pass in the `tsseverlibrary` object too.

packages/typescript-estree/src/create-program/createProjectService.ts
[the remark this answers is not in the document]
jakebailey: I sent https://github.com/microsoft/TypeScript/pull/53896 for this, but theoretically you could just do an `undefined as any` for now if you wanted to skip implementing this bit? But, what you have is equivalent to the `noopTypingsInstaller` so is probably fine.

packages/typescript-estree/src/parser.ts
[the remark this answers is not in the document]
jakebailey: Did you have a chance to get some numbers? I'm definitely interested.

packages/typescript-estree/src/useProgramFromProjectService.ts
```diff
@@ -0,0 +1,34 @@
...
+  projectService.setCompilerOptionsForInferredProjects({
+    rootDir: parseSettings.tsconfigRootDir,
+  });
+
+  projectService.openClientFile(
+    parseSettings.filePath,
+    parseSettings.codeFullText,
+    /* scriptKind */ undefined,
+    parseSettings.tsconfigRootDir,
+  );
```
JoshuaKGoldberg, line 21: We're seeing unit test failures like `Cannot read properties of undefined (reading 'fileName')` on `sourceFile.fileName`:

```js
// (code simplified from node_modules/tslint/lib/linter.js)
var sourceFile = this.program.getSourceFile(fileName);

// ...

if (error_1.isError(error) && error.stack !== undefined) {
    error_1.showRuleCrashWarning(error.stack, rule.getOptions().ruleName, sourceFile.fileName);
}
```

That `sourceFile` is `undefined` because the project (`this.program`) created by the project service has a different `currentDirectory` than our `parseSettings.tsconfigRootDir`:

```diff
- /Users/josh/repos/typescript-eslint/packages/eslint-plugin-tslint
+ /Users/josh/repos/typescript-eslint/packages/eslint-plugin-tslint/tests/fixtures/fixture-project
```

It comes from `ConfiguredProject`'s constructor providing `currentDirectory` as `getDirectoryPath(configFileName)`:

https://github.com/microsoft/TypeScript/blob/df3bec66748f48badad61faa3fb4f9a79a11a612/src/server/project.ts#L2600

@bradzacher @jakebailey I'm not sure whether we're holding this wrong. I _think_ we would want to be able to resolve the client file using its path relative to typescript-eslint's configured root dir is, not relative to its specific project, no?

jakebailey: I would think that if using ProjectService, you should totally ignore any and all other configuration knobs in ts-eslint for the project; they're not going to be used at all by tsserver. The paths you give it should be in whatever format tsserver requires (absolute? normalized? some combo?).

packages/typescript-estree/src/parser.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Sorry, not yet - definitely this week! (conferences and other committments)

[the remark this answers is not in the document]
JoshuaKGoldberg: Updated the OP. tl;dr: it's a little bit faster, maybe 5% - but not earth-shattering.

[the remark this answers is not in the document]
jakebailey: I'm not totally surprised. Is there any way to test it in the editor or some sort of watch mode, to show that long-lived processes are more responsive?

I'd also be interested in profiling it (may test it tomorrow) to make sure it's not duplicating work.

Also, if someone with a project which previously had issues with tsconfig configuration, monorepo setup, etc, can try it (since theoretically this PR makes all of the options surrounding tsconfig no longer required; it "just works").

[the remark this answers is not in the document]
bradzacher: @JoshuaKGoldberg the more interesting comparison would be subset linting rather than complete project linting (which is what the editor does essentially).

So pick one subfolder in a package and run a lint - because it only calculates a minimal subset of type info it should be faster than a classic run.

packages/typescript-estree/src/create-program/createProjectService.ts
JoshuaKGoldberg: Interestingly, we're already moving towards wrapping TypeScript APIs with our own services (e.g. https://github.com/typescript-eslint/typescript-eslint/pull/6404). So that might coincidentally make many rules out-of-the-box. Though I worry about edge cases around folks importing from `typescript` themselves...

packages/typescript-estree/src/parser.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Updated the OP - I tried a few globs and didn't see any real changes. Maybe I'm running it wrong?

packages/typescript-estree/src/create-program/createOrUpdateProjectService.ts
```diff
@@ -0,0 +1,94 @@
...
+          rootFiles: [
+            {
+              content: codeFullText,
+              fileName: filePath,
+            },
+          ],
+        });
+      }
+    }
+  }
```
JoshuaKGoldberg, line 43: When `project` is set to something other than `./tsconfig.json`, the project service doesn't seem to natively understand to ask for it. You can see this in unit tests such as `no-unsafe-call.test.ts` (any `RuleTester` with a `parserOptions.project` like `./tsconfig.*.json`). [[Example CI failure](https://github.com/typescript-eslint/typescript-eslint/actions/runs/4800211437/jobs/8540946706)]

```plaintext
    Message:
      A fatal parsing error occurred: Parsing error: Cannot read file '/home/runner/work/typescript-eslint/typescript-eslint/packages/eslint-plugin/tests/fixtures/tsconfig.noImplicitThis.json'
```

It feels very weird to specifically indicate that a single file name & content should match for the project. But calling `openExternalProject` caused a significant slowdown in unit tests in `eslint-plugin`: `time TYPESCRIPT_ESLINT_EXPERIMENTAL_TSSERVER=true yarn jest naming-convention` went from ~90s to ~120s on my Mac Studio.

Is there a better strategy for what we're trying to do?

packages/eslint-plugin-tslint/src/rules/config.ts
```diff
@@ -118,7 +119,7 @@ export default createRule<Options, MessageIds>({
     context,
     [{ rules: tslintRules, rulesDirectory: tslintRulesDirectory, lintFile }],
   ) {
-    const fileName = context.getFilename();
+    const fileName = path.resolve(context.getCwd(), context.getFilename());
```
JoshuaKGoldberg, line 122: Explaining: with `EXPERIMENTAL_ProjectService`, relative paths might no longer match up with how the project views its root file. So switching to absolute paths is more reliable.

packages/typescript-estree/src/create-program/createOrUpdateProjectService.ts
jakebailey: My impression is still that there's no reason to respect `project` set to any path when run with the project service, because it's not going to be respectd in VS Code either. Adding the project service means that no configuration is needed at all for any tsconfigs, besides one toggle to say whether TS is loaded or not.

Am I missing something?

JoshuaKGoldberg: We do need to respect some compiler options in our lint rules though. Some lint rules behave differently when e.g. `noImplicitThis` or `strictNullChecks` are enabled. `/Users/josh/repos/typescript-eslint/packages/eslint-plugin/tests/rules/non-nullable-type-assertion-style.test.ts` shows tests that are specifically set up for `noUncheckedIndexedAccess`.

Separately, if someone is using a `tsconfig.eslint.json` to set up different `paths` or other significant changes, we need to respect that in our rules. One common strategy is to have a `tsconfig.eslint.json` that specifies `allowJs: true` to typed rules are allowed to run on `.eslintrc.js` & other config files.

jakebailey: > We do need to respect some compiler options in our lint rules though. Some lint rules behave differently when e.g. `noImplicitThis` or `strictNullChecks` are enabled. `/Users/josh/repos/typescript-eslint/packages/eslint-plugin/tests/rules/non-nullable-type-assertion-style.test.ts` shows tests that are specifically set up for `noUncheckedIndexedAccess`.

Is this just exclusive for your testing?

> Separately, if someone is using a `tsconfig.eslint.json` to set up different `paths` or other significant changes, we need to respect that in our rules. One common strategy is to have a `tsconfig.eslint.json` that specifies `allowJs: true` to typed rules are allowed to run on `.eslintrc.js` & other config files.

The former is the thing that using the project service solves, though; you don't need to trick TS into loading the code into one project anymore.

Loading js, yeah, not sure there.

JoshuaKGoldberg: > Is this just exclusive for your testing?

The testing is reflective of real users. We've seen cases where people set varying TSConfig options in different parts of their projects -sometimes in a formalized monorepo, sometimes in a large vaguely-organized single repo- and want that respected in their lint rules.

> add some sort of config for the "implicit" project used for files that aren't in any other project?

Interesting - like a default project that always exists, provided by our tooling? Based on `tsconfig.json`? But what if the user doesn't have a `tsconfig.json` visible (e.g. is linting from a subdirectory)?

jakebailey: > The testing is reflective of real users. We've seen cases where people set varying TSConfig options in different parts of their projects -sometimes in a formalized monorepo, sometimes in a large vaguely-organized single repo- and want that respected in their lint rules.

Right, but if they have that, then the tsconfigs on disk should already reflect that setup, right? So they can typecheck and get tsserver working?

> Interesting - like a default project that always exists, provided by our tooling? Based on `tsconfig.json`? But what if the user doesn't have a `tsconfig.json` visible (e.g. is linting from a subdirectory)?

So the implicit project is what you're using when you open `.eslintrc.js` in a project with a tsconfig that doesn't include it. All code has to be in some project in order to analyze it. So, I would expect that with the ProjectService, you can open any file and get results.

Whether or not the defualt settings are good, i'm not sure. VS Code configures it differently: https://github.com/microsoft/vscode/blob/2d28a86e1d8cb556d06c1f646ca7370ed761e5b0/extensions/typescript-language-features/src/tsconfig.ts#L30

JoshuaKGoldberg: We talked on Discord - I'll post a summary soon. tl;dr is - I'm going to prioritize getting this integration simple & shippable, perhaps to the point of excluding / ignoring these unit tests.

JoshuaKGoldberg: Ok! Summarizing our chat:
* I (Josh) have been trying to come up with repro cases showing why users would want to have their own `project`(s) (`tsconfig.eslint.json`, `tsconfig.differentCompilerOption`, ...) - and have yet to come up with a broadly compelling reason 
    * `allowJs: true` is given to us for free by virtue of the default inferred TS project (see the vscode link 👆)
    * Even if we had a great use case, it would be an _unusual_ and _weird_ one: because it would mean people are running ESLint in ways different from how the editor views files!
* The vast majority of users who would want behavior akin to today's `project: true` would likely be happy with `useProjectService`. 
    * If we do want to use an API for adding manual non-inferred projects (e.g. `openExternalProject`) then we can always add it in later - it's not necessary for an MVP
* My (Josh) next steps will be to:
    1. Remove the `openExternalProject` usage
    2. Try to work around it in unit tests, and fall back to skipping tests that can't
    3. Un-draft this PR for review
    4. _(hopeful)_ get this PR release & available
    5. _(hopeful)_ update docs to mention this experimental project service

packages/eslint-plugin/tests/rules/consistent-type-imports.test.ts
```diff
@@ -14,6 +14,7 @@ const ruleTester = new RuleTester({
 });
 
 const withMetaParserOptions = {
+  EXPERIMENTAL_useProjectService: false,
```
JoshuaKGoldberg, line 17: The project service only supports TSConfigs with names like `./tsconfig.json`. So in testing these rules, we explicitly indicate _not_ to use it despite CI potentially setting `process.env.TYPESCRIPT_ESLINT_EXPERIMENTAL_TSSERVER`.

packages/typescript-estree/src/parseSettings/index.ts
```diff
@@ -57,6 +58,13 @@ export interface MutableParseSettings {
    */
   errorOnUnknownASTType: boolean;
 
+  /**
+   * Experimental: TypeScript server to power program creation.
+   */
+  EXPERIMENTAL_projectService:
+    | tsserverlibrary.server.ProjectService
```
jakebailey, line 65: Will including this have a negative effect for those who are importing this, since now two big TS d.ts files will get loaded? Maybe not too many use this, but, I'm just thinking ahead.

packages/typescript-estree/src/useProgramFromProjectService.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const configFilePath = program.getCompilerOptions().configFilePath as string;
+  if (
+    !parseSettings.projects
+      .map(absolutify)
+      .map(getCanonicalFileName)
+      .includes(getCanonicalFileName(configFilePath))
+  ) {
+    throw new Error(`Cannot read file '${parseSettings.projects[0]}'`);
+  }
```
jakebailey, line 44: Is this bit needed? If I'm understanding this correctly, this still forces users to specify tsconfigs?

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -0,0 +1,55 @@
...
+  // TODO: see getWatchProgramsForProjects
+  // We don't watch the disk, we just refer to these when ESLint calls us
+  // there's a whole separate update pass in maybeInvalidateProgram at the bottom of getWatchProgramsForProjects
+  // (this "goes nuclear on TypeScript")
+  const system: tsserver.server.ServerHost = {
+    ...tsserver.sys,
+    clearImmediate,
+    clearTimeout,
+    setImmediate,
+    setTimeout,
```
jakebailey, line 19: These are already provided by `ts.sys`, technically; tsserver actually just does `sys as ServerHost`, which is unsafe, but, seemingly okay if you want to write less here.

packages/typescript-estree/src/parseSettings/index.ts
JoshuaKGoldberg: Hmm, I hope not! So far it hasn't hurt anything I've noticed.

packages/typescript-estree/src/useProgramFromProjectService.ts
JoshuaKGoldberg: We have a trio of tests that fail without it:
* `parseAndGenerateServices › projectFolderIgnoreList › ignores a folder when given a string glob`
* `semanticInfo › non-existent project file`
* `semanticInfo › fail to read project file`

...but I think you're right, with the project service is managing projects I don't see a need to preserve that behavior. I'll exclude them as well.

packages/typescript-estree/src/create-program/createProjectService.ts
JoshuaKGoldberg: Hmm, that assertion does make me nervous. Do you think it'd be reasonable to indicate that `ts.sys` is a `System` with those properties filled out?

jakebailey: It's tricky because outside of node, those may not actually exist. It's fine to leave it as-is, I think.

.github/workflows/ci.yml
```diff
@@ -91,10 +91,31 @@ jobs:
...
-
-      - name: Run Check
+      - name: Run Lint
         run: yarn ${{ matrix.lint-task }}
 
+  lint_with_build_from_root:
+    name: Lint From Root
+    # because we lint with our own tooling, we need to build
+    needs: [build]
+    runs-on: macos-latest
```
JoshuaKGoldberg, line 101: > `macos-latest`

https://github.com/typescript-eslint/typescript-eslint/actions/runs/4825613286/jobs/8596705424?pr=6754

```plaintext
Run TYPESCRIPT_ESLINT_EXPERIMENTAL_TSSERVER=true npx eslint .

<--- Last few GCs --->

[233[8](https://github.com/typescript-eslint/typescript-eslint/actions/runs/4825613286/jobs/8596705424?pr=6754#step:5:9):0x5[9](https://github.com/typescript-eslint/typescript-eslint/actions/runs/4825613286/jobs/8596705424?pr=6754#step:5:10)76d80]    72536 ms: Mark-sweep 2020.0 (2088.4) -> 2011.9 (2080.8) MB, 1230.0 / 0.0 ms  (average mu = 0.618, current mu = 0.305) allocation failure; scavenge might not succeed
[2338:0x5976d80]    72619 ms: Scavenge 2021.0 (2081.3) -> 2018.5 (2081.3) MB, 17.1 / 0.0 ms  (average mu = 0.618, current mu = 0.305) allocation failure; 
[2338:0x5976d80]    73028 ms: Scavenge 2021.3 (2081.6) -> 2019.3 (2[10](https://github.com/typescript-eslint/typescript-eslint/actions/runs/4825613286/jobs/8596705424?pr=6754#step:5:11)2.6) MB, 372.5 / 0.0 ms  (average mu = 0.618, current mu = 0.305) allocation failure; 


<--- JS stacktrace --->

 1: 0xb7a940 node::Abort() [node]
```

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

bradzacher:
I think I need to point out that we can't truly use this flag in production until TS deduplicates the packages so types etc are shared between them.

The reason we've never supplied the `typescript` instance via parser services is because there are so many tools/utils which aren't built to accept it as an argument - they just import `typescript` directly and use it that way. Not to mention any 3rd party rules do exactly the same thing right now.

I realise that there shouldn't actually be any differences between the two packages - but there is a small chance there are that could cause "undefined" behaviour for users.

jakebailey:
I talked about this a little with Josh, and I actually think it may not be quite as bad as first thought. My thinking is this:

There are only two types of mismatches that can occur:

- Differing versions of typescript in `node_modules`.
- Differing behaviors between `typescript.js` and `tsserverlibrary.js` on the same version of TypeScript.

The former is already a risk factor right now because `ts` isn't a part of the context; it's effectively required that the version that a library loads is the same as what eslint is going to run.

Given the above is already possible, the only new observable difference is between the two bundles in the same version of TypeScript. But, unless we start introducing hidden state via WeakMaps or something very, very odd:

- All of the enums will contain the same values, so be interchangable.
- Anything to do with type checking is actually going to be properties of `TypeChecker`, `Node`, `Symbol`, and therefore you're not really going to be able to use the "wrong" side's values.

So honestly, I think things might "just work".

But, I think that I'm interested in seeing whether or not we could push ProjectService into `typescript.js`. But, @sheetalkamat pointed out after I proposed this that we could feasibly at that point just move everything into `typescript.js` and make `tsserverlibrary.js` reexport from it, which also interesting and actually quite a savings overall. But, I haven't looked into the impact of that besides `typescript.js` taking 10% longer to load, but the whole package becoming `8.11 MB` smaller.

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -0,0 +1,55 @@
+import * as tsserver from 'typescript/lib/tsserverlibrary';
```
jakebailey, line 1: Is there any way that you can use a lazy import here, e.g. move this into a require in `createProjectService`? This will end up increasing the startup time of the plugin by another 140ms or so just from loading TS twice.

JoshuaKGoldberg: Love it! Applying now, then will re-measure performance.

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -0,0 +1,59 @@
+import type * as tsserverType from 'typescript/lib/tsserverlibrary';
```
jakebailey, line 1: FWIW you can probably just call this `ts`, unless you're trying to avoid potential confusion.

packages/typescript-estree/src/parseSettings/createParseSettings.ts
```diff
@@ -47,6 +51,13 @@ export function createParseSettings(
         : new Set(),
     errorOnTypeScriptSyntacticAndSemanticIssues: false,
     errorOnUnknownASTType: options.errorOnUnknownASTType === true,
+    EXPERIMENTAL_projectService:
+      (options.EXPERIMENTAL_useProjectService === true &&
+        process.env.TYPESCRIPT_ESLINT_EXPERIMENTAL_TSSERVER !== 'false') ||
+      (process.env.TYPESCRIPT_ESLINT_EXPERIMENTAL_TSSERVER === 'true' &&
+        options.EXPERIMENTAL_useProjectService !== false)
+        ? (TSSERVER_PROJECT_SERVICE ??= createProjectService())
+        : undefined,
```
bradzacher, line 60: lol our logic for all of this is getting pretty complicated.
maybe we want to convert all these flags into an enum or something to simplify from `n` boolean flags to a more understandable representation?

packages/typescript-estree/src/parseSettings/index.ts
bradzacher: This will actually leak out into the public types accidentally.
`index` has `export { createProgramFromConfigFile as createProgram } from './create-program/useProvidedPrograms';`, and `./create-program/useProvidedPrograms` has `declare function useProvidedPrograms(programInstances: Iterable<ts.Program>, parseSettings: ParseSettings): ASTAndProgram | undefined;` - i.e. `import type { ParseSettings } from '../parseSettings'`

To properly isolate this Josh we'll probably want to refactor `create-program/useProvidedPrograms` into two files so that we don't import it from the `index` - i.e. just have `createProgramFromConfigFile` in its own file so that we don't leak the internal types.

I was poking around via `unpkg` but might be worth just double checking all of the paths with intellisense from a local build to double check that is the only place it leaks into `index`'s dependency graph.

packages/typescript-estree/src/useProgramFromProjectService.ts
```diff
@@ -0,0 +1,40 @@
...
+    parseSettings.tsconfigRootDir,
+  );
+  if (!opened.configFileName) {
+    return undefined;
+  }
+
+  const scriptInfo = projectService.getScriptInfo(parseSettings.filePath);
+  const program = projectService
+    .getDefaultProjectForFile(scriptInfo!.fileName, true)!
+    .getLanguageService(/*ensureSynchronized*/ true)
```
bradzacher, line 26: the TS code style is bleeding in here :P

bradzacher (approved):
as I mentioned in discord - I'm happy to get this in so people can start testing this at home - it all looks pretty solid as far as I can tell.

If @jakebailey is happy then i'm happy.

packages/eslint-plugin/tests/rules/consistent-type-imports.test.ts
JoshuaKGoldberg: Noting - I almost have this working locally I think, just finding issues with the project program not finding the file after I update the root dir...

packages/type-utils/tests/TypeOrValueSpecifier.test.ts
```diff
@@ -198,7 +198,7 @@ describe('TypeOrValueSpecifier', () => {
           'type Foo = {prop: string}; type Test = Foo;',
         ].map((test): [string, TypeOrValueSpecifier] => [
           test,
-          { from: 'file', name: 'Foo', path: 'tests/fixtures/file.ts' },
+          { from: 'file', name: 'Foo', path: 'file.ts' },
```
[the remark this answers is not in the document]
JoshuaKGoldberg: The tests now are based on `./tests/fixtures/` as the cwd - instead of `./`. The path is relative to the cwd.

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -0,0 +1,58 @@
...
+      attach: (): void => {},
+      enqueueInstallTypingsRequest: (): void => {},
+      installPackage: (): Promise<never> => {
+        throw new Error('This should never be called.');
+      },
+      isKnownTypesPackageName: () => false,
+      onProjectClosed: (): void => {},
+      globalTypingsCacheLocation: '',
+    },
+    logger: {
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Ooh good point. I'll leave that as a followup issue 👍 it'd be a good fodder for getting someone other than me involved in this area of code.

packages/typescript-estree/src/parseSettings/createParseSettings.ts
JoshuaKGoldberg: Yeah agreed, this is  ... a lot. I think it'd be good to go even further and have some kind of discriminated union for the area of project settings. That way we can be more clear on e.g. not allowing `project: './tsconfig.eslint.json'` with `useProjectService` (until/unless that becomes supported in TypeScript APIs). A good followup to investigate.

packages/typescript-estree/src/parseSettings/index.ts
JoshuaKGoldberg: I see what you mean - as a quicker / less change-y step, I just changed the parameters for these functions to be their own types (or just a `string`). How does that feel to you?

JoshuaKGoldberg:
Re-requesting review from @bradzacher because I applied a couple of small touchups & answered a question. But otherwise I feel comfortable merging this after we get v6 out the door (just to be safe & not conflict with the v6 release). Woop woop! 🚀

packages/typescript-estree/src/parseSettings/index.ts
bradzacher: that fixes the issue, yeah because we no longer leak `ParseSettings` into a file that's imported via the `index` - meaning consumers won't ever implicitly load the tssserverlibrary defs.

bradzacher (dismissed):

packages/typescript-estree/src/parseSettings/index.ts
jakebailey: This file still contains the import from `tsserverlibrary`; from the above it seemed like it was able to be eliminated. Is that still the case?

JoshuaKGoldberg: We're just using it here for `EXPERIMENTAL_projectService: tsserverlibrary.server.ProjectService | undefined;`, and it's a `type` import. I'd think this is enough to fix the issues around importing both TypeScript files? Unless we're talking about different things?

jakebailey: I was actually originally referring to the type checking, in that uses of your API will type check TS's huge API twice.

ritschwumm:
may i ask what happened with this PR - is it merged now, or had it to be abandoned?

JoshuaKGoldberg:
Now that `v6` is merged into `main`, I'm planning on merging this as well some time in the next 1-2 weeks. Depending on whether there are any major issues reported with `6.0.0`:

* If there are, we'll fix them in next week's release (July 17th) and I'll plan to merge this this in time for the following week (July 24th)
* If there aren't, I'll plan to merge this in time for next week's release (July 17th)

Note that we're keeping the _`EXPERIMENTAL`_ on for now. We're going to want to think a bit more deeply on how the `parserOptions` type shape should look.

FYI @jakebailey - we're moving quickly this month! ⚡

packages/typescript-estree/src/parseSettings/index.ts
jakebailey: Is there any alternative that would prevent this from being visible externally, or is this definitely a requirement?

JoshuaKGoldberg: 🤔 long-term I think we really want this to be visible externally. I want to investigate integrating with other project service consumers such as https://github.com/Quramy/typescript-eslint-language-service (cc @Quramy).

I don't think I follow what you mean by users of our API type checking the huge API twice - do you mean because if this file's `.d.ts` output is in the `node_modules/` of projects, TypeScript (e.g. via `tsc`) will type check it?

If so: I suppose in theory we could hand-write a `ProjectService` type that coincidentally matches the `tsserverlibrary.server.ProjectService` type, with comments explaining why. Does that feel reasonable to you?

jakebailey: If someone imports `@typescript-eslint/estree`, they will then access the `d.ts` files for this, which then references both `typescript` and `tsserverlibrary`, and since they're not the same file and both large, it'll increase the check time of anyone who imports this package.

Probably I should revisit the "merge typescript and tsserverlibrary" idea, I had just been working on other stuff.

JoshuaKGoldberg: Ah, got it thanks. I think this is fine for now. I'm under the impression that very few consumers import from any `@typescript-eslint/*` package - and it's mostly authors of custom rules. Good to note as a potential followup if someone reports anything.

JoshuaKGoldberg:
https://github.com/typescript-eslint/typescript-eslint/milestone/10 has a small enough number of bugs -and no major architectural issues- reported that I feel comfortable merging this in. Hooray! 🚀 


_(knocking on wood, fingers crossed 🤞)_

JoshuaKGoldberg:
This change terrifies me.

Tom Hanks in a captain's outfit saying 'May God have mercy on my soul'

.github/workflows/ci.yml
```diff
@@ -193,6 +193,38 @@ jobs:
...
+        uses: ./.github/actions/prepare-install
+        with:
+          node-version: 18
+      - name: Build
+        uses: ./.github/actions/prepare-build
+      - name: Run unit tests for ${{ matrix.package }}
+        run: npx nx test ${{ matrix.package }} --coverage=false
+        env:
+          CI: true
+          TYPESCRIPT_ESLINT_EXPERIMENTAL_TSSERVER: true
```
bradzacher, line 226: This was a great thing to add!

PR #6594, opened 2023-03-10
feat(eslint-plugin): [no-unnecessary-cond] check logical assignments

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3553
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Fixes #3553

PR #6762, opened 2023-03-24
fix(eslint-plugin): [no-unnecessary-condition] ignore optionally accessing index signatures

### LINKED ISSUE #6632: Bug: [no-unnecessary-condition] warns about conditional assignment to index signature

### Playground Link

https://typescript-eslint.io/play#ts=4.9.5&sourceType=module&code=MYewdgzgLgBAZiEAuGAlApqATgEwDzRYCWYA5gDQxgCuAtgEbpYB8MAvDAN4C+A3ALAAoUJFj0AhlhSESpdjADkAa3QBPBQMFCEIANoSsAXRgB+ExwAMvGAHob8RPsnGiEKiFgkiUIuIA2RABe6DiU9NSwIjjeROD+MOIQEESkYLToYJ5ucH7ipKQhCW7UYGCY6EmSqgB02o4GxgDUHACMmkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK+SpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A

### Repro Code

```typescript
const foo: Record<string, number> = {};
const bar: string = 'key';

foo[bar] ??= 0; // foo[bar] is not initialized, but conditional assignment is flagged as unnecessary.
foo[bar] += 1;
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-unnecessary-condition": "error",
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true
  }
}
```

### Expected Result

No warning.

### Actual Result

Unnecessary conditional, expected left-hand side of `??` operator to be possibly null or undefined. 4:1 - 4:9

### Additional Info

This is basically https://github.com/typescript-eslint/typescript-eslint/issues/6264 which was fixed in https://github.com/typescript-eslint/typescript-eslint/pull/6452; I guess it resurfaced because the rule now checks logical assignments as of https://github.com/typescript-eslint/typescript-eslint/pull/6594.

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6632

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -182,17 +182,32 @@ export default createRule<Options, MessageId>({
       return checker.isTupleType(nodeType);
     }
 
-    function isArrayIndexExpression(node: TSESTree.Expression): boolean {
+    // Index accesses are exempt from certain checks about nullish-ness, because
+    // the type returned by the checker is not the strictest
+    function isIndexAccessExpression(node: TSESTree.Expression): boolean {
+      if (node.type !== AST_NODE_TYPES.MemberExpression) {
+        return false;
+      }
+      const objType = getNodeType(node.object);
+      if (checker.isTupleType(objType)) {
+        // If indexing a tuple with a literal, the type will be sound
+        return node.property.type !== AST_NODE_TYPES.Literal;
+      }
+      if (node.computed) {
+        const indexType = getNodeType(node.property);
+        if (isTypeAnyType(indexType)) {
+          // No need to check anything about any
+          return true;
+        }
+        if (isTypeFlagSet(indexType, ts.TypeFlags.NumberLike)) {
+          return (
+            checker.getIndexTypeOfType(objType, ts.IndexKind.Number) !==
+            undefined
+          );
+        }
+      }
       return (
-        // Is an index signature
-        node.type === AST_NODE_TYPES.MemberExpression &&
-        node.computed &&
-        // ...into an array type
-        (nodeIsArrayType(node.object) ||
-          // ... or a tuple type
-          (nodeIsTupleType(node.object) &&
-            // Exception: literal index into a tuple - will have a sound type
-            node.property.type !== AST_NODE_TYPES.Literal))
+        checker.getIndexTypeOfType(objType, ts.IndexKind.String) !== undefined
       );
     }
 
@@ -215,7 +230,7 @@ export default createRule<Options, MessageId>({
       // Since typescript array index signature types don't represent the
       //  possibility of out-of-bounds access, if we're indexing into an array
       //  just skip the check, to avoid false positives
-      if (isArrayIndexExpression(node)) {
+      if (isIndexAccessExpression(node)) {
         return;
       }
 
@@ -276,7 +291,7 @@ export default createRule<Options, MessageId>({
         //  possibility of out-of-bounds access, if we're indexing into an array
         //  just skip the check, to avoid false positives
         if (
-          !isArrayIndexExpression(node) &&
+          !isIndexAccessExpression(node) &&
           !(
             node.type === AST_NODE_TYPES.ChainExpression &&
             node.expression.type !== AST_NODE_TYPES.TSNonNullExpression &&
@@ -490,7 +505,7 @@ export default createRule<Options, MessageId>({
     ): boolean {
       const lhsNode =
         node.type === AST_NODE_TYPES.CallExpression ? node.callee : node.object;
-      if (node.optional && isArrayIndexExpression(lhsNode)) {
+      if (node.optional && isIndexAccessExpression(lhsNode)) {
         return true;
       }
       if (
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
@@ -318,15 +318,41 @@ if (returnsArr?.()[42]) {
 }
 returnsArr?.()[42]?.toUpperCase();
     `,
-    // nullish + array index
+    // nullish + index signature
     `
 declare const arr: string[][];
+declare const x: any;
 arr[x] ?? [];
     `,
-    // nullish + optional array index
     `
 declare const arr: { foo: number }[];
 const bar = arr[42]?.foo ?? 0;
+    `,
+    `
+declare const foo: Record<string, number>;
+declare const bar: string;
+
+foo[bar] ??= 0;
+    `,
+    `
+declare const foo: Record<string, number>;
+declare const bar: string;
+
+foo[bar] ?? 0;
+    `,
+    `
+function getElem(dict: Record<string, { foo: string }>, key: string) {
+  if (dict[key]) {
+    return dict[key].foo;
+  } else {
+    return '';
+  }
+}
+    `,
+    `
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
```diff
@@ -1670,26 +1686,6 @@ pick({ foo: 1, bar: 2 }, 'bar');
     },
     {
       code: `
-function getElem(dict: Record<string, { foo: string }>, key: string) {
-  if (dict[key]) {
-    return dict[key].foo;
-  } else {
-    return '';
-  }
-}
```
Josh-Cena, line 1679: This was arguably a bug fix in the existing test case, because it shouldn't be reported either.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -182,17 +182,32 @@ export default createRule<Options, MessageId>({
...
+    function isIndexAccessExpression(node: TSESTree.Expression): boolean {
+      if (node.type !== AST_NODE_TYPES.MemberExpression) {
+        return false;
+      }
+      const objType = getNodeType(node.object);
+      if (checker.isTupleType(objType)) {
+        // If indexing a tuple with a literal, the type will be sound
+        return node.property.type !== AST_NODE_TYPES.Literal;
+      }
+      if (node.computed) {
```
bradzacher, line 196: It's not just computed expressions - it's all expressions! TS specifically just cares about whether or not the index signature access has been guarded or not!

[playground](https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true#code/PTAEAEDsHsFVIMYAsCmCDWKAmBJSWUAPbAQQQRQGdKAoGghAGwEMAnFUBaSSgF1GgAjAFYAuUACU00VlgA8fVgEtIAcwA0oSAFcAtoJSsAfAG46Q4QG0A5M2sBdAHS9oAMSXEsACgCUZi46Czm4e2L5m9Ggs7JzcfKDo4ooqqv4iluhOLu6e4UA)

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
bradzacher: This actually should be reported because the fixture config doesn't use `noUncheckedIndexAccess`!

I think we should probably guard the new code around whether or not the flag is on.

bradzacher (changes requested):
Thanks for getting straight on this!
Almost there I think!

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
Josh-Cena: @bradzacher This code is the exact same principle as `dictionary[key] ??= defaultVal`. If we want to ignore that, we have to ignore others as well.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
Josh-Cena: @bradzacher If it's not a computed member access, then it necessarily uses the string index signature, which is checked below.

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
bradzacher: They are and they aren't. You're correct in that they're both record index access in a logical position - so they're semantically the same case. However TS reports the types differently for the two cases.

In this instance TS will report the read type, meaning that if `noUncheckedIndexAccess` is turned on then TS will report the type as `T | undefimed` to us.

However in `x ?? y`, TS will report the write type, meaning it reports just `T` - no matter whether or not `noUncheckedIndexAccess` is on or not.

This is the fundamental difference between the two cases and why we are false positiving on the assignment operators, but _not_ on this case.

Josh-Cena: However, we already ignore array index access, even without `noUncheckedIndexAccess`. See this test case:

```ts
declare const arr: { foo: number }[];
const bar = arr[42]?.foo ?? 0;
```

I find it quite natural to extend the scope to all index signatures, since there isn't something special about array indices.

bradzacher: The reason we did it for arrays and not objects is because defining arrays as `(T | undefined)[]` is similar to `Record<string, T | undefined>`, but far, far less ergonomic syntactically and is arguably worse given how common it is to iterate arrays directly using an iterator method or for/of.

For objects it's pretty natural to union in `undefined` because most usecases for reading is an index signature - the direct iteration methods (`Object.entries` / `Object.values`) are both rarer usecases so the `undefined` doesn't get in the way as much.

Now that we have `noUncheckedIndexAccess` - we don't really need the logic at all, but I understand that some people think the compiler option goes too far and can get in the way (which is true to a degree).

The motivation behind this change isn't bringing the two sets up to parity (that's a much larger change that we should discuss as a group to decide if we think we should treat objects the same as arrays in this regard) - instead it's just fixing the current false positives we have due to the logical operator handling we added in #6594

JoshuaKGoldberg:
👋 @Josh-Cena just checking in, is this still something you have time for?

Josh-Cena:
Yes! I'll get to it. Sorry for the wait. I still don't know what the best way to fix this is, because I feel like accounting for index signature optionality, even without the compiler option, should be acceptable.

Josh-Cena:
Yes! I'll get to it. Sorry for the wait. I still don't know what the best way to fix this is, because I feel like accounting for index signature optionality, even without the compiler option, should be acceptable.

hbiede:
Hoping this will be available in the near-term future. I'd love to use this rule without having to use dozens of casts

karlhorky:
It seems like https://github.com/microsoft/TypeScript/pull/54777 was merged, which closed the issue in the TypeScript repo:

- https://github.com/microsoft/TypeScript/issues/53305#event-10100428639

Does that mean that this PR is now obsolete?

Josh-Cena:
@karlhorky That PR changes the quick info of the write type so it excludes `undefined` (and the programmatic API always excludes it), which makes this PR even more necessary :)

Josh-Cena:
@bradzacher Sorry it's been a really long time and I think neither of us are up to speed on this now. Let's recap—what (potential) false-positives should we fix and what should we continue erroring on? Currently we ignore all `object[index]` where `object` has index signatures, because that would always be potentially undefined.

bradzacher (changes requested):
originally I was against us bringing arrays and objects into parity.
however given the state `noUncheckedIndexAccess` is in esp in regards to `??=` - I shall acquiesce here and we can loosen the rule to just ignore really anything with an index signature.

A few comments but otherwise looking good.

JoshuaKGoldberg:
Closing this PR as it's been stale for a while without activity. If anybody wants to drive it forward, please do post your own PR - and if you use this as a start, consider adding a co-author attribution at the end of your PR description. Thanks! 😊

PR #5750, opened 2022-10-06
feat(utils): add dependency constraint filtering for `RuleTester`

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: ref #1752
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

While working on #5573 I realised that we need infrastructure that allows us to skip tests depending on dependency versions (i.e. the current ESLint or TS version).

This PR adds that functionality into our extended version of the `RuleTester` class.
You will be able to specify constraints like so:

```ts
ruleTester.run('my-rule', rule, {
  valid: [
    {
      code: 'const x = 1;',
      dependencyConstraints: {
        // enforces the semver ">=7.32.0"
        eslint: '7.32.0',

        // i also added this form which allows the user to fully control the semver.satisfies call
        'some-other-module': {
          type: 'semver',
          version: '~6',
          options: {
            includePrerelease: true,
          },
        },
      },
    },
  ],
  invalid: [
    {
      code: 'const x = 1;',
      errors: [ /* ... */ ],
      dependencyConstraints: {
        eslint: '7.32.0',
        // etc - same as above
      },
    },
  ],
});
```

Issue #1752, opened 2020-03-17
Add automated regression testing against old TS versions


bradzacher:
#1698 introduced a regression for TS<3.8.x.
This wasn't caught by the type checker, nor was it caught by the tests because both were using TS3.8.

It was a simple mistake, but one that shouldn't have happened.
We should add an integration test in order to make sure that this doesn't happen again.

We cannot run builds against older versions of TS, because our code relies upon property types etc added by the latest versions, but we can easily run tests against the old versions.
I don't think we need to test _every_ version we support, but at least testing against the previous one or two minors should be good enough.

JoshuaKGoldberg:
Copying the useful bits from my dup issue:

The simplest approach might be extending CI's [`unit_tests_on_other_node_versions`](https://github.com/typescript-eslint/typescript-eslint/blob/f785965a12e20e82de1c7f3d039023c225115eb1/.github/workflows/ci.yml#L231) to also switch on different TypeScript package versions ([example syntax](https://github.community/t/create-matrix-with-multiple-os-and-env-for-each-one/16895/2)). 

It'd be great as a next step _(followup issue?)_ to also make use of [CodeCov's report merging](https://docs.codecov.io/docs/merging-reports) to combine unit test coverage from all of them. #2516 is an example PR where code under test changes behavior on the TypeScript version.

bradzacher:
The thing that makes this difficult is ensuring that the rules that have variant functionality on TS versions will have to pass on the other TS versions (this is the same problem that makes  #2464 difficult)

To make this work we'll have to extend and customise the core ESLint RuleTester and design a config system for it.

----

There's a lot of codecov features we should totally leverage more of - I've just never bothered to read the docs 😅

JoshuaKGoldberg:
To start, can that be as simple as checking the TypeScript version at Jest runtime when constructing the `.test.ts` file tests?

`tests/rules/some-rule.test.ts`:

```ts
ruleTester.run('some-rule', rule, {
  // ...
});

if (util.isAboveTsVersion('3.9')) {
  ruleTester.run('some-rule typescript@>=3.9', rule, {
    // ...
  });
} else {
  ruleTester.run('some-rule typescript@<3.9', rule, {
    // ...
  });
}
```

`util/versions.ts`:

```ts
// (may need to add overrides or some logic if any TS releases had unusual version numbers...
export const isAboveTsVersion = (version: string) =>
  semver.satisfies(
    ts.version,
[...]

PR #6777, opened 2023-03-27
feat: add new package `rule-tester`

bradzacher:
## PR Checklist

References:
- https://github.com/typescript-eslint/typescript-eslint/discussions/6499
- #5750
- #1752

## Overview

The changes I implemented in #5750 were pretty complex to build because I was working around how the base `RuleTester` class did things. I ran into a lot of weirdness due to things like the class expecting specific properties on objects.

In looking to implement #6499 - I don't want to hit the same issues, so I think it's best if we just draw a line in the sand and fork it so that we can build our enhancements directly into the tooling.

This is also good because it means that we can provide a new baseline for testing and don't need to worry about what the installed ESLint version supports - which will make our eventual dependency testing easier.

This PR:
- fork the base class and adds types
    - The logic itself is unchanged, but I did move some things around to make it a bit easier to type and manage.
- clones the base eslint test for `RuleTester`
    - This test is completely unchanged because it'll take a complete rewrite to make it jest-compatible.
- adds the changes we've added in our existing subclass fork:
    - adds `afterAll` as additional static "test framework providers" to the class.
    - adds a new per-test option - `skip` which runs the test using `it.skip`.
    - adds automatic parser cache cleaning using `afterAll`.
    - adds automatic file naming for type-aware linting
        - I implemented this slightly differently than before:
            - it uses a different default name for `tsx` files now
            - the default filenames can be overridden in the constructor (previously you HAD to do it per test).
    - adds per-test dependency constraints
        - I implemented this slightly differently than before:
          Before our hands were tied and I had to lean on setting `only` on every test to filter out the skipped tests - now with the new `skip` functionality we only need set `skip` on these tests which is a much, much nicer implementation!
- adds new features
    - adds a new per-test option - `skip` which runs the test case using `it.skip`.
    - adds `itSkip` and `describeSkip` as additional static "test framework providers" to the class.
    - adds a new `defaultFilenames` constructor option to allow configuring the default filenames used for type-aware testing.

Documentation:
https://deploy-preview-6777--typescript-eslint.netlify.app/architecture/rule-tester

TODO:
- [x] add a new package
- [x] fork the base class
- [x] implement the changes we have in [our current subclass fork](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/utils/src/eslint-utils/rule-tester)
    - [x] implement automatic parser cache cleaning
    - [x] implement automatic file naming for type-aware linting
    - [x] implement the dependency checks from #5750
- [ ] ~~migrate codebase to this ruletester~~
    - i'll do this in a separate PR
- [ ] ~~implement snapshot testing #6499~~
    - i'll do this in a separate PR


packages/eslint-plugin-tslint/src/rules/config.ts
```diff
@@ -1,10 +1,27 @@
...
+    if (cache.has(key)) {
+      return cache.get(key)!;
+    }
+    const result = func(...args);
+    cache.set(key, result as ReturnType<T>);
+    return result;
+  } as T;
+  return memoized;
+}
+
```
armano2, line 24: can we extract this change to standalone PR?

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -424,6 +425,19 @@ interface RuleListener {
...
+interface RuleListenerCatchAllBaseCase {
+  [nodeSelector: string]: RuleFunction | undefined;
+}
+// Interface to merge into for anyone that wants to add more selectors
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+interface RuleListenerExtension {}
+
+type RuleListener = RuleListenerBaseSelectors &
+  RuleListenerExitSelectors &
+  RuleListenerCatchAllBaseCase;
```
armano2, line 440: https://github.com/typescript-eslint/typescript-eslint/pull/6679/files#diff-53e6896dd0f15121e29e2dfc7309a38d6d6637fc0a5a3bfe64bc4007ab5fca64R457

maybe we could add missing codePath functions to type?

bradzacher: I just haven't bothered with them just because honestly I've never found an actual use for them in our lint rules.
Our users are apparently in the same boat because nobody has asked for them (317 dependents of our utils package listed in npm).

Maybe one day but I don't think it's worth messing up the DevX of node selectors for until we have an actual motivating usecase for.

packages/eslint-plugin-tslint/src/rules/config.ts
bradzacher: Yeah will do. I was just looking to see if we had a lodash dependency (because the rule tester needed one) and I noticed this plugin used the entire lodash package for just one function!

I'll extract it out because it's not worth the massive dependency, esp given we can simplify the function massively.

packages/utils/src/ts-eslint/Rule.ts
armano2: https://github.com/SonarSource/eslint-plugin-sonarjs/blob/4dfd992eeccafca92caed52889b4f83bb1e38c3c/src/rules/no-one-iteration-loop.ts#L41

one of examples :P

packages/eslint-plugin-tslint/src/rules/config.ts
armano2: if possible we should just drop it

bradzacher: https://github.com/typescript-eslint/typescript-eslint/pull/6810

packages/rule-tester/typings/eslint.d.ts
```diff
@@ -0,0 +1,24 @@
...
+      // ...
+    };
+    environments: Map<string, Linter.Environment>;
+    // ...
+  };
+}
+
+declare module 'eslint' {
+  export { SourceCode } from '@typescript-eslint/utils/ts-eslint';
+}
```
JoshuaKGoldberg, line 24: Lol.

JoshuaKGoldberg (approved):
Whew, what a PR! This is great - looking forward to having native dependency contraints!

docs/architecture/Rule_Tester.mdx
```diff
@@ -0,0 +1,216 @@
+---
+id: rule-tester
+sidebar_label: rule-tester
+---
+
+import CodeBlock from '@theme/CodeBlock';
+
+# `@typescript-eslint/rule-tester`
```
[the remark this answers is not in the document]
bradzacher: As per our chats on discord - @armano2 and I think we should push to make this a breaking change replacement in v6 - people should switch to it cos it's better in many ways and it should mostly be a 1:1 replacement.

Also switching to the package means people get the `only: boolean` test prop for free, regardless of their ESLint version.

docs/architecture/Rule_Tester.mdx
```diff
@@ -0,0 +1,216 @@
...
+
+With that config the parser will automatically run in type-aware mode and you can write tests just like before.
+
+### Test Dependency Constraints
+
+Sometimes it's desirable to test your rule against multiple versions of a dependency to ensure backwards and forwards compatibility.
+With backwards-compatibility testing there comes a complication in that some tests may not be compatible with an older version of a dependency.
+For example - if you're testing against an older version of TypeScript, certain features might cause a parser error!
+
+import DependencyConstraint from '!!raw-loader!../../packages/rule-tester/src/types/DependencyConstraint.ts';
```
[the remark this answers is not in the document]
bradzacher: It's much nicer than us copy-pasting and manually syncing things fo sho

packages/rule-tester/src/RuleTester.ts
```diff
@@ -0,0 +1,1028 @@
...
+  AnyRuleModule,
+  ParserOptions,
+  RuleContext,
+  RuleModule,
+} from '@typescript-eslint/utils/ts-eslint';
+import { Linter } from '@typescript-eslint/utils/ts-eslint';
+// we intentionally import from eslint here because we need to use the same class
+// that ESLint uses, not our custom override typed version
+import { SourceCode } from 'eslint';
+import merge from 'lodash.merge';
```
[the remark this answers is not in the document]
bradzacher: I thought about that - but ESLint has this exact import for its rule tester.
So I figured that we'd just use the exact same code to avoid a duplicate install

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -424,6 +425,18 @@ interface RuleListener {
...
   YieldExpression?: RuleFunction<TSESTree.YieldExpression>;
 }
+type RuleListenerExitSelectors = {
+  [K in keyof RuleListenerBaseSelectors as `${K}:exit`]: RuleListenerBaseSelectors[K];
+};
+interface RuleListenerCatchAllBaseCase {
+  [nodeSelector: string]: RuleFunction | undefined;
+}
+// Interface to merge into for anyone that wants to add more selectors
+interface RuleListenerExtension {}
```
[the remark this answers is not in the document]
bradzacher: That was just me adding some future-proofing in case anyone was doing something weird.
Actually don't need it - but I think it's fine to leave it undocumented for the moment.

packages/utils/src/json-schema.ts
```diff
@@ -1,24 +1,2 @@
-// Note - @types/json-schema@7.0.4 added some function declarations to the type package
-// If we do export *, then it will also export these function declarations.
-// This will cause typescript to not scrub the require from the build, breaking anyone who doesn't have it as a dependency
```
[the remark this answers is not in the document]
bradzacher: yeah because in the olden days we did `export * from 'json-schema'` and we'd export the types for functions that don't actually exist.
But now a days we can do `export type *` and it'll not include the runtime bits!

Though #6963 would just be better than making this change at all TBH

PR #6780, opened 2023-03-27
fix(type-utils): fixed TypeOrValueSpecifier not accounting for scoped DT packages

### LINKED ISSUE #6698: Bug: type-utils > TypeOrValueSpecifier doesn't account for org-scoped packages

### Relevant Package

ast-spec

### Description

From https://github.com/typescript-eslint/typescript-eslint/pull/4436#discussion_r1134628059: when checking whether a `'package'` matches in `TypeOrValueSpecifier.ts`, we check whether the declaration file name includes `specifier.package` directly:

https://github.com/typescript-eslint/typescript-eslint/blob/346fdbc1c2cfe63a6cd006c22b652d536332cefc/packages/type-utils/src/TypeOrValueSpecifier.ts#L170-L176

This is a potential issue because package names with `@` in them have a `__` replacing the `@`. And there might be other replacements I'm not remembering - haven't looked deeply into it. 

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/cc222f340546ff85b7c9271f4ecbd07b9d661c94/types/11ty__eleventy-img

We should handle any name substitutions in npm packages well, and add unit tests for it in `TypeOrValueSpecifier.test.ts`.

### Fail

```typescript
n/a
```

### Pass

```typescript
n/a
```

### Additional Info

_No response_

marekdedic:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6698

## Overview

- Added TypeOrValueSPecifier tests for package specifiers

### DIFF
--- a/packages/type-utils/src/TypeOrValueSpecifier.ts
+++ b/packages/type-utils/src/TypeOrValueSpecifier.ts
@@ -146,6 +146,22 @@ function typeDeclaredInFile(
   );
 }
 
+function typeDeclaredInPackage(
+  packageName: string,
+  declarationFiles: ts.SourceFile[],
+): boolean {
+  const typesPackageName =
+    '@types/' +
+    (packageName[0] === '@'
+      ? packageName.substring(1).replaceAll('/', '__')
+      : packageName);
+  return declarationFiles.some(
+    declaration =>
+      declaration.fileName.includes(`node_modules/${packageName}/`) ||
+      declaration.fileName.includes(`node_modules/${typesPackageName}/`),
+  );
+}
+
 export function typeMatchesSpecifier(
   type: ts.Type,
   specifier: TypeOrValueSpecifier,
@@ -170,12 +186,6 @@ export function typeMatchesSpecifier(
         program.isSourceFileDefaultLibrary(declaration),
       );
     case 'package':
-      return declarationFiles.some(
-        declaration =>
-          declaration.fileName.includes(`node_modules/${specifier.package}/`) ||
-          declaration.fileName.includes(
-            `node_modules/@types/${specifier.package}/`,
-          ),
-      );
+      return typeDeclaredInPackage(specifier.package, declarationFiles);
   }
 }
--- a/packages/type-utils/src/TypeOrValueSpecifier.ts
+++ b/packages/type-utils/src/TypeOrValueSpecifier.ts
@@ -153,10 +153,11 @@ function typeDeclaredInPackage(
   // Handle scoped packages - if the name starts with @, remove it and replace / with __
   const typesPackageName =
     '@types/' + packageName.replace(/^@([^/]+)\//, '$1__');
-  return declarationFiles.some(
-    declaration =>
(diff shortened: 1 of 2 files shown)


packages/type-utils/tests/TypeOrValueSpecifier.test.ts
```diff
@@ -139,7 +139,8 @@ describe('TypeOrValueSpecifier', () => {
         .program!.getTypeChecker()
         .getTypeAtLocation(
           services.esTreeNodeToTSNodeMap.get(
-            (ast.body[0] as TSESTree.TSTypeAliasDeclaration).id,
+            (ast.body[ast.body.length - 1] as TSESTree.TSTypeAliasDeclaration)
+              .id,
```
marekdedic, line 143: I actually think this fixes a minor issue that is probably present in more tests - I have taken this directly out of some other tests in this repo

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -146,6 +146,22 @@ function typeDeclaredInFile(
...
 
+function typeDeclaredInPackage(
+  packageName: string,
+  declarationFiles: ts.SourceFile[],
+): boolean {
+  const typesPackageName =
+    '@types/' +
+    (packageName[0] === '@'
+      ? packageName.substring(1).replace('/', '__')
+      : packageName);
```
JoshuaKGoldberg, line 157: [Refactor] ✨ 

```suggestion
  const typesPackageName = '@types/' + packageName.replace(/^@(.+)\//, '$1__');
```

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -146,6 +146,22 @@ function typeDeclaredInFile(
...
+  return declarationFiles.some(
+    declaration =>
+      declaration.fileName.includes(`node_modules/${packageName}/`) ||
+      declaration.fileName.includes(`node_modules/${typesPackageName}/`),
+  );
```
JoshuaKGoldberg, line 162: [Refactor] Just to remove the need for two iterations over the string... if you dislike this I think it's totally reasonable to ignore the suggestion though 😄 

```suggestion
  const matcher = new RegExp(
    `node_modules/(?:${packageName}|${typesPackageName})/`,
  );

  return declarationFiles.some(declaration =>
    matcher.test(declaration.fileName),
  );
```

JoshuaKGoldberg (changes requested):
LGTM in general, thanks for the fast follow! Just requesting changes on 1-2 cleanups, lmk what you think?

packages/type-utils/src/TypeOrValueSpecifier.ts
marekdedic: Thanks, I tweaked the regexp a bit more FYI

packages/type-utils/src/TypeOrValueSpecifier.ts
marekdedic: Thanks!

PR #6781, opened 2023-03-27
fix(type-utils): file variant of TypeOrValueSpecifier uses canonical filenames instead of lowercasing

### LINKED ISSUE #6697: Bug?: type-utils > TypeOrValueSpecifier's file detection is case insensitive

### Relevant Package

type-utils

### Description

Following up from https://github.com/typescript-eslint/typescript-eslint/pull/4436/files#r1134626465: in order to get the PR merged, the _"is this type declared in one of these files?"_ logic (`typeDeclaredInFile`) logic was changed to be case _insensitive_. It `.toLowerCase()`s all strings.

That means this config:

```js
{ from: 'file', name: 'MyType', path: './filename' }
```

...will match `./filename`, `./Filename`, `./FILENAME`, and so on.

This case insensitivity _probably_ won't matter in any projects. But most file systems are in fact case sensitive. So we really should get it to work at some point.

### Additional Info

To start on this issue, remove the `.toLowerCase()` calls in `typeDeclaredInFile`. https://github.com/typescript-eslint/typescript-eslint/blob/346fdbc1c2cfe63a6cd006c22b652d536332cefc/packages/type-utils/src/TypeOrValueSpecifier.ts#L130

Re-run unit tests for `type-utils` and `eslint-plugin`. You'll also want to add unit tests to `TypeOrValueSpecifier.test.ts` to make sure that case sensitivity is respected.

marekdedic:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6697


marekdedic:
Hi,
since this was previously an issue on your machine, I am tagging @JoshuaKGoldberg in particular (I cannot assign you as a reviewer) to verify this works for them, thanks!

packages/typescript-estree/src/index.ts
```diff
@@ -16,6 +16,7 @@ export { simpleTraverse } from './simple-traverse';
 export * from './ts-estree';
 export { createProgramFromConfigFile as createProgram } from './create-program/useProvidedPrograms';
 export * from './create-program/getScriptKind';
+export { getCanonicalFileName } from './create-program/shared';
```
marekdedic, line 19: Is this OK?

JoshuaKGoldberg (changes requested):
Yeah I think this is very reasonable 👍. Just, it needs some test coverage please?

marekdedic:
Well, I added a test for the normalization, however, I can't really test for capitalization - the test would then fail on some platforms, right? Does it pass on your machine?

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -1,3 +1,4 @@
+import { getCanonicalFileName } from '@typescript-eslint/typescript-estree';
```
JoshuaKGoldberg, line 1: @bradzacher just confirming on architecture things - is there a better dependency setup we can use here?

JoshuaKGoldberg (approved):
Yeah I see what you mean - testing this is hard. But what's here so far looks good to me, thanks!

Since this exports a previously internal helper I'll defer to its author for a second review.

packages/type-utils/src/TypeOrValueSpecifier.ts
bradzacher: No this is probably the best way to do it and ensures we keep things in sync with the parser's logic.

PR #3460, opened 2021-05-28
fix: generate library types for TypeScript v4.3

bradzacher:
Forgot to regenerate these.

Fixes #3449

PR #6782, opened 2023-03-27
fix: update getLib for new TypeScript targets ES2021, ES2022

andersk:
## PR Checklist

## Overview

Support for the `es2021.full` and `es2022.full` libs was added in #3460 and #4615, but the `getLib` function wasn’t updated to use them; if `target` is set to `"ES2021"` or `"ES2022"` in `tsconfig.json`, `getLib` would incorrectly fall back to `['lib']` rather than `['es2021.full']` or `['es2022.full']`. Add these missing cases.

### DIFF
--- a/packages/parser/src/parser.ts
+++ b/packages/parser/src/parser.ts
@@ -54,10 +54,14 @@ function getLib(compilerOptions: CompilerOptions): Lib[] {
   }
 
   const target = compilerOptions.target ?? ScriptTarget.ES5;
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (approved):
Nice find, thanks @andersk!

Out of curiosity - how did you notice this? 😄

andersk:
I found this in https://github.com/zulip/zulip/pull/24875 where the `no-undef` rule started complaining about more globals when we switched the TypeScript target from ES2020 to ES2022. (I know you don’t recommend `no-undef` with typescript-eslint, but we want to prohibit some globals like jQuery that TypeScript thinks are globally available but we want to be explicit about importing.)

Issue #5449, opened 2022-08-09
Bug: [no-unnecessary-type-constraint] False positive for generics on arrow functions in .mts files


spalger:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Abug) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/docs/linting/troubleshooting) and my problem is not listed.

### Issue Description

When using the new ([as of TS 4.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#new-file-extensions)) `.mts` file extension the `const foo = <T>() => ...` syntax for generic arguments is reserved similar to JSX files. The suggestion from tsc recommends defining an explicit constraint:

<img width="1633" alt="image" src="https://user-images.githubusercontent.com/1329312/183544789-01f13bf7-6b1b-4d1c-bf9c-140aff9613e5.png">

Since my function doesn't need or nor do I want a type constraint, but I need one for the syntax, I reached for the noop constraint of `extends any`, but then I was violating the `no-unnecessary-type-constraint` rule.

<img width="1785" alt="image" src="https://user-images.githubusercontent.com/1329312/183545544-05e4430b-c230-4196-a326-d691d0dc5204.png">

In this case, with the `.mts` extension and it's limited syntax I think it's unfair to say it "doesn't do anything and is unnecessary".

### Reproduction Repository Link

https://github.com/spalger/reproduce-issues/tree/ts-eslint-no-unnecessary-type-constraint

### Repro Steps

1) clone the repo with `--branch`

    `git clone git@github.com:spalger/reproduce-issues.git --branch ts-eslint-no-unnecessary-type-constraint`

2) `npm i`
3) `npm run lint`

### Versions

| package                                | version |
| -------------------------------------- | ------- |
| `@typescript-eslint/eslint-plugin`     | `5.33.0` |
| `@typescript-eslint/parser`            | `5.33.0` |
| `TypeScript`                           | `4.7.4` |
| `ESLint`                               | `8.21.0` |
| `node`                                 | `18.7.0` |

JoshuaKGoldberg:
Somehow, I had a feeling this line of code would come back to haunt us 😄:

[...]

PR #6785, opened 2023-03-28
chore(website): [playground] allow to choose file extensions

armano2:
## PR Checklist

## Overview

> extracted from #6656 

This change replaces `jsx` checkbox with dropdown of supported file extensions

image

ref #5449


packages/website/src/components/editor/LoadedEditor.tsx
```diff
@@ -28,12 +28,25 @@ export interface LoadedEditorProps extends CommonEditorProps {
...
+    return 'typescript';
+  }
+  if (/\.[mc]?js(x)?$/.test(file)) {
+    return 'javascript';
+  }
+  if (/\.(json|eslintrc)$/.test(file)) {
+    return 'json';
+  }
+  return 'plaintext';
+}
```
bradzacher, line 42: could this use this function instead?

https://github.com/typescript-eslint/typescript-eslint/blob/399b3b1263665288316be6faff759cf514e24b1b/packages/typescript-estree/src/create-program/getScriptKind.ts#L4-L34

packages/website/src/components/config.ts
```diff
@@ -1,6 +1,20 @@
...
+  'ts',
+  'tsx',
+  'js',
+  'jsx',
+  'd.ts',
+  'cjs',
+  'mjs',
+  'cts',
+  'mts',
 ];
```
bradzacher, line 20: could this be `Object.values(ts.Extension)` instead?

packages/website/src/components/types.ts
```diff
@@ -14,14 +14,27 @@ export interface RuleDetails {
...
+export type ConfigFileType =
+  | 'ts'
+  | 'tsx'
+  | 'js'
+  | 'jsx'
+  | 'd.ts'
+  | 'cjs'
+  | 'mjs'
+  | 'cts'
+  | 'mts';
```
bradzacher, line 26: could this be `ts.Extension` instead?

packages/website/src/components/Playground.tsx
```diff
@@ -50,7 +50,7 @@ function rangeReducer<T extends SelectedRange | null>(
 }
 function Playground(): JSX.Element {
   const [state, setState] = useHashState({
-    jsx: false,
+    fileType: 'ts',
```
bradzacher, line 53: I think we should default this to TSX.
IMO it's the rarer case that people want the non-TSX supported syntax (which really is just angle bracket assertions), and much more common that people want to use JSX in the playground.

packages/website/src/components/OptionsSelector.tsx
```diff
@@ -74,11 +74,12 @@ function OptionsSelectorContent({
...
       <Expander label="Options">
-        <InputLabel name="Enable jsx">
-          <Checkbox
-            name="jsx"
-            checked={state.jsx}
-            onChange={(e): void => setState({ jsx: e })}
+        <InputLabel name="File type">
+          <Dropdown
+            name="fileType"
+            value={state.fileType ?? 'ts'}
```
bradzacher, line 80: ditto below - can we default to TSX?

packages/website/src/components/types.ts
armano2: my issue with using ts in this context is that we do not have access to it yet, we have to wait for  playground to be fully initialized, and to do so we have to know what extensions has to be choosen,
addtionally we will have to parse remove some options from it `TsBuildInfo` and `Json`

packages/website/src/components/editor/LoadedEditor.tsx
armano2: in theory we could use it in here

```
export function determineLanguage(file: string): string {
 const kind = getScriptKind()
 switch (extension) { 
   case window.ts.ScriptKind.JS:
   
 }
}
```

but we still need extra check for `eslintrc`

packages/website/src/components/config.ts
armano2: a tthis point we do not have access to typescript instance yet as editor may be not loaded

packages/website/src/components/editor/LoadedEditor.tsx
armano2: after further considiration, i actually don't need it i can just let ts decide what language to choose base on available extensions

armano2:
fallback cases to test:

- old url without jsx
  https://deploy-preview-6785--typescript-eslint.netlify.app/play/#ts=5.0.2&sourceType=module&showAST=ts&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA
- old url wiht jsx
  https://deploy-preview-6785--typescript-eslint.netlify.app/play/#ts=5.0.2&jsx=true&sourceType=module&showAST=ts&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

packages/website/src/components/hooks/useHashState.ts
```diff
@@ -165,66 +172,33 @@ const writeStateToLocalStorage = (newState: ConfigModel): void => {
...
+      const url = `${window.location.pathname}#${newHash}`;
 
-  const _setState = useCallback((cfg: Partial<ConfigModel>) => {
-    console.info('[State] updating config diff', cfg);
-    setTmpState(cfg);
+      if (cfg.ts) {
+        window.location.href = url;
+        window.location.reload();
+      } else {
+        window.history.replaceState(undefined, document.title, url);
```
armano2, line 195: note: this should be changed to use react router to fix issues with navigation 

but this is oo scope for this change

packages/website/src/components/hooks/useHashState.ts
```diff
@@ -65,20 +73,22 @@ const parseStateFromUrl = (hash: string): ConfigModel | undefined => {
       );
     }
 
+    let fileType = readFileType(searchParams.get('fileType'));
+    if (searchParams.get('jsx') === 'true') {
+      fileType = '.tsx';
+    }
```
JoshuaKGoldberg, line 79: [Nit] 

```suggestion
    const fileType =
      searchParams.get('jsx') === 'true'
        ? '.tsx'
        : readFileType(searchParams.get('fileType'));
```

JoshuaKGoldberg (approved):
🔥

PR #6795, opened 2023-03-30
fix(eslint-plugin): [no-unnecessary-type-constraint] correctly fix in cts/mts files

### LINKED ISSUE #5449: Bug: [no-unnecessary-type-constraint] False positive for generics on arrow functions in .mts files

WoodyWoodsta:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5449 

## Overview

- Suggest comma disambiguation on `.mts`, `cts` as well as jsx variant files


WoodyWoodsta:
I've specifically left out unit tests for this, as it doesn't appear that any of the fixtures for `eslint-plugin` include .mts or .cts files. It seems like something which should require test-wide/project-wide consensus. Please correct me if I'm wrong!

WoodyWoodsta:
I've noticed two things:
- My changes don't address the problem (at least in my project via linking)
- I can add tests as you can specify the `filename` on the `RuleTester`. Will do that now.

packages/typescript-estree/src/create-program/getScriptKind.ts
```diff
@@ -1,6 +1,32 @@
 import path from 'path';
 import * as ts from 'typescript';
 
+function isNewModuleExtension(filePath: string): boolean {
```
Josh-Cena, line 4: I suggest a better name, because "new" doesn't age where. What about `isNodeNextModuleExtension`

Josh-Cena: The reason why I didn't go with `Node16` is exactly because it's tied to a specific feature and if that resolution mode goes out of fashion (e.g. when `Node24` is introduced, etc.), the function would seem awkward. Unless you plan to make this function do exactly `Node16` logic and do not intend it to accomodate new extensions in the future, I suggest using something more fluid like `NodeNext`.

WoodyWoodsta: That's fine - more just aware that "node-next" is something that changes meaning over time, but the method logic won't unless updated. Ultimately happy with either, so will make it `isNodeNextModuleExtension`.

packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts
```diff
@@ -60,14 +65,21 @@ export default util.createRule({
...
       : new Map([[AST_NODE_TYPES.TSUnknownKeyword, 'unknown']]);
 
-    const inJsx = context.getFilename().toLowerCase().endsWith('tsx');
+    const filename = context.getFilename();
+    const scriptKind = getScriptKind(
+      filename,
+      !!context.parserOptions.ecmaFeatures?.jsx,
+    );
+    const inJsx = getLanguageVariant(scriptKind) === ts.LanguageVariant.JSX;
+    const newModuleExtension = isNewModuleExtension(filename);
```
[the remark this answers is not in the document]
WoodyWoodsta: I don't know the codebase well enough to understand where `program` comes from.

packages/typescript-estree/src/create-program/getScriptKind.ts
WoodyWoodsta: https://github.com/typescript-eslint/typescript-eslint/pull/6795/commits/5a5bc878a65e62682cba2c61cf8ccafc34b35e58

packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts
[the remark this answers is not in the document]
armano2: i updated comment with an example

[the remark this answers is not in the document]
WoodyWoodsta: Oh right ok I understand where `program` comes from now. It looks like you get the script kind as well, so no need to export and use `getLanguageVariant` and `getScriptKind` from `typescript-estree`.

packages/typescript-estree/src/create-program/getScriptKind.ts
```diff
@@ -1,6 +1,32 @@
 import path from 'path';
 import * as ts from 'typescript';
 
+function isNodeNextModuleExtension(filePath: string): boolean {
```
bradzacher, line 4: we don't need to put this within `typescript-estree` as t's not used here and won't ever be used here - this can just live directly in our plugin - I think it can probably live directly within the rule's file.

I'm don't think we really want to expose this - I don't think many other plugins (if any) are going to be wanting this sort of thing.

bradzacher (changes requested):
Apologies - it appears that I've lead you astray due to my misunderstanding of TS.
I've never actually used `.mts` for JSX, but I knew it errored on JSX-ambiguous syntax, so I had assumed it actually supported JSX.

In testing I now see that this isn't actually the case - `.mts`/`.cts` doesn't support JSX, it just errors on ambiguous syntax, and does not actually support JSX in the files. I assume that this just the TS team future-proofing - reading the design meeting notes it appears like they didn't want to lock in one-way or the other for the first release, so they just made sure they could make it JSX later if they wanted to by banning ambiguous syntax.

This obviously the reason why the `getScriptKind` function returns "TS" as the kind for the file instead of "TSX" - which was another thing I hadn't noticed in passing earlier.

Which provides a problem as you've noticed our suggested solution obviously won't fix it.

So what would be a good solution?
Now that I know what we know - I think we should just go back to hard-coding this. Something like this would do the job:
```ts
function requiresGenericDeclarationDisambiguation(filename: string) {
  const ext = path.ext(filename).toLowerCase();
  switch (ext) {
    case ts.Extension.Tsx:
    case ts.Extension.Cts:
    case ts.Extension.Mts:
      return true;

    default:
      return false;
  }
}
```

packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts
```diff
@@ -60,14 +65,22 @@ export default util.createRule({
...
     const source = context.getSourceCode();
 
     const checkNode = (
       node: TypeParameterWithConstraint,
       inArrowFunction: boolean,
     ): void => {
-      const constraint = unnecessaryConstraints.get(node.constraint.type);
+      const constraint =
+        !nodeNextModuleExtension &&
+        unnecessaryConstraints.get(node.constraint.type);
```
bradzacher, line 83: note that if we're in a `.mts` or `.cts` file - we still want to report and apply a fix!
It just needs to be the JSX-style fix, not the non-JSX fix.
I.e. instead of fixing `<T extends any>` to `<T>`, we need to fix it to `<T,>`.

packages/typescript-estree/src/create-program/getScriptKind.ts
WoodyWoodsta: Sure 👍

packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts
WoodyWoodsta: Is this correct though? In my experience, something like `prettier` will then format `<T,>` to `<T>` as also highlighted [here](https://github.com/typescript-eslint/typescript-eslint/issues/5449#issuecomment-1265318049). If you have a pipeline which employs `eslint` **and** `prettier` fixes, then the behaviour will be unexpected.

WoodyWoodsta:
@bradzacher To sum up your comments and relate to the issue:

### Original Issue
- `<T>` requires disambiguation in `mts`, `cts` and JSX language variant files
- `<T extends any>` as a disambiguation reports `no-unnecessary-type-constraint` in `mts` and `cts` files
- rule fix, `<T>` -> `<T,>`, is stripped of the dangling comma with at least `prettier` as a formatter returning it to the place of needing disambiguation

### Fix
- Only report `no-unnecessary-type-constraint` for `<T extends any>` in files which do not require disambiguation (everything other than `mts`, `cts` and JSX language variant files)

No changes to eslint fix required.

Let me know if this sounds correct.

---

Of course this could also be resolved in some way in formatters (i.e. to not strip the dangling comma here), but:
- it's a matter of opinion which of `<T,>` and `<T extends any>` is "better"
- stripping dangling commas from single-line argument lists **except** for in this case requires quite a lot of contextual awareness which I'm not sure whether these formatters have 🤷

bradzacher:
> rule fix, <T> -> <T,>, is stripped of the dangling comma with at least prettier as a formatter returning it to the place of needing disambiguation

If prettier is stripping the syntactically necessary comma when formatting a `.cts/mts`  file, then that's a prettier bug to be reported and fixed.

It has already been reported: https://github.com/prettier/prettier/issues/14518

> stripping dangling commas from single-line argument lists except for in this case requires quite a lot of contextual awareness which I'm not sure whether these formatters have

Formatters are powered by the AST - they know exactly and unambiguously that `<T,>() => {}` is an arrow function with a type parameter declaration because the parser has encoded that in the AST.

The formatter then decides how to print the code - if it prints the type parameter on a single line, its trivial to also insert the comma.

bradzacher:
> Only report no-unnecessary-type-constraint for <T extends any> in files which do not require disambiguation (everything other than mts, cts and JSX language variant files)

This is not correct, no.

The rule should **_always_** report on the unnecessary `extends any` and the fix is to remove it in `.ts` files, and replace it with a comma in `.tsx/cts/mts` files.

WoodyWoodsta:
> Formatters are powered by the AST

@bradzacher Ok you confirm what I thought. So if I understand correctly, there is actually nothing to do here in `typscript-eslint`? 😆

bradzacher:
My [last comment](https://github.com/typescript-eslint/typescript-eslint/pull/6795#issuecomment-1493443420) describes the change we need to make to the rule - we need to fix to `<T,>` in `mts/cts` files.

WoodyWoodsta:
@bradzacher Hopefully I've now understood. Please see https://github.com/typescript-eslint/typescript-eslint/pull/6795/commits/6bfccc6f6987b8552dd6c16a659cc09dae7dad92

packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts
```diff
@@ -1,3 +1,5 @@
+import { extname } from 'node:path';
```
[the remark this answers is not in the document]
WoodyWoodsta: Or I can just change the import to `'path'` if that's what is used throughout the project. Looks like it is.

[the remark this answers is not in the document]
WoodyWoodsta: @armano2 https://github.com/typescript-eslint/typescript-eslint/pull/6795/commits/ff2a40310de43f1a96dc195ebab38006e67c7c25

JoshuaKGoldberg:
@bradzacher is this still something you have time to review?

bradzacher (dismissed):
this LGTM!

bradzacher (approved):

PR #6816, opened 2023-04-02
feat: remove `RuleTester` in `/utils` in favour of the new `/rule-tester` package

bradzacher:
BREAKING CHANGE:
Removes `RuleTester` from the `@typescript-eslint/utils/eslint-utils` package in favour of the new `@typescript-eslint/rule-tester` package.

## PR Checklist

## Overview

Migrates our codebase to use the package added in #6777 as a thorough test of the fork to make sure it works as expected.
- Most changes were just to the file imports
- Some tests I was able to remove the `filename` config. The old rule tester did `file.tsx` for JSX files - which actually didn't work at all because TS wouldn't pick up both `file.ts` and `file.tsx`.
- I removed the `batchedSingleLineTests` util from some places - but it's probably worth removing it entirely in future (once I add snapshot tests we shouldn't need it)


JoshuaKGoldberg:
For future reviewing reference...

```js
[...document.querySelectorAll(".file-header:has(.rgh-seen--11765709499)")].filter(e => e.textContent.includes("packages/eslint-plugin/tests/rules/") && e.textContent.includes(".test.ts")).forEach(e => e.querySelector("form label").click())
```

packages/utils/src/eslint-utils/rule-tester/RuleTester.ts
```diff
@@ -63,6 +63,9 @@ function isDescribeWithSkip(
   );
 }
 
+/**
+ * @deprecated use `@typescript-eslint/rule-tester` instead
```
JoshuaKGoldberg, line 67: Can you also file a delection tracking issue & link to it here? I'm guessing it'll be a v7 breaking change.

JoshuaKGoldberg (approved):
🎊

bradzacher:
re-requesting reviews because I've made this a breaking change.
This completely removes `RuleTester` from `/utils` now - will update the blog post to point users at the new package

packages/eslint-plugin/tests/rules/no-unsafe-assignment.test.ts
```diff
@@ -269,78 +268,84 @@ const x: Set<Set<Set<string>>> = new Set<Set<Set<any>>>();
...
-          column: 12,
-          endColumn: 25,
-        },
-        {
-          messageId: 'unsafeArraySpread',
-          line: 3,
-          column: 12,
-          endColumn: 28,
-        },
-      ],
```
armano2, line 290: is there a reason why we are not preserving loine/col/endCol ? 

its a good practice to add those to validate if error message is reported in correct place

bradzacher: because I couldn't be bothered remapping each one before and after the prettier format.
it already took long enough to extract those ones as it is.

soon™️ I'll implement snapshot testing in the fork which will snapshot an error underline and will make the line/col assertions unnecessary cruft anyways.

packages/website/blog/2023-03-13-announcing-typescript-eslint-v6-beta.md
```diff
@@ -275,9 +276,47 @@ For more information, see:
...
+  - You can control the default filenames by passing `defaultFilenames` to the `RuleTester` constructor.
+
+New features:
+
+- `skip: boolean` - the inverse option of `only: boolean`. When `true` we will use your test framework's test skip functionality (`it.skip`) to mark the test as skipped. This is useful during development as it enables you to control which tests run without needing to comment blocks out.
+- Dependency version filtering. It's useful to test your rule against multiple versions of your dependencies to ensure it doesn't break on older versions. However in some cases certain tests will not work on older versions of some dependencies due to features that didn't exist until recently - for example a test might use newer syntax that didn't exist in an older version of TypeScript. Our rule tester includes options that allow you to declare the allowed version ranges for a test so that it is automatically skipped when necessary.
+
+For more information on the package, [see the package documentation](/architecture/rule-tester).
```
JoshuaKGoldberg, line 316: [Accessibility] Making this link a bit more comprehensible on its own...

```suggestion
For more information on the package, [see the `rule-tester` package documentation](/architecture/rule-tester).
```

JoshuaKGoldberg (approved):
Cute drawn Gengar-like creature dancing to the caption 'heck yes'

PR #6836, opened 2023-04-03
chore: Switch to ESLint flat config

nzakas:
## PR Checklist

:)

## Overview

In testing the new flat config, I thought a good way to test typescript-eslint would be to convert the project itself to use flat config. I've got most of the way there, but I'm afraid my knowledge of how to configure Nx is pretty basic and so I can't quite get the linting to run. Happy to continue iterating on this if someone can point me in the right direction.


bradzacher:
thanks heaps for doing this! Even went and fixed up the nx config for us!
will have a proper look at it shortly.

quick question from a glance: `FlatCompat` what are the steps the ecosystem needs to take in order for us to migrate away from the compat class? Is the big ticket item defining flat config compatible configurations from our plugin?

eslint.config.js
```diff
@@ -0,0 +1,422 @@
...
+          glob: 'Infinity',
+        },
+      },
+    },
+    rules: {
+      // make sure we're not leveraging any deprecated APIs
+      'deprecation/deprecation': 'error',
+
+      //
+      // our plugin :D
```
JoshuaKGoldberg, line 96: [Praise] Thank you for keeping the comments. This one has always made me smile 😄

yarn.lock
```diff
@@ -6826,6 +6846,15 @@ espree@^9.5.0:
...
     eslint-visitor-keys "^3.3.0"
 
+espree@^9.5.1:
+  version "9.5.1"
+  resolved "https://registry.yarnpkg.com/espree/-/espree-9.5.1.tgz#4f26a4d5f18905bf4f2e0bd99002aab807e96dd4"
+  integrity sha512-5yxtHSZXRSW5pvv3hAlXM5+/Oswi1AUFqBmbibKb5s6bp3rGIDkyXU6xCoyuuLhijr4SFwPrXRoZjz0AZDN9tg==
+  dependencies:
+    acorn "^8.8.0"
+    acorn-jsx "^5.3.2"
+    eslint-visitor-keys "^3.4.0"
```
JoshuaKGoldberg, line 6856: [Nit] Could probably dedupe this with `espree@9.5.0`. But we also have so many changes in the v6 branch that it's fine to leave as-is IMO. Not a blocker at all, not even a request for changes.

nzakas:
> quick question from a glance: FlatCompat what are the steps the ecosystem needs to take in order for us to migrate away from the compat class? Is the big ticket item defining flat config compatible configurations from our plugin?

Yes. So basically my thinking was first, get the repo's config files swapped over to `eslint.config.js` and make sure everything works with `FlatCompat` as a sanity check. Then, the next step would be to create flat config versions of the other configs you include the plugin. I'm also happy to help with that once we can get the repo linted properly using `eslint.config.js`.

nzakas:
So this is where I need help. I just don't know enough about Nx to understand what the problem is with the lint task.

JoshuaKGoldberg:
I'll defer to @JamesHenry as the Nx expert in the org, but I _suspect_ that Nx doesn't yet recognize `eslint.config.js`. You may need to have the Nx config specify the path explicitly: https://nx.dev/packages/linter/executors/eslint#eslintconfig. (found via https://github.com/nrwl/nx/issues/8982)

(I'm holding time on my calendar to try this out tomorrow morning, if nobody else has gotten it fixed by then)

JamesHenry:
Hi @nzakas, thanks again for this!

@JoshuaKGoldberg is exactly right, for now in all project.json files with a "lint" target you will need to add the `eslintConfig` property.

Now that you have confirmed that flat config is feature complete we will add automatic detection for this in an upcoming Nx version but for now you'll need to configure it, e.g for the website project:

image

There seem to be some remaining issues, however, and I wanted to hand it back to you before I spend too long spinning my wheels on stuff I'm not familiar with.

2 issues that seem like they have easy solutions:

1. The references to `require('@typescript-eslint')` need to be updated because that isn't a module that can be resolved (I think you used the value from the old config directly which was relying on ESLint's magical resolution of the eslint-plugin package under that namespace). The requires should be:

image

2. `eslintrc.extends` is being given arrays here but seems to be expecting multiple strings so it throws

At least one remaining issue that I can't easily understand from the schema validation error message:

image

I wonder if anything can be done to build on top of ajv and improve that error message?

nzakas:
@JoshuaKGoldberg @JamesHenry great, I'll make those changes and see how far I can get.

nzakas:
Okay, so the root level problem I'm dealing with right now is that because Nx uses `--config` with the specified configuration file, that automatically is treated as eslintrc instead of flat config. In order to force ESLint use flat config with `--config`, we need to set the environment variable `ESLINT_USE_FLAT_CONFIG=true`. I've tried a few different things to set an environment variable but it seems like however Nx is using ESLint isn't respecting it. Any ideas?

JamesHenry:
@nzakas Nx doesn't use the CLI `--config` it uses the `ESLint` class API. If I add a console.log to `node_modules/@nrwl/linter/src/executors/eslint/lint.impl.js` I can see the environment variable being respected, is that definitely enough? Or does Nx need to conditionally use the new class in its executor implementation?

JamesHenry:
The exact class usage is here: `node_modules/@nrwl/linter/src/executors/eslint/utility/eslint-utils.js`

nzakas:
@JamesHenry ah! Yes, that makes a big difference. It's the CLI that recognizes the environment variable and then switches between `ESLint` and `FlatESLint` classes, so the Nx implementation would have to do the same (assuming it wants to support flat config).

JamesHenry:
@nzakas cool, I'll submit a PR to Nx next week when I'm back from vacation to allow it to:

A) automatically detect `eslint.config.js` files (once it is the default format we will also switch Nx to generating/updating them like it does today with .eslintrc.json files)

B) Conditionally change the implementation class depending on the config used/the presence of the environment variable

Until then I've mark this one as blocked, I'll circle back as soon as there is an Nx version for us to apply here

nzakas:
Awesome, thanks so much. Enjoy your vacation!

JamesHenry:
Just a follow up on this, now that we (Nx) are a larger company with more resources to work on product, we have a more robust planning cycle. I have added supporting flat config as an item for our next iteration which starts in a couple of weeks. I should hopefully be working on it personally.

I will update this PR once there is a version to try it out with.

Thanks again @nzakas!

nzakas:
Sure thing. I'll stay tuned in to this PR.

JamesHenry:
Hi @nzakas I'm sorry it took me a while to loop back to this. I added the support for Flat Config to Nx in the last cycle and have updated your branch here to contain those changes.

However, when I run the linting for a particular project, it seems that the root config is currently invalid based on the new structure?

image

nzakas:
Sorry about that. I think the config file was so long that I just missed that section. Indeed, there are no overrides in flat config, so I just flattened that section out and pushed the new config file. (I'm on my way out so didn't have time to verify it 100% works but at least it's the correct format.)

JamesHenry:
@nzakas it did fail to validate again just FYI

GitMurf:
Hi team! Nice work so far. I have a couple projects coming up that I am proposing and they are requiring some clarity on the transition to the new flat configs. Is this the PR to follow for the transition of typescript-eslint to using the new flat config file format `eslint.config.js`? If not, is there a better place to "follow" / comment? Thanks!

JoshuaKGoldberg:
Just a heads up, we've also got #7273 in-progress. Might be relevant here.

JoshuaKGoldberg:
@nzakas now that https://github.com/eslint/eslint/pull/17640 is merged, are there any changes you want to make here?

nzakas:
That PR doesn't really apply here. This PR's purpose was to switch the repo to use flat config. Updating the plugin for flat config would be separate.

JoshuaKGoldberg:
Got it - what I really mean is, is there anything on your end you plan on doing to this PR? Or is it ready for us to take over?

JamesHenry:
I think a win for everybody here would be if we abandon this PR and instead run the Nx automated migration for migrating from .eslintrc to flag config:

- it saves Nicholas's time
- it tests out the migration on a real world repo and helps us find any remaining rough edges

WDYT?

karlhorky:
@JamesHenry I guess you mean this Nx automated migration to flat config?

- https://nx.dev/recipes/tips-n-tricks/flat-config

Looks pretty cool 👀 didn't know there was a tool to automate this! I guess it could work for other configs elsewhere too... like [our `eslint-config-upleveled` shared config](https://github.com/upleveled/eslint-config-upleveled) 🤔 Added to our issue tracking the flat config: https://github.com/upleveled/eslint-config-upleveled/issues/160

JamesHenry:
@karlhorky to be clear the automation is not for plugin authors convert their plugin implementations to flat config, it is for users to convert their setups.

This PR relates to typescript-eslint’s own usage of eslint

karlhorky:
Yeah, maybe it still applies to shared configs like the one I shared above? (it's not a plugin, although we also have one of those)

Issue #6839, opened 2023-04-04
Enhancement: v6 `typeMatchesSpecifier` add ability to match via a pattern


RebeccaStevens:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#supported-rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Relevant Package

utils

### My proposal is suitable for this project

- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Description

Add ability to provide a regex expression of the type name instead of a string that has to match exactly. 
Additionally, allow for a glob pattern to be given as the "path" for the `FileSpecifier`.

### Fail

```typescript
// N/A
```

### Pass

```typescript
// Match all local types prefix with mutable
typeMatchesSpecifier(
  type,
  {
    from: "file",
    name: /^(M|m)utable.+/u,
  },
  program);

// Match all local types defined in a test or tests director.
typeMatchesSpecifier(
  type,
  {
    from: "file",
    name: /^.+/u,
    path: "**/tests?/**"
  },
  program);

// Match all build in "Readonly" types
typeMatchesSpecifier(
  type,
  {
    from: "lib",
    name: /^Readonly.+/u,
  },
  program);

// Match all types in all @foo/* packages
typeMatchesSpecifier(
  type,
  {
    from: "package",
    name: /^.+/u,
    package: /^@foo\/.+/u,
  },
  program);
```

### Additional Info

New specifier definitions will look something like this:

```ts
interface FileSpecifier {
  from: 'file';
  name: string | RegExp | (string | RegExp)[];
  path?: string;
}

interface LibSpecifier {
  from: 'lib';
  name: string | RegExp | (string | RegExp)[];
}

interface PackageSpecifier {
  from: 'package';
  name: string | RegExp | (string | RegExp)[];
  package: string | RegExp;
}
```

RebeccaStevens:
@marekdedic

bradzacher:
Globs are difficult because it means we have to add a dependency on a glob library and do fuzzy matches.

[...]

PR #6860, opened 2023-04-07
fix(type-utils): allow `FileSpecifier` to exclude paths

### LINKED ISSUE #6819: Bug: v6 `typeMatchesSpecifier` `FileSpecifier` checks in node modules

### Relevant Package

utils

### Playground Link

_No response_

### Repro Code

```typescript
typeMatchesSpecifier(
  type,
  {
    from: "file",
    name: "RegExp",
  },
  parserServices.program)
```

### ESLint Config

_No response_

### tsconfig

_No response_

### Expected Result

It shouldn't find this type (as I haven't defined it).

### Actual Result

I finds it in "node_modules/.pnpm/typescript@5.0.3/node_modules/typescript/lib/lib.es5.d.ts".

### Additional Info

I would expect a `FileSpecifier` to ignore node module files by default.
`typescript/lib/*.d.ts` files should only match with a `LibSpecifier` and other files in node module to only match with a `PackageSpecifier`.

Maybe `FileSpecifier` needs an extra property like `ignorePath` or `ignoreGlob` that defaults to `"node_modules"`.
Or alternatively, `path` could be made a glob that defaults to selecting files not in `"node_modules"`.

### Versions

| `@typescript-eslint/eslint-plugin`     | `6.0.0-alpha.106` |
| `@typescript-eslint/parser`            | `6.0.0-alpha.106` |
| `@typescript-eslint/scope-manager`     | `6.0.0-alpha.106` |
| `@typescript-eslint/typescript-estree` | `6.0.0-alpha.106` |
| `@typescript-eslint/type-utils`        | `6.0.0-alpha.106` |
| `@typescript-eslint/utils`             | `6.0.0-alpha.106` |
| `TypeScript`                           | `5.0.3` |
| `ESLint`                               | `8.36.0` |
| `node`                                 | `18.15.0` |

RebeccaStevens:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6819

## Overview

Adds `excludePaths` option to `FileSpecifier` (has no effect if a `path` is given).
This new option defaults to `["node_modules"]`.

### DIFF
--- a/packages/type-utils/src/TypeOrValueSpecifier.ts
+++ b/packages/type-utils/src/TypeOrValueSpecifier.ts
@@ -1,11 +1,19 @@
 import path from 'path';
 import type * as ts from 'typescript';
 
-interface FileSpecifier {
+type FileSpecifier = {
   from: 'file';
   name: string | string[];
-  path?: string;
-}
+} & (
+  | {
+      path?: string;
+      excludePaths?: undefined;
+    }
+  | {
+      path?: undefined;
+      excludePaths?: string | string[]; // defaults to `["node_modules"]`.
+    }
+);
 
 interface LibSpecifier {
   from: 'lib';
@@ -129,14 +137,26 @@ function specifierNameMatches(type: ts.Type, name: string | string[]): boolean {
 
 function typeDeclaredInFile(
   relativePath: string | undefined,
+  excludePaths: string | string[] | undefined,
   declarationFiles: ts.SourceFile[],
   program: ts.Program,
 ): boolean {
   if (relativePath === undefined) {
     const cwd = program.getCurrentDirectory().toLowerCase();
-    return declarationFiles.some(declaration =>
-      declaration.fileName.toLowerCase().startsWith(cwd),
-    );
+    const excludePathsArray =
+      excludePaths === undefined
+        ? ['node_modules']
+        : Array.isArray(excludePaths)
+        ? excludePaths
+        : [excludePaths];
+
+    return declarationFiles.some(declaration => {
+      const fileName = declaration.fileName.toLowerCase();
+      return (
+        fileName.startsWith(cwd) &&
+        excludePathsArray.every(p => !fileName.startsWith(path.join(cwd, p)))
+      );
+    });
   }
   const absolutePath = path
     .join(program.getCurrentDirectory(), relativePath)
@@ -164,7 +184,12 @@ export function typeMatchesSpecifier(
       ?.map(declaration => declaration.getSourceFile()) ?? [];
   switch (specifier.from) {
     case 'file':
-      return typeDeclaredInFile(specifier.path, declarationFiles, program);
+      return typeDeclaredInFile(
+        specifier.path,
+        specifier.excludePaths,
+        declarationFiles,
+        program,
+      );
     case 'lib':
       return declarationFiles.some(declaration =>
         program.isSourceFileDefaultLibrary(declaration),


packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -129,14 +137,26 @@ function specifierNameMatches(type: ts.Type, name: string | string[]): boolean {
...
+    return declarationFiles.some(declaration => {
+      const fileName = declaration.fileName.toLowerCase();
+      return (
+        fileName.startsWith(cwd) &&
+        excludePathsArray.every(p => !fileName.startsWith(path.join(cwd, p)))
+      );
+    });
   }
   const absolutePath = path
     .join(program.getCurrentDirectory(), relativePath)
```
marekdedic, line 162: Just to point this out - you are **not** disallowing files in the `excludesPath` if they are explicitly reference in `path`. I actually think they should be disallowed as well - that would make the specifier more consistent.

See https://github.com/typescript-eslint/typescript-eslint/issues/6819#issuecomment-1500269177

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -1,11 +1,19 @@
...
+} & (
+  | {
+      path?: string;
+      excludePaths?: undefined;
+    }
+  | {
+      path?: undefined;
+      excludePaths?: string | string[]; // defaults to `["node_modules"]`.
+    }
+);
```
marekdedic, line 16: I have to say I'm not a fan of the new configuration option - what would the use-case be? I can only imagine if you want to specify a local file but prevent the same type defined in another local from being used - but then, why not just use the already-present `path` property to specify the exact type you want?

marekdedic (changes requested):
Thanks for the PR, you truly beat me to it by just a few minutes :D 

As I wrote in the comments, I'm not a fan of the paths being configurable and would just hard-code `node_modules` there.

However, if you do decide to go ahead with the `excludePaths` option, `typeOrValueSpecifierSchema` needs to be modified as well :)

packages/type-utils/src/TypeOrValueSpecifier.ts
marekdedic: Looking closer I see that the `FileSpecifier` type disallows adding both `path` and `excludePaths` at the same time - so this wouldn't be such an issue - except for the situation where you leave `excludePaths` empty (so it defaults to `node_modules`) and add `path: "node_modules/..."` - then I think that would still actually match, right?

packages/type-utils/src/TypeOrValueSpecifier.ts
RebeccaStevens: The main use case I thought of and thus why I included it, is to support yarn 2 and 3. Yarn doesn't use the `node_modules` folder but rather is own `.yarn` path.

We could hard code this in too, but yarn allows it to be changed. We could maybe read the config file...

But what if there's another package manager out there doing its own thing too. When bower was still around it used `bower_components` I believe.

RebeccaStevens: An alternative would be to allow the `path` option to be a prefix. In which case specifically `src` will stop it from looking in `node_modules`.

marekdedic: Ugh, yarn... Yeah, that makes sense. But I think we actually *need* to read this from the config file anyway if we are to support non-standard locations - the `PackageSpecifier` currently hard-codes `node_modules` as well :( So doing that would actually be my preferred solution...

A simpler one would be to just hard-code `node_modules` and `.yarn` and if the user uses custom location, they're on their own :/

packages/type-utils/src/TypeOrValueSpecifier.ts
RebeccaStevens: `excludedPaths` is only used when `path` is `undefined` so that should be an issue.

packages/type-utils/src/TypeOrValueSpecifier.ts
marekdedic: Or #6861 could actually help - I have written that issue with `PackageSpecifier` in mind, but it could be repurposed for `FileSpecifier` as well - the type just should **not** be found in the map...

RebeccaStevens: Sounds promising.

JoshuaKGoldberg:
I'm going to wait on discussion in #6839 - it might impact the config option here.

JoshuaKGoldberg:

JoshuaKGoldberg:
Sorry for the delay - #6839 is accepting PRs so this is unblocked!

packages/type-utils/src/TypeOrValueSpecifier.ts
JoshuaKGoldberg, line 16: Starting a new conversation here: why make `excludePaths` and `path` mutually exclusive? What if someone wants to provide both? E.g.:

```js
{
  path: "examples",
  excludePaths: ["examples/*/node_modules"]
}
```

Proposal: can we switch this back to a straightforward interface?

```ts
interface FileSpecifier {
  from: 'file';
  name: string | string[];
  path?: string;
  excludePaths?: string | string[]; // defaults to `["node_modules"]`.
}
```

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -1,11 +1,19 @@
...
-  path?: string;
-}
+} & (
+  | {
+      path?: string;
+      excludePaths?: undefined;
+    }
+  | {
+      path?: undefined;
+      excludePaths?: string | string[]; // defaults to `["node_modules"]`.
```
JoshuaKGoldberg, line 14: I also wonder if we should call this `pathExcludes`, to visually lump it in with `path`... 🤔

JoshuaKGoldberg (changes requested):
Sorry for the delay here! I think the changes generally make sense, but I'm leaning towards keeping it less complex by not making `excludePaths` and `path` mutually exclusive. What do you think?

JoshuaKGoldberg:
👋 ping @RebeccaStevens, just checking in - is this still something you have time for? No worries if not! I just don't want to leave it hanging.

JoshuaKGoldberg:
Closing this PR as it's been stale for a while without activity. Feel free to reopen @RebeccaStevens if you have time - but no worries if not!

If anybody wants to drive it forward, please do post your own PR - and if you use this as a start, consider adding a co-author at the end of your PR description. Thanks! 😊

PR #6873, opened 2023-04-09
fix(eslint-plugin): [no-unused-vars] check if any variable definition is exported

### LINKED ISSUE #6188: Bug: [no-unused-vars] false positive if name is type/value merged and exported

### Playground Link

https://typescript-eslint.io/play/#ts=4.9.3&jsx=true&sourceType=module&showAST=es&code=C4TwDgpgBAZg9nKBeKBvAvgbgFAQB5hwBOwUAxnAHYDOp8iKAFAIZEDmAXLAgJTIB8abFChEIwAK5FKUVmxzpsQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tidmUQAmtAG4BDaKgyRE0aB2iRwYAL4hlQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
type foo = {};
export const foo = (arg: foo) => {
  return arg;
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

no error because the type `foo` is used and the value `foo` is exported

### Actual Result

```
'foo' is assigned a value but never used. 2:14 - 2:17
```

### Additional Info

_No response_

Maxim-Mazurok:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6188

## Overview

As per https://github.com/typescript-eslint/typescript-eslint/issues/6188#issuecomment-1489530919 suggestion, instead of checking the first declaration of a variable to determine if it's exported or not, I'm checking all declarations. Test included.


packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
```diff
@@ -1072,6 +1072,12 @@ export class Foo {
         typescript: '4.4',
       },
     },
+    `
+interface Foo {
+  bar: string;
+}
+export const Foo = 'bar';
+    `,
```
JoshuaKGoldberg, line 1080: [Testing] I peeked through the existing tests and couldn't find one that verifies this still gives an error when the variable _isn't_ exported. Am I missing it? If not, could you add one in please? (basically this without the `export`)

JoshuaKGoldberg (changes requested):
Super, thanks! Implementation looks great 🙌  - just requesting changes on a teeny bit more testing (unless I missed it already existing?).

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
Maxim-Mazurok: Ah, good spot.
Even though we already have tests for 
```
interface Foo {
  bar: string;
  baz: Foo['bar'];
}
```
and 
```
interface Foo {
  a: string;
}
interface Foo {
  b: Foo;
}
```
there were no tests for interface + const so I've added one. 

Also since I'm here, added a case for when `export const` is before `interface` to ensure that we are checking all the definitions and not only the last one.

JoshuaKGoldberg: Ah great, thanks!

packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -423,9 +423,7 @@ function isMergableExported(variable: TSESLint.Scope.Variable): boolean {
  * @returns True if the variable is exported, false if not.
  */
 function isExported(variable: TSESLint.Scope.Variable): boolean {
-  const definition = variable.defs[0];
-
-  if (definition) {
+  const exportedDefinition = variable.defs.find(definition => {
```
JoshuaKGoldberg, line 426: [Nitpick] This can be ever so slightly simplified:

```suggestion
  return variable.defs.some(definition => {
```

Definitely not a blocker for review. I can make this change since it's the only comment I have 🙂 otherwise this PR looks great!

JoshuaKGoldberg (dismissed):
Super, thanks @Maxim-Mazurok! This looks clean and great. And the extra tests are always lovely, much appreciated!

Aside - my apologies for taking so long without getting back to you. I got sidetracked during the v6 work and dropped priority on the PR queue. But I should have let you know (or, honestly, just reviewed this PR - it's a pretty straightforwardly great one). Sorry about that!

JoshuaKGoldberg (approved):

PR #6876, opened 2023-04-09
feat(eslint-plugin): [ban-types] add NonNullable suggestion and allow custom suggestions

### LINKED ISSUE #6486: Enhancement: [ban-types] Suggest `NonNullable<unknown>` as an alternative for `{}` and `Object`

https://typescript-eslint.io/rules/ban-types/

### Description

The rule forbids the use of the type `{}` (and its synonym `Object`), which means "any non-nullish value", and gives some helpful suggestions on what to use instead, depending on what the developer actually might have meant.

However, there seems to be no suggestion (outside [disabling the rule](https://github.com/typescript-eslint/typescript-eslint/issues/5542#issuecomment-1227114340)) for what to use when the intended meaning _really_ is "any non-nullish value". The suggested alternative for "any object", namely `object`, doesn't include primitive types such as `number`, so it's not an equivalent replacement in this case.

Rather than bypassing or disabling the rule, and leaving the easily misunderstood `{}` in place, I think it would be useful to suggest an explicit alternative for this case as well; something along the lines of
```
- If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead.
```
(placed _after_ all the other suggestions).

TypeScript includes the [`NonNullable` utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype) since 2.8, so I don't think there shouldn't be any compatibility issues.

PS: I wasn't sure if this issue should go under Documentation or Enhancement; since it changes the output of the rule I opted for the latter. My apologies if I got it wrong :)

### Fail

```typescript
/**
 * This is a generic combinator to add a fallback value to a function, to handle a null input.
 * However, we don't want to accidentally use it when the function _already_ handles null,
 * otherwise our fallback will clobber the original value.
 */
const withFallback: <Input extends {}, Output>( // This fails the rule and that's OK :)
  mapping: (input: Input) => Output,
  fallback: Output
) => (input: Input | null) => Output = (mapping, fallback) => (input) =>
  input === null ? fallback : mapping(input);

const plusOne: (input: number) => number
const plusOneWithFallback = withFallback(plusOne, 0); // This won't compile if we replace {} with object
```

### Pass

```typescript
const withFallback: <Input extends NonNullable<unknown>, Output>( // Same meaning, explicit, passes :)
  mapping: (input: Input) => Output,
  fallback: Output
) => (input: Input | null) => Output = (mapping, fallback) => (input) =>
  input === null ? fallback : mapping(input);

const plusOne: (input: number) => number
const plusOneWithFallback = withFallback(plusOne, 0); // This compiles
```

### Additional Info

_No response_

### LINKED ISSUE #6875: Enhancement: [ban-types] Add `suggest` option to types

https://typescript-eslint.io/rules/ban-types

### Description

Sometimes a banned type can support an autofix, but you'd prefer to pass an array of potential replacements rather than one autofix. 

Updating the `Types` option to:
```diff
{
  message: string;
  fixWith?: string;
+ suggest?: readonly string[];
}
```
would allow for replacements to be show as ESLint suggestions.

This would also allow the default `Object` and `{}` banned types to use suggestions.

### Fail

```typescript
function foo(x: {}) {}
```

### Pass

```typescript
function foo(x: object) {} // replaced by using ESLint suggestion
```

### Additional Info

_No response_

NotWoods:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6486 and fixes #6875
- [/] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22) (will revert if the issue is rejected)

## Overview

This PR adds `NonNullable<unknown>` to the list of alternatives for `Object` and `{}`, and introduces suggestions to automatically replace those types with `NonNullable<unknown>`.

### DIFF
--- a/packages/eslint-plugin/src/rules/ban-types.ts
+++ b/packages/eslint-plugin/src/rules/ban-types.ts
@@ -88,6 +88,7 @@ const defaultTypes: Types = {
       'The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.',
       '- If you want a type meaning "any object", you probably want `object` instead.',
       '- If you want a type meaning "any value", you probably want `unknown` instead.',
+      '- If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead.',
     ].join('\n'),
   },
   '{}': {
@@ -96,6 +97,7 @@ const defaultTypes: Types = {
       '- If you want a type meaning "any object", you probably want `object` instead.',
       '- If you want a type meaning "any value", you probably want `unknown` instead.',
       '- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.',
+      '- If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead.',
     ].join('\n'),
   },
 };
--- a/packages/eslint-plugin/src/rules/ban-types.ts
+++ b/packages/eslint-plugin/src/rules/ban-types.ts
@@ -11,6 +11,7 @@ type Types = Record<
   | {
       message: string;
       fixWith?: string;
+      suggest?: readonly string[];
     }
 >;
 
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (approved):
💯 looks great, thanks @NotWoods!

'yay' balloon with confetti

Issue #5692, opened 2022-09-26
Bug: [no-floating-promises] Incorrect handling of finally calls


aleclarson:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Abug+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/docs/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.3&sourceType=module&code=MYewdgzgLgBAJgUwLYgGIFczCgS3ALhgAocwAHdKQgIxBABsEBDMAShgF4A+GABQCcQSHBAQAeaP1IBzHh2KkKVGLQbM2nHgG8AUDBg4AZgvKV2u-fv4Io6fmD6DhogHTWIDAG4IiAcgAWCPT0IDAA7iD89HC+rADcejAAvjBBojAWlta29o5CIghuCABWCNh+ipThTBAwhkz0orEJ+kk6bTqIKBhYuOBEUPzoCKwuUIFgRETs3BmJoJBqLiHSfgCiYFAI1nAw4whgze2jhqQN9ACeUzPa8+AejMsgq74bWzt1Z8EXR0nxQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDN4OBDfMwDmtYtA4BbSshTooiaBOiRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
demoFunction(true).then(() => {
  console.log('Entered then');
}).finally(() => {
  console.log('Entered finally');
});
```

### ESLint Config

```javascript
module.exports = {
  parser: "@typescript-eslint/parser",
  rules: {
    "@typescript-eslint/no-floating-promises": ["error"],
  },
};
```

### Expected Result

There is no `.catch` call, and the rule should report it.

### Actual Result

Nothing reported

### Versions

| package                                | version |
[...]

PR #6881, opened 2023-04-11
fix(eslint-plugin): [no-floating-promises] false negative calling .then with second argument undefined

### LINKED ISSUE #6850: Bug: [no-floating-promises] False negative calling .then with second argument undefined

### Playground Link

https://typescript-eslint.io/play#ts=5.0.2&sourceType=module&code=PTAEGEHsCdoUwMYBcA2BPUAzFBDA5nnACYCwAUAArSQC2AlgM5wB08AVokgBQCUzSACzgA7Lr1ABeAHygA3qAC+PciFAAVAY1BbBcUAxw09gusLzbhoAFIBlZqACSSfUjooU2hqByWArsIEfIhRiUHZOOkhhZnIqWkYWcORefiFRcWk5RQAaUH8iOExTYh4gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDN4OBDfMwDmtYtA4BbSshTooiaBOiRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
Promise.reject().then(() => { }, undefined)
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-floating-promises": ["error"],
  },
};
```

### tsconfig

_No response_

### Expected Result

I expect that .then(f, undefined) should be flagged as an error the same as .then(f)

### Actual Result

No error

### Additional Info

_No response_

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6850 

## Overview

In no-floating-promises, the presence of an `onRejected` handler for `.then` or `.catch` is one of the valid ways of terminating an expression involving a promise. However, not all values supplied for `onRejected` are actually valid. Namely, anything that is not a function (looking at you, `null` and `undefined`) is simply ignored in EcmaScript. This PR adds validation to the rejection handler to verify that it is actually a function.

In short,
```
Promise.reject().then(() => {}) // previously was invalid, still is invalid
Promise.reject().then(() => {}, undefined) // previously was valid, now is invalid
```

In addition to that narrow issue, I have added the same validation for the `.catch` handler, since the same principles applies.
```
Promise.reject().catch() // previously was invalid, still is invalid
Promise.reject().catch(undefined) // previously was valid, now is invalid
```

In general, any `onRejected` argument to `.then` or `.catch` is now required to be callable.

There is a special case error message per discussion in the issue. 

I do have several things to bring to the attention of reviewers

1. I notice that this rule considers any "Promise-like" thenables, not just EcmaScript Promises. The claim that all rejection handlers are ignored if they are not callable is explicitly based on EcmaScript Promises, and may not be true for other Promise implementations. Is this a concern? Or should we go ahead with flagging errors assuming the ES implementation, despite the fact that, in principle, they could be false positives for an arbitrary thenable?

2. The rule at present never flags on a `.finally` that has a nonzero number of arguments. I don't actually understand why (is this intentional behavior or a bug??), since finally can make a resolved promise reject, but it can never make a rejected promise resolve. Because I don't understand the intention, and because the finally handler is not a rejection handler, I made no attempt to add validation to the handler in `.finally`.

3. I'm not totally happy with passing the result object up the recursion in `isHandledPromise`, but I'm not sure I know of a simpler way of passing the additional info required in order to special case the error message. Open to suggestions...

PS this is my first open source PR. Apologies in advance if I commit any major faux pas

### DIFF
--- a/packages/eslint-plugin/src/rules/no-floating-promises.ts
+++ b/packages/eslint-plugin/src/rules/no-floating-promises.ts
@@ -61,7 +61,7 @@ export default util.createRule<Options, MessageId>({
           },
           ignoreIIFE: {
             description:
-              'Whether to ignore async IIFEs (Immediately Invocated Function Expressions).',
+              'Whether to ignore async IIFEs (Immediately Invoked Function Expressions).',
             type: 'boolean',
           },
         },
--- a/packages/eslint-plugin/src/rules/no-floating-promises.ts
+++ b/packages/eslint-plugin/src/rules/no-floating-promises.ts
@@ -15,9 +15,22 @@ type Options = [
 
 type MessageId =
   | 'floating'
+  | 'floatingVoid'
+  | 'floatingUselessRejectionHandler'
+  | 'floatingUselessRejectionHandlerVoid'
   | 'floatingFixAwait'
   | 'floatingFixVoid'
-  | 'floatingVoid';
+  | 'floatingFixAwait';
+
+const messageBase =
+  'Promises must be awaited, end with a call to .catch, or end with a call to .then with a rejection handler.';
+
+const messageBaseVoid =
+  'Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler' +
+  ' or be explicitly marked as ignored with the `void` operator.';
+
+const messageRejectionHandler =
+  'A rejection handler that is not a function will be ignored.';
 
 export default util.createRule<Options, MessageId>({
   name: 'no-floating-promises',
@@ -30,13 +43,14 @@ export default util.createRule<Options, MessageId>({
     },
     hasSuggestions: true,
     messages: {
-      floating:
-        'Promises must be awaited, end with a call to .catch, or end with a call to .then with a rejection handler.',
+      floating: messageBase,
       floatingFixAwait: 'Add await operator.',
-      floatingVoid:
-        'Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler' +
-        ' or be explicitly marked as ignored with the `void` operator.',
+      floatingVoid: messageBaseVoid,
       floatingFixVoid: 'Add void operator to ignore.',
+      floatingUselessRejectionHandler:
+        messageBase + ' ' + messageRejectionHandler,
+      floatingUselessRejectionHandlerVoid:
+        messageBaseVoid + ' ' + messageRejectionHandler,
     },
     schema: [
       {
@@ -80,11 +94,18 @@ export default util.createRule<Options, MessageId>({
           expression = expression.expression;
         }
 
-        if (isUnhandledPromise(checker, expression)) {
+        const { isUnhandled, nonFunctionHandler } = isUnhandledPromise(
+          checker,
+          expression,
+        );
+
+        if (isUnhandled) {
           if (options.ignoreVoid) {
             context.report({
               node,
-              messageId: 'floatingVoid',
+              messageId: nonFunctionHandler
+                ? 'floatingUselessRejectionHandlerVoid'
+                : 'floatingVoid',
               suggest: [
                 {
                   messageId: 'floatingFixVoid',
@@ -110,7 +131,9 @@ export default util.createRule<Options, MessageId>({
           } else {
             context.report({
               node,
-              messageId: 'floating',
+              messageId: nonFunctionHandler
+                ? 'floatingUselessRejectionHandler'
+                : 'floating',
               suggest: [
                 {
                   messageId: 'floatingFixAwait',
@@ -168,16 +191,32 @@ export default util.createRule<Options, MessageId>({
       );
     }
 
+    function isValidRejectionHandler(rejectionHandler: TSESTree.Node): boolean {
+      // Note - when merged with main, getTypeAtLocation will be available as an
+      // instance method and these expressions can be chained.
+
+      return (
+        services.esTreeNodeToTSNodeMap
+          .get(rejectionHandler)
+          .getTypeAtLocation()
+          .getCallSignatures().length > 0
+      );
+    }
+
     function isUnhandledPromise(
       checker: ts.TypeChecker,
       node: TSESTree.Node,
-    ): boolean {
+    ): { isUnhandled: boolean; nonFunctionHandler?: boolean } {
       // First, check expressions whose resulting types may not be promise-like
       if (node.type === AST_NODE_TYPES.SequenceExpression) {
         // Any child in a comma expression could return a potentially unhandled
         // promise, so we check them all regardless of whether the final returned
         // value is promise-like.
-        return node.expressions.some(item => isUnhandledPromise(checker, item));
+        return (
+          node.expressions
+            .map(item => isUnhandledPromise(checker, item))
+            .find(result => result.isUnhandled) ?? { isUnhandled: false }
+        );
       }
 
       if (
@@ -192,24 +231,45 @@ export default util.createRule<Options, MessageId>({
 
       // Check the type. At this point it can't be unhandled if it isn't a promise
       if (!isPromiseLike(checker, services.esTreeNodeToTSNodeMap.get(node))) {
-        return false;
+        return { isUnhandled: false };
       }
 
       if (node.type === AST_NODE_TYPES.CallExpression) {
         // If the outer expression is a call, it must be either a `.then()` or
         // `.catch()` that handles the promise.
-        return (
-          !isPromiseCatchCallWithHandler(node) &&
-          !isPromiseThenCallWithRejectionHandler(node) &&
-          !isPromiseFinallyCallWithHandler(node)
-        );
+
+        const catchRejectionHandler = getRejectionHandlerFromCatchCall(node);
+        if (catchRejectionHandler) {
+          if (isValidRejectionHandler(catchRejectionHandler)) {
+            return { isUnhandled: false };
+          } else {
+            return { isUnhandled: true, nonFunctionHandler: true };
+          }
+        }
+
+        const thenRejectionHandler = getRejectionHandlerFromThenCall(node);
+        if (thenRejectionHandler) {
+          if (isValidRejectionHandler(thenRejectionHandler)) {
+            return { isUnhandled: false };
+          } else {
+            return { isUnhandled: true, nonFunctionHandler: true };
+          }
+        }
+
+        if (isPromiseFinallyCallWithHandler(node)) {
+          return { isUnhandled: false };
+        }
+
+        return { isUnhandled: true };
       } else if (node.type === AST_NODE_TYPES.ConditionalExpression) {
         // We must be getting the promise-like value from one of the branches of the
         // ternary. Check them directly.
-        return (
-          isUnhandledPromise(checker, node.alternate) ||
-          isUnhandledPromise(checker, node.consequent)
-        );
+        const alternateResult = isUnhandledPromise(checker, node.alternate);
+        if (alternateResult.isUnhandled) {
+          return alternateResult;
+        } else {
+          return isUnhandledPromise(checker, node.consequent);
+        }
       } else if (
         node.type === AST_NODE_TYPES.MemberExpression ||
         node.type === AST_NODE_TYPES.Identifier ||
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -293,26 +362,42 @@ function isFunctionParam(
   return false;
 }
 
-function isPromiseCatchCallWithHandler(
+function tryParseCatchCallWithhandler(
```
JoshuaKGoldberg, line 365: [Nitpick] 

```suggestion
function tryParseCatchCallWithHandler(
```

JoshuaKGoldberg, line 365: [Naming] Also on the name, my instinct for things named _"try*"_ is that they literally use a `try`/`catch` and/or may throw an error. This doesn't. Perhaps a more direct name would be `parseCatchCallWithHandler`?

(same for the "then" equivalent later on in the file)

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -1427,5 +1427,56 @@ async function foo() {
...
+        {
+          line: 12,
+          messageId: 'floatingUselessRejectionHandlerVoid',
+        },
+        {
+          line: 13,
+          messageId: 'floatingUselessRejectionHandlerVoid',
+        },
+      ],
+    },
```
JoshuaKGoldberg, line 1480: [Non-actionable] Heh, this reminded me to file #6887. No need to make any change here, as this is following the existing standard in the file.

packages/eslint-plugin/src/rules/no-floating-promises.ts
JoshuaKGoldberg, line 1: > I notice that this rule considers any "Promise-like" thenables, not just EcmaScript Promises. The claim that all rejection handlers are ignored if they are not callable is explicitly based on EcmaScript Promises, and may not be true for other Promise implementations. Is this a concern? Or should we go ahead with flagging errors assuming the ES implementation, despite the fact that, in principle, they could be false positives for an arbitrary thenable?

This is a really good point. We've had discussions about this before: that the vast majority of users only know "thenables" as Promises (e.g. #6113). For now this is fine. Eventually we may take action 🤷

JoshuaKGoldberg, line 1: > The rule at present never flags on a .finally that has a nonzero number of argument ...

Good point - and something to file a separate docs issue for. The fact that this isn't explicit in the docs means the docs are lacking!

JoshuaKGoldberg, line 1: > I'm not totally happy with passing the result object up the recursion in `isHandledPromise`

Haha I thought it was clever 😄. No cleaner alternatives come to mind. Let's do it!

JoshuaKGoldberg (changes requested):
Congrats on an excellent first open source PR! 🚀 

This generally looks good to me, except for a few small touchups - and the big `TODO`.

No major (or minor) faux pas 😄 - I appreciate you asking for clarification and adhering to existing styles really well!

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -194,24 +220,56 @@ export default util.createRule<Options, MessageId>({
...
         // If the outer expression is a call, it must be either a `.then()` or
         // `.catch()` that handles the promise.
-        return (
-          !isPromiseCatchCallWithHandler(node) &&
-          !isPromiseThenCallWithRejectionHandler(node) &&
-          !isPromiseFinallyCallWithHandler(node)
-        );
+        //
+        // TODO (Kirk) - This is not what the code does. It also allows a .finally
+        // so long as it has a handler.
```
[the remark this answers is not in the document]
kirkwaiblinger: Oh, whoops, I'll remove that.

I had made that note to myself while contemplating whether to include validation to the `.finally` handler, since I noticed that the comment there is out of sync with the code. (i.e. "if the outer expression is a call, it must be either a .then or .catch that handles the promise" is incomplete, since the code also permits .finally).

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger: Looks like this issue is under discussion here: https://github.com/typescript-eslint/typescript-eslint/issues/5692

Do you think that updating the ambiguity in the docs is a worthy issue on its own or is that implicitly handled by that issue?

JoshuaKGoldberg: Aha nice find - yes that issue is perfect on its own.

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -220,22 +279,33 @@ export default util.createRule<Options, MessageId>({
...
-        return true;
+        return { isUnhandled: true };
       } else if (node.type === AST_NODE_TYPES.LogicalExpression) {
-        return (
-          isUnhandledPromise(checker, node.left) ||
-          isUnhandledPromise(checker, node.right)
-        );
+        const leftResult = isUnhandledPromise(checker, node.left);
+        if (leftResult.isUnhandled) {
+          return leftResult;
```
JoshuaKGoldberg, line 286: [Testing] Oop sorry just noticed - codecov is reporting that this isn't covered by tests. We'll want everything to be tested. Either:

* this is code that can be hit, and a test should be added
* this is code that can't be hit, and should be removed

kirkwaiblinger:
Hi @JoshuaKGoldberg, it's been about 3 months since the last activity on this PR (which is the amount of time that [the contributing guide](https://typescript-eslint.io/contributing/pull-requests) suggests to wait before pinging pending reviewers, so I am gently pinging you. Would you be able to take a final pass on this? I believe I have made the minor changes requested.

JoshuaKGoldberg:
Yes! @kirkwaiblinger so sorry about the delay - you've done everything right (and are even being very kind about my blatantly ignoring you 😅). Very sorry to do that. Now that we're done with v6, I'm going through the PR queue.

But I shouldn't have let this go without at least an ACK - sorry about that. I'll try to make sure to not leave things open for this long again.

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -171,13 +197,17 @@ export default util.createRule<Options, MessageId>({
     function isUnhandledPromise(
       checker: ts.TypeChecker,
       node: TSESTree.Node,
-    ): boolean {
+    ): { isUnhandled: boolean; specialReason?: SpecialReasons } {
```
JoshuaKGoldberg, line 200: [Nitpick]

> `specialReason?: SpecialReasons`

Nit: I think I see why you made the `SpecialReasons` type (leaving the code open to more reasons in the future), but for now it's not really necessary to. We can always expand to more cases later. But in the meantime it's extra cognitive complexity for a feature we aren't 100% likely to need. I'd prefer having a simpler type if possible.

```suggestion
    ): { isUnhandled: boolean; nonFunctionHandler?: boolean } {
```

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -293,26 +363,42 @@ function isFunctionParam(
   return false;
 }
 
-function isPromiseCatchCallWithHandler(
+function parseCatchCallWithHandler(
   expression: TSESTree.CallExpression,
-): boolean {
-  return (
+): { onRejected: TSESTree.CallExpressionArgument } | undefined {
```
JoshuaKGoldberg, line 368: [Nitpick] Similar to https://github.com/typescript-eslint/typescript-eslint/pull/7092/files#r1262762950, we can simplify this type.

```suggestion
): TSESTree.CallExpressionArgument | undefined {
```

JoshuaKGoldberg (changes requested):
OK! Again so very sorry about taking so many months to get to this (and on your first PR even! 😭). 

The only blocking comment is adding a test case for the `ignoreVoid` option, to make sure that error ID logic is tested. Everything else is nitpicking and non-blocking - no worries if you ignore those comments 😄.

If you don't have time to address the changes, no worries - I'll set a reminder for myself to tackle them on Sunday or Monday morning before our next scheduled release. 

Thanks again - really great set of PRs here! ❤️ 

Tony the Tiger saying ggggrreeeaatt

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger: Changed, and changed function names correspondingly

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -1427,5 +1427,67 @@ async function foo() {
         },
       ],
     },
+    {
+      code: `
+const maybeCallable: string | (() => {});
+const definitelyCallable: () => {};
```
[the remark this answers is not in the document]
kirkwaiblinger: Yeah this confused me but i had seen it in the other test cases. Changed em all to use declare

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -194,24 +224,53 @@ export default util.createRule<Options, MessageId>({
...
-          !isPromiseFinallyCallWithHandler(node)
-        );
+
+        const catchCallResult = parseCatchCallWithHandler(node);
+        if (catchCallResult) {
+          const { onRejected } = catchCallResult;
+          const onRejectedAsTsNode =
+            parserServices.esTreeNodeToTSNodeMap.get(onRejected);
+
+          if (isValidRejectionHandler(checker, onRejectedAsTsNode)) {
```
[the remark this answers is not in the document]
kirkwaiblinger: done

kirkwaiblinger:
> Yes! @kirkwaiblinger so sorry about the delay - you've done everything right (and are even being very kind about my blatantly ignoring you 😅). Very sorry to do that. Now that we're done with v6, I'm going through the PR queue.
> 
> But I shouldn't have let this go without at least an ACK - sorry about that. I'll try to make sure to not leave things open for this long again.

No worries!

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger: Makes sense. Changed

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -110,7 +133,10 @@ export default util.createRule<Options, MessageId>({
           } else {
             context.report({
               node,
-              messageId: 'floating',
+              messageId:
+                specialReason === 'nonFunctionHandler'
+                  ? 'floatingUselessRejectionHandler'
+                  : 'floating',
```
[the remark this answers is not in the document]
kirkwaiblinger: Added several test cases. Hoping the next coverage report comes back clean!

(to clarify - the default is `ignoreVoid: true`, so the ones that were added are the ones with `ignoreVoid: false`)

kirkwaiblinger:
I'm not really sure what caused the CI failures, but neither appears to be directly related to my changes as far as I can tell. Looking for advice on this

packages/eslint-plugin/src/rules/no-floating-promises.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Awesome, thanks!

In case it's helpful, you can always view the coverage report locally. It's just [Jest with `coverage`](https://jestjs.io/docs/cli#--coverageboolean), outputting to `packages/*/coverage/lcov-report/index.html`.

JoshuaKGoldberg:
> I'm not really sure what caused the CI failures, but neither appears to be directly related to my changes as far as I can tell. Looking for advice on this

```plaintext
<--- Last few GCs --->
[2598:0x4d801c0]   237024 ms: Mark-sweep 2024.6 (2087.5) -> 2013.3 (2092.7) MB, 1363.3 / 0.0 ms  (average mu = 0.274, current mu = 0.103) allocation failure scavenge might not succeed
[2598:0x4d801c0]   238998 ms: Mark-sweep 2029.9 (2093.3) -> 2020.3 (2099.0) MB, 1822.2 / 0.0 ms  (average mu = 0.169, current mu = 0.077) allocation failure scavenge might not succeed
<--- JS stacktrace --->
 1: 0xb090e0 node::Abort() [/opt/hostedtoolcache/node/16.20.1/x64/bin/node]
 2: 0xa1b70e  [/opt/hostedtoolcache/node/16.20.1/x64/bin/node]
 3: 0xce1a20 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/opt/hostedtoolcache/node/16.20.1/x64/bin/node]
```

🙃 love it. We might have too many unit tests for the Jest+Node setup we're using.

Passed on re-run: https://github.com/typescript-eslint/typescript-eslint/actions/runs/5558453479/jobs/10164457558 & https://github.com/typescript-eslint/typescript-eslint/actions/runs/5558453479/jobs/10164509125. 🤷. If it keeps happening we can file an issue and take a look at reducing the memory footprint.

JoshuaKGoldberg (approved):
💯 thanks - excited to land this one!

PR #6885, opened 2023-04-11
fix(eslint-plugin): [consistent-type-assertions] wrap object return value with parentheses

### LINKED ISSUE #6884: Bug: [consistent-type-assertions] autofix breaks type assertion of return value

### Playground Link

https://typescript-eslint.io/play/#ts=5.0.3&sourceType=module&code=C4TwDgpgBAYg9nKBeKBvKAjAhgJwFxQB2ArgLYYQ4DcUAvgFADGchAzsFAB7JQAUAlMgB8UADzw4Q9NnxQArHSpA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6MgeyeUuX0Ra0ipWgENkyRNHyUuqDAG1IU6B2iQANJnDYc4ydNlMAykSToo4yDogBfALo7bIW0A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
type Foo = { bar: number; }
const x = () => <Foo>{ bar: 5 };
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/consistent-type-assertions": ["error", {
      "assertionStyle": "as",
    }],
  },
};
```

### tsconfig

_No response_

### Expected Result

I expected the following autofix.

```typescript
type Foo = { bar: number; }
const x = () => ({ bar: 5 } as Foo);
```

### Actual Result

The following autofix breaks TypeScript syntax.

```typescript
type Foo = { bar: number; }
const x = () => { bar: 5 } as Foo;
```

### Additional Info

Ref: #6641

dora1998:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6884 

## Overview

Wrap type assertion with parentheses if object literal type and used as return value of arrow function.


packages/eslint-plugin/src/rules/consistent-type-assertions.ts
```diff
@@ -132,16 +132,27 @@ export default util.createRule<Options, MessageIds>({
...
+
+                if (
+                  node.expression.type === AST_NODE_TYPES.ObjectExpression &&
+                  node.parent?.type ===
+                    AST_NODE_TYPES.ArrowFunctionExpression &&
+                  !util.isParenthesized(node.expression, sourceCode)
+                ) {
+                  yield fixer.insertTextBefore(node, '(');
+                  yield fixer.insertTextAfter(node, ')');
+                }
```
JoshuaKGoldberg, line 149: [Refactor] Aha yet another case of precedence around parenthesis and such getting in the way of auto-fixing! 😂. We've been collecting utilities around this common use case in `getWrappingFixer`. Could you check to see if this kind of logic is already handled there? And if not, add a generalized handler to it? I think knowing whether a one-line return is an object is something multiple rules are likely to need to know.

JoshuaKGoldberg (changes requested):
LGTM in general - and _love_ that you're fixing an issue you filed so quickly! 😍 

Just requesting changes on generalizing the fix if possible - since it's very likely to come up again. Thanks!

packages/eslint-plugin/src/rules/consistent-type-assertions.ts
dora1998: Thank you! I didn't know that 😂
I tried adding the logic whether a one-line return is an object to `getWrappingFixer` function, and refactored to use it in type assertion autofix.

packages/eslint-plugin/src/util/getWrappingFixer.ts
```diff
@@ -205,3 +215,21 @@ function isLeftHandSide(node: TSESTree.Node): boolean {
...
+/**
+ * Checks if a node's parent is arrow function expression and a inner node is object expression
+ */
+function isObjectExpressionInOneLineReturn(
+  node: TSESTree.Node,
+  innerNode: TSESTree.Node,
+): boolean {
+  if (
+    node.parent?.type === AST_NODE_TYPES.ArrowFunctionExpression &&
+    node.parent?.body === node &&
```
JoshuaKGoldberg, line 228: [Nitpick] If `node.parent?.type` is something defined, we know `node.parent` exists. ✨ 

```suggestion
    node.parent.body === node &&
```

JoshuaKGoldberg: Also -and this one is very optional, just posting in case you like it more, as I ever so slightly do 😄- you could simplify a bit to:

```ts
return (
  node.parent?.type === AST_NODE_TYPES.ArrowFunctionExpression &&
  node.parent.body === node &&
  innerNode.type === AST_NODE_TYPES.ObjectExpression
);
```

JoshuaKGoldberg (approved):
LGTM, great test cases! Glad to see a reuse of the existing shared code. Thanks! 🙌 

Since I haven't touched the `getWrappingFixer` util area much, marking as 1 approval and waiting for someone else to approve as well.

packages/eslint-plugin/src/util/getWrappingFixer.ts
JoshuaKGoldberg, line 228: Btw, I see the test coverage is dropping on this line... I _suspect_ it's fine, as a general-purpose thing. But it'd be nice not to have to run unused code. 🤔 

(delete the `node.parent?.type` line -> no tests fail)

dora1998: Oh I missed that case 😅 I added it!
I also want to add a case for `node.parent.body` line, but I cannot come up with...

packages/eslint-plugin/src/util/getWrappingFixer.ts
```diff
@@ -42,6 +42,13 @@ export function getWrappingFixer(
         code = `(${code})`;
       }
 
+      // check the inner node is used as return body of arrow function
+      if (isObjectExpressionInOneLineReturn(node, innerNode)) {
```
armano2, line 46: q: should we do `else if` here instead? as we do not want code to be wrapped twice

dora1998: Yeah that's true. I fixed it!

JoshuaKGoldberg:
cc @armano2 - is this still something you want to look at?

If not, no worries, I feel comfortable merging :)

bradzacher (commented):
big oof - the entire `packages/eslint-plugin/src/util/getWrappingFixer.ts` file probably needs to be rewritten because right now it uses its own hard-coded precedence logic, rather than using our [precedence utilities](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/util/getOperatorPrecedence.ts)

dora1998:
@bradzacher 
I added `getWrappedCode` util function that wrap text in parens if needed because of precedence.
but It's hard to replace `getWrapped` with `getWrappedCode` because now we have to know precedence of node and all innerNodes..., so I'd like to hear your opinion.

***

By the way, lint and unit test is failing in CI but successful in my local environment. Will merging the main branch fix this...?

JoshuaKGoldberg:
> By the way, lint and unit test is failing in CI but successful in my local environment. Will merging the main branch fix this...?

Yeah that should fix it. I don't like throwing extra work your way so I'll go ahead and merge now.

JoshuaKGoldberg (dismissed):
@armano2 do you still have time for this?

No worries if not, now that you've taken at least one look I feel comfortable re-reviewing and approving & merging if it's good to go.

dora1998:
@JoshuaKGoldberg Sorry for my late reply...
Thank you so much for taking over my PR ✨

JoshuaKGoldberg (approved):
Swell, thanks for such a thorough change! 🙌

polyzen:
> Does this not make `consistent-type-assertions` typed?

Seems to be the case. I'm using the following:
```
'plugin:@typescript-eslint/strict',
'plugin:@typescript-eslint/stylistic',
```
and getting the following error:
`Error: Error while loading rule '@typescript-eslint/consistent-type-assertions': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.`

JoshuaKGoldberg:
I can't reproduce any issues using the rule in versions 6.5.0 of the parser and plugin. See https://github.com/JoshuaKGoldberg/repros/tree/consistent-type-assertions-type-information.

By the way, if you think there's a bug in typescript-eslint, the right way to ask is to [file a new issue](https://github.com/typescript-eslint/typescript-eslint/issues/new/choose). The issue templates include helpful & necessary practices such as making sure you're on the latest version of all our packages. And it's much easier for us as maintainers to not lose track of things posted as issues.

Our contributing guide doesn't _explicitly_ ask not to post tangential questions on closed PRs, so I filed an issue to add those docs.

---

Just to emphasize: if you're experiencing an issue with typescript-eslint, please [file an issue with the appropriate template](https://github.com/typescript-eslint/typescript-eslint/issues/new/choose). Please. 🙂

Issue #6892, opened 2023-04-12
Bug: [array-type] Does not validate shape of config


domdomegg:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.0.3&sourceType=module&code=MYewdgzgLgBAtgTwMpQE4EswHMIC4bQbYDaAujALwxkDcAUKJLCFABYCmqKRO+AgqlQBDBAB5CmLAD5K1UvSA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AQ2mncNqNPRgA2pERcA9tEgAaKJ26FIAXXBgAviDVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
const myStrings: string[] = [];
const otherStrings: Array<string> = [];
```
```


### ESLint Config

```javascript
module.exports = {
  parser: "@typescript-eslint/parser",
  rules: {
    "@typescript-eslint/array-type": ["error", "array"]
  },
};
```

### Expected Result

I initially expected that I'd enforce array types (e.g. `string[]` rather than `Array<string>` everywhere), becuase I skimmed https://github.com/typescript-eslint/typescript-eslint/blob/HEAD/packages/eslint-plugin/docs/rules/array-type.md and didn't realise this should be an options object.

This is a user error, and should instead be configured:

```
    "@typescript-eslint/array-type": ["error", { default: "array" }]
```

That said, I then expected the rule would give me some kind of warning or error that my eslint config was invalid, so I could debug this quickly.

### Actual Result

The rule doesn't complain about its configuration, but instead highlights every array type it sees as an error - and tries to convert them to the other type in a loop.

### Additional Info

[...]

PR #6894, opened 2023-04-12
fix(eslint-plugin): fix schemas across several rules and add schema tests

### LINKED ISSUE #6852: Bug: [array-type] the config `simple-array` should be disallowed by the schema

### Playground Link

### Repro Code

```typescript
const b: number[] = [];
const d: Array<number> = [];
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
  }
};
```

### tsconfig

_No response_

### Expected Result

### Actual Result

### Additional Info

_No response_

domdomegg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6852, #6892

## Overview

This PR became a lot bigger than I was expecting, and was much more fiddly than I expected too. If we want, we can take d4ce93bacdc3a8afdb97fc73088ecf18343721df on its own, and the rest of the changes as a separate PR - but then it won't have tests to enforce that the schema stays fixed.

### Specific array-type problem

The ultimate problem was that we were using `prefixItems` rather than `items`. This was a problem because eslint uses an old version of ajv (v6) and and old version of the JSON schema spec (v4). See https://github.com/eslint/eslint/issues/13888#issuecomment-872591875 and https://github.com/eslint/eslint/blob/9d1b8fc60cc31f12618e58c10a2669506b7ce9bf/lib/shared/ajv.js#L12

As noted on https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation v4 of the spec didn't actually have `prefixItems`, and instead the `items` value was assumed to be prefixItems if it was an array of schemas:

> In Draft 4 - 2019-09, tuple validation was handled by an alternate form of the items keyword. When items was an array of schemas instead of a single schema, it behaved the way prefixItems behaves.

This in turn I think made the schema fail silently, and it just was okay validating any old random value you'd pass in.

I've fixed this by changing from `prefixItems` to `items`.

### Adding tests

Actually figuring this out was pretty difficult. To help me get there, I wrote some tests that validated the schemas generally against the same options as eslint.

This helped find other problems with array-type, which was that you could provide any arbitrary arguments as it did not have additionalProperties disabled.

Expanding this across all the rules, using their defaultOptions as sample config (which isn't a perfect solution, but at least is easy and works in every case except 2), managed to catch a few other badly configured rules. I've addressed these problems, applying the same fixes. I've left some comments in the PR to explain changes and why they were necessary too.

The tests are by no means perfect. They don't cover every edge case, and it's very possible to have nested objects with additional arbitrary properties that shouldn't be there. But they are a step in the right direction and what I had time to implement right now :)

### Note on adding ajv dependency

This is added as a devDependency, as its only used in tests. Additionally, as eslint uses the same version this doesn't actually add a real dependency, it just makes it explicit - so it doesn't slow down installs etc.


packages/eslint-plugin/src/rules/array-type.ts
```diff
@@ -124,7 +126,6 @@ export default util.createRule<Options, MessageIds>({
                 'The array type expected for readonly cases. If omitted, the value for `default` will be used.',
             },
           },
-          type: 'object',
```
domdomegg, line 127: Moved this up to the top for consistency with most of the other rules I saw.

packages/eslint-plugin/src/rules/lines-between-class-members.ts
```diff
@@ -9,16 +9,20 @@ const baseRule = getESLintCoreRule('lines-between-class-members');
...
 type MessageIds = util.InferMessageIdsTypeFromRule<typeof baseRule>;
 
-const schema = util.deepMerge(
-  { ...baseRule.meta.schema },
-  {
-    1: {
-      exceptAfterOverload: {
-        type: 'boolean',
-        default: true,
+const schema = Object.values(
```
domdomegg, line 12: Before, we had:

```ts
{ "0": { /* schema A */ }, "1": { /* schema B */ } }
```

AJV is actually not able to interpret this correctly, and no validation actually occurs.

This change fixes it to:

```ts
[{ /* schema A */ },  { /* schema B */ }]
```

Using Object.values here should be safe. The baseRule.meta.schema is an array itself, so will be spread in order (guaranteed by ECMA spec). deepMerge will merge the keys in the order of the first object. Object.values will take the values in order, so we'll get back the array in the right order.

packages/eslint-plugin/src/rules/no-parameter-properties.ts
```diff
@@ -51,7 +51,6 @@ export default util.createRule<Options, MessageIds>({
                 'public readonly',
               ],
             },
-            minItems: 1,
```
domdomegg, line 54: Removed this constraint so the default is able to be validated against the schema.

Alternative could be to exclude this from the schema validation list, or have an 'override' schema for testing against the schema.

I don't think this matters for actual usage, and I think it's not harmful: in fact it might be useful in some configs to be able to specify some allowed things by default, and then specify none to be allowed in an override config explicitly to be clear about why it's there.

packages/eslint-plugin/src/rules/no-restricted-imports.ts
```diff
@@ -27,18 +27,18 @@ const allowTypeImportsOptionSchema = {
 const schemaForMergeArrayOfStringsOrObjects = {
   items: {
     anyOf: [
-      {},
```
domdomegg, line 30: These seem incorrect and unnecessary. I think the intention was for this to merge with the base `anyOf` items, but deepMerge doesn't handle array items: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/utils/src/eslint-utils/deepMerge.ts

Although thinking about this now, I'm not certain the updated rule is 100% correct, would appreciate careful eyes on it. Might need to refactor how this extends from the base.

packages/eslint-plugin/src/rules/parameter-properties.ts
```diff
@@ -52,22 +52,21 @@ export default util.createRule<Options, MessageIds>({
...
         {
           type: 'object',
+          additionalProperties: false,
           properties: {
             allow: {
               type: 'array',
               items: {
                 $ref: '#/$defs/modifier',
               },
-              minItems: 1,
```
domdomegg, line 64: Same minItems reasoning as before

packages/eslint-plugin/tests/schemas.test.ts
```diff
@@ -0,0 +1,91 @@
+import Ajv from 'ajv';
+import type { JSONSchema4 } from 'json-schema';
+
+import eslintPlugin from '../src';
+import { RuleModule } from '@typescript-eslint/utils/src/ts-eslint';
+
+describe("Validating rule's schemas", () => {
+  // These two have defaults which cover multiple arguments that are incompatible
+  const excluded = ['semi', 'func-call-spacing'];
```
domdomegg, line 9: Instead of just excluding these, we could have overrides so we test the schema does validate something reasonable.

packages/eslint-plugin/src/rules/no-restricted-imports.ts
domdomegg: Have redone this. It's fairly horrible how it indexes into the baseRule - happy to take feedback on better approaches.

packages/eslint-plugin/tests/schemas.test.ts
domdomegg: Have done this, think it's a better approach.

packages/eslint-plugin/src/rules/array-type.ts
```diff
@@ -111,8 +111,10 @@ export default util.createRule<Options, MessageIds>({
           enum: ['array', 'generic', 'array-simple'],
         },
       },
-      prefixItems: [
+      items: [
```
JoshuaKGoldberg, line 114: Hmm, looks the [`generated-rule-docs.ts` file](https://github.com/typescript-eslint/typescript-eslint/blob/72989344bf459552e3a4865c5bcbf44cf5d02a50/packages/website/plugins/generated-rule-docs.ts) will need to be updated to handle these new practices. Right now it assumes `meta.schema.prefixItems` exists on all schemas with the `'array'` type.

https://github.com/typescript-eslint/typescript-eslint/blob/72989344bf459552e3a4865c5bcbf44cf5d02a50/packages/website/plugins/generated-rule-docs.ts#L268

```plaintext
Compiled with problems:X

ERROR in ../eslint-plugin/docs/rules/array-type.md

Module build failed (from ../../node_modules/@docusaurus/mdx-loader/lib/index.js):
TypeError: Cannot read properties of undefined (reading '0')
    at /Users/josh/repos/typescript-eslint/packages/website/plugins/generated-rule-docs.ts:268:61
    at wrapped (/Users/josh/repos/typescript-eslint/node_modules/trough/wrap.js:25:19)
```

packages/eslint-plugin/src/rules/lines-between-class-members.ts
JoshuaKGoldberg: Nifty trick to use `Object.values` to convert the string index style object to an array!

packages/eslint-plugin/tests/schemas.test.ts
```diff
@@ -0,0 +1,91 @@
...
+import eslintPlugin from '../src';
+
+describe("Validating rule's schemas", () => {
+  // These two have defaults which cover multiple arguments that are incompatible
+  const overrideOptions: Record<string, unknown> = {
+    semi: ['never'],
+    'func-call-spacing': ['never'],
+  };
+
+  for (const [ruleName, rule] of Object.entries(eslintPlugin.rules)) {
```
JoshuaKGoldberg, line 14: [Praise] I like this too! +1 to having tests for options being allowed.

packages/eslint-plugin/tests/schemas.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    test(`${ruleName} rejects arbitrary arguments`, () => {
+      if (isValid(rule, [{ 'arbitrary-schemas.test.ts': true }])) {
+        throw new Error(`Options succeeded validation for arbitrary options`);
+      }
+    });
+  }
+});
+
+describe('Individual rule schema validation', () => {
```
JoshuaKGoldberg, line 33: [Suggestion] Hmm, we normally put rule-specific things in each rule's specific test files. What do you think about moving these individual ones to closer to their rules?

JoshuaKGoldberg (changes requested):
Really digging this direction in general - especially with the added test coverage! Thanks for sending!

Gave a general / high-level review, but since `generated-rule-docs.ts` is breaking, I'll wait to get more detailed. Hopefully getting that to work isn't too tricky? Explicitly marking this PR as draft till then - happy to re-review once it is (or give input if getting it to work is unclear). Cheers!

packages/eslint-plugin/src/rules/no-restricted-imports.ts
JoshuaKGoldberg: Yeah seems like you're certainly following the standard practice: https://github.com/eslint/eslint/blob/1fea2797801a82a2718814c83dad641dab092bcc/lib/rules/no-restricted-imports.js#L19.

You might want to cast `const baseSchema = baseRule.meta.schema as SomeTypeDescribingWhatWeKnow`, since this is making an assumption on the base rule's schema type.

packages/eslint-plugin/src/rules/no-parameter-properties.ts
JoshuaKGoldberg: Sorry, I don't follow - why is this `minItems` not good? It's a correct restriction, no?

domdomegg: For this PR, the main reason I did it is beacuse it would prefer the default options from validating against the schema itself, and this is a simple way of getting around that.

But also I think semantically it makes sense that it should accept an array of nothing for the exceptions allowed, meaning that no parameter properties are allowed.

packages/eslint-plugin/src/rules/no-restricted-imports.ts
domdomegg: Good idea, have done this.

packages/eslint-plugin/tests/schemas.test.ts
domdomegg: Agreed, this makes sense. Have put the schema validation logic in a separate helper, and moved this out to the appropriate test files.

packages/eslint-plugin/src/rules/array-type.ts
domdomegg: Should be fixed now. Uses `items` instead of `prefixItems`, and also checks that it exists first.

domdomegg:
@JoshuaKGoldberg thanks for the quick review! I think I've got everything working now and have addressed the initial feedback, so it's ready for a more detailed review.

JoshuaKGoldberg (dismissed):
LGTM, thanks! Since @bradzacher has been looking in this area, I'll defer to Brad's second opinion for a review.

packages/eslint-plugin/src/rules/no-restricted-imports.ts
```diff
@@ -5,61 +5,160 @@ import type {
...
 export type MessageIds = InferMessageIdsTypeFromRule<typeof baseRule>;
 
-const allowTypeImportsOptionSchema = {
+// In some versions of eslint, the base rule has a completely incompatible schema
+// This helper function is to safely try to get parts of the schema. If it's not
+// possible, we'll fallback to less strict checks.
+const tryAccess = <T>(getter: () => T, fallback: T): T => {
+  try {
+    return getter();
+  } catch (error) {
```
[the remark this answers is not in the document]
domdomegg: Applied

bradzacher:
What I'll get you to do is rebase this on top of #6939 (which includes #6899 that's already landed in v6).
I've already merged your schema changes in as I built out the schema gen tooling - so what we're missing now is the explicit schema test cases.

So if you rebase this on my branch, we'll be able to merge your awesome validation tests into v6 and we can release it all together!

Thanks so much for the work you did here - it was a good set of pointers for me to look for in the generated types when writing the schema gen tooling 😄

bradzacher:
PRs are both merged so feel free to rebase on v6 at your leisure :)

domdomegg:
Closing in favour of #6947

PR #6896, opened 2023-04-12
fix(eslint-plugin): refactor schemas that use `$ref`

### LINKED ISSUE #6852: Bug: [array-type] the config `simple-array` should be disallowed by the schema

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6852 

## Overview

Updates rule schemas that use $ref to use an array-based schema instead of an object-based one and updates the doc gen to match.

ESLint will wrap an array schema in an array object: `[schema] -> { type: 'array', items: [schema] }`, so we can use that knowledge to reference the `$defs` with `#/items/0/$defs/...`.

This saves us manually recreating the tuple validation that ESLint already does for us.
It does require us to rewrite the `$ref` for our doc gen because we reference the schema object directly (eg `#/items/0/$defs/... -> #/$defs/...`

### DIFF
--- a/packages/eslint-plugin/src/rules/comma-dangle.ts
+++ b/packages/eslint-plugin/src/rules/comma-dangle.ts
@@ -47,6 +47,7 @@ export default util.createRule<Options, MessageIds>({
       recommended: false,
       extendsBaseRule: true,
     },
+    // intentionally a non-array schema to mirror the base rule
     schema: {
       $defs: {
         value: {
@@ -61,19 +62,19 @@ export default util.createRule<Options, MessageIds>({
         {
           oneOf: [
             {
-              $ref: '#/$defs/value',
+              $ref: '#/items/0/$defs/value',
             },
             {
               type: 'object',
               properties: {
-                arrays: { $ref: '#/$defs/valueWithIgnore' },
-                objects: { $ref: '#/$defs/valueWithIgnore' },
-                imports: { $ref: '#/$defs/valueWithIgnore' },
-                exports: { $ref: '#/$defs/valueWithIgnore' },
-                functions: { $ref: '#/$defs/valueWithIgnore' },
-                enums: { $ref: '#/$defs/valueWithIgnore' },
-                generics: { $ref: '#/$defs/valueWithIgnore' },
-                tuples: { $ref: '#/$defs/valueWithIgnore' },
+                arrays: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                objects: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                imports: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                exports: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                functions: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                enums: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                generics: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                tuples: { $ref: '#/items/0/$defs/valueWithIgnore' },
               },
               additionalProperties: false,
             },
--- a/packages/eslint-plugin/src/rules/padding-line-between-statements.ts
+++ b/packages/eslint-plugin/src/rules/padding-line-between-statements.ts
@@ -595,6 +595,7 @@ export default util.createRule<Options, MessageIds>({
     },
     fixable: 'whitespace',
     hasSuggestions: false,
+    // intentionally a non-array schema to mirror the base rule
     schema: {
       $defs: {
         paddingType: {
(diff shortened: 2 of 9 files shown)


packages/eslint-plugin/src/rules/comma-dangle.ts
```diff
@@ -61,19 +62,19 @@ export default util.createRule<Options, MessageIds>({
...
-                tuples: { $ref: '#/$defs/valueWithIgnore' },
+                arrays: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                objects: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                imports: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                exports: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                functions: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                enums: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                generics: { $ref: '#/items/0/$defs/valueWithIgnore' },
+                tuples: { $ref: '#/items/0/$defs/valueWithIgnore' },
               },
```
bradzacher, line 78:
```suggestion
                arrays: { $ref: '#/$defs/valueWithIgnore' },
                objects: { $ref: '#/$defs/valueWithIgnore' },
                imports: { $ref: '#/$defs/valueWithIgnore' },
                exports: { $ref: '#/$defs/valueWithIgnore' },
                functions: { $ref: '#/$defs/valueWithIgnore' },
                enums: { $ref: '#/$defs/valueWithIgnore' },
                generics: { $ref: '#/$defs/valueWithIgnore' },
                tuples: { $ref: '#/$defs/valueWithIgnore' },
              },
```

domdomegg:
Hey, just wanted to flag that I saw this limitation in ESLint when looking around there to understand how the schema logic worked and was considering contributing to ESLint to remove this limitation altogether: so the $refs are just hoisted to the top level schema.

Not sure if that would be accepted, but given their docs I don't think it would be a breaking change (as consumers are suppoesd to expect that $refs don't work with array types).

In any case, if we're using an undocumented feature of eslint we should probably have contract tests around eslint that it does validate these as expected, in case that logic does change.

packages/eslint-plugin/src/rules/padding-line-between-statements.ts
```diff
@@ -595,6 +595,7 @@ export default util.createRule<Options, MessageIds>({
     },
     fixable: 'whitespace',
     hasSuggestions: false,
+    // intentionally a non-array schema to mirror the base rule
```
armano2, line 598: while you are at it, can we fix/improve this schema here?


```diff
-      items: {
+      items: [{
        type: 'object',
        properties: {
          blankLine: { $ref: '#/$defs/paddingType' },
          prev: { $ref: '#/$defs/statementType' },
          next: { $ref: '#/$defs/statementType' },
        },
        additionalProperties: false,
        required: ['blankLine', 'prev', 'next'],
-      },
+      }],
```

packages/website/plugins/generated-rule-docs.ts
```diff
@@ -34,6 +34,59 @@ function nodeIsParent(node: unist.Node<unist.Data>): node is unist.Parent {
   return 'children' in node;
 }
 
+const prettierConfig = resolveConfig.sync(__dirname);
```
JoshuaKGoldberg, line 37: [Nitpick] Now we're mixing our style & the default, heh.

Screenshot of array-type mixing single and double apostrophes

JoshuaKGoldberg (approved):
LGTM!

Would normally request changes on using Prettier for `defaultOptions` too but I trust you 😄 and think it's fine as a followup.

packages/eslint-plugin/src/rules/padding-line-between-statements.ts
bradzacher: no  because that schema is actually not a tuple - it's an array!
https://eslint.org/docs/latest/rules/padding-line-between-statements#rule-details

```json
{
    "padding-line-between-statements": [
        "error",
        { "blankLine": LINEBREAK_TYPE, "prev": STATEMENT_TYPE, "next": STATEMENT_TYPE },
        { "blankLine": LINEBREAK_TYPE, "prev": STATEMENT_TYPE, "next": STATEMENT_TYPE },
        { "blankLine": LINEBREAK_TYPE, "prev": STATEMENT_TYPE, "next": STATEMENT_TYPE },
        { "blankLine": LINEBREAK_TYPE, "prev": STATEMENT_TYPE, "next": STATEMENT_TYPE },
        ...
    ]
}
```

bradzacher:
closing in favour of #6899 and #6894

PR #6899, opened 2023-04-13
feat: improve rule schemas, add test to validate schemas, add tooling to generate schema types

bradzacher:
## Overview

@JoshuaKGoldberg I went through with my threat for no reason other than it seemed fun.

Previously we used `json-schema-to-typescript` to generate our types, however we have found that the package has a number of edge-cases that don't do what we want. Attempting to change the package is... difficult because it is built to be super flexible so it can attempt to handle every single permutation and valid state of a JSON schema.

However in reality we don't actually need all that power - just a fraction of it because rule schemas are some degree of sane (usually... at least our are). This means we can easily constrain the transformation code and thus simplify it significantly.

This PR:
1) introduces a new (currently internal) package that contains tooling to convert a rule's schema to TypeScript types.
1) introduces snapshot tests for the plugin that snapshots the schema and the generated TS types
    - I realised that running this against our rules is a great way for us to validate our schemas are doing what we think they are - I personally find that seeing the fully resolved schema along side the computed TS types are incredibly helpful in understanding the schemas.
This has the added benefit that we can validate the types that will get embedded in the website without needing to do a website build - which is a nice, quick sanity check.
1) introduces a test for the plugin that validates that the schemas don't include invalid properties
    - Whilst building the tool I noticed a lot of invalid states in the schema, so I threw together a test to shout about unknown properties. This in turn helped me find a bunch of incorrect schemas!
1) updates the website build to use the new package

Note: I purposely haven't written tests for the tool as it's pretty extensively tested by running on our rules already. I didn't see any need in additional tests.


packages/eslint-plugin/tests/schema-snapshots/naming-convention.shot
```diff
@@ -0,0 +1,3473 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Rule schemas should be convertible to TS types for documentation purposes naming-convention 1`] = `
```
armano2, line 3: we should refactor this one to use `$refs` as its duplicating a lot of stuff

bradzacher: https://github.com/typescript-eslint/typescript-eslint/blob/deee1e97d58f34eda1ab46cdc41b2e73b3812a47/packages/eslint-plugin/tests/schema-snapshots/naming-convention.shot#L936-L1409

packages/eslint-plugin/tests/schema-snapshots/type-annotation-spacing.shot
```diff
@@ -0,0 +1,83 @@
...
+      arrow?: SpacingConfig;
+      colon?: SpacingConfig;
+      parameter?: SpacingConfig;
+      property?: SpacingConfig;
+      returnType?: SpacingConfig;
+      variable?: SpacingConfig;
+    };
+  },
+];
+"
```
armano2, line 82: code looks fine here, but there seem to be an issue at 
https://deploy-preview-6899--typescript-eslint.netlify.app/rules/type-annotation-spacing

packages/website/plugins/generated-rule-docs.ts
```diff
@@ -243,7 +253,10 @@ export const generatedRuleDocs: Plugin = () => {
 
       optionsH2Index += 2;
 
-      if (meta.schema.length === 0) {
+      const hasConfig = Array.isArray(meta.schema)
+        ? meta.schema.length === 0
+        : Object.keys(meta.schema).length === 0;
+      if (!hasConfig) {
```
armano2, line 259: this condition is not correct, `hasConfig` should be changed to `noConfig` or condition should be changed

```suggestion
      const noConfig = Array.isArray(meta.schema)
        ? meta.schema.length === 0
        : Object.keys(meta.schema).length === 0;
      if (noConfig) {
```

```suggestion
      const hasConfig = Array.isArray(meta.schema)
        ? meta.schema.length > 0
        : Object.keys(meta.schema).length > 0;
      if (!hasConfig) {
```

packages/eslint-plugin/tests/schema-snapshots/semi.shot
```diff
@@ -0,0 +1,75 @@
...
+  | (
+      | []
+      | ['never']
+      | [
+          'never',
+          {
+            beforeStatementContinuationChars?: 'always' | 'any' | 'never';
+          },
+        ]
+    );
```
armano2, line 73: do you think its viable to generate array's with `minItems` / `maxItems` to use optional tuple elements?

in some other rules this is getting overly complicated because of that

```ts
type Options =
  | [
      "always"?,
      {
        omitLastInOneLineBlock?: boolean;
      }?
    ]
  | [
      "never"?,
      {
        beforeStatementContinuationChars?: "always" | "any" | "never";
      }?
    ];
```

packages/website/src/components/editor/createProvideCodeActions.ts
```diff
@@ -34,9 +34,9 @@ export function createProvideCodeActions(
               edits: [
                 {
                   resource: model.uri,
-                  // @ts-expect-error monaco for ts >= 4.8
+                  // monaco for ts >= 4.8
                   textEdit: editOperation,
-                  // monaco for ts < 4.8
+                  // @ts-expect-error monaco for ts < 4.8
```
armano2, line 39: i pushed lint fix to v6 for this and other issues that where caused by incorrect merge, should be gone after rebase/merge

packages/website/plugins/generated-rule-docs.ts
```diff
@@ -22,6 +19,19 @@ const COMPLICATED_RULE_OPTIONS = new Set([
...
   'naming-convention',
 ]);
+/**
+ * Rules that do funky things with their defaults and require special code
+ * rather than just JSON.stringify-ing their defaults blob
+ */
+const SPECIAL_CASE_DEFAULTS = new Map([
+  //
+  ['ban-types', '[{ /* See below for default options */ }]'],
+]);
```
armano2, line 29: do we want to addres this in separate PR?

packages/website/plugins/generated-rule-docs.ts
bradzacher: Oops. I extracted the logic when I patched the json types package to fix a type error and I got it wrong.

packages/eslint-plugin/tests/schema-snapshots/padding-line-between-statements.shot
```diff
@@ -0,0 +1,272 @@
...
+        | 'while'
+        | 'with'
+      )[],
+    ];
+
+type Options = {
+  blankLine: PaddingType;
+  next: StatementType;
+  prev: StatementType;
+}[];
```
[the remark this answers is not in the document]
bradzacher: https://github.com/eslint/eslint/blob/main/lib/rules/padding-line-between-statements.js#L441-L469

This is just how the rule defines its schema!
They incorrectly defined it when they switched to an object schema to use defs.

packages/eslint-plugin/tests/schema-snapshots/semi.shot
bradzacher: Optional elements are bad because it allows assigning `undefined` - it's less safe.

https://www.typescriptlang.org/play?#code/C4TwDgpgBAKlC8UDaBnYAnAlgOwOYH4AaKbAVwFsAjCdfAXQG4AoJgCiTqhQENhMUAZpggpYASmbsARFM48+g4aJgS2SGcQCMc3vyEjxkpKWwATCEOwRTOhfuWr2J85evFnFnNdt6lhlqCQUACqCMicAD7IaFh4kdEYOLjEZFQ0jCzsPooGwY7qsly6OaJ5RhpQ2kV2fmVqHq421b65+Q1epu5mnlZN8i2lEkA

docs/linting/Troubleshooting.mdx
```diff
@@ -71,6 +71,42 @@ See our docs on [type aware linting](./Typed_Linting.mdx) for more information.
 You're using an outdated version of `@typescript-eslint/parser`.
 Update to the latest version to see a more informative version of this error message, explained [above](#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file 'backlink to I get errors telling me ESLint was configured to run ...').
 
+## How do I turn on a `@typescript-eslint` rule?
```
JoshuaKGoldberg, line 74: Meant for another PR?

packages/eslint-plugin/src/rules/array-type.ts
```diff
@@ -105,30 +105,27 @@ export default util.createRule<Options, MessageIds>({
...
-            },
-            readonly: {
-              $ref: '#/$defs/arrayOption',
-              description:
-                'The array type expected for readonly cases. If omitted, the value for `default` will be used.',
-            },
+        properties: {
+          default: {
+            $ref: '#/items/0/$defs/arrayOption',
+            description: 'The array type expected for mutable cases...',
```
JoshuaKGoldberg, line 118: [Question] 🤔 existing code, but why the `...`?

packages/eslint-plugin/src/rules/ban-types.ts
```diff
@@ -34,9 +34,9 @@ function stringifyNode(
 }
 
 function getCustomMessage(
-  bannedType: null | string | { message?: string; fixWith?: string },
+  bannedType: null | true | string | { message?: string; fixWith?: string },
 ): string {
-  if (bannedType == null) {
+  if (bannedType == null || bannedType === true) {
```
JoshuaKGoldberg, line 39: [Testing] Should this come with a `ban-types.test.ts` change? I don't think this is already covered?

JoshuaKGoldberg (commented):
also what armano2 said 😄 - but this looks great so far! I'll take a deeper look upon re-request

packages/website/plugins/generated-rule-docs.ts
bradzacher: 🤷‍♂️ I can split it out if you think it's worth the effort?
I put it in here cos I was working on the docs as a whole here and leaving this as `const defaultOptions: Options = [{}];` seemed wrong.

docs/linting/Troubleshooting.mdx
bradzacher: Intentionally added as part of cleaning up the holistic docs update in this PR.
We've seen this asked a few times and the ESLint team doesn't think their docs are unclear, so as part of improving the rule config docs - this FAQ article to quick-link people to.

packages/eslint-plugin/src/rules/ban-types.ts
bradzacher: Yeah this was accidentally allowed by the schema (it was set to `{type: 'boolean'}` instead of `{enum: [false]}`) - but the value wasn't handled by the logic or properly documented.
Will add a test

packages/eslint-plugin/tests/schema-snapshots/semi.shot
bradzacher: The better solution here would be to improve the algorithm to collapse the union.
EG instead of `([] | [ ..1.. ]) | ([] | [ ..2.. ])`
it should be `[] | [ ..1.. ] | [ ..2.. ]`

For this example it could be:
```ts
type Options =
  | []
  | ['always']
  | [
      'always',
      {
        omitLastInOneLineBlock?: boolean,
      }
    ]
  | ['never']
  | [
      'never',
      {
        beforeStatementContinuationChars?: 'always' | 'any' | 'never',
      }
    ];
```

packages/website/src/components/lib/jsonSchema.ts
```diff
@@ -8,6 +8,11 @@ const defaultRuleSchema: JSONSchema4 = {
   enum: ['off', 'warn', 'error', 0, 1, 2],
 };
 
+// https://github.com/microsoft/TypeScript/issues/17002
+function isArray(arg: unknown): arg is readonly unknown[] {
+  return isArray(arg);
+}
```
armano2, line 14:
```suggestion
function isArray(arg: unknown): arg is readonly unknown[] {
  return Array.isArray(arg);
}
```

as is i think this is an infinite loop

packages/website/src/components/lib/jsonSchema.ts
```diff
@@ -18,30 +23,18 @@ const defaultRuleSchema: JSONSchema4 = {
...
     return {
       type: 'array',
       items: [defaultRuleSchema, ...ruleSchema],
       minItems: 1,
       additionalItems: false,
     };
   }
-  // TODO: delete this once we update schemas
-  // example: ban-ts-comment
-  if (Array.isArray(ruleSchema.prefixItems)) {
```
armano2, line 33: after infinite loop is fixed, we should verify if eslint is not using it:

```
monaco.languages.json.jsonDefaults.diagnosticsOptions.schemas.filter(item => item.schema.prefixItems)
```

packages/website/src/components/lib/jsonSchema.ts
bradzacher: Hahahaha "select all occurrences in the file and replace" did not even notice.

packages/website/src/components/lib/jsonSchema.ts
bradzacher: We are not using it, nor is eslint.
Its not a valid Json Schema v4 property.

It wasn't caught before because the JSONSchema type had a `[k: string]: any` signature that allowed additional, incorrect properties.
I patched the package to remove that signature (hence I found this error too).

Using `prefixItems` was what caused a few of our rules to not validate their options (because v6 of avj only supports v4 of json schema)

armano2: yes i just validated it and its no longer used, great job 👍

armano2: > Using prefixItems was what caused a few of our rules to not validate their options (because v6 of avj only supports v4 of json schema)

ajv supports other versions of schema but eslint is forcing it to use v4, 
https://github.com/eslint/eslint/blob/1fea2797801a82a2718814c83dad641dab092bcc/lib/shared/ajv.js#L11-L12

packages/eslint-plugin/tests/schema-snapshots/naming-convention.shot
```diff
@@ -1572,7 +1572,6 @@ type Options = /** Selector 'default' */
       suffix?: PrefixSuffixConfig;
       trailingUnderscore?: UnderscoreOptions;
     }
-  /** Multiple selectors in one config */
```
armano2: do we want to preserve description here?

packages/rule-schema-to-typescript-types/src/index.ts
```diff
@@ -78,10 +83,16 @@ export function compile(
...
-  return prettierFormat(
-    [...refTypes, optionsType].join('\n\n'),
-    prettierConfig,
+
+  const codeWithUnsimplifiedOptions = [...refTypes, optionsType].join('\n\n');
+  const simplifiedOptionsType = useTsToSimplifyOptionsType(
+    codeWithUnsimplifiedOptions,
+  );
+  const codeWithSimplifiedOptions = [...refTypes, simplifiedOptionsType].join(
+    '\n\n',
```
armano2, line 93: can we correct order of items, to reproduce how this was generated before?

in main/v6 we print defaults and options in same order

main:
image

this branch:
image

packages/eslint-plugin/tests/schema-snapshots/naming-convention.shot
bradzacher: hmm - this is a result of my last change to use TS to automatically simplify unions.
it appears that TS does not retain the comments on union members 😢

Maybe we can't do this then and we're just stuck with uncollapsed unions sadge

packages/rule-schema-to-typescript-types/src/index.ts
bradzacher: yeah I can update the defaults printer so that it sorts alphabetically as well!

packages/eslint-plugin/tests/schema-snapshots/naming-convention.shot
armano2: hmm wdyt about changing `printUnionType` to

```ts
function printUnionType(members: (string | GeneratedResult)[]): string {
  return members
    .map(member => {
      if (typeof member === 'string') {
        return `${member}`;
      }
      return `${printComment(member)} ${member.code}`;
    })
    .join(' | \n');
}
```

there is few differences tho:

image

image

packages/rule-schema-to-typescript-types/src/index.ts
armano2: > yeah I can update the defaults printer so that it sorts alphabetically as well!

can we use `natural-compare` instead of `localeCompare` for this?

that should sort it as `'array' | 'array-simple' | 'generic'` instead of `'array-simple' | 'array' | 'generic'`

packages/eslint-plugin/tests/schema-snapshots/naming-convention.shot
bradzacher: The big problem right now is that we are building strings - and it's really hard to tell if a string returned from the function needs to be wrapped or not - so we just always wrap it.

I think we can just do better here by rewriting the algorithm to build a pseudo-AST instead of building strings - that way we don't need to worry about superfluous parentheses because we can make decisions on the fly as to whether or not we need parentheses or how things get nested.
Can also do a deduplication pass as well.
It'll be much easier to do an optimisation pass when printing the AST from the top down as opposed to trying to optimise the strings from the bottom up.

armano2: > I think we can just do better here by rewriting the algorithm to build a pseudo-AST

I get that, i'm just wandering how granule you want to make it, in teory something like this for now could be sufficient:

```ts
type GeneratedResult =
  | {
      readonly code: string;
      readonly type: 'raw';
      readonly commentLines?: string[];
    }
  | {
      readonly code: string[];
      readonly type: 'union';
      readonly commentLines?: string[];
    };
```

----------------

i'm not sure how high this is on your PRIO list, I really like this change :>

bradzacher: yeah I wasn't going to go full ts-estree AST - just a light-weight object to do exactly what you described!

I wonder actually - I actually might just be able to remove the parentheses cos we only generate literals, arrays, objects, and unions - none of which require parentheses...

I'll play around and see.

Removing TS is good anyway cos it's a pretty slow way of doing things (makes the schema test take ~10s). But it's good to know how easy it is to use `@typescript/vfs` to do a quick typecheck like this.

packages/rule-schema-to-typescript-types/src/index.ts
bradzacher: After playing for a bit - natural compare isn't a good fit for code, IMO.

For example natural compare will sort tuples in a nearly reverse order to what we'd want:
```ts
[
  "['always']",
  "['never',{ignoreForLoopInit?: boolean;},'option']",
  "['never',{ignoreForLoopInit?: boolean;},]",
  "['never']",
  '[]'
]

// we really want this order:
[
  '[]',
  "['always']",
  "['never']",
  "['never',{ignoreForLoopInit?: boolean;},]",
  "['never',{ignoreForLoopInit?: boolean;},'option']",
]
```

Unfortunately it ignores whitespace amongst non-letters, so we can't control this without doing some really janky stuff.
I'll stick with localeCompare instead

package.json
```diff
@@ -71,6 +71,7 @@
     "@types/jest": "^29.0.2",
     "@types/jest-specific-snapshot": "^0.5.5",
     "@types/marked": "^4.0.3",
+    "@types/natural-compare-lite": "^1.4.0",
```
armano2, line 74:
```suggestion
    "@types/natural-compare": "^1.4.0",
```

armano2 (approved):
this looks awesome, thank you for you hard work

yarn.lock
```diff
@@ -3888,7 +3865,12 @@
   resolved "https://registry.yarnpkg.com/@types/ms/-/ms-0.7.31.tgz#31b7ca6407128a3d2bbc27fe2d21b345397f6197"
   integrity sha512-iiUgKzV9AuaEkZqkOLDIvlQiL6ltuZd9tGcW3gwpnX8JbuiuhFlEGmmFXEXkN50Cvq7Os88IY2v0dkDqXYWVgA==
 
-"@types/natural-compare@^1.4.0":
+"@types/natural-compare-lite@^1.4.0":
+  version "1.4.0"
+  resolved "https://registry.yarnpkg.com/@types/natural-compare-lite/-/natural-compare-lite-1.4.0.tgz#90724682da3c304dd8d643b4e9ba00f53f29d454"
+  integrity sha512-ZDcj/yWsRPacqKPpCExWWFq9X1JQwEOfHsu8deN1Qfa6M02z6tN4DK6AMf2IkM7709gp3QW6Bo7m2NFDhA485w==
+
+"@types/natural-compare@*":
```
armano2, line 3873: "@types/natural-compare@*", "@types/natural-compare@^1.4.0":

Issue #6827, opened 2023-04-03
Repo: Enable strict & stylistic configs internally


JoshuaKGoldberg:
### Suggestion

~Labeling this in the v6 milestone because if we go for this, I'd like it to be used as dogfood / real world user feedback.~ Similar benefits tracked in #6759, so removing this from the v6 milestone.

Right now in our internal ESLint config we just enable `recommended-type-checked`: https://github.com/typescript-eslint/typescript-eslint/commit/5346b5bbdbba81439ba761c282ba9cdcec7b45c8

It'd sure be nice if we used our own strict & stylistic configs!

PR #6915, opened 2023-04-15
chore(node-utils): simplify some util functions and add some unit tests

bertyhell:
## PR Checklist

## Overview

It uses some more modern syntax to make some utility functions more concise. I also added a few unit tests.
This PR doesn't fix a specific issue. I'm not sure what I would put in the issue description?

These code suggestions were done by chatgpt

### DIFF
--- a/packages/typescript-estree/src/node-utils.ts
+++ b/packages/typescript-estree/src/node-utils.ts
@@ -97,10 +97,7 @@ export function hasModifier(
  */
 export function getLastModifier(node: ts.Node): ts.Modifier | null {
   const modifiers = getModifiers(node);
-  if (modifiers == null) {
-    return null;
-  }
-  return modifiers[modifiers.length - 1] ?? null;
+  return modifiers?.[modifiers.length - 1] ?? null;
 }
 
 /**
@@ -338,13 +335,10 @@ export function findFirstMatchingAncestor(
   node: ts.Node,
   predicate: (node: ts.Node) => boolean,
 ): ts.Node | undefined {
-  while (node) {
-    if (predicate(node)) {
-      return node;
-    }
+  while (node && !predicate(node)) {
     node = node.parent;
   }
-  return undefined;
+  return node;
 }
 
 /**
@@ -396,9 +390,7 @@ export function isComputedProperty(
 export function isOptional(node: {
   questionToken?: ts.QuestionToken;
 }): boolean {
-  return node.questionToken
-    ? node.questionToken.kind === SyntaxKind.QuestionToken
-    : false;
+  return node.questionToken?.kind === SyntaxKind.QuestionToken ?? false;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (changes requested):
Ooh, thank you for looking out for our code quality! Much appreciated 🙂. But...

> This PR doesn't fix a specific issue

Indeed, there's a reason at least 1-2 of the pages you had to get through to send a PR mentioned attaching to an issue. These changes generally seem reasonable, but are all code points that could be statically detected by a linter. If we want to prefer this more succinct style _(I do!)_ we should have our lint rules tackle it. Tracked in #6827. Perhaps you'd have a good time working on that one?

JoshuaKGoldberg:
> code suggestions were done by chatgpt

😂.

If you're just mentioning that it was authored by ChatGPT as a cool point, nifty! That is cool! Don't feel the following rant applies to you.

But, because people look at our PRs (and AI bots are approaching their _peak of inflated expectations_ part of the [Gartner hype cycle](https://en.wikipedia.org/wiki/Gartner_hype_cycle)), I want to make something clear about code suggestions done by ChatGPT: deferring to an AI bot is _not the same as code ownership_.

I'm sure previous PRs have had code authored by ChatGPT. Plenty of people at the very least use GitHub Copilot. There's no way for us to know. But the idea that an author puts some code into a commit and sends it means they should have an intellectual _understanding_ of it. PR authors should "own" the code they send -- ownership in the sense of being able to advocate for the code.

For PR authors _attempting_ to understand code written by person(s) and/or bot(s), that's another matter. I'd welcome questions posted as comments in the PR.

bertyhell:
> But, because people look at our PRs (and AI bots are approaching their _peak of inflated expectations_ part of the [Gartner hype cycle](https://en.wikipedia.org/wiki/Gartner_hype_cycle)), I want to make something clear about code suggestions done by ChatGPT: deferring to an AI bot is _not the same as code ownership_.

I understand. To be fair. More than half of the suggestions it made were non nonsensical.
But i was insipred by this article that found a one line fix for a performance problem: https://jpcamara.com/2023/03/07/making-tanstack-table.html

So i looked at that PR and gave chatGPT the old code from the PR and asked it to optimize it. It also fixed the performance issue.

It seems like an interesting way to automate performance improvements:
* Run a library and identify hot code (code that is called a lot of times) https://www.npmjs.com/package/v8-profiler-next
* Extract stateless functions from that code
* Ask ChatGPT to generate unit tests for that function
* Ask ChatGPT to optimize the function
* automatically run the unit tests with the original function
* automatically run the unit tests with the optimized function
* if both succeed, add the list of possible optimizations
* Keep find more optimizations
* Present the programmer with a list of possible optimizations and how much it cuts down on execution time when run against a real life codebase.
* Profit?

I gave it a go, but this code base seems to already be quite optimized :p

PR #6923, opened 2023-04-17
chore: bump minimum TS to 4.3.5

bradzacher:
## Overview

We've taken a minute here to finish our v6 so v4.2.4 has now fallen outside the 2 year support window for versions.
This bumps the minimum to v4.3.5.
This also makes our lives easier for #6723


.gitignore
```diff
@@ -84,6 +84,7 @@ _ts4.2
 # the downlevel-dts output folders
 _ts3.4
```
JoshuaKGoldberg, line 85: Can we get rid of `_ts3.4` now?

JoshuaKGoldberg (approved):
Great!

.gitignore
armano2: we should keep it in gitignore until we merge it, as it will prevent us from accidently pushing those files while switching between v6 and main

PR #6930, opened 2023-04-18
chore: remove duplicated and unused dependencies

armano2:
## PR Checklist

## Overview

- We are no longer using `lodash` and we can safely remove `@types` for it
- `natural-compare-lite` and `natural-compare` is a same package with exact same code, and instead of using `natural-compare-lite` we should use `natural-compare` as its already required by `eslint`
  - this package had marginal amounts of downloads before was added to our project :>


JoshuaKGoldberg (approved):
Llama flipping on sunglasses and saying 'cool'

PR #6939, opened 2023-04-20
fix: correct handling of additionalProperties in object schemas

bradzacher:
## PR Checklist

## Overview

@domdomegg's #6894 made me realise I'd incorrectly handled this case in the schema gen.
When it's missing/`undefined` it's the same as being `true` - but I treated it as `false`


domdomegg (approved):

armano2 (approved):
good catch 👍

PR #6947, opened 2023-04-22
fix(eslint-plugin): fix schemas across several rules and add schema tests

### LINKED ISSUE #6852: Bug: [array-type] the config `simple-array` should be disallowed by the schema

domdomegg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6852, #6892

## Overview

Rebase of #6894 on top of v6, as per https://github.com/typescript-eslint/typescript-eslint/pull/6894#issuecomment-1516224308


packages/eslint-plugin/src/rules/no-restricted-imports.ts
```diff
@@ -19,37 +19,95 @@ const baseRule = getESLintCoreRule('no-restricted-imports');
...
+// possible, we'll fallback to less strict checks.
+const tryAccess = <T>(getter: () => T, fallback: T): T => {
+  try {
+    return getter();
+  } catch {
+    return fallback;
+  }
+};
+
+const baseSchema = baseRule.meta.schema as {
```
bradzacher, line 33: question: why did you switch back to basing off of the base rule's schema here?
Any reason or was it just to go back to de-duping things?

packages/eslint-plugin/src/rules/parameter-properties.ts
```diff
@@ -59,7 +59,6 @@ export default util.createRule<Options, MessageIds>({
             items: {
               $ref: '#/items/0/$defs/modifier',
             },
-            minItems: 1,
```
bradzacher, line 62: question: why remove this? isn't it better if we keep this for strictness to prevent empty arrays?

packages/eslint-plugin/tests/areOptionsValid.ts
```diff
@@ -0,0 +1,43 @@
...
+    return false;
+  }
+
+  return valid;
+}
+
+function normalizeSchema(
+  schema: JSONSchema4 | readonly JSONSchema4[],
+): JSONSchema4 {
+  if (!Array.isArray(schema)) {
```
bradzacher, line 25: We could export this from the `index.ts` in the package and use it here
https://github.com/typescript-eslint/typescript-eslint/blob/3e3b7891a73136136929b6912ace9301b30fb0f3/packages/rule-schema-to-typescript-types/src/isArray.ts

Which will fix the safety caused by the TS bug

packages/eslint-plugin/tests/rules/array-type.test.ts
```diff
@@ -2156,3 +2157,19 @@ type BrokenArray = {
...
+    }
+  });
+
+  // https://github.com/typescript-eslint/typescript-eslint/issues/6892
+  test('array-type does not accept non object option', () => {
+    if (areOptionsValid(rule, ['array'])) {
+      throw new Error(`Options succeeded validation for bad options`);
+    }
+  });
+});
```
bradzacher, line 2175: I wonder if we should do a rule-tester style API here to remove the boilerplate.
For example something like this:

```ts
ruleOptionsTester(rule, {
  valid: [ ... ],
  invalid: [
    [{ default: 'simple-array' }],
    ['array'],
  ],
});
```

packages/eslint-plugin/src/rules/no-restricted-imports.ts
domdomegg: No strong reason: just to dedupe things / keep things in sync.

packages/eslint-plugin/tests/areOptionsValid.ts
domdomegg: Nice - didn't see this helper existed. Have moved it from the rule-schema-to-typescript-types package to the utils package and have used here.

packages/eslint-plugin/tests/rules/array-type.test.ts
domdomegg: Yeah, I think something like this would be neat if we end up doing a lot of these kinds of tests. For now I think it's simple enough that we can leave this to be done in future if necessary?

JoshuaKGoldberg:
Removing from the v6 milestone as it doesn't actually block releasing v6 (I believe). Someone yell at me if that's wrong. 🙂

JoshuaKGoldberg:
Oh also, @domdomegg just checking in - are you waiting on us for anything? I believe our last status was waiting on you to re-request review to take another look.

packages/eslint-plugin/tests/rules/array-type.test.ts
bradzacher: Note that there is
https://github.com/eslint/rfcs/pull/103 and the implementing PR https://github.com/eslint/eslint/pull/16823

This would allow us to include these tests in the rule like:
```
tester.run('rule', rule, {
  valid: [ ... ],
  invalid: [ ... ],
  fatal: [
    {
      options: ['simple-array'],
      error: {
        name: 'SchemaValidationError',
        message: 'Whatever the json schema error is',
      },
    },
    {
      options: ['array'],
      error: {
        name: 'SchemaValidationError',
        message: 'Whatever the json schema error is',
      },
    },
  ],
}
```

bradzacher (dismissed):

bradzacher:
@domdomegg if you want to update this against `v6` and resolve the conflicts, and get the CI passing - then we're good to merge this in!

domdomegg:
I think we're ready @bradzacher / @JoshuaKGoldberg :)

JoshuaKGoldberg:

domdomegg:
Thanks @JoshuaKGoldberg :raised_hands:

PR #6962, opened 2023-04-27
chore: bump deps

bradzacher:
bumps all our dependency versions across all packages and tidies up some of the package.jsons (removed local `@types/` references to deduplicate things, removed some unused packages)


patches/eslint-visitor-keys+3.4.0.patch
```diff
@@ -0,0 +1,33 @@
+diff --git a/node_modules/eslint-visitor-keys/dist/eslint-visitor-keys.d.cts b/node_modules/eslint-visitor-keys/dist/eslint-visitor-keys.d.cts
```
bradzacher, line 1: https://github.com/eslint/eslint-visitor-keys/pull/47

JoshuaKGoldberg (approved):
Barbie sitting down playing an electric guitar. Caption: 'you rock'

PR #6963, opened 2023-04-27
feat: fork json schema types for better compat with ESLint rule validation

bradzacher:
BREAKING CHANGE:
`@typescript-eslint/utils` now only exports the v4 JSON schema types - the only version ESLint validates rules with.
Removed the unsafe `[string]: any` indexer from the `JSONSchema4` type to help ensure invalid properties aren't used.

---

This was something I noticed whilst working on our schema enhancements - we re-exported all these types that were mostly useless for lint rules because ESLint is hard-locked to v4 of the schema. So I decided to remove those newer schema types from the export.

I was going to leave it there, but then I remember that I had added a local patch that strictified the types by removing the unsafe indexer (which helped us catch bugs in our schemas). So I decided we can just fork the types entirely and apply my patch so all consumers get that same benefit.


packages/utils/src/json-schema.ts
```diff
@@ -1,24 +1,253 @@
...
+
+//==================================================================================================
+// JSON Schema Draft 04
+//==================================================================================================
+
+/**
+ * @see https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.1
+ */
+export type JSONSchema4TypeName =
+  | 'string' //
```
JoshuaKGoldberg, line 16: [Nitpick] Here and elsewhere

```suggestion
  | 'string'
```

JoshuaKGoldberg (approved):
🔥 looks great! I wonder if we could export this as our own types so other ESLint projects (maybe ESLint itself, if they want a devDependency to help with types) could use it? Given that it's already exported by `@typescript-eslint/utils` I think it's fine to stay as-is.

armano2 (approved):

bradzacher:
@JoshuaKGoldberg / @armano2 I just pushed a new commit (178cf35b67421ca91e76975a659679f7519c4226) that further strictifies the types to a discriminated union.
This allows us to use types to validate that you don't provide mismatched properties.
This caught a bug in two of our schema defs:
- [one of them defined `additionalProperties` instead of `additionalItems`](https://github.com/typescript-eslint/typescript-eslint/pull/6963/files#diff-7cd04f626b0e300cad28c7da0af5ca89f3d768766c8ea0f8f568423e80a230ec)
- [one of them defined `minLength` instead of `minItems`](https://github.com/typescript-eslint/typescript-eslint/pull/6963/files#diff-f7d4ffd3171ad6097effd69dcfaa1aba256c01b961e8e0e193b6de535e157a4d)

The downside to doing this is that you can be significantly less flexible in how you write schemas. For reference, you now must always define a `type` for each schema so that it can correctly pick the type, and it restricts some funky combos like adding `anyOf` + `$ref` + any concrete type.
Given that we always want to represent schemas as types - I don't think this is necessarily a bad thing though? Across our codebase we didn't end up changing anything aside from adding `type` in a few missing places.

If you guys think it's better to just stick to the spec and be loose - more than happy to revert the schema changes.

It's really hard to judge the impact of such a change, given that we don't know how all our consumers define schemas - but I'd think they're not too crazy?

JoshuaKGoldberg:
Oof this hurts me though 😞. It seems like it'd be very annoying to have to write out the `type` when it can be inferred: 

> ```ts
> arrayOption: {
>     type: 'string',
>     enum: ['array', 'generic', 'array-simple'],
> },
> ```

Maybe there's a happy medium we can go for? Only require `type` when it's not directly inferable?

bradzacher:
> It seems like it'd be very annoying to have to write out the type when it can be inferred

The flip side is that the explicit `type` will prevent cases like `enum: [1, 2, "3"]` that would otherwise make the 3rd option inaccessible.

PR #6998, opened 2023-05-06
chore: apply provenance data to canary releases

JamesHenry:


JoshuaKGoldberg:
Side note: very impressive how little config was needed to enable this. I'd been worrying about needing to set up complex build tooling... Nicely done @JamesHenry 😄

JamesHenry:
@JoshuaKGoldberg thanks, although worth noting I forgot the token permission initially. I also switched from env var to .npmrc so that we get it for free in the private automations pipeline without changes 

https://github.com/typescript-eslint/typescript-eslint/commit/6d5e1b9efa2e631ab74f11700202e8bb08ca877c

Issue #6581, opened 2023-03-07
[restrict-template-expressions] should clarify what it means when it refers to a useful toString


tom-seddon:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

Quoting the docs:

> This rule reports on values used in a template literal string that aren't primitives and don't define a more useful .toString() method

But, as per https://github.com/typescript-eslint/typescript-eslint/issues/3538, it is by design that a definition of a more useful `toString` is in fact ignored. The docs should ideally be updated to reflect this. Or perhaps just clarified? - clearly it's not just me that doesn't quite follow what the rules are.

### Affected URL(s)

What's CRUD? The source code for the docs appears to be here, if that's any help: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/restrict-template-expressions.md

felixfbecker:
Agree, this is very confusing.

The behavior also doesn't stack well with `no-base-to-string` like this. It is impossible at the moment to enforce that e.g. putting a `URL` object into a template string is okay, but `URL | undefined` is not.

Josh-Cena:
@felixfbecker, `URL` not being allowed in `no-base-to-string` is a bug. You should be able to do `` `${url.toString()}` ``.

bradzacher:
[...]

PR #7009, opened 2023-05-07
docs: be more explicit about what restrict-template-expressions restricts

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/6581

### DIFF
--- a/packages/eslint-plugin/docs/rules/restrict-template-expressions.md
+++ b/packages/eslint-plugin/docs/rules/restrict-template-expressions.md
@@ -13,6 +13,8 @@ This rule reports on values used in a template literal string that aren't string
 :::note
 
 This rule intentionally does not allow objects with a custom `toString()` method to be used in template literals, because the stringification result may not be user-friendly.
+
+For example, arrays have a custom [`toString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toString) method, which only calls `join()` internally, which joins the array elements with commas. This means that (1) array elements are not necessarily stringified to useful results (2) the commas don't have spaces after them, making the result not user-friendly. The best way to format arrays is to use [`Intl.ListFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat), which even supports adding the "and" conjunction where necessary.
 You must explicitly call `object.toString()` if you want to use this object in a template literal.
 The [`no-base-to-string`](./no-base-to-string.md) rule can be used to guard this case against producing `"[object Object]"` by accident.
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (dismissed):
💯

Darkly lit dungeon-y scene; hooded figure with visible face saying 'flawless victory'

packages/eslint-plugin/docs/rules/restrict-template-expressions.md
```diff
@@ -6,9 +6,17 @@ description: 'Enforce template literal expressions to be of `string` type.'
...
-JavaScript will call `toString()` on an object when it is converted to a string, such as when `+` adding to a string or in `${}` template literals.
-The default Object `.toString()` returns `"[object Object]"`, which is often not what was intended.
-This rule reports on values used in a template literal string that aren't primitives and don't define a more useful `.toString()` method.
+JavaScript automatically [converts an object to a string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#string_coercion) in a string context, such as when concatenating it with a string using `+` or embedding it in a template literal using `${}`.
+The default `toString()` method of objects returns `"[object Object]"`, which is often not what was intended.
+This rule reports on values used in a template literal string that aren't strings, numbers, or BigInts, optionally allowing other data types that provide useful stringification results.
+
+:::note
+
+This rule intentionally does not allow objects with a custom `toString()` method to be used in template literals, because the stringification result may not be user-friendly.
```
bradzacher, line 15: Might be worth mentioning like "for example arrays have a custom toString which is akin to calling `.join(',')` - which is not a user-friendly representation as it does not include spaces between the commas."

bradzacher (approved):

packages/eslint-plugin/docs/rules/restrict-template-expressions.md
Josh-Cena:
```suggestion
This rule intentionally does not allow objects with a custom `toString()` method to be used in template literals, because the stringification result may not be user-friendly. For example, arrays have a custom [`toString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toString) method, which only calls `join()` internally, which joins the array elements with commas. This means that (1) array elements are not necessarily stringified to useful results (2) the commas don't have spaces after them, making the result not user-friendly. The best way to format arrays is to use [`Intl.ListFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat), which even supports adding the "and" conjunction where necessary.
```

PR #7022, opened 2023-05-11
feat: drop support for node v14 and test against node v20

bradzacher:
## Overview

As of April 30 2023 node14 is now out-of-support.
This PR drops it from our testing pipeline and bumps the engines config.
This PR also bumps the highest tested version to v20.


JoshuaKGoldberg (approved):
🚀

JoshuaKGoldberg:
```plaintext
```

Fun! 🙃

Issue #6225, opened 2022-12-16
Enhancement: add strict `parent` types for nodes that have well-defined parents


bradzacher:
Right now a node's `.parent` property is loosely typed as `TESTree.Node`. This is pretty sucky as it means that users need to either add unnecessary logic or explicit casts to refine types in some cases.

There are a number of nodes wherein we know exactly what the parent node will always be due to the defined structure of the AST.
For example, we know that a `VariableDeclarator` will always have a `VariableDeclaration` parent.

We should define some of these simpler relationships in our types to improve the experience of consuming the AST types.

Some cases we can define:
- [x] `AccessorProperty` has parent `ClassBody` (#9487)
- [ ] `CatchClause` has parent `TryStatement`
- [ ] `ClassBody` has parent `ClassDeclaration | ClassExpression | TSAbstractClassDeclaration`
- [ ] `ExportSpecifier` has parent `ExportNamedDeclaration`
- [ ] `ImportAttribute` has parent `ImportDeclaration | ImportExpression`
- [ ] `ImportDefaultSpecifier` has parent `ImportDeclaration`
- [ ] `ImportNamespaceSpecifier` has parent `ImportDeclaration`
- [ ] `ImportSpecifier` has parent `ExportAllDeclaration | ExportNamedDeclaration | ImportDeclaration`
- [ ] `JSXAttribute` has parent `JSXOpeningElement`
- [ ] `JSXClosingElement` has parent `JSXElement`
- [ ] `JSXClosingFragment` has parent `JSXFragment`
- [ ] `JSXOpeningElement` has parent `JSXElement`
- [ ] `JSXOpeningFragment` has parent `JSXFragment`
- [ ] `JSXSpreadAttribute` has parent `JSXOpeningElement`
- [ ] `MethodDefinition` has parent `ClassBody`
- [ ] `Property` has parent `ObjectExpression | ObjectPattern`
- [ ] `PropertyDefinition` has parent `ClassBody`
- [ ] `SpreadElement` has parent `ArrayExpression | CallExpression | ObjectExpression`
- [ ] `StaticBlock` has parent `ClassBody`
- [ ] `SwitchCase` has parent `SwitchStatement`
- [ ] `TemplateElement` has parent `TemplateLiteral | TSTemplateLiteralType`
- [ ] `TSAbstractAccessorProperty` has parent `ClassBody`
- [ ] `TSAbstractMethodDefinition` has parent `ClassBody`
[...]

PR #7065, opened 2023-05-25
feat(typescript-estree): split TSMappedType typeParameter into constraint and key

### LINKED ISSUE #6433: AST cleanup: type parameter in `TSMappedType`
```ts
type MappedType = {
  [Key in Type]: Value;
};
```

Mapped types are a self-contained thing in TS; they can declare exactly one signature - the index signature defining the mapping. They can declare no other properties.
As such we emit a special AST node for them - `TSMappedType`
https://github.com/typescript-eslint/typescript-eslint/blob/a9cb860c37590b01222879f07a4c63e6ce95a37d/packages/ast-spec/src/type/TSMappedType/spec.ts#L6-L13

The problem with this AST is that we emit a `TSTypeParameter` for the indexer.
```ts
[ Key in Type ] : Value;
  ^^^^^^^^^^^ TSTypeParameter
  ^^^ .name = Identifier
         ^^^^ .constraint = TypeNode
```
We emit this because the underlying TS AST uses a `ts.TypeParameter` node for this.

This is a pretty weird AST considering the following things:
1) `TSTypeParameter` supports things not syntactically[1] valid in a mapped type indexer (eg variance sigils (`in`, `out`) and default value `T = Default`).
1) `.constraint` is marked as optional on `TSTypeParameter`, even though it is a syntactically[1] required.
1) in all other usages of `TSTypeParameter` the existence of the `.constraint` property implies the `extends` keyword: `Name extends Constraint`. However just for the `TSMappedType` the `.constraint` instead implies the `in` keyword: `Name in Constraint`.
  Note that in this location it is syntactically[1] invalid to use `[Key extends Value]`.
1) it's possible to define a [key remapping](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#key-remapping-in-mapped-types) in a mapped type. This remapping node is defined as a sibling of the `TSTypeParameter`, but is part of the `TSMappedType` node:
    ```ts
    type MappedType = {
      [ Key in Type as Remapping ]: Value;
        ^^^^^^^^^^^ TSTypeParameter
                       ^^^^^^^^^ TSMappedType.nameType
    };
    ```

[1] Note that by syntactically valid I mean fatal parser error from TS - not just a semantic error!

It's pretty clear to me that our AST shape here is wrong and should be changed.

I propose that we switch to the following AST shape:
```diff
 export interface TSMappedType extends BaseNode {
   type: AST_NODE_TYPES.TSMappedType;
-  typeParameter: TSTypeParameter;
+  key: Identifier;
+  constraint: TypeNode;
   readonly?: boolean | '-' | '+';
   optional?: boolean | '-' | '+';
   typeAnnotation?: TypeNode;
   nameType: TypeNode | null;
 }
```

**_Migration Plan_**
With this sort of breaking AST change it can be difficult to roll out to the ecosystem without breaking the plugin ecosystem that may rely on the existing shape.
I propose the following migration plan:
1) implement the above additions (`key` and `constraint`) to the AST.
2) use JSDoc to mark `typeParameter` as `@deprecated`.
3) (optional) add test-time logging to warn consumers that they're using a deprecated, to-be-removed property.
4) broadcast this change (perhaps via a blog post for easy sharing?).
5) breaking change remove `typeParameter` in a major release.

JoshuaKGoldberg:
BREAKING CHANGE:
Deprecates an existing property on the AST

## PR Checklist

- [x] Addresses an existing open issue: fixes #6433

## Overview

Per the issue, deprecates the `typeParameter` property in favor of `constraint` and `key` properties.


packages/scope-manager/tests/fixtures/type-declaration/mapped-named.ts.shot
```diff
@@ -13,7 +13,7 @@ ScopeManager {
       ],
```
bradzacher, line 13: i realise now that our tests for named mapped types is lacking.
Could you please add the following two cases:

```ts
type T = 1; // should have 0 references
type M = { [T in string as T]: 1 };
```

```ts
type T = 1; // should have 1 reference
type M = { [K in string as T]: 1 };
```

packages/scope-manager/tests/fixtures/type-declaration/mapped-named.ts.shot
```diff
@@ -95,21 +66,39 @@ ScopeManager {
...
+          isTypeReference: true,
+          isValueReference: false,
+          isWrite: false,
+          resolved: null,
+        },
+        Reference$5,
       ],
-      set: Map {
-        "k" => Variable$4,
-      },
```
bradzacher, line 107: this is a problem - this shouldn't have disappeared!
the type variable `k` is still defined in this scope.

packages/scope-manager/tests/fixtures/type-declaration/mapped.ts.shot
```diff
@@ -87,20 +66,31 @@ ScopeManager {
...
+          isTypeReference: true,
+          isValueReference: false,
+          isWrite: false,
+          resolved: null,
+        },
+        Reference$4,
       ],
-      set: Map {
-        "k" => Variable$4,
-      },
```
bradzacher, line 98: ditto - this shouldn't have disappeared!

packages/visitor-keys/src/visitor-keys.ts
```diff
@@ -212,7 +212,7 @@ const additionalKeys: AdditionalKeys = {
   TSIntersectionType: ['types'],
   TSIntrinsicKeyword: [],
   TSLiteralType: ['literal'],
-  TSMappedType: ['nameType', 'typeParameter', 'typeAnnotation'],
+  TSMappedType: ['constraint', 'key', 'nameType', 'typeAnnotation'],
```
bradzacher, line 215: I think changing this order is what borked the scope analysis.

The ordering of keys should be the ordering of the nodes in the AST.

So it should be `['key', 'constraint', 'nameType', 'typeAnnotation']`

It looks like we used to have a bug in our code because the `nameType` is allowed to reference the `key` (eg `type U = {[T in string as T]: T}` is valid TS!)

bradzacher (changes requested):
this change introduces a bug into the scope analysis because previously it used to implicitly rely on the fact that `typeParameter` was a `TSTypeParemeter` - i.e. declared a type variable.

So now we'll need to update our visitor so that we declare the key as a type variable, then visit the constraint, then the name type, then the type annotation.

https://github.com/typescript-eslint/typescript-eslint/blob/2f46341a14ca6f05334301315bc54dd20129c408/packages/scope-manager/src/referencer/TypeVisitor.ts#L198-L203

packages/scope-manager/tests/fixtures/type-declaration/mapped-named.ts.shot
JoshuaKGoldberg: Aha! Nice spot, thanks.

packages/scope-manager/src/referencer/TypeVisitor.ts
```diff
@@ -197,7 +197,21 @@ class TypeVisitor extends Visitor {
...
+    this.#referencer
+      .currentScope()
+      .defineIdentifier(
+        node.key,
+        new TypeDefinition(
+          node.key,
+          node.parent as
+            | TSESTree.TSInterfaceDeclaration
+            | TSESTree.TSTypeAliasDeclaration
+            | TSESTree.TSTypeParameter,
```
JoshuaKGoldberg, line 209: I added #6225 to the v7 milestone. It's never too early to get excited about the next major version! 😄

packages/scope-manager/src/referencer/TypeVisitor.ts
```diff
@@ -197,7 +197,21 @@ class TypeVisitor extends Visitor {
...
+        ),
+      );
+    this.visit(node.constraint);
+    this.visit(node.nameType);
+    this.visit(node.typeAnnotation);
```
JoshuaKGoldberg, line 214: @bradzacher I'm low confidence I understood this right... going off the other type visitors with identifier/key-like nodes, this seemed like the right way to go?...

packages/scope-manager/tests/fixtures/type-declaration/mapped-named.ts.shot
```diff
@@ -41,7 +41,6 @@ ScopeManager {
       defs: [
         TypeDefinition$3 {
           name: Identifier<"k">,
-          node: TSTypeParameter$3,
```
JoshuaKGoldberg, line 44: ...except, I suspect the removal of the type parameter node from the type definition is probably wrong?

JoshuaKGoldberg:
We didn't make it in time. I'll move this to v7.

JoshuaKGoldberg:

JoshuaKGoldberg:
Or, actually, since this needs to wait till v7 anyway - I'll just leave it closed.

packages/typescript-estree/src/convert.ts
```diff
@@ -3518,6 +3527,45 @@ export class Converter {
...
+        Object.defineProperty(node, deprecatedKey, {
+          enumerable: true,
+          writable: true,
+          value,
+        });
+      },
+    });
+
+    return node as Properties & Record<Key, Value>;
+  }
```
JoshuaKGoldberg, line 3567: Intentionally the same new method as in #8920.

packages/eslint-plugin/src/rules/indent.ts
```diff
@@ -382,6 +382,7 @@ export default createRule<Options, MessageIds>({
       TSMappedType(node: TSESTree.TSMappedType) {
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         const squareBracketStart = context.sourceCode.getTokenBefore(
+          // eslint-disable-next-line deprecation/deprecation
```
JoshuaKGoldberg, line 385: Since this rule is getting removed in #8074 anyway.

packages/scope-manager/src/referencer/TypeVisitor.ts
```diff
@@ -199,7 +199,21 @@ class TypeVisitor extends Visitor {
...
+    this.#referencer
+      .currentScope()
+      .defineIdentifier(
+        node.key,
+        new TypeDefinition(
+          node.key,
+          node.parent as
+            | TSESTree.TSInterfaceDeclaration
+            | TSESTree.TSTypeAliasDeclaration
+            | TSESTree.TSTypeParameter,
```
bradzacher, line 211: RC: the `TypeDefinition` class should be expanded to include `TSMappedType` here:
https://github.com/typescript-eslint/typescript-eslint/blob/210e2810f6e83be0f3b565cb16bf476a05d945dd/packages/scope-manager/src/definition/TypeDefinition.ts#L8-L10

```suggestion
          node,
```

packages/scope-manager/src/referencer/TypeVisitor.ts
bradzacher: First up - we need to visit things in the order that makes sense from a variable reference POV.

The previous code just did `visitChildren`. This meant that the visitor would work off of the ordering defined in the `visitor-keys`, eg: `TSMappedType: ['nameType', 'typeParameter', 'typeAnnotation'],` [1] and `typeParameter` defines a variable for the `.name` and then manually visits the `.constraint` then the `.default`.

[1] Note: I don't remember why I put `nameType` first in #2748... It seems wrong 🤔 Not sure...

---

You've done a good job of re-interpreting the existing code.
But to explain what we do here in the scope analyser.

1) Visit the mapped type
2) We know it has one index signature and nothing but one index signature...
3) That signature defines a type variable (the "key") and that variable can be referenced later in the mapped type - so we need to define a type variable can be referenced by the "constraint" and the "name type".
4) then we visit the children and do whatever needs to be done for them.
    - Because we had to pull apart the `key` to create a variable - we can't just `visitChildren` automatically and have to instead manually visit each child.

So your logic is all sound.

In terms of the ordering of these calls - you've defined it in source order and that should be all good. There's no way for the `constraint` or `nameType` to declare a "variable" that can be referenced later - which means that the ordering doesn't explicitly matter - so source ordering is the best choice.

packages/scope-manager/tests/fixtures/type-declaration/mapped-named.ts.shot
bradzacher: yeah this should still be there - everything must have a `node`.
This is because of your code here:
https://github.com/JoshuaKGoldberg/typescript-eslint/blob/210e2810f6e83be0f3b565cb16bf476a05d945dd/packages/scope-manager/src/referencer/TypeVisitor.ts#L208-L211

In the scope analyser we don't actually have access to the `.parent` cos that property is actually set by ESLint during traversal! The AST we're analysing only goes down.

So `node.parent` is `undefined` - hence this snapshot serialises without its `node`.

packages/scope-manager/tests/fixtures/type-declaration/mapped.ts.shot
```diff
@@ -41,7 +41,6 @@ ScopeManager {
       defs: [
         TypeDefinition$3 {
           name: Identifier<"k">,
-          node: TSTypeParameter$3,
```
bradzacher, line 44: ditto to above

bradzacher (changes requested):
overall looks good - one definite bug to fix though.

packages/scope-manager/src/referencer/TypeVisitor.ts
JoshuaKGoldberg: Ok cool thanks, this makes sense I think!

packages/scope-manager/tests/fixtures/type-declaration/mapped-named.ts.shot
JoshuaKGoldberg: 👍 makes sense. And now it's back to existing, this time as the `TSMappedType$*` itself. Great!

bradzacher (approved):

PR #7110, opened 2023-06-16
feat(eslint-plugin): apply final v6 changes to configs

### LINKED ISSUE #6759: Configs: Apply changes to config presets for v6

### Description

https://github.com/typescript-eslint/typescript-eslint/discussions/6014 discusses breaking changes to config presets for the v6 branch. We've applied most of those changes in the `v6` branch already. But there may be some changes to those changes that come out of the discussion.

At the very least:

* #5728: was merged into `main` after `v6` was released in beta
* #6110: at time of filing this issue, not yet merged 

Filing this tracking issue to make sure we apply any changes we need before v6 is released as stable.

### Impacted Configurations

All of them! ⚡ 

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6759

## Overview

Applies the final set of changes to preset configs for v6, as discussed in https://github.com/typescript-eslint/typescript-eslint/discussions/6014.


packages/eslint-plugin/src/rules/consistent-type-exports.ts
```diff
@@ -84,6 +84,7 @@ export default util.createRule<Options, MessageIds>({
       const symbol = services.getSymbolAtLocation(specifier.exported);
       const aliasedSymbol = checker.getAliasedSymbol(symbol!);
 
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
       if (!aliasedSymbol || aliasedSymbol.escapedName === 'unknown') {
```
JoshuaKGoldberg, line 88: Lots of complaints from `no-unsafe-enum-comparison`, and I think there's a bug in the rule... ran out of time today to investigate though.

JoshuaKGoldberg: Just to be clear, I think this PR is still ready to review despite these annoying `eslint-disable-next-line`s. If it is a bug in the rule then I'll file an issue & send a PR resolving the bug.

.eslintrc.js
```diff
@@ -52,6 +52,10 @@ module.exports = {
     // make sure we're not leveraging any deprecated APIs
     'deprecation/deprecation': 'error',
 
+    // Pending: https://github.com/typescript-eslint/typescript-eslint/issues/4820
+    // (too many false positives)
+    '@typescript-eslint/prefer-nullish-coalescing': 'off',
```
bradzacher, line 57: > too many false positives

What false positives? The rule is type-aware, isn't it?
We already had it turned on as an error as well didn't we?

packages/eslint-plugin/src/rules/no-base-to-string.ts
```diff
@@ -5,7 +5,7 @@ import * as ts from 'typescript';
 import * as util from '../util';
 
 enum Usefulness {
-  Always,
+  Always = 'always',
   Never = 'will',
   Sometimes = 'may',
 }
```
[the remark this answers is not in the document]
bradzacher: I think making it `strict` is probably better. It does seem like a lot of cases would be "okay" with mixed enums.

packages/eslint-plugin/src/rules/no-empty-interface.ts
```diff
@@ -84,6 +84,7 @@ export default util.createRule<Options, MessageIds>({
 
             const isInAmbientDeclaration = !!(
               util.isDefinitionFile(filename) &&
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
               scope.type === 'tsModule' &&
```
bradzacher, line 88:
```suggestion
              scope.type === ScopeType.tsModule &&
```

this is a legit error

packages/eslint-plugin/src/rules/consistent-type-exports.ts
bradzacher: so `.escapedName` is typed as [`ts.__String`](https://github.com/microsoft/TypeScript/blob/5021dd87ef36e87bae39cbb5a30dabd295410920/src/compiler/types.ts#L5981):
```ts
export type __String =
  | (string & { __escapedIdentifier: void; })
  | (void & { __escapedIdentifier: void })
  | InternalSymbolName;
```

Note the `InternalSymbolName` type in that union - [that is an enum](https://github.com/microsoft/TypeScript/blob/5021dd87ef36e87bae39cbb5a30dabd295410920/src/compiler/types.ts#L5953-L5971)!

The rule has handling for some cases, but it doesn't appear to detect branded types. Filed https://github.com/typescript-eslint/typescript-eslint/issues/7114

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -126,7 +126,7 @@ function specifierNameMatches(type: ts.Type, name: string | string[]): boolean {
   if (symbol === undefined) {
     return false;
   }
-  return name.some(item => item === symbol.escapedName);
+  return name.some(item => (item as ts.__String) === symbol.escapedName);
```
bradzacher, line 129: I think just keep using a disable comment for these cases and that way when we fix https://github.com/typescript-eslint/typescript-eslint/issues/7114 we'll be able to just remove the unnecessary disables

packages/type-utils/src/isTypeReadonly.ts
```diff
@@ -86,7 +86,7 @@ function isTypeReadonlyArrayOrTuple(
       ESLintUtils.NullThrowsReasons.MissingToken('symbol', 'array type'),
     );
     const escapedName = symbol.getEscapedName();
-    if (escapedName === 'Array') {
+    if (escapedName === ('Array' as ts.__String)) {
```
bradzacher, line 89: ditto

packages/scope-manager/src/ScopeManager.ts
```diff
@@ -109,6 +109,7 @@ class ScopeManager {
    */
   public acquire(node: TSESTree.Node, inner = false): Scope | null {
     function predicate(testScope: Scope): boolean {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
       if (testScope.type === 'function' && testScope.functionExpressionScope) {
```
bradzacher, line 113:
```suggestion
      if (testScope.type === ScopeType.function && testScope.functionExpressionScope) {
```

packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -97,6 +97,7 @@ class UnusedVarsVisitor<
       const scope = this.#scopeManager.acquire(node, inner);
 
       if (scope) {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
         if (scope.type === 'function-expression-name') {
```
bradzacher, line 101:
```suggestion
        if (scope.type === ScopeType.functionExpressionName) {
```

packages/eslint-plugin/src/rules/no-shadow.ts
```diff
@@ -364,6 +364,7 @@ export default util.createRule<Options, MessageIds>({
     ): TSESLint.Scope.Scope | null {
       const upper = scope.upper;
 
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
       if (upper?.type === 'function-expression-name') {
```
bradzacher, line 368:
```suggestion
      if (upper?.type === ScopeType.functionExpressionName) {
```

packages/eslint-plugin/src/rules/no-redeclare.ts
```diff
@@ -255,6 +255,7 @@ export default util.createRule<Options, MessageIds>({
 
         // Node.js or ES modules has a special scope.
         if (
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
           scope.type === 'global' &&
```
bradzacher, line 259:
```suggestion
          scope.type === ScopeType.global &&
```

bradzacher (approved):

.eslintrc.js
JoshuaKGoldberg: ...huh. I could have sworn I received many errors until I added this line. Never mind!

packages/eslint-plugin/src/rules/restrict-plus-operands.ts
```diff
@@ -72,19 +72,24 @@ export default util.createRule<Options, MessageIds>({
   },
   defaultOptions: [
     {
+      allowAny: true,
+      allowBoolean: true,
+      allowNullish: true,
+      allowNumberAndString: true,
+      allowRegExp: true,
```
bradzacher, line 79: @JoshuaKGoldberg i don't remember our discussion here - why did we turn all of the options on?
I thought we discussed that turning all the rule options on was bad as it made the rule effectively useless.

bradzacher: Your example is template strings - but this is restrict-plus-operands.

So was this an accidental mix-up?

JoshuaKGoldberg: I think it applies to both of them. I've seen both happen with users. Which is why I keep mixing them up in comments 😅 to me, they're similar -just not exactly the same- cases.

bradzacher: I guess the question is - why even have these turned on at all if all the options are on by default?
The rules enforce pretty nothing like this so it's just wasted lint time.

[Eg for RPO there's the following cases](https://typescript-eslint.io/play/#ts=5.1.6&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3pmAubGMAligOYDcAUKJLAsuligK4C2%2BLrARiDFTdDiJUmAjAByzCBA5SI8AD5iSFauEH0RTNpOmzpi%2BJx581tIQ1FckSGfGu2QUFP3V1hjeEi4ArfAG8AX1dzTU9nAE98SJCND1FmFABrFCQAdxR8RJT0l0pKHHgAajEqTmKjNioCwgrOavKSzl0IMrYKnBaqQpLOuW7app1%2B-Mb7G1aamAqHSdmZiYb2ku8fcngAeg34GAAHJHgMAAsoLCQAN15oXbR4ABUAZXheGCQYKYrV9a2d-dHl%2BAxD4lIGUMbZVIZb7bPYHY6nLyXGDXW6PZ4wV7vHrwCG5aG-JCUIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6aFfKsu4%2BWZLQD2paAEMmAE1QZIiaNCHRI4MAF8QaoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false). Of these 16 cases the default options mean the rule only matches FOUR cases - two of which are TS already errors.

So we're recommending a rule to everyone to stop them from doing: `string + object` and `string + unknown`. The former is also banned under `no-base-to-string`, and one could argue the latter should be allowed under the `allowAny` option.

tl;dr - why are we even turning on the rules in the recommended set if they don't do anything with an overly permissive default option?

JoshuaKGoldberg: A couple reasons swayed me:

* (lesser) There is benefit to `string + unknown`
* (greater) I've seen a projects turn off `no-base-to-string` without turning off `restrict-plus-operands`

From a performance perspective, if one of these two rules is already asking for the types of these, I some level of caching on the TypeScript side would make the other rule not a significant change.

bradzacher: The thing I'd point to is that we added this big caution block to the RPO docs because we didn't think people using the options was good - and yet it's the default that everyone uses these options?
https://typescript-eslint.io/rules/restrict-plus-operands#:~:text=%2C%0A%5D%3B-,CAUTION,-We%20generally%20recommend

Just seems like we're contradicting ourselves "don't use these options but also they're all turned on by default"

JoshuaKGoldberg: Yeah :/ agreed. I don't like this.

bradzacher: Personal preference:
Given this rule is really doing nothing and a lot of people don't like it - we should just undo this change and only include it in `strict-type-checked`

Issue #723, opened 2019-07-18
[prefer-destructuring] Variable declaration type annotation


ark120202:
```ts
const foo = { bar: 1 as const };
const bar: number = foo.bar;
// Currently fixed to
const { bar } = foo;
```
1. Fixer should change code to `const { bar }: { bar: number } = foo`, removing type annotation is unsafe
1. I think it would make sense to add an option to disable rule in this case, annotations on destructured variables aren't very pretty

armano2:
[Playground](https://typescript-eslint.io/play#ts=4.7.2&sourceType=module&code=MYewdgzgLgBAZiEMC8MDeMBGBDATgLhgEYZsIZRJYBfAbgFgAoS6LPQsAVwFtMBTXCniIAdDlwNGAeikwAwp1y4+YKABsAnvACWADz4ATGFBBMWsDOJjUhCELSA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQA7UQDNFoBaAExQBc4BjK2aASwDsBzdKE6Ae2knBgAviCFA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA)

### Code

```ts
const foo = { bar: 1 as const };
const bar: number = foo.bar;
// Currently fixed to
const { bar } = foo;
```

### Eslint config

```json
{
  "rules": {
    "prefer-destructuring": "error"
  }
}
```

HolgerJeromin:
Can this issue be closed? ref PR #7117

Issue #1931, opened 2020-04-24
New Rule: `prefer-destructuring`


0xCLARITY:
I'd like to use the [`prefer-destructuring`](https://eslint.org/docs/rules/prefer-destructuring) rule from ESLint, however, when I try to use array destructuring with an Express request object, I get the following TypeScript Error:

`Type 'ParamsDictionary' must have a '[Symbol.iterator]()' method that returns an iterator.ts(2488)`.

That seems reasonable, but then the base ESLint `prefer-destructuring` rule gives me a warning, even though TypeScript claims I'm not allowed to use array destructuring here.

I think having a type-aware version of this rule would be beneficial.

bradzacher:
could you please give me a self-contained example?

hbergren:
Does this work?

```ts
import { Request, Response, NextFunction } from 'express'

async function example(
  req: Request,
  _res: Response,
  _next: NextFunction,
): Promise<void> {
  // Type 'ParamsDictionary' must have a '[Symbol.iterator]()' method that returns an iterator
  const [param] = req.params

  console.log(param)
}
```

bradzacher:
what was the code _before_ the rule fixed it?
`const param = req.params[0]`?

hbergren:
Yes, which the base ESLint `prefer-destructuring` rule warns against.

bradzacher:
Thanks.
Happy to accept a PR if you'd like to add a type-aware version of the rule to prevent array destructuring of non-iterable types.
In that case, could also implement a fixer, as we know for sure the type is fixable.

phaux:
The correct way to destructure this is
```js
const { 0: param } = req.params
```
because `params` is just an object with numeric properties.

danvk:
Another issue with the eslint `prefer-destructuring` rule in TS is if you apply a type declaration to the resulting variable:

```ts
const obj = {xs: ['a', 'b', 'c']};  // type is {xs: string[]}
const xs: readonly string[] = obj.xs;
xs.push('d');
// ~~~~ Property 'push' does not exist on type 'readonly string[]'.
```

Line two gets flagged by eslint's rule:

![image](https://user-images.githubusercontent.com/98301/90832432-c0fe5a80-e313-11ea-9d19-9f0da3208684.png)

However, applying the quick fix removes the type annotation, an unsafe operation which makes the `push` call pass the type checker:

```ts
[...]

PR #7117, opened 2023-06-18
feat(eslint-plugin): add new extended rule `prefer-destructuring`

seiyab:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/723  https://github.com/typescript-eslint/typescript-eslint/issues/1931

## Overview
Add a new base rule extension `@typescript-eslint/prefer-destructuring` to resolve following issues.
- https://github.com/typescript-eslint/typescript-eslint/issues/723
- https://github.com/typescript-eslint/typescript-eslint/issues/1931

https://github.com/typescript-eslint/typescript-eslint/issues/723 requests two things: 1. keep type annotation, 2. add an option to allow non-destructuring assignment for type annotated ones.
For the first one, I just disable auto fixer for declaration with type annotation, rather than providing correct auto fix. It's just because difficulty and maitainability.
For the second one, I add `enforceForDeclarationWithTypeAnnotation` option. I set default value `false` because I agree with "annotations on destructured variables aren't very pretty" in the issue.

https://github.com/typescript-eslint/typescript-eslint/issues/1931 requests type-aware version and I have implemented it for numeric index access for non-iterable object, that is the only case reported in the issue (excluding comments that are duplication of https://github.com/typescript-eslint/typescript-eslint/issues/723).
There is a notable thing for it. I treated numeric index access for non-iterable object as renamed one. So it is affected by `enforceForRenamedProperties` option.


packages/eslint-plugin/docs/rules/prefer-destructuring.md
```diff
@@ -0,0 +1,91 @@
...
+description: 'Require destructuring from arrays and/or objects.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/prefer-destructuring** for documentation.
+
+## Examples
+
+This rule extends the base [`eslint/prefer-destructuring`](https://eslint.org/docs/rules/prfer-destructuring) rule.
```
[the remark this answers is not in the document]
seiyab: I just copied & pasted current format. I changed it to the final url.

packages/eslint-plugin/src/rules/prefer-destructuring.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+function isTypeIterableType(
+  type: ts.Type,
+  typeChecker: ts.TypeChecker,
+): boolean {
+  if (isTypeAnyType(type)) {
+    return true;
+  }
+  if (!type.isUnion()) {
```
JoshuaKGoldberg, line 211: [Cleanup] Interestingly, `require-await.ts` already has some code around this pattern of checking all types for a well known symbol property:

https://github.com/typescript-eslint/typescript-eslint/blob/3cdf5c93163036fdc60325129d9746ff57574336/packages/eslint-plugin/src/rules/require-await.ts#L120-L122

https://github.com/typescript-eslint/typescript-eslint/blob/3cdf5c93163036fdc60325129d9746ff57574336/packages/eslint-plugin/src/rules/require-await.ts#L237-L242

What do you think about making a shared helper to deduplicate some of the code? Think it's worth it?

(not a blocker at all)

packages/eslint-plugin/src/rules/prefer-destructuring.ts
```diff
@@ -0,0 +1,232 @@
...
+    },
+  });
+}
+
+function isTypeIterableType(
+  type: ts.Type,
+  typeChecker: ts.TypeChecker,
+): boolean {
+  if (isTypeAnyType(type)) {
+    return true;
```
JoshuaKGoldberg, line 209: [Question] We can't assume that `any`s are or are not iterable. Seeing this return `true` confused me. Let's either return `false` or rename the function to `isTypeAnyOrIterableType`.

packages/eslint-plugin/tests/rules/prefer-destructuring.test.ts
```diff
@@ -0,0 +1,1364 @@
...
+      options: [{ VariableDeclarator: { object: false } }],
+    },
+    {
+      code: "var foo = object['foo'];",
+      options: [{ VariableDeclarator: { object: false } }],
+    },
+    '({ foo } = object);',
+    {
+      // Fix #8654
+      code: 'var foo = array[0];',
```
JoshuaKGoldberg, line 628: [Bug] A lot of these tests use variables that aren't declared, but interact with source logic that uses the type checker. For example, arrays care about type checking from `isTypeIterableType`. The tests will need to be fixed up to give variables proper types.

Example:

```suggestion
      code: `
        declare const array: string[];
        var foo = array[0];
      `,
```

JoshuaKGoldberg (changes requested):
A great start, thanks for working on this @seiyab! 🚀 

I suspect once the tests are cleaned up for type checking, there might be some rule code changes. So I can take a deeper look once those are fixed up.

JoshuaKGoldberg:
I also took the liberty of merging in from `main`, since `v6` was merged & changed a few internals after this PR was sent. Cheers!

packages/eslint-plugin/src/rules/prefer-destructuring.ts
```diff
@@ -0,0 +1,232 @@
...
+
+const baseRule = getESLintCoreRule('prefer-destructuring');
+
+type BaseOptions = InferOptionsTypeFromRule<typeof baseRule>;
+type FullBaseOptions = BaseOptions & [unknown, unknown];
+type Options0 = FullBaseOptions[0];
+type Options1 = FullBaseOptions[1] & {
+  enforceForDeclarationWithTypeAnnotation?: boolean;
+};
+type Options = [Options0, Options1?];
```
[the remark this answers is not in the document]
seiyab: It makes sense. I committed the suggestion. Thanks.

packages/eslint-plugin/src/rules/prefer-destructuring.ts
seiyab: I think we can't use `expandUnionOrIntersectionType` is not suitable for `prefer-destructuring`.
If we use it, it returns same results for types both like `object & unknown[]` and `object | unknown[]` though former is iterable and latter is not iterable.
I need to inspect `require-await` to make a shared helper. If `expandUnionOrIntersectionType` works properly for `require-await`, maybe we don't need shared helper. Or not, I think making a shared helper looks reasonable and the helper will fix a bug of `require-await`.
I suppose that shared helper should be added in another PR because it will change the behavior `require-await` a bit.

packages/eslint-plugin/tests/rules/prefer-destructuring.test.ts
seiyab: They are regression tests copied from base rule (See [this comment](https://github.com/typescript-eslint/typescript-eslint/pull/7117/files/521506c2fecdbd6ac652879b9debedb2854107a3#diff-01b15f30c37a23cb5d826daba20a0ea9402405002adf703c7dc46d98fc875fe3R544-R548)).
So I intentionally keep them same as original ones.
Should I declare them?

JoshuaKGoldberg: Hmm, unfortunate to have to fork the base rule's tests this way. But yes they do need to change. The base rule's tests weren't written with TypeScript type checking in mind. Our adding in type checking means they may no longer be valid tests.

.cspell.json
```diff
@@ -89,6 +89,7 @@
     "nullish",
     "OOM",
     "OOMs",
+    "openjsf",
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Ah, I'd missed that this is in the not-expanded-by-default large tests file. Never mind!

packages/eslint-plugin/tests/rules/prefer-destructuring.test.ts
```diff
@@ -0,0 +1,1364 @@
...
+const ruleTester = new RuleTester({
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    sourceType: 'module',
+    tsconfigRootDir: rootPath,
+    project: './tsconfig.json',
+  },
+});
+
+ruleTester.run('prefer-desctructuring', rule, {
```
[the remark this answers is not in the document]
seiyab: I have fixed it. https://github.com/typescript-eslint/typescript-eslint/pull/7117/commits/4d69749905db386f343bce1ab1f96105fbfc728c

packages/eslint-plugin/tests/rules/prefer-destructuring.test.ts
seiyab: I see. I will write from scratch ✏️

packages/eslint-plugin/src/rules/prefer-destructuring.ts
```diff
@@ -0,0 +1,232 @@
...
+function isArrayLiteralIntegerIndexAccess(
+  node: TSESTree.Expression,
+): node is TSESTree.MemberExpression {
+  if (node.type !== AST_NODE_TYPES.MemberExpression) {
+    return false;
+  }
+  if (node.property.type !== AST_NODE_TYPES.Literal) {
+    return false;
+  }
+  return Number.isInteger(node.property.value);
```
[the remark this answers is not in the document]
JoshuaKGoldberg: As long as the unit tests test the logic & are passing, all is well!

packages/eslint-plugin/tests/rules/prefer-destructuring.test.ts
seiyab: I have rewritten them.

seiyab:
I have responded to your requests🚀

packages/eslint-plugin/src/rules/prefer-destructuring.ts
```diff
@@ -0,0 +1,239 @@
...
+        object: true,
+      },
+      AssignmentExpression: {
+        array: true,
+        object: true,
+      },
+    },
+    {},
+  ],
+  create(context, [enabledTypes, options1 = {}]) {
```
JoshuaKGoldberg, line 91: [Naming] IMO there's no need to call this `options1`; it's clear from the code context that it's the rule's `options`.

```suggestion
  create(context, [enabledTypes, options = {}]) {
```

packages/eslint-plugin/src/rules/prefer-destructuring.ts
```diff
@@ -0,0 +1,239 @@
...
+      }
+      return enabledTypes[nodeType as keyof typeof enabledTypes][
+        destructuringType as keyof (typeof enabledTypes)[keyof typeof enabledTypes]
+      ];
+    }
+
+    function baseRulesWithoutFix(): ReturnType<typeof baseRule.create> {
+      if (baseRulesWithoutFixCache == null) {
+        baseRulesWithoutFixCache = baseRule.create(noFixContext(context));
+      }
```
JoshuaKGoldberg, line 178: [Nitpick] Just a little cleaner here:

```suggestion
      baseRulesWithoutFixCache ??= baseRule.create(noFixContext(context));
```

I'm actually surprised we don't have https://eslint.org/docs/latest/rules/logical-assignment-operators enabled... #7461

JoshuaKGoldberg (dismissed):
LGTM, thanks for both sending & updating it (with so many tests)! 🔥

I'll defer to an additional review from @typescript-eslint/triage-team because this is the first instance I can find of calling `baseRule.create` in two different potential ways.

seiyab:
Thank you for your review.
I have responded to https://github.com/typescript-eslint/typescript-eslint/pull/7117#discussion_r1292517629 and https://github.com/typescript-eslint/typescript-eslint/pull/7117#discussion_r1292518239 but I'm wondering why `create(context, [enabledTypes, options = {}]) {` get  uncovered renaming a variable.

seiyab:
@JoshuaKGoldberg 
Could you tell me the current status?

JoshuaKGoldberg:
Heh, I was hoping for an additional review, but it seems that everyone is slammed right now. Let's merge as-is 😄. Thanks!

JoshuaKGoldberg (approved):
Large ship with caption 'it's shipping time'

PR #7138, opened 2023-06-25
chore: enabled stylistic-type-checked internally

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: continues #6827

## Overview

Adds `stylistic-type-checked` to the list of `extends` configs and removes manual enables of rules contained in it. Most of the changes seem to be from `no-confusing-void-expression`, followed by `consistent-indexed-object-style`.

A couple rules received special treatment:
* Doesn't enable `prefer-nullish-coalescing` yet, as there were a lot of complaints from it and I'll look into it separately
* `no-empty-function` seemed to be popular for tests, so I moved it to the tests overrides

### DIFF
--- a/packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts
+++ b/packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts
@@ -267,7 +267,7 @@ export default createRule<Options, MessageIds>({
 
       if (literal.loc.end.line === literal.loc.start.line) {
         // don't use template strings for single line tests
-        context.report({
+        return context.report({
           node: literal,
           messageId: 'singleLineQuotes',
           fix(fixer) {
@@ -288,7 +288,6 @@ export default createRule<Options, MessageIds>({
             ];
           },
         });
-        return;
       }
 
       const lines = text.split('\n');
@@ -299,7 +298,7 @@ export default createRule<Options, MessageIds>({
       const isEndEmpty = lastLine.trimStart() === '';
       if (!isStartEmpty || !isEndEmpty) {
         // multiline template strings must have an empty first/last line
-        context.report({
+        return context.report({
           node: literal,
           messageId: 'templateLiteralEmptyEnds',
           *fix(fixer) {
@@ -318,12 +317,11 @@ export default createRule<Options, MessageIds>({
             }
           },
         });
-        return;
       }
 
       const parentIndent = getExpectedIndentForNode(literal, sourceCode.lines);
       if (lastLine.length !== parentIndent) {
-        context.report({
+        return context.report({
           node: literal,
           messageId: 'templateLiteralLastLineIndent',
           fix(fixer) {
@@ -333,7 +331,6 @@ export default createRule<Options, MessageIds>({
             );
           },
         });
-        return;
       }
 
       // remove the empty lines
@@ -349,14 +346,13 @@ export default createRule<Options, MessageIds>({
       const requiresIndent = firstLineIndent.length > 0;
       if (requiresIndent) {
         if (firstLineIndent.length !== expectedIndent) {
-          context.report({
+          return context.report({
             node: literal,
             messageId: 'templateStringRequiresIndent',
             data: {
               indent: expectedIndent,
             },
           });
-          return;
         }
 
         // quick-and-dirty validation that lines are roughly indented correctly
@@ -370,14 +366,13 @@ export default createRule<Options, MessageIds>({
 
           const indent = matches[1];
           if (indent.length < expectedIndent) {
-            context.report({
+            return context.report({
               node: literal,
               messageId: 'templateStringMinimumIndent',
               data: {
                 indent: expectedIndent,
               },
             });
-            return;
           }
         }
 
@@ -402,7 +397,7 @@ export default createRule<Options, MessageIds>({
               .join('\n')
           : formatted;
 
-        context.report({
+        return context.report({
           node: literal,
           messageId: isErrorTest
             ? 'invalidFormattingErrorTest'
@@ -417,7 +412,6 @@ export default createRule<Options, MessageIds>({
             );
           },
         });
-        return;
       }
     }
 
--- a/packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts
+++ b/packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts
@@ -267,7 +267,7 @@ export default createRule<Options, MessageIds>({
 
       if (literal.loc.end.line === literal.loc.start.line) {
         // don't use template strings for single line tests
-        return context.report({
+        context.report({
           node: literal,
           messageId: 'singleLineQuotes',
           fix(fixer) {
@@ -288,6 +288,7 @@ export default createRule<Options, MessageIds>({
             ];
           },
         });
+        return;
       }
 
       const lines = text.split('\n');
@@ -298,7 +299,7 @@ export default createRule<Options, MessageIds>({
       const isEndEmpty = lastLine.trimStart() === '';
       if (!isStartEmpty || !isEndEmpty) {
         // multiline template strings must have an empty first/last line
-        return context.report({
+        context.report({
           node: literal,
           messageId: 'templateLiteralEmptyEnds',
           *fix(fixer) {
@@ -317,11 +318,12 @@ export default createRule<Options, MessageIds>({
             }
           },
         });
+        return;
       }
 
       const parentIndent = getExpectedIndentForNode(literal, sourceCode.lines);
       if (lastLine.length !== parentIndent) {
-        return context.report({
+        context.report({
           node: literal,
           messageId: 'templateLiteralLastLineIndent',
           fix(fixer) {
@@ -331,6 +333,7 @@ export default createRule<Options, MessageIds>({
             );
           },
         });
+        return;
       }
 
       // remove the empty lines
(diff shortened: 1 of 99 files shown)


packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts
```diff
@@ -397,7 +402,7 @@ export default createRule<Options, MessageIds>({
               .join('\n')
           : formatted;
 
-        return context.report({
```
bradzacher, line 400: which rule suggested this?
TBH I prefer the `return report` style here as it's easier to see that the report is terminal.

packages/eslint-plugin/src/rules/naming-convention-utils/format.ts
```diff
@@ -17,26 +17,26 @@ https://gist.github.com/mathiasbynens/6334847
 function isPascalCase(name: string): boolean {
   return (
     name.length === 0 ||
-    (name[0] === name[0].toUpperCase() && !name.includes('_'))
```
bradzacher, line 20: I actually think that the previous form is clearer for this specifica case.
[It's also substantially faster (not that it matters at the this speed... but it is twice as fast - ~1.2B ops/s vs ~2.4B ops/s)](https://jsbench.me/buljbfvcn4/1).

I wonder if we should add an option to skip this case from the rule?

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -657,10 +663,14 @@ export default createRule<Options, MessageId>({
       AssignmentExpression: checkAssignmentExpression,
       BinaryExpression: checkIfBinaryExpressionIsNecessaryConditional,
       CallExpression: checkCallExpression,
-      ConditionalExpression: (node): void => checkNode(node.test),
+      ConditionalExpression: (node): void => {
+        checkNode(node.test);
+      },
```
bradzacher, line 668: TBH I personally don't think that multilining arrows is necessarily clearer.
Is there a reason the stylistic set is pushing for this style?

packages/rule-schema-to-typescript-types/src/index.ts
```diff
@@ -69,9 +69,7 @@ function compileSchema(
 
   const refMap = new Map<string, string>();
   // we only support defs at the top level for simplicity
-  const defs = (schema.$defs ?? schema.definitions) as
-    | Record<string, JSONSchema4>
-    | undefined;
```
bradzacher, line 74: IIRC I did this intentionally because both of the keys were not techincally `undefined` - is that not the case any more?

packages/rule-tester/src/types/DependencyConstraint.ts
```diff
@@ -12,9 +12,4 @@ export type AtLeastVersionConstraint =
 export type VersionConstraint =
   | AtLeastVersionConstraint
   | SemverVersionConstraint;
-export interface DependencyConstraint {
-  /**
-   * Passing a string for the value is shorthand for a '>=' constraint
-   */
-  readonly [packageName: string]: VersionConstraint;
-}
```
bradzacher, line 20: Making this a record is actually less clear due to the lost comment

packages/scope-manager/src/scope/ScopeBase.ts
```diff
@@ -232,7 +232,7 @@ abstract class ScopeBase<
     block: TBlock,
     isMethodDefinition: boolean,
   ) {
-    const upperScopeAsScopeBase = upperScope as Scope;
```
bradzacher, line 235: I'm surprised TS allowed this - the reason this cast was there was because `upperScope` is typed as `TUpper` - and there was a lot of weird behaviour that stemmed from it being a generic instead of the concrete type.

I think there was also some weirdness due to the cyclic nature of the files? I don't fully remmeber though

packages/typescript-estree/tests/lib/convert.test.ts
```diff
@@ -21,73 +21,77 @@ describe('convert', () => {
     );
   }
 
-  it('deeplyCopy should convert node correctly', () => {
-    const ast = convertCode('type foo = ?foo<T> | ?(() => void)?');
+  /* eslint-disable @typescript-eslint/dot-notation */
```
bradzacher, line 24: why is this disabled?

packages/typescript-estree/tests/lib/persistentParse.test.ts
```diff
@@ -148,47 +176,69 @@ function baseTests(
...
     // parse once to: assert the config as correct, and to make sure the program is setup
-    expect(() => parseFile('foo', PROJECT_DIR)).not.toThrow();
+    expect(() => {
+      parseFile('foo', PROJECT_DIR);
+    }).not.toThrow();
     // bar should throw because it doesn't exist yet
-    expect(() => parseFile(bazSlashBar, PROJECT_DIR)).toThrow();
+    expect(() => {
+      parseFile(bazSlashBar, PROJECT_DIR);
+    }).toThrow();
```
bradzacher, line 195: this is the sort of reason I think that the concise arrow function style is better - satisfying this lint rule just added some 160 lines to this file and (IMO) made the code *harder* to read

packages/utils/src/ts-eslint/Linter.ts
```diff
@@ -130,12 +130,8 @@ namespace Linter {
   export type GlobalVariableOptionBase = 'off' | 'readonly' | 'writable';
   export type GlobalVariableOption = GlobalVariableOptionBase | boolean;
 
-  export interface GlobalsConfig {
-    [name: string]: GlobalVariableOption;
-  }
-  export interface EnvironmentConfig {
-    [name: string]: boolean;
-  }
```
bradzacher, line 138: I believe that these should stay as interfaces to allow for global augmentation from consumers.

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -428,10 +429,9 @@ interface RuleListenerBaseSelectors {
 type RuleListenerExitSelectors = {
   [K in keyof RuleListenerBaseSelectors as `${K}:exit`]: RuleListenerBaseSelectors[K];
 };
-interface RuleListenerCatchAllBaseCase {
-  [nodeSelector: string]: RuleFunction | undefined;
-}
```
bradzacher, line 433: Ditto - interface on purpose for augmentation

packages/utils/src/ts-eslint/SourceCode.ts
```diff
@@ -384,9 +384,7 @@ namespace SourceCode {
     visitorKeys: VisitorKeys | null;
   }
 
-  export interface VisitorKeys {
-    [nodeType: string]: string[];
-  }
```
bradzacher, line 389: I do like the `Record` style more in general, but you can lose some nice implicit documentation sometimes (eg the variable name here).

I'm so torn about using a disable comment for this though because it's so minor.

packages/eslint-plugin/src/rules/naming-convention-utils/format.ts
JoshuaKGoldberg: Agreed. The main 👎 on this change in my head was that the rule is suggesting _longer_ code... #7140

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
JoshuaKGoldberg: Yup it is! +1 to the visibility @rubiesonthesky 😄.

I wonder if it should be moved to the `strict-type-checked` config instead... https://github.com/typescript-eslint/typescript-eslint/discussions/7130#discussioncomment-6285562

JoshuaKGoldberg (commented):
(accidentally started a review, but getting pulled out for something - will continue commenting soon)

packages/rule-schema-to-typescript-types/src/index.ts
JoshuaKGoldberg: Yup! `no-unnecessary-type-assertion` is complaining on them. Both `schema.$defs` and `schema.definitions` are `Record<string, JSONSchema4> | undefined`.

packages/utils/src/ts-eslint/SourceCode.ts
JoshuaKGoldberg: Yeah this is an interesting discussion. I'll undo the changes from `consistent-indexed-object-style` and file them as a todo to look at later, similar to `prefer-nullish-coalescing`.

packages/typescript-estree/tests/lib/convert.test.ts
JoshuaKGoldberg: `deeplyCopy` is a private member. `['deeplyCopy']` bypasses that.

packages/scope-manager/src/scope/ScopeBase.ts
JoshuaKGoldberg: 🤷

packages/typescript-estree/tests/lib/convert.test.ts
```diff
@@ -21,73 +21,77 @@ describe('convert', () => {
...
+  describe('deeplyCopy', () => {
+    it('deeplyCopy should convert node correctly', () => {
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Aha! I'd meant to but must have missed it in a merge. Thanks 😄

PR #7139, opened 2023-06-25
chore: bump to latest lerna and apply custom release entry markdown

JamesHenry:


nx.json
```diff
@@ -1,6 +1,5 @@
 {
   "$schema": "./node_modules/nx/schemas/nx-schema.json",
-  "npmScope": "typescript-eslint",
```
JamesHenry, line 3: This is deprecated and no longer really used in nx, it will derive it from your existing root package

lerna.json
```diff
@@ -1,5 +1,11 @@
 {
+  "$schema": "node_modules/lerna/schemas/lerna-schema.json",
```
JamesHenry, line 2: We added this a while ago, always nice to get IDE autocomplete etc

bradzacher (approved):

bradzacher:
Thanks for cracking onto this so fast

lerna.json
```diff
@@ -1,5 +1,11 @@
...
   "version": "5.60.0",
   "npmClient": "yarn",
-  "stream": true
+  "stream": true,
+  "command": {
+    "version": {
+      "changelogEntryAdditionalMarkdown": "\nOn our website you can read about our [versioning strategy](https://main--typescript-eslint.netlify.app/users/versioning) and [releases](https://main--typescript-eslint.netlify.app/users/releases)"
```
Josh-Cena, line 8: One leading `\n` is enough to start a new paragraph, right?

I wonder why we are using a netlify domain instead of our own?

Also, missing a period, I guess.

JamesHenry: The custom domain points at the docs that were true for the latest release, but I think we always want to point folks to the freshest copy of the docs (the latest commit on main) when it comes to releases and versioning?

Yes we probably don’t even need the leading newline, just added it to be safe. We can amend it after the first time it is used tomorrow

Josh-Cena (approved):
LGTM!

JoshuaKGoldberg (approved):
🔥

PR #7191, opened 2023-07-11
chore: bump dependency versions

bradzacher:
## Overview

This just does a bump of all our dependencies to pin their versions to the latest in all packages except the website packages.

The remaining outdated packages are intentionally outdated ones

image

- nx deps to be upgraded later (cc @JamesHenry)
  - @nrwl/jest      
  - @nrwl/nx-cloud  
  - @nrwl/workspace 
  - @nx/jest        
  - @nx/linter      
  - @nx/workspace
  - nx
- prettier has a major to work through
- `@types/node` - TBH we should probably downgrade this to v16 - given that's the lowest version we support. For now I just left it at the highest v18
- `ts-node` because of the bug (we need to finish migrating off this to tsx tbh)


JamesHenry:
Did you use nx migrate for this?

bradzacher:
> Did you use nx migrate for this?

@JamesHenry I haven't bumped `nx` in this PR - I figured we could do that separately given that might need nx config changes too.

JamesHenry:
D’oh sorry I misread your description - yeah we always wanna use Nx migrate for that reason. I can do that in a follow up later

packages/ast-spec/src/declaration/TSImportEqualsDeclaration/fixtures/external-module-ref-string/snapshots/5-AST-Alignment-AST.shot
```diff
@@ -52,8 +52,7 @@ exports[`AST Fixtures declaration TSImportEqualsDeclaration external-module-ref-
         },
       },
     ],
--   sourceType: 'module',
-+   sourceType: 'script',
```
JoshuaKGoldberg, line 56: ...cool.

packages/repo-tools/src/postinstall.mts
```diff
@@ -1,5 +1,12 @@
-import * as execa from 'execa';
+import { $ as $_config } from 'execa';
```
JoshuaKGoldberg, line 1: Nit: I might call this `$config`...

patches/@microsoft+api-extractor+7.36.1.patch
```diff
@@ -1,12 +1,13 @@
...
+index 31b46f8..e7081c6 100644
 --- a/node_modules/@microsoft/api-extractor/lib/api/ExtractorConfig.js
 +++ b/node_modules/@microsoft/api-extractor/lib/api/ExtractorConfig.js
-@@ -669,5 +669,5 @@ ExtractorConfig.FILENAME = 'api-extractor.json';
+@@ -668,6 +668,6 @@ ExtractorConfig.FILENAME = 'api-extractor.json';
   */
  ExtractorConfig._tsdocBaseFilePath = path.resolve(__dirname, '../../extends/tsdoc-base.json');
  ExtractorConfig._defaultConfig = node_core_library_1.JsonFile.load(path.join(__dirname, '../schemas/api-extractor-defaults.json'));
 -ExtractorConfig._declarationFileExtensionRegExp = /\.d\.ts$/i;
 +ExtractorConfig._declarationFileExtensionRegExp = /\.d\.(c|m)?ts$/i;
```
JoshuaKGoldberg, line 10: Just bumped https://github.com/microsoft/rushstack/pull/4023#issuecomment-1631687085

JoshuaKGoldberg (approved):
Cartoony/drawn ship choppily animated to be bouncing on some waves

PR #7202, opened 2023-07-12
feat(typescript-estree): throw on invalid update expressions

### LINKED ISSUE #7196: Enhancement: Throw errors for invalid `UpdateExpression`s

### Relevant Package

parser

### Description

These expressions should be invalid.

```js
1 ++;

(1)++;
(1)--;

++(1);
--(1);

(1 + 2)++;
(1 + 2)--;

++(1 + 2);
--(1 + 2);

(x + x)++;
(x + x)--;

++(x + x);
--(x + x);
```

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7196

## Overview

We should be able to report on things like `a() = 1` and `1 = 1` as well, but I didn't do this because things like `{ a } = 1` makes it a little more complicated.


bradzacher:
I had a look at the TS codebase and it's a bit messy with regard to this class of errors.

Here's a few snippets I can find:

I can't link the `checker.ts` file in GH cos it's too big but here are the relevant lines:
```ts
    function checkReferenceExpression(expr: Expression, invalidReferenceMessage: DiagnosticMessage, invalidOptionalChainMessage: DiagnosticMessage): boolean {
        // References are combinations of identifiers, parentheses, and property accesses.
        const node = skipOuterExpressions(expr, OuterExpressionKinds.Assertions | OuterExpressionKinds.Parentheses);
        if (node.kind !== SyntaxKind.Identifier && !isAccessExpression(node)) {
            error(expr, invalidReferenceMessage);
            return false;
        }
        if (node.flags & NodeFlags.OptionalChain) {
            error(expr, invalidOptionalChainMessage);
            return false;
        }
        return true;
    }
```
https://github.com/microsoft/TypeScript/blob/94f03cf0c6eb5cb3a391e817db9c9a7bb9f1de6c/src/compiler/utilities.ts#L7877-L7880

and

```ts
    function checkReferenceAssignment(target: Expression, sourceType: Type, checkMode?: CheckMode): Type {
        const targetType = checkExpression(target, checkMode);
        const error = target.parent.kind === SyntaxKind.SpreadAssignment ?
            Diagnostics.The_target_of_an_object_rest_assignment_must_be_a_variable_or_a_property_access :
            Diagnostics.The_left_hand_side_of_an_assignment_expression_must_be_a_variable_or_a_property_access;
        const optionalError = target.parent.kind === SyntaxKind.SpreadAssignment ?
            Diagnostics.The_target_of_an_object_rest_assignment_may_not_be_an_optional_property_access :
            Diagnostics.The_left_hand_side_of_an_assignment_expression_may_not_be_an_optional_property_access;
        if (checkReferenceExpression(target, error, optionalError)) {
            checkTypeAssignableToAndOptionallyElaborate(sourceType, targetType, target, target);
        }
        if (isPrivateIdentifierPropertyAccessExpression(target)) {
            checkExternalEmitHelpers(target.parent, ExternalEmitHelpers.ClassPrivateFieldSet);
        }
        return sourceType;
    }
    function checkReferenceExpression(expr: Expression, invalidReferenceMessage: DiagnosticMessage, invalidOptionalChainMessage: DiagnosticMessage): boolean {
        // References are combinations of identifiers, parentheses, and property accesses.
        const node = skipOuterExpressions(expr, OuterExpressionKinds.Assertions | OuterExpressionKinds.Parentheses);
        if (node.kind !== SyntaxKind.Identifier && !isAccessExpression(node)) {
            error(expr, invalidReferenceMessage);
            return false;
        }
        if (node.flags & NodeFlags.OptionalChain) {
            error(expr, invalidOptionalChainMessage);
            return false;
        }
        return true;
    }
```
https://github.com/microsoft/TypeScript/blob/94f03cf0c6eb5cb3a391e817db9c9a7bb9f1de6c/src/compiler/utilitiesPublic.ts#L1555-L1558

JoshuaKGoldberg (approved):
LGTM - glad you're diving into this @Josh-Cena!

@bradzacher - since you commented in the issue I'll wait for your review for a bit.

bradzacher (changes requested):
These are also semantically valid assignment:
```ts
declare let x: 1;

(x as 1) = 1;
(<1>x) = 1;
(x satisfies 1) = 1;
```
Which is covered by the `skipOuterExpressions(expr, OuterExpressionKinds.Assertions | OuterExpressionKinds.Parentheses);` line in the TS codebase.

Also currently this is invalid ([for now](https://github.com/tc39/proposal-optional-chaining-assignment)):
```ts
declare const y: { z: number };

y?.z = 1;
```

I don't know if we have handling for cases like this (if we don't - do we want to add them in this PR?):
```ts
declare const z: string[];

let [ ...a = [] ] = z;
let { ...b = {} } = z;
```

Other than that looking good!

Josh-Cena:
> ```ts
> let [ ...a = [] ] = z;
> let { ...b = {} } = z;
> ```

This PR only deals with update expressions, not assignment or declaration, so not those atm.

I'll add the assertion ones.

packages/typescript-estree/src/convert.ts
```diff
@@ -1950,6 +1951,12 @@ export class Converter {
          * ESTree uses UpdateExpression for ++/--
          */
         if (operator === '++' || operator === '--') {
+          if (!isValidAssignmentTarget(node.operand)) {
+            this.#throwUnlessAllowInvalidAST(
+              node,
```
fisker, line 1956:
```suggestion
              node.operand,
```

fisker:
@Josh-Cena 

Side question: when I work on invalid AST checks, I feel it's hard to debug, maybe we need find a better way to run test on single fixture, do you feel the same?

bradzacher:
@fisker I did build this into the infra to make it possible to filter the tests:

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/tests/fixtures.test.ts#L21-L25

But I agree that it's not a very good DevX and pretty hacky. Open to suggestions to improve it!

fisker:
I hope we run test from each directory, so we can
`jest packages/ast-spec/src/expression/UpdateExpression/fixtures/_error_/literal/`.

packages/typescript-estree/src/node-utils.ts
```diff
@@ -934,6 +934,26 @@ export function nodeCanBeDecorated(node: TSNode): boolean {
...
+    case SyntaxKind.Identifier:
+      return true;
+    case SyntaxKind.PropertyAccessExpression:
+    case SyntaxKind.ElementAccessExpression:
+      if (node.flags & ts.NodeFlags.OptionalChain) {
+        return false;
+      }
+      return true;
+    case SyntaxKind.ParenthesizedExpression:
+    case SyntaxKind.TypeAssertionExpression:
```
bradzacher, line 948: I might have missed it but I think we're missing a case for this

bradzacher (changes requested):
just one comment otherwise
youhavemystamp

packages/typescript-estree/src/node-utils.ts
Josh-Cena: @bradzacher Wow, thanks for catching that! In doing so I realized `TypeAssertionExpression` is only for `<number>foo` and `foo as number` is a separate syntax kind.

bradzacher (approved):
just need to get CI passing and we can merge this

Josh-Cena:
The test failure is coming from https://github.com/typescript-eslint/typescript-eslint/pull/8052

PR #6457, opened 2023-02-13
feat(eslint-plugin): [class-methods-use-this] add extension rule

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1103
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

This adds a standard AST fork of the base rule that can be configured to ignore one or both of the cases where a method has the `override` keyword, or the parent class implements an interface.

Based on #7260

PR #7260, opened 2023-07-18
feat(eslint-plugin): sync getFunctionHeadLoc implementation with upstream

bradzacher:
Splitting this out of #6457 to better focus the PR.
Syncs the utility from upstream giving us better coverage and alignment with ESLint core.


JoshuaKGoldberg (approved):
South park style politician stamping a paper in their office

PR #7267, opened 2023-07-19
fix(typescript-estree): fix TSNode type error on old ts versions

bradzacher:
## Overview

Found this when upgrading an internal codebase that's on TS4.9
Also realised we forgot to clean this up with the v6 min version bump!


JoshuaKGoldberg (approved):
Aha!

PR #7273, opened 2023-07-19
feat: add types for flat config files

bradzacher:
## Overview

Adds in strict type definitions for the new "flat config" style config.
I took the time to do a little bit of refactoring to attempt to share as many types as possible and better organise things.
In particular I tried to move things a little so everything didn't hang off the `Linter` namespace like before.

I believe I got everything right - it was pieced together from the docs. Sadly there is no json schema like with classic configs and the validation function is quite loose - so some of it is pieced together based on past config structure.
Feel free to eyeball [the docs](https://eslint.org/docs/latest/use/configure/configuration-files-new) and compare them to make sure I got it right.

This change should be completely backwards-compatible - please flag if you think I made any incompatible changes though.

TODO - https://eslint.org/docs/latest/extend/plugins#metadata-in-plugins


packages/utils/src/ts-eslint/Parser.ts
```diff
@@ -0,0 +1,70 @@
...
+         */
+        parse(text: string, options?: ParserOptions): TSESTree.Program;
+      }
+    | {
+        /**
+         * Information about the parser to uniquely identify it when serializing.
+         */
+        meta?: ParserMeta;
+        /**
+         * Parses the given text into an AST with optional
```
JoshuaKGoldberg, line 36: > with optional

?

JoshuaKGoldberg (dismissed):
OK! From what I can tell, this is a very accurate and good representation of how things will be in the new world. But, as you mentioned in the description, this is all conjecture and guessing. I think we'd want some response to ~https://github.com/eslint/eslint/issues/17391~ https://github.com/eslint/eslint/issues/17242 and/or an explicit review from someone on ESLint core. 🤷

JoshuaKGoldberg:
@bradzacher now that https://github.com/eslint/eslint/pull/17640 is merged, are there any changes you want to make here?

karlhorky:
@JoshuaKGoldberg @bradzacher while we wait for the PR merge + release, is it possible to publish a WIP version of this to npm? Or is there another easy way for package authors to help test out these new types in their projects? (before this PR is merged)

karlhorky:
## Workaround

For now, I'm using `@types/eslint`, which has the `Linter.FlatConfig` export:

```js
/** @type {import('eslint').Linter.FlatConfig} */
const config = {
};

export default config;
```

Types aren't as nice as the `@typescript-eslint/util` types, but it works so far for now...

packages/utils/src/ts-eslint/Parser.ts
JoshuaKGoldberg:
```suggestion
         * Parses the given text into an AST
```

JoshuaKGoldberg:
> Or is there another easy way for package authors to help test out these new types in their projects? (before this PR is merged)

https://typescript-eslint.io/contributing/local-development -> https://typescript-eslint.io/contributing/local-development/local-linking would be the right way.

Looks like this has some merge conflicts right now. Were it not for those and builds on `main` being broken (#7899) I'd merge as-is.

bradzacher:
One of the reasons I hadn't pushed forward with this was I was trying to decide if I should build some features for typing the config or not.

There's a whole project around that so might be worth just not doing it for the moment.

karlhorky:
> features for typing the config

> There's a whole project around that

I assume you mean [`eslint-define-config`](https://github.com/eslint-types/eslint-define-config)  by @Shinigami92 ?

Alternative: I have been [using the types directly with JSDoc](https://github.com/upleveled/eslint-config-upleveled/blob/6bea5040ec38d4d393efe86e181699aa4a8dceaa/index.cjs#L441) and the experience hasn't been too bad.

bradzacher:
Yes, like that but built in.
I spent a while musing on this idea but I ultimately got stuck on trying to make it truly safe.
Sadly the way that flat configs are defined are just not compatible with true type safety.

For example this is a valid config

```ts
import foo from 'eslint-plugin-foo';
import bar from 'eslint-plugin-bar';

export default [
  { plugins: {foo} }, 
  {
    files: ["path1"],
    plugins: {bar}, 
  }, 
  {
    // foo plugin rules are available here
    // but not bar plugin rules
  } 
];
```

And you can see how the generics just don't work because config merging. So we'd need to go with a less sound approach (which is fine enough but not ideal). There's also the complicated thing of renaming plugins which would need to be handled (via template literal types probs?) 

The other thing I was looking at was our schema generation tooling. To make type-safe configs work we would want a way to auto-generate types for the plugin based on the json schemas. We now have that tooling but we don't have it designed for public consumption - or even published.

Theres a lot of work to do to define this space as a general solution that any plugin at all can easily leverage and users can "just use".

---

Which is a long and rambly confirmation that this probably just needs to land as-is and the type-safe rule configs can come later.

JoshuaKGoldberg:
cc @bradzacher - GitHub won't let me request your review on your own PR (because nobody has ever swapped author/reviewer role on a PR, ever) but I think I got it merged from `main` all right.

packages/utils/src/ts-eslint/Config.ts
```diff
@@ -0,0 +1,244 @@
...
+    [name: string]: GlobalVariableOption;
+  }
+  export interface EnvironmentConfig {
+    [name: string]: boolean;
+  }
+
+  export type ParserOptions = ParserOptionsTypes.ParserOptions;
+}
+
+export namespace ClassicConfig {
```
bradzacher, line 36: I was being kind of petty when I called this "classic config" - realistically given that v9 will physically kill them, it probably should be called "legacy config".

packages/utils/src/ts-eslint/Linter.ts
```diff
@@ -118,91 +113,23 @@ namespace Linter {
...
+  export type GlobalVariableOptionBase = SharedConfig.GlobalVariableOptionBase;
+  export type ParserOptions = SharedConfig.ParserOptions;
+  export type RuleEntry = SharedConfig.RuleEntry;
+  export type RuleLevel = SharedConfig.RuleLevel;
+  export type RuleLevelAndOptions = SharedConfig.RuleLevelAndOptions;
+  export type RulesRecord = SharedConfig.RulesRecord;
+  export type Severity = SharedConfig.Severity;
+  export type SeverityString = SharedConfig.SeverityString;
+
+  /** @deprecated use ClassicConfig.ConfigType instead */
```
bradzacher, line 128:
```suggestion
  /** @deprecated use Linter.ConfigType instead */
```

packages/utils/src/ts-eslint/Linter.ts
```diff
@@ -118,91 +113,23 @@ namespace Linter {
...
+  export type Config = ClassicConfig.Config;
+  export type ConfigType = ClassicConfig.Config | FlatConfig.ConfigArray;
+  /** @deprecated use ClassicConfig.ConfigOverride instead */
```
bradzacher, line 131:
```suggestion
  /** @deprecated use Linter.ConfigOverride instead */
```

JoshuaKGoldberg:
oop - @bradzacher if you want to send another PR I should be able to review before our auto-release today

(I'm about to step out but will be back soon)

packages/utils/src/ts-eslint/Config.ts
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * The definition of plugin processors.
+     * Can be referenced by string in the config (i.e., `"pluginName/processorName"`).
+     */
+    processors?: Record<string, Processor>;
+    /**
+     * The definition of plugin rules.
+     */
+    rules?: Record<string, RuleCreateFunction | AnyRuleModule>;
+  }
```
bradzacher, line 144: https://eslint.org/docs/latest/extend/plugins#meta-data-in-plugins

```suggestion
  export interface Plugin {
    /**
     * The definition of plugin processors.
     * Can be referenced by string in the config (i.e., `"pluginName/processorName"`).
     */
    processors?: Record<string, Processor>;
    /**
     * The definition of plugin rules.
     */
    rules?: Record<string, RuleCreateFunction | AnyRuleModule>;
    /**
     * Metadata about your plugin for easier debugging and more effective caching of plugins
     */
    meta?: {
      /**
       * The meta.name property should match the npm package name for your plugin.
       */
      name: string,
      /**
       * The meta.version property should match the npm package version for your plugin.
       */
      version: string,
    },
  }
```

bradzacher: There's also https://eslint.org/docs/latest/extend/plugins#configs-in-plugins that wasn't added.
I was waiting on them to update the docs to show what the new configs look like but IDK when they'll do that.........

packages/utils/src/ts-eslint/Config.ts
```diff
@@ -0,0 +1,244 @@
...
+  export interface Plugin {
+    /**
+     * The definition of plugin processors.
+     * Can be referenced by string in the config (i.e., `"pluginName/processorName"`).
+     */
+    processors?: Record<string, Processor>;
+    /**
+     * The definition of plugin rules.
+     */
+    rules?: Record<string, RuleCreateFunction | AnyRuleModule>;
```
bradzacher, line 143: I think that just specifying `RuleCreateFunction` might be dead for flat configs?
I know it's deprecated I just don't know if it's dead dead.

packages/utils/src/ts-eslint/Config.ts
bradzacher: Probably need to check these types against https://eslint.org/docs/latest/extend/plugin-migration-flat-config to double check they are correct.

I originally wrote them blindly slash in conjunction with `@types/eslint` and the eslint source code.

bradzacher:
🤷‍♂️ was just drive by commenting before i went to bed.
i'd forgotten what I'd even put in this pr TBH heh.
anything older than a month is mostly deleted from my memory #toddlerdadlife

JoshuaKGoldberg:
Well, this is released in https://github.com/typescript-eslint/typescript-eslint/releases/tag/v6.11.0 (cc @karlhorky). If folks don't like it we can always fix it up 😄

karlhorky:
Hmm... just trying out `@typescript-eslint/utils@6.11.0` now, and I'm having trouble getting my JSDoc imports working

I first tried this:

```js
/** @type {import('@typescript-eslint/utils').TSESLint.Linter.Config[]} */
const config = [
  { 
    plugins: {
      '@next/next': next,
    },
    ...
  },
  ...
```

But this resulted in the error that the type of `plugins` doesn't match the eslintrc-style config (`string[]`):

```
Type '{ '@next/next': ESLint.Plugin; '@typescript-eslint': ESLint.Plugin; 'jsx-a11y': ESLint.Plugin; 'jsx-expressions': ESLint.Plugin; 'react-hooks': ESLint.Plugin; ... 5 more ...; upleveled: ESLint.Plugin; }' is not assignable to type 'string[]'.
  Object literal may only specify known properties, and ''@next/next'' does not exist in type 'string[]'.
```


Another option I tried was to try to import the `FlatConfig` namespace directly from `dist/ts-eslint/Config.js`, but I think this doesn't work either:

```js
/** @type {import('./node_modules/@typescript-eslint/utils/dist/ts-eslint/Config.js').FlatConfig[]} */
```

This resulted in a different error:

```
Namespace '"/Users/k/p/eslint-config-upleveled/node_modules/@typescript-eslint/utils/dist/ts-eslint/Config"' has no exported member 'FlatConfig'.
```

---

Is there a better way to import this type for usage in my shared config?

bradzacher:
@karlhorky you should be able to use
```ts
import('@typescript-eslint/utils/ts-eslint').Config.FlatConfig.ConfigArray
```

karlhorky:
Thanks! Switched over here:

- https://github.com/upleveled/eslint-config-upleveled/commit/a3830a4ca426a019bee0f060f70f91b1122430cf

Still some manual work needed (see below), but looks like it's working 🙌 

- extra `.d.ts` files including [a custom type for `configs.recommended` on ESLint plugins](https://github.com/upleveled/eslint-config-upleveled/blob/a3830a4ca426a019bee0f060f70f91b1122430cf/types/eslint-plugin-jsx-a11y.d.ts#L5-L7)
- casting (eg. for [conditional modification of the config](https://github.com/upleveled/eslint-config-upleveled/blob/a3830a4ca426a019bee0f060f70f91b1122430cf/index.js#L934-L947)) - `satisfies` would be nicer to use here, but I was unable to get [`@satisfies` in JSDoc](https://github.com/microsoft/TypeScript/pull/51753) working

Issue #3440, opened 2021-05-24
Bug: [prefer-reduce-type-parameter] unfixable reporting


JounQin:
- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.
- [x] I have [read the FAQ](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/FAQ.md) and my problem is not listed.

**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/<rule>": ["prefer-reduce-type-parameter"]
  }
}
```

```TS
// first case
export function normalizeParams<P>(params: P): P {
  return Object.entries(params).reduce((acc, [key, value]) => {
    value = typeof value === 'string' ? value.trim() : value;
    if (value) {
      acc[key as keyof P] = value;
    }
    return acc;
  }, {} as P);
}

// second case
export type Keys<T> = { [Key in keyof T]: Key };

export const toKeys = <T>(source: T): Keys<T> =>
  Object.keys(source).reduce((acc, k) => {
    const key = k as keyof T;
    acc[key] = key;
    return acc;
  }, {} as Keys<T>);
```

**Expected Result**

Ignore such cases

**Actual Result**

Error reporting

**Additional Info**

**Versions**

| package                            | version |
| ---------------------------------- | ------- |
| `@typescript-eslint/eslint-plugin` | `4.24.0` |
| `@typescript-eslint/parser`        | `4.24.0` |
| `TypeScript`                       | `4.2.4` |
| `ESLint`                           | `7.27.9` |
| `node`                             | `12.22.1` |

bradzacher:
Both of your examples are ***completely unsound*** due to the `as` though.
You're *forcing* typescript to treat `P`/`T` as an object, but they are unbounded generics - meaning they can be non-objects.

```ts
const str1 = normalizeParams('foo');
// typeof str1 === 'foo'
// but at runtime
// str1 === {0: 'f', 1: 'o', 2: 'o'}

const num1 = normalizeParams(1);
// typeof num1 === 1
// but at runtime
// num1 === {}



const str2 = normalizeParams('foo');
// typeof str2 === string
// but at runtime
// str2 === {0: "0", 1: "1", 2: "2"}

const num1 = normalizeParams(1);
// typeof num1 === number
// but at runtime
// num1 === {}
```

Doing an `as` assertion really is unsafe here.
[...]

PR #7339, opened 2023-07-27
chore: enabled most of strict-type-checked internally

### LINKED ISSUE #6827: Repo: Enable strict & stylistic configs internally

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6827

## Overview

Ran `yarn lint --fix` and disabled a couple rules with a _lot_ of complaints.


packages/eslint-plugin/src/rules/naming-convention-utils/enums.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/prefer-literal-enum-member */
```
bradzacher, line 1: we should probably be explicit with the enums in this file

packages/eslint-plugin/src/rules/naming-convention-utils/parse-options.ts
```diff
@@ -87,6 +87,7 @@ function parseOptions(context: Context): ParsedOptions {
   const result = util.getEnumNames(Selectors).reduce((acc, k) => {
     acc[k] = createValidator(k, context, normalizedOptions);
     return acc;
+    // eslint-disable-next-line @typescript-eslint/prefer-reduce-type-parameter
```
bradzacher, line 90: i wonder if this should be an exception in the rule?
I don't think `{}` can ever be typed with the parameter, can it?

packages/eslint-plugin/src/rules/prefer-regexp-exec.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/prefer-literal-enum-member */
```
bradzacher, line 1:
```ts
enum ArgumentType {
  Other = 0,
  String = 1 << 0,
  RegExp = 1 << 1,
  Both = String | RegExp,
}
```

🤮

packages/eslint-plugin/src/util/getOperatorPrecedence.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/prefer-literal-enum-member */
```
bradzacher, line 1:
```suggestion
/* eslint-disable @typescript-eslint/prefer-literal-enum-member -- the enums come from TS so to make merging upstream easier we purposely avoid adding literal values. */
```

packages/eslint-plugin/tools/generate-configs.ts
```diff
@@ -76,10 +76,7 @@ async function main(): Promise<void> {
   );
   const EXTENDS = ['./configs/base', './configs/eslint-recommended'];
 
-  type RuleEntry = [
-    string,
-    TSESLint.RuleModule<string, readonly unknown[], TSESLint.RuleListener>,
-  ];
+  type RuleEntry = [string, TSESLint.RuleModule<string, readonly unknown[]>];
```
bradzacher, line 79: TODO - should really migrate to `import { RuleModule } from '@typescript-eslint/utils/ts-eslint'`
That was the entire reason for adding the export maps

packages/scope-manager/src/definition/CatchClauseDefinition.ts
```diff
@@ -6,8 +6,7 @@ import { DefinitionType } from './DefinitionType';
 class CatchClauseDefinition extends DefinitionBase<
   DefinitionType.CatchClause,
   TSESTree.CatchClause,
-  null,
-  TSESTree.BindingName
```
bradzacher, line 10: I'll be honest when I say that keeping these for documentation is nice.
Can we invert this problem and remove the default from `DefinitionBase`?

packages/utils/src/ts-eslint/SourceCode.ts
```diff
@@ -208,19 +208,7 @@ declare class TokenStore {
   getTokensBetween<T extends SourceCode.CursorWithCountOptions>(
```
bradzacher, line 208:
```diff
-* @param options The option object. If this is a function then it's `options.filter`.
+* @param options The option object. If this is a number then it's `options.count`. If this is a function then it's `options.filter`.
```

so that it matches the other signatures

bradzacher (dismissed):

packages/eslint-plugin/src/rules/naming-convention-utils/enums.ts
JoshuaKGoldberg: Just confirming - do you mean you're not in favor of the bitwise `1 << 0`s, etc.?

packages/eslint-plugin/src/rules/naming-convention-utils/parse-options.ts
JoshuaKGoldberg: https://github.com/typescript-eslint/typescript-eslint/issues/3440#issuecomment-1356641311 

... https://github.com/microsoft/TypeScript/issues/9879 🙃

bradzacher (dismissed):

packages/eslint-plugin/src/rules/prefer-regexp-exec.ts
Josh-Cena: Doesn't it make you feel like a pro? 😎

packages/eslint-plugin/src/rules/naming-convention-utils/enums.ts
JoshuaKGoldberg: @bradzacher was there any action you wanted me to take here?

PR #7355, opened 2023-07-28
Update no-explicit-any.md

### LINKED ISSUE #7354: Docs: add allowance in no-explicit-any docs that any is sometimes a necessary last resort

### Suggested Changes

The current documentation reads as if explicit any is almost always an error that needs fixing, but due to a lack of TypeScript features (e.g. higher-kinded types or Haskell-style custom infix operators) explicit any is often the only correct fix for typing higher-order functions.

e.g.:

```TypeScript
// eslint-disable-next-line no-explicit-any
type a2a = (x: any) => any;
type compose = (...fns: a2a[]) => a2a;
const pipe: compose = (...fns) => x => fns.reduce((y, f) => f(y), x);
const compose: compose = (...fns) => x => fns.reduceRight((y, f) => f(y), x);

type n2n = (n: number) => number;
const g: n2n = n => n + 1;
const f: n2n = n => n * 2;

const h: n2n = pipe(g, f);
const j: n2n = compose(f, g);
```

### Affected URL(s)

https://typescript-eslint.io/rules/no-explicit-any/

ericelliott:
Fix potentially misleading statement about explicit `any` - which is sometimes required for functional code.

## PR Checklist

**Notes:**

This checklist discourages people from contributing PRs by trying to force us to wait around for "accepting prs". If you don't want to accept a PR, just close it. Bonus: It gives you one spot to decide whether to accept a PR, instead of 2.

- [x] Addresses an existing open issue: fixes #7354

## Overview

Provide an example of when explicit `any` may be needed, and how to disable the rule on an as-needed basis.

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-explicit-any.md
+++ b/packages/eslint-plugin/docs/rules/no-explicit-any.md
@@ -161,7 +161,7 @@ interface Garply {
 **Difficult-to-represent TypeScript types**.
 Some code patterns can be difficult to represent exactly in the TypeScript type system. For example, functional programming concepts such as composing and piping sometimes necessitate using `any`.
 
-```TypeScript
+```ts
 // eslint-disable no-explicit-any
 type a2a = (x: any) => any;
 const pipe = (...fns: a2a[]) => (x: any) => fns.reduce((y, f) => f(y), x);
--- a/packages/eslint-plugin/docs/rules/no-explicit-any.md
+++ b/packages/eslint-plugin/docs/rules/no-explicit-any.md
@@ -158,7 +158,8 @@ interface Garply {
 
 ## When Not To Use It
 
-**Higher Order Functions**. If you're typing a higher order function such as `compose`, `pipe`, etc, you may need `any` because TypeScript types can't always fully represent functional JavaScript patterns. For example, functions like `compose` and `pipe` compose many different functions with many different types which TypeScript can't accurately represent. Instead, you can assign explicit types later to the returned functions, e.g.:
+**Difficult-to-represent TypeScript types**.
+Some code patterns can be difficult to represent exactly in the TypeScript type system, and/or expose legitimate edge cases for using `any`. For example, functional programming concepts such as composing and piping sometimes necessitate using `any`.
 
 ```TypeScript
 // eslint-disable no-explicit-any
--- a/packages/eslint-plugin/docs/rules/no-explicit-any.md
+++ b/packages/eslint-plugin/docs/rules/no-explicit-any.md
@@ -158,21 +158,23 @@ interface Garply {
 
 ## When Not To Use It
 
-**Higher Order Functions**. If you're typing a higher order function such as `compose`, `pipe`, etc, you may need `any` because TypeScript types are not fully co-expressive with idiomatic functional JavaScript. Such functions are intentionally generic, and you can assign explicit types later to the returned functions, e.g.:
+**Higher Order Functions**. If you're typing a higher order function such as `compose`, `pipe`, etc, you may need `any` because TypeScript types can't always fully represent functional JavaScript patterns. For example, functions like `compose` and `pipe` compose many different functions with many different types which TypeScript can't accurately represent. Instead, you can assign explicit types later to the returned functions, e.g.:
 
 ```TypeScript
-// eslint-disable-next-line no-explicit-any
+// eslint-disable no-explicit-any
 type a2a = (x: any) => any;
-type compose = (...fns: a2a[]) => a2a;
-const pipe: compose = (...fns) => x => fns.reduce((y, f) => f(y), x);
-const compose: compose = (...fns) => x => fns.reduceRight((y, f) => f(y), x);
+const pipe = (...fns: a2a[]) => (x: any) => fns.reduce((y, f) => f(y), x);
+const compose = (...fns: a2a[]) => (x: any) => fns.reduceRight((y, f) => f(y), x);
+// eslint-enable no-explicit-any
 
 type n2n = (n: number) => number;
 const g: n2n = n => n + 1;
 const f: n2n = n => n * 2;
+type n2s = (n: number) => string;
+const exclaim:n2s = (n) => `${n}!`;
 
-const h: n2n = pipe(g, f);
-const j: n2n = compose(f, g);
+const h: n2s = pipe(g, f, exclaim);
+const j: n2s = compose(exclaim, f, g);
 ```
 
 **Unknown Types**. If an unknown type or a library without typings is used
--- a/packages/eslint-plugin/docs/rules/no-explicit-any.md
+++ b/packages/eslint-plugin/docs/rules/no-explicit-any.md
@@ -6,8 +6,8 @@ description: 'Disallow the `any` type.'
 >
 > See **https://typescript-eslint.io/rules/no-explicit-any** for documentation.
 
-The `any` type in TypeScript is a dangerous "escape hatch" from the type system.
-Using `any` disables many type checking rules and is generally best used only as a last resort or when prototyping code.
+The `any` type in TypeScript is a potentially dangerous "escape hatch" from the type system.
+Using `any` disables many type checking rules and is generally best used only when it's required or when prototyping code.
 This rule reports on explicit uses of the `any` keyword as a type annotation.
 
 > TypeScript's `--noImplicitAny` compiler option prevents an implied `any`, but doesn't prevent `any` from being explicitly used the way this rule does.
@@ -158,7 +158,24 @@ interface Garply {
 
 ## When Not To Use It
 
-If an unknown type or a library without typings is used
+**Higher Order Functions**. If you're typing a higher order function such as `compose`, `pipe`, etc, you may need `any` because TypeScript types are not fully co-expressive with idiomatic functional JavaScript. Such functions are intentionally generic, and you can assign explicit types later to the returned functions, e.g.:
+
+```TypeScript
+// eslint-disable-next-line no-explicit-any
+type a2a = (x: any) => any;
+type compose = (...fns: a2a[]) => a2a;
+const pipe: compose = (...fns) => x => fns.reduce((y, f) => f(y), x);
+const compose: compose = (...fns) => x => fns.reduceRight((y, f) => f(y), x);
+
+type n2n = (n: number) => number;
+const g: n2n = n => n + 1;
+const f: n2n = n => n * 2;
+
+const h: n2n = pipe(g, f);
+const j: n2n = compose(f, g);
+```
+
+**Unknown Types**. If an unknown type or a library without typings is used
 and you want to be able to specify `any`.
 
 ## Related To
--- a/packages/eslint-plugin/docs/rules/no-explicit-any.md
+++ b/packages/eslint-plugin/docs/rules/no-explicit-any.md
@@ -7,7 +7,7 @@ description: 'Disallow the `any` type.'
 > See **https://typescript-eslint.io/rules/no-explicit-any** for documentation.
 
 The `any` type in TypeScript is a potentially dangerous "escape hatch" from the type system.
-Using `any` disables many type checking rules and is generally best used only when it's required or when prototyping code.
+Using `any` disables many type checking rules and is generally best used only as a last resort or when prototyping code.
 This rule reports on explicit uses of the `any` keyword as a type annotation.
 
 > TypeScript's `--noImplicitAny` compiler option prevents an implied `any`, but doesn't prevent `any` from being explicitly used the way this rule does.
--- a/packages/eslint-plugin/docs/rules/no-explicit-any.md
+++ b/packages/eslint-plugin/docs/rules/no-explicit-any.md
@@ -159,7 +159,7 @@ interface Garply {
 ## When Not To Use It
 
 **Difficult-to-represent TypeScript types**.
-Some code patterns can be difficult to represent exactly in the TypeScript type system, and/or expose legitimate edge cases for using `any`. For example, functional programming concepts such as composing and piping sometimes necessitate using `any`.
+Some code patterns can be difficult to represent exactly in the TypeScript type system. For example, functional programming concepts such as composing and piping sometimes necessitate using `any`.
 
 ```TypeScript
 // eslint-disable no-explicit-any


packages/eslint-plugin/docs/rules/no-explicit-any.md
```diff
@@ -158,7 +158,24 @@ interface Garply {
 
 ## When Not To Use It
 
-If an unknown type or a library without typings is used
+**Higher Order Functions**. If you're typing a higher order function such as `compose`, `pipe`, etc, you may need `any` because TypeScript types are not fully co-expressive with idiomatic functional JavaScript. Such functions are intentionally generic, and you can assign explicit types later to the returned functions, e.g.:
```
JoshuaKGoldberg, line 161: > fully co-expressive with idiomatic functional JavaScript

The vast majority of our readers won't know what this means 🙂. Let's stick with phrases that keep docs approachable for newcomers.

IMO docs generally don't need to be exactly precise for topics that don't need it. I think it'd be ok here to lose a little precision and go with something like:

```diff
- TypeScript types are not fully co-expressive with idiomatic functional JavaScript.
+ TypeScript types can't always fully represent functional JavaScript patterns.
```

Thoughts?

(I'll separately comment that I think we can make this more generalized - this is more a discussion for the fun of it than anything else)

packages/eslint-plugin/docs/rules/no-explicit-any.md
```diff
@@ -6,8 +6,8 @@ description: 'Disallow the `any` type.'
 >
 > See **https://typescript-eslint.io/rules/no-explicit-any** for documentation.
 
-The `any` type in TypeScript is a dangerous "escape hatch" from the type system.
-Using `any` disables many type checking rules and is generally best used only as a last resort or when prototyping code.
+The `any` type in TypeScript is a potentially dangerous "escape hatch" from the type system.
+Using `any` disables many type checking rules and is generally best used only when it's required or when prototyping code.
```
JoshuaKGoldberg, line 10: > when it's required

-1 from me on this change: "required" is, at best, just as vague as "last resort". While "last resort" both conveys our disdain for `any` (good) and indicates you should have tried many things first (also good).

packages/eslint-plugin/docs/rules/no-explicit-any.md
JoshuaKGoldberg, line 161: > **Higher Order Functions**

Functional shenanigans are a common example of type system edge cases, yes, but they're not the only one. And -similar to my other comment- functional terminology can be off-putting when phrased as a top-level reason rather than an example of a more approachable term. How about we make this more general?

```diff
- **Higher Order Functions**. etc. etc.
+ **Difficult-to-represent TypeScript types**.
+ Some code patterns can be difficult to represent exactly in the TypeScript type system, and/or expose legitimate edge cases for using `any`.
+ For example, functional programming concepts such as composing and piping sometimes necessitate using `any`.
```

Thoughts?

JoshuaKGoldberg (changes requested):
You know it's going to be a fun time when someone files an issue mentioning Haskell and the PR uses a phrase like _"fully co-expressive"_. 😄 

A good start, but let's workshop the direction & phrasing. Thanks!

packages/eslint-plugin/docs/rules/no-explicit-any.md
```diff
@@ -6,8 +6,8 @@ description: 'Disallow the `any` type.'
 >
 > See **https://typescript-eslint.io/rules/no-explicit-any** for documentation.
 
-The `any` type in TypeScript is a dangerous "escape hatch" from the type system.
-Using `any` disables many type checking rules and is generally best used only as a last resort or when prototyping code.
+The `any` type in TypeScript is a potentially dangerous "escape hatch" from the type system.
```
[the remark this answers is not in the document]
bradzacher: What isn't dangerous about `any`?
It literally opts you out of every single check that TS might provide - there's nothing that isn't dangerous about that.

packages/eslint-plugin/docs/rules/no-explicit-any.md
```diff
@@ -158,7 +158,24 @@ interface Garply {
...
+type compose = (...fns: a2a[]) => a2a;
+const pipe: compose = (...fns) => x => fns.reduce((y, f) => f(y), x);
+const compose: compose = (...fns) => x => fns.reduceRight((y, f) => f(y), x);
+
+type n2n = (n: number) => number;
+const g: n2n = n => n + 1;
+const f: n2n = n => n * 2;
+
+const h: n2n = pipe(g, f);
+const j: n2n = compose(f, g);
```
Josh-Cena, line 175: It's not hard to make this type-safe:

```ts
type a2a<T> = (x: T) => T;
const pipe = <T,>(...fns: a2a<T>[]) => (x: T) => fns.reduce((y, f) => f(y), x);
const compose = <T,>(...fns: a2a<T>[]) => (x: T) => fns.reduceRight((y, f) => f(y), x);

type n2n = (n: number) => number;
const g: n2n = n => n + 1;
const f: n2n = n => n * 2;

const h: n2n = pipe(g, f);
const j: n2n = compose(f, g);
```

So I don't think this should be shown. To date, I haven't seen a single case where you cannot express patterns fully type-safely (or, if so, the pattern is probably awkward in the first place anyway), if you think hard enough about it.

ericelliott: Those are not correct types for compose and pipe, which can take any number of functions of different types:

```
type a2a<T> = (x: T) => T;
const pipe = <T,>(...fns: a2a<T>[]) => (x: T) => fns.reduce((y, f) => f(y), x);
const compose = <T,>(...fns: a2a<T>[]) => (x: T) => fns.reduceRight((y, f) => f(y), x);

type n2n = (n: number) => number;
const g: n2n = n => n + 1;
const f: n2n = n => n * 2;
type n2s = (n: number) => string;
const exclaim:n2s = (n) => `${n}!`;

const h: n2s = pipe(g, f, exclaim);
const j: n2s = compose(exclaim, f, g);
```

Errors:

```
Type '(x: number) => number' is not assignable to type 'n2s'.
  Type 'number' is not assignable to type 'string'.
Argument of type 'n2s' is not assignable to parameter of type 'a2a<number>'.
  Type 'string' is not assignable to type 'number'.
Type '(x: number) => number' is not assignable to type 'n2s'.
  Type 'number' is not assignable to type 'string'.
Argument of type 'n2s' is not assignable to parameter of type 'a2a<number>'.
```

vs

```
type a2a = (x: any) => any;
const pipe = (...fns: a2a[]) => (x: any) => fns.reduce((y, f) => f(y), x);
const compose = (...fns: a2a[]) => (x: any) => fns.reduceRight((y, f) => f(y), x);

type n2n = (n: number) => number;
const g: n2n = n => n + 1;
const f: n2n = n => n * 2;
type n2s = (n: number) => string;
const exclaim:n2s = (n) => `${n}!`;

const h: n2s = pipe(g, f, exclaim);
const j: n2s = compose(exclaim, f, g);
```

No errors.

packages/eslint-plugin/docs/rules/no-explicit-any.md
ericelliott: Huge improvement!

packages/eslint-plugin/docs/rules/no-explicit-any.md
ericelliott: I can get on-board with that as long as it's clear that `any` is not always an error that needs fixing.

packages/eslint-plugin/docs/rules/no-explicit-any.md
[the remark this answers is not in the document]
JoshuaKGoldberg: Re-thinking after reading your https://github.com/typescript-eslint/typescript-eslint/issues/7354#issuecomment-1654905592: yeah you're right. I was coming from the perspective that it can be necessary and, when handled carefully, relatively safe. But as you said:

> Your example is only safe purely because you've explicitly annotated the variables to "hide" the `any`. Essentially you've allowed the unsafety and covered it up quickly.

(and the rest of that paragraph)

Marking as resolved, rescinding my +1 😄

packages/eslint-plugin/docs/rules/no-explicit-any.md
ericelliott: Love it.

packages/eslint-plugin/docs/rules/no-explicit-any.md
Josh-Cena: FWIW, most FP libraries define overloads for 1–10 piped functions and _then_ as a last resort have the `(...fns: (a: any) => any) => any` signature. This does seem awkward but you don't take type safety away from your consumers.

packages/eslint-plugin/docs/rules/no-explicit-any.md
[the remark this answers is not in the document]
ericelliott: @bradzacher I could go deep into the empirical data behind [type safety](https://medium.com/javascript-scene/the-typescript-tax-132ff4cb175b) but this isn't the right forum for a debate. I think it's enough to say that there are some useful things that you just can't do in TypeScript without `any` - foundational things like proper function composition (see the code example) which can [eliminate a huge amount of code in your application](https://medium.com/javascript-scene/why-every-react-developer-should-learn-function-composition-23f41d4db3b1) and with it, eliminate all the bugs that may have cropped up in the code that got eliminated. In other words, there are cases where using `any` is a net win for safety, by a huge margin.

packages/eslint-plugin/docs/rules/no-explicit-any.md
ericelliott: 1. It is *extremely common* to declare compose and pipe inline because they are convenient one-liners that many functional programmers can crank out faster than they can import a library. Inline definitions also reduce dependencies and are usually far simpler than the library alternatives (including typings).

Even if that could work, even doing it for all the common basic combos like `string => Array<Number>` quickly leads to an untenable combinatorial explosion. In other words, the "additional type safety" provided by overloading is a fragile illusion:

* string
* number
* bigint
* boolean
* symbol
* object
* Array<T>
* Promise<T>
* Map<K, V>
* WeakMap<K extends object, V>
* Set<T>
* WeakSet<T>
* Date
* RegExp
* Error
* Function (oh no! functions can take and return functions! Inception time...)
... -- Remember, we have to raise this whole list to an exponent, and getting into all the combinations of those <T> is just not a workable solution.

Josh-Cena: I think you misunderstood about overloads. I'm talking about these:

```ts
function pipe<T1, R>(f1: (a: T1) => R): R;
function pipe<T1, T2, R>(f1: (a: T1) => T2, f2: (a: T2) => R): R;
function pipe<T1, T2, T3, R>(f1: (a: T1) => T2, f2: (a: T2) => T3, f3: (a: T3) => R): R;
```

ericelliott:
Thanks for the quick attention on this @JoshuaKGoldberg @Josh-Cena @bradzacher . 🎉 

I think I have addressed all of the concerns.

packages/eslint-plugin/docs/rules/no-explicit-any.md
ericelliott: Ah, but those don't work because the return value of pipe is a function of a type from `A => C for A => B |> B => C`:

```
// eslint-disable no-explicit-any
// type a2a = (x: any) => any;
// const pipe = (...fns: a2a[]) => (x: any) => fns.reduce((y, f) => f(y), x);
function pipe<T1, R>(f1: (a: T1) => R): R;
function pipe<T1, T2, R>(f1: (a: T1) => T2, f2: (a: T2) => R): R;
function pipe<T1, T2, T3, R>(f1: (a: T1) => T2, f2: (a: T2) => T3, f3: (a: T3) => R): R;
function pipe(...fns: Array<(...args: any[]) => any>): (...args: any[]) => any {
  return (input: any) => fns.reduce((prev, fn) => fn(prev), input);
}
// eslint-enable no-explicit-any

type n2n = (n: number) => number;
const g: n2n = n => n + 1;
const f: n2n = n => n * 2;
type n2s = (n: number) => string;
const exclaim:n2s = (n) => `${n}!`;
```

Error:
```
Type 'string' is not assignable to type 'n2s'.
```

We can try to fix it by expecting the return value to be a function, but TypeScript can't automatically infer the correct function type to expect:

```
function pipe<T1, R>(f1: (a: T1) => R): (a: T1) => R;
function pipe<T1, T2, R>(f1: (a: T1) => T2, f2: (a: T2) => R): (a: T1) => R;
function pipe<T1, T2, T3, R>(f1: (a: T1) => T2, f2: (a: T2) => T3, f3: (a: T3) => R): (a: T1) => R;
function pipe(...fns: Array<(...args: any[]) => any>): (...args: any[]) => any {
  return (input: any) => fns.reduce((prev, fn) => fn(prev), input);
}
// eslint-enable no-explicit-any

type n2n = (n: number) => number;
const g: n2n = n => n + 1;
const f: n2n = n => n * 2;
type n2s = (n: number) => string;
const exclaim:n2s = (n) => `${n}!`;
type s2b = (s: string) => boolean;
const isHuge:s2b = (s) => s.length > 10;

const h: n2s = pipe(g, exclaim, isHuge);
// const j: n2s = compose(exclaim, f, g);

h(20);
```

Error

```
Type '(a: number) => boolean' is not assignable to type 'n2s'.
  Type 'boolean' is not assignable to type 'string'.
```

It seems TypeScript is confused about what types to expect - because we don't know what types to expect until we pass all our functions to the pipe, and it may be the case that none of the function types we pass end up being the correct type for the whole composition.

ericelliott: Oops, no, this doesn't work, either. The type expectations are still too rigid. It fails even with the fallback:

```TypeScript
// eslint-disable no-explicit-any
// type a2a = (x: any) => any;
// const pipe = (...fns: a2a[]) => (x: any) => fns.reduce((y, f) => f(y), x);
type a2a = (x: any) => any;

function pipe<T1, R>(f1: (a: T1) => R): (a: T1) => R;
function pipe<T1, T2, R>(f1: (a: T1) => T2, f2: (a: T2) => R): (a: T1) => R;
function pipe<T1, T2, T3, R>(f1: (a: T1) => T2, f2: (a: T2) => T3, f3: (a: T3) => R): (a: T1) => R;
function pipe(...fns: a2a[]): a2a {
  return (input: any) => fns.reduce((prev, fn) => fn(prev), input);
}
// eslint-enable no-explicit-any

type n2n = (n: number) => number;
const g: n2n = n => n + 1;
const f: n2n = n => n * 2;
type n2s = (n: number) => string;
const exclaim:n2s = (n) => `${n}!`;
type s2b = (s: string) => boolean;
const isHuge:s2b = (s) => s.length > 10;
```

Error:
```
Errors in code
Argument of type 'n2s' is not assignable to parameter of type '(a: number) => number'.
  Type 'string' is not assignable to type 'number'.
```

ericelliott: Note: [Ramda's type is 54 LOC of overloads](https://github.com/ramda/types/blob/develop/types/pipe.d.ts) and still breaks with this example. These problems have [plagued](https://github.com/typed-typings/npm-ramda/issues/29) all of the functional programming libraries for years when people tried to apply TypeScript to them.

packages/eslint-plugin/docs/rules/no-explicit-any.md
```diff
@@ -158,7 +158,27 @@ interface Garply {
 
 ## When Not To Use It
 
-If an unknown type or a library without typings is used
+**Difficult-to-represent TypeScript types**.
+Some code patterns can be difficult to represent exactly in the TypeScript type system, and/or expose legitimate edge cases for using `any`. For example, functional programming concepts such as composing and piping sometimes necessitate using `any`.
+
+```TypeScript
```
Josh-Cena, line 164:
```suggestion
```ts
```

packages/eslint-plugin/docs/rules/no-explicit-any.md
```diff
@@ -158,7 +158,27 @@ interface Garply {
 
 ## When Not To Use It
 
-If an unknown type or a library without typings is used
+**Difficult-to-represent TypeScript types**.
+Some code patterns can be difficult to represent exactly in the TypeScript type system, and/or expose legitimate edge cases for using `any`. For example, functional programming concepts such as composing and piping sometimes necessitate using `any`.
```
Josh-Cena, line 162:
```suggestion
Some code patterns can be difficult to represent exactly in the TypeScript type system. For example, functional programming concepts such as composing and piping sometimes necessitate using `any`.
```

Josh-Cena (approved):
I'm okay with the current wording :)

bradzacher:
> This checklist discourages people from contributing PRs by trying to force us to wait around for "accepting prs". If you don't want to accept a PR, just close it.

This is on purpose - we don't want people to spend hours or days doing work that we won't accept into the project - it wastes their time and deters them from making future contributions.

We've had a number of people in the past raise PRs for new options or docs changes that we as maintainers have disagreed with in the context of the project. It feels sucky for us to have to close a PR that someone was excited about and spent time on, and it feels even worse for the author as well.

Additionally the default search on the issues tab only includes issues. So if you just raise a PR and we discuss it within the PR then people won't easily be able to discover the discussion. This increases the chance of duplicates and the like.
This also has the benefit of focusing the discussion in the PR - the issue is for discussing the idea and whether or not it makes sense, then the PR is for reviewing the change itself.

---

TL;DR by asking for an issue first it means
1) the discussion is easily discoverable in future,
2) the threads can be more focused, and
3) the contributor is less likely to "waste" time on work that's in the wrong direction.

packages/eslint-plugin/docs/rules/no-explicit-any.md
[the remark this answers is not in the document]
bradzacher: I disagree that `any` is safe anywhere.
I get that there are cases that you can't type with TS and are stuck using an `any` with - but that doesn't make it safe to use.

As mentioned in the issue - the only reason your example code is "safe" is because you've hidden the any with an explicit annotation. Had you forgotten the explicit annotation then your code would be in an unsafe state. This approach is simply not scalable though - in small teams it can be sufficient to find and catch in code reviews and thus "make things safe" - but at scale you can't rely on such tribal knowledge and the pattern quickly devolves into an unsafe mess.

Using `any` ***is*** unsafe. Not potentially - it just is.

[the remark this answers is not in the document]
ericelliott: That dogmatic *opinion* is **not supported by evidence.** If you have evidence to support your statement, please supply it. Otherwise, I refer you to the provably true statement: "[function composition] can [eliminate a huge amount of code in your application](https://medium.com/javascript-scene/why-every-react-developer-should-learn-function-composition-23f41d4db3b1) and with it, eliminate all the bugs that may have cropped up in the code that got eliminated." In other words, whether or not static types actually have any causal relationship with bug density (a statement which lacks empirical support), we can say with **objective certainty**, in the specific case of function composition, `any` can be empirically shown to reduce the likelihood of bugs by **reducing the surface area of code** - which, regardless of bug density, will reduce the total amount of code that bugs could appear in, and hence, reduce the total number of bugs over-all.

Empirical Evidence:
* [This article](https://medium.com/javascript-scene/why-every-react-developer-should-learn-function-composition-23f41d4db3b1) explains how function composition reduces the total amount of code needed, with specific examples of both React Higher Order Component composition and server side request handling middleware.
* There is no known causal empirical relationship between language choice (including type safety) and defect rates http://janvitek.org/pubs/toplas19.pdf

[This paper specifically looking at TypeScript vs JavaScript](https://www.researchgate.net/publication/359389871_To_Type_or_Not_to_Type_A_Systematic_Comparison_of_the_Software_Quality_of_JavaScript_and_TypeScript_Applications_on_GitHub) found:

* TypeScript apps were actually more bug prone with longer bug resolution times compared to JavaScript. This goes against the expectation that static typing improves bug rates.
* TS projects needed on average **more than an additional day to fix bugs** (31.86 vs. 33.04 days) (no causal relationship proven either way).
* Within TypeScript projects, reduced usage of the "any" type (which ignores type safety) correlated with better code quality and faster bug resolution times. But it **did not correlate with fewer bugs**. This supports using any as a last resort, but does not support the statement that `any` is "unsafe".

The paper concludes that while TypeScript has benefits like code quality, it **does not directly reduce bugs** versus JavaScript in this sample. **The perceived advantage of type safety for bug prevention was not observed.**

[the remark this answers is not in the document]
ericelliott: If you're thinking of pointing to the widely cited ["To Type or Not to Type"](https://earlbarr.com/publications/typestudy.pdf) whose conclusions were specifically rejected by the last article I mentioned above, you should know the original "To Type or Not to Type" article is the primary source I used to show a very small relationship between static types and bug density in [The TypeScript Tax](https://medium.com/javascript-scene/the-typescript-tax-132ff4cb175b) because the study showed that  "specification errors caused about 78% of the publicly classified bugs studied on GitHub". This implies that type safety cannot be the primary safety net that you use to prevent bug defects, and after applying other safety measures (design review, tdd, code review) the number of remaining bugs is very small, so even factoring for a further 20% bug reduction from TypeScript (which the paper did not achieve, even knowing the root cause of the bugs studied and specifically trying to solve for them using TypeScript typings), the difference in total bugs found is quite small, relatively speaking:

typescript-bug-reduction-after-quality-measures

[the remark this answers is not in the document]
ericelliott: One last point: The back-and-forth about attempts to improve type safety using features like generics to try to make composition type safe reveals several examples of the attempt to eliminate `any` in the type definition actually introduced errors, breaking the type checking for `compose` and `pipe` by making the types more restrictive than they should have been, or leading to incorrect typings for the functions.

This supports my assertion that sometimes, just using `any` is *safer* than trying to remove it from your code.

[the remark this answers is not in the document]
JoshuaKGoldberg: I think we're talking past each other. The following points are ones we seem to agree on:

* Type correctness does not guarantee program correctness _(intentionally quoting your article)_: agreed ✅ 
* Functional programming paradigms improve program quality: agreed ✅
* `any` is sometimes necessary in TypeScript to fully represent type signatures not representable in its type system: agreed ✅ 
* `any` should not be used when a more appropriate type representation is available: agreed ✅ with the caveat that we might not agree on _what_ is more appropriate _when_
* Many cases of `any` _outside of limited necessary uses_ reduce TypeScript's usefulness: agreed ✅ because by definition, features such as go-to-definition and property name checking are turned off on `any`s

The following two opinions are considered canon in this repository. It would be a waste of time to debate them here, as they're generally-accepted TypeScript best practices and discussing them would quickly devolve into huge discussions that likely won't change anybody's mind:

* **`any` is dangerous**: even if it's necessary, it's very easy to get wrong
    * Something can be dangerous and necessary. Using `as` type assertions is sometimes necessary too, but they're also -to a lesser extent- dangerous
* **Type correctness improves program quality**
    * Caveat: it can't be the primary safety net (of course!)
    * You might only agree with us that it improves developer capabilities, but I don't think it's necessary for our discussion to debate whether it reduces bugs


The main contentious point seems to be the intent and wording around whether to use `any` as a backup / last resort. Yes, there are cases where using `any` is better than an incorrect type description. If (or, I hope, _when_) we do take in a PR mentioning that, it'll probably end up being 2-3 sentences at most. We don't need to pull in all of programming and paradigm type theory to discuss those 2-3 sentences.

I'll post a root comment on this PR.

JoshuaKGoldberg:
@ericelliott in this PR you have written us over 500 words -most of which are in the latest few posts in https://github.com/typescript-eslint/typescript-eslint/pull/7355#discussion_r1278599904- and linked articles & papers containing well over 10,000 more explaining your points. I truly appreciate and respect your enthusiasm and willingness to espouse best practices and work with us so we understand where you're coming from. But:

* Per our contributing guidelines and @bradzacher's comment just above this one (https://github.com/typescript-eslint/typescript-eslint/pull/7355#issuecomment-1656940978), our PRs are not a good place for this kind of deep discussion.
* Per https://github.com/typescript-eslint/typescript-eslint/pull/7355#discussion_r1278605493, the conversation is devolving into a much grander scope than it needs to be for this issue.
* Although it's lovely that you have the time to get into deep debates/discussions on this on the internet, neither our maintainer time nor the time of most of our prospective contributors is not well spent getting dragged into them. Having to read through hundreds of comment words + skim -let alone deeply read- thousands of words of _academic papers_ to participate in a repository's discussion is a significant barrier to entry.

I'm going to close this PR. If you want to discuss _**specifically the phrasing around how to say that `any` is a sometimes-necessary, dangerous escape hatch**_ then great!, let's do it in #7354.

If you want to bring up functional programming theory, theory of type safety, or any other grandiose topic, I would encourage you to do that separately from here.

ericelliott:
@JoshuaKGoldberg  I believe your statement "any is a sometimes-necessary, dangerous escape hatch" resolves the discussion beautifully, and I'm willing to amend the PR to say that. My concern was that people see an absolute, unqualified statement like "any is dangerous" and feel like they need to shoot themselves in the foot to eliminate `any`.

I agree [this is not the right forum for the deep debate](https://github.com/typescript-eslint/typescript-eslint/pull/7355#discussion_r1277901606), and it was never my intention to debate anybody.

That said, the suggested changes (with your excellent addition of "any is a sometimes-necessary, dangerous escape hatch") seem like a great fix for the issue.

The PR is now closed, I can't  conveniently make the change. I don't feel like my contributions are welcome or appreciated at this stage. I'll leave you to it.

PR #7390, opened 2023-07-31
feat: add `no-unsafe-unary-minus` rule

### LINKED ISSUE #7076: Rule proposal: no-implicit-coercion using a negative/minus (`-`) symbol

### Description

ESLint's [no-implicit-coercion](https://eslint.org/docs/latest/rules/no-implicit-coercion) rule can't check for casting a string to a number by using `-` because it would disallow negating a regular number. By using type information, we can provide additional coercion checks.

### Fail Cases

```typescript
const foo = '32px'
var n = -foo;
```

### Pass Cases

```typescript
const bar = 32
var n = -bar;
var n = -Number(foo);
var n = -parseFloat(foo);
var n = -parseInt(foo, 10);
```

### Additional Info

_No response_

samestep:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7076

## Overview

This PR pulls the implementation of the [npm package `eslint-plugin-unary-minus`](https://www.npmjs.com/package/eslint-plugin-unary-minus) into this repo.

Note that this implementation uses the `TypeChecker` methods `getUnionType` and `isTypeAssignableTo` which are not public in the TypeScript API; see microsoft/TypeScript#9879.


packages/eslint-plugin/src/rules/no-unsafe-unary-minus.ts
```diff
@@ -0,0 +1,55 @@
+import type * as ts from 'typescript';
+
+import * as util from '../util';
+
+interface TypeChecker extends ts.TypeChecker {
+  // https://github.com/microsoft/TypeScript/issues/9879
+  isTypeAssignableTo(source: ts.Type, target: ts.Type): boolean;
+  getUnionType(types: ts.Type[]): ts.Type;
```
JoshuaKGoldberg, line 8: Oh 😞 this might be a problem for us. We try very hard not to use TypeScript-internal APIs in our rules. While these type assignability APIs have coincidentally been stable for quite some time, we've tried to be good citizens and not rely on them...

packages/eslint-plugin/src/rules/no-unsafe-unary-minus.ts
```diff
@@ -0,0 +1,55 @@
...
+        const services = util.getParserServices(context);
+        const argType = services.getTypeAtLocation(node.argument);
+        const checker = services.program.getTypeChecker() as TypeChecker;
+        if (
+          !checker.isTypeAssignableTo(
+            argType,
+            checker.getUnionType([
+              checker.getNumberType(), // first exposed in TypeScript v5.1
+              checker.getBigIntType(), // first added in TypeScript v5.1
+            ]),
```
JoshuaKGoldberg, line 43: ...that being said, I don't think you need the type assignability APIs for this specific rule. I think you should be able to just compare with `ts.TypeFlags`. Search on examples of `isTypeFlagSet` and `ts.TypeFlags.Num` in code for examples.

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  valid: [
+    '+42;',
+    '-42;',
+    '-42n;',
+    '(a: number) => -a;',
+    '(a: bigint) => -a;',
+    '(a: number | bigint) => -a;',
+    '(a: any) => -a;',
+    '(a: 1 | 2) => -a;',
+  ],
```
JoshuaKGoldberg, line 26: [Testing] Some missing coverage for:

- A valid case that would be invalid with a different operator (e.g. `(a: string) => +a;`)
- Arrays
- Intersections
- Objects
- Strings
- Template literal strings
- `never`
- `unknown`
- `void`

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  },
+  parser: '@typescript-eslint/parser',
+});
+
+ruleTester.run('no-unsafe-unary-minus', rule, {
+  valid: [
+    '+42;',
+    '-42;',
+    '-42n;',
+    '(a: number) => -a;',
```
JoshuaKGoldberg, line 21: [Bug] This isn't valid syntax. It might coincidentally work for now, but there's no guarantee it always will.

For this and the other `=>` cases:

```suggestion
    'const _ = (a: number) => -a;',
```

...though, I'd personally suggest using a `declare const` instead, as it's a little more clear what's being tested. 🤷 up to you!

```ts
declare const a: number;
-a;
```

JoshuaKGoldberg (changes requested):
A solid start, thanks for sending in! ✨ 

Requesting changes on using more direct, public APIs, and some more testing.

packages/eslint-plugin/src/rules/no-unsafe-unary-minus.ts
samestep: Does this actually work? I wrote an initial version using just `TypeFlags`, and it failed even on a few of the simple test cases I currently have. Are you saying `TypeFlags.Num` is just straight-up equivalent to being assignable to `number | bigint`, or is there more logic I'd need to write on top of that?

JoshuaKGoldberg: I haven't tried it locally, but in the past I've had to use variants like `TypeFlags.NumberLike` (see `no-unsafe-enun-comparison`) and/or `TypeFlags.BigIntLike` (see `restrict-template-expressions`).

samestep: Ah OK; I tried those in my initial implementation and it didn't work on unions, for instance. (And I believe `any` and literals, but those have their own flags.) I didn't go further down that route because I don't have any idea how more complicated types like intersections or mapped types etc would interact with it.

JoshuaKGoldberg: > unions

Ah yes unions can be tricky. Rules often have to `tsutils.unionTypeParts` to get them, then run a check on each of the constituents.

> intersections or mapped types

Fortunately, you probably don't need to do anything special for them. `getTypeAtLocation` will retrieve whatever the result of their type is.

---

And, this reminds me - there should also be some tests for generics / type parameters. `util.getConstrainedTypeAtLocation` should be used.

* `<T>(t: T) => -t` -> that might be unsafe, since we don't know what `T` is
* `<T extends number>(t: T) => -t` -> that isn't unsafe, since although we don't know what `T` is, we do know it's _constrained_ to `number`

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
Josh-Cena: > This isn't valid syntax

In what sense is it not? 🤔

JoshuaKGoldberg: OH wow I am wrong! That is valid syntax! I guess it makes sense that you could have an arrow function as an expression... never mind! 😄

JoshuaKGoldberg:
👋 Hey @samestep! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

samestep:
@JoshuaKGoldberg thanks for the ping! Sorry for dropping the ball on this; I'm taking a look now, will see if I can figure out how to restructure it according to the feedback you gave in August.

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
samestep: Done, thanks!

packages/eslint-plugin/docs/rules/no-unsafe-unary-minus.md
```diff
@@ -0,0 +1,37 @@
...
+TypeScript does not prevent you from putting a minus sign before things other than numbers:
+
+```ts
+const s = 'hello';
+const x = -s; // x is NaN
+```
+
+This rule restricts the unary `-` operator to `number | bigint`.
+
+## Examples
```
JoshuaKGoldberg, line 18: [Docs] I like that this shows a more real-world usage (functions) than just literals. But it's a bit more to read when they're all functions. What do you think about making most of the examples literals?

Another syntax I like to use in cases like this is `declare`:

```ts
declare const eitherNumeric: number | bigint;
-eitherNumeric;
```

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    '(a: bigint) => -a;',
+    '(a: number | bigint) => -a;',
+    '(a: any) => -a;',
+    '(a: 1 | 2) => -a;',
+    '(a: string) => +a;',
+    '(a: number[]) => -a[0];',
+    '<T,>(t: T & number) => -t;',
+    '(a: { x: number }) => -a.x;',
+    '(a: never) => -a;',
```
JoshuaKGoldberg, line 30: [Functionality] Interesting. What use would there be for `-a`? I suspect this should be an error, but am not sure...

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  valid: [
+    '+42;',
+    '-42;',
+    '-42n;',
+    '(a: number) => -a;',
+    '(a: bigint) => -a;',
+    '(a: number | bigint) => -a;',
+    '(a: any) => -a;',
+    '(a: 1 | 2) => -a;',
+    '(a: string) => +a;',
```
JoshuaKGoldberg, line 26: [Functionality] ...huh. I at first thought this was a bug, and that `+` shouldn't be allowed. But the rule is just for `unary-minus`. Maybe this is evidence that the rule should be a more broad `no-unsafe-unary`, for `+` on non-numeric values too? Thoughts?

(if so we should probably discuss in the issue)

JoshuaKGoldberg (changes requested):
Looking great! A few comments, mostly more questions than anything else. But I am starting to wonder about increasing the rule's scope... 🤔

packages/eslint-plugin/docs/rules/no-unsafe-unary-minus.md
samestep: OK! I replaced all the functions with `declare`s.

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
samestep: No, the `never` type is supposed to have no elements, to any operation on it is valid.

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
Josh-Cena: `+x` is a common synonym for `Number(x)`, so this will surely make some unhappy.

samestep: Yeah, we could check `+` too, but in my opinion there's no need: any person using this rule would probably already be using ESLint's [no-implicit-coercion](https://eslint.org/docs/latest/rules/no-implicit-coercion) rule, which would already catch all instances of unary `+`. I could remove this particular test case if you want, though, if it's confusing.

JoshuaKGoldberg: Right, that makes sense. I think it's fine as-is. @Josh-Cena if you want to push either direction I'll go with that 😄

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
JoshuaKGoldberg: But why would someone `-` it? To me, that's more likely a sign of a mistake: that they thought it had a value but it actually doesn't.

samestep: Sure, but that's entirely orthogonal: if we want to make a rule saying not to use a `never` type in a context that allows a broader type, then we could (although it seems a bit weird), but it is unrelated to this rule about `-`.

JoshuaKGoldberg: cc @typescript-eslint/triage-team - I'm of the opinion this rule would make sense for that. Anybody have thoughts?

samestep: To clarify my position: I'm actually pretty strongly opposed to disallowing `never` in this rule. The whole point of `never` is that it serves as a [bottom type](https://en.wikipedia.org/wiki/Bottom_type), and it arises naturally when you take the intersection of two disjoint types. It would be inconsistent and surprising for it to be disallowed here.

Josh-Cena: I don't see why forbidding it either—it's mathematically sound. Our heuristic is basically "allow only operands assignable to `number`/`bigint`" and `never` fits.

packages/eslint-plugin/tests/rules/no-unsafe-unary-minus.test.ts
Josh-Cena: Yes, let's keep it as-is. I don't think it's worth removing either—always good to delineate the boundaries of a rule so future behavior changes are obvious.

JoshuaKGoldberg (approved):
🚀 swell, thanks!

PR #7394, opened 2023-08-01
fix(eslint-plugin): [strict-boolean-expressions] replace dangerous fixer with suggestions

### LINKED ISSUE #7312: Bug: @typescript-eslint/strict-boolean-expressions breaks code on --fix

### Issue Description

Running `eslint --fix` breaks code when the rule `@typescript-eslint/strict-boolean-expressions` has the option `allowNullableObject` set to false

Code before fix
```
type Items = {
    foo: string
}
const items: Items[] = [
    {foo: 'bar'},
    {foo: 'bar'},
]
function someFn (items?: Items[]) {
    return JSON.parse(
        JSON.stringify(items || []),
    );
}
const test = someFn(items)
console.log(test); // [ { foo: 'bar' }, { foo: 'bar' } ]
```
code after fix
```
...
function someFn (items?: Items[]) {
    return JSON.parse(
        JSON.stringify((items != null) || []),
    );
}
const test = someFn(items)
console.log(test); // true
```

**What did you expect to happen?**
Code gets changed with the `??` operator or not changed at all

https://github.com/fasenderos/typescript-eslint-strict-boolean

### Repro Steps

Clone the repo and run `npm install`, then try the code before fix

- `npm run build`
- `npm start`

The output will be `[ { foo: 'bar' }, { foo: 'bar' } ]`

Now run the fix and try again

- `npm run lint:fix`
- `npm run build`
- `npm start`

The output will be `true`

### Versions

| `@typescript-eslint/eslint-plugin`     | `^6.2.0` |
| `@typescript-eslint/parser`            | `^6.2.0` |
| `TypeScript`                           | `^5.1.6` |
| `ESLint`                               | `^8.45.0` |
| `node`                                 | `18.16.1` |

kemilbeltre:
Remove explicit rule check for logical negation expressions

## PR Checklist

- [x] Addresses an existing open issue: fixes #7312


packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -715,11 +715,6 @@ export default util.createRule<Options, MessageId>({
             context.report({
               node,
               messageId: 'conditionErrorNullableObject',
-              fix: util.getWrappingFixer({
-                sourceCode,
-                node,
-                wrap: code => `${code} != null`,
-              }),
```
JoshuaKGoldberg, line 722: Could you move this to a [_suggestion_ (`suggest`)](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions) instead? It's useful to offer this to users, just we can't auto-apply it under `--fix`.

JoshuaKGoldberg (changes requested):
Close!

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -715,11 +715,16 @@ export default util.createRule<Options, MessageId>({
...
+              suggest: [
+                {
+                  messageId: 'conditionErrorNullableObject',
+                  fix: util.getWrappingFixer({
+                    sourceCode,
+                    node,
+                    wrap: code => `${code} != null`,
+                  }),
+                },
+              ],
```
JoshuaKGoldberg, line 727: [Testing] And now that this is a `suggest` instead of a `fix`, we'll need unit testing for it. Could you add them in please?

`suggestions: [` is a good search term for other rules that give suggestions.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -715,11 +715,16 @@ export default util.createRule<Options, MessageId>({
...
               node,
               messageId: 'conditionErrorNullableObject',
-              fix: util.getWrappingFixer({
-                sourceCode,
-                node,
-                wrap: code => `${code} != null`,
-              }),
+              suggest: [
+                {
+                  messageId: 'conditionErrorNullableObject',
```
JoshuaKGoldberg, line 720: [Bug] Ah, suggestion messages should generally to be different from the parent report. You'll want to add a new `messageId` for this case.

Filed https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/368 to help automate enforcing this.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -715,11 +715,16 @@ export default util.createRule<Options, MessageId>({
             context.report({
               node,
               messageId: 'conditionErrorNullableObject',
```
JoshuaKGoldberg, line 717: [Bug] So I just scrolled up a bit in the file and found another usage of `messageId: 'conditionErrorNullableObject'`. Looks like there's another `fix` that uses this same message ID. I think this'll need to also switch that one to a `suggestion`.

JoshuaKGoldberg (changes requested):
Getting closer! 🙌

JoshuaKGoldberg (approved):
Super, thanks @kemilbeltre! 🙌

PR #7404, opened 2023-08-03
docs: modify examples for explicit-module-boundary-types

### LINKED ISSUE #7430: Docs: modify examples for explicit-module-boundary-types

### Suggested Changes

1.
image

2.
image

3.
image

I had difficulty understanding the `explicit-module-boundary-types` rule when I first encountered it due to unclear and non-intuitive examples.

In the initial description of this rule, it states, "Require explicit return and argument types on exported functions and classes public class methods." When writing code that violates this rule in the playground, it shows an error message like "Missing return type on function," as seen in the image above. Furthermore, in the third image, the comment for Incorrect code reads, "Should indicate that no value is returned (void)."

Based on this information, one way to address this issue is to avoid exporting meaningless functions or classes. However, to better understand and adhere to this rule in a clear and intuitive manner, I believe it is more appropriate to follow the approach described in the document: explicitly specifying types to avoid errors. By doing so, we can easily grasp the essence of the rule, which emphasizes "Requiring explicit return and argument types on exported functions and classes public class methods", creating more straightforward and intuitive examples.

# before

### ✅ Correct

```ts
// Function is not exported
function test() {
  return;
}

// A return value of type number
export var fn = function (): number {
  return 1;
};
// A return value of type string
export var arrowFn = (): string => 'test';
// All arguments should be typed
export var arrowFn = (arg: string): string => `test ${arg}`;
export var arrowFn = (arg: unknown): string => `test ${arg}`;

// Class is not exported
class Test {
  method() {
    return;
  }
}
```

# after

### ✅ Correct

```ts
// A function with no return value (void)
export function test(): void {
  return;
}

// A return value of type number
export var fn = function (): number {
  return 1;
};
// A return value of type string
export var arrowFn = (): string => 'test';
// All arguments should be typed
export var arrowFn = (arg: string): string => `test ${arg}`;
export var arrowFn = (arg: unknown): string => `test ${arg}`;

// A class method with no return value (void)
export class Test {
  method(): void {
    return;
  }
}
```

## 📌 Answer : Josh-Cena

image

## 📌 Answer : ryan-dia

When I first encountered these two points in the rule, having both of them in a single example made it harder to understand. Therefore, as you suggested, instead of putting both points in one example, dividing them into separate examples would make it much more intuitive, clear, and easier to comprehend.   [applied code](https://github.com/typescript-eslint/typescript-eslint/pull/7404/commits/8eaf4cd07005d20195d70574c5d7933ed56ffb14)

Alternatively, in my opinion, since the rule name itself is not "explicit-types" but "explicit-module-boundary-types," it is natural to assume that it won't apply unless it's related to module boundaries. Hence, there might be no need to include examples for cases outside module boundaries. Moreover, in the initial explanation of the rule, it states "on exported functions' and classes' public class methods," and in the "When Not To Use It" section, it mentions, "If you wish to make sure all functions have explicit return types, as opposed to only the module boundaries, you can use explicit-function-return-type." This clearly indicates that the rule applies only to module boundaries. Users can infer this, and if you still wish to include the examples, it would be better to present them as follows.

# 🛠️  case 1

### ❌ Incorrect

```ts
// Should indicate that no value is returned (void)
export function test() {
  return;
}

// Should indicate that a number is returned
export default function () {
  return 1;
}

// Should indicate that a string is returned
export var arrowFn = () => 'test';

// All arguments should be typed
export var arrowFn = (arg): string => `test ${arg}`;
export var arrowFn = (arg: any): string => `test ${arg}`;

export class Test {
  // Should indicate that no value is returned (void)
  method() {
    return;
  }
}

// The function does not apply because it is not an exported function.
function test() {
  return;
}
```

### ✅ Correct

```ts
// A function with no return value (void)
export function test(): void {
  return;
}

// A return value of type number
export var fn = function (): number {
  return 1;
};
// A return value of type string
export var arrowFn = (): string => 'test';
// All arguments should be typed
export var arrowFn = (arg: string): string => `test ${arg}`;
export var arrowFn = (arg: unknown): string => `test ${arg}`;

// A class method with no return value (void)
export class Test {
  method(): void {
    return;
  }
}

// The function does not apply because it is not an exported function.
function test() {
  return;
}
```

# 🛠️ case 2

### ❌ Incorrect

```ts
// Should indicate that no value is returned (void)
export function test() {
  return;
}

// Should indicate that a number is returned
export default function () {
  return 1;
}

// Should indicate that a string is returned
export var arrowFn = () => 'test';

// All arguments should be typed
export var arrowFn = (arg): string => `test ${arg}`;
export var arrowFn = (arg: any): string => `test ${arg}`;

export class Test {
  // Should indicate that no value is returned (void)
  method() {
    return;
  }
}
```

### ✅ Correct

```ts
// If you remove the export, you have the option to either write explicit return and argument types or not. 

// A function with no return value (void)
export function test(): void {
  return;
}

// A return value of type number
export var fn = function (): number {
  return 1;
};
// A return value of type string
export var arrowFn = (): string => 'test';
// All arguments should be typed
export var arrowFn = (arg: string): string => `test ${arg}`;
export var arrowFn = (arg: unknown): string => `test ${arg}`;

// A class method with no return value (void)
export class Test {
  method(): void {
    return;
  }
}
```

### Affected URL(s)

https://typescript-eslint.io/rules/explicit-module-boundary-types/

### Reference
#7404

Ryan-Dia:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7430

## Overview

I had difficulty understanding the `explicit-module-boundary-types` rule when I first encountered it due to unclear and non-intuitive examples.

In the initial description of this rule, it states, "Require explicit return and argument types on exported functions and classes public class methods." When writing code that violates this rule in the playground, it shows an error message like "Missing return type on function," as seen in the image above. Furthermore, in the third image, the comment for Incorrect code reads, "Should indicate that no value is returned (void)."

Based on this information, one way to address this issue is to avoid exporting meaningless functions or classes. However, to better understand and adhere to this rule in a clear and intuitive manner, I believe it is more appropriate to follow the approach described in the document: explicitly specifying types to avoid errors. By doing so, we can easily grasp the essence of the rule, which emphasizes "Requiring explicit return and argument types on exported functions and classes public class methods", creating more straightforward and intuitive examples.

### Reference

https://typescript-eslint.io/rules/explicit-module-boundary-types/


Josh-Cena (commented):
The point of this rule is:

1. Exported members must have explicit signatures
2. Non-exported members may or may not have explicit signatures

Your change makes everything exported, so the second case is lost. I'm not sure how this is an improvement. If you wish, though, we can have both an exported and a non-exported example.

Ryan-Dia:
> The point of this rule is:
> 
> 1. Exported members must have explicit signatures
> 2. Non-exported members may or may not have explicit signatures
> 
> Your change makes everything exported, so the second case is lost. I'm not sure how this is an improvement. If you wish, though, we can have both an exported and a non-exported example.


When I first encountered these two points in the rule, having both of them in a single example made it harder to understand. Therefore, as you suggested, instead of putting both points in one example, dividing them into separate examples would make it much more intuitive, clear, and easier to comprehend.

As you mentioned, I created two separate examples for each of the two points you made. Now, what do you think would be the best way to modify them?

packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md
```diff
@@ -62,14 +64,50 @@ export var arrowFn = (): string => 'test';
...
-  method() {
+// A class method with no return value (void)
+export class Test {
+  method(): void {
     return;
   }
 }
 ```
 
+### 2. Non-exported members may or may not have explicit signatures
```
JoshuaKGoldberg, line 75: Formatting is a little funky here:

Screenshot of "2. Non-exported ..." heading interpreted as a middle tab amongst five

JoshuaKGoldberg (changes requested):
Per https://deploy-preview-7404--typescript-eslint.netlify.app/contributing/pull-requests, we ask that each sent PR either addresses an open issue or is for an "extremely minor" documentation typos. This isn't minor. Could you file an issue please? I think we'd want to discuss this a bit more.

Ryan-Dia:
> Per https://deploy-preview-7404--typescript-eslint.netlify.app/contributing/pull-requests, we ask that each sent PR either addresses an open issue or is for an "extremely minor" documentation typos. This isn't minor. Could you file an issue please? I think we'd want to discuss this a bit more.

At first, I thought it was a minor matter, but as we talked, it seemed that there were more significant aspects to consider. As you mentioned, I have filed the issue. Thank you.

Issues : #7430

packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md
Ryan-Dia: thank you 👍
The formatting has been adjusted

packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md
```diff
@@ -12,9 +12,11 @@ It can also improve TypeScript type checking performance on larger codebases.
 
 ## Examples
 
+### 1. Exported members must have explicit signatures
+
 <!--tabs-->
```
JoshuaKGoldberg, line 17: Looking at this first pair of tabs:

1. ✏️ `export function test() {`: Great!
2. ✏️ `export default function () {`: Odd that the _Correct_ version is different. I'd say it should probably instead have an explicit return type.
3. `export var arrowFn`: Great!
   - ✏️ I wonder if we can delete number 2 (`export default function() {`), since this `arrowFn` also serves the purpose of showing different syntax?
4. `export var arrowFn = (arg):`: Great!
5. `export var arrowFn = (arg: any):`: Great!
6. ✏️ `export class Test {`: Mostly great, but can you move the _Correct_ code's comment to just above `method()`? It's confusing IMO to see it change location (and the new location is inaccurate to what the comment says)

✏️ indicates a request for changes :)

packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md
```diff
@@ -62,14 +64,50 @@ export var arrowFn = (): string => 'test';
...
+export class Test {
+  method(): void {
     return;
   }
 }
 ```
 
+### 2. Non-exported members may or may not have explicit signatures
+
+<!--tabs-->
```
JoshuaKGoldberg, line 77: I also find it confusing how this second pair of tabs has such different code between _Incorrect_ and _Correct_. And all it's trying to convey is that an alternative to adding an explicit type annotation is to not export a function.

Maybe instead of having a whole second tabs pair, a single mention could be made in the first tabs pair? I think that'd be easier to read & would result in a more streamlined docs page.

JoshuaKGoldberg (changes requested):
Thanks for the issue & discussion! I like the direction this is taking 😄 just requesting some touchups to streamline a bit. 🚀

Ryan-Dia:
>3. export var arrowFn: Great!
✏️ I wonder if we can delete number 2 (export default function() {), since this arrowFn also serves the purpose of showing different syntax?
>6. ✏️ export class Test {: Mostly great, but can you move the Correct code's comment to just above method()? It's confusing IMO to see it change location (and the new location is inaccurate to what the comment says)   

All of these requests have been reflected. 😁
 
 
> I also find it confusing how this second pair of tabs has such different code between Incorrect and Correct. And all it's trying to convey is that an alternative to adding an explicit type annotation is to not export a function.

>Maybe instead of having a whole second tabs pair, a single mention could be made in the first tabs pair? I think that'd be easier to read & would result in a more streamlined docs page.   

I agree. 
I also want to address this only in the first tab pair.
So, I suggested two additional styles in the [issue](https://github.com/typescript-eslint/typescript-eslint/issues/7430)
Which one do you like?
(The code currently in the PR is for case 3)

>case 1
```ts

// A function with no return value (void)
export function test(): void {
  return;
}

// A return value of type string
export var arrowFn = (): string => 'test';

// All arguments should be typed
export var arrowFn = (arg: string): string => `test ${arg}`;
export var arrowFn = (arg: unknown): string => `test ${arg}`;

export class Test {
// A class method with no return value (void)
  method(): void {
    return;
  }
}

// The function does not apply because it is not an exported function.
function test() {
  return;
}
```

>case 2
```ts
// If you remove the export, you have the option to either write explicit return and argument types or not. 

// A function with no return value (void)
export function test(): void {
  return;
}

// A return value of type string
export var arrowFn = (): string => 'test';

// All arguments should be typed
export var arrowFn = (arg: string): string => `test ${arg}`;
export var arrowFn = (arg: unknown): string => `test ${arg}`;

export class Test {
// A class method with no return value (void)
  method(): void {
    return;
  }
}
```

>case 3 (pr)
```ts
// case 1: A function with no return value (void)
export function test(): void {
  return;
}
// case 2: Function is not exported
function test() {
  return;
}

// case 1: A return value of type string
export var arrowFn = (): string => 'test';
// case 2: arrowFn is not exported
var arrowFn = (arg): string => `test ${arg}`;

// case 1: All arguments should be typed
export var arrowFn = (arg: string): string => `test ${arg}`;
export var arrowFn = (arg: unknown): string => `test ${arg}`;
// case 2: arrowFn is not exported
var arrowFn = (arg): string => `test ${arg}`;
var arrowFn = (arg: any): string => `test ${arg}`;

export class Test {
  // case 1: A class method with no return value (void)
  method(): void {
    return;
  }
}
// case 2: Class is not exported
class Test {
  method() {
    return;
  }
}
```

JoshuaKGoldberg:
Err, I think case 1? It's been a few weeks and there's a lot of text to read through 😅 sorry.

Ryan-Dia:
> Err, I think case 1? It's been a few weeks and there's a lot of text to read through 😅 sorry.

That's all right.

They said it was confusing because it was divided into two tabs before, so I would like to write it as one tab.

I prepared 3 cases of styles.

Currently, case 3 style is applied to PR.

Does your response mean that Case 1 style is good?

JoshuaKGoldberg:
Let's go with case 1 and then we can review from there 👍

Ryan-Dia:
> 👋 Hey @Ryan-Dia! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

I'm sorry for the delay. I applied it to case 1 you mentioned last time.

JoshuaKGoldberg (approved):
This is great, thanks! The changes make the correct/incorrect flipping much more indicative of what the rule enforces. 🚀 love it!

Michael Scott saying 'I love it! I love it!'

PR #7406, opened 2023-08-03
fix(utils): add to JSONSchema4Type missing Array and Object

### LINKED ISSUE #7412: Bug: Typescript error in default value for new rule schema

### Issue Description

In a new rule meta.schema an entry of, for example:

```
{
                type: 'object',
                properties: {
                  allowedTypeAnnotations: {
                    type: 'array',
                    items: { type: 'string' },
                    default: [],
                  },
                },
                additionalProperties: false,
 },
```

Gives an error for the default value of `[]`, where I expect it to be a valid value
This is due to missing types for JSONSchema4Array and JSONSchema4Object in JSONSchema4Type.
Those types exist in json-schema: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/json-schema/index.d.ts#L30

PR with the fix:
https://github.com/typescript-eslint/typescript-eslint/pull/7406 

https://github.com/wespickett/typescript-eslint-ex1/tree/main

### Repro Steps

1) clone the repo
2) npm install
3) Open in an editor like VSCode to see the error in example-rule.ts, or run `npx tsc`

### Versions

| `@typescript-eslint/eslint-plugin`     | `6.3.0` |
| `@typescript-eslint/parser`            | `6.3.0` |
| `@typescript-eslint/scope-manager`     | `6.3.0` |
| `@typescript-eslint/typescript-estree` | `6.3.0` |
| `@typescript-eslint/type-utils`        | `6.3.0` |
| `@typescript-eslint/utils`             | `6.3.0` |
| `TypeScript`                           | `5.1.6` |
| `ESLint`                               | `8.46.0` |
| `node`                                 | `18.14.2` |

wespickett:
Updates type to add missing Array and Object

From https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/json-schema/index.d.ts#L30

## PR Checklist

- [ ] Addresses an existing open issue: fixes #7412


bradzacher:
As per our [contributing guide](https://typescript-eslint.io/contributing/pull-requests) please file an issue first so that we can discuss the change and the usecases you're intending to target with it.

Closing pending discussion.

packages/utils/src/json-schema.ts
```diff
@@ -25,7 +25,23 @@ export type JSONSchema4TypeName =
 /**
  * @see https://tools.ietf.org/html/draft-zyp-json-schema-04#section-3.5
  */
-export type JSONSchema4Type = boolean | number | string | null;
+export type JSONSchema4Type =
+  | boolean
+  | JSONSchema4Array
+  | JSONSchema4Object
```
bradzacher, line 31: I don't think it is the best thing for us to add this here because `JSONSchema4Type` is used in the `enum` property where arrays and objects are disallowed (slash non-sensical).

So I think it makes sense for us to define a new type `JSONSchema4DefaultValue` which includes all of the valid default values.

packages/utils/src/json-schema.ts
```diff
@@ -460,7 +476,7 @@ export interface JSONSchema4NumberSchema extends JSONSchema4Base {
 /**
  * @see https://json-schema.org/understanding-json-schema/reference/boolean.html
  */
-export interface JSONSchema4BoleanSchema extends JSONSchema4Base {
+export interface JSONSchema4BooleanSchema extends JSONSchema4Base {
```
bradzacher, line 479: goodcatch

Spelling is hard 😅
Technically this is a breaking change but given it's a cock up I'm fine with doing this here. Given nobody noticed I doubt anyone is using it directly yet!

packages/utils/src/json-schema.ts
wespickett: @bradzacher Updated with a separate type named `JSONSchema4TypeExtended`, which I thought could be a better name, but can change it to `JSONSchema4DefaultValue` if needed

bradzacher (approved):
I'll leave this as a future exercise - but we can go one step further with this and narrow the default prop for each type. Eg a boolean type schema can only have a boolean default, and an array type schema can only have an array.

Thanks for fixing this!

PR #7440, opened 2023-08-07
docs: warn against progressive type enhancement in rules

### LINKED ISSUE #7158: Docs: Advise against "progressive enhancement" with rule type information

### Suggested Changes

Coming over from https://github.com/typescript-eslint/typescript-eslint/issues/7124#issuecomment-1615640551, quoting @bradzacher:

> ... we've always given the advice that you shouldn't do progressive enhancement of lint rules depending on if type information might be there.
>
> Thats the same hard-and-fast rule that we have taken with every single one of our lint rules.
>
> The reason we take this line is because config is hard and if you just silently enhance a rule with type information but the user misconfigures it how can you alert the user to the misconfiguration?

This is good advice, but I don't believe it's in the public rule docs? We should add it so folks know. This comes up once in a while on social media too.

### Affected URL(s)

https://typescript-eslint.io/developers/custom-rules#typed-rules

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7158

## Overview

Adds a caution to the docs.

I wonder if this is overkill...?

### DIFF
--- a/docs/developers/Custom_Rules.mdx
+++ b/docs/developers/Custom_Rules.mdx
@@ -273,10 +273,10 @@ This can be necessary for TypeScript APIs not wrapped by the parser services.
 :::
 
 :::caution
-We recommend against changing rule logic based on whether `services.program` exists.
+We recommend against changing rule logic based solely on whether `services.program` exists.
 In our experience, users are generally surprised when rules behave differently with or without type information.
 Additionally, if they misconfigure their ESLint config, they may not realize why the rule started behaving differently.
-Consider creating two versions of the rule instead, or only creating a type checked version.
+Consider either gating type checking behind an explicit option for the rule or creating two versions of the rule instead.
 :::
 
 ## Testing
--- a/docs/developers/Custom_Rules.mdx
+++ b/docs/developers/Custom_Rules.mdx
@@ -276,6 +276,7 @@ This can be necessary for TypeScript APIs not wrapped by the parser services.
 We recommend against changing rule logic based on whether `services.program` exists.
 In our experience, users are generally surprised when rules behave differently with or without type information.
 Additionally, if they misconfigure their ESLint config, they may not realize why the rule started behaving differently.
+Consider creating two versions of the rule instead, or only creating a type checked version.
 :::
 
 ## Testing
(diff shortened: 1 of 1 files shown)


bradzacher (dismissed):

docs/developers/Custom_Rules.mdx
```diff
@@ -272,6 +272,12 @@ Rules can retrieve their full backing TypeScript type checker with `services.pro
 This can be necessary for TypeScript APIs not wrapped by the parser services.
 :::
 
+:::caution
+We recommend against changing rule logic on whether `services.program` exists.
```
Josh-Cena, line 276:
```suggestion
We recommend against changing rule logic based on whether `services.program` exists.
```

docs/developers/Custom_Rules.mdx
```diff
@@ -272,6 +272,12 @@ Rules can retrieve their full backing TypeScript type checker with `services.pro
...
+In our experience, users are generally surprised when rules behave differently with or without type information.
+Additionally, if they misconfigure their ESLint config, they may not realize why the rule started behaving differently.
```
Josh-Cena, line 278: This sounds vague. Maybe a specific case?

JoshuaKGoldberg: Hmm yeah good point. Agreed.

armano2: as possible solution we could recommend to create 2 separate rules one with type-info and one without

JoshuaKGoldberg: I like it! Added :)

bradzacher: Also gating it behind options is *okay*, I think.
That way you still have that hard point of a user explicitly opting in to the type-awareness and thus the rule can explicitly crash if it's not provided.

`naming-convention` does this (the only rule we have that does) for the type-based enforcement of names.

JoshuaKGoldberg: This actually is very relevant to https://github.com/veritem/eslint-plugin-vitest/issues/251. Nice.

PR #7447, opened 2023-08-10
docs: add 'When Not To Use It' or an intentional omission notice on all rules

### LINKED ISSUE #7257: Repo: Enforce that all rules have a "When Not To Use It" docs section?
### Suggestion

Right now, about 70 rule docs files have a `## When Not To Use It` heading. Is this practice something we want to standardize for all rules and enforce as a CI/lint/test check?

FWIW I'd be in favor. I think it's good to always portray a nuanced view of when you would or wouldn't want to use a rule.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7257

## Overview

**Context**: We'd like to bolster the sections in our docs that explain when _not_ to use certain lint rules. Virtually all of our rules -even the exceptionally useful ones- have tradeoffs that can make them harder to use in _some_ projects. We'd love to know what those tradeoffs are so we can document them!

~Posting as a reference draft for #7257.~ Marking as ready for review. I'd love any and all feedback on this: what reasons have you seen to _not_ use specific lint rules? Please do comment them here, in the file for the corresponding lint rule! _(or DM/email me somewhere if you want to send private feedback)_

I can't promise we'll agree with all reasons 😄 but at the very least they're good for us to know. And oftentimes _"our project wasn't architected in a way to allow this code pattern"_ is a good reason for us to mention.


packages/eslint-plugin/docs/rules/ban-types.md
```diff
@@ -181,3 +181,7 @@ Example configuration:
   ]
 }
 ```
+
+## When Not To Use It
+
+If your project intentionally deals with the class equivalents of primitives, it might not be worthwhile to enable `ban-types`.
```
bradzacher, line 187: So this isn't really correct - it is for the defaults, sure, but this rule could ban anything!

packages/eslint-plugin/docs/rules/no-duplicate-enum-values.md
```diff
@@ -48,3 +48,9 @@ enum E {
   B = 'B',
 }
 ```
+
+## When Not To Use It
+
+It can sometimes be useful to include duplicate enum members.
+For example, when renaming an enum member, it can sometimes be useful to keep the old name until a scheduled major breaking change.
```
bradzacher, line 55: This isn't a reason to avoid the rule, IMO - it's a reason for a disable comment though!

packages/eslint-plugin/docs/rules/no-duplicate-type-constituents.md
```diff
@@ -57,3 +57,7 @@ When set to true, duplicate checks on intersection type constituents are ignored
 ### `ignoreUnions`
 
 When set to true, duplicate checks on union type constituents are ignored.
+
+## When Not To Use It
+
+If you don't mind excess types taking up space in intersections or unions, you can avoid this rule.
```
bradzacher, line 63: This is the sort of sentence I was meaning by my comment.
All this really does is invert the rule name.

"If you don't want your types to have no duplicate type constituents, use this rule"
"If you don't mind your types having duplicate type constituents, don't use this rule"

Personally I don't think this one adds any value to the rule docs.
Which is why I say "prefer to add one, but not enforce they all have one".

packages/eslint-plugin/docs/rules/no-extra-non-null-assertion.md
```diff
@@ -50,3 +50,7 @@ function foo(bar?: { n: number }) {
   return bar?.n;
 }
 ```
+
+## When Not To Use It
+
+If you don't mind excess `!`s in your code, you can avoid this rule.
```
bradzacher, line 56: same as above

"Use this rule if you want to remove extra non-null-assertions"
"Don't use this rule if you are okay with extra non-null-assertions"

packages/eslint-plugin/docs/rules/no-duplicate-enum-values.md
Josh-Cena: It occurs to me that "When Not To Use It" can mean both "use a disable comment" and "not use this rule altogether"—often the former is more meaningful because it delineates acknowledged edge-cases that the rule will not special case.

bradzacher: Yeah now that's something I could get behind - documenting clear edge-cases where we recommend using a disable comment!

packages/eslint-plugin/docs/rules/no-extra-non-null-assertion.md
JoshuaKGoldberg: ...yeah I truly cannot think of a reason not to use this rule 😄

packages/eslint-plugin/docs/rules/no-duplicate-type-constituents.md
JoshuaKGoldberg: Finally getting back to this! I think I got this intent in all the rules I could. For any others I added a `<!-- Intentional omission: When Not To Use It -->` comment. WDYT?

packages/website/src/components/Playground.module.css
```diff
@@ -66,7 +66,9 @@
   justify-content: stretch;
   align-items: stretch;
   outline: none;
-  transition: border-color 0.2s linear, background-color 0.2s linear;
+  transition:
+    border-color 0.2s linear,
+    background-color 0.2s linear;
```
JoshuaKGoldberg, line 71: 🤔

packages/eslint-plugin/docs/rules/no-unsafe-member-access.md
Faithfinder, line 1: You've asked for comments - we find this rule (and others from the family) redundant with both no explicit and implicit `any` turned on. If something is of type `any`, it's because somebody made it so, disabling the linter. No point hassling them any further.

packages/eslint-plugin/docs/rules/restrict-template-expressions.md
Faithfinder, line 1: Doesn't play too well with [referring to other components](https://styled-components.com/docs/advanced#referring-to-other-components) in styled components

packages/eslint-plugin/docs/rules/no-duplicate-type-constituents.md
bradzacher: Yeah I'm more than happy if we have a way to explicitly opt-out!

packages/eslint-plugin/docs/rules/no-unsafe-member-access.md
JoshuaKGoldberg: Ooh interesting, thanks! If you have any references I'd love to see them. IME it does sometimes come up (e.g. `JSON.parse`) but I can definitely see a world where that's not a real problem (e.g. using [`ts-reset`](https://github.com/total-typescript/ts-reset)).

packages/eslint-plugin/docs/rules/restrict-template-expressions.md
JoshuaKGoldberg: Ah super interesting, thanks! I wonder if folks have ever gotten this rule working usefully in a codebase that uses Emotion, styled-components, or similar...

packages/eslint-plugin/docs/rules/no-unsafe-member-access.md
Faithfinder: Here are some of the things that light up when I turn these rules on in our codebase:

Legacy pre-typescript tests, decided to keep them since we still have JS code
image

Some shitty code that shouldn't really be using `any`, but it does, so, why are you hassling me?
image

Explicitly disabled `as any`, what more do you want?
image

These are mostly artifacts of a legacy codebase that was JS before it was TS, so it's a variant of "too much effort", really, but point is, "No explicit/implicit any` already highlights all these problems, I don't need a second rule shouting at me

packages/eslint-plugin/docs/rules/restrict-template-expressions.md
JoshuaKGoldberg: Huh, I was about to file an issue for this, but I can't repro any issues when using `styled`.

```ts
import { styled } from "styled-components";

export const Link = styled.a`
  background: papayawhip;
`;

export const Icon = styled.svg`
  ${Link}:hover & {
    fill: rebeccapurple;
  }
`;

export function test() {
  console.log(`hi ${test}`);
}
```

The `hi ${test}` does result in a complaint, but not the `${Link}`.

packages/eslint-plugin/docs/rules/no-unsafe-member-access.md
JoshuaKGoldberg: Thought about this a bit: changed the wording of the new message to _If your codebase has many existing `any`s or areas of unsafe code, it may be difficult to enable this rule._. Those three examples really are unsafe code that's not written with TypeScript best practices. So the lint rule will naturally complain on them.

(not saying it's always easily doable to adhere to best practices - just that the lint rule is written primarily for them, and may need to be disabled)

packages/eslint-plugin/docs/rules/restrict-template-expressions.md
Josh-Cena: The rule ignores tagged templates:

https://github.com/typescript-eslint/typescript-eslint/blob/9bd0c393faadd52f20985f6a5209479d2531a1ca/packages/eslint-plugin/src/rules/restrict-template-expressions.ts#L136-L139

Because they can do whatever they want.

packages/eslint-plugin/docs/rules/no-unsafe-member-access.md
JoshuaKGoldberg: Anyway, thanks! I feel good about where things are now 🙂 but let me know if there's anything else!

bradzacher (approved):

packages/eslint-plugin/docs/rules/restrict-template-expressions.md
Faithfinder: Yeah, ok, I lied to you (because misremembered and didn't double check). What's lighting up for us is [stitches](https://stitches.dev/docs/tokens), not styled components, and they're not maintained anymore. Sorry about the wild goose chase!
image
image

packages/eslint-plugin/docs/rules/no-unsafe-member-access.md
Faithfinder: I can also rephrase my complaint as "These make `any` behave like `unknown` and we already have `unknown` for that" =P

But sounds good to me :)

PR #7460, opened 2023-08-12
fix: add new properties since eslint 8.40.0

### LINKED ISSUE #7453: `context.sourceCode` does not exist in type definition

### Issue Description

First of all, thanks for this awesome project!

(Not sure if I'm using the correct template)

Since eslint 8.40.0, a new property called `sourceCode` was added to `context`, and the old `context.getSourceCode` has been deprecated. However, typescript-eslint doesn't include these types.

N/A

### Repro Steps

1) Just check `context` in any rule

### Versions

| `@typescript-eslint/eslint-plugin`     | `latest(6.x.x)` |
| `@typescript-eslint/parser`            | `latest(6.x.x)` |
| `@typescript-eslint/scope-manager`     | `latest(6.x.x)` |
| `@typescript-eslint/typescript-estree` | `latest(6.x.x)` |
| `@typescript-eslint/type-utils`        | `latest(6.x.x)` |
| `@typescript-eslint/utils`             | `latest(6.x.x)` |
| `TypeScript`                           | `5.1.6` |
| `ESLint`                               | `8.45.0` |
| `node`                                 | `16.18.0` |

so1ve:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7453

## Overview

Add type definition for new eslint context properties

### DIFF
--- a/packages/utils/src/ts-eslint/Rule.ts
+++ b/packages/utils/src/ts-eslint/Rule.ts
@@ -220,19 +220,41 @@ interface RuleContext<
    * Returns the current working directory passed to Linter.
    * It is a path to a directory that should be considered as the current working directory.
    * @since 6.6.0
+   * @deprecated
    */
   getCwd(): string;
 
   /**
+   * The current working directory passed to Linter.
+   * It is a path to a directory that should be considered as the current working directory.
+   * @since 8.40.0
+   */
+  cwd: string;
+
+  /**
    * Returns the filename associated with the source.
+   * @deprecated
    */
   getFilename(): string;
 
   /**
+   * The filename associated with the source.
+   * @since 8.40.0
+   */
+  filename: string;
+
+  /**
    * Returns the full path of the file on disk without any code block information (unlike `getFilename()`).
    * @since 7.28.0
+   * @deprecated
    */
   getPhysicalFilename?(): string;
+
+  /**
+   * The full path of the file on disk without any code block information (unlike `filename`).
+   * @since 8.40.0
+   */
+  physicalFilename?: string;
 
   /**
    * Returns the scope of the currently-traversed node.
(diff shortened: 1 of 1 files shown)


packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -220,20 +220,42 @@ interface RuleContext<
    * Returns the current working directory passed to Linter.
    * It is a path to a directory that should be considered as the current working directory.
    * @since 6.6.0
+   * @deprecated
```
JoshuaKGoldberg, line 223: Come to think of it, it's probably best to explain why deprecated. What to you think about this?

```suggestion
   * @deprecated Use {@link `cwd`} instead.
```

JoshuaKGoldberg (changes requested):
LGTM, thanks! ✨ 

Just the one suggestion on adding a notice to the deprecation, what do you think?

packages/utils/src/ts-eslint/Rule.ts
so1ve: Yeah, looks great!

JoshuaKGoldberg (approved):
💯 thanks!

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -220,20 +220,42 @@ interface RuleContext<
    * Returns the current working directory passed to Linter.
    * It is a path to a directory that should be considered as the current working directory.
    * @since 6.6.0
+   * @deprecated Use {@link `cwd`} instead.
```
bradzacher, line 223: we probably want to remove these `@deprecated` tags.
most plugins will support ESLint v7 and v8 - they're not limited to 8.40.0
this means that trying to direct consumers to use these new APIs is an impossible state as they have to do things like `context.cwd ?? context.getCwd()` - which is just worse code than `context.getCwd()`.

so1ve: You are 100% right - shall I open a new PR?

JoshuaKGoldberg: Hmm, that makes sense.

@so1ve if you want to get another PR to typescript-eslint under your belt 😄 we can merge it any time before Monday midday EST. I'll set a reminder to send one in if one hasn't yet been by then.

PR #7466, opened 2023-08-13
fix(eslint-plugin): [no-unnecessary-condition] false positives with branded types

### LINKED ISSUE #7293: Bug: [no-unnecessary-condition] False positives for branded strings

### Playground Link

https://typescript-eslint.io/play#ts=5.1.6&fileType=.tsx&code=C4TwDgpgBAsiBCAnAhgOwCYXQZWIglqgOZQC8UAznoSQGRQDeUA%2BswEYoYBcUArqgGtUAewDuqKAF8A3AFgAUAoDGw1FSgc0mHNWJkoAIgNRkFWAk7bcBYnPn4AZlAAUmjFms0AlIwVQoAPQBUAAiwpTCALYQwAAWNAB0CpIKCkFQACrxZujCEGYiwFAANhDI6FDA4cglhMAQiFANiMKISfIqakWIEES8xciInnrkRiZmVDZEdo4uPX0DQ7pEPgx%2BgcFhEdFxiclAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK%2BSpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```typescript
type MyBrandedString = string & { __brand: unknown };

const brandedString = "" as MyBrandedString;
if (brandedString) {
  // Do something.
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I did not expect to get a linter error, since `brandedString` in the example can be falsy.

### Actual Result

The rule reports:

```
Unnecessary conditional, value is always truthy.
```

This looks like a false positive to me, because `string & { }` does not disallow empty strings or the like.

### Additional Info

#2506 could be somewhat related, since they also had issues with a similar intersection.

MBuchalik:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7293

## Overview

This PR fixes false positives in rule `no-unnecessary-condition` caused by branded types, as described in #7293. 

Note: #7293 only talks about branded strings, but I noticed that the problem occurred for any type.

> Please include your favorite emoji at the bottom of your pull request to hint to us that you did in fact read this file.

💩

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -28,6 +28,9 @@ const isTruthyLiteral = (type: ts.Type): boolean =>
 const isPossiblyFalsy = (type: ts.Type): boolean =>
   tsutils
     .unionTypeParts(type)
+    // Intersections like `string & {}` can also be possibly falsy,
+    // requiring us to look into the intersection.
+    .flatMap(type => tsutils.intersectionTypeParts(type))
     // PossiblyFalsy flag includes literal values, so exclude ones that
     // are definitely truthy
     .filter(t => !isTruthyLiteral(t))
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -37,7 +37,14 @@ const isPossiblyFalsy = (type: ts.Type): boolean =>
     .some(type => isTypeFlagSet(type, ts.TypeFlags.PossiblyFalsy));
 
 const isPossiblyTruthy = (type: ts.Type): boolean =>
-  tsutils.unionTypeParts(type).some(type => !tsutils.isFalsyType(type));
+  tsutils
+    .unionTypeParts(type)
+    .map(type => tsutils.intersectionTypeParts(type))
+    .some(intersectionParts =>
+      // It is possible to define intersections that are always falsy,
+      // like `"" & { __brand: string }`.
+      intersectionParts.every(type => !tsutils.isFalsyType(type)),
+    );
 
 // Nullish utilities
 const nullishFlag = ts.TypeFlags.Undefined | ts.TypeFlags.Null;
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
@@ -83,6 +83,11 @@ const result2 = foo() == null;
     necessaryConditionTest('null | object'),
     necessaryConditionTest('undefined | true'),
     necessaryConditionTest('void | true'),
+    // "branded" type
+    necessaryConditionTest('string & {}'),
+    necessaryConditionTest('string & { __brand: string }'),
+    necessaryConditionTest('number & {}'),
+    necessaryConditionTest('boolean & {}'),
 
     necessaryConditionTest('any'), // any
     necessaryConditionTest('unknown'), // unknown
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -28,6 +28,9 @@ const isTruthyLiteral = (type: ts.Type): boolean =>
 const isPossiblyFalsy = (type: ts.Type): boolean =>
   tsutils
     .unionTypeParts(type)
+    // Intersections like `string & {}` can also be possibly falsy,
+    // requiring us to look into the intersection.
+    .flatMap(type => tsutils.intersectionTypeParts(type))
```
MBuchalik, line 33: This works because:
- if it is not an intersection, `intersectionTypeParts(type)` just returns the original type, making the function behave like before
- if it is something like `string & {}`, `intersectionTypeParts(type)` returns an array containing the types `string` and `{}`. `string` in this example is possibly falsy, making our function correctly report that the entire thing is possibly falsy
- if we have something like `string & number`, the type checker directly reports this type as `never`, i.e. here we shouldn't even see that this was ever defined as an intersection in the first place

MBuchalik:
I don't know why the type checker is failing. Looks like it is not related to this PR. Type checking succeeds locally on my machine.

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
```diff
@@ -83,6 +83,11 @@ const result2 = foo() == null;
     necessaryConditionTest('null | object'),
     necessaryConditionTest('undefined | true'),
     necessaryConditionTest('void | true'),
+    // "branded" type
+    necessaryConditionTest('string & {}'),
+    necessaryConditionTest('string & { __brand: string }'),
+    necessaryConditionTest('number & {}'),
+    necessaryConditionTest('boolean & {}'),
```
JoshuaKGoldberg, line 90: [Testing] Normally we'll want both `invalid` and `valid` cases added. This just adds `valid`. Could you add some more coverage?

Additionally, some maybe missing areas:

* `bigint`
* ...I don't think symbol brands are a thing, so maybe it's fine to skip them?
* `{} & { __brand: string }`: should this still cause a report?
* `{ a: number } & { b: string }`: should this still cause a report?
* `string & string`

MBuchalik: Thanks for the feedback! 😃

I have added more tests, especially for more complex scenarios that looked like potential edge cases to me.

While adding tests for `invalid` cases, I noticed that `isPossiblyTruthy` also needed to be fixed: If we have a scenario like `"" & { __brand: string }`, the rule would not report any error. Thus, we need to look at the intersection members and make sure that all of them are not falsy.  
(Interesting side note: The rule as currently found on `main` would report `alwaysTruthy`.)

---

Two points regarding your comments:

> I don't think symbol brands are a thing, so maybe it's fine to skip them?

I think so too 👍 

> `{} & { __brand: string }` and `{ a: number } & { b: string }`

I believe that both should behave just like if you wrote `{ __brand: string }`, since the first is equal to `{ __brand: string }`, and the second is equal to `{ a: number, b: string }`.  
Thus, `{ a: number } & { b: string } & string` should not cause a report. But `{ a: number } & { b: string } & "foo"` should cause a report.


---

Btw: I tried to find a scenario where we need to recursively unwrap unions/intersections. Something like `foo & (bar | (baz & frub))`. But I simply couldn't come up with one where the type checker actually reports it in this particular shape. Or, in other words: I couldn't find a scenario where the current implementation would fail. Thus, I decided not to add any recursive logic for now.

JoshuaKGoldberg (approved):
Swell, thanks! ✨

PR #7467, opened 2023-08-13
feat(rule-tester): allow to create empty tests

### LINKED ISSUE #7481: Enhancement: Add the ability to skip valid or invalid tests

### Relevant Package

rule-tester

### Description

I would like to be able to skip valid or invalid scenarios in some tests. In some cases, such examples are simply not needed.

### Fail

```typescript
import noLetRule from '../rule/no-let'

ruleTester.run(
  `disallow to use let`,
  rule,
  {
    valid: [],
    invalid: [`let a = 1`],
  },
)
```

This test is currently crashing with an error: `Error: No test found in suite valid`.

azat-io:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7481

## Overview

In some cases I only need to test for valid or invalid scenarios.

For example, if I want to check that a rule for the ESLint plugin works with different code stylings. Or that a rule does not apply to a certain file extension.

Example:

```ts
ruleTester.run(
  `${RULE_NAME}: works consistently with an empty array or an array with one element`,
  rule,
  {
    valid: ['[].includes(person)', "['Decim'].includes(bartender)"],
    invalid: [],
  },
)
```

This test is currently crashing with an error: `Error: No test found in suite invalid`

### DIFF
--- a/packages/rule-tester/tests/RuleTester.test.ts
+++ b/packages/rule-tester/tests/RuleTester.test.ts
@@ -833,7 +833,18 @@ describe('RuleTester', () => {
           ],
         });
 
-        expect(mockedDescribe.mock.calls).toHaveLength(2);
+        expect(mockedDescribe.mock.calls).toMatchInlineSnapshot(`
+          [
+            [
+              "my-rule",
+              [Function],
+            ],
+            [
+              "invalid",
+              [Function],
+            ],
+          ]
+        `);
       });
 
       it('does not call describe with invalid if no invalid tests are provided', () => {
@@ -848,7 +859,18 @@ describe('RuleTester', () => {
           invalid: [],
         });
 
-        expect(mockedDescribe.mock.calls).toHaveLength(2);
+        expect(mockedDescribe.mock.calls).toMatchInlineSnapshot(`
+          [
+            [
+              "my-rule",
+              [Function],
+            ],
+            [
+              "valid",
+              [Function],
+            ],
+          ]
+        `);
       });
     });
   });
--- a/packages/rule-tester/tests/RuleTester.test.ts
+++ b/packages/rule-tester/tests/RuleTester.test.ts
@@ -819,6 +819,37 @@ describe('RuleTester', () => {
         expect(mockedDescribeSkip.mock.calls).toHaveLength(0);
         // expect(mockedIt.mock.lastCall).toMatchInlineSnapshot(`undefined`);
       });
+
+      it('does not call describe with valid if no valid tests are provided', () => {
+        const ruleTester = new RuleTester();
+
+        ruleTester.run('my-rule', NOOP_RULE, {
+          valid: [],
+          invalid: [
+            {
+              code: 'invalid',
+              errors: [{ messageId: 'error' }],
+            },
+          ],
+        });
+
+        expect(mockedDescribe.mock.calls).toHaveLength(2);
+      });
+
+      it('does not call describe with invalid if no invalid tests are provided', () => {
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (changes requested):
The code changes look nifty but let's discuss in a backing issue. The rule tester is an important part of the public API and I'm not sure we want to do this without talking it over first.

azat-io:
I would like to be able to skip valid or invalid scenarios in some tests. In some cases, such examples are simply not needed.

Some simple example:

```ts
import noLetRule from '../rule/no-let'

ruleTester.run(
  `disallow to use let`,
  rule,
  {
    valid: [],
    invalid: [`let a = 1`],
  },
)
```

azat-io:
Issue #7481

azat-io:
@JoshuaKGoldberg Could you review this PR?

This pull request is the last step to upgrade to version 6 before the big release in our ESLint plugin: https://github.com/azat-io/eslint-plugin-perfectionist/issues/32

JoshuaKGoldberg (changes requested):
Changes look good, thanks for keeping on this!

Could you please add some unit tests for the new logic though? Conditionally skipping tests is a kind of behavior that's easy to trip up over time.

JoshuaKGoldberg:
Yes, sorry for the delay!

azat-io:
@JoshuaKGoldberg Hmm. Can you tell me the best way to do that? Since this problem occurs only when using RuleTester with Vitest. Because of this: https://github.com/vitest-dev/vitest/blob/main/packages/runner/src/run.ts#L333-L337

JoshuaKGoldberg:
See the tests at the bottom of `RuleTester.test.ts` in the `constructor constraints` block for reference: https://github.com/JoshuaKGoldberg/typescript-eslint/blob/6d3d162ce032ebcf5f892a4edfb897797fc96191/packages/rule-tester/tests/RuleTester.test.ts#L762

They show running `ruleTester.run('my-rule", ...)` and then asserting on the `mockedDescribe.mock`.

```ts
it('does not call describe with valid if no valid tests are provided', () => {
  const ruleTester = new RuleTester();

  ruleTester.run('my-rule', NOOP_RULE, {
    valid: [],
    invalid: [
      {
        code: 'invalid',
        errors: [{ messageId: 'error' }],
      },
    ],
  });

  expect(mockedDescribe.mock.calls) // ...
```

azat-io:
@JoshuaKGoldberg Thanks. Done.

packages/rule-tester/tests/RuleTester.test.ts
```diff
@@ -819,6 +819,37 @@ describe('RuleTester', () => {
...
+        ruleTester.run('my-rule', NOOP_RULE, {
+          valid: [
+            {
+              code: 'valid',
+            },
+          ],
+          invalid: [],
+        });
+
+        expect(mockedDescribe.mock.calls).toHaveLength(2);
```
JoshuaKGoldberg, line 851: These two tests have the same assertions. I'd think it good to verify that they're called with the right portions. Maybe `toHaveBeenCalledWith` or `toMatchInlineSnapshot` or something like that. Thoughts?

azat-io: Thanks! Did it with `toMatchInlineSnapshot`.

JoshuaKGoldberg (approved):
Awesome! I actually stumbled onto a want for this recently 😄 so it's very exciting to see it come in. Thanks as always @azat-io!

Issue #6909, opened 2023-04-15
Bug: [no-unsafe-enum-comparison] does not work with bit masks


mrazauskas:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.0.3&sourceType=module&code=KYDwDg9gTgLgBAYwgOwM72MgrgWzgFQE8xgAxAGwEMBzVOAbwCg44BBZQuAXjgEY4APALgAGADTM4AVWQBrZBADuybn0HDeElgGUYUAJbJqq-kLgAmLXAByuAEbAoJ9XADMVgEIQI5YJRU8psIALIwAvoyMoJCwcABmWMgIMPoocAAWlKhEJBQ0ABSoEFhQCMAAXATEZFS0YnAwlFDUwDCVOTU0qACUDJJQrSUqhcWlwHAAZA1NLTC9XAvTza0A3OFAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUFvcgAQBcBPABxQGNoBLEggWhXioDsCB6ZgeztmeQEMAZogbNYAWzrlO4kv2rJOzdFETRonaJHBgAviF1A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
export const enum TypeFlags {
  Any = 1 << 0,
  Unknown = 1 << 1,
  String = 1 << 2,
  Number = 1 << 3,
  Boolean = 1 << 4
}

export function hasTypeFlag(source: TypeFlags, target: TypeFlags) {
  return (source & target) === target;
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
        "@typescript-eslint/no-unsafe-enum-comparison": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

[...]

PR #7479, opened 2023-08-15
feat: support Explicit Resource Management syntax for TS 5.2

sajikix:
## PR Checklist

- [x] Addresses an existing open issue: RC support of #7155

## Overview
Supported the `Explicit Resource Management` syntax supported in TS 5.2 RC.
( Note: I know that @sosukesuzuki  is working on this #7155 Issue. I am his colleague and worked on this PR with his support. )

 I am concerned about the following points and would like to take feedback.
- Should I throw an error in the `typescript-eslint` layer when I parse the wrong syntax?
  - For example, `using {a} = {a:1}` or `using a;`.
- Should I add a snapshot test that uses `using declaration` in `forOfStatement`?
  - I couldn't find any test cases for `let` or `const` either.


Josh-Cena:
> * Should I throw an error in the `typescript-eslint` layer when I parse the wrong syntax?

Yes; this is mostly to support downstream parser consumers like Prettier.

> * Should I add a snapshot test that uses `using declaration` in `forOfStatement`?

Would be good to have one. Snapshots can never be exhaustive.

packages/typescript-estree/src/node-utils.ts
```diff
@@ -335,14 +335,21 @@ export function isJSXToken(node: ts.Node): boolean {
...
-    return 'let';
-  }
-  if (node.flags & ts.NodeFlags.Const) {
-    return 'const';
+): 'const' | 'let' | 'var' | 'using' | 'await using' {
+  switch (node.flags) {
+    case ts.NodeFlags.Using:
+      return 'using';
+    case ts.NodeFlags.AwaitUsing:
+      return 'await using';
```
JoshuaKGoldberg, line 343: ...huh. `ts.NodeFlags` is not a proper exponents-of-two enum:

```ts
enum NodeFlags {
    None = 0,
    Let = 1,
    Const = 2,
    Using = 4,
    AwaitUsing = 6,
    NestedNamespace = 8,
    Synthesized = 16,
    Namespace = 32,
    // ...
}
```

https://github.com/microsoft/TypeScript/pull/54505/files#diff-e9fd483341eea176a38fbd370590e1dc65ce2d9bf70bfd317c5407f04dba9560R796 isn't the first instance of TS adding a combination value to the enum, but it does mess with our logic.

This direct equality checking doesn't work when `node.flags` has additional values on it. For example, putting it inside an `async function` causes an `await using`'s `kind` to be `"const"` and a `using`'s `kind` to be `"var"`.

So missing at least these cases:
* `using` inside a `function`
* `using` inside an `async function`
* `await using` inside an `async function`

packages/ast-spec/src/declaration/VariableDeclaration/spec.ts
```diff
@@ -29,5 +29,5 @@ export interface VariableDeclaration extends BaseNode {
    * var z = 3;
    * ```
    */
-  kind: 'const' | 'let' | 'var';
+  kind: 'await using' | 'const' | 'let' | 'using' | 'var';
```
JoshuaKGoldberg, line 32: In theory, we could adhere more strictly to the [Explicit Resource Management > Declarations > VariableDeclaration spec](https://github.com/estree/estree/blob/0362bbd130e926fed6293f04da57347a8b1e2325/stage3/explicit-resource-management.md#variabledeclaration):

> If `kind` is `"using"` or `"await using"`, for every declarator `d` of `declarations`, `d.id` must be an Identifier. If the variable declaration is the `left` of a ForOfStatement, `d.init` must be `null`, otherwise `d.init` must be an Expression.

I'm not familiar with our practices around specifics like that. Do we normally?

Without any prior context, I would in theory prefer to make our AST more strict & compliant (at the cost of having more types). That way AST types are more accurate. For example, if someone is writng a rule around `using` statements, they probably wouldn't want to have to assert `declarations.id` to be `Identifier` instead of the default.

JoshuaKGoldberg (changes requested):
Thanks for getting started on this, really exciting work! 🚀 

A couple of comments: one bug report on the node flags, and one discussion.

packages/typescript-estree/src/node-utils.ts
sajikix: I appreciate your clarification about the misunderstanding regarding the determination of NodeFlags 🙏 

As you rightly mentioned,  AwaitUsing corresponds to `Const | Using`. To address this, I'll be making the following adjustments:
- Before distinguishing`Const` or `Using`, I will compare `node.flags & ts.NodeFlags.AwaitUsing`(result of `&` operation) with `ts.NodeFlags.AwaitUsing`.
  - This comparison will return true only when the `kind` of the node is `await using`.
- If the comparison mentioned above is not applicable, it indicates that the `kind` of the node is not `await using`. In this case, I will individually evaluate the `Const` and `Using` flags using `&`.

packages/ast-spec/src/declaration/VariableDeclaration/spec.ts
bradzacher: yup I'd definitely agree here!
the spec is much stricter, so we can separate the types to make things easier to refine.
we can also enforce this from the parsing side as well - eg error if the `id` is not an `Identifier`.

bradzacher (changes requested):
Thanks for getting to this so fast!

1) please make the `VariableDeclaration` a disjoint union type to refine the "using" case for the `.id`.
2) please add error tests for the bad cases (eg `using { a } = {}`, `await using [ b ] = [];`, etc)

JoshuaKGoldberg:
Ha, when I prototyped it locally I couldn't think of anything better than `TraditionalVariableDeclaration`. Which I don't like at all. 🤷

packages/typescript-estree/src/node-utils.ts
```diff
@@ -335,13 +335,20 @@ export function isJSXToken(node: ts.Node): boolean {
  */
 export function getDeclarationKind(
   node: ts.VariableDeclarationList,
-): 'const' | 'let' | 'var' {
+): 'const' | 'let' | 'var' | 'using' | 'await using' {
```
JoshuaKGoldberg, line 338: [Question] What do you think about making a dedicated type for this? Just as a convenience point, not a blocker for the PR.

packages/typescript-estree/src/node-utils.ts
```diff
@@ -335,13 +335,20 @@ export function isJSXToken(node: ts.Node): boolean {
...
   if (node.flags & ts.NodeFlags.Let) {
     return 'let';
   }
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
+  if ((node.flags & ts.NodeFlags.AwaitUsing) === ts.NodeFlags.AwaitUsing) {
```
JoshuaKGoldberg, line 343: [Aside] Amusing: https://github.com/typescript-eslint/typescript-eslint/issues/6909. (not requesting changes, just amused)

packages/ast-spec/src/declaration/VariableDeclaration/spec.ts
JoshuaKGoldberg: Looks like the latest commit gets the first bit around `.id`s, but not the ForOfStatement narrowing. Progress!

JoshuaKGoldberg (changes requested):
Great so far! I think ForOfStatements should also be updated for _If the variable declaration is the `left` of a ForOfStatement, `d.init` must be `null`, otherwise `d.init` must be an Expression._, right?

packages/typescript-estree/src/node-utils.ts
sajikix: Looks good!

sajikix:
Thanks again for the review!
I had overlooked that the `left` node type in `ForOfStatements` can be made even stricter...

To make this type stricter, I changed the type of `UsingDeclaration` to two types of union(`UsingInForOfDeclaration` and  `UsingInNomalConextDeclaration`) again. (I can't come up with a good idea for naming these two types either. ....)

Also added AST validation for using in ForStatement.

bradzacher (approved):
this is looking good to me!
thanks heaps for helping us support the next TS version!

PR #7482, opened 2023-08-15
docs(website): add Open in Playground button to code examples

### LINKED ISSUE #7315: Website: Consider aligning to eslint.org 'Open in Website'

### Suggested Changes

https://github.com/eslint/eslint/issues/17018 -> https://github.com/eslint/eslint/pull/17306 added playground links to correct and incorrect code blocks in the ESLint playground. The design the ESLint team asked for is different from ours:

* We (typescript-eslint.io) have a single _Try this rule in the playground ↗_ link on a rules page, floating underneath the config example
* eslint.org instead has an _Open in Playground_ link in the bottom-right of each rule (in)correct code snippet

I very much like how the eslint.org approach lets users directly try out code snippets (thanks @nzakas! ref: https://github.com/eslint/eslint/issues/17018#issuecomment-1490970585). How about we add the same type to our website?

I'm actually leaning towards keeping the existing _Try this rule in the playground ↗_ link -perhaps de-emphasizing it visually- so that folks still get the link above the fold & in a consistent location.

### Affected URL(s)

haocheng6:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7315

## Overview

- This PR updates the `generatedRuleDocs` plugin to add an `eslintrcHash` to the meta string of each correct/incorrect code sample. By default, only code blocks inside a tab will be processed by  the `generatedRuleDocs` plugin in this way. The meta string will be parsed by the `CodeBlockString` component. When `eslintrcHash` is detected, the `CodeBlockString` component will render an "Open in Playground" button at the bottom of the code block.
- This PR also manually updates some docs to add additional commands in the meta string of code blocks.
  - If `showPlaygroundButton` is present, the `generatedRuleDocs` plugin will also add an `eslintrcHash` to the meta string. This is useful for code blocks that are not inside a tab.
  - If `option='VALUE'` is present, its value will be used in the playground link so that users will see the correct option in the editor if they open the playground by clicking the generated button.
- This PR also decreases the font size of the "Try this rule in the playground ↗" link to de-emphasize it visually.

<details><summary>Examples</summary>

### Example 1: the "Open in Playground" button

example

### Example 2: the eslintrc editor has the correct option

eslintrc_example

</details>


armano2:
We should try to align design of 
image

and 

image

---------------

wdyt about printing options in title of code block?

haocheng6:
@armano2 Thanks for the feedback!

I agree that the playground button/link style should be consistent for both the `.eslintrc.cjs` code block and example code blocks. Since we have three options here, which one do you think is the best?

<details><summary>Expand me to see the three options</summary>

1. "Open in Playground" text button

open_in_playground_button

2. Icon button next to the copy button

external_icon_button

3. Existing style for the "Try this rule in the playground" link.

text_link

</details>

I am leaning towards option 1 or 2. My only concern of option 2 is the button is only visible when the code block is hovered on (although we could tweak the CSS to make it always visible) and the text button might be easier to understand than the icon button.

> wdyt about printing options in title of code block?

For some code examples, the option is very long which might not fit in the title line. For example, some code examples in `packages/eslint-plugin/docs/rules/member-ordering.md` are associated with a long option like below.

long_option_text

Also, the option used for most code examples is already obvious in the paragraph above them, like

> The following patterns are considered correct with the default options { ignoredTypeNames: ["RegExp"] }:

So I think printing options in the title of code blocks might look repetitive in some cases. What do you think?

packages/website/src/theme/MDXComponents/TryInPlayground.tsx
```diff
@@ -1,20 +1,32 @@
...
     <Link
-      className={styles.tryInPlaygroundLink}
-      to={`/play#eslintrc=${eslintrcHash}`}
+      className={clsx(styles.tryInPlaygroundLink, className)}
+      to={`/play#${params.toString()}`}
       target="_blank"
     >
-      Try this rule in the playground ↗
+      {children}
     </Link>
```
JoshuaKGoldberg, line 30: [Design] The button looks great in general, but there is a bit of extra vertical spacing that I find a little awkward:

Screenshot of the no-non-null-asserted-chain examples showing more vertical space dedicated to the new button than two existing code lines

I tried these styles and they reduced the vertical height a bit. What do you think?

```css
margin-top: -4rem;
position: relative;
```

Screenshot of the same example but with much less vertical space just used for the button

JoshuaKGoldberg (changes requested):
Snazzy! I really like how you added to Docusaurus generation. Thanks for sending this in! 🎉 

Just a couple of small things, then we should be ready to go. What do you think?

packages/website/src/theme/MDXComponents/TryInPlayground.tsx
haocheng6: I agree for most code blocks, moving up the button a little bit looks better. But I noticed some edge cases that might need special handling:

- In some code blocks, the last line is very long. If we move the button up, then it will make the last one or two lines unreadable.
button_over_text

- For code blocks that only contain one or two lines, the "Open in Playground" may overlap the copy button on the top right corner.
buttons_overlap

My two cents on how to solve these issues:

- If the last one or two lines in a code block is too long, we can reformat the code a little bit, especially if it is a long comment line which can be broken into multiple lines without affecting the semantics of the original code.
- In `website/src/theme/CodeBlock/Content/String.tsx`, we can check if the code snippet only has one or two lines or the last couple of lines are too long. If not, we can add a special class name to the button to move it a little up.
- On mobile screens, I think it might be better to never move the button up due to the limited horizontal space.

What do you think?

packages/website/plugins/generated-rule-docs.ts
```diff
@@ -393,6 +398,59 @@ export const generatedRuleDocs: Plugin = () => {
...
+        continue;
+      }
+
+      if (
+        nodeIsCode(node) &&
+        (insideTab || node.meta?.includes('showPlaygroundButton')) &&
+        !node.meta?.includes('eslintrcHash=')
+      ) {
+        let playgroundEslintrc = eslintrc;
+        const option = node.meta?.match(optionRegex)?.groups!.option;
```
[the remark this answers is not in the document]
haocheng6: Thanks for the suggestion! Updated here: 6e0929245895497ecad3f28ca94902adcdb14ed9

packages/website/src/theme/MDXComponents/TryInPlayground.tsx
JoshuaKGoldberg: Blurgh yeah. I do dislike these edge cases.

> we can reformat the code a little bit

Yeah. Long comment lines feel difficult to read for docs to me anyway, and probably something to avoid. Maybe if we just move them all to a comment above the code, that'll get us most of the way?

```diff
+ // etc etc etc etc etc etc etc
+ let something;
- let something; // etc etc etc etc etc etc etc
```

haocheng6: Updated in 0f7d581e785f6530b8fe7c0f6e2d51557ffb9821 and e2ed8608dad951f793cf7f8ba51a04137536ccfd.

I reformatted one .md file to shorten the last lines of a few code blocks and tweaked the style of the playground button. Now the button will not overlap the copy button and its position will be dynamically adjusted based on the screen size.

Btw, I am not sure why one of the CI check is still in progress for ~2 hours.

JoshuaKGoldberg (dismissed):
This looks great, thanks!

There are a couple of things I think can be fine as followups:

* Applying @armano2's suggestion around unifying the button/tooling design 
* Auto-detecting and/or standardizing the options headings

...but neither are blocking IMO. Let's ship! 🚀 

(sorry for taking so long, this dropped off my radar)

packages/website/src/theme/CodeBlock/Content/styles.module.css
```diff
@@ -6,6 +6,7 @@
   /* rtl:ignore */
   direction: ltr;
   border-radius: inherit;
+  min-height: 5.5rem;
```
JoshuaKGoldberg, line 9: Oops: https://github.com/typescript-eslint/typescript-eslint/issues/8110

cc @haocheng6 as fyi, this was my mistake, I should have caught this in review.

Issue #7534, opened 2023-08-25
Repo: `yarn generate-lib` fails on main


bradzacher:
### Suggestion

I think there's something wrong with how `yarn`3 resolves binaries or something - after running a `yarn install` in the root I then tried running `yarn generate-lib` in the root and it fails with

```
$ yarn generate-lib        

> nx run scope-manager:generate-lib

Usage Error: Couldn't find a script named "tsx".
$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] <scriptName> ...
Warning: run-commands command "yarn tsx tools/generate-lib.ts" exited with non-zero status code

 ———————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 >  NX   Ran target generate-lib for project scope-manager (531ms)
 
    ✖    1/1 failed
    ✔    0/1 succeeded [0 read from cache]
 
   View structured, searchable error logs at https://nx.app/runs/rhV5iHc7zv
```

CDing into the scope manager package:
```
$ cd packages/scope-manager
$ yarn generate-lib
command not found: nx
```

So it looks like `yarn` won't auto read the packages from the parent dir?

Josh-Cena:
@bradzacher: `generate-lib` now runs but the ESLint run at the end of the script fails:

```
ConfigurationNotFoundError: No ESLint configuration found in packages/scope-manager/src/lib.
    at CascadingConfigArrayFactory._finalizeConfigArray (node_modules/@eslint/eslintrc/lib/cascading-config-array-factory.js:521:19)
    at CascadingConfigArrayFactory.getConfigArrayForFile (node_modules/@eslint/eslintrc/lib/cascading-config-array-factory.js:312:21)
    at FileEnumerator._iterateFilesWithFile (node_modules/eslint/lib/cli-engine/file-enumerator.js:368:43)
    at FileEnumerator._iterateFiles (node_modules/eslint/lib/cli-engine/file-enumerator.js:349:25)
    at FileEnumerator.iterateFiles (node_modules/eslint/lib/cli-engine/file-enumerator.js:299:59)
    at iterateFiles.next (<anonymous>)
    at CLIEngine.executeOnFiles (node_modules/eslint/lib/cli-engine/cli-engine.js:797:48)
    at LegacyESLint.lintFiles (node_modules/eslint/lib/eslint/eslint.js:551:23)
    at main (packages/repo-tools/src/generate-lib.mts:305:30) {
[...]

PR #7535, opened 2023-08-25
feat: bump supported TS version to 5.2

### LINKED ISSUE #7155: TypeScript 5.2 Support

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7155

## Overview

- bump the version
- regenerate lib types
- regenerate playground types


packages/scope-manager/tools/generate-lib.ts
```diff
@@ -270,6 +280,17 @@ function main(): void {
...
+  await ESLint.outputFixes(results);
+  console.log('Autofixed lint errors');
 }
 
-main();
+main().catch(e => {
+  console.error(e);
+  // eslint-disable-next-line no-process-exit
+  process.exit(1);
+});
```
JoshuaKGoldberg, line 296: What's going on here? Is it related to #7534?

JoshuaKGoldberg (approved):
🚀

packages/scope-manager/tools/generate-lib.ts
bradzacher: Nah so I ran the generator and noticed that it changed all of the lib files and caused a bunch of lint errors due to unsorted imports and such.

I figured that instead of manually running the linter afterwards was dumb cos we can automate it :)

packages/scope-manager/tools/generate-lib.ts
```diff
@@ -1,5 +1,6 @@
 import type { TSESTree } from '@typescript-eslint/types';
 import { AST_TOKEN_TYPES } from '@typescript-eslint/types';
+import { ESLint } from '@typescript-eslint/utils/ts-eslint';
```
armano2, line 3: @bradzacher this seem to be generating some issues on my machine

```bash
> yarn build

NX   Could not execute command because the task graph has a circular dependency
rule-schema-to-typescript-types:build --> type-utils:build --> utils:build --> scope-manager:build --> utils:build
```

PR #7551, opened 2023-08-28
feat(utils): add ESLint `CodePath` selector types

### LINKED ISSUE #6993: Enhancement: Support CodePath rule listener methods

### Relevant Package

utils

### Description

Rules created using `createRule` currently don't have TypeScript support for the [Code Path Analysis](https://eslint.org/docs/latest/extend/code-path-analysis) events. This would be great to add since they still work fine, its just that TypeScript complains.

I'd like to be able to use `onCodePathStart`, `onCodePathEnd`, `onCodePathSegmentStart`, `onCodePathSegmentEnd`, and `onCodePathSegmentLoop` as well as use an interface for CodePath and CodePathSegment,

### Fail

```typescript
N/A
```

### Pass

```typescript
createRule({
  meta: { ... },
  defaultOptions: [],
  create(context) {
    return {
      /**
       * @param {import('eslint').Rule.CodePath} codePath
       * @param {import('@typescript-eslint/utils').TSESTree.Node} node
       */
      onCodePathStart(codePath, node) {
        const isEffect = ASTUtils.isFunction(node) && isUseEffectCall(node.parent);
        functionInfoStack.push({ codePath, node, isEffect, listeners: [] });
      }
    }
  }
})
```

### Additional Info

_No response_

Zamiell:
## PR Checklist

- [X] Addresses an existing open issue: fixes #6993

## Overview

Adds exported `CodePath` interfaces and un-commented the `onCodePathStart` and `onCodePathEnd` methods, which makes things slightly less painful for people who have custom rules that use this feature.

Based on: https://github.com/typescript-eslint/typescript-eslint/pull/6679


Zamiell:
CI is drunk, the website builds just fine for me locally. Can a maintainer reset the CI cache or something?

Zamiell:
Oh, actually on second thought, it just looks like it is failing a spell check on code that is unrelated to my PR, so a maintainer should fix that.

packages/eslint-plugin/typings/eslint-rules.d.ts
```diff
@@ -817,17 +817,90 @@ declare module 'eslint/lib/rules/no-invalid-this' {
...
+       *
+       * If you want to use this method, you can either ignore those parameters
+       * or specify them with a type assertion, depending on what you want to
+       * do.
+       */
+      onCodePathSegmentLoop?: (
+        fromSegment: TSESLint.CodePathSegment,
+        // toSegment: TSESLint.CodePathSegment,
+        // node: TSESTree.Node,
+      ) => void;
```
bradzacher, line 903: we don't need to add these here.
our extension rule doesn't care about the code path selectors so no need to complicate things

bradzacher (changes requested):
looking good so far!

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -284,6 +284,61 @@ interface RuleContext<
...
+ */
+interface CodePath {
+  id: string;
+  initialSegment: CodePathSegment;
+  finalSegments: CodePathSegment[];
+  returnedSegments: CodePathSegment[];
+  thrownSegments: CodePathSegment[];
+  currentSegments: CodePathSegment[];
+  upper: CodePath | null;
+  childCodePaths: CodePath[];
```
bradzacher, line 304: question - is there documentation on what each property means?
Generally we've tried to add as much jsdoc to these as physically possible

armano2: https://eslint.org/docs/latest/extend/code-path-analysis#codepath

Zamiell: this link is already included in the JSDoc for the interface, so there's nothing else to do, right?

packages/eslint-plugin/typings/eslint-rules.d.ts
Zamiell: disagree - I think it is good for DX, as it adds auto-complete to VSCode and is a valuable indication that these methods actually exist, as well as provide good mouse-over documentation on exactly what to do if you need to use them.

packages/utils/src/ts-eslint/Rule.ts
bradzacher: Sorry I mean annotate each property with the documentation from the web page.
This brings all of the web docs into the user's IDE - which is a big win as it saves someone having to open their browser just to read that the ID means "A unique string. Respective rules can use id to save additional information for each code path."

For example - have a look at the types for the `ESLint` class!
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/utils/src/ts-eslint/ESLint.ts

Zamiell: ok, I added the jsdoc now

packages/eslint-plugin/typings/eslint-rules.d.ts
bradzacher: This particular set of types is ***ONLY*** used internally within our project and it exists to let us expose the selectors that this particular rule has so that we can call them within our extension version of the rule.

But because our extension version does not touch the code path analysis - these aren't used. So uncommenting them does not change or fix anything.

Zamiell: ok, i will delete them then.
can you tell me how to make autocomplete work though for plugin authors?

bradzacher: we can't do autocomplete as that would require adding the signatures to the selectors object.
the types you've added are in the realm of "good enough" to allow people to define their rule in the very rare case that they want the code path analysis.

Zamiell: ok

packages/eslint-plugin/typings/eslint-rules.d.ts
bradzacher, line 1: if you revert the changes to this file entirely we can merge this in.

Zamiell: Since you claim at that auto-complete is impossible, I propose that the greatly improved comments that I authored will act as documentation for plugin authors, and thus should be kept as-is.

When I needed to use this functionality in my own ESLint plugin, the first thing I did was to grep the codebase for "onCodePathStart", which got me started on the right direction, because I could copy paste the commented out code into my own code as selectors.

If that is not convincing to you and you still think that having commented out code here is appropriate, then I think we need to:
1. remove the existing commented out code in the file you want to leave alone, since that code is no longer accurate now due to the inclusion of the real types inside of the `typescript-eslint` codebase
2. officially document how to accurately use these selectors on the website (I'm not really sure where the correct place would be, you can direct me where)

packages/eslint-plugin/typings/eslint-rules.d.ts
```diff
@@ -817,17 +817,60 @@ declare module 'eslint/lib/rules/no-invalid-this' {
...
+      */
+
+      /*
+      onCodePathSegmentLoop?: (
+        fromSegment: TSESLint.CodePathSegment,
+        toSegment: TSESLint.CodePathSegment,
+        node: TSESTree.Node,
+      ) => void;
+      */
```
bradzacher, line 873: This particular file (`packages/eslint-plugin/typings/eslint-rules.d.ts`) is private and internal to our repo - it isn't included in any releases or even really intended for viewing by anyone.

I'm more than happy for these comments to exist! But they should be put in a place that consumers can actually see them. So probably within [this currently empty interface?](https://github.com/IsaacScript/typescript-eslint/blob/interfaces/packages/utils/src/ts-eslint/Rule.ts#L553)

Could you please revert changes to this file and move the comments into the other file?

once that's done we can land this!

bradzacher (changes requested):

Zamiell:
ok, i addressed the feedback and merged with the main branch

bradzacher (approved):

bradzacher:
For future reference - I'd avoid working from a repo in an org account and instead work from a fork on your personal account. We maintainers can't modify your PR at all - even the standard "update branch" button which auto-merges main doesn't work. This can make it difficult to manage your OSS contributions because we have to ping you and wait for your action even for trivial things like merging main - which we sometimes need to do if main was broken (eg in this case the CI was broken!).

Please merge main so we can kick off the CI again with the latest deps then we can merge this.

Zamiell:
> Please merge main 

done!

JoshuaKGoldberg:
...aaaand lint is broken on `main`. Sorry 😄  we can just take this as-is and fix linting separately. #7913

PR #7553, opened 2023-08-28
fix(eslint-plugin): [explicit-function-return-type] support JSX attributes in `allowTypedFunctionExpressions`

### LINKED ISSUE #7552: [explicit-function-return-type] `allowTypedFunctionExpressions` option should be aware of JSX

### Issue Description

The [`explicit-function-return-type` rule](https://typescript-eslint.io/rules/explicit-function-return-type/) has the [`allowTypedFunctionExpressions` option](https://typescript-eslint.io/rules/explicit-function-return-type/#allowtypedfunctionexpressions). This option is enabled as part of the [default option set for the rule](https://typescript-eslint.io/rules/explicit-function-return-type/#options).

This option is meant to allow omitting a return type when it will be enforced during assignment against a LHS type assertion, similar to how omitted argument types will be handled by TypeScript itself. This option should take into account assignment to JSX attributes but does not. Such assignment will be fully enforced by TypeScript JSX handling and is semantically the same as the object-attribute-assignment logic that is also allowed under this option.

image

As seen in this screenshot [from the linked repo](https://github.com/merrywhether/return-type/blob/main/src/pages/index.tsx), examples 2 and 3 are functionally the same but treated differently by the rule. Example 3 should not be a lint error.

[Demo PR](https://github.com/merrywhether/typescript-eslint/pull/2) with a potential version of this change.

https://github.com/merrywhether/return-type

### Repro Steps

1) clone the repo
2) `pnpm install`
3) `pnpm lint`

(other package managers will work as well)

```sh
% pn lint

> return-type@0.1.0 lint .../return-type
> next lint


./src/pages/index.tsx
15:5  Error: Missing return type on function.  @typescript-eslint/explicit-function-return-type
33:22  Error: Missing return type on function.  @typescript-eslint/explicit-function-return-type
```

(only the first error should be valid)

### Versions

| `@typescript-eslint/eslint-plugin`     | `6.5.0` |
| `@typescript-eslint/parser`            | `6.5.0` |
| `TypeScript`                           | `5.2.2` |
| `ESLint`                               | `8.48.0` |
| `node`                                 | `20.3.1` |

merrywhether:
## PR Checklist

- [X] Addresses an existing open issue: fixes #7552

## Overview

Adds support for JSX attributes to the `allowTypedFunctionExpressions` option of the `explicit-function-return-type` rule by augmenting the `isTypedFunctionExpression()` function to include more parent types. Also DRYes out code between common and recursive object-property cases.


packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
```diff
@@ -37,6 +37,51 @@ function isPropertyDefinitionWithTypeAnnotation(
   );
 }
 
+/**
+ * Checks if a node belongs to:
+ * ```
+ * foo(() => 1)
+ * ```
+ */
+function isFunctionArgument(
```
merrywhether, line 46: No changes, just moved from below to be colocated with siblings.

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
```diff
@@ -37,6 +37,51 @@ function isPropertyDefinitionWithTypeAnnotation(
...
+  );
+}
+
+/**
+ * Checks if a node is an expression in a JSX attribute assignment
+ * ```
+ * <Comp x={...} />
+ * ```
+ */
+function isJSXAttribute(
```
merrywhether, line 63: Open to feedback on the name here.

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
```diff
@@ -37,6 +37,51 @@ function isPropertyDefinitionWithTypeAnnotation(
...
+  node: TSESTree.Node,
+): node is TSESTree.JSXExpressionContainer | TSESTree.JSXSpreadAttribute {
```
merrywhether, line 65: It seems like this (and the others) could just be boolean (especially the ones that are asserting more than just a type).

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
```diff
@@ -37,6 +37,51 @@ function isPropertyDefinitionWithTypeAnnotation(
...
+  return (
+    node.type === AST_NODE_TYPES.JSXExpressionContainer ||
```
merrywhether, line 67: This is technically the relevant RHS type of `JSXAttribute`, so this check could also recurse one level higher if that is preferred. This should be the only way to have a function supplied in JSX though.

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
```diff
@@ -37,6 +37,51 @@ function isPropertyDefinitionWithTypeAnnotation(
...
+    node.type === AST_NODE_TYPES.JSXSpreadAttribute
+  );
+}
+
+function isTypedParent(
```
merrywhether, line 72: Created a wrapping function for these to make it deduplicate between the common case and recursive object-property case. Open to feedback on the name here.

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
```diff
@@ -194,11 +216,8 @@ function isTypedFunctionExpression(
   }
 
   return (
-    isTypeAssertion(parent) ||
-    isVariableDeclaratorWithTypeAnnotation(parent) ||
-    isPropertyDefinitionWithTypeAnnotation(parent) ||
+    isTypedParent(parent, node) ||
     isPropertyOfObjectWithType(parent) ||
-    isFunctionArgument(parent, node) ||
     isConstructorArgument(parent)
```
merrywhether, line 221: DRYing these out, it makes it more obvious that the recursive object-property doesn't include this `isConstructorArgument()` case, but it seems like it should?

Happy to move it into `isTypedParent()` if this is indeed a bug.

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
bradzacher: I think we'll want to check the parent because this is another valid location for a `JSXExpressionContainer`:

```ts
<>{() => {}}</>
```

packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts
```diff
@@ -180,6 +180,46 @@ class App {
...
+  return <button {...{ onClick: () => {} }} />;
+};
+      `,
+      options: [{ allowTypedFunctionExpressions: true }],
+      parserOptions: {
+        ecmaFeatures: {
+          jsx: true,
+        },
+      },
+    },
```
bradzacher, line 222: for completeness can we also get this as an `invalid` test?

bradzacher (changes requested):
this is looking good so far!
good work adding all the test cases!

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
merrywhether: @bradzacher Ah, I totally forgot about the render-props pattern. Running this down actually led me across an existing issue in TS where fragment shorthands are not typed properly: https://github.com/microsoft/TypeScript/issues/50429

I think all `JSXExpressionContainer` instances should be fully bound by types if it weren't for that issue ([reference CodeSandbox with render props](https://codesandbox.io/s/wispy-frog-sywpdm?file=/src/App.tsx)):

1. Attribute assignment: original case; type can be reliably inferred from assignment
1. Render props of built-ins and `<Fragment />`: not allowed due to limited child types (`JSX.ElementType` or `JSX.Element | null`); all functions will be type errors here
1. Render props of `<>` fragment shorthand: all functions _should_ be type errors here 
1. Render props of components with custom function children: properly typed-constrained by explicit `children` prop, the same as if children was defined via attribute

I went around on this for a while, but eventually realized that while something like this is unsafe:
```tsx
<>{() => 'bug'}</>
```
Having a lint error that turns it into this this doesn't make things better as it's still not a type error:
```tsx
<>{(): string => 'bug'}</>
```
While this is a currently an issue with TypeScript, it doesn't seem like the lint rule can do anything better here in place of proper type-checking.

[Reference TS-ESLint playground](https://typescript-eslint.io/play/#ts=5.2.1-rc&showAST=es&fileType=.tsx&code=MYewdgzgLgBAHjAvDAFAKBjAPAPg57HAbxQEokdUiYBDCAEwDMAuGKAJwFcBTGAX1J8sAejwFsAMXY0A5gFtuYKMTIUqtBizZdeAocKmyFSsQSwBhEHIAOK8okopqdJqw49%2BgkZZumReUgBuIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts
merrywhether: Added negated counterparts of all new tests

merrywhether:
@bradzacher Updated this, LMK what you think

merrywhether:
@bradzacher Anything I can do to help land this?

bradzacher (approved):
this looks good to me - thanks for this!

PR #7564, opened 2023-08-30
fix(eslint-plugin): [consistent-type-imports] do not add new semicolons

AlessioGr:
## PR Checklist

- [ ] Addresses an existing open issue: I have not opened an issue for this

## Overview

Previously, the consistent-type-imports rule would add new semicolons at the end of some import even though the original code did not end with a semicolon.

This conflicted with the "semi: false" option in prettier, which I used in my project.

In this PR, I have set it to only add this semicolon if the original code has a semicolon as well. This resolves the conflict without adding a new property to this eslint rule.

### DIFF
--- a/packages/eslint-plugin/src/rules/consistent-type-imports.ts
+++ b/packages/eslint-plugin/src/rules/consistent-type-imports.ts
@@ -680,7 +680,10 @@ export default util.createRule<Options, MessageIds>({
             afterFixes.push(insertTypeNamedSpecifiers);
           }
         } else {
-          // The import is both default and named.  Insert named on new line because can't mix default type import and named type imports
+          const sourceEndsWithSemicolon =
+            sourceCode.getTokenAfter(node.source)?.value === ';';
+
+          // The import is both default and named. Insert named on new line because can't mix default type import and named type imports
           if (fixStyle === 'inline-type-imports') {
             yield fixer.insertTextBefore(
               node,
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
We generally don't try to put effort into rule fixers towards respecting nuances of original source file formatting. It's not the responsibility of the lint rules. If the user cares deeply about formatting, they should use a dedicated formatter.

> Addresses an existing open issue: I have not opened an issue for this

Indeed, this is why we ask for issues to be opened first. That way you don't go through the effort of making a PR for something we wouldn't take in internally.

Josh-Cena:
Also note that this is not how ESLint works. See https://eslint.org/docs/latest/extend/custom-rules#applying-fixes:

> Since all rules are run again after the initial round of fixes is applied, it’s not necessary for a rule to check whether the code style of a fix will cause errors to be reported by another rule.

You should generally run Prettier as the last step of your linting fixes.

packages/eslint-plugin/src/rules/consistent-type-imports.ts
```diff
@@ -680,7 +680,10 @@ export default util.createRule<Options, MessageIds>({
             afterFixes.push(insertTypeNamedSpecifiers);
           }
         } else {
-          // The import is both default and named.  Insert named on new line because can't mix default type import and named type imports
+          const sourceEndsWithSemicolon =
+            sourceCode.getTokenAfter(node.source)?.value === ';';
```
JoshuaKGoldberg, line 684: FWIW the code changes themselves are pretty well crafted 😄 nicely done!

PR #7616, opened 2023-09-07
fix(eslint-plugin): check possibly undefined ancestor.parent

MichaelBelousov:
## PR Checklist

## Overview

Check if `ancestor.parent` is defined before passing it to typescript which will otherwise crash with:
`TypeError: Cannot read properties of undefined (reading 'kind')` on an undefined value.

### DIFF
--- a/packages/eslint-plugin/src/rules/return-await.ts
+++ b/packages/eslint-plugin/src/rules/return-await.ts
@@ -95,6 +95,7 @@ export default util.createRule({
 
       while (ancestor && !ts.isFunctionLike(ancestor)) {
         if (
+          ancestor.parent &&
           ts.isTryStatement(ancestor.parent) &&
           ts.isBlock(ancestor) &&
           ancestor.parent.end === ancestor.end


bradzacher:
Please file an issue with a repro case before raising a PR.

MichaelBelousov:
> Please file an issue with a repro case before raising a PR, as per our contributing guidelines

Yes sorry, thought it was ok to not do that while this is still a draft, was working on that, but admittedly having trouble reproducing, the failure seems to occur in my repro because the parser thinks there's a parenthetical group when there isn't... will reopen with a repro

MichaelBelousov:
I cannot reproduce in the playground in spite of locally using the latest packages. I will just use a local patch and let this rot until someone else encounters this.

bradzacher:
Try reproducing it in an isolated repro repo?
That's our recommendation usually as you can setup whatever configs, dependencies, and state you might need.

MichaelBelousov:
> Try reproducing it in an isolated repro repo? That's our recommendation usually as you can setup whatever configs, dependencies, and state you might need.

I'll try extracting it out of proprietary code, it's just a lot of work and I'm also in a bit of a crunch, so it won't be timely, which I'm sure is fine

PR #7645, opened 2023-09-14
chore: add discord and stackoverflow as links to issue template

bradzacher:
IDK if this is the best as it adds more noise
But people can't easily find these options right now without jumping through some hoops.


JoshuaKGoldberg (approved):
We'll see how this goes 😄

JoshuaKGoldberg:
(once lint passes)

Josh-Cena:
Another possibility: add a "community" page to the website, and only add a link to that page. The benefit is we won't be constantly updating the issue wizard.

JoshuaKGoldberg:
> Another possibility: add a "community" page to the website, and only add a link to that page. The benefit is we won't be constantly updating the issue wizard.

Agreed! Filed #7736.

Issue #7628, opened 2023-09-11
Bug: [prefer-nullish-coalescing] Default option will change runtime behavior


ide:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.2.2&showAST=ts&fileType=.tsx&code=MYewdgzgLgBAZiEAuG0BOBLMBzGAfGAVzABMBTOLMkmAXhgHIGYBDCVKTHfI0iqkgG4AUKEggANmQB0EkNgAUCEPgIMWHLtgYBKEUA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6MAG1IvaOOiQANJnDYcleU02IAClQC2lfJQBuKJVgPZIyfFSaKMXackT6BgC%2BgWBBALr6IUFAA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```typescript
const foo: string | undefined = '' as string | undefined;
console.log(foo || 'a string');
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/prefer-nullish-coalescing": ["error", {
      "ignorePrimitives": {
        "string": false
      }
    }]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expected no lint errors. The [docs](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md#ignoreprimitives) say this is correct code with `ignorePrimitives: { string: false }`:

```ts
const foo: string | undefined = 'bar';
foo || 'a string';
```

[...]

PR #7650, opened 2023-09-15
docs(eslint-plugin): fix ignorePrimitives examples in prefer-nullish-coalescing docs

ide:
## PR Checklist

- [x] Addresses an existing open issue: brought up in (but doesn't close) #7628

## Overview

The examples for the `ignorePrimitives` option were flipped. Setting `ignorePrimitives` to true means that any logical OR expression with a primitive operand is ignored by the lint rule.

I also made two other edits to this docs page. The first is to change "you can safely replace" to "you may consider replacing" in the lint rule's description. The reason for this change is that this lint rule suggests changes that alter the code's behavior in subtle cases where `x || y !== x ?? y`. My interpretation of "safe" for lint suggestions is they don't change the behavior of the code, please let me know if this expectation isn't correct!

There is also a small change to reword the tense in one of the sentences to help it read more smoothly.


JoshuaKGoldberg (approved):
💯 love it, thanks!

PR #7666, opened 2023-09-17
docs: better rules table (filter rules by extension, etc.)

### LINKED ISSUE #7329: Docs: Formatting / Deprecated Rules are not clearly delineated in the rule list

### Suggested Changes

The [rules reference](https://eslint.org/docs/latest/rules/) lists all of the rules. It's the first place that someone goes to when they want to start looking through all of the rules that `typescript-eslint` offers.

Right now, they are separated into 2 categories: Supported Rules & Extension Rules.

However, this layout does not capture which rules are formatting rules (which are greatly discouraged from being used) and it does not capture which rules are deprecated (which obviously shouldn't be used). In other words, the "bad" rules are lumped in together with everything else, which makes the experience painful as an end-user who is going through the list to find out which ones they want to enable for their codebase.

On the left sidebar, there is a "Formatting Rules" and "Deprecated Rules" section, but this is pretty obscure and hard to miss, especially because they are all the way at the bottom behind a huge list of rules. Furthermore, the fact that these sidebars exist isn't represented on the main rules page. So I motion that these sidebar categories are completely deleted and the categorization is merged into the main rules page in some way.

First, I'd mention that we might want to take inspiration from [the ESLint rules page](https://eslint.org/docs/latest/rules/), which delineates rules into four categories:
- Possible Problems
- Suggestions
- Layout & Formatting
- Deprecated

(I'm not counting the "Removed" category here.)

So, we could fix this problem in a few different ways. I think one way is to replicate the ESLint rule layout, such that the ordering of rules on the page would look something like this:

- Possible Problems
  - Possible Problems (Main Rules)
  - Possible Problems (ESLint Extension Rules)
- Suggestions
  - Suggestions (Main Rules)
  - Suggestions (ESLint Extension Rules)
- Layout & Formatting
  - Layout & Formatting (Main Rules)
  - Layout & Formatting (ESLint Extension Rules)
- Deprecated
  - Deprecated (Main Rules)
  - Deprecated (ESLint Extension Rules)

At the end of the day, the problem we are looking to solve is that users can go through the menu of rules without tripping over "bad" rules. Or, they can easily see from a glance which rules are formatting rules when browsing the master rule list.

### Affected URL(s)

https://eslint.org/docs/latest/rules/

Zamiell:
## PR Checklist

- [X] Addresses an existing open issue: fixes #7329

## Overview

- This PR changes the docs in the way that Brad discusses in the linked issue.
- I got rid of all of the existing sidebar categories, since they do not make sense in the context of having a sortable table.
- I also created a "constants.ts" file to remove some existing DRY violations that I found in the repository.


packages/website/src/components/RulesTable/index.tsx
```diff
@@ -191,29 +217,69 @@ export default function RulesTable({
...
+          />
+          <RuleFilterCheckBox
+            mode={filters.formatting}
+            setMode={(newMode): void => changeFilter('formatting', newMode)}
+            label={`${FORMATTING_RULE_EMOJI} formatting rule`}
+          />
+          <RuleFilterCheckBox
+            mode={filters.deprecated}
+            setMode={(newMode): void => changeFilter('deprecated', newMode)}
+            label={`${DEPRECATED_RULE_EMOJI} deprecated rule`}
```
JoshuaKGoldberg, line 247: Screenshot of the rule filters wrapping the last one, deprecated, to the next line

It's going to bug me that on my monitor, these wrap to a next line... maybe one of the titles can be shrunk by a word?

packages/website/src/components/RulesTable/index.tsx
```diff
@@ -191,29 +217,69 @@ export default function RulesTable({
...
         <thead>
           <tr>
             <th className={styles.ruleCol}>Rule</th>
-            <th className={styles.attrCol}>Config</th>
-            <th className={styles.attrCol}>Fixer</th>
-            <th className={styles.attrCol}>Typed</th>
+            <th className={styles.attrCol}>
+              <button title="The config group that the rule belongs to, if any.">
+                C
+              </button>
```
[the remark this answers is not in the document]
Zamiell: First, I needed to shorten the columns to a single letter to minimize the width of all the columns. (Otherwise, there's very little room for the actual rule name + rule description in the left most column.)

However, once you do that, the columns names are no longer easily identifiable from just their first letter alone, so I needed to add a text-on-hover functionality, and using a button with `title` text seemed to be the easiest way to do that. For example, if I just use normal text with a `title` element, then its non-obvious that it actually has mouseover functionality (since it just looks like normal text).

I'm open to other ideas if you want to workshop this further

packages/website/src/components/RulesTable/index.tsx
Zamiell: that seems arbitrary, but i'll get rid of "rule" in the three places that it appears

packages/website/src/components/RulesTable/index.tsx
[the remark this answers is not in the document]
JoshuaKGoldberg: Makes sense how you got there. The problem with a button is that it's an accessibility issue: buttons are meant to be interactive. A non-interactive button is confusing. Both for mouse users and non-mouse (e.g. keyboard) users, but especially for the latter.

I think using the emojis as a key would help. Will comment in the other thread.

packages/website/src/components/RulesTable/index.tsx
```diff
@@ -191,29 +217,69 @@ export default function RulesTable({
...
+            </th>
+            <th className={styles.attrCol}>
+              <button title="Whether the rule has an auto-fixer and/or has suggestions.">
+                F
```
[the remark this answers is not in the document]
JoshuaKGoldberg: [`white-space: nowrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space)?

Screenshot of the first two rules table rows with non-collapsed attribute columns with emojis

packages/eslint-plugin/docs/rules/README.md
```diff
@@ -6,18 +6,27 @@ pagination_prev: null
...
+- `💭 requires type information` refers to whether the rule needs to leverage the power of the TypeScript compiler in order to work properly. Rules that require type information are usually much slower than rules that don't. Thus, if linting performance becomes an issue in a gigantic codebase, a good first step might be make rules that require type information run only in CI.
+- `🧱 extension rule` means that the rule was originally a [core ESLint rule](https://eslint.org/docs/latest/rules/). Some core ESLint rules do not support TypeScript syntax: either they crash, ignore the syntax, or falsely report against it. In these cases, we create an "extension rule": a rule within our plugin that has the same functionality, but also supports TypeScript.
+- `📐 formatting rule` means that the rule has to do with formatting. Formatting rules are mostly here for legacy purposes, because we [strongly recommend against using ESLint for formatting](/linting/troubleshooting/formatting).
+- `💀 deprecated rule` means that the rule should no longer be used and will be removed from the plugin in a future version.
```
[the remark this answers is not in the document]
Zamiell: I think the emoji + category name are pretty descriptive enough such that a one sentence description doesn't add much value.
instead, to solve the problem you are describing, i will simply move the sorting documentation to the bottom of the page with a link next to the sort UI elements.

packages/website/src/components/RulesTable/index.tsx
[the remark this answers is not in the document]
Zamiell: in that screenshot, it doesn't look like `nowrap` is doing anything.
in other words, the size of the columns 2-6 still look to be doubled in size (compared to just having a single letter).
thus, I think that we are wasting a lot of space here.

packages/website/src/components/RulesTable/index.tsx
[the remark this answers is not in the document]
Zamiell: ok, i can replace the single letter with an emoji, but then I still think it runs into the same problem.
in other words, an emoji is just "normal text", so how do we indicate to the end-user that this particular emoji has some special mouseover text functionality?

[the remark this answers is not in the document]
Zamiell: > I think you can add title to other elements too

Yes, as I said above, I can add `title` to e.g. `p` or `div` element, but then there is no visual indication that mouseover functionality exists, so we need a solution to that problem. (changing the cursor when mousing over doesn't solve the problem because you have to mouse over it first and you would have no reason to do that in the first place)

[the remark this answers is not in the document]
Zamiell: > I think it would make sense to remove package name from the rule names.

I kind of like the full name for copy-pasting purposes from the docs into a e.g. eslint-disable comment, and then also for the inverse e.g. ctrl+f for a rule that was triggered in my codebase without manually having to mutate the text.

what does josh think about this?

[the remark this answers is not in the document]
JoshuaKGoldberg: which Josh?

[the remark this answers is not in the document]
JoshuaKGoldberg: Just kidding - I'm still ruminating. It's a surprisingly hard UX issue...!

packages/eslint-plugin/docs/rules/README.md
```diff
@@ -6,18 +6,27 @@ pagination_prev: null
 slug: /
 ---
 
-`@typescript-eslint/eslint-plugin` includes over 100 rules that detect best practice violations, bugs, and/or stylistic issues specifically for TypeScript code.
-See [Configs](/linting/configs) for how to enable recommended rules using configs.
+## Introduction
```
Josh-Cena, line 9: H2s called "Introduction" are not customary in our docs. I suggest removing it, since the first paragraph is introduction by default.

packages/eslint-plugin/docs/rules/README.md
```diff
@@ -6,18 +6,27 @@ pagination_prev: null
...
-In these cases, we create what we call an extension rule; a rule within our plugin that has the same functionality, but also supports TypeScript.
+## Sorting
 
-<RulesTable ruleset="extension-rules" />
+- "Config Group" refers to the configuration preset. We offer [three different config groups](/linting/configs) that allow users to enable a large set of recommended rules all at once.
+- `🔧 fixable` refers to whether the rule contains an ESLint auto-fixer. If the rule has an auto-fixer, then some rule violations can be fixed by running `eslint` with the `--fix` flag. This will automatically change the code, which can save a lot of time! (It is a common pattern for developers to configure their IDE to automatically run `eslint --fix` when saving a TypeScript file.)
+- `💡 has suggestions` refers to whether the rule will offer a suggestion of how to fix the code. Sometimes, it is not safe to automatically fix the code with an auto-fixer. But in these cases, we often have a good guess of what the correct fix should be, and we can provide it as a suggestion to the developer.
```
Josh-Cena, line 28: These concepts exist in ESLint. I'm surprised there doesn't seem to be a detailed description of this in ESLint docs.

packages/eslint-plugin/docs/rules/README.md
```diff
@@ -6,18 +6,27 @@ pagination_prev: null
...
+- `💭 requires type information` refers to whether the rule needs to leverage the power of the TypeScript compiler in order to work properly. Rules that require type information are usually much slower than rules that don't. Thus, if linting performance becomes an issue in a gigantic codebase, a good first step might be make rules that require type information run only in CI.
+- `🧱 extension rule` means that the rule was originally a [core ESLint rule](https://eslint.org/docs/latest/rules/). Some core ESLint rules do not support TypeScript syntax: either they crash, ignore the syntax, or falsely report against it. In these cases, we create an "extension rule": a rule within our plugin that has the same functionality, but also supports TypeScript.
```
Josh-Cena, line 30: Sometimes we do more than that: e.g. we use type information to refine the check. I think the only thing that's for sure is:

1. The rule either supports TypeScript syntax or leverages type informatoin
2. There exists a base rule that does similar stuff but in a less powerful way

Zamiell: i didn't author this passage, it was taken from the existing page. if there are problems, do you want to suggest a commit?

Josh-Cena: Ah, didn't realize that. Let's keep it as-is for now, then.

packages/eslint-plugin/docs/rules/README.md
```diff
@@ -6,18 +6,25 @@ pagination_prev: null
 slug: /
 ---
 
-`@typescript-eslint/eslint-plugin` includes over 100 rules that detect best practice violations, bugs, and/or stylistic issues specifically for TypeScript code.
-See [Configs](/linting/configs) for how to enable recommended rules using configs.
+`@typescript-eslint/eslint-plugin` includes over 100 rules that detect best practice violations, bugs, and/or stylistic issues specifically for TypeScript code. All of our rules are listed below.
 
-## Supported Rules
+Instead of enabling rules one by one, we recommend using one of [our pre-defined configs](/linting/configs) to easily enable a large set of
+recommended rules.
```
Josh-Cena, line 12:
```suggestion
Instead of enabling rules one by one, we recommend using one of [our pre-defined configs](/linting/configs) to easily enable a large set of recommended rules.
```

packages/website/src/components/RulesTable/index.tsx
```diff
@@ -68,14 +80,32 @@ function RuleRow({
             : undefined
         }
       >
-        {fixable ? '🔧\n' : '\n'}
-        {hasSuggestions ? '💡' : ''}
+        {fixable ? `${FIXABLE_EMOJI}\n` : '\n'}
+        {hasSuggestions ? SUGGESTIONS_EMOJI : ''}
```
Josh-Cena, line 84:
```suggestion
        {fixable ? FIXABLE_EMOJI : ''}<br />
        {hasSuggestions ? SUGGESTIONS_EMOJI : ''}
```

I believe it should do the same thing

packages/eslint-plugin/docs/rules/README.md
Josh-Cena: Actually, where is it copied from? I don't find existing content that contains the same text.

Josh-Cena (approved):
Looks great to me, but will wait on @JoshuaKGoldberg

packages/eslint-plugin/docs/rules/README.md
```diff
@@ -6,18 +6,25 @@ pagination_prev: null
...
 
-<RulesTable ruleset="supported-rules" />
+import RulesTable from "@site/src/components/RulesTable";
 
-## Extension Rules
+<RulesTable />
 
-In some cases, ESLint provides a rule itself, but it doesn't support TypeScript syntax; either it crashes, or it ignores the syntax, or it falsely reports against it.
-In these cases, we create what we call an extension rule; a rule within our plugin that has the same functionality, but also supports TypeScript.
+## Sorting
```
Josh-Cena, line 22: This is not "sorting"—it's "filtering".

packages/eslint-plugin/docs/rules/README.md
Zamiell: The main branch of this repo contains this text:

image

packages/website/src/components/RulesTable/index.tsx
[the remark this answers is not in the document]
Zamiell: > Is it possible to hide the plugin name if screen is narrow?

i vote no to keep it simple

packages/website/src/components/RulesTable/index.tsx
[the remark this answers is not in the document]
Zamiell: should i resolve this?

packages/website/src/components/RulesTable/index.tsx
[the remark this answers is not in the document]
Zamiell: the current version of the PR has emojis, but the outstanding issue is that the emojis need to have some signifier that they have mouseover functionality (or the table could be reworked in some other way).

Josh-Cena (dismissed):

packages/eslint-plugin/docs/rules/README.md
```diff
@@ -6,18 +6,24 @@ pagination_prev: null
...
+- `💭 requires type information` refers to whether the rule needs to leverage the power of the TypeScript compiler in order to work properly. Rules that require type information are usually much slower than rules that don't. Thus, if linting performance becomes an issue in a gigantic codebase, a good first step might be make rules that require type information run only in CI.
+- `🧱 extension rule` means that the rule was originally a [core ESLint rule](https://eslint.org/docs/latest/rules/). Some core ESLint rules do not support TypeScript syntax: either they crash, ignore the syntax, or falsely report against it. In these cases, we create an "extension rule": a rule within our plugin that has the same functionality, but also supports TypeScript.
+- `📐 formatting rule` means that the rule has to do with formatting. Formatting rules are mostly here for legacy purposes, because we [strongly recommend against using ESLint for formatting](/linting/troubleshooting/formatting).
+- `💀 deprecated rule` means that the rule should no longer be used and will be removed from the plugin in a future version.
```
JoshuaKGoldberg, line 29: [Content] Moving this all down to the bottom of the page does help make the page easier to skim. But it doesn't address the issue that this text is quite a lot to read. Which I think is a symptom of a deeper issue: that the text is duplicating docs that should really exist elsewhere (and mostly do). We don't need to re-explain, say, what fixers are, or what typed linting is. There are other docs for that - and we want people to all read the same docs.

What do you think about splitting this into smaller list items, each with links to relevant docs?

```md
### Config Group (⚙️)

"Config Group" refers to the [pre-defined config](/linting/configs) that includes the rule.
Extending from a configuration preset allow for enabling a large set of recommended rules all at once.

### Metadata

- `🔧 fixable` refers to whether the rule contains an [ESLint `--fix` auto-fixer](https://eslint.org/docs/latest/use/command-line-interface#--fix).
- `💡 has suggestions` refers to whether the rule contains an ESLint suggestion fixer.
  - Sometimes, it is not safe to automatically fix the code with an auto-fixer. But in these cases, we often have a good guess of what the correct fix should be, and we can provide it as a suggestion to the developer.
- `💭 requires type information` refers to whether the rule requires [typed linting](/linting/typed-linting).
- `🧱 extension rule` means that the rule was originally a [core ESLint rule](https://eslint.org/docs/latest/rules/).
  - Some core ESLint rules do not support TypeScript syntax: either they crash, ignore the syntax, or falsely report against it.
  - In these cases, we create an "extension rule": a rule within our plugin that has the same functionality, but also supports TypeScript.
- `📐 formatting rule` means that the rule has to do with formatting.
  - We [strongly recommend against using ESLint for formatting](/linting/troubleshooting/formatting).
  - Soon, formatting rules will be moved to the [ESLint stylistic plugin](https://eslint.style).
- `💀 deprecated rule` means that the rule should no longer be used and will be removed from the plugin in a future version.
```

The only points here that aren't covered elsewhere are:

* Suggestion fixers: which should really be documented on eslint.org. I'll file an issue.
* Extension rules: good info, left in - but split into sub-items for readability
    * Now I wonder: should we have more dedicated docs about them? 🤔 
* Formatting rules: eslint.style has become the right place to point folks to 🚀 

I don't want to be a stick in the mud and keep dragging on this PR. I'd be up for merging if you two are happy with how things are now. But I do think we can trim this down a bit. Thoughts?

cc @Josh-Cena

JoshuaKGoldberg (changes requested):
Looking great! I'm mostly happy - just suggesting some streamlining to the docs. WDYT? ✨

Zamiell:
> Looking great! I'm mostly happy - just suggesting some streamlining to the docs. WDYT?

All the feedback should be committed now.

JoshuaKGoldberg (dismissed):
🚀 thanks, this is looking great!

packages/eslint-plugin/docs/rules/README.md
```diff
@@ -6,18 +6,36 @@ pagination_prev: null
...
 import RulesTable from "@site/src/components/RulesTable";
 
-<RulesTable ruleset="supported-rules" />
+<RulesTable />
+
+## Filtering
+
+### Config Group (⚙️)
 
-## Extension Rules
```
JoshuaKGoldberg, line 18: I just realized https://typescript-eslint.io/rules/#extension-rules is probably the only canonical link to an explanation of extension rules out there. So we probably don't want to get rid of the heading. But I've pestered you so much on this PR 😅 I'll just add it in, and if you @Josh-Cena / @Zamiell disagree we can always revert before the release on Monday.

Zamiell: do we really need a separate heading though? it is fairly self-explanatory what an extension rule is.
like, for example, consider `class-methods-use-this`.
if you go to the page for it, it says:
> This rule extends the base [eslint/class-methods-use-this](https://eslint.org/docs/rules/class-methods-use-this) rule. It adds support for ignoring override methods or methods on classes that implement an interface.
After reading this, i am not left wondering, "I have more questions about what an extension rule is, I wonder if there is a more formal definition out there somewhere?"

JoshuaKGoldberg: I think folks need somewhere to hard-link to. It's similar to ESLint's issue for fixers vs suggestions: even if they're fairly straightforward for many, some folks coming in fresh have a hard time with them. I personally had a hard time both with fixers-vs-suggestions _and_ what extension rules are.

JoshuaKGoldberg: Ooh and - pretty soon I think we'll be able to trim down the metadata list down to one sentence per item!

* Suggestions: can link to https://github.com/eslint/eslint/pull/17657 once it's in
* Formatting rules: https://eslint.style is coming soon

...which just leaves the deeper explanation on extension rules.

packages/website/src/components/RulesTable/index.tsx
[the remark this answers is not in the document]
JoshuaKGoldberg: Oh yes now that it's all emojis we're 💯

packages/website/src/components/RulesTable/index.tsx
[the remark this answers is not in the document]
JoshuaKGoldberg: a problem for future us

JoshuaKGoldberg:
Whew, what a PR! Thanks again @Zamiell for iterating it on it with us. Shoutout @rubiesonthesky too!

@Josh-Cena I really liked your feedback and hope I didn't clobber over it in my eagerness to get the PR merged. :3

Zamiell:
do we have a plan for mouseover-functionality-indication?

JoshuaKGoldberg:
I personally have none. If you want to tackle that as a followup issue it could be nice. I'm also wondering if we'd want to rely on a more rich table library that has sorting, filtering, etc. built-in.

packages/website/src/components/RulesTable/index.tsx
JoshuaKGoldberg: #7992 -> #7993

PR #7669, opened 2023-09-18
chore: use named import for `util`

antfu:
## PR Checklist

(sorry for sending this PR without an issue, as I consider it a straightforward refactor)

## Overview

Currently, we are mixing the usage for `* as util` and named import in different rules (and even both in the same file).

I think named import is a generally better syntax as it enables tree-shaking, improves overall consistency, and would help https://github.com/eslint-stylistic/eslint-stylistic to migrate away those formatting rules (we want tree-shaking to avoid bundling all utils again).


antfu:
The failing tests seems to inherit from the main bench.

packages/eslint-plugin/src/rules/adjacent-overload-signatures.ts
```diff
@@ -1,7 +1,7 @@
 import type { TSESTree } from '@typescript-eslint/utils';
 import { AST_NODE_TYPES } from '@typescript-eslint/utils';
 
-import * as util from '../util';
+import { createRule, getNameFromMember, MemberNameType } from '../util';
```
JoshuaKGoldberg, line 4: I... kind of like the `import * as util` approach more personally 😅 . When there are a lot of utilities that files swap between, it's nice not having large imports...

antfu:
Well, to my surprise, [rollup does tree-shake well](https://rollupjs.org/repl/?version=3.29.2&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMiUyRiUyRiUyMFRSRUUtU0hBS0lORyU1Q25pbXBvcnQlMjAqJTIwYXMlMjBtYXRoJTIwZnJvbSUyMCcuJTJGbWF0aHMuanMnJTNCJTVDbiU1Q25jb25zb2xlLmxvZyhtYXRoLmN1YmUoNSkpJTNCJTIwJTJGJTJGJTIwMTI1JTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlMkMlMjJuYW1lJTIyJTNBJTIybWFpbi5qcyUyMiU3RCUyQyU3QiUyMmNvZGUlMjIlM0ElMjIlMkYlMkYlMjBtYXRocy5qcyU1Q24lNUNuJTJGJTJGJTIwVGhpcyUyMGZ1bmN0aW9uJTIwaXNuJ3QlMjB1c2VkJTIwYW55d2hlcmUlMkMlMjBzbyU1Q24lMkYlMkYlMjBSb2xsdXAlMjBleGNsdWRlcyUyMGl0JTIwZnJvbSUyMHRoZSUyMGJ1bmRsZS4uLiU1Q25leHBvcnQlMjBjb25zdCUyMHNxdWFyZSUyMCUzRCUyMHglMjAlM0QlM0UlMjB4JTIwKiUyMHglM0IlNUNuJTVDbiUyRiUyRiUyMFRoaXMlMjBmdW5jdGlvbiUyMGdldHMlMjBpbmNsdWRlZCU1Q24lMkYlMkYlMjByZXdyaXRlJTIwdGhpcyUyMGFzJTIwJTYwc3F1YXJlKHgpJTIwKiUyMHglNjAlNUNuJTJGJTJGJTIwYW5kJTIwc2VlJTIwd2hhdCUyMGhhcHBlbnMhJTVDbmV4cG9ydCUyMGNvbnN0JTIwY3ViZSUyMCUzRCUyMHglMjAlM0QlM0UlMjB4JTIwKiUyMHglMjAqJTIweCUzQiU1Q24lNUNuJTJGJTJGJTIwVGhpcyUyMCU1QyUyMnNpZGUlMjBlZmZlY3QlNUMlMjIlMjBjcmVhdGVzJTIwYSUyMGdsb2JhbCU1Q24lMkYlMkYlMjB2YXJpYWJsZSUyMGFuZCUyMHdpbGwlMjBub3QlMjBiZSUyMHJlbW92ZWQuJTVDbndpbmRvdy5lZmZlY3QxJTIwJTNEJTIwJ2NyZWF0ZWQnJTNCJTVDbiU1Q25jb25zdCUyMGluY2x1ZGVFZmZlY3QlMjAlM0QlMjBmYWxzZSUzQiU1Q25pZiUyMChpbmNsdWRlRWZmZWN0KSUyMCU3QiU1Q24lNUN0JTJGJTJGJTIwT24lMjB0aGUlMjBvdGhlciUyMGhhbmQlMkMlMjB0aGlzJTIwaXMlMjBuZXZlciU1Q24lNUN0JTJGJTJGJTIwZXhlY3V0ZWQlMjBhbmQlMjB0aHVzJTIwcmVtb3ZlZC4lNUNuJTVDdHdpbmRvdy5lZmZlY3QxJTIwJTNEJTIwJ25vdCUyMGNyZWF0ZWQnJTNCJTVDbiU3RCUyMiUyQyUyMmlzRW50cnklMjIlM0FmYWxzZSUyQyUyMm5hbWUlMjIlM0ElMjJtYXRocy5qcyUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJvdXRwdXQlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiU3RCUyQyUyMnRyZWVzaGFrZSUyMiUzQXRydWUlN0QlN0Q=) (but I guess you need to be careful not reference the object entirely, it's not a concern with the current codebase).


Then, I think the issue would fall into mainly consistency/preferences. In that case, I don't have strong opinions over which.

Josh-Cena:
I prefer named imports. The name `util` by itself doesn't add much info.

armano2:
this was partially a reason why we swapped website-eslint to esbuild and started compiling it from source, as currently we are compiling code co cjs that is hard to safely tree-shake,

changing those imports from * as to named reduces complexity of otput code by removing some of required helpers, but its still dosn't solve issue with tree-shaking, esp in rollup

packages/eslint-plugin/src/rules/adjacent-overload-signatures.ts
armano2: this was partially a reason why we swapped website-eslint to esbuild and started compiling it from source, as currently we are compiling code co cjs that is hard to safely tree-shake,

changing those imports from * as to named reduces complexity of otput code by removing (not using) some of required helpers, but its still dosn't solve issue with tree-shaking, esp in rollup

see #6716

----------------
https://www.typescriptlang.org/play?target=7&module=1#code/JYWwDg9gTgLgBAKjgQwM5wK42AGzgMyghDgHIA6cgei11IG4AoRgYwgDtV4YBTAD3gBeTNhzl8ECAAoAlIyA
for code:
```ts

const text = util.foo()
```
currently we are building it to:
```ts
"use strict";
var __createBinding = (this && this.__createBinding) || (Object.create ? (function(o, m, k, k2) {
    if (k2 === undefined) k2 = k;
    var desc = Object.getOwnPropertyDescriptor(m, k);
    if (!desc || ("get" in desc ? !m.__esModule : desc.writable || desc.configurable)) {
      desc = { enumerable: true, get: function() { return m[k]; } };
    }
    Object.defineProperty(o, k2, desc);
}) : (function(o, m, k, k2) {
    if (k2 === undefined) k2 = k;
    o[k2] = m[k];
}));
var __setModuleDefault = (this && this.__setModuleDefault) || (Object.create ? (function(o, v) {
    Object.defineProperty(o, "default", { enumerable: true, value: v });
}) : function(o, v) {
    o["default"] = v;
});
var __importStar = (this && this.__importStar) || function (mod) {
    if (mod && mod.__esModule) return mod;
    var result = {};
    if (mod != null) for (var k in mod) if (k !== "default" && Object.prototype.hasOwnProperty.call(mod, k)) __createBinding(result, mod, k);
    __setModuleDefault(result, mod);
    return result;
};
Object.defineProperty(exports, "__esModule", { value: true });
const util = __importStar(require("../util"));
const text = util.foo();

```

--------------

and for code: https://www.typescriptlang.org/play?target=7&module=1#code/JYWwDg9gTgLgBAbzgMwhOBfFUIjgcgDpCB6AVxmABt8BuAKHoGMIA7AZ3hgFMAPeALwo0ACgCU9IA

```ts
import { foo } from '../util';

const text = foo()
```

we are generating 

```ts
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const util_1 = require("../util");
const text = (0, util_1.foo)();
```

JoshuaKGoldberg:
@typescript-eslint/triage-team I'm up for merging this as-is (so, moving to named imports), for consistency's sake and because tree shaking can be complex sometimes even when everything is configured well. Any objections?

JoshuaKGoldberg (dismissed):
Approving knowing that we might have to clean some stuff up after merging from `main`.

JamesHenry (approved):

JamesHenry:
Thanks @antfu!

Josh-Cena (dismissed):

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -181,24 +185,24 @@ export default util.createRule<Options, MessageIds>({
...
           if (contextualType) {
             // in strict mode you can't assign null to undefined, so we have to make sure that
             // the two types share a nullable type
-            const typeIncludesUndefined = util.isTypeFlagSet(
+            const typeIncludesUndefined = tsutils.isTypeFlagSet(
               type,
               ts.TypeFlags.Undefined,
             );
-            const typeIncludesNull = util.isTypeFlagSet(
+            const typeIncludesNull = tsutils.isTypeFlagSet(
```
JoshuaKGoldberg, line 196: Aha! A behavior change. I'll fix.

antfu:
Should be good now :)

JoshuaKGoldberg:
Thanks!

Issue #7630, opened 2023-09-11
Repo: Playground code refactor: Playground.tsx state switches


JoshuaKGoldberg:
### Suggestion

From https://github.com/typescript-eslint/typescript-eslint/pull/6843#discussion_r1157795745, referring to all the logical ternaries in `Playground.tsx`:

> i have a feeling that i should extract this to separate component, this and/or condition chain is getting complex

https://github.com/typescript-eslint/typescript-eslint/pull/6843/files#diff-c76545bc907e11e58eb71533f91f9887abb2a21dc4d8d2a389317dffbbf5a72c

There are indeed a lot of logical conditions in that file. It'd be nice to streamline the logic a bit to be more readable / straightforward / less complex. Maybe a dedicated component or two? Maybe an enum that gets switched on in render logic?

PR #7672, opened 2023-09-19
chore(website): [playground] improve mobile design

armano2:
## PR Checklist

## Overview

Improve playground design on mobile devices

this change has been extracted from #6656


bradzacher (changes requested):
The layout seems a bit wrong on mobile - the bottom panel is real small and I can't scroll to see the content.

armano2:
can you tell me what browser / os is that, i will check it out

packages/website/src/components/Playground.tsx
```diff
@@ -60,133 +74,151 @@ function Playground(): React.JSX.Element {
     }
   }, []);
 
+  useEffect(() => {
+    if (windowSize === 'mobile') {
+      playgroundMenuRef.current?.collapse();
+    } else if (windowSize === 'desktop') {
+      playgroundMenuRef.current?.expand();
+    }
+  }, [windowSize, playgroundMenuRef]);
```
StyleShit, line 83: why do you need the ref as a dependency here? doesn't it keep the same reference on each re-render anyway?

```suggestion
  }, [windowSize]);
```

packages/website/src/components/Playground.tsx
```diff
@@ -60,133 +74,151 @@ function Playground(): React.JSX.Element {
...
+                  filter={state.showAST === 'es' ? esQueryFilter : undefined}
+                  value={
+                    state.showAST === 'ts'
+                      ? astModel.storedTsAST
+                      : state.showAST === 'scope'
+                      ? astModel.storedScope
+                      : state.showAST === 'es'
+                      ? astModel.storedAST
+                      : undefined
+                  }
```
StyleShit, line 212: maybe this is cleaner?
```suggestion
                  value={ {
                    ts: astModel.storedTsAST,
                    scope: astModel.storedScope,
                    es: astModel.storedAST,
                  }[state.showAST] }
                  
```

packages/website/src/components/Playground.module.css
```diff
@@ -3,89 +3,95 @@
...
 
-.codeBlocks {
-  display: flex;
-  flex: 1 1 0;
-  flex-direction: row;
-  height: 100%;
-  width: calc(100vw - 20rem);
+.hidden {
+  display: none;
+  visibility: hidden;
```
StyleShit, line 19: `display: none` doesn't render at all, why do you need `visibility: hidden`?

StyleShit (changes requested):

packages/website/src/components/Playground.module.css
armano2: we do not need it, i just moved this css from bellow

armano2:
```suggestion
```

packages/website/src/components/Playground.tsx
armano2: agree, https://github.com/typescript-eslint/typescript-eslint/issues/7630

bradzacher (approved):
this is looking good to me from a design perspective and the react code sure looks like react code (it's been a while since I've properly written any react so :nodders:

my only point of feedback is this logic here -  chained, mixed logicals within a JSX expression are pretty gnarly to read, but we've already got #7630 for that so :nodders: again

bradzacher:
something about this PR is breaking in CI. The initial install step is failing to finish (it was auto-cancelled after SIX hours)

JoshuaKGoldberg (approved):
Swell, thanks!

PR #7680, opened 2023-09-20
feat: add a new ESLint parser built on top of SWC

bradzacher:
## Overview

This is an experiment I'm hacking through.
I don't know if we'll end up landing this or not - it's a curiosity.

The idea is simple - for non-type-aware lint setups we don't need to use TS itself as the parser - we just need something that can parse TS code.
TS's parser isn't super fast - it's really general-purpose and has things like error-recovery - features that we don't really want. Also it's written in JS and we all know that JS sucks 😉
OTOH SWC is quite a fast parser! It's designed with performance in mind and is strict.

Sadly SWC doesn't produce an ESTree-compatible AST - so we need to include a converter. This isn't the end of the world, however, considering we currently also do the same thing for the TS AST.

The big question is exactly how much faster it is as an ESLint parser in practice.
Once I've sufficiently hacked away at the converter I'll give it a whirl on Canva's codebase of ~60k files (parsed in a non-type-aware setup, obviously). If I see good enough wins there - chances are regular, less insane codebases would also see wins - and we can consider landing this.

---

In terms of design for the converter - I've chosen to go with a different approach to the `typescript-estree` converter. TSE uses a switch/case style and recurses through the AST in a vague, unsafe manner - it assumes that the converted AST matches the expected types (it does this by liberal use of `any` 😭).

In this new conversion logic I'm trying out explicit, strictly typed conversion functions which ensure that the output type exactly matches the expected AST types. The hope is that this will prevent weird cases of type-misalignment.

If this style works well we can consider converting TSE to this style too (I've been thinking about doing this for a while TBH).

---

TODO:
- [ ] finish conversion logic
- [ ] test performance
- [ ] finish the workspace setup for the new package


liuxingbaoyu:
I'm guessing this might not be much faster or even slower since AST serialization and deserialization in rust and js is very slow.
rollup is currently doing this another way. (Serialize the entire tree through the buffer)
Maybe this will help you!
https://github.com/rollup/rollup/pull/5073

bradzacher:
@liuxingbaoyu note that rollup's comparison is switching from `acorn` to `swc`. Acorn is a much more performant parser than TS is - so it makes sense that doing a quick swap would be on-par/slower.

From my experience migrating from TS to SWC can net some decent wins even accounting for the cost of serialization.

As a side note - I thought that SWC used NAPI and a `SharedArrayBuffer` to pass the data back across the boundary? Am I wrong in thinking that and instead it serialises the object to a string and then marshals the string across the boundary instead?

liuxingbaoyu:
emmm. There is an issue here from two years ago, I'm not sure if there are any improvements now.
https://github.com/swc-project/swc/blob/main/node-swc/src/index.ts#L91

> As a side note - I thought that SWC used NAPI and an `ArrayBuffer` to pass the data back across the boundary? Am I wrong in thinking that and instead it serialises the object to a string and then marshals the string across the boundary instead?

It seems to be using `JSON.parse`. :)
https://github.com/swc-project/swc/blob/8231f874e759ad32ef75a79b18afc4ae163381b8/node-swc/src/index.ts#L91

JoshuaKGoldberg:
Lots of merge conflicts and this is taking up space in the PR queue. I'm going to close this old draft just to keep the queue small.

PR #7691, opened 2023-09-24
feat(eslint-plugin): [no-unsafe-enum-comparison] add switch suggestion

### LINKED ISSUE #7643: Enhancement: [no-unsafe-enum-comparison] Suggestion fixer to switch to enum value

https://typescript-eslint.io/rules/no-unsafe-enum-comparison

### Description

If the rule detects an unsafe enum<>literal comparison that happens to match with a member of the enum's type, the rule report could include a suggestion to switch the literal to that literal value.

```diff
- fruit === "apple";
+ fruit === Fruit.Apple;
```

### Fail

```typescript
enum Fruit {
  Apple = 'apple',
}

declare let fruit: Fruit;

fruit === "apple";
```

### Pass

```typescript
enum Fruit {
  Apple = 'apple',
}

declare let fruit: Fruit;

fruit === Fruit.Apple;
```

### Additional Info

_No response_

StyleShit:
Closes #7643


packages/eslint-plugin/src/rules/no-unsafe-enum-comparison.ts
```diff
@@ -100,11 +123,44 @@ export default util.createRule({
...
+            return context.report({
+              messageId: 'mismatchedSimilar',
+              node,
+              data: {
+                replacement: rightEnumKey,
+              },
+            });
+          }
+
+          return context.report({
```
JoshuaKGoldberg, line 163: By "suggestion" I meant the actual ESLint feature: https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions. I don't think we need to add any new reports?

StyleShit: Oh ok, sorry, didn't get that 

so you mean something like this?
```TS
if (typeViolates(leftTypeParts, right)) {
  const enumKey = getEnumKeyForLiteral(...);

  return context.report({
    messageId: 'mismatched',
    node,
    suggest: [
      {
         desc: "Replace with the literal value with an enum member.",
         fix(fixer) {
           return fixer.replaceText(...);
         }
      },
    ],
  });
}
```

In addition, is there any place where I can find some documentation about this?
My current implementation of `getEnumKeyForLiteral()` is basically me guessing using TS, and some basic knowledge of ASTs, but it's not 100% working and I can't seem to find a good resource about this stuff.

___
(sorry if those are dumb questions, I just watched [this](https://www.youtube.com/watch?v=OVNQWzdhCQA) video and wanted to help 😅)

JoshuaKGoldberg: Yeah along those lines! I was thinking vaguely like:

```diff
context.report({
  messageId: 'mismatched',
  node,
+  suggest: [
+    {
+      fix(fixer) {
+        return fixer.replaceText(...);
+      },
+      memessageId "replaceValueWithEnum",
+    },
  ]
});
```

The idea here is that we're not changing when the rule reports. We're only adding in a new `suggest` property to add in a suggestion.

I haven't looked too deeply, but it _might_ get tricky trying to figure out what text to add in. If the enum is declared in another file then I don't know that we have a good way to add in an `import` for it automatically. No worries if you don't implement that part just yet.

...and cool! Really happy to see folks come in from MiTS 😄 welcome!

StyleShit: The main reason I split it into two conditions is to determine which of them violated and to replace the text accordingly.
Will try to look at it again and see if I can get any progress

StyleShit:
@JoshuaKGoldberg 

OK, should be better now, but I still can't figure out the types there 😓

StyleShit:
How stupid is it to tell you that i'm in your live twitch but can't send messages? 😂

packages/eslint-plugin/src/rules/enum-utils/shared.ts
```diff
@@ -33,8 +48,36 @@ export function getEnumTypes(
...
+ * getEnumKeyForLiteral([Fruit.Apple, Fruit.Banana], 'orange') --> null
+ * ```
+ */
+export function getEnumKeyForLiteral(
+  enumLiterals: ts.Type[],
+  literal: unknown,
+): string | null {
+  for (const enumLiteral of enumLiterals) {
+    // @ts-expect-error Not sure why `value` is not on `enumLiteral`.
+    if (enumLiteral.value === literal) {
```
JoshuaKGoldberg, line 74: It's because `ts.Type` is the general type for any type in code. But it seems like this is guaranteed to be the more specific `ts.LiteralType`, since it has a `value` and is from the enum members. (I found that type by going to definition on `ts.Type` and searching for `value: `)

You'll want to put an assertion or type predicate or similar somewhere to indicate that the value being provided to `enumLiterals` is `ts.LiteralType[]`.

Tip: I think this might be helpful...

```plaintext
.filter((subType): subType is ts.LiteralType => ...;
```

packages/eslint-plugin/src/rules/enum-utils/shared.ts
```diff
@@ -33,8 +48,36 @@ export function getEnumTypes(
...
+      const { symbol } = enumLiteral;
+
+      // @ts-expect-error Not sure why `parent` is not on `symbol`.
+      return `${symbol.parent.name}.${symbol.name}`;
```
JoshuaKGoldberg, line 78: 🤔 great question. I don't know. It must be part of the private API...

In the past we've had to augment TypeScript's types. See `packages/*/typescript/typescript.d.ts`. But this is very risky (what if they update their internals?) so we try not to.

Instead, you could go through the AST: access `symbol.valueDeclaration` and assert that it's an `EnumMember`. You'll then be encouraged by the type system to handle all the possible allowed cases (what if the member name is a literal like `'abc'`?).

JoshuaKGoldberg (changes requested):
Really good start so far! The code looks clean and I like the general direction. 😄 

Left a few comments and suggestions for how to overcome the `// @ts-expect-error`s. I suspect they'll end up showing you a bunch more test cases to add. Have fun!

packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
```diff
@@ -565,5 +565,63 @@ ruleTester.run('strict-enums-comparison', rule, {
...
+            },
+          ],
+        },
+      ],
+    },
+    {
+      code:
+        `enum Mixed {A = 1, B = 'b'} ` +
+        `declare const mixed: Mixed; ` +
+        `mixed === 1 || mixed === 'b';`,
```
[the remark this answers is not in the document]
StyleShit: It was actually just to avoid line breaks, but you're right, it should be consistent. Done.

packages/eslint-plugin/src/rules/no-unsafe-enum-comparison.ts
```diff
@@ -107,6 +113,53 @@ export default util.createRule({
...
+                  const leftEnumKey = getEnumKeyForLiteral(
+                    getEnumLiterals(left),
+                    util.getStaticValue(node.right)?.value,
+                  );
+
+                  if (leftEnumKey != null) {
+                    return fixer.replaceText(
+                      node,
+                      `${leftExpression} ${node.operator} ${leftEnumKey}`,
+                    );
```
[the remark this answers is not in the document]
StyleShit: TIL, thanks :)

packages/eslint-plugin/src/rules/enum-utils/shared.ts
StyleShit: I think I got it right this time, and it's simpler than what you've suggested, thanks!

packages/eslint-plugin/src/rules/enum-utils/shared.ts
```diff
@@ -33,8 +50,39 @@ export function getEnumTypes(
...
+ * match. For example:
+ * ```ts
+ * enum Fruit {
+ *   Apple = 'apple',
+ *   Banana = 'banana',
+ * }
+ *
+ * getEnumKeyForLiteral([Fruit.Apple, Fruit.Banana], 'apple') --> 'Fruit.Apple'
+ * getEnumKeyForLiteral([Fruit.Apple, Fruit.Banana], 'banana') --> 'Fruit.Banana'
+ * getEnumKeyForLiteral([Fruit.Apple, Fruit.Banana], 'cherry') --> null
```
StyleShit, line 67: @JoshuaKGoldberg 

Nit: changed from `orange` to `cherry`

packages/eslint-plugin/src/rules/enum-utils/shared.ts
```diff
@@ -33,8 +50,40 @@ export function getEnumTypes(
...
+    if (enumLiteral.value === literal) {
+      const { symbol } = enumLiteral;
+
+      const memberDeclaration = symbol.valueDeclaration as ts.EnumMember;
+      const enumDeclaration = memberDeclaration.parent;
+
+      const enumName = enumDeclaration.name.getText();
+      const memberName = memberDeclaration.name.getText();
+
+      return `${enumName}.${memberName}`;
```
StyleShit, line 84: This area could definitely have fewer variables, but TBH, the only reason I have so many of them is to make it look & feel better and more readable (e.g. prettier breaks long lines, and naming those things seems weird).

IDK if that's a performance concern, should I change it?

packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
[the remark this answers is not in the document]
StyleShit: Done

packages/eslint-plugin/src/rules/enum-utils/shared.ts
```diff
@@ -20,6 +20,23 @@ function getBaseEnumType(typeChecker: ts.TypeChecker, type: ts.Type): ts.Type {
...
 
+/**
+ * Retrieve only the Enum literals from a type. for example:
+ * - 123 --> []
+ * - {} --> []
+ * - Fruit.Apple --> [Fruit.Apple]
+ * - Fruit.Apple | Vegetable.Lettuce --> [Fruit.Apple, Vegetable.Lettuce]
+ * - Fruit.Apple | Vegetable.Lettuce | 123 --> [Fruit.Apple, Vegetable.Lettuce]
+ * - T extends Fruit --> [Fruit]
+ */
```
JoshuaKGoldberg, line 31: [Praise] Great comments, thanks for continuing the existing standard 🙂

packages/eslint-plugin/src/rules/enum-utils/shared.ts
```diff
@@ -33,8 +50,40 @@ export function getEnumTypes(
...
+  literal: unknown,
+): string | null {
+  for (const enumLiteral of enumLiterals) {
+    if (enumLiteral.value === literal) {
+      const { symbol } = enumLiteral;
+
+      const memberDeclaration = symbol.valueDeclaration as ts.EnumMember;
+      const enumDeclaration = memberDeclaration.parent;
+
+      const enumName = enumDeclaration.name.getText();
```
JoshuaKGoldberg, line 81: [Cleanup] We generally try to avoid `getText()` when possible. It's a blunt function that directly draws from the source code text - which means it papers over nuances we might care about, such as if an `id` is actually a computed property.

```suggestion
      const enumName = enumDeclaration.name.text;
```

packages/eslint-plugin/src/rules/enum-utils/shared.ts
```diff
@@ -33,8 +50,40 @@ export function getEnumTypes(
...
+      const memberName = memberDeclaration.name.getText();
```
JoshuaKGoldberg, line 82: [Testing] Although TypeScript says its enums can't have computed property names, `["key"]` literals are allowed. There should be some tests for that. I see you live added one for `'a' = 1,`, let's do `['a'] = 1,` (or an equivalent) too? And `['a' /* gotcha */] = 1,` (or an equivalent) - comments tend to get in the way when using `getText()`!

Fortunately, `[someVariableId]` and `[123]` aren't allowed. So if you need an assertion or two, that might make sense.

packages/eslint-plugin/src/rules/enum-utils/shared.ts
JoshuaKGoldberg: Yeah the number of variables doesn't really impact performance. I like how you set it up 😄

JoshuaKGoldberg (changes requested):
Looking great! Structure is solid, good test coverage, super readable code. Nice!

Just requesting changes on a bit more test coverage and some `getText()` -> `name` touchups.

Shaun the Sheep giving two thumbs up

packages/eslint-plugin/src/rules/enum-utils/shared.ts
StyleShit: Done

packages/eslint-plugin/src/rules/enum-utils/shared.ts
StyleShit: Done

packages/eslint-plugin/src/rules/enum-utils/shared.ts
```diff
@@ -33,8 +50,52 @@ export function getEnumTypes(
...
+
+      switch (memberNameIdentifier.kind) {
+        case ts.SyntaxKind.Identifier:
+          return `${enumName}.${memberNameIdentifier.text}`;
+
+        case ts.SyntaxKind.StringLiteral:
+          return `${enumName}['${memberNameIdentifier.text}']`;
+
+        case ts.SyntaxKind.ComputedPropertyName:
+          return `${enumName}[${memberNameIdentifier.expression.getText()}]`;
```
StyleShit, line 92: had to use `.getText()` here, there is no `.text` for some reason
seems like it's fine? See the test for that as a reference

packages/eslint-plugin/src/rules/enum-utils/shared.ts
```diff
@@ -33,8 +50,52 @@ export function getEnumTypes(
...
+
+      const memberNameIdentifier = memberDeclaration.name;
+      const enumName = enumDeclaration.name.text;
+
+      switch (memberNameIdentifier.kind) {
+        case ts.SyntaxKind.Identifier:
+          return `${enumName}.${memberNameIdentifier.text}`;
+
+        case ts.SyntaxKind.StringLiteral:
+          return `${enumName}['${memberNameIdentifier.text}']`;
```
StyleShit, line 89: not sure what to do in this case, should I put single quotes? double quotes? somehow read it from the source?

packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
```diff
@@ -565,5 +565,308 @@ ruleTester.run('strict-enums-comparison', rule, {
...
+          ],
+        },
+      ],
+    },
+    {
+      code: `
+        enum ComputedKey {
+          [\`test-
+          key\` /* with comment */] = 1,
+        }
```
StyleShit, line 847: (Why is this even legal?!)

JoshuaKGoldberg: to make our lives harder 🫠

packages/eslint-plugin/src/rules/enum-utils/shared.ts
JoshuaKGoldberg: This brings up a couple necessary test cases:

* `'` with a `\'` inside
* `"` with a `\"` inside

There... may be an existing utility for this? I haven't searched for it.

packages/eslint-plugin/src/rules/enum-utils/shared.ts
JoshuaKGoldberg: The type of `memberNameIdentifier.expression` is `ts.Expression`, which is a pretty wide type. You could `as` assert it to the union of types that are actually allowed as enum member names... or just go with `.getText()`. If the unit tests aren't showing any bad behavior, it's probably fine. 😎

packages/eslint-plugin/src/rules/enum-utils/shared.ts
JoshuaKGoldberg: Oh, and https://github.com/typescript-eslint/typescript-eslint/issues/7747 implies the `switch-exhaustiveness-check` fixer for enums might be useful.

StyleShit: Seems like it also has some bugs. Enums are awful.
Filed #7768

JoshuaKGoldberg: Love it. Let's not introduce new bugs here, and anything existing can be tackled there.

packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
StyleShit: It's hard enough without it! ☹

packages/eslint-plugin/src/rules/enum-utils/shared.ts
StyleShit: @JoshuaKGoldberg 

[I hate enums](https://github.com/typescript-eslint/typescript-eslint/pull/7691/commits/cf560404a626abebff0dc29fcb19d8c21c4cfda4)

anything I've missed?

packages/eslint-plugin/src/rules/enum-utils/shared.ts
StyleShit: I would probably keep it like this, seems fine in the tests (unless I'm missing some cases?)

JoshuaKGoldberg: The existing bugs in #7768 make me not _too_ motivated to get every single possible edge case in this PR. The tests you've added are pretty great. Thanks!

packages/eslint-plugin/src/rules/enum-utils/shared.ts
JoshuaKGoldberg: Ah found this thread - yeah we can tackle other issues in #7768. I think you got all the introduced ones!

PR #7696, opened 2023-09-27
feat(eslint-plugin): [max-params] don't count `this: void` parameter

### LINKED ISSUE #7538: Base rule extension: [max-params] don't count this: void

### Link to the base rule

https://eslint.org/docs/latest/rules/max-params

### Description

It would be great ignore "this: void" by default.

Eslint cries about "this: void" even though it is not a real parameter.

### Fail

```typescript
// Fail. Still report an error if too much parameters are there:

private doSomethingWild (x: number, y: number, z: number, k: any) {
 //
}
```

### Pass

```typescript
// Pass. Don't count "this: void".

private doSomethingWild (this: void, x: number, y: number, z: number) {
 //
}
```

### Additional Info

I don't have time to make it myself. :(

StyleShit:
Closes #7538


packages/eslint-plugin/src/rules/max-params.ts
```diff
@@ -0,0 +1,99 @@
...
+              countVoidThis: {
+                type: 'boolean',
+              },
+            },
+          },
+        ],
+      },
+    },
+  ),
+) as JSONSchema4[];
```
StyleShit, line 39: for some reason, I can't get `deepMerge` to work properly here. 

I've tried adding these lines to add support for array overrides, but I'm not sure why it doesn't work:
```diff
if (isObjectNotArray(firstValue) && isObjectNotArray(secondValue)) {
  // object type
  acc[key] = deepMerge(firstValue, secondValue);
+ } else if (Array.isArray(firstValue)) {
+   // array type
+   acc[key] = Object.values(
+     deepMerge({ ...firstValue }, { ...(secondValue as ObjectLike) }),
+   );
+ } else {
  // value type
  acc[key] = secondValue;
}
```

seems to work fine when isolated though:
https://tsplay.dev/wX1GLN

packages/eslint-plugin/src/rules/max-params.ts
```diff
@@ -0,0 +1,97 @@
...
+        type: 'integer',
+        minimum: 0,
+      },
+      countVoidThis: {
+        type: 'boolean',
+      },
+    },
+    additionalProperties: false,
+  },
+] as JSONSchema4[];
```
StyleShit, line 37: for some reason, I can't get `deepMerge` to work properly here. 

I've tried adding these lines to add support for array overrides, but I'm not sure why it doesn't work:
```diff
 if (isObjectNotArray(firstValue) && isObjectNotArray(secondValue)) {
   // object type
   acc[key] = deepMerge(firstValue, secondValue);
+ } else if (Array.isArray(firstValue)) {
+   // array type
+   acc[key] = Object.values(
+     deepMerge({ ...firstValue }, { ...(secondValue as ObjectLike) }),
+   );
  } else {
   // value type
   acc[key] = secondValue;
 }
```

seems to work fine when isolated though:
https://tsplay.dev/wX1GLN
___
Anyway... I should've probably still hard-coded the values here since I don't support passing only a number like the base rule does (maybe I should?)

packages/eslint-plugin/src/rules/max-params.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (countVoidThis === true) {
+      return baseRules;
+    }
+
+    const removeVoidThisParam = <T extends FunctionLike>(node: T): T => {
+      if (node.params.length === 0) {
+        return node;
+      }
+
+      const params = [...node.params];
```
JoshuaKGoldberg, line 62: [Performance] This is probably not impactful for most users, but just in case: it's best not to create arrays unnecessarily. It's a bit extra memory.

Much of this function's _"create a shifted `params` if necessary"_ could be reduced down to:

```ts
const params =
  node.params.length &&
  node.params[0].type === AST_NODE_TYPES.Identifier &&
  node.params[0].name === 'this' &&
  node.params[0].typeAnnotation?.typeAnnotation.type ===
    AST_NODE_TYPES.TSVoidKeyword
    ? node.params.slice(1)
    : node.params;
```

packages/eslint-plugin/src/rules/max-params.ts
JoshuaKGoldberg: Extended rules generally directly go off `context.options`, is that what you're looking for?

https://github.com/typescript-eslint/typescript-eslint/blob/ca1abb5cdaecd8f039a807b47db2a642e14fbc1e/packages/eslint-plugin/src/rules/brace-style.ts#L30-L32

JoshuaKGoldberg (changes requested):
Looks great to me! Nice and straightforward, I like it. 🙂 

Left a couple of suggestions - what do you think?

packages/eslint-plugin/src/rules/max-params.ts
StyleShit: Nope, I'm talking about `baseRule.meta.schema`.
I had some issues when trying to add an option to the base schema, the deepMerge function seems to break it.
Is it OK to keep it hard-coded instead of extending the original schema in runtime?

packages/eslint-plugin/docs/rules/max-params.md
```diff
@@ -0,0 +1,10 @@
+---
+description: 'Enforce a maximum number of parameters in function definitions.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/max-params** for documentation.
+
+This rule extends the base [`eslint/max-params`](https://eslint.org/docs/rules/max-params) rule.
+This version adds support for TypeScript `this` parameter so it won't be counted as a parameter.
```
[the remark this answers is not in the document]
StyleShit: I'm not a native speaker and both sound ok to me, so I'm not sure if this is an issue, should probably just trust you? 😅 

___
I actually _love_ nit-picking in CRs, go for it 😄

packages/eslint-plugin/src/rules/max-params.ts
StyleShit: yup, I actually had thoughts about performance while writing this, but didn't think it's gonna much of an issue. 
You're totally right (even though it's just a shallow copy of, and let's be real here, _maybe_ 10 items max, but yeah totally agree).

TBH, I don't like huge ternaries, but having it inside a regular condition doesn't look good either, so I just took the cursed ternary path.

JoshuaKGoldberg: Yeah to me the important bit is only making a new array if needed.

packages/eslint-plugin/src/rules/max-params.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    const removeVoidThisParam = <T extends FunctionLike>(node: T): T => {
+      if (node.params.length === 0) {
+        return node;
+      }
+
+      const params =
+        node.params[0] &&
+        node.params[0].type === AST_NODE_TYPES.Identifier &&
```
JoshuaKGoldberg, line 64: [Nit] If `node.params.length` isn't `0` then we're guaranteed `node.params[0]` exists:

```suggestion
        node.params[0].type === AST_NODE_TYPES.Identifier &&
```

packages/eslint-plugin/src/rules/max-params.ts
JoshuaKGoldberg: Ah got it - yeah this is fine. Base rules don't change often and many others already hard-code it (e.g. `dot-notation`). Thanks for asking!

packages/eslint-plugin/src/rules/max-params.ts
```diff
@@ -0,0 +1,91 @@
+import { AST_NODE_TYPES, type TSESTree } from '@typescript-eslint/utils';
+
+import * as util from '../util';
```
JoshuaKGoldberg, line 3: Nit: In the meantime, we switched the style to named imports (#7669). I can make the change here since it's good to merge other than these nits.

JoshuaKGoldberg (approved):
Looks great, thanks for iterating on this! Great work 🚀 

Just a couple of small things that I can touch up before merging.

Woman at computer saying 'oh, hell yes'

packages/eslint-plugin/src/rules/max-params.ts
JoshuaKGoldberg: ...come to think of it, we can get rid of the new object altogether by also running `return node;` if we don't need to slice/shift the params. I'll quick edit that in.

StyleShit: Yup, good catch
I want to blame copilot for this, but I don't remember whose fault it is 🫣

JoshuaKGoldberg: 😄 if you have to ask 😛 

(I also missed this until just now!)

PR #7708, opened 2023-10-01
fix(eslint-plugin): [no-unsafe-return] allow returning anything if explicitly returning any

### LINKED ISSUE #3549: [no-unsafe-return] does not allow return any[] when function is typed as `any`

**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/no-unsafe-return": "error"
  }
}
```

```TS
function foo(): any {
    if (Math.random()) {
        let res: any = {};
        return res;
    }
    let resArr: any[] = [];
    resArr.push(true);
    return resArr; // <-- eslint unhappy
}
```

**Expected Result**

`any` is any and the rule should be happy with anything :-)

**Actual Result**

error @typescript-eslint/no-unsafe-return : Unsafe return of an `any[]` typed value.

**Additional Info**

Follow up to #3482 

**Versions**

| `@typescript-eslint/eslint-plugin` | `4.27.0` |
| `@typescript-eslint/parser`        | `4.27.0` |
| `TypeScript`                       | `4.1` |
| `ESLint`                           | `7.27.0` |

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3549

## Overview

I'm not sure if we should handle the following case: `const x: () => any = () => [] as any[];`

Overall the docs does not make it clear which cases are exempted, so I guess we have some room for interpretation.

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-unsafe-return.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unsafe-return.test.ts
@@ -114,6 +114,11 @@ function foo(): Set<number> {
         return [] as any[];
       }
     `,
+    `
+      function foo(): unknown {
+        return [] as any[];
+      }
+    `,
   ],
   invalid: [
     {
--- a/packages/eslint-plugin/tests/rules/no-unsafe-return.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unsafe-return.test.ts
@@ -108,6 +108,12 @@ function foo(): Set<number> {
         return new Map();
       }
     `,
+    // https://github.com/typescript-eslint/typescript-eslint/issues/3549
+    `
+      function foo(): any {
+        return [] as any[];
+      }
+    `,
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/tests/rules/no-unsafe-return.test.ts
```diff
@@ -108,6 +108,12 @@ function foo(): Set<number> {
         return new Map();
       }
     `,
+    // https://github.com/typescript-eslint/typescript-eslint/issues/3549
+    `
+      function foo(): any {
+        return [] as any[];
+      }
+    `,
```
JoshuaKGoldberg, line 116: [Testing] Could you add a test for `unknown` too?

JoshuaKGoldberg (approved):
👍 very reasonable!

JoshuaKGoldberg (approved):
🚀

PR #7709, opened 2023-10-01
feat(eslint-plugin): [no-restricted-imports] support import = require

### LINKED ISSUE #2563: [no-restricted-imports] No way to use this rule with `import = require()` syntax.

**Repro**

```JSON
{
  "rules": {
    "no-restricted-imports": [
      "error",
      {
        paths: ["lodash"],
      },
    ],
  }
}
```

```TS
import _ = require("lodash");
```

**Expected Result**

Should result in a linting error.

**Actual Result**

No errors.

**Additional Info**

**Versions**

| package                                 | version |
| --------------------------------------- | ------- |
| `@typescript-eslint/eslint-plugin`      | `4.1.0` |
| `@typescript-eslint/parser`             | `4.1.0` |
| `@typescript-eslint/typescript-estree`  | `4.1.0` |
| `@typescript-eslint/experimental-utils` | `4.1.0` |
| `TypeScript`                            | `3.5.2` |
| `node`                                  | `12.16.3` |
| `npm`                                   | `6.14.7` |

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2563


packages/eslint-plugin/src/rules/no-restricted-imports.ts
```diff
@@ -269,28 +269,55 @@ export default createRule<Options, MessageIds>({
...
         if (
-          node.importKind === 'type' ||
-          (node.specifiers.length > 0 &&
-            node.specifiers.every(
-              specifier =>
-                specifier.type === AST_NODE_TYPES.ImportSpecifier &&
-                specifier.importKind === 'type',
-            ))
+          !isAllowedTypeImportPath(importSource) &&
+          !isAllowedTypeImportPattern(importSource)
```
JoshuaKGoldberg, line 285: [Cleanup] Nit: there's a teeny bit of duplication with calling `!isAllowedTypeImportPath(importSource) && !isAllowedTypeImportPattern(importSource)`. Maybe worth a helper function? It's only two places, so maybe not...

JoshuaKGoldberg (approved):
Swell, LGTM!

I think this is a breaking change and blocked on going into our next major release?

Josh-Cena:
I don't think this qualifies as a breaking change (if people really think so it can be behind an option). We usually don't automatically categorize "more errors" as breaking change, especially in the case where an error should be highly expected and the lack of one can only be explained as a bug.

JoshuaKGoldberg:
Yeah I think you're right. In theory, if a lot of devs still used `import = require` syntax, then I think this'd be more debatable. But it's not a super-used feature.

🚀 ship it!

PR #4436, opened 2022-01-12
feat(eslint-plugin): [prefer-readonly-parameter-types] added an optional type allowlist

marekdedic:
BREAKING CHANGE:
Changes arguments to a public API

## PR Checklist

-   [x] Addresses an existing issue: fixes #4185
-   [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
-   [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

I have added an option type whitelist to the rule `prefer-readonly-parameter-types`.

~This PR is not yet ready:~ ✔️ 
- [x] Check that the type is from the standard library
- [x] Decide whether the whitelist really should be user-configurable
- [x] Decide what to do with third-party types that cannot be made readonly, such as `JQueryStatic` (which leads to sindresorhus/type-fest#337)

PR #7710, opened 2023-10-01
feat(eslint-plugin): [no-var-requires, no-require-imports] allow option

### LINKED ISSUE #1902: @typescript-eslint/no-var-requires leads to broken npm publications
**Repro**

Create a typescript package that imports fields from `package.json`, e.g.,

```js
import { engines as supportedEngines } from '../package.json';
```

Attempt to build and publish the module. You will get a build that looks something like this:

```bash
npm notice === Tarball Contents === 
npm notice 11.4kB LICENSE                                 
npm notice 1.9kB  build/package.json                      
npm notice 1.7kB  package.json                            
npm notice 1.2kB  README.md                               
npm notice 757B   build/src/backoff-timeout.d.ts          
npm notice 2.7kB  src/backoff-timeout.ts                  
npm notice 684B   build/src/call-credentials-filter.d.ts  
npm notice 2.5kB  src/call-credentials-filter.ts        
```

☝️ this bundle should include `backoff-timeout.js`, `call-credentials-filter.js`. There appears to be a bug (_expected behavior?_) with npm, such that if there's a `package.json` file in your `build/` directory, no JavaScript is included in the publish.

**Expected Result**

`@typescript-eslint/no-var-requires` should not advise users to `import` from `package.json`, as it can lead to bad publications to `npm`.

**Actual Result**

`@typescript-eslint/no-var-requires` suggests that you switch any `require`s from `package.json` to `imports`.

**Additional Info**

here's where we ran into the issue: https://github.com/grpc/grpc-node/issues/1357

**Versions**

`N/A`, this is related specifically to the `@typescript-eslint/no-var-requires` plugin, and is a suggestion for behavior.

## Thoughts

To help protect folks from this hiccup, it might be nice if `@typescript-eslint/no-var-requires` was able to specifically suggest that folks use `require` on a `package.json` import; while suggesting `import` for other JSON files.

### LINKED ISSUE #5314: Enhancement: [no-require-imports] Add option `allowJson`

https://typescript-eslint.io/rules/no-require-imports/

### Description

If `module` in `tsconfig.json` is set to `Node16`, TypeScript 4.7 will convert `import … = require('…')` statements to require calls created using `_createRequire`.

As long as import assertions are not allowed in Node.js, I believe this is a nice syntax to read `package.json` files.

I propose to add an option named `allowJson` to `no-require-imports` that allows require imports of JSON files.

### Fail

```typescript
import lodash = require('lodash')
import lib = require('./lib')
```

### Pass

```typescript
These are currently also errored on


import lodashPkg = require('lodash/package.json')
import pkg = require('./package.json')
```

### Additional Info

In:

```ts
import pkg = require('./package.json')

console.log(pkg.version)
```

Out:

```js
import { createRequire as _createRequire } from "module";
const __require = _createRequire(import.meta.url);
const pkg = __require("./package.json");
console.log(pkg.version);
```

[TypeScript playground](https://www.typescriptlang.org/play?target=99&module=100#code/JYWwDg9gTgLgBGA1gczgXjlApgRwK7DYAUA5AHQD0YAhgMaLXJZkBWAzhAHYkCUAUH1pcOAG2YiIyIkmRkAblihtgXHkA)

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1902, fixes #5314


Josh-Cena:
#5314 asks for `allowJson`. This PR narrows the scope, because in fact most JSON can be correctly imported in Node, as long as it lives inside `src` (although you would have to use import attributes, which is highly experimental). Not sure if we should expand the scope to cover all JSON.

packages/eslint-plugin/docs/rules/no-require-imports.md
```diff
@@ -28,6 +28,28 @@ import { lib2 } from 'lib2';
 import * as lib3 from 'lib3';
 ```
 
+## Options
+
+### `allowPackageJson`
```
JoshuaKGoldberg, line 33: I've also seen libraries want to import from `turbo.json`, `project.json`, etc. Maybe this should be the standard format from https://github.com/typescript-eslint/typescript-eslint/discussions/6017, with docs showing how to use it for `package.json` files? WDYT?

Josh-Cena: I'm okay with that. What's your proposed API? My thought is something similar to `no-restricted-imports`, with an array of `paths`.

JoshuaKGoldberg: I was thinking the exact format from  https://github.com/typescript-eslint/typescript-eslint/pull/4436 -> https://typescript-eslint.io/rules/prefer-readonly-parameter-types.

```json
{
  "@typescript-eslint/no-var-requires": [
    "error",
    {
      ignore: [
        {
          name: "package.json",
          source: "file",
        },
      ],
    },
  ],
}
```

Josh-Cena: This looks couunterintuitive as a user, TBH. This API is for _types_, but what is `"package.json"` as a type? What does `source: "file"` even mean?

JoshuaKGoldberg: What happens if they want to import from a node module, e.g. `import emojis from "emojilib/dist/emoji-en-US.json"`? Should we just have a `string[]` as the option to capture all those cases? Feels off to me that we have the big thought-out format and then don't use it... but it is overhead, yes. 🤔

JoshuaKGoldberg: ping @Josh-Cena ?

Josh-Cena: Sorry, I did not forget this 😅 Will try to garner some energy and clear my notifications this afternoon.

Josh-Cena: @JoshuaKGoldberg I thought about it. Here are my thoughts:

1. The only thing this PR changes is allowing `require`. It doesn't change anything about `import`—they can still be used as normal.
2. Sure, we have a thought-out format for specifying _types_, but here we are talking about import _paths_, and this design doesn't come without its precedent either—we are being consistent with `no-restricted-imports`.
3. About having `string[]` as an option. You mean `{ ignorePaths: string[] }`? My initial design is `{ ignorePaths: TheOptionsOfNoRestrictedImports }`, i.e. it supports multiple ignore paths, each one with a path, a glob, a custom message, etc. But I now think that's hugely unnecessary for an _ignore_ option. So sure, I can do that.
4. If you want to differentiate between node modules and local JSON files, you would have to support regex in the ignore path (I think) to be able to tell between `./foo/data.json` and `foo/data.json`. That means the option has to be more granular than a `string[]`; it would at minimum need to be `{ ignorePaths: { paths: string[] /* globs */, patterns: string[] /* regexes */ }`. Does that make sense? Am I overthinking it?

I'm very curious what your current idea for the option's shape is.

JoshuaKGoldberg: The intent of https://github.com/typescript-eslint/typescript-eslint/discussions/6017 was: _"RFC: Common format to specify a type or value as a rule option"_. I was really hoping it'd be usable for situations exactly like this one. But if we're seeing a practical shortcoming of the format -it being too heavy-handed- then I like this as a motivation to improve it. Or maybe it's just irrelevant here...

Maybe just `allow: string[]` or `ignore: string[]`, where each of the string elements is a regex? It'd be a little verbose to have to say something like `"package\.json$"]`, but not the worst thing in the world IMO.

Thinking on the sticking point of relativity: from what I can think of either:
* the rule gets told a root ESLint path to resolve from _(but what about monorepos with config in the root vs per-package?)_
* we tell users to either increase regex specificity or allow sibling level `package.json` imports (`/[.\\\/]*package.json/`)

I think having two things to think about, like `paths` and `patterns` from 4., would be a bit much. Is there anything that can only be captured in that system, not the single option?

bradzacher: The complicated thing is that sometimes you have tsconfig paths which make things look non-relative. We can't build in any specific assumptions around the path here if we want people to allow local but disallow node_module packages done this way.

So a regex is probably the most flexible option for this so codebases can easily devise some regex that matches their usecase. I'd suspect that most codebases would just want something like you suggested joshg to allow a relative package.json

Josh-Cena:
After https://github.com/typescript-eslint/typescript-eslint/issues/8092 we can finally get rid of this sucky duplication...

packages/eslint-plugin/src/rules/no-require-imports.ts
```diff
@@ -1,27 +1,59 @@
...
+    }
     return {
       'CallExpression[callee.name="require"]'(
         node: TSESTree.CallExpression,
       ): void {
+        if (
+          node.arguments[0]?.type === AST_NODE_TYPES.Literal &&
+          typeof node.arguments[0].value === 'string' &&
+          isImportPathAllowed(node.arguments[0].value)
+        ) {
```
StyleShit, line 54: does it work with string constants?

```TS
const path = 'package.json';
const pkg = require(path);
```

Josh-Cena: No, that needs to make the rule type-aware, or otherwise implement a go-to-definition mechanism, which I'm not sure we have the precedent. I'm also not sure if it happens more than a handful of times.

packages/eslint-plugin/src/rules/no-var-requires.ts
```diff
@@ -18,14 +22,39 @@ export default createRule<Options, MessageIds>({
...
+      return allowPatterns.some(pattern => importPath.match(pattern));
+    }
     return {
       'CallExpression[callee.name="require"]'(
         node: TSESTree.CallExpression,
       ): void {
+        if (
+          node.arguments[0]?.type === AST_NODE_TYPES.Literal &&
+          typeof node.arguments[0].value === 'string' &&
+          isImportPathAllowed(node.arguments[0].value)
```
JoshuaKGoldberg, line 54: [Bug] I think this should use `getStaticStringValue` or an equivalent tailored just to strings. Then it would support template literals.

I did a quick search for `type === AST_NODE_TYPES.Literal` and found a few files `Referencer.ts`, `prefer-regexp-exec.ts`, `prefer-reduce-type-parameter.ts`, `padding-line-between-statements.ts`, `no-restricted-imports.ts`...) that do similar _"get the value if it's a string literal"_ logic. Maybe good to make a new utility?

Thoughts?

JoshuaKGoldberg (approved):
LGTM! The template literal support I trust you on 😄 it's not high priority IMO. Otherwise this looks great, thanks!

packages/eslint-plugin/src/rules/no-var-requires.ts
JoshuaKGoldberg: #8186. I wanted to get this PR merged 😄 and I don't think template literals in `require()` calls are pretty rare.

PR #7712, opened 2023-10-01
fix(eslint-plugin): [prefer-string-starts-ends-with] only report slice/substring with correct range

### LINKED ISSUE #4181: [prefer-string-starts-ends-with] False positive for substring with length limit
- [x] I have [read the FAQ](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/TROUBLESHOOTING.md) and my problem is not listed.

**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/prefer-string-starts-ends-with": ["error"]
  }
}
```

```TS
export function isThreeCharPrefixMatch(str: string, candidate: string): boolean {
  return str.substring(0, 3) === candidate;
}
```

`tsconfig`: the default as generated by `tsc --init`
**Expected Result**
No error, since a `substring` call other than `substring(0)` can not be replaced with `startsWith` alone.

**Actual Result**
The following error was printed:
```
2:10  error  Use 'String#startsWith' method instead  @typescript-eslint/prefer-string-starts-ends-with
```

**Additional Info**
Consider the code with the change suggested by the linter:
```TS
export function isThreeCharPrefixMatchB(str: string, candidate: string): boolean {
  return str.startsWith(candidate);
}
```
Now
```
> isThreeCharPrefixMatch("int32", "in");
false
> isThreeCharPrefixMatchB("int32", "in");
true
```
**Versions**

| `@typescript-eslint/eslint-plugin` | `5.4.0` |
| `@typescript-eslint/parser`        | `5.4.0` |
| `TypeScript`                       | `4.5.2` |
| `ESLint`                           | `8.2.0` |
| `node`                             | `12.22.5` |

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4181

## Overview

Avoids reporting `slice`/`substring` where the sliced range does not match the expected substring's. This may lead to false negatives, but is "safe" in the general sense.

### DIFF
--- a/packages/eslint-plugin/src/rules/prefer-string-starts-ends-with.ts
+++ b/packages/eslint-plugin/src/rules/prefer-string-starts-ends-with.ts
@@ -161,23 +161,22 @@ export default createRule({
     }
 
     /**
-     * Check if a given node is a negative index expression
-     *
-     * E.g. `s.slice(- <expr>)`, `s.substring(s.length - <expr>)`
-     *
-     * @param node The node to check.
-     * @param expectedIndexedNode The node which is expected as the receiver of index expression.
+     * Returns true if `node` is `-substring.length` or
+     * `parentString.length - substring.length`
      */
-    function isNegativeIndexExpression(
+    function isLengthAheadOfEnd(
       node: TSESTree.Node,
-      expectedIndexedNode: TSESTree.Node,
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (approved):
Tricky! Nicely done 👌

PR #7752, opened 2023-10-15
feat(typescript-estree): add allowDefaultProjectForFiles project service allowlist option

### LINKED ISSUE #7435: Bug: `EXPERIMENTAL_useProjectService: true` throws unclear error if file is not included in a program

### Relevant Package

parser

### Playground Link

_No response_

### Repro

To the root of a project that is using `EXPERIMENTAL_useProjectService: true` recently I added `rollup.config.js`. That file was not added to tsconfig's `include` list. ESLint throws: 

> Error while loading rule '@typescript-eslint/consistent-type-exports': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.

By the way, it was so exciting to see the `EXPERIMENTAL_useProjectService` option added. The project I talk about is rather experimental so it was perfect playground to tryout the `ProjectService`.

### ESLint Config

```javascript
module.exports = {
  parserOptions: {
    "EXPERIMENTAL_useProjectService": true
  }
};
```

### tsconfig

```jsonc
{
  "include": ["./src"]
}
```

### Expected Result

Perhaps the error could be more useful. If this should be an error at all. Not sure what is going on.

### Actual Result

 Looking at the error, I thought that maybe the option was removed or stoped working.

### Additional Info

_No response_

### Versions

| `@typescript-eslint/parser`            | `6.2.1` |

### LINKED ISSUE #7871: Enhancement: Allow project service to provide type information for out-of-project files

### Relevant Package

typescript-estree

### Description

Splitting out of #7435: that issue indicates that the error message for trying to generate type information for a file not recognized by the `tsconfig.json` is not ideal. But as we see in #7752, if we don't check to make sure that the file is recognized... type information still works! This issue tracks: can we allow "unknown" (not included in the TSConfig) files to have type info generated by the project service?

Not needing to explicitly include files in the TSConfig would be ideal from the perspective of simple configs. It'd mean we can get rid of `tsconfig.eslint.json`-style strategies of having a different TSConfig for your linted code. Those are commonly used to enable `allowJs` for root-level config files (e.g. `eslint.config.js`) not included in the project's TS-only `tsconfig.json`.

We can't enable that kind of "config-less typed JS linting" with traditional `project: true` styles of approaches because of the following combination of tech restrictions:

1. The _default program_ in TypeScript is a "builder": each time you add a new file to it, it recreates & recomputes a large swathe of nodes and types.
    * See also https://github.com/typescript-eslint/typescript-eslint/issues/7761#issuecomment-1769504267 for points on misconfiguration.
3. ESLint does not provide us with APIs around when files are added, updated, or removed. We're restricted to just setting everything up once in the parser. See https://github.com/eslint/rfcs/pull/102.
4. As a result, what was renamed to `DEPRECATED__createDefaultProgram` in v6 is known to cause significant slowdowns in non-trivial uses

If we can use the project service in a way that doesn't suffer from the same drawback, then that'd be a massive win for users.

### Additional Info

Shoutout+thanks as always @jakebailey for helping us with the TypeScript side!

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7871

## Overview

Removes the check that `projectService.openClientFile` should return an opened file. In my testing, it looks like we can still generate type info for the file just with the default program for it?

Standalone repro here: ~https://github.com/JoshuaKGoldberg/repros/tree/project-service-without-opened-file~ https://github.com/JoshuaKGoldberg/repros/tree/allow-default-project-for-files

Asked for confirmation about this approach on the TypeScript Discord: https://discord.com/channels/508357248330760243/640177429775777792/1163133940115197952. Edit: confirmed, and filed #7761 as a followup ✅


packages/typescript-estree/src/useProgramFromProjectService.ts
JoshuaKGoldberg, line 1: cc @mrazauskas @probablyup from #7435: I'd love to know, does this fix actually fix the issue for you? I have a reproduction in https://github.com/JoshuaKGoldberg/repros/tree/project-service-without-opened-file showing it working on my end. https://github.com/mrazauskas/x-use-project-service also works for me now.

mrazauskas:
Thanks for working on this.

I wonder, should it perhaps throw an error or at least log a warning is the default TSConfig is used? For instance, in my case there was no intention to use the default config. Seeing a warning or error would make me either to include the file into the program or to exclude it from lint.

Just thinking out loud. Can be I misunderstood the fix.

JoshuaKGoldberg:
Hmm yeah it's an interesting idea. I don't think we'd want to warn by default, because one of the primary motivations for the option is to allow unlisted files to be linted (e.g. `.eslintrc.cjs`). But I can definitely see why you'd want that. I'll wait on getting explicit feedback around the approach from the TypeScript team (asked in the TS Discord & linked in the OP) before maybe filing an issue or discussion.

jakebailey (approved):
Theoretically looks correct to me; the only actual change is not returning early.

JoshuaKGoldberg:
Excellent. I'll hold off on considering this mergeable until the conversation in #7761 is resolved. Just in case we end up radically overhauling everything.

JoshuaKGoldberg:
Re-targeting this to specifically focus on #7871.

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -9,7 +13,15 @@ const createStubFileWatcher = (): ts.FileWatcher => ({
 
 export type TypeScriptProjectService = ts.server.ProjectService;
 
-export function createProjectService(): TypeScriptProjectService {
+export interface ProjectServiceSettings {
+  allowDefaultProjectForFiles: ReadonlySet<string>;
```
JoshuaKGoldberg, line 17: This name is not particularly good but it's under an experimental option so I'd expect we'll rethink all names for v7.

packages/typescript-estree/src/parser-options.ts
```diff
@@ -88,6 +88,16 @@ interface ParseOptions {
...
 
+/**
+ * Granular options to configure the project service.
+ */
+export interface ProjectServiceOptions {
+  /**
+   * Files to allow running with the default inferred project settings.
+   */
+  allowDefaultProjectForFiles?: string[];
+}
```
bradzacher, line 99: don't forget to document this in `docs/packages/TypeScript_ESTree.mdx`

bradzacher: also would be worth explicitly documenting that this accepts a glob

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -46,5 +58,16 @@ export function createProjectService(): TypeScriptProjectService {
     },
     session: undefined,
   });
+
+  return {
+    allowDefaultProjectForFiles: new Set(
+      typeof options === 'object'
+        ? options.allowDefaultProjectForFiles
+            ?.flatMap(pattern => globby.sync(pattern))
+            .map(filePath => path.join(tsconfigRootDir, filePath))
```
bradzacher, line 67: this should probably use the `getCanonicalFileName` util to ensure it's in the standard form?

packages/typescript-estree/src/useProgramFromProjectService.ts
```diff
@@ -1,26 +1,38 @@
...
     parseSettings.codeFullText,
     /* scriptKind */ undefined,
     parseSettings.tsconfigRootDir,
   );
-  if (!opened.configFileName) {
-    return undefined;
+
+  if (hasFullTypeInformation) {
+    if (opened.configFileName) {
+      if (allowDefaultProjectForFiles.has(parseSettings.filePath)) {
```
bradzacher, line 22: ditto here - probably worth passing the path through `getCanonicalFileName` to ensure we get a match.

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -46,5 +58,16 @@ export function createProjectService(): TypeScriptProjectService {
     },
     session: undefined,
   });
+
+  return {
+    allowDefaultProjectForFiles: new Set(
+      typeof options === 'object'
+        ? options.allowDefaultProjectForFiles
+            ?.flatMap(pattern => globby.sync(pattern))
```
bradzacher, line 66: hmmmm I wonder if we could do this check after-the-fact.
EG like store the glob(s) themselves and then if a file fails to be found - then compare the glob with the file path to look for a match.

that way we never ever resolve the globs to disk paths - saving time and ensuring we don't fall out-of-sync with disk state.

one weakness of the current approach is that new files (eg if you add a new `.eslintrc.js` in your codebase) would fail the IDE match and the user would need to restart their IDE to get it working.

packages/typescript-estree/src/useProgramFromProjectService.ts
```diff
@@ -34,6 +46,6 @@ export function useProgramFromProjectService(
   function absolutify(filePath: string): string {
     return path.isAbsolute(filePath)
       ? filePath
-      : path.join(projectService.host.getCurrentDirectory(), filePath);
+      : path.join(service.host.getCurrentDirectory(), filePath);
   }
```
bradzacher, line 50: i realise this is late - but maybe `ensureAbsolutePath` would be better here, eg `ensureAbsolutePath(filePath, service.host.getCurrentDirectory())` just so we're reusing things.

packages/typescript-estree/src/create-program/createProjectService.ts
JoshuaKGoldberg: I like the lazy-loading idea for the file-not-found case a lot, but we also need the resolved globs in the case of the file-found case:

```ts
if (allowDefaultProjectForFiles.has(filePath)) {
  throw new Error(
    `${filePath} was included by allowDefaultProjectForFiles but also was found in the project service. Consider removing it from allowDefaultProjectForFiles.`,
  );
}
```

Given we already break down in the case of files changing, is there much benefit to lazy-globbing here?

bradzacher: do we need to resolve them, though?
Shouldn't we be able to convert that check instead to
```
if (file matches globs) {
  throw new Error();
}
```

a library like minimatch would allow us to handle this without any disk ops

bradzacher (approved):
approving - pending that last convo being resolved

packages/typescript-estree/src/create-program/createProjectService.ts
JoshuaKGoldberg: Hmm, I was thinking there may be edge cases around files existing on disk but not being included in the TSConfig - or some inverted version of that... But that doesn't seem very likely. And I suppose we can cross that bridge if we get to it.

JoshuaKGoldberg: Amusingly, `minimatch` was already many times over a sub-dependency. No significant change to the lockfile.

PR #7808, opened 2023-10-20
fix(eslint-plugin): [prefer-literal-enum-member] allows self-referential enums with allowBitwiseExpressions

### LINKED ISSUE #7763: Bug: [prefer-literal-enum-member] allowBitwiseExpressions with self-referential enums does not disable error

### Playground Link

https://typescript-eslint.io/play/#ts=5.2.2&fileType=.ts&code=PTAEDMCcHsFtQBYBckAcDOAuEB3PA6JAT1QFN0BjSAS1SQBsBDAOwHN9pJXgATaC9MAQseAI2jQA1sFLMArrHT5ksegChZC0ADFq9UgEEKFculABvNaFAA5aM1IAaK6ABKpRj1ABeUAEZQAB5A-2drAHUaJFIffyCQgCYwtw8eSOpo2PdPUAAfUHTo5wBfIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wlsfy8AhvHpNYAW1oTEEgEa90YANpRe0APbRIAGkxQR8DQHcAQpXzHKyRAFEAHhxTJKGpqgz44iMAF8wALrgfiC%2BQA&tsconfig=&tokens=false

### Repro Code

```typescript
// from https://www.typescriptlang.org/docs/handbook/enums.html
enum FileAccess {
  None,
  Read = 1 << 1,
  Write = 1 << 2,
  ReadWrite = Read | Write,
}
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/prefer-literal-enum-member": [
      "error", { "allowBitwiseExpressions": true }
    ]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

There should be no reported error. This is a fairly normal use-case for enums and bitwise operators. The official ts docs even have it as an example.

I would like a way to only allow literal enums, but also allow bitwise in this use case. I can get the code to pass by changing it to, but I feel it makes things less clear:

```
enum FileAccess {
  None,
  Read = 1 << 1, // 0b0010
  Write = 1 << 2, // 0b0100
  ReadWrite = 0b0010 | 0b0100,
}
```

### Actual Result

Recieve error:

> @typescript-eslint/prefer-literal-enum-member:
> Explicit enum value must only be a literal value (string, number, boolean, etc). 6:3 - 6:12

### Additional Info

In the actual code I an writing, I am extending "plugin:@typescript-eslint/strict-type-checked"

jbs-marcus:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7763

## Overview

Allows enums with binary bitwise expressions if both operands are literals or reference members of own enum

### DIFF
--- a/packages/eslint-plugin/src/rules/prefer-literal-enum-member.ts
+++ b/packages/eslint-plugin/src/rules/prefer-literal-enum-member.ts
@@ -59,16 +59,44 @@ export default createRule({
           return;
         }
 
+        // Allow bitwise operators when both operands are literals or refer to another member on this enum
         if (
           allowBitwiseExpressions &&
           node.initializer.type === AST_NODE_TYPES.BinaryExpression &&
           ['|', '&', '^', '<<', '>>', '>>>'].includes(
             node.initializer.operator,
           ) &&
-          node.initializer.left.type === AST_NODE_TYPES.Literal &&
-          node.initializer.right.type === AST_NODE_TYPES.Literal
+          node.parent.type === AST_NODE_TYPES.TSEnumDeclaration
         ) {
-          return;
+          const leftIsLiteral =
+            node.initializer.left.type === AST_NODE_TYPES.Literal;
+          const rightIsLiteral =
+            node.initializer.right.type === AST_NODE_TYPES.Literal;
+
+          // early branch, so gathering member identifiers only happens when needed
+          if (leftIsLiteral && rightIsLiteral) {
+            return;
+          }
+
+          const members = node.parent.members.reduce((names, member) => {
+            if ('name' in member.id) names.push(member.id.name);
+            return names;
+          }, new Array<string>());
+
+          const leftIsValidIdentifier =
+            node.initializer.left.type === AST_NODE_TYPES.Identifier &&
+            members.includes(node.initializer.left.name);
+
+          const rightIsValidIdentifier =
+            node.initializer.right.type === AST_NODE_TYPES.Identifier &&
+            members.includes(node.initializer.right.name);
+
+          if (
+            (leftIsLiteral || leftIsValidIdentifier) &&
+            (rightIsLiteral || rightIsValidIdentifier)
+          ) {
+            return;
+          }
         }
 
         context.report({
--- a/packages/eslint-plugin/src/rules/prefer-literal-enum-member.ts
+++ b/packages/eslint-plugin/src/rules/prefer-literal-enum-member.ts
@@ -73,7 +73,6 @@ export default createRule({
           const rightIsLiteral =
             node.initializer.right.type === AST_NODE_TYPES.Literal;
 
-          // early branch, so gathering member identifiers only happens when needed
           if (leftIsLiteral && rightIsLiteral) {
             return;
           }
--- a/packages/eslint-plugin/src/rules/prefer-literal-enum-member.ts
+++ b/packages/eslint-plugin/src/rules/prefer-literal-enum-member.ts
@@ -78,12 +78,15 @@ export default createRule({
             return;
           }
 
-          const members = node.parent.members.reduce((names, member) => {
-            if ('name' in member.id) {
-              names.push(member.id.name);
-            }
-            return names;
-          }, new Array<string>());
+          const members = node.parent.members.reduce<string[]>(
+            (names, member) => {
+              if ('name' in member.id) {
+                names.push(member.id.name);
+              }
+              return names;
+            },
+            [],
+          );
 
           const leftIsValidIdentifier =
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/prefer-literal-enum-member.ts
```diff
@@ -59,16 +59,46 @@ export default createRule({
...
+          const rightIsLiteral =
+            node.initializer.right.type === AST_NODE_TYPES.Literal;
+
+          // early branch, so gathering member identifiers only happens when needed
+          if (leftIsLiteral && rightIsLiteral) {
+            return;
+          }
+
+          const members = node.parent.members.reduce((names, member) => {
+            if ('name' in member.id) {
```
JoshuaKGoldberg, line 82: [Code] We generally try not to use `'in'` checks or similar for AST code. If a property (here, `id`) doesn't exist it's because the node legitimately might be a type that doesn't have it. See test cases across the enum-related rules for property names based on literals:

```ts
enum Example {
  ['abc'] = 0,
}
```

The code here should be able to handle names like that, including test coverage.

packages/eslint-plugin/src/rules/prefer-literal-enum-member.ts
```diff
@@ -59,16 +59,46 @@ export default createRule({
...
+              names.push(member.id.name);
+            }
+            return names;
+          }, new Array<string>());
```
JoshuaKGoldberg, line 86: [Style] Heh, I'm surprised this didn't trigger any enabled lint rules. `.reduce<string[]>( ..., [])` would be our preferred syntax here.

packages/eslint-plugin/src/rules/prefer-literal-enum-member.ts
```diff
@@ -59,16 +59,46 @@ export default createRule({
...
-          node.initializer.right.type === AST_NODE_TYPES.Literal
+          node.parent.type === AST_NODE_TYPES.TSEnumDeclaration
         ) {
-          return;
+          const leftIsLiteral =
+            node.initializer.left.type === AST_NODE_TYPES.Literal;
+          const rightIsLiteral =
+            node.initializer.right.type === AST_NODE_TYPES.Literal;
+
+          // early branch, so gathering member identifiers only happens when needed
```
JoshuaKGoldberg, line 76: [Proofreading] I like the previous comment, but this one doesn't really help with anything IMO. Early returns like this one are a common pattern.

```suggestion
```

packages/eslint-plugin/tests/rules/prefer-literal-enum-member.test.ts
```diff
@@ -73,6 +73,8 @@ enum Foo {
   E = 1 & 0,
   F = 1 ^ 0,
   G = ~1,
+  E = G | 1,
+  F = G | E,
```
JoshuaKGoldberg, line 77: [Testing] Let's add a new test case for each of these two added cases: https://typescript-eslint.io/contributing/local-development#granular-unit-tests

Here and below.

JoshuaKGoldberg (changes requested):
A good start, thanks! 🚀

jbs-marcus:
> A good start, thanks! 🚀

Okay, thank you! I will try to resolve these soon

packages/eslint-plugin/tests/rules/prefer-literal-enum-member.test.ts
jbs-marcus: I'm sorry, I don't think I understand what is being asked here

JoshuaKGoldberg: Ah sorry - what I mean to say is:

1. Each of the entries in the `invalid` and `valid` is its own "unit test"
2. In theory, we like having each unit test check for a specific behavior
3. The added code is testing a newly added behavior
4. Therefore, the added code should be added in a new unit test (new entries in the `invalid` or `valid` arrays)

The existing code in our repo doesn't always respect the one-behavior-per-unit-test preference... but new code should ideally.

Does that make sense?

JoshuaKGoldberg:
👋 Ping @jbs-marcus, is this still something you have time and interest for?

JoshuaKGoldberg:
Closing this PR as it's been stale for a while without activity. Feel free to reopen @jbs-marcus if you have time - but no worries if not!

If anybody wants to drive it forward, please do post your own PR - and if you use this as a start, consider adding a co-author attribution at the end of your PR description. Thanks! 😊

PR #7812, opened 2023-10-20
feat(utils): update types to reflect `RuleContext` and `SourceCode` changes and deprecations

G-Rath:
## PR Checklist

- [x] ~Addresses an existing open issue: fixes #000~
- [x] ~That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)~

## Overview

Reference: https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9


G-Rath:
Are folks ok if I ignore the deprecation linting errors? I can help getting them addressed but would prefer to do so in dedicated PR since they'll require actual runtime changes and there's a few different ways they can be approached

JoshuaKGoldberg:
Well, we need typescript-eslint to work on both v8 and v9 for the time being. Dropping v8 would be a breaking change. I'd presume `eslint-disable` comments would be the way to go.

G-Rath:
My point is you'll need to make changes to support v9 since some of these are being removed in that major - it's pretty straight forward ([I've already done this in `eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest/pull/1453)) but I'd still prefer to do it in a dedicated PR because there are two different changes here (adding types, moving off deprecated functions and properties) & how I've done it for `eslint-plugin-jest` might not be how you want to do it here

I'll add disable directives for now

JoshuaKGoldberg:
Agreed. https://github.com/jest-community/eslint-plugin-jest/blob/3c12c9842df7950dd214855fa69b1785c4d1574d/src/rules/utils/misc.ts is roughly the strategy I'm betting we'll want to take. 🙂

JoshuaKGoldberg (approved):
💯 thanks!

JoshuaKGoldberg:
Adding `1 approval` label for now because this I'd love someone else from @typescript-eslint/triage-team to take a look at this area code. Both to raise visibility, and because I'm not 100% confident I'm not missing something. A gut hunch.

JuanSebastianDiazVera (approved):

packages/utils/src/ts-eslint/SourceCode.ts
```diff
@@ -299,6 +299,31 @@ declare class SourceCodeBase extends TokenStore {
...
+   * This information can be used to track references to variables.
+   * @since 8.38.0
+   */
+  getDeclaredVariables(node: TSESTree.Node): readonly Scope.Variable[];
+  /**
+   * Marks a variable with the given name in the current scope as used.
+   * This affects the no-unused-vars rule.
+   * @since 8.39.0
+   */
+  markVariableAsUsed(name: string, node: TSESTree.Node): boolean;
```
bradzacher, line 326: With these brand new methods I think it's probably best to mark them as optional so  that people have to do version checks to use them.

it's going to be a rare, rare, rare case that a plugin specifically supports `>= 8.39.0` and can only use these methods. ESLint core rules are the only case.

It's going to be an absolute dog's breakfast upgrading to support v9 when they remove the old ones :sigh:

bradzacher (changes requested):

bradzacher (approved):

JoshuaKGoldberg:
Thanks again @G-Rath!

JoshuaKGoldberg:
Amusingly, there are now a bunch of lint failures on `main`: #7913. But this PR had a passing ✅. So nobody should feel bad 😄

PR #7819, opened 2023-10-22
chore: cancel concurrent ci workflow runs

StyleShit:
This might help with freeing up some machines of the free github actions tier


JoshuaKGoldberg (approved):
Right! I forgot this was an option 😄 good idea. Thanks!

Billions guy saying 'that's a good point'

PR #7822, opened 2023-10-23
feat(eslint-plugin): [no-misused-spread] add new rule `no-misused-spread`

### LINKED ISSUE #748: Rule proposal: prevent array, iterable and function spreads into objects
**Repro**

```TS
// Array spread
const squares = [1, 4, 9, 16];
const cubes = [1, 8, 27];
const bothArray = { ...squares, ...cubes };
// rule would error here, as spreading an array causes surprising results
bothArray === {0: 1, 1: 8, 2: 27, 3: 16}
// TS: typeof bothArray === (typeof Array<number>) ❌

// Iterable spread
const map = new Map([[1,2], [3,4]]);
const set = new Set([5,6,7,8]);
const bothIterable = { ...map, ...set };
// rule would error here as spreading an iterable does nothing
bothIterable === {}
// TS: typeof allFunc === (typeof Set<number> & typeof Map<number, number>) ❌

// Function spread
function funcDecl() {}
const funcExpr = function () {};
const arrowFunc = () => {};
const allFunc = { ...funcDecl, ...funcExpr, ...arrowFunc };
// rule would error here as spreading a function does nothing
allFunc === {};
// TS: typeof allFunc === ({}) ✅

// Object with call-signature
interface FuncWithProps {
  property?: string;
  (): number;
}
const funcWithProps: FuncWithProps = () => 1;
funcWithProps.property = 'foo';
const spreadFuncWithProps = {...funcWithProps};
// rule would NOT error here as custom function props are spread as expected
spreadFuncWithProps === {property: 'foo'};
// TS: typeof allFunc === (FuncWithProps) ✅
```

TypeScript does not have an understanding of how iterable object spreads work, and borks the types.
This leads to unsoundness and broken code.

TypeScript does understand function object spreads, but they are noops, so they shouldn't be allowed.

**Additional Information**
Detecting this should be a matter of detecting whether the type of each expression being spread within an _object_ implements `Iterable<T>`, defined within `lib.es2015.iterable.d.ts`.

The converse rule of preventing objects being spread within an array seems unnecessary/out-of-scope, given that the TypeScript type system is able to determine that each expression spread within an array is iterable.

lcharlois-neotys:
## PR Checklist

## Overview

Partially addresses #748.
Add a new rule that prevent spreading a function.


Josh-Cena (changes requested):
This rule's scope is too narrow. We should probably rename it to something like `no-misused-spread`.

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
StyleShit, line 1: maybe also add tests for spreading `Map`, `Set`, and a class instance?

StyleShit (changes requested):

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,47 @@
...
+  defaultOptions: [],
+  name: 'no-misused-spread',
+  meta: {
+    docs: {
+      description: 'Disallow spread operator on function',
+      recommended: 'stylistic',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      forbidden:
```
StyleShit, line 19: in case we'll expand this rule in the future, maybe change the message id?
(just something from top of my head, not sure it's a good one though)

```suggestion
      forbiddenFunctionSpread:
```

StyleShit (changes requested):

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,47 @@
...
+} from '@typescript-eslint/utils';
+
+import { createRule } from '../util';
+
+export default createRule({
+  defaultOptions: [],
+  name: 'no-misused-spread',
+  meta: {
+    docs: {
+      description: 'Disallow spread operator on function',
```
StyleShit, line 14: should probably be updated to a more generic "no-spread" description

StyleShit (changes requested):

StyleShit:
If you've used some of @OliverJAsh's code from [this comment](https://github.com/typescript-eslint/typescript-eslint/issues/748#issuecomment-1450102392), consider adding them as a co-author 😄

lcharlois-neotys:
> If you've used some of @OliverJAsh's code from [this comment](https://github.com/typescript-eslint/typescript-eslint/issues/748#issuecomment-1450102392), consider adding him as a co-author 😄

I will be happy. Could you tell me how to add co-author?
Is adding something in the PR description enough?

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,47 @@
...
+      recommended: 'stylistic',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      forbidden:
+        'Spreading a function is almost always a mistake. Did you forgot to call the function?',
+    },
+    schema: [],
+    type: 'suggestion',
```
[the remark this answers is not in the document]
lcharlois-neotys: IMO this is most of the time a problem, but untill today there is nothing warning the user that it is one.
So I guess this is an expected usage and I choose `suggestion` rather than `problem`.
I can change it if needed.

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    `,
+    `
+const a = () => ({ value: 33 });
+const b = ({ value: number }) => ({ value: value });
+b({ ...a() });
+    `,
+    `
+const a = () => ({ value: 33 });
+const b = ({ value: number }) => ({ value: value });
+    `,
```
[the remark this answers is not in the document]
lcharlois-neotys: I think it is a copy/paste from the previous in order to test another case, but honestly I don't remember since last time I worke on this PR.
I remove it.

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { createRule } from '../util';
+
+export default createRule({
+  defaultOptions: [],
+  name: 'no-misused-spread',
+  meta: {
+    docs: {
+      description: 'Disallow spread operator on function',
+      recommended: 'stylistic',
```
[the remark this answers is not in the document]
lcharlois-neotys: What will be your recommendation?
Should it be `strict`?

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
lcharlois-neotys: Do you mean in the valid part?

packages/eslint-plugin/src/rules/no-misused-spread.ts
lcharlois-neotys: I've proposed a change.

packages/eslint-plugin/src/rules/no-misused-spread.ts
lcharlois-neotys: It makes sense.

packages/eslint-plugin/src/rules/no-misused-spread.ts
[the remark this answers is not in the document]
StyleShit: Not sure... 
let's wait for @JoshuaKGoldberg I guess?

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
StyleShit: yup

packages/eslint-plugin/src/configs/all.ts
```diff
@@ -97,6 +97,7 @@ export = {
     '@typescript-eslint/no-misused-new': 'error',
     '@typescript-eslint/no-misused-promises': 'error',
     '@typescript-eslint/no-mixed-enums': 'error',
+    '@typescript-eslint/no-misused-spread': 'error',
```
StyleShit, line 100: [Nit]
This line should probably need to be one line above

StyleShit (commented):
looks good to me overall except for a small nitpicking, and maybe add the tests we've discussed earlier

anyway, I'm not a maintainer, just stumbled across this PR, so let's wait for one 😄

packages/eslint-plugin/src/rules/no-misused-spread.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: https://typescript-eslint.io/linting/configs - this is a functional, bug-catching rule so it wouldn't be `'stylistic'`. It would be `'strict'` for now to avoid a breaking change. In each new next major version we look at which rules might be promoted to `'recommended'`.

packages/eslint-plugin/src/rules/no-misused-spread.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: It's a `'problem'`. Anything that catches directly incorrect _logic_ is a problem.

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,46 @@
...
+
+type MessageIds = 'forbiddenFunctionSpread';
+
+export default createRule<[], MessageIds>({
+  defaultOptions: [],
+  name: 'no-misused-spread',
+  meta: {
+    docs: {
+      description:
+        "Disallow spread operator that shouldn't be spread most of the time",
```
JoshuaKGoldberg, line 13: [Proofreading] "most of the time" is kind of an imprecise, vague statement. Users may not understand it.

Instead, let's speak to what this rule is specifically trying to avoid: spreads that don't do anything.

```suggestion
        "Disallow using the spread operator on values that can't be spread",
```

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,46 @@
...
+    },
+    messages: {
+      forbiddenFunctionSpread:
+        'Spreading a function is almost always a mistake. Did you forget to call the function?',
+    },
+    schema: [],
+    type: 'suggestion',
+  },
+  create: (context): TSESLint.RuleListener => {
+    const listener = (node: TSESTree.SpreadElement): void => {
```
JoshuaKGoldberg, line 25: [Style] Is there a reason you'd wanted to split this out instead of having it inline at the bottom of the function? Inline would mean an inferrable type for `node`:

```ts
return {
  SpreadElement(node): void {
    // ...
  }
};
```

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,46 @@
...
+      const services = getParserServices(context);
+      const checker = services.program.getTypeChecker();
+
+      const tsNode = services.esTreeNodeToTSNodeMap.get(node.argument);
+      const type = checker.getTypeAtLocation(tsNode);
```
JoshuaKGoldberg, line 30: [Cleanup] We have a new thing for this 😄 https://typescript-eslint.io/blog/announcing-typescript-eslint-v6#type-checker-wrapper-apis

```suggestion
      const type = services.getTypeAtLocation(node.argument);
```

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,46 @@
...
+  meta: {
+    docs: {
+      description:
+        "Disallow spread operator that shouldn't be spread most of the time",
+      recommended: 'stylistic',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      forbiddenFunctionSpread:
+        'Spreading a function is almost always a mistake. Did you forget to call the function?',
```
JoshuaKGoldberg, line 19: [Proofreading] Maybe...

```suggestion
        'Spreading a function with no custom properties does nothing. Did you forget to call the function?',
```

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,46 @@
...
+  name: 'no-misused-spread',
+  meta: {
+    docs: {
+      description:
+        "Disallow spread operator that shouldn't be spread most of the time",
+      recommended: 'stylistic',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      forbiddenFunctionSpread:
```
JoshuaKGoldberg, line 18: [Feature] If we add the rule as-is (only looking at functions), it'd be a breaking change to later add in Maps, Sets, etc. I think the rule should include all of what's requested in the issue to start.

LMK if any of them give you grief, happy to help!

JoshuaKGoldberg (changes requested):
Looking good! Requesting expanding and a few small cleanups, but this is a great start 🙂

JoshuaKGoldberg:
👋 Ping @lcharlois-neotys, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

JoshuaKGoldberg:
Closing this PR as it's been stale for a while without activity. cc @lcharlois-neotys - if you want to keep working on it, that'd be swell. But no worries if you don't have time. If anybody wants to drive it forward, please do post your own PR - and if you use this as a start, consider adding a co-author attribution at the end of your PR description. Thanks! 😊

Issue #7338, opened 2023-07-27
Repo: Enable prefer-string-starts-with internally


JoshuaKGoldberg:
### Suggestion

Our `.eslintrc.js` generally has a line that looks something like:

https://github.com/typescript-eslint/typescript-eslint/blob/593d37fe0cf5e23c07611d13537dc8bab10f1c5b/.eslintrc.js#L56-L58

Filing this tracking issue to investigate enabling them. For each, we'll want to:

1. Try enabling it in a commit/branch/PR
2. Fix up any legitimate complaints it points out
3. If it points too many things incorrectly, either:
    * File bug/feature issues on the rule to help it improve
    * Keep it disabled altogether

JoshuaKGoldberg:
At this point, it's just two rules left:

https://github.com/typescript-eslint/typescript-eslint/blob/5b78f2240b0f18e1fb745532b8cdd36cc4c7d6ae/eslint.config.mjs#L106-L108

`no-confusing-void-expression` is blocked & tracked by #8538. Retargeting this issue for `prefer-string-starts-with`.

PR #7835, opened 2023-10-25
chore(typescript-estree): fixed no-unnecessary-condition complaints

JoshuaKGoldberg:
## PR Checklist

## Overview

Makes progress on fixing existing [`@typescript-eslint/no-unnecessary-condition`](https://typescript-eslint.io/rules/no-unnecessary-condition) violations under `packages/typescript-estree`. Part of a series of PRs that split this up, as there are many of them.


packages/typescript-estree/src/parseSettings/warnAboutTSVersion.ts
```diff
@@ -25,8 +25,7 @@ let warnedAboutTSVersion = false;
 
 export function warnAboutTSVersion(parseSettings: ParseSettings): void {
   if (!isRunningSupportedTypeScriptVersion && !warnedAboutTSVersion) {
-    const isTTY =
-      typeof process === 'undefined' ? false : process.stdout?.isTTY;
+    const isTTY = typeof process === 'undefined' ? false : process.stdout.isTTY;
```
bradzacher, line 28: IIRC this is actually necessary
https://github.com/typescript-eslint/typescript-eslint/pull/1619

JoshuaKGoldberg: 🤔 I don't like adding support for something that's not possible per type definitions... `process.stdout` is supposed to exist. If some bundler removes it then they're telling a lie to packages like us. Hmm. #7896

bradzacher: Yeah generally I'd say "I don't care" but the reason we accepted it was because it's prettier and they're one of our biggest consumers.

packages/typescript-estree/src/node-utils.ts
```diff
@@ -424,7 +424,7 @@ export function findFirstMatchingAncestor(
   node: ts.Node,
   predicate: (node: ts.Node) => boolean,
 ): ts.Node | undefined {
-  while (node) {
+  while (node as ts.Node | undefined) {
```
Josh-Cena, line 427: This will eventually cause a `no-param-reassign` error if we have that rule. Can we use another variable instead?

packages/typescript-estree/src/parseSettings/warnAboutTSVersion.ts
JoshuaKGoldberg: I do ❤️ Prettier 😄

PR #7836, opened 2023-10-25
chore(website): fixed most no-unnecessary-condition complaints

JoshuaKGoldberg:
## PR Checklist

## Overview

Makes progress on fixing existing [`@typescript-eslint/no-unnecessary-condition`](https://typescript-eslint.io/rules/no-unnecessary-condition) violations under `packages/website`. Part of a series of PRs that split this up, as there are many of them.


bradzacher (approved):

packages/website/src/components/editor/LoadedEditor.tsx
```diff
@@ -267,7 +267,7 @@ export const LoadedEditor: React.FC<LoadedEditorProps> = ({
     return debounce(() => editor.layout(), 1);
   }, [editor]);
 
-  const container = editor.getContainerDomNode?.() ?? editor.getDomNode();
+  const container = editor.getContainerDomNode();
```
armano2, line 270: this is not correct, as this part of code is neccesary for some older versions of playground (older version of ts)

packages/website/src/components/lib/jsonSchema.ts
```diff
@@ -50,7 +50,7 @@ export function getRuleJsonSchemaWithErrorLevel(
       };
     }
     // example: naming-convention rule
-    if (typeof ruleSchema.items === 'object' && ruleSchema.items) {
+    if (typeof ruleSchema.items === 'object') {
```
armano2, line 53:
```
typeof null === 'object'
```

packages/website/src/components/typeDetails/TypeInfo.tsx
```diff
@@ -67,7 +67,7 @@ export function TypeInfo({
   onHoverNode,
 }: TypeInfoProps): React.JSX.Element {
   const computed = useMemo(() => {
-    if (!typeChecker || !value) {
+    if (!typeChecker) {
```
armano2, line 70: this is not correct

packages/website/src/components/typeDetails/TypesDetails.tsx
```diff
@@ -48,15 +48,13 @@ export function TypesDetails({
           value={value}
         />
       </div>
-      {selectedNode && (
```
armano2, line 51: `selectedNode` may be undefined

packages/website/src/components/lib/jsonSchema.ts
bradzacher: I think in the context of the types it's probably defined as `items?: T[]` - hence the `typeof` is enough to refine the types correctly?
This was a strictness change we added in v6.

packages/website/src/components/typeDetails/TypeInfo.tsx
bradzacher: @armano2 according to the types it is - the property `value` is defined as `readonly value: Node` and it's never reassigned in the component body

packages/website/src/components/typeDetails/TypesDetails.tsx
bradzacher: @armano2 based on the types it can't be.
the initial value passed to `useState` is typed as `readonly value: Node`.
Then the only call to `setSelectedNode` is guarded by `if (item.node) { setSelectedNode(item.node) }`

packages/website/src/components/editor/LoadedEditor.tsx
JoshuaKGoldberg: 🤔 I'm confused, where are those types defined? The `editor.api.d.ts` I see has `getContainerDomNode(): HTMLElement;` - and I thought we're settled on just one editor version?

packages/website/src/components/lib/jsonSchema.ts
JoshuaKGoldberg: I don't see any `null` values for `items` in any ESLint or typescript-eslint rule. Is there a reason to support it?

PR #7849, opened 2023-10-27
feat(eslint-plugin): [ban-ts-comments] suggest ts-expect-error over ts-ignore

### LINKED ISSUE #7843: Enhancement: [ban-ts-comment] Autofix ts-ignore to ts-expect-error depending on options

https://typescript-eslint.io/rules/ban-ts-comment/

### Description

A fairly common use-case for `ban-ts-comment` is to ban @ts-ignore in favor of @ts-expect-error.

However, if developers are unfamiliar with @ts-expect-error, the error message doesn't help them find out about it.

The error message should recommend switching, and an autofix should be provided. If @ts-expect-error is also banned, the above original error message can be used instead.

I think it would be useful to autofix even if the rule is configured to require descriptions, as it still saves keystrokes. I don't know if autofixes are allowed to adjust code into a "fail" state however.

### Fail

```typescript
// @ts-ignore
```

### Pass

```typescript
// @ts-expect-error
// ^autofix
```

### Additional Info

_No response_

NotWoods:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7843

## Overview

Updates error message and adds a suggestion to replace ts-ignore with ts-expect-error, as long as the option for banning ts-expect-error is not set to true.

### DIFF
--- a/packages/eslint-plugin/src/rules/ban-ts-comment.ts
+++ b/packages/eslint-plugin/src/rules/ban-ts-comment.ts
@@ -1,4 +1,4 @@
-import { AST_TOKEN_TYPES } from '@typescript-eslint/utils';
+import { AST_TOKEN_TYPES, type TSESLint } from '@typescript-eslint/utils';
 
 import { createRule, getStringLength } from '../util';
 
@@ -19,8 +19,10 @@ export const defaultMinimumDescriptionLength = 3;
 
 type MessageIds =
   | 'tsDirectiveComment'
+  | 'tsIgnoreInsteadOfExpectError'
   | 'tsDirectiveCommentDescriptionNotMatchPattern'
-  | 'tsDirectiveCommentRequiresDescription';
+  | 'tsDirectiveCommentRequiresDescription'
+  | 'replaceTsIgnoreWithTsExpectError';
 
 export default createRule<[Options], MessageIds>({
   name: 'ban-ts-comment',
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/ban-ts-comment.ts
```diff
@@ -130,11 +137,37 @@ export default createRule<[Options], MessageIds>({
...
           const option = options[fullDirective];
           if (option === true) {
-            context.report({
-              data: { directive },
-              node: comment,
-              messageId: 'tsDirectiveComment',
-            });
+            if (directive === 'ignore' && options['ts-expect-error'] !== true) {
+              // Special case to suggest @ts-expect-error instead of @ts-ignore,
+              // as long as @ts-expect-error is banned outright.
```
JoshuaKGoldberg, line 142: I would consider `// @ts-expect-error` to still be superior over `// @ts-ignore` even if it's banned too. Sometimes I have to use it with an eslint-disable comment, but it's still preferable. I'd say let's always suggest it.

JoshuaKGoldberg (changes requested):
Looks great overall - and love the thorough tests! Just the one point, what do you think?

JoshuaKGoldberg (approved):
Cartoonish skeleton with an oversized glowing jack-o-lantern head, dancing in a nighttime cemetery

Issue #7867, opened 2023-11-02
Docs: [restrict-plus-operands] skipCompoundAssignments flag in the example is reversed


noshiro-pf:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

https://typescript-eslint.io/rules/restrict-plus-operands/#skipcompoundassignments

```diff
- Examples of code for this rule with { skipCompoundAssignments: true }:
+ Examples of code for this rule with { skipCompoundAssignments: false }:
```

This seems to have been forgotten when renaming in https://github.com/typescript-eslint/typescript-eslint/pull/7027.

### Affected URL(s)

https://typescript-eslint.io/rules/restrict-plus-operands/#skipcompoundassignments

PR #7868, opened 2023-11-02
docs: flip the flag of skipCompoundAssignments in the example

noshiro-pf:
Fixed https://github.com/typescript-eslint/typescript-eslint/issues/7867.


JoshuaKGoldberg (approved):
Nice catch, thank you! It took me a few moments to figure this out even with your explanation here 😄 very good.

PR #7880, opened 2023-11-04
feat(eslint-plugin): [switch-exhaustiveness-check] add requireDefaultForNonUnion option

### LINKED ISSUE #2959: [switch-exhaustiveness-check] Require default case if discriminant type is not a union

**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/switch-exhaustiveness-check": "warn"
  }
}
```

```TS
type Day = string; // or number

const day = 'Monday' as Day;
let result = 0;

switch (day) {
  case 'Monday': {
    result = 1;
    break;
  }
}
```
or
```TS
type Day = any; // or unknown, or a JS variable outside of a TS context

const day = 'Monday' as Day;
let result = 0;

switch (day) {
  case 'Monday': {
    result = 1;
    break;
  }
}
```

**Expected Result**

I would expect to see an error message compatible with the [`default-case` ESLint rule](https://eslint.org/docs/rules/default-case), indicating that the switch-exhaustiveness-check rule is invalidated because the expression is not a union type.

**Actual Result**

No error message.

**Additional Info**

Essentially, I think the best of both worlds is to just have the predictable functionality of the `default-case` rule operate normally except in the event that a switch statement's expression is a union type. Today, there doesn't seem to be any way for the two rules to operate in tandem. Adding this functionality might be a breaking change, but I'm honestly a bit curious how controversial it would be.

ST-DDT:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2959

## Overview

This PR adds the `requireDefaultForNonUnion` option to the [`switch-exhaustiveness-check`](https://typescript-eslint.io/rules/switch-exhaustiveness-check/) rule.
If enabled, all switches on non-union types require a default case to ensure they are "exhaustive".
The rule is disabled by default, although I would recommend changing this in a future major release, because why would you not want all of your switch cases to be exhaustive.

The rule already has the check for union types so it was easy to use the else branch to add the check for the default case.

````ts
const value: number = Math.floor(Math.random() * 3);

switch (value) {
  case 0:
    return 0;
  case 1:
    return 1;
}
````

->

````ts
const value: number = Math.floor(Math.random() * 3);

switch (value) {
  case 0:
    return 0;
  case 1:
    return 1;
  default: { throw new Error('default case') }
}
````

This cannot be achieved with the [`default-case`](https://eslint.org/docs/latest/rules/default-case) rule, because that also requires the default case for union types which would hide the potential errors caused by extending the union with another case.

### CODE REVIEW

File: packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.md [hunk 1]
```diff
@@ -101,6 +103,25 @@ switch (day) {
...
+
+switch (value) {
+  case 0:
+    return 0;
+  case 1:
+    return 1;
+}
+```
+
+Since `value` is a non-union type it requires the switch case to have a default clause.
```
JoshuaKGoldberg, line 123:
```suggestion
Since `value` is a non-union type it requires the switch case to have a default clause only with `requireDefaultForNonUnion` enabled.
```

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -163,6 +189,28 @@ export default createRule({
...
+          context.report({
+            node: node.discriminant,
+            messageId: 'switchIsNotExhaustive',
+            data: {
+              missingBranches: 'default',
+            },
+            suggest: [
+              {
+                messageId: 'addMissingCases',
+                fix(fixer): TSESLint.RuleFix | null {
```
JoshuaKGoldberg, line 207: [Cleanup] There's no `| null` returned by this function or the others, so none of the three `| null`s need to be here.

```suggestion
                fix(fixer): TSESLint.RuleFix {
```

packages/eslint-plugin/tests/rules/switch-exhaustiveness-check.test.ts
```diff
@@ -209,6 +209,22 @@ function test(value: ObjectUnion): number {
   }
 }
     `,
+    // switch with default clause on non-union type
+    {
+      code: `
+const value: number = Math.floor(Math.random() * 3);
```
JoshuaKGoldberg, line 215: [Cleanup] Just nitpicking, what you're looking to do is more succinctly described by:

```suggestion
declare const value: number;
```

Not a blocking request if you hate this 😄

JoshuaKGoldberg (approved):
LGTM, thanks! Just a little bit of nitpicking left - if you don't like the suggestions or don't have a chance to address them, no worries. We can merge roughly as-is.

JoshuaKGoldberg (approved):
✨ swell, thanks!

JoshuaKGoldberg:
Ah, merge conflicts. I'll take care of them and merge.

JoshuaKGoldberg:
Integration test failures are unrelated. 5dbf6221ea603914c0b9eecdc76a0555df75da10

ArnaudBarre:
Thank you very much for this! Love this <3

But more me this more close to the bug fix, this option should not exist and always be enabled. I don't think anyone choosing to enable this rule will be agaisnt having this check that really catch bugs or shows a missing union in your types

ST-DDT:
@ArnaudBarre Should we create a new issue for that?

JoshuaKGoldberg:
That would be the general right practice, yes 🙂. But I can tell you right now that this rule is far opinionated to not have this as an option. As a user of it I would _maybe_ only ever use it for unions. It'd be quite restrictive to have it apply to _all_ `switch` statements. Sometimes you really just want to `switch` over a few values! It's IMO not our place as a linter to tell people not to use a totally valid pattern of code like that.

I'm not saying this to shut your ideas down. If you feel strongly that I'm wrong here and we should discuss in an issue, then absolutely let's continue the conversation there. We can always discuss there, and even if it's not immediately accepted we can always put the `evaluating community engagement` label on it. Just posting now as a warning before you go through the effort of filing an issue. 😄

ST-DDT:
> @ArnaudBarre Should we create a new issue for that?

Done: https://github.com/typescript-eslint/typescript-eslint/issues/7966

- https://github.com/typescript-eslint/typescript-eslint/issues/7966

PR #7897, opened 2023-11-11
feat(eslint-plugin): [no-floating-promises] flag result of .map(async)

### LINKED ISSUE #5958: [no-floating-promises] flag `.map(async)` result

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.4&sourceType=module&code=MYewdgzgLgBAJgQygmBeGBtA5FAptLAXQG4AoUxZAOgFsEAHACgQgE8xgZGBKNAPhgBvUjBgIA7ggCWsMLnEwACgCcQNKRFyNGAfWX4ANDH0ArXqgGaoAFSk1cIAK5RGpowEYADN+7cyAXz9SIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6aRAR1ko9oEMA7v0r50URNGgB7aJAA04bHiKlkFanQbM2TabQBm8af3zMA5rWIyAtpWQpxkSTLlKAviHdA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
const data = ['test'];

data.map(async () => {
  await new Promise((_res, rej) => setTimeout(rej, 1000));
});
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/require-await": "error",
    "@typescript-eslint/no-floating-promises": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.

### Actual Result

No eslint error, but ERR_UNHANDLED_REJECTION when running

### Additional Info

_No response_

### Versions

| package | version |
| -- | -- |
| `@typescript-eslint/eslint-plugin` | `5.42.1` |
| `@typescript-eslint/parser` | `5.42.1` |
| `TypeScript` | `4.8.4` |
| `ESLint` | `8.15.0` |
| `node` | `web` |

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5958 

## Overview

This PR adds functionality to the rule to check for unawaited arrays of promises, where we might expect to use Promise.all/Promise.allSettled, etc to handle the result.

```
async function rejectIn(t: number) {
  new Promise<void>((resolve) => {
    setTimeout(() => {
      resolve();
    }, t);
  });

  throw new Error();
}

// Unhandled rejections. Will now flag.
[1, 2, 3].map(async (t) => {
  await rejectIn(t);
});
```

I haven't attempted to add editor suggestions for fixes because the fixes seem rather complicated, depending on how you ended up with an array of promises and what you want to do with it. For example, we might fix the above code with 

```
// should we use Promise.all(), Promise.allSettled(), Promise.race(), etc? 
// it's unclear to me that picking one for the user is a good idea.
// Plus, they will still have to handle the resulting promise with catch or something.
Promise.all(
  [1, 2, 3].map(async (t) => {
    await rejectIn(t);
  }),
).catch(() => {});

// This fix would seem plausible but actually violates no-misused-promises with checksVoidReturn: true.
// Also, seems like it would be challenging to implement since it involves inspecting how we got to the
// floating promise in the first place.
[1, 2, 3].forEach(async (t) => {
  await rejectIn(t);
});

// This one should be easy to implement.
void [1, 2, 3].map(async (t) => {
  await rejectIn(t);
});
```

This seems like enough of an edge case that maybe the user should be responsible for manually resolving it. Please advise if adding editor suggestion would be a requirement for this PR.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-floating-promises.ts
+++ b/packages/eslint-plugin/src/rules/no-floating-promises.ts
@@ -120,71 +120,69 @@ export default createRule<Options, MessageId>({
                 ? 'floatingPromiseArrayVoid'
                 : 'floatingPromiseArray',
             });
+          } else if (options.ignoreVoid) {
+            context.report({
+              node,
+              messageId: nonFunctionHandler
+                ? 'floatingUselessRejectionHandlerVoid'
+                : 'floatingVoid',
+              suggest: [
+                {
+                  messageId: 'floatingFixVoid',
+                  fix(fixer): TSESLint.RuleFix | TSESLint.RuleFix[] {
+                    const tsNode = services.esTreeNodeToTSNodeMap.get(
+                      node.expression,
+                    );
+                    if (isHigherPrecedenceThanUnary(tsNode)) {
+                      return fixer.insertTextBefore(node, 'void ');
+                    }
+                    return [
+                      fixer.insertTextBefore(node, 'void ('),
+                      fixer.insertTextAfterRange(
+                        [expression.range[1], expression.range[1]],
+                        ')',
+                      ),
+                    ];
+                  },
+                },
+              ],
+            });
           } else {
-            if (options.ignoreVoid) {
-              context.report({
-                node,
-                messageId: nonFunctionHandler
-                  ? 'floatingUselessRejectionHandlerVoid'
-                  : 'floatingVoid',
-                suggest: [
-                  {
-                    messageId: 'floatingFixVoid',
-                    fix(fixer): TSESLint.RuleFix | TSESLint.RuleFix[] {
-                      const tsNode = services.esTreeNodeToTSNodeMap.get(
-                        node.expression,
+            context.report({
+              node,
+              messageId: nonFunctionHandler
+                ? 'floatingUselessRejectionHandler'
+                : 'floating',
+              suggest: [
+                {
+                  messageId: 'floatingFixAwait',
+                  fix(fixer): TSESLint.RuleFix | TSESLint.RuleFix[] {
+                    if (
+                      expression.type === AST_NODE_TYPES.UnaryExpression &&
+                      expression.operator === 'void'
+                    ) {
+                      return fixer.replaceTextRange(
+                        [expression.range[0], expression.range[0] + 4],
+                        'await',
                       );
-                      if (isHigherPrecedenceThanUnary(tsNode)) {
-                        return fixer.insertTextBefore(node, 'void ');
-                      }
-                      return [
-                        fixer.insertTextBefore(node, 'void ('),
-                        fixer.insertTextAfterRange(
-                          [expression.range[1], expression.range[1]],
-                          ')',
-                        ),
-                      ];
-                    },
+                    }
+                    const tsNode = services.esTreeNodeToTSNodeMap.get(
+                      node.expression,
+                    );
+                    if (isHigherPrecedenceThanUnary(tsNode)) {
+                      return fixer.insertTextBefore(node, 'await ');
+                    }
+                    return [
+                      fixer.insertTextBefore(node, 'await ('),
+                      fixer.insertTextAfterRange(
+                        [expression.range[1], expression.range[1]],
+                        ')',
+                      ),
+                    ];
                   },
-                ],
-              });
-            } else {
-              context.report({
-                node,
-                messageId: nonFunctionHandler
-                  ? 'floatingUselessRejectionHandler'
-                  : 'floating',
-                suggest: [
-                  {
-                    messageId: 'floatingFixAwait',
-                    fix(fixer): TSESLint.RuleFix | TSESLint.RuleFix[] {
-                      if (
-                        expression.type === AST_NODE_TYPES.UnaryExpression &&
-                        expression.operator === 'void'
-                      ) {
-                        return fixer.replaceTextRange(
-                          [expression.range[0], expression.range[0] + 4],
-                          'await',
-                        );
-                      }
-                      const tsNode = services.esTreeNodeToTSNodeMap.get(
-                        node.expression,
-                      );
-                      if (isHigherPrecedenceThanUnary(tsNode)) {
-                        return fixer.insertTextBefore(node, 'await ');
-                      }
-                      return [
-                        fixer.insertTextBefore(node, 'await ('),
-                        fixer.insertTextAfterRange(
-                          [expression.range[1], expression.range[1]],
-                          ')',
-                        ),
-                      ];
-                    },
-                  },
-                ],
-              });
-            }
+                },
+              ],
+            });
           }
         }
       },
--- a/packages/eslint-plugin/src/rules/no-floating-promises.ts
+++ b/packages/eslint-plugin/src/rules/no-floating-promises.ts
@@ -63,7 +63,6 @@ export default createRule<Options, MessageId>({
         messageBase + ' ' + messageRejectionHandler,
       floatingUselessRejectionHandlerVoid:
         messageBaseVoid + ' ' + messageRejectionHandler,
-
       floatingPromiseArray: messagePromiseArray,
       floatingPromiseArrayVoid: messagePromiseArrayVoid,
     },
--- a/packages/eslint-plugin/src/rules/no-floating-promises.ts
+++ b/packages/eslint-plugin/src/rules/no-floating-promises.ts
@@ -113,7 +113,14 @@ export default createRule<Options, MessageId>({
           isUnhandledPromise(checker, expression);
 
         if (isUnhandled) {
-          if (!promiseArray) {
+          if (promiseArray) {
+            context.report({
+              node,
+              messageId: options.ignoreVoid
+                ? 'floatingPromiseArrayVoid'
+                : 'floatingPromiseArray',
+            });
+          } else {
             if (options.ignoreVoid) {
               context.report({
                 node,
@@ -178,13 +185,6 @@ export default createRule<Options, MessageId>({
                 ],
               });
             }
-          } else {
-            context.report({
-              node,
-              messageId: options.ignoreVoid
-                ? 'floatingPromiseArrayVoid'
-                : 'floatingPromiseArray',
-            });
           }
         }
       },
--- a/packages/eslint-plugin/src/rules/no-floating-promises.ts
+++ b/packages/eslint-plugin/src/rules/no-floating-promises.ts
@@ -23,7 +23,9 @@ type MessageId =
   | 'floatingUselessRejectionHandler'
   | 'floatingUselessRejectionHandlerVoid'
   | 'floatingFixAwait'
-  | 'floatingFixVoid';
+  | 'floatingFixVoid'
+  | 'floatingPromiseArray'
+  | 'floatingPromiseArrayVoid';
 
 const messageBase =
   'Promises must be awaited, end with a call to .catch, or end with a call to .then with a rejection handler.';
@@ -34,6 +36,13 @@ const messageBaseVoid =
 
 const messageRejectionHandler =
   'A rejection handler that is not a function will be ignored.';
+
+const messagePromiseArray =
+  "An array of Promises may be unintentional. Consider handling the promises' fulfillment or rejection with Promise.all or similar.";
+
+const messagePromiseArrayVoid =
+  "An array of Promises may be unintentional. Consider handling the promises' fulfillment or rejection with Promise.all or similar," +
+  ' or explicitly marking the expression as ignored with the `void` operator.';
 
 export default createRule<Options, MessageId>({
   name: 'no-floating-promises',
@@ -54,6 +63,9 @@ export default createRule<Options, MessageId>({
         messageBase + ' ' + messageRejectionHandler,
       floatingUselessRejectionHandlerVoid:
         messageBaseVoid + ' ' + messageRejectionHandler,
+
+      floatingPromiseArray: messagePromiseArray,
+      floatingPromiseArrayVoid: messagePromiseArrayVoid,
     },
     schema: [
       {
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -296,11 +325,75 @@ export default createRule<Options, MessageId>({
...
 });
 
+function isPromiseArray(checker: ts.TypeChecker, node: ts.Node): boolean {
+  const type = checker.getTypeAtLocation(node);
+
+  // NOTE - I've chosen isArrayType rather than isArrayLikeType because the
+  // main real use case here is to catch the result of `.map(f)`,
+  // which should always be a native array type.
+  // PR QUESTION - should we be using isArrayLikeType instead?
+  if (checker.isArrayType(type)) {
```
kirkwaiblinger, line 335: I've chosen isArrayType rather than isArrayLikeType because the main real use case here is to catch the result of `.map(f)`, which should always be a native array type. 

Looking for advice on which to put here.

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -296,11 +325,70 @@ export default createRule<Options, MessageId>({
...
+    const arrayType = checker.getTypeArguments(type)[0];
+    if (isPromiseLikeType(checker, arrayType)) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+function isPromiseLikeType(checker: ts.TypeChecker, type: ts.Type): boolean {
```
kirkwaiblinger, line 340: 1. I generally struggled to find out info about the TS Compiler API. If anyone knows where to find details please link me to it.

2. I made a new utility function here in order to check whether the `T` in an `Array<T>` is Promise-like, basing it off the existing function (now renamed to isPromiseLikeNode) . I didn't use that function because it depends on the specific location of the node it's inspecting, not just the `ts.Type` and so I couldn't figure out if it was possible to use it for checking an `Array<T>` (See item 1). Please advise if this is a bad idea. I'm guessing it won't have 100% test coverage as written, since I didn't bother yet writing test cases for all the NonFunctionHandlerThenable cases. If it turns out this is the right way to go, I can go ahead and fill those out.

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -35,6 +37,13 @@ const messageBaseVoid =
 const messageRejectionHandler =
   'A rejection handler that is not a function will be ignored.';
 
+const messagePromiseArray =
+  "An array of Promises may be unintentional. Consider handling the promises' fulfillment or rejection with Promise.all or similar.";
```
kirkwaiblinger, line 41: Discussion is probably needed on the wording of these special case error messages. I just did my best.

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -296,12 +327,31 @@ export default createRule<Options, MessageId>({
   },
 });
 
+function isPromiseArray(checker: ts.TypeChecker, node: ts.Node): boolean {
+  const type = checker.getTypeAtLocation(node);
+  for (const ty of tsutils
+    .unionTypeParts(type)
+    .map(t => checker.getApparentType(t))) {
+    if (checker.isArrayType(ty)) {
```
kirkwaiblinger, line 335: I've chosen isArrayType rather than isArrayLikeType because the main real use case here is to catch the result of .map(f), which should always be a native array type.

Looking for advice on whether this was the correct call or ArrayLike would be better.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -1651,5 +1685,90 @@ Promise.reject(new Error('message')).finally(() => {});
...
+      errors: [{ line: 2, messageId: 'floatingPromiseArrayVoid' }],
+    },
+    {
+      code: `
+[1, 2, Promise.reject().catch(() => {}), 3];
+      `,
+      errors: [{ line: 2, messageId: 'floatingPromiseArrayVoid' }],
+    },
+    {
+      code: `
```
kirkwaiblinger, line 1732: test case from the issue report

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -296,12 +327,31 @@ export default createRule<Options, MessageId>({
...
+      const arrayType = checker.getTypeArguments(ty)[0];
```
kirkwaiblinger, line 336: This doesn't appear to be a sound way to get the element type. A few lines in the test cases have more than one type argument, but I don't know why. For example, 
```
function f<T extends Array<Promise<number>>>(a: T): void {
  a;
}
```

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger, line 335: `isArrayType` covers most cases, but not all. Specifically tuples are trouble. I'd need to add something additional (not sure what exactly) in order to get the following test cases to pass
```
[Promise.reject()] as const; // should error
```
```
declare function cursed(): [Promise<unknown>];
cursed(); // should flag 
```

packages/eslint-plugin/src/rules/no-floating-promises.ts
JoshuaKGoldberg: Yeah it's a little spooky but you should be good. If `ty` is an array type then the first type argument should be matched to the array type parameter. Remaining type arguments get assigned to whatever the element type is (here, `Promise<...>`).

packages/eslint-plugin/src/rules/no-floating-promises.ts
JoshuaKGoldberg, line 1: > I haven't attempted to add editor suggestions for fixes because the fixes seem rather complicated

Yeah I think it's reasonable to leave those out for now. Let's just get the core rule working first.

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger: So, yes, but I wasn't sure then how the type arguments/generic types would work (and haven't been able to find any docs).

I'll change it to my best guess, but then I'd appreciate confirmation it's sound.

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger: Hmm.. I guess, I don't totally follow, since
```
checker.getTypeArguments(/* node corresponding to a */) // results in [Promise, T],
```
whereas I think I would expect `[T, Promise]`, or `[Promise, number]` depending on how I interpret the wording of your comment.

However I'm perfectly happy to take your word for it if you think it's good :)

packages/eslint-plugin/src/rules/no-floating-promises.ts
JoshuaKGoldberg: Yeah the APIs for TypeScript are ... not particularly thoroughly documented. We've mostly just stumbled along with what's exposed publicly (or we've felt comfortable writing manual types to make exposed). If it works then I'd say 👍 go for it.

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger: I have never been more grateful for test infrastructure :)

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -97,75 +109,82 @@ export default createRule<Options, MessageId>({
...
-                  messageId: 'floatingFixAwait',
-                  fix(fixer): TSESLint.RuleFix | TSESLint.RuleFix[] {
-                    if (
-                      expression.type === AST_NODE_TYPES.UnaryExpression &&
-                      expression.operator === 'void'
-                    ) {
-                      return fixer.replaceTextRange(
-                        [expression.range[0], expression.range[0] + 4],
-                        'await',
+            if (options.ignoreVoid) {
```
Josh-Cena, line 124: `} else if (options.ignoreVoid) {`

Josh-Cena (commented):
The tests are great—I cannot devise new test cases. I'll also admit I don't completely understand the problem domain. Just a few nits; the error message looks fine to me.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -475,6 +475,43 @@ Promise.reject()
...
+declare const promiseArray: Array<Promise<unknown>>;
+void promiseArray;
+      `,
+    },
+    {
+      // This one is a bit of a head-scratcher, knowing that the rule recursively
+      // checks .finally() expressions. However, ultimately, the whole expression
+      // is not a promise (it's invalid TS), so it's not a floating promise.
+      code: `
```
JoshuaKGoldberg, line 493: [Testing] Fun 😄 but I don't think we need to include test cases for invalid type errors. Since Array doesn't have a `.finally`, I don't think there's a need to test for the user trying it.

packages/eslint-plugin/src/rules/no-floating-promises.ts
JoshuaKGoldberg: Yeah this is tricky. Thinking...

> of Promises

It might not be an array "of Promises". It might just happen to have one Promise.

> Promise.all

I hesitate to suggest this specific strategy because it's not always the right direction. Sometimes e.g. it's that they accidentally introduced async behavior in a loop that should be in series.

> Consider

We normally try to be a little more assertive in messages. Users get confused if we' seem not sure. 😄 

---

Maybe something like: _"This array may contain at least one floating Promise. Promises must be awaited, end with a call to .catch, or end with a call to .then with a rejection handler."_?

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -54,6 +63,9 @@ export default createRule<Options, MessageId>({
         messageBase + ' ' + messageRejectionHandler,
       floatingUselessRejectionHandlerVoid:
         messageBaseVoid + ' ' + messageRejectionHandler,
+
```
JoshuaKGoldberg, line 66: [Style] Nit: no need for this 🙂 

```suggestion
```

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -296,12 +325,39 @@ export default createRule<Options, MessageId>({
...
+function isPromiseArray(checker: ts.TypeChecker, node: ts.Node): boolean {
+  const type = checker.getTypeAtLocation(node);
+  for (const ty of tsutils
+    .unionTypeParts(type)
+    .map(t => checker.getApparentType(t))) {
+    if (checker.isArrayType(ty)) {
+      const arrayType = checker.getTypeArguments(ty)[0];
+      if (isPromiseLike(checker, node, arrayType)) {
+        return true;
+      }
```
JoshuaKGoldberg, line 337: [Style] Just nitpicking, the type (at least in the current tests) is only ever up to one of these two possibilities (arrays and tuples). So returning early is doable.

```ts
if (checker.isArrayType(ty)) {
  return isPromiseLike(checker, node, checker.getTypeArguments(ty)[0]);
}
```

packages/eslint-plugin/src/rules/no-floating-promises.ts
JoshuaKGoldberg, line 1: [Feature] Whether to include this under an option... I suspect it's a useful enough case that having it on by default wouldn't offend too many people. But having an option to turn off the behavior IMO might be good.

I think it's fine for now, and if someone complains we can always add a feature request to add an option in.

cc @bradzacher @Josh-Cena

JoshuaKGoldberg, line 1: [Docs] Right, and adding an example of this to the docs page would be great.

JoshuaKGoldberg (changes requested):
LGTM! Very close to approval, just some nits, cleaning up some testing, and (mainly) the message text. I'll wait a bit to approve and then merge in case another maintainer has an opinion on whether to add an option. But this is great stuff!

Cartoon Chip and Dale character in an office saying 'great stuff'

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger: Hmmmmm. My concern with this is that it doesn't give someone a way to solve it. Because if
```
[1, 2, Promise.resolve(), 3]
```
comes up, you can't solve it with
```
[1, 2, Promise.resolve().catch(() => {}), 3]
```

That's still a floating promise array (with the current impl at least). But you _could_ solve it with 
```
[1, 2, void Promise.resolve(), 3]; // effectively suppressed
Promise.all([1, 2, Promise.resolve(), 3]).catch(() => {}); // addressed
```
Basically the procedure is either
   a) don't put any promise in the array, _even if you have the handlers on it_
   b) wrap the  array with an array promise method, followed by resolving the ensuing unhandled promise

But I worry I would get misled into trying to resolve a) with the handlers, which is a red herring, and that I wouldn't know to do option b) at all.

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger: Could you clarify this? I'm not following, sorry.

packages/eslint-plugin/src/rules/no-floating-promises.ts
bradzacher: I think this is just a check that people would always want on.
I don't understand when you wouldn't want this, tbh

packages/eslint-plugin/src/rules/no-floating-promises.ts
JoshuaKGoldberg: Yeah :/ I'm not a huge fan of the current wording for being a little too prescriptive too... maybe we can give a quick message and suggest reading the docs for more info?

kirkwaiblinger: My idea for now...
```
This expression may produce an array containing a floating Promise. See the docs for more info on handling Promise arrays. 
```
Really struggling with coming up with something convincing haha.

packages/eslint-plugin/src/rules/no-floating-promises.ts
kirkwaiblinger: added 👍

JoshuaKGoldberg (approved):
Cartoony Octopus bringing up a 'SHIP IT' sign

PR #7922, opened 2023-11-13
chore(deps): update babel to 7.23.3

JoshuaKGoldberg:
## Overview

Following the directions in https://typescript-eslint.io/maintenance/dependency-version-upgrades#babel.

Also sent #7921.


bradzacher (approved):

JoshuaKGoldberg:
Re-requesting review because I updated the snapshots.

packages/ast-spec/src/legacy-fixtures/basics/fixtures/dynamic-import-with-import-assertions/snapshots/3-Babel-AST.shot
```diff
@@ -87,6 +87,85 @@ Program {
             end: { column: 42, line: 3 },
           },
         },
+        options: ObjectExpression {
```
bradzacher, line 90: why the heck do they emit both `.attributes` and `.options`?!?

bradzacher (approved):

PR #7933, opened 2023-11-15
fix(utils): add missing fields to flat config types

bradzacher:
## Overview

This adds the missing fields from #7273

cc @karlhorky

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -119,6 +119,7 @@
     "serializers",
     "Sourcegraph",
     "stringification",
+    "stringly",
     "superset",
     "thenables",
     "transpiled",
--- a/packages/utils/src/ts-eslint/Config.ts
+++ b/packages/utils/src/ts-eslint/Config.ts
@@ -131,14 +131,36 @@ export namespace FlatConfig {
   export type SeverityString = SharedConfig.SeverityString;
   export type SourceType = ParserOptionsTypes.SourceType | 'commonjs';
 
+  export interface PluginMeta {
+    /**
+     * The meta.name property should match the npm package name for your plugin.
+     */
+    name: string;
+    /**
+     * The meta.version property should match the npm package version for your plugin.
+     */
+    version: string;
+  }
   export interface Plugin {
     /**
+     * Shared configurations bundled with the plugin.
+     * Users will reference these directly in their config (i.e. `plugin.configs.recommended`).
+     */
+    configs?: Record<string, Config>;
+    /**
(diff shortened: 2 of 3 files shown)


JoshuaKGoldberg (approved):
TIL "stringly"

karlhorky:
Nice, thanks! I'll check this out when this lands in [`@typescript-eslint/utils@6.11.1`](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v6.11.1) or  [`@typescript-eslint/utils@6.12.0`](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v6.12.0)

JoshuaKGoldberg:
For visibility, we publish canary builds on every commit to `main`: https://main--typescript-eslint.netlify.app/users/versioning.

(not saying you need to try them - just that if anybody urgently wants this ASAP, the canaries are an option 😄)

karlhorky:
Ok thanks, I decided to try out the canary / alpha releases and upgraded to `@typescript-eslint/*@6.11.1-alpha.20` here, and simplified some types:

https://github.com/upleveled/eslint-config-upleveled/commit/bb2a8600b2d9c6fca3aae93d83ea4eab2ff09102#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346

bradzacher:
@karlhorky FYI you're doing things a little incorrectly here!
Specifically the type of the plugin isn't a `FlatConfig.Plugin` because the plugin doesn't support flat configs yet.

If you have a look at https://github.com/typescript-eslint/typescript-eslint/pull/7935 you can see how I typed things.

We can probably expose better types here though to make this easier. I.e. We should probably expose a type like `FlatConfig.ClassicPlugin<T>` where `T` can be used to define the shared configs.

You also should be using the compat tooling instead of deeply pulling apart the config. I still don't know the "best" way to do this style of "extend multiple configs for a given set of files. I asked on E ESLint discord (https://discord.com/channels/688543509199716507/1174310123791990866) but I got ignored... Might need try pinging again.

karlhorky:
Yeah, I'm aware I'm faking the types a bit in some places (this is to avoid having to know / keep updated on which plugins expose which type of config). Maybe I should open PRs to add types to each of the plugins I use... 🤔 so that I don't need to maintain them in my project and then everyone has types.

This ESLint config is pretty runtime-tested as is (in many student projects), so it's mostly just making the type system happy and reducing the faking to the least amount possible.

But I will for sure watch all new things including #7935 and use anything to make it simpler!

PR #7951, opened 2023-11-19
chore(website): auto-generate type checked rule notice in rule docs

### LINKED ISSUE #7918: Docs: Auto-add a performance warning to typed rules' "When Not To Use It"

### Suggested Changes

Several typed rules have a similar kind of _"performance is slower with typed rules"_ notice in their _`## When Not To Use It`_. Examples:

https://github.com/typescript-eslint/typescript-eslint/blob/3cc3d0eff8e9d574d35c05c97b607a8a6ba59d9b/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md?plain=1#L98

https://github.com/typescript-eslint/typescript-eslint/blob/3cc3d0eff8e9d574d35c05c97b607a8a6ba59d9b/packages/eslint-plugin/docs/rules/prefer-reduce-type-parameter.md?plain=1#L58

How about we standardize this by adding that automatically in rule docs generation (https://github.com/typescript-eslint/typescript-eslint/blob/3cc3d0eff8e9d574d35c05c97b607a8a6ba59d9b/packages/website/plugins/generated-rule-docs.ts)?

### Affected URL(s)

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7918

## Overview

Adds the following notice to the _When Not To Use It_ section of rule docs, auto-generating that h2 if it doesn't yet exist:

> Type checked lint rules are more powerful than traditional lint rules, but also require configuring [type checked linting](http://localhost:3000/linting/typed-linting). See [Performance Troubleshooting](http://localhost:3000/troubleshooting/performance-troubleshooting) if you experience performance degredations after enabling type checked rules.

Bikeshedding welcome on that message.

I also got fed up trying to work with the >500 lines of `generated-rule-docs.ts`, so I enacted two refactors:

1. Splitting up the major parts of its logic into individual files/functions with a few shared utilities. Now `generated-rule-docs/index.ts` has IMO a much more readable flow.
2. Adding a `spliceChildrenAndAdjustHeadings` function that adjusts all headings' indices.  Previously the code mostly manually did it.


packages/website/plugins/generated-rule-docs/spliceWhenNotToUseIt.ts
```diff
@@ -0,0 +1,64 @@
+import type * as mdast from 'mdast';
+import type * as unist from 'unist';
+
+import type { RequiredHeadingIndices } from './requiredHeadings';
+import { spliceChildrenAndAdjustHeadings } from './splicing';
+import type { RuleMetaDataWithDocs } from './types';
+
+export function spliceWhenNotToUseIt(
```
Josh-Cena, line 8: Can we not call functions `spliceX`, `unshiftX`, or `pushX`? They both function as `insert` and which one is used is an implementation detail.

packages/website/plugins/generated-rule-docs/nodes.ts
```diff
@@ -0,0 +1,14 @@
+import type * as mdast from 'mdast';
+import type * as unist from 'unist';
+
+export function nodeIsCode(node: unist.Node): node is mdast.Code {
```
Josh-Cena, line 4: I suggest moving all non-tree-modifying helper functions to the same file.

Josh-Cena (commented):
Just some tiny nits

packages/website/plugins/generated-rule-docs/spliceWhenNotToUseIt.ts
JoshuaKGoldberg: 😍 I love this comment so much. Yes!

packages/website/plugins/generated-rule-docs/nodes.ts
JoshuaKGoldberg: Sorry, not sure how to parse this - which helper functions are you thinking of?

bradzacher:
Perhaps it would be good to add this as a subsection if there already is a "When not to use it" section?
That way our hand-crafted text sits more prominently and is separated from the auto-generated text.

EG `await-thenable`:

screenshot of docs

It's hard to tell that there are two separate points there and additionally the hand-crafted point is placed second.

packages/website/plugins/generated-rule-docs/nodes.ts
Josh-Cena: I meant these files: nodes.ts, requiredHeadings.ts, spliceChildrenAndAdjustHeadings.ts, strings.ts, types.ts. These are all small utilities and I don't entertain the idea of splitting them into 5 files well.

packages/website/plugins/generated-rule-docs/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  nodeIsParent,
+} from './utils';
+
+export const generatedRuleDocs: Plugin = () => {
+  return (root, file) => {
+    if (!nodeIsParent(root) || !isVFileWithStem(file)) {
+      return;
+    }
+
+    // const rule = file.stem in pluginRules ? pluginRules[file.stem] : undefined;
```
JoshuaKGoldberg, line 26:
```suggestion
```

bradzacher (approved):

PR #7955, opened 2023-11-19
chore: enable prefer-nullish-coalescing internally

JoshuaKGoldberg:
## PR Checklist

## Overview

Fixes up most of the last few violations of [@typescript-eslint/prefer-nullish-coalescing](https://typescript-eslint.io/rules/prefer-nullish-coalescing). Ignores primitives because we tend to do "fun" things with strings.


packages/eslint-plugin/src/rules/consistent-type-imports.ts
```diff
@@ -611,7 +611,7 @@ export default createRule<Options, MessageIds>({
         const { namedSpecifiers: valueImportNamedSpecifiers } =
           classifySpecifier(sourceImports.valueImport);
         if (
-          sourceImports.valueOnlyNamedImport ||
+          sourceImports.valueOnlyNamedImport ??
           valueImportNamedSpecifiers.length
```
bradzacher, line 615: personally - I hate this style of `if` condition.
I think the `||` is much clearer here.

bradzacher (commented):
Can we turn it on with `ignoreConditionalTests`?
I think that using `??` in conditional tests is less clear.

Consider the downlevelled output:
```ts
if (x || y) { }

// vs

if (x ?? y) { } 
// downlevelled to
if (x != null ? x : y) { }
```

I think it's a whole lot less clear.
Actually I'm going to go as far as advocating for flipping the default in v7.

packages/eslint-plugin/src/rules/consistent-type-imports.ts
JoshuaKGoldberg: Finally getting to this: enabled `ignoreConditionalTests` in the ESLint config and reverted the now-unnecessary changes.

bradzacher (approved):

PR #7957, opened 2023-11-19
feat(eslint-plugin): [no-useless-template-literals] add new rule

### LINKED ISSUE #2846: Rule proposal: detect useless template literal interpolation
Hi there,

First of all, thanks for this great set of ESLint + TS utilities, these are great!

I am unsure if I'm doing this right, because the issue template was about bug reports, but I would like to propose a new rule. Since I didn't see a template for this, I'm going to go ahead like this, and then can change it / resubmit if necessary.

What I wanted to bring up was the idea of a `@typescript-eslint/no-useless-interpolation` rule, similar to the https://github.com/eslint/eslint/issues/10798 proposal to ESLint, which as far as I can tell, never got implemented.

The idea would be, if there is useless interpolation with a template string, this would be an error.

Example of **incorrect** code:

```jsx
`${'abc'}` // incorrect

const str = 'def';
`${str}` // incorrect

`123${'abc'}` // incorrect

`${'abc'}${'def'}` // incorrect
```

Examples of **correct** code:

```jsx
'abc' // correct

const str = 'def';
str // correct

'123abc' // correct

'abcdef' // correct
```

I would suggest only doing this for strings.

## Remaining template things

StyleShit:
Closes #2846


packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -0,0 +1,90 @@
...
+          node.quasis[1].value.raw === '' &&
+          node.expressions.length === 1 &&
+          node.expressions[0].type === AST_NODE_TYPES.Identifier &&
+          isUnderlyingTypeString(node.expressions[0]);
+
+        if (hasSingleStringVariable) {
+          context.report({
+            node,
+            messageId: 'noUselessTemplateLiteral',
+          });
```
StyleShit, line 71: should probably add a fixer in another PR
will raise a new issue after this one is merged

StyleShit:
@JoshuaKGoldberg 

Not sure why the tests are failing
Should I run `yarn generate:configs`? It doesn't seem to do anything useful

packages/eslint-plugin/src/rules/index.ts
StyleShit, line 1: Not sure why the tests are failing
Should I run `yarn generate:configs`? It doesn't seem to do anything useful

JoshuaKGoldberg: See #6899 - you can run `yarn test -u` in `packages/eslint-plugin` to get these updated.

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
JoshuaKGoldberg, line 1: [Testing] A few more cases come to mind:

* `never`
* `unknown`
* Generic types (making sure type parameter constraints are respected)
* Template literal that includes a string variable but also other text content (so should be `valid`)

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
```diff
@@ -0,0 +1,194 @@
...
+    tsconfigRootDir: rootPath,
+    project: './tsconfig.json',
+  },
+});
+
+ruleTester.run('no-useless-template-literals', rule, {
+  valid: [
+    "const string = 'a';",
+
+    // allow variables & literals concatenation
```
JoshuaKGoldberg, line 20: [Style] Not a blocker IMO, but a preference - I don't think there's a need for the comments in the test file explaining sections of tests. They tests are pretty self-explanatory. And over time these comments have a bad habit of falling out of sync.

```suggestion
```

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    return {
+      TemplateLiteral(node: TSESTree.TemplateLiteral): void {
+        // don't check tagged template literals
+        if (node.parent.type === AST_NODE_TYPES.TaggedTemplateExpression) {
+          return;
+        }
+
+        // don't allow a single variable in a template literal
```
JoshuaKGoldberg, line 58: [Style] Similar to the test file, the comments in this function don't really add anything IMO. "don't check tagged template literals" just above a check that the node's parent isn't a tagged template expression is roughly the same. This variable's name and the variable name below are both roughly the same as their comments.

```suggestion
```

packages/eslint-plugin/docs/rules/no-useless-template-literals.md
```diff
@@ -0,0 +1,64 @@
...
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-useless-template-literals** for documentation.
+
+This rule reports template literals that can be simplified to a normal string literal.
+
+## Examples
+
+<!--tabs-->
```
JoshuaKGoldberg, line 13: [Docs] Many of our rule files intentionally have symmetry between the correct and incorrect examples in the docs. As in, if you switch between the two of them, you can see that lines switch between correct and incorrect code in-place. That's a pleasantry for readers so they can more quickly see the diffs between the two. Can you apply that here?

JoshuaKGoldberg (changes requested):
🚀 a great start! Requested some changes that I suspect might require some code changes. Can take a deeper look once they're resolved. Many thanks for sending this in!

Mr. Burns saying 'excellent' from The Simpsons

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
```diff
@@ -0,0 +1,194 @@
...
+      const concatenated = \`\${number}b\`;
+    `,
+
+    `
+      const boolean = true;
+      const concatenated = \`\${boolean}b\`;
+    `,
+
+    `
+      const nullish = nullish;
```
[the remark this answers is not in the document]
StyleShit: whoops, that's a typo
fixed

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
StyleShit: yup, you're probably right
I just can't get used to seeing this type of tests instead of seeing something like `it('should work...')` 😂 
anyway, removed (I think losing sync is a good point here)

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
StyleShit: yeah, TBH this condition and its comment were copied from somewhere else, and I just followed the same pattern
I agree that `self-explanatory code > comment` (hence the variables/functions names) 
removed

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
StyleShit: never - hmm, not sure what should happen in this case, in the current implementation it's valid
unknown - done
generic - good one, done
template with other text - already exists

packages/eslint-plugin/docs/rules/no-useless-template-literals.md
StyleShit: that's a great idea, I like it!

packages/eslint-plugin/src/rules/index.ts
StyleShit: seems like I should've also used `yarn generate:configs`
thanks!

StyleShit:
IDK why prettier is failing, seems to be fine locally

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -0,0 +1,97 @@
...
+          });
+
+          return;
+        }
+
+        const allowedChars = ['\r', '\n', "'", '"'];
+
+        const hasStringWithAllowedChars = node.quasis.some(quasi => {
+          return new RegExp(`[${allowedChars.join('')}]`).test(quasi.value.raw);
+        });
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Or even better, `/[\r\n'"]/`. No need to dynamically compute the regular expression.

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -0,0 +1,97 @@
...
+
+        if (hasStringWithAllowedChars) {
+          return;
+        }
+
+        const allAreLiterals = node.expressions.every(expression => {
+          return expression.type === AST_NODE_TYPES.Literal;
+        });
+
+        if (allAreLiterals) {
```
JoshuaKGoldberg, line 88: [Bug] It's possible to have a situation where the rule should complain on something that doesn't have _all_ literals though.

Here, the rule should report specifically on the <code>\`${'def'}\`</code>:

```ts
const other = 'abc';
`before${'def'}middle${other}after`;
```

Instead of having the rule report on the entire `node`, let's have the rule report on each violating literal. Then we get a smaller, more precise squiggly range.

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -396,8 +396,8 @@ function getNodeType(node: Member): MemberKind | null {
       return node.value && functionExpressions.includes(node.value.type)
         ? 'method'
         : node.readonly
-          ? 'readonly-field'
-          : 'field';
+        ? 'readonly-field'
+        : 'field';
```
JoshuaKGoldberg, line 400: [Style] Unrelated changes, can you revert?

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
JoshuaKGoldberg, line 88: [Bug] Also on this `.every` usage: I don't think we'd want this rule to report on template literals with _no_ `expressions`. The stylistic decision of whether to use template literals vs other forms of apostrophes when there's no interpolation is out of scope for this bug-catching rule. https://eslint.org/docs/latest/rules/quotes -> https://typescript-eslint.io/rules/quotes/ should catch that.

Request: revert the stylistic changes in other code, and add a test?

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
JoshuaKGoldberg, line 1: [Testing] A bit more comes to mind:

* Bigint
* Symbol
* Function
* `null` keyword
* `undefined` literal
* `T extends number` in the `invalid` cases

JoshuaKGoldberg (changes requested):
Getting closer, fantastic! 🥳 

Mostly requesting changes on making the rule not flag stylistic points. On deeper thought I think there's some more testing to be done too.

Thanks!

packages/eslint-plugin/docs/rules/no-useless-template-literals.md
```diff
@@ -0,0 +1,53 @@
...
+
+```ts
+const ab1 = 'ab';
+const ab2 = 'ab';
+
+const stringWithNumber = '1 + 1 = 2';
+
+const stringWithBoolean = 'true is true';
+
+const string = 'a';
```
JoshuaKGoldberg, line 42: [Docs] It's sometimes confusing for folks to see the name of a primitive used as a variable name. 

```suggestion
const text = 'a';
```

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
```diff
@@ -0,0 +1,240 @@
...
+ruleTester.run('no-useless-template-literals', rule, {
+  valid: [
+    "const string = 'a';",
+
+    `
+      const string = 'a';
+      const concatenated = \`\${string}b\`;
+    `,
+
+    `
```
park-jemin, line 25: Question: How are we handling conflicts with [`quotes`](https://typescript-eslint.io/rules/quotes/)? Been on a couple projects where the backtick is preferred, and it would be nice if we could scope this to only strings _with_ template literals if possible.

Should add tests for these imo:

```ts
// For example, these should be valid (even if they flag other rules)
const a = `A`; 
const b = `$B`;
const c = '${"C"}';
const d = "${'D'}";
const e = '${`C`}';
const f = "${`D`}";
```

JoshuaKGoldberg: https://github.com/typescript-eslint/typescript-eslint/pull/7957#discussion_r1416275932 - yeah this rule shouldn't try to enforce style preferences. It should only target unnecessary logic.

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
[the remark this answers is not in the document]
StyleShit: yup
I actually don't need this anymore, since it's basically a stylistic check

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -0,0 +1,97 @@
...
+
+type MessageId = 'noUselessTemplateLiteral';
+
+export default createRule<[], MessageId>({
+  name: 'no-useless-template-literals',
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow unnecessary template literals',
+      recommended: 'recommended',
```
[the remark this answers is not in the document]
StyleShit: I will eventually learn all of the rules (pun intended) here...

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -0,0 +1,97 @@
...
+        if (node.parent.type === AST_NODE_TYPES.TaggedTemplateExpression) {
+          return;
+        }
+
+        const hasSingleStringVariable =
+          node.quasis.length === 2 &&
+          node.quasis[0].value.raw === '' &&
+          node.quasis[1].value.raw === '' &&
+          node.expressions.length === 1 &&
+          node.expressions[0].type === AST_NODE_TYPES.Identifier &&
```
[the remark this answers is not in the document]
StyleShit: added some tests

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
StyleShit: Done

packages/eslint-plugin/src/rules/member-ordering.ts
StyleShit: Sure

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
StyleShit: Done

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
StyleShit: Not sure what's the test for `T extends numbers`
others are done

packages/eslint-plugin/docs/rules/no-useless-template-literals.md
StyleShit: Done

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
StyleShit: Fixed, thanks!

StyleShit:
no idea why the build is failing

JoshuaKGoldberg:
Btw @StyleShit I appreciate you answering every thread 🙂  if it's easier for you, you can just mark them as resolved.

JoshuaKGoldberg:
> no idea why the build is failing

Yeah 🙁 not related to this PR. Should be fixed by #8048. I'll update the branch.

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
JoshuaKGoldberg: There's already a test with `function func<T extends number>(arg: T) {` in the `valid` cases. But for `invalid`... ok I forget what I was thinking here. Never mind then 😄

StyleShit:
> Btw @StyleShit I appreciate you answering every thread 🙂 if it's easier for you, you can just mark them as resolved.

@JoshuaKGoldberg
Hehe that's the force of a habit... 
At Elementor, we usually answer every thread to make sure we didn't miss anything, and let the reviewer resolve the thread.
It usually helps the reviewers keeping track of the discussions and put themselves back into context on a re-review.

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
```diff
@@ -0,0 +1,303 @@
...
+
+ruleTester.run('no-useless-template-literals', rule, {
+  valid: [
+    "const string = 'a';",
+
+    'const string = `a`;',
+
+    `
+      const string = 'a';
+      const concatenated = \`\${string}b\`;
```
JoshuaKGoldberg, line 24: [Style] This is non-blocking, but just in case you prefer any of these:

* Instead of declaring variables, you can use `declare`. Example: `declare const string: 'a';`. It's a bit more semantically close to what the test is trying to do (use an identifier with a particular type). Though admittedly it's more characters...
   * (I do see there are some cases that do this now - reposting this point so it's alongside the others)
* Omit the variable for a template literal, and just have it as the statement. Example: <code>\`\\${string}b\`;</code>.
* If there's only one line in the test case, not have an extra blank line in front & end.

I won't request changes on any of these. Just figured I'd mention them 😄

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    `
+      const concatenated = \`true && false = \${true && false}\`;
+    `,
+
+    `
+      tag\`\${'a'}\${'b'}\`;
+    `,
+
+    `
+      const wrappedNumber = \`\${1}\`;
```
JoshuaKGoldberg, line 74: [Suggestion] Should this be in `invalid`, though? Right now the rule only looks at string types, but I can't think of a reason why it wouldn't flag the useless template literal here. Same with the bigints, booleans, `null`, and `undefined`.

JoshuaKGoldberg:
I, yet again, and resentful that GitHub's PR process doesn't give a way to distinguish between _"one of us in this conversation thinks it's resolved"_ and _"both of us in this conversation think it's resolved"_... 🙃

JoshuaKGoldberg (changes requested):
Ok this is looking great! Thanks for keeping working on it. 🙌 

Apologies for the late-breaking request: I think the only blocker (other than reverting unrelated changes) is also reporting on unnecessary non-string literals such as `bigint` and `number`. Or at least, resolving the conversation around them is.

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
StyleShit: hmm,
I wondered the same, but I think this is valid since it has a different meaning than `const value = 1`
it's being converted to a string here, which might be the intention of the developer

packages/eslint-plugin/tests/docs.test.ts
```diff
@@ -89,16 +89,16 @@ describe('Validating rule docs', () => {
...
             `See **https://typescript-eslint.io/rules/${ruleName}** for documentation.`,
-            ``,
+            '',
           ].join('\n'),
           type: 'blockquote',
         });
       });
 
-      test(`headings must be title-cased`, () => {
+      test('headings must be title-cased', () => {
```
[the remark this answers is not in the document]
StyleShit: Oops
those are from the days when this rule was stylistic 🫣

reverted 😄

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
StyleShit: 1. declare is longer, yeah, but I don't really have a preference about it, so why not
2. hmm, I gotta admit that having a statement without a variable seems weird to me, but same as above, don't really have any strong preference
3. I've changed some of the cases but kept others with empty lines. I think sometimes it's not readable and feels too condensed.

See this: 
image

Versus this:
image

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
JoshuaKGoldberg: What I mean is, it's same as `const value = '1'`, no?

JoshuaKGoldberg (changes requested):
(just https://github.com/typescript-eslint/typescript-eslint/pull/7957#discussion_r1424188459 left)

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
StyleShit: hmm
so you're saying that in any case where there is any type of literal inside an expression it should be invalid?

JoshuaKGoldberg: Yeah that's what I'm thinking. Unless I've missed something -entirely possible, JS coercion/stringification rules are wacky- I think the invalid cases there are unnecessary.

StyleShit: yeah, so this is kinda similar to my initial approach, using the `areAllLiterals` if you remember
should be fine (theoretically? I mean.. it's JS after all...)

StyleShit: What about things like this?

```TS
`1 + 1 = ${1 + 1}`;
```

JoshuaKGoldberg: > initial approach ... `areAllLiterals`

Yeah! I think my concern originally was putting the report range on the whole template literal vs. individual expressions. If an individual expression is any literal, then it should be reported on IMO.

> `${1 + 1}`

Expressions I think we can avoid. We can draw a line at just literals. Anything dynamic, such as a binary expression, can get difficult to report on.

StyleShit: Ok, see this one https://github.com/typescript-eslint/typescript-eslint/pull/7957/commits/408e5c721d6e00c7c329d0d355171c67a5fcb2c3

JoshuaKGoldberg (approved):
🎆 this is great, thanks for all your work on it! 

Looking forward to getting this into _recommended_ for v7 😄

StyleShit:
filed a new issue as promised:
https://github.com/typescript-eslint/typescript-eslint/issues/8056

karlhorky:
Just tested this in [`@typescript-eslint/eslint-plugin@6.15.0`](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v6.15.0) and it works well, thanks @StyleShit and @JoshuaKGoldberg and everyone else involved! 🙌

Zamiell:
thanks Style, this is a very nice rule!

PR #7968, opened 2023-11-21
feat: support TS 5.3

### LINKED ISSUE #7719: TypeScript 5.3 Support
https://devblogs.microsoft.com/typescript/announcing-typescript-5-3-beta/

*Please be patient.*

# ✅ Import Attributes (#7821)

We should take this opportunity to deprecate our [import/export `.assertions`](https://github.com/estree/estree/blob/master/stage3/import-assertions.md) property, and instead implement the [import/export `.attributes`](https://github.com/estree/estree/blob/master/experimental/import-attributes.md) property. The AST shape is otherwise unchanged so it should just be a property deprecation.

From the TS AST side - they deprecated the `.assertClause` property that we currently use to get the attributes and the intended replacement is `.attributes`. Otherwise no changes.

# ✅ Optimizations by Skipping JSDoc Parsing (#7821)

Our usecase was one of the motivators for this change!
We'll want to ensure we are parsing with `host.jsDocParsingMode = JSDocParsingMode.ParseForTypeInfo`.

This should reduce memory usage and improve parsing performance.

# 🔲  lib.d.ts Updates (#7923)

---

- Stable Support `resolution-mode` in Import Types
- `switch (true)` Narrowing
  - it's not treated the same as an `if` condition - so we don't need to worry about updating our rules to match this, I think?
- Narrowing On Comparisons to Booleans
- `instanceof` Narrowing Through `Symbol.hasInstance`
- Checks for `super` Property Accesses on Instance Fields
- Interactive Inlay Hints for Types
- Optimizations by Comparing Non-Normalized Intersections
- Consolidation Between tsserverlibrary.js and typescript.js
  - Once we drop <5.3 we can switch everything to import from the root export. But for now we need backwards compat.
- Breaking Changes and Correctness Improvements
lib.d.ts Changes
  - Checks for `super` Accesses on Instance Properties

### LINKED ISSUE #7956: Repo: Use less inconvenient ways of enforcing our usage of TypeScript APIs than patching
### Suggestion

Splitting out of https://github.com/typescript-eslint/typescript-eslint/pull/7923#discussion_r1398071314: we're finding it rather inconvenient to have to work with [`yarn patch`](https://yarnpkg.com/cli/patch) to apply our patches to TypeScript's `.d.ts` file.

For context, we want to enforce a few somewhat-specific-to-use ways of using TypeScript:

* Adding in removed fields as `@deprecated` optionals for backwards compatibility
* `@deprecating` fields not present in all our supported versions

Example patch: https://github.com/typescript-eslint/typescript-eslint/blob/8ec3800d6c4eef078ec20b5434805e731242a49d/.yarn/patches/typescript-npm-5.3.0-beta-33b1e90865.patch

Unfortunately, Yarn patching presents a couple inconveniences:
* https://github.com/yarnpkg/berry/issues/4441 gets in the way (though does not seem to fully block) patching TypeScript.
* Even ignoring ☝️, we have to manually re-apply patches every time there's a new TypeScript that changes the hash of the file. Ugh. 

I'd rather we did something less getting-in-the-way-of-tooling / needing-manual-work-every-TS-version-change, such as declaration merging and/or linting.

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7719, fixes #7956

## Overview

- Updates the TS version ranges
- Fixes a bug in the types that causes type-errors in userland installs with TS 5.3
- Simplifies our local TS types patch by moving it to declaration merging (ref #7956)
    - We do lose one safety patch (mark `SyntaxKind.JSDoc` as deprecated as it was added in 4.7) - but that's not an enum member we'll ever used so I figured it was worth it.
- Cleans up the scope-manager lib generator so it generates lint compliant code.
    - I ran the generator and it generated changes in some files.
    - Turns out that we formatted the code, then lint + fixed the code after it was written.
    - This would leave incorrectly formatted code behind in the imports.
    - So either we could add another format pass - or just make the generated code lint compliant! I did the former cos it was easy.


docs/users/Dependency_Versions.mdx
```diff
@@ -7,7 +7,7 @@ import packageJson from '../../package.json';
 
 ## ESLint
 
-> The version range of ESLint currently supported is `^6.0.0 || ^7.0.0 || ^8.0.0`.
+> The version range of ESLint currently supported is `^7.0.0 || ^8.0.0`.
```
bradzacher, line 10: i went to update this file cos I'd forgotten we'd automated it via importing the `package.json`. But I noticed this was wrong so I fixed it.

packages/typescript-estree/src/node-utils.ts
```diff
@@ -191,6 +191,7 @@ export function isComment(node: ts.Node): boolean {
  * @returns is JSDoc comment
  */
 function isJSDocComment(node: ts.Node): node is ts.JSDoc {
+  // eslint-disable-next-line deprecation/deprecation -- SyntaxKind.JSDoc was only added in TS4.7 so we can't use it yet
```
bradzacher, line 194: this is the compromise to removing the patch.
you can't declaration merge a comment onto an enum member.
small price to pay though.

packages/typescript-estree/typings/typescript.d.ts
```diff
@@ -10,4 +10,48 @@ declare module 'typescript' {
...
+   */
+  function getModifiers(node: HasModifiers): readonly Modifier[] | undefined;
+  /**
+   * @deprecated don't use this directly as it does not exist pre-4.8; instead use getModifiers from `src/getModifiers.ts`.
+   */
+  function canHaveModifiers(node: Node): node is HasModifiers;
+  /**
+   * @deprecated don't use this directly as it does not exist pre-4.8; instead use getDecorators from `src/getModifiers.ts`.
+   */
+  function canHaveDecorators(node: Node): node is HasDecorators;
```
bradzacher, line 56: This is what we used to patch!

packages/typescript-estree/src/ts-estree/ts-nodes.ts
```diff
@@ -4,9 +4,9 @@ import type * as ts from 'typescript';
 // Eg: https://github.com/typescript-eslint/typescript-eslint/issues/2388, https://github.com/typescript-eslint/typescript-eslint/issues/2784
 /* eslint-disable @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error, @typescript-eslint/no-empty-interface */
 declare module 'typescript' {
-  // @ts-ignore - added in TS 4.5
+  /** @ts-ignore - added in TS 4.5, deprecated and converted to a type-alias in TS 5.3 */
   export interface AssertClause extends ts.Node {}
-  // @ts-ignore - added in TS 4.5
+  /** @ts-ignore - added in TS 4.5, deprecated and converted to a type-alias in TS 5.3 */
   export interface AssertEntry extends ts.Node {}
```
bradzacher, line 10: https://github.com/microsoft/TypeScript/issues/56477

This is a problem for us.
Testing in the work repo this breaks consumers on TS5.3
Even with the jsdoc style - TS will error on the TS declaration.

screenshot of type error

We don't get this error locally because we use `skipLibChecks`.

Note that this blocks the full support.

JoshuaKGoldberg (approved):
Great, thanks! 🚀 

Happy to see the patches removed, even considering the cost.

bradzacher:
https://github.com/microsoft/TypeScript/issues/56477 has a fix merged and will release with v5.3.3. We can merge this in now and just say "fuck it" and tell plugin authors to just turn on `skipLibChecks` until 5.3.3.

For lint users we do have full support for 5.3 so it feels wrong to not merge this.

WDYT @typescript-eslint/triage-team

fisker:
Ready to release?

bradzacher:
https://github.com/typescript-eslint/typescript-eslint/issues/7719#issuecomment-1826964205

fisker:
Thank you!

PR #7988, opened 2023-11-24
chore: prefix all nx package scripts with npx

### LINKED ISSUE #7987: Repo: Many dev commands require a global nx command
### Suggestion

Commands prefixed with `nx` (e.g. [`nx run ...`](https://nx.dev/nx-api/nx/documents/run)) require either something like `npx nx` in front or a global `nx` script, such as from `npm i -g npm`. If you don't then trying to run them gives an error:

```plaintext
josh ~/repos/typescript-eslint/packages/website $ yarn start
command not found: nx
```

Running a search for `": "nx ` in `package.json` files, we get 40 results in 18 files right now. 

I think we have two likely workarounds:

* Explicitly require dev machines have run something like `npm i -g nx`
* Prefix all those commands with something like `npx ` (e.g. instead of `nnx

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7987

## Overview

Prefixes all the commands that use `nx` with `npx `.


merceyz (changes requested):
I'm not too familiar with the inner workings of `npx` but wont this use whatever version is installed globally and if none is found prompt the user to install it?

This repository also doesn't use `npm` so this doesn't seem like an ideal solution.

If what you're after is using the `nx` version installed in the root workspace then you should use `yarn run -T nx` or `run -T nx`.
Ref https://github.com/typescript-eslint/typescript-eslint/issues/7987#issuecomment-1825457681.

bradzacher:
Could we just add nx as a dep and then prefix with yarn instead?

JoshuaKGoldberg:
Swapping `"start": "npx nx start"` with `"start": "yarn nx start"` yields:

```plaintext
josh ~/repos/typescript-eslint/packages/website $ yarn start
Usage Error: Couldn't find a script named "nx".
```

Same with `yarn run nx start`. I've been reading through https://nx.dev/core-features/run-tasks & co but don't see any recommendations around this?

bradzacher:
Yeah I'm not quite sure - I would have thought that `yarn nx` would run the locally installed verison of `nx` - do we need to add `nx` as a dependency to each and every package so that yarn can see it?

IDK how Yarn3 works - you didn't need to do that with yarn1 as it'd just look in the parent folder.

JamesHenry:
How frustrating yeah this definitely worked fine in yarn 1. FWIW you can also install nx globally one time and it will always work as well, it will resolve your most local installation possible before attempting to run the actually globally installed code.

I’ll need to play with this to get a real sense of what might optimal here, I’m afraid I lost all of the weekend because of moving apartments and my next availability will be this coming weekend, so feel free to do whatever makes the most sense in the immediate term if this is blocking

PR #7997, opened 2023-11-27
feat(typescript-estree): pass jsDocParsingMode everywhere

bradzacher:
## Overview

@fisker made me realise that we hadn't wired this new option up to every possible spot.
This PR ensures that all locations that we create a program or source file have the new option set to improve our perf.

### DIFF
--- a/packages/typescript-estree/src/create-program/createDefaultProgram.ts
+++ b/packages/typescript-estree/src/create-program/createDefaultProgram.ts
@@ -57,7 +57,10 @@ function createDefaultProgram(
 
   const program = ts.createProgram(
     [parseSettings.filePath],
-    commandLine.options,
+    {
+      ...commandLine.options,
+      jsDocParsingMode: ts.JSDocParsingMode?.ParseForTypeInfo,
+    },
     compilerHost,
   );
   const ast = program.getSourceFile(parseSettings.filePath);
--- a/packages/typescript-estree/src/create-program/createProjectService.ts
+++ b/packages/typescript-estree/src/create-program/createProjectService.ts
@@ -1,6 +1,6 @@
-/* eslint-disable @typescript-eslint/no-empty-function -- for TypeScript APIs*/
 import type * as ts from 'typescript/lib/tsserverlibrary';
 
+// eslint-disable-next-line @typescript-eslint/no-empty-function
 const doNothing = (): void => {};
 
 const createStubFileWatcher = (): ts.FileWatcher => ({
@@ -45,6 +45,6 @@ export function createProjectService(): TypeScriptProjectService {
       startGroup: doNothing,
     },
     session: undefined,
+    jsDocParsingMode: tsserver.JSDocParsingMode?.ParseForTypeInfo,
   });
 }
-/* eslint-enable @typescript-eslint/no-empty-function */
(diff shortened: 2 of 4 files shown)


JoshuaKGoldberg (approved):
⚡

jakebailey:
One day you may want to export this as some sort of setting; if lint rules want to do anything with Program besides query types or access the original `JSDoc` nodes, `ParseForTypeInfo` won't provide the full picture anymore.

bradzacher:
@jakebailey for sure - if we have some concrete usecases that would be an option!
The problem is that plugins run well after the parse+typecheck is done - which means that a plugin cannot control this setting. Instead a user would have to pass it via their eslint config as instructed by the plugin's docs.

> they want to access the original JSDoc nodes

AFAIK the parsed JSDoc AST isn't consumed by anyone (at least nobody's ever mentioned a usecase for it -- people always avoid touching the TS AST if they can).

> if lint rules want to do anything with Program besides query types ... ParseForTypeInfo / ParseNone won't provide the full picture anymore

What part of the picture is missing from the program? Isn't it just some scope analysis / usage information in terms of "this thing was used in JSDoc within a TS file"?

jakebailey:
> a user would have to pass it via their eslint config as instructed by the plugin's docs.

Yep, this is precisely what I meant.

> AFAIK the parsed JSDoc AST isn't consumed by anyone (at least nobody's ever mentioned a usecase for it -- people always avoid touching the TS AST if they can).

I'm still holding out hope to reuse info from TS's parser to make eslint-plugin-jsdoc faster; TS parses the same info way faster and it would have already been doing it unconditionally until the new option.

> What part of the picture is missing from the program? Isn't it just some scope analysis / usage information in terms of "this thing was used in JSDoc within a TS file"?

So concretely, you won't get any info about something being deprecated, because that is part of what's skipped (depending on the mode). That and a rule that could use any other kind of tag.

jakebailey:
Right; Josh had told me that you used something that didn't use the TS info directly before I proposed/implemented the new option.

Mainly that was just an easy example.

The other case is see/link, which aren't parsed in the modes used after this PR, which would be needed to create an edge-case free version of no-unused-var, my main hope for the future.

bradzacher:
Okay now I see it. I had to push through all of the nx caching to see it properly.
This change actually did break `eslint-plugin-deprecation`.

bradzacher:
I'll fix this and we'll do an out-of-band release to get this out soon.

jakebailey:
Ugh, yeah, they are pulling it out of jsdoc nodes via `getJsDocTags`:

https://github.com/gund/eslint-plugin-deprecation/blob/ae8ca953b1b4cc2021525089f56ede97613e9311/src/rules/deprecation.ts#L311-L318

On the TS side, there's no jsdoc parsing mode besides `ParseAll` that would give this info; potentially this indicates that there should be more, or that `deprecated` should be included in others, but it sorta messes with the neat buckets I gave that relate to being able to produce accurate diagnostics or accurate type info (as deprecation is "its own thing" which realistically was about editor feedback).

bradzacher:
Yeah there was a user in discord that also mentioned that they were using a custom JSDoc tag to power a custom lint rule and it OFC broke with this change.

So we can fit a lot of users in a neat box - but we also need to allow people to get out of the box.

PR #7999, opened 2023-11-28
fix: default to parse all JSDoc and provide options to configure it

bradzacher:
## Overview

In hindsight #7997 was sadly a breaking change.
There's a number of usecases that do rely on accessing JSDoc on the TS AST (eg `eslint-plugin-deprecation`).

This PR amends this change so that:
- we accept the setting via parser options so it's user-configurable.
- we default JSDoc parsing to `all` (which was the pre-5.3 behaviour) for type-aware linting.
- we default JSDoc parsing to `none` for non-type-aware-linting.
    - Setting this to `none` by default for this case should provide all non-type-aware-linting users with a free perf boost.
    - This is *technically* a breaking change - but the scope should be zero.
    - The only usecase this impacts is "accessing JSDoc from the TS AST when not using type-aware linting".
    - This usecase is possible - but I don't think it's something anyone would ever do as this would only give you JSDoc for the current file. Most single-file rules assume there's no TS AST.
    - If we do encounter a breakage we can revert this back to `all` - but I don't think we'll see it.

In a future we can consider switching the type-aware default to `'type-info'` and force users to pass the config if and only if they use a rule that requires `'all'`.

FYI @jakebailey / @fisker

### DIFF
--- a/packages/typescript-estree/src/create-program/createProjectService.ts
+++ b/packages/typescript-estree/src/create-program/createProjectService.ts
@@ -9,7 +9,9 @@ const createStubFileWatcher = (): ts.FileWatcher => ({
 
 export type TypeScriptProjectService = ts.server.ProjectService;
 
-export function createProjectService(): TypeScriptProjectService {
+export function createProjectService(
+  jsDocParsingMode?: ts.JSDocParsingMode,
+): TypeScriptProjectService {
   // We import this lazily to avoid its cost for users who don't use the service
   // TODO: Once we drop support for TS<5.3 we can import from "typescript" directly
   const tsserver = require('typescript/lib/tsserverlibrary') as typeof ts;
@@ -45,6 +47,6 @@ export function createProjectService(): TypeScriptProjectService {
       startGroup: doNothing,
     },
     session: undefined,
-    jsDocParsingMode: tsserver.JSDocParsingMode?.ParseForTypeInfo,
+    jsDocParsingMode,
   });
 }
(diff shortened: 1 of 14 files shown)


packages/typescript-estree/src/parseSettings/createParseSettings.ts
```diff
@@ -136,6 +162,17 @@ export function createParseSettings(
     });
   }
 
+  // No type-aware linting which means that cross-file (or even same-file) JSDoc is useless
+  // So in this specific case we default to 'none' if no value was provided
+  if (
+    options.jsDocParsingMode == null &&
+    parseSettings.projects.length === 0 &&
+    parseSettings.programs == null &&
+    parseSettings.EXPERIMENTAL_projectService
```
jakebailey, line 171: Should this be `!parseSettings.EXPERIMENTAL_projectService`?

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -9,7 +9,9 @@ const createStubFileWatcher = (): ts.FileWatcher => ({
 
 export type TypeScriptProjectService = ts.server.ProjectService;
 
-export function createProjectService(): TypeScriptProjectService {
+export function createProjectService(
+  jsDocParsingMode?: ts.JSDocParsingMode,
```
jakebailey, line 13: Preemptive check; IIRC this is public. Should this be some sort of options bag rather than a parameter?

fisker:
Thanks for mentioning, no problem for us.

packages/typescript-estree/src/parseSettings/createParseSettings.ts
bradzacher: goodcatch

packages/typescript-estree/src/create-program/createProjectService.ts
bradzacher: Nah this isn't public - we don't export it from either `index.ts` or `use-at-your-own-risk.ts`.

JoshuaKGoldberg (approved):

bradzacher:
This has been released in an out-of-band patch release https://github.com/typescript-eslint/typescript-eslint/releases/tag/v6.13.1

PR #8011, opened 2023-12-01
feat(eslint-plugin): [prefer-promise-reject-errors] add rule

### LINKED ISSUE #7687: Base rule extension: [prefer-promise-reject-errors]

### Link to the base rule

https://eslint.org/docs/latest/rules/prefer-promise-reject-errors

### Description

the `prefer-promise-reject-errors` rule should be extended as the base rule does not catch cases where the value is not a literal

### Fail

```typescript
const foo = "asdf"
Promise.reject(foo) // no error
```

### Pass

```typescript
const foo = new Error("asdf")
Promise.reject(foo) // no error
```

### Additional Info

originally raised #7673 and https://github.com/eslint/eslint/issues/17592 for `no-throw-literal` because i wasn't aware there was already a base rule for this

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7687

## Overview

I tried to migrate the [original rule](https://github.com/eslint/eslint/blob/main/lib/rules/prefer-promise-reject-errors.js) with only type information added (it allows to detect `PromiseConstructor` like and `Error` like variables)

Also I moved `isErrorLike` function to the `type-utils` package (I hope it's the right place for this function) as its logic may be reused both in `prefer-promise-reject-errors` and `no-throw-literal`

I copied incorrect and correct examples from [`no-throw-literal`](https://typescript-eslint.io/rules/no-throw-literal) because I found them applicable. But maybe there are too many examples for this rule


packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        });
+      `,
+      options: [
+        {
+          allowEmptyReject: true,
+        },
+      ],
+    },
+    'new Promise();',
+    'new Promise(5);',
```
StyleShit, line 84: those 2 are not valid code regardless of the rule

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
```diff
@@ -0,0 +1,227 @@
...
+      declare const foo: Error | undefined
+      Promise.reject(foo)
+    `,
+    `
+      declare const foo: () => Promise<string>
+      Promise.reject(await foo())
+    `,
+    `
+      declare const foo: boolean
+      Promise.reject(foo && new Error())
```
StyleShit, line 159: maybe also add a passing case for `reject(true && new Error())`?

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    'new Promise((resolve, reject) => reject(5))',
+    `
+      new Promise((resolve, reject) => {
+        fs.readFile('foo.txt', (err, file) => {
+          if (err) reject('File not found')
+          else resolve(file)
+        })
+      })
+    `,
+    'new Promise((yes, no) => no(5))',
```
StyleShit, line 193: same as above, maybe add a passing case with the same approach of renaming

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    'new Promise(({foo, bar, baz}, reject) => reject(5))',
+    'new Promise(function(reject, reject) { reject(5) })',
+    'new Promise(function(foo, arguments) { arguments(5) })',
+    'new Promise((foo, arguments) => arguments(5))',
```
StyleShit, line 197: hmm, WDYT about this one (in both valid and invalid cases)?
```TS
'new Promise((...args) => args[1](5))',
```

StyleShit (changes requested):

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
auvred: I just added them to improve test coverage. This is a valid syntax, although it throws an exception in runtime

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
auvred: I thought about it, but I think we can do it in one of the following prs to make it easier to review the current one

Also, if we add support for this case, we would probably need to support a plenty of similar cases such as:

```tsx
new Promise((args) => { const [,reject] = args; reject(); })
```

```tsx
let reject
new Promise((resolve, _reject) => reject = _reject)
reject()
```

```tsx
new Promise(resolve, _reject) => {
  reject = _reject
  reject()
})
```
and so on...

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
```diff
@@ -0,0 +1,233 @@
...
+      declare const foo: {
+        bar: PromiseConstructor
+      }
+      new (foo?.bar)((resolve, reject) => reject(5))
+    `,
+    `
+      const foo = Promise
+      new foo((resolve, reject) => reject(5))
+    `,
+  ].map<TSESLint.InvalidTestCase<MessageIds, Options>>(invalidCase => {
```
JoshuaKGoldberg, line 223: [Testing] We've been moving away from auto-generated tests. They have a habit of being imprecise and hard to statically analyze. We'd also want to be able to customize/see more directly what line & column ranges the errors would be reported on. -1 from me on the `.map`, can you switch to direct object literals please?

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
```diff
@@ -0,0 +1,147 @@
...
+        const rejectVariable = getDeclaredVariables(context, executor).find(
+          variable => variable.identifiers.includes(rejectParamNode),
+        );
+        /* istanbul ignore if */ if (!rejectVariable) {
+          return;
+        }
+
+        rejectVariable.references.forEach(ref => {
+          if (
+            !ref.isRead() ||
```
JoshuaKGoldberg, line 135: [Testing] I tried removing `!ref.isRead() ||` and none of the unit tests failed. Is there a reason to include it that can be tested?

JoshuaKGoldberg (changes requested):
Awesome, really solid work here - thanks for working on it @auvred!

It's interesting that this is the first rule to really need to check whether something is actually a `'PromiseConstructor'`. Others have just approximated comparisons to `Thenable`. 

A few touchups & refactors requested, but the general direction looks great to me! 🚀

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
JoshuaKGoldberg: No need to include these IMO, the rule doesn't really check for them and they'd give a runtime + TypeScript complaint.

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
JoshuaKGoldberg: Yeah I don't think we need to check these kinds of `...args` cases 🙂 that's rather more complex than what most rules try to handle.

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
```diff
@@ -0,0 +1,147 @@
...
+  {
+    allowEmptyReject?: boolean;
+  },
+];
+
+export default createRule<Options, MessageIds>({
+  name: 'prefer-promise-reject-errors',
+  meta: {
+    type: 'suggestion',
+    docs: {
```
JoshuaKGoldberg, line 26: What do you think about recommending this in the `strict-type-checked` config? I think it'd be reasonable to as we already include https://typescript-eslint.io/rules/no-throw-literal in there.

packages/eslint-plugin/docs/rules/prefer-promise-reject-errors.md
```diff
@@ -0,0 +1,164 @@
...
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/prefer-promise-reject-errors** for documentation.
+
+This rule extends the base [`eslint/prefer-promise-reject-errors`](https://eslint.org/docs/rules/prefer-promise-reject-errors) rule.
+It uses type information to guarantee that `Promise` can only be rejected with `Error` objects.
+
+## Examples
+
+<!--tabs-->
```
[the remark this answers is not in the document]
auvred: Yep, thanks! Reduced it https://github.com/typescript-eslint/typescript-eslint/pull/8011/commits/5263fa80ed318f9ae7cf8abbb71c6b3dc91b529f

packages/eslint-plugin/docs/rules/prefer-promise-reject-errors.md
```diff
@@ -0,0 +1,164 @@
...
+}
+Promise.reject(err());
+
+const foo = {
+  bar: '',
+};
+Promise.reject(foo.bar);
+
+// new Promise
+
```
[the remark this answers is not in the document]
auvred: Yep, I implied that these comments would separate the sections. With less count of examples it looks cleaner. Please take a look again, I'm not sure if I should delete these comments

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
```diff
@@ -0,0 +1,147 @@
...
+        for (const declaration of declarations) {
+          const sourceFile = declaration.getSourceFile();
+          if (program.isSourceFileDefaultLibrary(sourceFile)) {
+            return true;
+          }
+        }
+      }
+
+      return false;
+    }
```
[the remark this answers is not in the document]
auvred: Not sure about the function name, but I extracted it: https://github.com/typescript-eslint/typescript-eslint/pull/8011/files#diff-75626015398ba4cc6293d6ec23077170a50f23e306b326a44966e44448c3edf2

---

TODO: refactor these places to use this function
https://github.com/typescript-eslint/typescript-eslint/blob/705370ac0d9c54081657b8855b398e57d6ea4ddb/packages/eslint-plugin/src/rules/no-implied-eval.ts#L82-L90 https://github.com/typescript-eslint/typescript-eslint/blob/705370ac0d9c54081657b8855b398e57d6ea4ddb/packages/eslint-plugin/src/rules/no-implied-eval.ts#L145-L153

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
```diff
@@ -0,0 +1,147 @@
...
+        }
+        const rejectParamNode = executor.params.at(1);
+        if (!rejectParamNode || !isIdentifier(rejectParamNode)) {
+          return;
+        }
+
+        const rejectVariable = getDeclaredVariables(context, executor).find(
+          variable => variable.identifiers.includes(rejectParamNode),
+        );
+        /* istanbul ignore if */ if (!rejectVariable) {
```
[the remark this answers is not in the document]
auvred: Thx! https://github.com/typescript-eslint/typescript-eslint/pull/8011/commits/a680b8df2b2c5472e7fe7c7a639d6849b81188ce

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
auvred: You're right, thanks, check for `CallExpression` is already a check for `isRead`

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
auvred: Done https://github.com/typescript-eslint/typescript-eslint/pull/8011/commits/f2ee5d967496beda88be1e8c25feb317551a5097

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
auvred: Ah if we remove these 2 cases, the coverage is no longer 100% :(
100% coverage


This branch won't be tested: https://github.com/typescript-eslint/typescript-eslint/pull/8011/files#diff-801508874bd1d1cb9f122a9ca577fdf2350bfd756c5e63572487b65b859d6799R112-R115

But if we remove this branch, the rule will throw runtime error during linting `new Promise()`. I suppose it shouldn't break on valid syntax
Is `/* istanbul ignore */` applicable for such a case?

JoshuaKGoldberg: `/* istanbul ignore */` is almost never appropriate (even most or all of our existing cases should probably be replaced with assertions or general-use utilities). 

If the rule would _crash_ in a particular case then I guess including the test makes sense yeah.

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
```diff
@@ -0,0 +1,836 @@
...
+          column: 30,
+          endColumn: 39,
+        },
+      ],
+    },
+    {
+      code: `
+class Foo extends Promise<number> {}
+Foo.reject(5);
+      `,
```
auvred, line 790: Thanks to @Tjstretchalot for pointing on this! https://github.com/typescript-eslint/typescript-eslint/pull/8094#issuecomment-1864771039

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: LGTM! If you want to do that refactor in this PR I'd happily approve. Or we can leave it as a `good first issue` followup. Either way.

packages/eslint-plugin/docs/rules/prefer-promise-reject-errors.md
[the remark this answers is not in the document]
JoshuaKGoldberg: One of the issues with setting a precedent of `//` comment separators is that a lot of rules end up having many different sections of responsibility. It's a slippery slope, if you'll pardon the known fallacy...

The `foo.bar` examples don't really speak to me. They don't introduce any new type concepts. They just show that the rule works in edge cases. What do you think about removing them and then removing section comments?

JoshuaKGoldberg (approved):
This looks lovely, thanks! 🙌 

No blocking requests (other than the _"guarantee"_ nit) from me. Just some touchup suggestions. Feel free to wontfix or do them, and then we can merge. Or if you don't have time I'll take another look soon. Cheers!

packages/eslint-plugin/docs/rules/prefer-promise-reject-errors.md
```diff
@@ -0,0 +1,94 @@
+---
+description: 'Require using Error objects as Promise rejection reasons.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/prefer-promise-reject-errors** for documentation.
+
+This rule extends the base [`eslint/prefer-promise-reject-errors`](https://eslint.org/docs/rules/prefer-promise-reject-errors) rule.
+It uses type information to guarantee that `Promise` can only be rejected with `Error` objects.
```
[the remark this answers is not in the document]
auvred: Thanks!

packages/type-utils/src/isBuiltinSymbolLike.ts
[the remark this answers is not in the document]
auvred: Makes perfect sense! Thank you

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
```diff
@@ -0,0 +1,141 @@
...
+        if (
+          (callee.computed
+            ? callee.property.type === AST_NODE_TYPES.Literal &&
+              callee.property.value !== 'reject'
+            : callee.property.name !== 'reject') ||
+          !(
+            isPromiseConstructorLike(program, calleeObjectType) ||
+            isPromiseLike(program, calleeObjectType)
+          )
+        ) {
```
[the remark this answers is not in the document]
auvred: For now I just split it up into variables, if I can improve that, please tell me

packages/eslint-plugin/docs/rules/prefer-promise-reject-errors.md
```diff
@@ -0,0 +1,94 @@
...
+```ts
+// Promise.reject
+
+Promise.reject(new Error());
+
+try {
+  // ...
+} catch (e) {
+  Promise.reject(e);
+}
```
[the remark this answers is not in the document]
auvred: Nice catch! +1 on asserting that examples under the *Correct* header don't contain any errors

> Perhaps we could also test that code under _**Incorrect**_ tab contains at least one lint error, and _**Correct**_ code contains no errors


https://github.com/typescript-eslint/typescript-eslint/issues/8134

packages/eslint-plugin/docs/rules/prefer-promise-reject-errors.md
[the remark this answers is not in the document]
auvred: Thanks for the explanation, I removed all of them: https://github.com/typescript-eslint/typescript-eslint/pull/8011/commits/eaab9eb552b9bc2b854625f286f7d0418062b27d

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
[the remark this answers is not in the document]
auvred: Let's open a followup with `good first issue` ;)

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Ah splitting to variables in this way and then having the `if` _after_ them is a performance papercut. I'll post a new thread.

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        const rejectMethodCalled = callee.computed
+          ? callee.property.type === AST_NODE_TYPES.Literal &&
+            callee.property.value === 'reject'
+          : callee.property.name === 'reject';
+        const calleeIsLikePromise =
+          isPromiseConstructorLike(services.program, calleeObjectType) ||
+          isPromiseLike(services.program, calleeObjectType);
+
+        if (!rejectMethodCalled || !calleeIsLikePromise) {
```
JoshuaKGoldberg, line 99: [Performance] Previously, when it was all in one `if` statement, the type-checking functions `isPromiseConstructorLike` and `isPromiseLike` wouldn't be called if the previous logical bits (what is now `rejectMethodCalled`) were falsy.

So now it's a bit more readable at the cost of performance. One step forward, one step back!

My intent was to suggest splitting into functions so the logic gets lazily evaluated:

```ts
if (!rejectMethodCalled(callee) || !calleeIsLikePromise(callee)
```

...where `calleeObjectType` is only created inside `calleeIsLikePromise`. That way we don't call to the type checker at all if we don't need. The type checker can be slow sometimes so this is very good.

Aside: I filed an issue on myself to find a lint rule to enforce this: https://github.com/JoshuaKGoldberg/create-typescript-app/issues/1188

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
JoshuaKGoldberg, line 1: [Testing] Sorry for the late request, just realized a few missing cases:

* `Error[]`
* Readonly<Error>`
* `[Error]` (a tuple)
* `never`
* `unknown`

JoshuaKGoldberg (changes requested):
Sorry, two blocking requests:

* The variable readability change actually introduced a performance regression - lmk if the suggestion doesn't make sense?
* Apologies, I missed commenting on a couple of missing test cases

Since I already approved I do feel bad now requesting changes. If you don't have energy+time then no worries. I can apply these soon.

packages/eslint-plugin/src/rules/prefer-promise-reject-errors.ts
auvred: Ohh, don't know how I didn't notice that, thanks! Also thx for detailed explanation, I changed it https://github.com/typescript-eslint/typescript-eslint/pull/8011/commits/f186084686347b0f3850bf8cbb1ad07818e267c2

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
auvred: Nice catch, thanks! https://github.com/typescript-eslint/typescript-eslint/pull/8011/commits/e4af6eda213478034efc29276ae4b62537c1b16f

---

> * Readonly`

_I hope it's `ReadonlyArray<Error>`, right?_

JoshuaKGoldberg: Oop, `Readonly<Error>`

auvred: It was a nice try haha, I will add it tomorrow :)

auvred: `Readonly<Error>` - is actually a great case!

I covered it in https://github.com/typescript-eslint/typescript-eslint/pull/8011/commits/232642964a8677e133ab283b15f096c8050763f8

---

TODO: after merging this PR, we should file an issue for `no-throw-literal` as it doesn't currently cover this case [playground](https://typescript-eslint.io/play/#ts=5.3.3&fileType=.tsx&code=GYVwdgxgLglg9mABMOcAUBTATlgXIgJQwEMATBAGwE8AeAURziwD4BKRAbwChFEoALLHADuibFi4BfIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1vwAtoOA7rUb5E0AIbx0UMf2iRwYAL4glQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

JoshuaKGoldberg (commented):
Super, thanks! This looks great. 🚀 

Neil Patrick Harris giving two thumbs up and grinning

JoshuaKGoldberg (approved):

PR #8019, opened 2023-12-03
chore: enable no-non-null-assertion internally, excluding tests

JoshuaKGoldberg:
## PR Checklist

## Overview

Moves the disable of [@typescript-eslint/no-non-null-assertion](https://typescript-eslint.io/rules/no-non-null-assertion) to test files. Adds a whole bunch of ~disables~ `nullThrows` to the remaining violations. I think we have so many because we're normally pretty good at not adding unnecessary `!`s, so any `!` that was added was generally there for a good reason.

Also adds a unit test to the actually-testable case in `nullThrows`.


JoshuaKGoldberg:
Yeah especially around the token stuff, agreed. Will draft & put in my queue...

Josh-Cena:
FWIW, this is one of the rules I passionately hate, and I think it just adds a lot of noise to most codebases that are not written by novices. I'm okay to have it just for dogfooding, though.

JoshuaKGoldberg:
I'll wait to merge these changes until the flat config work lands.

packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
```diff
@@ -94,7 +94,10 @@ export default createRule<Options, MessageIds>({
                   }
                   // If the property's computed, we have to attach the
                   // annotation after the square bracket, not the enclosed expression
-                  return context.sourceCode.getTokenAfter(node.key)!;
+                  return nullThrows(
+                    context.sourceCode.getTokenAfter(node.key),
+                    NullThrowsReasons.MissingToken(']', 'key'),
+                  );
```
bradzacher, line 100: this pattern is so common - i wonder if we should create a utility to wrap it up and make it even shorter?

eg `getRequiredTokenAfter(sourceCode, node.key, 'optional message')`

JoshuaKGoldberg: I like it!

JoshuaKGoldberg: Hmm, tried this out - but in order to get the same precision of logs we'd need to still pass in `']'` and `key` as well.

```ts
return getRequiredTokenAfter(
  context.sourceCode,
  node.key,
  ']',
  'key',
);
```

I feel like I'd rather get this merged in now, then look at more options in a followup. WDYT?

bradzacher: I wonder if we should just say "fuck it" and create an API without the message.
It makes the end user experience worse cos more vague error messages - but it would be much cleaner and better devx.

JoshuaKGoldberg: I don't really mind having the extra verbosity for our devx, personally. It's been forcing me to think more carefully about what we're actually asserting on.

bradzacher (dismissed):

Issue #8025, opened 2023-12-05
Repo: Netlify PR deploys fail on latest Node.js 18.19.0


auvred:
### Suggestion

Logs: https://app.netlify.com/sites/typescript-eslint/deploys/656cda722b027400084ae549#L64-L91

This is because the latest version 18.x of Node.js (18.19.0) is being used. [It contains breaking changes](https://github.com/nodejs/node/releases/tag/v18.19.0) for `tsx` package used in deployment. It has been fixed on `tsx` side in the [`v4.6.1`](https://github.com/privatenumber/tsx/releases/tag/v4.6.1), but as I understand it, we can't just bump the `tsx` version, because `tsx@4` [doesn't support](https://github.com/privatenumber/tsx/releases/tag/v4.0.0) Node.js v16, although v16 it's still used in ci:
https://github.com/typescript-eslint/typescript-eslint/blob/cc2c6d3a4a515f7ccc23f10421a8db9c88345876/.github/workflows/ci.yml#L162

So we can:
1. Drop Node.js 16 support ([it's already unmaintained](https://github.com/nodejs/release?tab=readme-ov-file#nodejs-release-working-group)) and bump tsx
2. Pin Node.js version to 18.18.2 in Netlify
3. Use something other than `tsx`

I guess the 1 is the best option?

JoshuaKGoldberg:
🤔 per https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript it looks like we've pinned Node to 16 in the UI... 

![Screenshot of Dependency management settings in Netlify showing Node 16](https://github.com/typescript-eslint/typescript-eslint/assets/3335181/c1b6c883-52bd-4223-ae17-be69816cd655)

We wouldn't be able to drop Node.js 16 support until our next major version, as that'd be a breaking change. So for now we can either use something other than `tsx` or pin the Node.js version in Netlify to <=18.18.2.

JoshuaKGoldberg:
Ah but _environment variables_ are used instead - and `NODE_VERSION` was set to `18`. I switched it to `18.18.2` just now. So this _should_ be fixed. Thanks for filing!

If not, we can re-open and take another look.

JoshuaKGoldberg:
Following #8045 -> #8048, we should be able to remove that env variable. I've set a reminder for January 3rd, just in case an older PR doesn't have the updated version of tsx.

PR #8048, opened 2023-12-11
chore(deps): update tsx to 4.6.2

### LINKED ISSUE #8045: Repo: PR check "Run Unit Tests with Experimental TSServer" fails on latest Node.js 18.19.0
### Suggestion

CI failure: https://github.com/typescript-eslint/typescript-eslint/actions/runs/7151333167/job/19475573267?pr=8044#step:3:186
stderr: https://cloud.nx.app/runs/KoN80i59UA/task/repo-tools%3Apostinstall-script

The reason is the same as in https://github.com/typescript-eslint/typescript-eslint/issues/8025

According to https://github.com/typescript-eslint/typescript-eslint/issues/8025#issuecomment-1841499381:
> We wouldn't be able to drop Node.js 16 support until our next major version, as that'd be a breaking change. So for now we can either use something other than tsx or pin the Node.js version in Netlify to <=18.18.2.

We can't just bump tsx. We can do one of the following:

1. Patch `tsx` with changes from https://github.com/privatenumber/tsx/pull/422
2. Use something other than `tsx`
3. Pin Node.js to 18.18.2

I assume the third option is undesirable, because we would have to pin Node.js to `18.18.2` everywhere

If we choose the second option, here is a good starting point: https://github.com/privatenumber/ts-runtime-comparison

I can open two draft PRs for 1 and 2 options, so we can compare which one would be better

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8045

## Overview

All tests pass, I think it's fine

We should also consider that Node.js is already pinned in the Netlify https://github.com/typescript-eslint/typescript-eslint/issues/8025


JoshuaKGoldberg (approved):
Super, thanks!

PR #8065, opened 2023-12-14
feat(eslint-plugin): [no-useless-template-literals] add fix suggestions

### LINKED ISSUE #8056: Enhancement: [no-useless-template-literals] Add fixing suggestions

https://typescript-eslint.io/rules/no-useless-template-literals

### Description

Add suggestions to auto-fix the invalid cases

Splits from #7957

### Fail

```TypeScript
const a = `${ 1 }`;
const b = `${ null }`;
const c = `a-${ true }`;
```

### Pass

```TypeScript
const a = `1`;
const b = `null`;
const c = `a-true`;
```

### Additional Info

_No response_

StyleShit:
Closes #8056

### DIFF
--- a/packages/eslint-plugin/src/rules/no-useless-template-literals.ts
+++ b/packages/eslint-plugin/src/rules/no-useless-template-literals.ts
@@ -10,7 +10,7 @@ import {
   isUndefinedIdentifier,
 } from '../util';
 
-type MessageId = 'noUselessTemplateLiteral';
+type MessageId = 'noUselessTemplateLiteral' | 'removeUselessTemplateLiteral';
 
 export default createRule<[], MessageId>({
   name: 'no-useless-template-literals',
@@ -25,6 +25,9 @@ export default createRule<[], MessageId>({
     messages: {
       noUselessTemplateLiteral:
         'Template literal expression is unnecessary and can be simplified.',
+
+      removeUselessTemplateLiteral:
+        'Remove unnecessary template literal expression.',
     },
     schema: [],
   },
@@ -71,7 +74,7 @@ export default createRule<[], MessageId>({
             messageId: 'noUselessTemplateLiteral',
             suggest: [
               {
-                messageId: 'noUselessTemplateLiteral',
+                messageId: 'removeUselessTemplateLiteral',
                 fix(fixer): TSESLint.RuleFix[] {
                   const [prevQuasi, nextQuasi] = node.quasis;
 
@@ -107,7 +110,7 @@ export default createRule<[], MessageId>({
             messageId: 'noUselessTemplateLiteral',
             suggest: [
               {
-                messageId: 'noUselessTemplateLiteral',
+                messageId: 'removeUselessTemplateLiteral',
                 fix(fixer): TSESLint.RuleFix[] {
                   const index = node.expressions.indexOf(expression);
                   const prevQuasi = node.quasis[index];
--- a/packages/eslint-plugin/src/rules/no-useless-template-literals.ts
+++ b/packages/eslint-plugin/src/rules/no-useless-template-literals.ts
@@ -135,7 +135,12 @@ export default createRule<[], MessageId>({
                     expression.type === AST_NODE_TYPES.Literal;
 
                   if (isStringLiteral) {
-                    fixes.push(fixer.replaceText(expression, expression.value));
+                    const escapedValue = expression.value.replace(
+                      /([`$])/g,
+                      '\\$1',
+                    );
+
+                    fixes.push(fixer.replaceText(expression, escapedValue));
                   }
 
                   return fixes;
--- a/packages/eslint-plugin/src/rules/no-useless-template-literals.ts
+++ b/packages/eslint-plugin/src/rules/no-useless-template-literals.ts
@@ -10,12 +10,12 @@ import {
   isUndefinedIdentifier,
 } from '../util';
 
-type MessageId = 'noUselessTemplateLiteral' | 'removeUselessTemplateLiteral';
+type MessageId = 'noUselessTemplateLiteral';
 
 export default createRule<[], MessageId>({
   name: 'no-useless-template-literals',
   meta: {
-    hasSuggestions: true,
+    fixable: 'code',
     type: 'suggestion',
     docs: {
       description: 'Disallow unnecessary template literals',
@@ -25,9 +25,6 @@ export default createRule<[], MessageId>({
     messages: {
       noUselessTemplateLiteral:
         'Template literal expression is unnecessary and can be simplified.',
-
-      removeUselessTemplateLiteral:
-        'Remove unnecessary template literal expression.',
     },
     schema: [],
   },
@@ -72,27 +69,22 @@ export default createRule<[], MessageId>({
           context.report({
             node: node.expressions[0],
             messageId: 'noUselessTemplateLiteral',
-            suggest: [
-              {
-                messageId: 'removeUselessTemplateLiteral',
-                fix(fixer): TSESLint.RuleFix[] {
-                  const [prevQuasi, nextQuasi] = node.quasis;
+            fix(fixer): TSESLint.RuleFix[] {
+              const [prevQuasi, nextQuasi] = node.quasis;
 
-                  // Remove the quasis and backticks.
-                  return [
-                    fixer.removeRange([
-                      prevQuasi.range[1] - 3,
-                      node.expressions[0].range[0],
-                    ]),
+              // Remove the quasis and backticks.
+              return [
+                fixer.removeRange([
+                  prevQuasi.range[1] - 3,
+                  node.expressions[0].range[0],
+                ]),
 
-                    fixer.removeRange([
-                      node.expressions[0].range[1],
-                      nextQuasi.range[0] + 2,
-                    ]),
-                  ];
-                },
-              },
-            ],
+                fixer.removeRange([
+                  node.expressions[0].range[1],
+                  nextQuasi.range[0] + 2,
+                ]),
+              ];
+            },
           });
 
           return;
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -83,6 +108,40 @@ export default createRule<[], MessageId>({
...
+                  ];
+
+                  // Remove quotes for string literals (i.e. `'a'` will become `a`).
+                  const isStringLiteral =
+                    isUnderlyingTypeString(expression) &&
+                    expression.type === AST_NODE_TYPES.Literal;
+
+                  if (isStringLiteral) {
+                    fixes.push(fixer.replaceText(expression, expression.value));
+                  }
```
auvred, line 139: We need to escape backticks here, because


```ts
`some ${'`'} string`
```

will be fixed to 

```ts
`some ` string`
```

[Playground reproduction](https://deploy-preview-8065--typescript-eslint.netlify.app/play#ts=5.3.2&fileType=.tsx&code=AYZw9gtgpgBAJAbwOTCQXxiALgJwJYB2A5sANxA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tmUSWWS18iALbF4AQ2G1Gw6BPioMkRNGgdokcGAC%2BIHUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

auvred: Ohh, also this one:

```ts
`some ${'${aaa}'} string`
```

will be fixed to 

```ts
`some ${aaa} string`
```

[Playground reproduction](https://deploy-preview-8065--typescript-eslint.netlify.app/play#ts=5.3.2&fileType=.tsx&code=AYZw9gtgpgBAJAbwOSIIboL5IzEAXAJwEsA7Ac2CA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tmUSWWS18iALbF4AQ2G1Gw6BPioMkRNGgdokcGAC%2BIHUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

JoshuaKGoldberg (changes requested):
What @auvred said 😄

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
StyleShit: ooh those are good ones
thanks!

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -68,6 +72,27 @@ export default createRule<[], MessageId>({
           context.report({
             node: node.expressions[0],
             messageId: 'noUselessTemplateLiteral',
+            suggest: [
```
JoshuaKGoldberg, line 75: [Question] Is there a reason to have this as a `suggest`ion, rather than a `fix`? It shouldn't change code or types (roughly the only dividing line between fixes and suggestions) that I can see?

JoshuaKGoldberg (changes requested):
(just asking, not fully in review mode right now)

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
StyleShit: no special reason other than that this is what we've done [here](https://github.com/typescript-eslint/typescript-eslint/pull/7691), so I didn't really put much thought into that.

can probably be a fix 😄 

how do you test multiple fixes in the same line though?

```TS
`${'a'}${true}`;
```

JoshuaKGoldberg: The fixer function can be a generator / return an array of fixes:

https://github.com/typescript-eslint/typescript-eslint/blob/99a026fe41b4b169914352fe66e7d5d687b300bf/packages/utils/src/ts-eslint/Rule.ts#L114-L116

A couple of examples of that:
* `no-unnecessary-boolean-literal-compare` is a `fix: function* (fixer) {`
* `array-type` has a couple fixers that return arrays

Is that what you're looking for?

StyleShit: Yeah, I know it can return an array, but each of those nodes is being reported separately, so I thought that eslint would fix everything before returning the result in the test.

[The docs](https://eslint.org/docs/latest/extend/custom-rules#applying-fixes) state that:

> After applying fixes, ESLint will run all the enabled rules again on the fixed code, potentially applying more fixes. This process will repeat up to 10 times, or until no more fixable problems are found.

So I need to report on all of the nodes together? (if so, it'll put the squiggly line on the whole template literal which is probably not what we want)

JoshuaKGoldberg: Ah gotcha, sorry, I think I see what you mean now. Since each report is about a single violation I think each report should only suggest fixing that violation. They won't overlap ranges so ESLint should be able to apply all the fixes in one pass.

I don't remember whether ESLint will hold off applying fixes that border each other (e.g. how <code>\`${'a'}${'b'}\`</code> might report)... but if that is the case, ESLint would still be able to apply alternating fixes. So it should be no more than two fix passes.

I.e. what I'm suggesting is roughly this psuedocode-ish:

```ts
context.report({
  loc: { end, start },
  fix: (fixer) => replaceRange(end, start),
});
```

StyleShit: if I run it locally, it auto-fixes both of them on save (not sure if it's in a single pass though), so theoretically it should be fine and (IMO) should behave the same in tests, no?

I mean... do we need to change the otherwise-working code only because the tests are not working?
maybe we should fix the rule tester to behave the same as in the real world? (or am I missing something here?)

Not sure what you mean in the pseudo-code BTW

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -83,6 +108,45 @@ export default createRule<[], MessageId>({
...
+                  // Remove quotes for string literals (i.e. `'a'` will become `a`).
+                  const isStringLiteral =
+                    isUnderlyingTypeString(expression) &&
+                    expression.type === AST_NODE_TYPES.Literal;
+
+                  if (isStringLiteral) {
+                    const escapedValue = expression.value.replace(
+                      /([`$])/g,
+                      '\\$1',
+                    );
```
auvred, line 141: [Playground](https://deploy-preview-8065--typescript-eslint.netlify.app/play/#ts=5.3.2&fileType=.tsx&code=AYZw9gtgpgBAJAbwOQB0WIIZYL5OzEAFwCcBLAOwHNgg&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tmUSWWS18iALbF4AQ2G1Gw6BPioMkRNGgdokcGAC%2BIHUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

Input:
```ts
`some ${'\\${aaa}'} string`
```
Output:
```ts
`some \\${aaa} string`
```

---

[Playground](https://deploy-preview-8065--typescript-eslint.netlify.app/play/#ts=5.3.2&fileType=.tsx&code=AYZw9gtgpgBAJAbwOQB0XCQXxiALgJwEsA7Ac2CA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tmUSWWS18iALbF4AQ2G1Gw6BPioMkRNGgdokcGAC%2BIHUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)
Input:
```ts
`some ${'\\`'} string`
```
Output:
```ts
`some \\` string`
```

auvred (commented):
It seems that escaping logic might become very complicated :(

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
StyleShit: haha those are clever ones

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
StyleShit: OK, after an investigation, seems like this is the issue:
https://github.com/typescript-eslint/typescript-eslint/blob/2e7ce16344253e530deddce2ebf044403342dc85/packages/rule-tester/src/utils/SourceCodeFixer.ts#L65-L69

The fixer in the tests _should_ fix everything, but we check for `lastPos >= start`, which in my test cases was true after the first fix.
I considered changing it to `lastPos > start`, but wasn't sure if it's OK or not (only 2 tests were failing, which I think had the same issue)

Anyway, adding a space between the 2 errors seems to fix the issue (pun intended)
(e.g. `${'a'}${'b'}` => `${'a'} ${'b'}`)

JoshuaKGoldberg: Oh, is this a bug in RuleTester then? That might be a separate issue.

StyleShit: I think it is, but it probably requires more investigation, will raise an issue after this PR.
It's ready for review btw, I can't re-request review for some reason

StyleShit:
Bumping despite only 2 weeks have passed only because it still has the `awaiting response` badge even though this PR is ready for review @JoshuaKGoldberg

JoshuaKGoldberg (approved):
💯 awesome! No complaints, this is looking very thorough. Thanks & nice!

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
StyleShit: Filed a new issue:
https://github.com/typescript-eslint/typescript-eslint/issues/8251

PR #6569, opened 2023-03-04
fix(eslint-plugin): [no-unnecessary-boolean-literal-compare] fixer should handle parentheses

armano2:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6567
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

Add missing handling for removal and adding parenthesis in rule fixer

new cases covered:

```ts
if ((x instanceof Error) === false) {}
if (false === (x instanceof Error)) {}
if (x instanceof Error === false) {}
```

old fixer result

```ts
if (!(x instanceof Error) { }
if (!x instanceof Error)) { }
if (!x instanceof Error) { }
```

new fixer result

```ts
if (!(x instanceof Error)) { }
if (!(x instanceof Error)) { }
if (!(x instanceof Error)) { }
```

--------------

initially i was planning to extend range of fixer to include `)(` but after further investigation this  started generating to many edge cases

eg. `(false) === x instanceof Error`

PR #8067, opened 2023-12-14
feat(eslint-plugin): [no-array-delete] add new rule

### LINKED ISSUE #4432: Rule proposal: no-array-delete

StyleShit:
Closes #4432


packages/eslint-plugin/tests/rules/no-array-delete.test.ts
```diff
@@ -0,0 +1,519 @@
...
+              output: `
+        function deleteFromArray<T extends number[]>(a: T) {
+          a.splice(0, 1);
+        }
+      `,
+            },
+          ],
+        },
+      ],
+    },
```
StyleShit, line 517: This test is failing and I have no idea why. Help would be greatly appreciated

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,84 @@
...
+
+type MessageId = 'noArrayDelete' | 'useSplice';
+
+export default createRule<[], MessageId>({
+  name: 'no-array-delete',
+  meta: {
+    hasSuggestions: true,
+    type: 'problem',
+    docs: {
+      description: 'Disallow the delete operator with array expressions',
```
JoshuaKGoldberg, line 15: [Docs] _"expression"_ isn't the right word here, it's really anything that's an array type. That can include identifiers, returned values from calls, etc.

```suggestion
      description: 'Disallow using the `delete` operator on array values',
```

For reference, https://typescript-eslint.io/rules/no-dynamic-delete currently goes with:

> `Disallow using the delete operator on computed key expressions.`

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,84 @@
...
+          return;
+        }
+
+        const shouldHaveParentheses =
+          node.argument.property.type === AST_NODE_TYPES.SequenceExpression;
+
+        const nodeMap = services.esTreeNodeToTSNodeMap;
+        const target = nodeMap.get(node.argument.object).getText();
+        const rawKey = nodeMap.get(node.argument.property).getText();
+        const key = shouldHaveParentheses ? `(${rawKey})` : rawKey;
```
JoshuaKGoldberg, line 63: [Praise] I personally would probably have forgotten to account for these parenthesis 😄 fair play!

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,84 @@
...
+    type: 'problem',
+    docs: {
+      description: 'Disallow the delete operator with array expressions',
+      recommended: 'strict',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      noArrayDelete:
+        'Using the `delete` operator with an array expression is unsafe.',
+      useSplice: 'Use `{{ target }}.splice({{ key }}, 1)` instead.',
```
JoshuaKGoldberg, line 22: [Refactor] We don't generally put nodes themselves in messages. Users like predictable messages and having them jump around in length per the nodes breaks that a bit. Also, if the nodes are super long `box.mySuperLongPropertyIUsedToBeAJavaDeveloper("hello", "etc")` then the messages get funky.

Saying "an array splice" or something like that should suffice.

Even better, once the message no longer has node texts in it, the logic to retrieve them can move to inside the `fix` func†ion.

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
JoshuaKGoldberg, line 1: [Testing] Thoughts on some more test cases:

* `never`
* tuples
* `readonly` arrays
* `ReadonlyArray`
* `ArrayLike`

For `ArrayLike`, I'm wondering whether the rule should flag that 🤔 ... it theoretically doesn't have to be used with actual `Array`s, but I can't imagine a use case where someone would _want_ to be able to `delete` on an `ArrayLike`... what do you think?

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
JoshuaKGoldberg: ✨ `getConstrainedTypeAtLocation` ✨ 

Type parameters are funky because they're not equal to the type they're constrained to. We just conceptually think of them as that type. This trips up rules a lot 😄

JoshuaKGoldberg (changes requested):
A lovely start, thanks for working on this! The foundation is very solid 💪 just some touchups & more testing mostly.

Spongebob ripping off his shirt to reveal an extreme amount of muscles

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
StyleShit: Thank you!

packages/eslint-plugin/src/rules/no-array-delete.ts
StyleShit: sure, done :)

packages/eslint-plugin/src/rules/no-array-delete.ts
StyleShit: haha, I actually saw that in another PR, I'm not _that_ smart

packages/eslint-plugin/src/rules/no-array-delete.ts
StyleShit: that's actually a good reason, done

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
StyleShit: - TIL you can "delete" from `never`
- tuples are good ones, thanks!
- you can't delete from readonly arrays, so not sure what's the test case for that
- never used `ArrayLike`, but IMO if it depends on the implementation (can you handle `delete`s yourself? I know in PHP you can) it shouldn't be flagged, since the whole reason for this rule is the current weird browser implementation

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,90 @@
...
+        );
+
+        if (!isUnderlyingTypeArray(type)) {
+          return;
+        }
+
+        context.report({
+          node,
+          messageId: 'noArrayDelete',
+          fix(fixer): TSESLint.RuleFix | null {
```
JoshuaKGoldberg, line 69: [Bug] Err, why was this switched from `suggestion` to `fix`? As undesirable as the reported behavior is, switching from `delete` to `.splice()` changes code behavior and so can't be a `fix`.

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
JoshuaKGoldberg, line 1: [Testing] Sorry, I thought of a couple more test cases 😅 

* Comments positions: in `delete abc[def]`, both single- and multi-line comments could be before or after each of `abc`, `[`, `def`, and `]`.
* Parenthesis: see `getWrappingFixer` and where it's used. These can be surprisingly tricky! `delete ((abc[((def))]))`

Please enjoy this hellish block of code:

```ts
delete /* gotcha
Gotcha */ abc[(
 // GOTCHA
 def /* GOTCHA! */ /* !!! */ )
) /* GOTCHA !!! */ ];
```

JoshuaKGoldberg (changes requested):
Getting there! A blocking request on the `fix`/`suggestion` difference. And some more testing I thought of just now. Thanks!~

packages/eslint-plugin/src/rules/no-array-delete.ts
StyleShit: hmm
actually, I have no idea why I did this haha
for some reason, I remember us talking about changing it to a fix, but I'm probably mixing it with another PR 😓 

fixed :)

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
StyleShit: I've added those test cases (and other ones that came to my mind), but I'm not wondering:
1. what should we do about the comments in this case?
2. in which cases the `getWrappingFixer` is necessary? I mean... it's not necessary in the above cases, and the code currently works with cases like `delete arr[(doWork(), 1)]` or `delete a[(b + 1) * (b + 2)]`, so when is it being used?

In addition, found this weird case:

```TS
const a = [1, 2, 3];

delete a[(void 1)]; // a = [1, 2, 3]

a.splice((void 1), 1); // a = [2, 3]
```

do we have anything to do about it? or since it's a suggestion we're fine with this behavior?

StyleShit:
not sure why the CI is failing. doesn't seem to be related to this PR 🤷

JoshuaKGoldberg:
Yeah the CI failures are fixed on `main`, sorry about that 😞

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
JoshuaKGoldberg: > what should we do about the comments in this case?

Great question. I'd say as long as they generally are preserved in somewhat near the original location, it's fine. Trying to map user intent from comments to AST nodes is a nigh impossible task.

> in which cases the `getWrappingFixer` is necessary?

...hhmmmm. I'm not sure anymore 😄. It's a generally useful utility for some edge cases but if this rule doesn't come across any of those edge cases then it's fine to not use it I suppose. #6569 is a good example of a fix using `getWrappingFixer`.

> this weird case

This gets a `Type 'undefined' cannot be used as an index type.` TypeScript complaint so it's going to be a pretty rare case for users to hit. Plus the `void` operator is already pretty rarely used. Since it's a suggestion, yes, I think it's fine.

---

Just confirming we're on the page: I'm thinking suggestions should preserve comments. Right now the comments are being erased.

JoshuaKGoldberg (changes requested):
Progress!

packages/eslint-plugin/src/rules/no-array-delete.ts
```diff
@@ -0,0 +1,96 @@
...
+        context.report({
+          node,
+          messageId: 'noArrayDelete',
+          suggest: [
+            {
+              messageId: 'useSplice',
+              fix(fixer): TSESLint.RuleFix | null {
+                if (node.argument.type !== AST_NODE_TYPES.MemberExpression) {
+                  return null;
+                }
```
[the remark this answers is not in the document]
StyleShit: I actually tried it and it didn't work for some reason. now it works 😅 

TS is afraid of you

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
StyleShit: I aggregated all of the comments and put them above the fixed code.
not sure that I like it...

took inspiration from here:
https://github.com/typescript-eslint/typescript-eslint/blob/958fecaef10a26792dc00e936e98cb19fd26d05f/packages/eslint-plugin/src/rules/prefer-function-type.ts#L141-L171

I'm wondering if we can extract it to a util, but it feels like a too-early abstraction, WDYT?

packages/eslint-plugin/src/rules/no-array-delete.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Spongebob laughing evilly over dramatic purple-lit black underwater sky

packages/eslint-plugin/tests/rules/no-array-delete.test.ts
JoshuaKGoldberg: Yeah I probably wouldn't put them in an abstraction just yet. I like "WET" (Write Everything Twice) as a vague methodology - and this is only the second rule that uses `sourceCode.getCommentsInside()`.

PR #5719, opened 2022-10-03
chore(website): add details to Configurations page

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5002
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

Reworks the _Configurations_ page to emphasize the recommended configs. Also mention the `all` and `base` configs alongside `eslint-recommended` as "other" (utility) configs, and adds the Prettified, stringified contents of each config to the page.

PR #8073, opened 2023-12-15
feat(eslint-plugin): deprecate formatting (meta.type: layout) rules

### LINKED ISSUE #8072: 💀 Rule deprecations: formatting rules
## Overview

Following https://github.com/eslint/eslint/issues/17522 > https://eslint.org/blog/2023/10/deprecating-formatting-rules/:

> The next minor release of ESLint will deprecate core formatting rules. We recommend you use a source code formatter instead.

Hooray! 🙌 

We've already closed all issues in this repo that pertain to formatting rules (e.g. https://github.com/typescript-eslint/typescript-eslint/issues/1824#issuecomment-1714716151) and previously added a notice to rule docs not to use them (https://github.com/typescript-eslint/typescript-eslint/pull/7733). Next up will be marking the rules as `deprecated` in their metdata.

To be clear, this is not _removing_ the formatting rules. They'll still work if you really want to go against the recommendation of both ESLint core and the typescript-eslint team. They'll just show up as deprecated in documentation. Removing them will be a breaking change in a later major version.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8072

## Overview

For every rule whose `meta` includes `type: 'layout'`, adds `deprecated: true` and removes the rule from the `all` config.

Note that **changes to the `all` config are explicitly _not_ considered breaking**. Per https://typescript-eslint.io/linting/configs (added in #5719, back in v5):

> With the exception of `all`, `strict`, and `strict-type-checked`, all configurations are considered "stable".


bradzacher (approved):
just to confirm - all the deprecated rules are the ones that have been moved to `eslint-stylistic`, right? Are there any others?

Also accepted but you should mark them all as `replacedBy: ['eslint-stylistic/foo']`.

JoshuaKGoldberg:
> Are there any others?

Just did another pass - I did miss a few. I'd searched on `type: 'layout'` before, but that misses some extension rules that `...` spread from the base rule. https://typescript-eslint.io/rules/?=formatting is a better query.

JoshuaKGoldberg:
> Are there any others?

Just did another pass - I did miss one, `object-curly-spacing`. I'd searched on `type: 'layout'` before, but that misses some extension rules that `...` spread from the base rule. https://typescript-eslint.io/rules/?=formatting is a better query.

👍 on the other two suggestions, good call! Done.

packages/eslint-plugin/tests/areOptionsValid.test.ts
```diff
@@ -4,6 +4,7 @@ import { areOptionsValid } from './areOptionsValid';
 const exampleRule = createRule<['value-a' | 'value-b'], never>({
   name: 'my-example-rule',
   meta: {
+    deprecated: true,
```
bradzacher, line 7: undo sir

packages/eslint-plugin/src/rules/object-curly-spacing.ts
```diff
@@ -24,10 +24,12 @@ export default createRule<Options, MessageIds>({
   // eslint-disable-next-line eslint-plugin/prefer-message-ids,eslint-plugin/require-meta-type,eslint-plugin/require-meta-schema,eslint-plugin/require-meta-fixable -- all in base rule - https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/274
   meta: {
     ...baseRule.meta,
+    deprecated: true,
     docs: {
       description: 'Enforce consistent spacing inside braces',
       extendsBaseRule: true,
     },
+    replacedBy: ['@stylistic/object-curly-spacing'],
```
bradzacher, line 32: OCD eye twitch: all the other ones you added on consecutive lines

packages/website/plugins/generated-rule-docs/insertions/insertFormattingNotice.ts
```diff
@@ -6,8 +6,8 @@ export function insertFormattingNotice(page: RuleDocsPage): void {
   if (page.rule.meta.type === 'layout') {
     page.spliceChildren(0, 0, {
       value: `
-<admonition type="warning">
-This rule will soon be moved to <a href="https://eslint.style">eslint-stylistic</a>.
+<admonition title="Deprecated" type="warning">
+This rule is deprecated. Formatting rules now live in <a href="https://eslint.style">eslint-stylistic</a>.
```
bradzacher, line 10: can we do one better here and link directly to the replacement rule's docs?
I'd assume their URLs are gonna stay static so we can just hard link (cc @antfu so we can ensure redirects are in place should you restructure your URLs).

eg:
```ts
const replacement = page.rule.meta.replacedBy.find(e => e.startsWith('@stylistic/'));
const url = replacement ? `https://eslint.style/rules/default/${replacement.replace('@stylistic/', '')}` : 'https://eslint.style';
// ...
```

bradzacher (approved):

packages/eslint-plugin/src/rules/object-curly-spacing.ts
JoshuaKGoldberg: Heh, the other ones are placed before a `type:`. So they somewhat don't make the sorting less alphabetical. This one doesn't have a `type:`...

PR #8083, opened 2023-12-17
chore: add MDX files to precommit hook

Josh-Cena:
`.mdx` is not formatted in pre-commit and may cause CI to fail.


JoshuaKGoldberg (approved):
Oh! Nice spot.

PR #8084, opened 2023-12-17
feat(eslint-plugin): [naming-convention] support the auto-accessor syntax

### LINKED ISSUE #7823: [naming-convention] support the auto-accessor syntax

### Playground Link

https://typescript-eslint.io/play/#ts=5.2.2&fileType=.tsx&code=KYDwDg9gTgLgBAYwDYEMDOa4FkCeBhVDOAbwFgAoASDCgEsA3FGYOAfSgFclgBJNAUQB2AM2gJgAEzgBeODE7AA3BQrUOAIyS0EcAObB4KBOIzR2XYK1ppWwEWMlwAFAEoSqylAMcoguQAtrADpzbj4hUShxCWUqAF8VKjANLR0jEzRoOABBQRwAJQtw%2ByjJAC44dQgIbhQ-WWEUJDQlCjigA&eslintrc=N4KABGBEAOCGBOBnApvSAuKABALgT2mUQGN4BLaHAWiIBsyA7HAejiVUgBpwppaBXAOaNEGMAG0eESLgJFSFanUY5IPALrdp8frSJjQEabMIlylGonpNmDWAFtGgqsQD2DAG7ImZd2MlGRpCo8K5oWoFghpHSKHrEOGFikMS0sIiiETGQAGZh9rCqmAExQQDKOOQJAArpxLC0AMLpyGqlYOpSRgC%2BWUbR2XHICUmYkLDExESISX2BufmF-l2DdgDWyAD69ShtpZ0xvStRx7HI8YloY-bI9gBGqAAyZBtcp1B58AVFEu-S9Tcmi09jFNO9IHpYAATJwAVQYUNQJDCrTGDVorgA7mUnHoAPLwAAirn4dz0IJ6cxO7UgQxGV1%2B7WkiJysF0qneYJpn2%2ByyZUAB52au05x26XQOYHF3SAA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```typescript
export class MyClass {
	private _ruleIsEnforced = true;

	public get accessor_rule_is_enforced () {
		return this._ruleIsEnforced;
	}

	public accessor AnyRuleIsEnforced: boolean = false;
}
```

### ESLint Config

```javascript
module.exports = {
	'parser': '@typescript-eslint/parser',
	'plugins': [
		'@typescript-eslint'
	],
	'rules': {
		'@typescript-eslint/naming-convention': [
			'error',

			// classes
			{
				selector: 'class',
				format: ['StrictPascalCase'],
			},

			// accessor
			{
				selector: 'accessor',
				format: ['snake_case'],
			},

			// members
			{
				selector: 'memberLike',
				format: ['camelCase'],
				leadingUnderscore: 'allowSingleOrDouble',
			},

			// default
			{
				selector: ['default'],
				format: ['camelCase'],
			},
		],
	}
}
```

### tsconfig

_No response_

### Expected Result

Line 8 should report that the field must be snake_case (if caught by the accessor rule) or camelCase (if caught by the default rule).

### Actual Result

No errors are reported in this file.

### Additional Info

_No response_

arka1002:
Fixes: #7823

## PR Checklist

- [x] Addresses an existing open issue: fixes #7823 

## Overview
Adds auto-accessor in the accessor selector.


arka1002:
I'm looking into the CI failures.

arka1002:
@Josh-Cena can you please take a look.

packages/eslint-plugin/src/rules/naming-convention.ts
JoshuaKGoldberg, line 1: [Docs] There'll need to be a corresponding mention of accessors in `packages/eslint-plugin/docs/rules/naming-convention.md`. That docs page explicitly says what's under each of the rule's selectors.

JoshuaKGoldberg (changes requested):
Functionally looks great, thanks! Nice and straightforward. 🙂 

Just requesting changes to get it documented too. Every selector option should be.

Btw, no need to ping individuals for review. We have a queue and we get to things when we can.

packages/eslint-plugin/src/rules/naming-convention.ts
arka1002: Thanks for the review.

> no need to ping individuals for review

Sorry for that.

I just have 1 doubt, do I have to mention about auto accessors here, in the modifiers of `accessor` point - https://typescript-eslint.io/rules/naming-convention/#individual-selectors. 

What I mean is - will the accessor keyword be a modifier ?

```js
/* eslint "@typescript-eslint/naming-convention": ["error", { "selector": "accessor", "format": ["snake_case"], "modifiers": ["autoAccessor"] } ] */
class Ignored { accessor StrictPascalCase = 10; } // error(?)
```

JoshuaKGoldberg: > do I have to mention about auto accessors here, in the modifiers of `accessor` point

Yup, exactly! The accessor keyword will be a modifier.

---

Amusing aside: the terminology is funky and not ideal anymore. What the rule & its docs call _"accessor"_ is a general term referring to any keyword that changes how a field is accessed. That includes:

* **Inheritance** accessors: `abstract`, `override`,
* **Location** accessor: `static`
* **Visibility** accessors: `private`, `protected`, etc.
* ...a few more along those lines

Interestingly, `static` is the only accessor right now that modifies the runtime behavior. All the other ones are purely type system oriented.

Now that there's a new `accessor` keyword in the language, the previous general _"accessor"_ term is kind of a misnomer... out of scope for this PR to change it though. 🤷 just a confusing point we'll have to live with until or unless we breaking change rename it.

packages/eslint-plugin/docs/rules/naming-convention.md
```diff
@@ -201,7 +201,7 @@ There are two types of selectors, individual selectors, and grouped selectors.
 Individual Selectors match specific, well-defined sets. There is no overlap between each of the individual selectors.
 
 - `accessor` - matches any accessor.
-  - Allowed `modifiers`: `abstract`, `override`, `private`, `protected`, `public`, `requiresQuotes`, `static`.
+  - Allowed `modifiers`: `abstract`, `override`, `private`, `protected`, `public`, `requiresQuotes`, `static`, `autoAccessor`.
```
bradzacher, line 204: It isn't a new modifier - it's a new selector entirely! So it'll need its own dot point listing the allowed modifiers and types.

bradzacher (changes requested):

packages/eslint-plugin/docs/rules/naming-convention.md
JoshuaKGoldberg: Hmm, yeah, I think you're right - I disagree with my my earlier review suggestion of just going along with the existing `accessor` one.

packages/eslint-plugin/docs/rules/naming-convention.md
```diff
@@ -203,6 +203,9 @@ Individual Selectors match specific, well-defined sets. There is no overlap betw
 - `accessor` - matches any accessor.
   - Allowed `modifiers`: `abstract`, `override`, `private`, `protected`, `public`, `requiresQuotes`, `static`.
   - Allowed `types`: `array`, `boolean`, `function`, `number`, `string`.
+- `autoAccessor` - matches any auto-accessor.
+  - Allowed `modifiers`: `abstract`, `override`, `private`, `protected`, `public`, `requiresQuotes`, `static`.
+  - Allowed `types`: `array`, `boolean`, `function`, `number`, `string`.
```
JoshuaKGoldberg, line 208: [Docs] I'm getting confused reviewing this and I'm a maintainer & have read the TC39 specs for much of these... so I can only imagine how mind-boggingly confusing the difference between _"accessor"_ (as the rule already calls things) and _"autoAccessor"_ (what's being added) is. Which is not your "fault" as the PR author, just ... a fact of how convoluted this rule has gotten.

I think this docs page will need some note explaining the nomenclature. People are going to get tripped up on the differences between the two, and get mad at us for having such confusing names. What do you think about explaining them briefly somewhere around this list?

At the very least I think we'd need definitions somewhere of what "accessor" and "autoAccessor" each mean. Many developers won't about the [`accessor` syntax](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#auto-accessors-in-classes).

packages/eslint-plugin/src/rules/naming-convention-utils/enums.ts
```diff
@@ -46,6 +46,9 @@ enum Selectors {
 
   // other
   import = 1 << 18,
+
+  // auto accessor
+  autoAccessor = 1 << 19,
```
JoshuaKGoldberg, line 51: [Style] Nit - this feels like it belongs in `memberLike`. It's used with classes. Thoughts?

packages/eslint-plugin/src/rules/naming-convention.ts
```diff
@@ -538,6 +540,16 @@ export default createRule<Options, MessageIds>({
...
+        validator: validators.autoAccessor,
+        handler: (
+          node: TSESTree.AccessorPropertyNonComputedName,
+          validator,
+        ): void => {
+          const modifiers = getMemberModifiers(node);
+          handleMember(validator, node, modifiers);
+        },
+      },
       // #endregion accessor
```
JoshuaKGoldberg, line 553: [Style] Nit: `#region accessor` refers to the existing accessor stuff. This auto-accessor / `AccessorProperty` should be in a new region IMO.

packages/eslint-plugin/tests/rules/naming-convention/cases/autoAccessor.test.ts
JoshuaKGoldberg, line 1: [Testing] The selector describes being able to take in other modifies (`private`, `readonly`, etc.) - but this test only tests `static`. Missing test coverage?

JoshuaKGoldberg (changes requested):
Progress, thanks for working on it!

I'll defer to @bradzacher for final approval - but I think having more docs and matching test coverage to the new feature would be good.

packages/eslint-plugin/tests/rules/naming-convention/cases/autoAccessor.test.ts
arka1002: Thanks for the reviews.

I think I just found another bug while writing the abstract version of this. https://github.com/typescript-eslint/typescript-eslint/issues/8242

packages/eslint-plugin/src/rules/naming-convention-utils/enums.ts
arka1002: I think it just might be a breaking change.

I mean, for those people who are using `memberLike` grouped accessor and arent following the naming conv for auto-accessor.

But, I dont really know, I'm very new to this btw.

packages/eslint-plugin/src/rules/naming-convention.ts
arka1002: Ye, added.

packages/eslint-plugin/src/rules/naming-convention-utils/enums.ts
bradzacher: As per [our versioning guidelines](https://typescript-eslint.io/users/versioning#eslint-plugin-and-eslint-plugin-tslint)

> A change to the plugins shall not be considered breaking if it:
> ...
> - Adds additional checks to an existing rule that causes new reports in a small-to-medium set of cases in an average codebase.
> ...

We do not classify this as a breaking change as in most codebases there shouldn't be too many violations because (a) auto-accessor is very new and (b) it's likely that the code already follows the naming convention for other members

packages/eslint-plugin/docs/rules/naming-convention.md
bradzacher: I think the distinction is "oaky" if you know the syntax:
- "accessors" are the classic top-level get/set accessors
- "auto accessors" are the members declared with the `accessor` keyword

```ts
class Foo {
  // accessor
  get a() { return 1 }
  set a(val: number) { }

  // auto-accessor field from the decorators proposal (supported in TS)
  accessor b: number = 1;

  // stage 1 auto-accessors proposal (not supported in TS)
  accessor c { get; set }
  accessor d { get; set } = 1;
  accessor e { get() { return 1 } }
  accessor f { set(val) { } }
  accessor g { #get; set }
  // etc
}
```

There's probably scope here to re-define `accessor` as a group selector that encompasses `autoAccessor`.

EG like:
`classicAccessor` = just the top-level, non-keyworded accessor syntax
`autoAccessor` = the new keyworded syntax
`accessor` = group of `classicAccessor | autoAccessor`

WDYT?

packages/eslint-plugin/tests/rules/naming-convention/cases/autoAccessor.test.ts
bradzacher: @arka1002 we can definitely include coverage for that within this PR - there's no reason we need to split it out into two changes, I think.

If you want to tackle it as a separate change after this one - that's okay too - just let us know!

bradzacher:
Awaiting response as two comments to reply to:
- https://github.com/typescript-eslint/typescript-eslint/pull/8084#discussion_r1451125434
- https://github.com/typescript-eslint/typescript-eslint/pull/8084#discussion_r1451126481

packages/eslint-plugin/docs/rules/naming-convention.md
arka1002: > re-define `accessor` as a group selector

Ye, nice point.

packages/eslint-plugin/tests/rules/naming-convention/cases/autoAccessor.test.ts
arka1002: Ye, I'll do in this pr.

JoshuaKGoldberg:
Adding back `awaiting response` as we haven't been explicitly re-requested for review. But if if / when this is ready, happy to take another look!

arka1002:
@JoshuaKGoldberg I don't have the button to re-request review :( 
Reviewer tab ss
From my side, changes are done, this might be worth a discussion - https://github.com/typescript-eslint/typescript-eslint/pull/8084#discussion_r1451125434 - wdyt ?

packages/eslint-plugin/docs/rules/naming-convention.md
JoshuaKGoldberg: Brad's suggestion sounds good to me!

packages/eslint-plugin/src/rules/naming-convention-utils/schema.ts
```diff
@@ -290,6 +290,24 @@ const SCHEMA: JSONSchema.JSONSchema4 = {
...
+      ]),
+      ...selectorSchema('autoAccessor', true, [
+        'abstract',
+        'private',
+        'protected',
+        'public',
+        'requiresQuotes',
+        'static',
+        'override',
+      ]),
```
JoshuaKGoldberg, line 310: From https://github.com/typescript-eslint/typescript-eslint/pull/8084#discussion_r1451125434:

> EG like:
> classicAccessor = just the top-level, non-keyworded accessor syntax
> autoAccessor = the new keyworded syntax
> accessor = group of classicAccessor | autoAccessor

So I'd think `autoAccessor` would have just the new keyworded syntax, _not_ the old keywords, no?

packages/eslint-plugin/tests/rules/naming-convention/cases/classicAccessor.test.ts
JoshuaKGoldberg, line 1: [Testing] It's good to have created a test file for `classicAccessor` - but there should also be a test case file for `accessor` too! (so, three `*.test.ts` files around accessors, not two)

JoshuaKGoldberg (changes requested):
I think the direction is good - but the two new sub-groups shouldn't have overlapping matches, no?

JoshuaKGoldberg:
> I don't have the button to re-request review :(

Oh, I think I got lost in the GitHub UI - I see you did re-request review from at least me. Sorry about that!

packages/eslint-plugin/src/rules/naming-convention-utils/schema.ts
arka1002: `abstract`, `private`, `static` -- they all are 'modifiers'.

A 'selector' has multiple modifiers under it.

`autoAccessor` is an 'individual selector'(same as `classicAccessor`)

`accessor` is a 'group selector'. 

& I did that because ... say the modifiers which are applicable to `get` syntax, shouldn't they be also applicable to the `accessor` syntax ? 

`static accessor` is a thing ig ? https://github.com/tc39/proposal-decorators/blob/master/README.md#class-auto-accessors , see 3rd code snippet.

packages/eslint-plugin/tests/rules/naming-convention/cases/classicAccessor.test.ts
arka1002: Ah yes! Added here [`11d81c1`](https://github.com/typescript-eslint/typescript-eslint/pull/8084/commits/11d81c1162b85df30f04470f8a049df5a7b3cf01). There are a bunch of tests in the other file too https://github.com/typescript-eslint/typescript-eslint/blob/1200b4c4929d28bf6eed58dcfa4c6f40153022eb/packages/eslint-plugin/tests/rules/naming-convention/naming-convention.test.ts#L610-L625

packages/eslint-plugin/docs/rules/naming-convention.md
```diff
@@ -277,6 +280,9 @@ Group Selectors are provided for convenience, and essentially bundle up sets of
 - `variableLike` - matches the same as `function`, `parameter` and `variable`.
   - Allowed `modifiers`: `async`, `unused`.
   - Allowed `types`: none.
+- `accessor` - matches the same as `classicAccessor` and `autoAccessor`.
+  - Allowed `modifiers`: `abstract`, `override`, `private`, `protected`, `public`, `requiresQuotes`, `static`.
+  - Allowed `types`: `array`, `boolean`, `function`, `number`, `string`.
```
JoshuaKGoldberg, line 285: [Style] The list is in alphabetical order to make it easier to scan through. Let's move this to the top.

JoshuaKGoldberg (approved):
Whoohoo, this is lovely! 🙌 

Thanks for working on it & iterating with us @arka1002. Really nicely done final result here. 

I just have the one docs request, but nothing left from me in code. Will defer to @bradzacher. 🙂

packages/eslint-plugin/src/rules/naming-convention.ts
```diff
@@ -519,27 +523,48 @@ export default createRule<Options, MessageIds>({
...
-            node: TSESTree.MethodDefinitionNonComputedName,
-            validator,
-          ): void => {
-            const modifiers = getMemberModifiers(node);
-            handleMember(validator, node, modifiers);
-          },
+      [[
+        'MethodDefinition[computed = false]:matches([kind = "get"], [kind = "set"])',
+        ':matches(TSAbstractMethodDefinition)[computed = false][kind="get"]',
+        ':matches(TSAbstractMethodDefinition)[computed = false][kind="set"]',
```
bradzacher, line 536:
```suggestion
        'TSAbstractMethodDefinition[computed = false]:matches([kind="get"], [kind="set"])'
```

- `:matches` accepts a list of things and asserts the node matches at least one of them. So `:matches(T)` is the same as just `T`.
- We can use `:matches` to join these two lines, as in the previous line.

bradzacher (commented):
one small change otherwise LGTM

PR #8089, opened 2023-12-18
fix(eslint-plugin): [no-unnecesary-type-assertion] treat unknown/any as nullable

### LINKED ISSUE #1082: [no-unnecessary-type-assertion] False negative on unknown!
**Repro**

[Playground](https://typescript-eslint.io/play#ts=5.3.2&fileType=.tsx&code=KYDwDg9gTgLgBAYwDYEMDOa4EE4G8BQccAJijCgPwBccArgHYDW9EA7vQNz4C%2B%2B%2BokWHHQBPegjgAzBghgBLCPTgxgaGAAoAlDQAKUCAFs5aYAB4GzNvQB8eQnCTB4KGjgC8cesFbYtXIlBOtFBKKAB0pOQAhFy8QA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0YANrhsOOdA7RIGiAF0NAXxAmgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

```JSON
{
  "rules": {
    "@typescript-eslint/no-unnecessary-type-assertion": ["error"]
  }
}
```

```TS
export class A {
  data?: unknown;
}

export async function test(): Promise<unknown> {
  let a: A = new A();
  return a.data!;
}

```

**Expected Result**
no errors

**Actual Result**
This assertion is unnecessary since it does not change the type of the expression. eslint(@typescript-eslint/no-unnecessary-type-assertion)

**Additional Info**

**Versions**

| `@typescript-eslint/eslint-plugin` | `2.3.3` |
| `@typescript-eslint/parser`        | `2.3.3` |
| `node`                             | `12.10.0` |
| `npm`                              | `6.10.3` |

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1082

## Overview

Please tell me if I missed anything, but it turns out we aren't using either of the options of `isNullableType`, and I'm not sure if they really matter, so I've removed them. Please let me know if they need to be kept for compatibility (I don't think they are part of the API surface but who knows)?

This change also touches several other rules, but their tests pass anyway, and from a rough look there shouldn't be extra errors.


JoshuaKGoldberg:
Yeah great point on the `type-utils` change... I don't know myself 😄 filed #8093.

packages/type-utils/src/predicates.ts
```diff
@@ -8,25 +8,18 @@ const log = debug('typescript-eslint:eslint-plugin:utils:types');
...
-  }
-  return (flags & ts.TypeFlags.Null) !== 0;
+  return (
+    (flags &
+      (ts.TypeFlags.Any |
+        ts.TypeFlags.Unknown |
+        ts.TypeFlags.Null |
+        ts.TypeFlags.Undefined)) !==
+    0
+  );
```
StyleShit, line 22: [Curious]

what's the difference between this and `isTypeFlagSet`?

```TS
return isTypeFlagSet( 
    ts.TypeFlags.Any |
    ts.TypeFlags.Unknown |
    ts.TypeFlags.Null |
    ts.TypeFlags.Undefined
);
```

JoshuaKGoldberg: https://github.com/JoshuaKGoldberg/ts-api-utils/blob/7d515a4a8ff25ec1aeafc3aeeaa874172c69827e/src/flags.ts#L104 -> https://github.com/JoshuaKGoldberg/ts-api-utils/blob/7d515a4a8ff25ec1aeafc3aeeaa874172c69827e/src/flags.ts#L10

Looks like none?

StyleShit: I was actually looking here:
https://github.com/typescript-eslint/typescript-eslint/blob/6128a02cb15d500fe22fe265c83e4d7a73ae52c3/packages/type-utils/src/typeFlagUtils.ts#L27

so isn't it more readable? 🤷

Josh-Cena: Yeah, I already did a lot of refactors but didn't go the last mile 🙂

bradzacher:
The package is part of our public API. We need to keep backwards compatibility.

Josh-Cena:
It is poorly documented and I don't even understand what `isReceiver` is supposed to do. Can we keep `isReceiver` in the type but make it do nothing?

bradzacher:
It's been a long long time since https://github.com/typescript-eslint/typescript-eslint/pull/478 so I can't remember the full context for why it was added.

I believe the idea was that in a receiving location (eg the assignment target - the thing that's on the "receiving" end of an assignment) the util would specifically treat any as a "null". Based on the type flags any isn't null IIRC so it needs separate handling.

I think I went in a different direction and so the flag wasn't ever used.

bradzacher (dismissed):

Issue #1804, opened 2020-03-26
[await-thenable] Unexpected `await` of a non-Promise does not work if wrapped with Promise.all


viestat:
**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/await-thenable": "error"
  }
}
```

```TS
const value = 'value';
const createValue = () => 'value';

await Promise.all([value, createValue]);
```

**Expected Result**
The rule should show a linting error:
```
Unexpected `await` of a non-Promise (non-"Thenable") value  @typescript-eslint/await-thenable
```

**Actual Result**
No liniting error.

**Additional Info**

**Versions**

| package                            | version |
| ---------------------------------- | ------- |
| `@typescript-eslint/eslint-plugin` | `2.24.0` |
| `@typescript-eslint/parser`        | `2.24.0` |
| `TypeScript`                       | `3.8.3` |
| `ESLint`                           | `6.8.0` |
| `node`                             | `12.10.0` |
| `npm`                              | `6.10.3` |

bradzacher:
I am unable to reproduce this against master

![image](https://user-images.githubusercontent.com/7462525/77668447-a7add980-6f40-11ea-9cee-3873441aeb10.png)


Could you please provide a repro repo for this to help me investigate?

viestat:
@bradzacher I will create the repro repo, In the meantime, from the image you shared I can see that you added this test case in the **valid** property,  the issue is that the code should be **invalid** and throw an error.

bradzacher:
wait sorry, I misunderstood your OP.

Why should it be invalid?
`Promise.all` returns a `Promise` always, regardless of the input values:
![image](https://user-images.githubusercontent.com/7462525/77673281-345b9600-6f47-11ea-989e-79aa21f38cce.png)
Which means that it should always be awaited.

viestat:
> Why should it be invalid?
> `Promise.all` returns a `Promise` always, regardless of the input values 

You are right,  While it is valid JavaScript to pass non-promise elements to `Promise.all` it seems to be inconsistent with this rule.
It would prevent code like the following to be written:
```TS
const promise1 = Promise.resolve();
const promise2 = Promise.resolve();
const test1 = (): void => {};

export const run = async (): Promise<void> => {
  await Promise.all([promise1, promise2, test1()]);
};
```
Here it makes no sense at all to use `await` for all the input values since one of them is not a promise.

bradzacher:
[...]

PR #8033, opened 2023-12-07
docs: base Testing Rules documentation

c0sta:
## PR Checklist

- [X] Addresses an existing open issue: fixes #4875 
- [X] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [X] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview
Fixes: #4875 

Submitting this as a draft to gather some suggestions/ideas and understand if I'm going the right way here. 

I did all these steps inside [Local Linking](https://typescript-eslint.io/contributing/local-development/local-linking) and I was able to see my console.log running at my downstream repository, from what I've tried these steps are very straightforward so I'm not sure if there any more steps or more examples that I could point out here.

Tonight I aim to review and be more descriptive on the "run your `lint` script" step because it seems kinda "empty".

Issue #7318, opened 2023-07-26
Configs: Allow strict configs to have more strict options than recommended?


JoshuaKGoldberg:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22preset+config+change%22+) and found none that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Description

Coming over from comment thread https://github.com/typescript-eslint/typescript-eslint/pull/7110#discussion_r1271360631: rules are occasionally in a situation where:

* Many users in the _recommended_ configs don't want to use their full options, as they're more "strict" than common use cases
* We still want to enable the strict options for users, especially those on the _strict_ configs

What if we added to the `defaultOptions` and/or `meta.docs` sections of rules so that different configs can have different default options?

### Impacted Configurations

* `strict`
* `strict-type-checked`

### Additional Info

The example from the thread is https://typescript-eslint.io/rules/restrict-plus-operands/#options.

Zamiell:
This sounds like a must-do!

As a user who wants the maximally strict linting config, I'd be super interested to know which rules that I'm using that are inherited from `strict` that aren't maximally strictly configured. To start with, can we just document the specific rules that are included in `recommended` and could be stricter so that people can fix it themselves?

JoshuaKGoldberg:
> document the specific rules that have dumbed-down options? 

Hmm, very interesting. I like the root ask of knowing which rules have opt-in "more strict" options...

bradzacher:
I'm of two minds - I like the idea but I also dislike it because it does make things quite a bit more complicated to define and strictly type. The options type we use is not for the config inputs - it's for the "after defaults" merging - which is one reason we haven't added strict types for user configs with our options types.

The last thing I'd want is for us to define a stricter config, roll it out, then find we misconfigured it and it's broken!

JoshuaKGoldberg:
Here's a list of rules I'd propose:

* [`ban-ts-comment`](https://typescript-eslint.io/rules/ban-ts-comment): `minimumDescriptionLength` could be raised from 3
* [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises): `ignoreVoid` could switch from `true` to `false`
[...]

Issue #6701, opened 2023-03-20
Repo: Merge codecov report from ast-spec into typescript-estree's report


JoshuaKGoldberg:
### Suggestion

#6240 shows an example of a PR where changes in one package are given unit test coverage by changes in another package. Is that something we want factored into the codecov reports? At first I wanted it, but now I'm wondering if we should consider coverage just from within each package's own unit tests...

If we do want this, from @RebeccaStevens's https://github.com/typescript-eslint/typescript-eslint/pull/6240#discussion_r1142216814:

> Codecov can merge multiple reports. I've already set this up in [`ts-api-utils`](https://github.com/JoshuaKGoldberg/ts-api-utils/blob/c8e570dded3f4def7b040d13eb1eb10f3d7b1858/.github/workflows/test.yml#L38) to merge the reports from different TypeScript versions.
> 
> See https://docs.codecov.com/docs/flags for more details on how it works.

bradzacher:
I think the only one we want to merge would be tests from `ast-spec` covering the code in `typescript-estree` - because those are the actual tests for that package!

However I don't think we would want our utility packages to be "covered" by the tests in `eslint-plugin` - implicit tests via usage seem like they aren't a great marker for test coverage.

PR #8094, opened 2023-12-18
feat(eslint-plugin): [thenable-in-promise-aggregators] disallow non-Thenables to promise aggregators

Tjstretchalot:
## PR Checklist

- [X] Addresses an existing open issue: adds requested feature #1804

## Overview

Adds a new rule, `thenable-in-promise-aggregators`, which verifies that array arguments of `Promise.all`, `Promise.allSettled`, and `Promise.race` are all Thenable. The purpose behind this PR is discussed in the issue; for me, I often wrap promises to make them cancelable, but it still makes sense to name the object e.g. `fooPromise`, and when mixed with regular promises (or even without), I will type `Promise.race([stdPromise, fooPromise])`, which will look correct and appear to behave correctly for a long time, before getting non-responsive if one of the promises is unexpectedly slow.

### DIFF
--- a/packages/eslint-plugin/docs/rules/thenable-in-promise-aggregators.md
+++ b/packages/eslint-plugin/docs/rules/thenable-in-promise-aggregators.md
@@ -9,7 +9,7 @@ description: 'Disallow passing non-Thenable values to promise aggregators.'
 A "Thenable" value is an object which has a `then` method, such as a Promise.
 The `await` keyword is generally used to retrieve the result of calling a Thenable's `then` method.
 
-When multiple Thenable's are running at the same time, it is sometimes desirable to wait until any one of them resolves (`Promise.race`), all of them resolve or any of them reject (`Promise.all`), or all of them resolve or reject (`Promise.allSettled`).
+When multiple Thenables are running at the same time, it is sometimes desirable to wait until any one of them resolves (`Promise.race`), all of them resolve or any of them reject (`Promise.all`), or all of them resolve or reject (`Promise.allSettled`).
 
 Each of these functions accept an iterable of promises as input and return a single Promise.
 If a non-Thenable is passed, it is ignored.
@@ -43,6 +43,6 @@ await Promise.race([
 
 ## When Not To Use It
 
-If you want to allow code to use `Promise.race`, `Promise.all`, or `Promise.allSettled` on arrays of non-promise values.
+If you want to allow code to use `Promise.race`, `Promise.all`, or `Promise.allSettled` on arrays of non-Thenable values.
 This is generally not preferred but can sometimes be useful for visual consistency.
 You might consider using [ESLint disable comments](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) for those specific situations instead of completely disabling this rule.
--- a/packages/eslint-plugin/docs/rules/thenable-in-promise-aggregators.md
+++ b/packages/eslint-plugin/docs/rules/thenable-in-promise-aggregators.md
@@ -1,0 +1,49 @@
+---
+description: 'Disallow passing non-Thenable values to promise aggregators.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/thenable-in-promise-aggregators** for documentation.
+
+A "Thenable" value is an object which has a `then` method, such as a Promise.
+The `await` keyword is generally used to retrieve the result of calling a Thenable's `then` method.
+
+When multiple Thenable's are running at the same time, it is sometimes desirable to wait until any one of them resolves (`Promise.race`), all of them resolve or any of them reject (`Promise.all`), or all of them resolve or reject (`Promise.allSettled`).
+
+Each of these functions accept an iterable of promises as input and return a single
+Promise.
+If a non-Thenable is passed, it is ignored.
+While doing so is valid JavaScript, it is often a programmer error, such as forgetting to unwrap a wrapped promise, or using the `await` keyword on the individual promises, which defeats the purpose of using one of these Promise aggregators.
+
+## Examples
+
+<!--tabs-->
+
+### ❌ Incorrect
+
+```ts
+await Promise.race(['value1', 'value2']);
+
+await Promise.race([
+  await new Promise(resolve => setTimeout(resolve, 3000)),
+  await new Promise(resolve => setTimeout(resolve, 6000)),
+]);
+```
+
+### ✅ Correct
+
+```ts
+await Promise.race([Promise.resolve('value1'), Promise.resolve('value2')]);
+
+await Promise.race([
+  new Promise(resolve => setTimeout(resolve, 3000)),
+  new Promise(resolve => setTimeout(resolve, 6000)),
+]);
+```
+
+## When Not To Use It
+
+If you want to allow code to use `Promise.race`, `Promise.all`, or `Promise.allSettled` on arrays of non-promise values.
+This is generally not preferred but can sometimes be useful for visual consistency.
+You might consider using [ESLint disable comments](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) for those specific situations instead of completely disabling this rule.
--- a/packages/eslint-plugin/docs/rules/thenable-in-promise-aggregators.md
+++ b/packages/eslint-plugin/docs/rules/thenable-in-promise-aggregators.md
@@ -11,8 +11,7 @@ The `await` keyword is generally used to retrieve the result of calling a Thenab
 
 When multiple Thenable's are running at the same time, it is sometimes desirable to wait until any one of them resolves (`Promise.race`), all of them resolve or any of them reject (`Promise.all`), or all of them resolve or reject (`Promise.allSettled`).
 
-Each of these functions accept an iterable of promises as input and return a single
-Promise.
+Each of these functions accept an iterable of promises as input and return a single Promise.
 If a non-Thenable is passed, it is ignored.
 While doing so is valid JavaScript, it is often a programmer error, such as forgetting to unwrap a wrapped promise, or using the `await` keyword on the individual promises, which defeats the purpose of using one of these Promise aggregators.
 
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -1,8 +1,11 @@
-import * as tsutils from 'ts-api-utils';
 import {
   isTypeAnyType,
   isTypeUnknownType,
 } from '@typescript-eslint/type-utils';
+import type { TSESTree } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+import * as tsutils from 'ts-api-utils';
+
 import { createRule, getParserServices } from '../util';
 
 export default createRule({
@@ -31,17 +34,17 @@ export default createRule({
     const checker = services.program.getTypeChecker();
 
     return {
-      CallExpression(node): void {
-        if (node.callee.type !== 'MemberExpression') {
+      CallExpression(node: TSESTree.CallExpression): void {
+        if (node.callee.type !== AST_NODE_TYPES.MemberExpression) {
           return;
         }
-        if (node.callee.object.type !== 'Identifier') {
+        if (node.callee.object.type !== AST_NODE_TYPES.Identifier) {
           return;
         }
         if (node.callee.object.name !== 'Promise') {
           return;
         }
-        if (node.callee.property.type !== 'Identifier') {
+        if (node.callee.property.type !== AST_NODE_TYPES.Identifier) {
           return;
         }
 
@@ -56,14 +59,14 @@ export default createRule({
         }
 
         const arg = args[0];
-        if (arg.type === 'ArrayExpression') {
+        if (arg.type === AST_NODE_TYPES.ArrayExpression) {
           const { elements } = arg;
           if (elements.length === 0) {
             return;
           }
 
           for (const element of elements) {
-            if (element === null) {
+            if (element == null) {
               continue;
             }
             const elementType = services.getTypeAtLocation(element);
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -5,7 +5,8 @@ import {
 import type { TSESTree } from '@typescript-eslint/utils';
 import { AST_NODE_TYPES } from '@typescript-eslint/utils';
 import * as tsutils from 'ts-api-utils';
-import ts from 'typescript';
+import * as ts from 'typescript';
+
 import { createRule, getParserServices } from '../util';
 
 export default createRule({
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -198,12 +198,11 @@ export default createRule({
         }
 
         if (checker.isArrayType(argType)) {
-          if (argType.typeArguments === undefined) {
-            return;
-          }
-
-          if (argType.typeArguments.length < 1) {
-            return;
+          if (
+            argType.typeArguments == null ||
+            argType.typeArguments.length < 1
+          ) {
+            throw new Error('Expected to find type arguments for an array.');
           }
 
           const typeArg = argType.typeArguments[0];
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -185,20 +185,15 @@ export default createRule({
               node: element,
             });
           }
-        } else {
-          const argType = services.getTypeAtLocation(arg);
-          if (isTypeAnyType(argType) || isTypeUnknownType(argType)) {
-            return;
-          }
+          return;
+        }
 
-          if (!checker.isArrayType(argType)) {
-            context.report({
-              messageId: 'nonArrayArg',
-              node: arg,
-            });
-            return;
-          }
+        const argType = services.getTypeAtLocation(arg);
+        if (isTypeAnyType(argType) || isTypeUnknownType(argType)) {
+          return;
+        }
 
+        if (checker.isArrayType(argType)) {
           if (argType.typeArguments === undefined) {
             return;
           }
@@ -221,7 +216,38 @@ export default createRule({
             messageId: 'arrayArg',
             node: arg,
           });
+          return;
         }
+
+        if (checker.isTupleType(argType)) {
+          if (argType.typeArguments === undefined) {
+            return;
+          }
+
+          const originalNode = services.esTreeNodeToTSNodeMap.get(arg);
+          for (const typeArg of argType.typeArguments) {
+            if (isTypeAnyType(typeArg) || isTypeUnknownType(typeArg)) {
+              continue;
+            }
+
+            if (tsutils.isThenableType(checker, originalNode, typeArg)) {
+              continue;
+            }
+
+            context.report({
+              messageId: 'arrayArg',
+              node: arg,
+            });
+            return;
+          }
+          return;
+        }
+
+        context.report({
+          messageId: 'nonArrayArg',
+          node: arg,
+        });
+        return;
       },
     };
   },
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -33,7 +33,7 @@ export default createRule({
     const services = getParserServices(context);
     const checker = services.program.getTypeChecker();
 
-    const aggregateFunctionNames = ['all', 'race', 'allSettled'];
+    const aggregateFunctionNames = ['all', 'race', 'allSettled', 'any'];
 
     function skipChainExpression<T extends TSESTree.Node>(
       node: T,
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -5,7 +5,7 @@ import {
 import type { TSESTree } from '@typescript-eslint/utils';
 import { AST_NODE_TYPES } from '@typescript-eslint/utils';
 import * as tsutils from 'ts-api-utils';
-
+import ts from 'typescript';
 import { createRule, getParserServices } from '../util';
 
 export default createRule({
@@ -33,28 +33,129 @@ export default createRule({
     const services = getParserServices(context);
     const checker = services.program.getTypeChecker();
 
+    const aggregateFunctionNames = ['all', 'race', 'allSettled'];
+
+    function skipChainExpression<T extends TSESTree.Node>(
+      node: T,
+    ): T | TSESTree.ChainElement {
+      return node.type === AST_NODE_TYPES.ChainExpression
+        ? node.expression
+        : node;
+    }
+
+    function isSymbolFromDefaultLibrary(
+      program: ts.Program,
+      symbol: ts.Symbol | undefined,
+    ): boolean {
+      if (!symbol) {
+        return false;
+      }
+
+      const declarations = symbol.getDeclarations() ?? [];
+      for (const declaration of declarations) {
+        const sourceFile = declaration.getSourceFile();
+        if (program.isSourceFileDefaultLibrary(sourceFile)) {
+          return true;
+        }
+      }
+
+      return false;
+    }
+
+    // Is like Promise, or a class/interface extending from Promise
+    function isPromiseClassLike(program: ts.Program, type: ts.Type): boolean {
+      if (type.isIntersection()) {
+        return type.types.some(t => isPromiseLike(program, t));
+      }
+      if (type.isUnion()) {
+        return type.types.every(t => isPromiseLike(program, t));
+      }
+
+      const symbol = type.getSymbol();
+      if (!symbol) {
+        return false;
+      }
+
+      if (
+        symbol.getName() === 'Promise' &&
+        isSymbolFromDefaultLibrary(program, symbol)
+      ) {
+        return true;
+      }
+
+      if (symbol.flags & (ts.SymbolFlags.Class | ts.SymbolFlags.Interface)) {
+        const checker = program.getTypeChecker();
+
+        for (const baseType of checker.getBaseTypes(type as ts.InterfaceType)) {
+          if (isPromiseLike(program, baseType)) {
+            return true;
+          }
+        }
+      }
+
+      return false;
+    }
+
+    // Is like PromiseConstructor, or a class/interface extending from Promise
+    function isPromiseLike(program: ts.Program, type: ts.Type): boolean {
+      if (type.isIntersection()) {
+        return type.types.some(t => isPromiseLike(program, t));
+      }
+      if (type.isUnion()) {
+        return type.types.every(t => isPromiseLike(program, t));
+      }
+
+      const symbol = type.getSymbol();
+      if (!symbol) {
+        return false;
+      }
+
+      if (
+        symbol.getName() === 'PromiseConstructor' &&
+        isSymbolFromDefaultLibrary(program, symbol)
+      ) {
+        return true;
+      }
+
+      if (symbol.flags & (ts.SymbolFlags.Class | ts.SymbolFlags.Interface)) {
+        const checker = program.getTypeChecker();
+
+        for (const baseType of checker.getBaseTypes(type as ts.InterfaceType)) {
+          if (isPromiseClassLike(program, baseType)) {
+            return true;
+          }
+        }
+      }
+
+      return false;
+    }
+
     return {
       CallExpression(node: TSESTree.CallExpression): void {
-        if (node.callee.type !== AST_NODE_TYPES.MemberExpression) {
-          return;
-        }
-        if (node.callee.object.type !== AST_NODE_TYPES.Identifier) {
-          return;
-        }
-        if (node.callee.object.name !== 'Promise') {
-          return;
-        }
-        if (node.callee.property.type !== AST_NODE_TYPES.Identifier) {
+        const callee = skipChainExpression(node.callee);
+        if (callee.type !== AST_NODE_TYPES.MemberExpression) {
           return;
         }
 
-        const { name } = node.callee.property;
-        if (!['race', 'all', 'allSettled'].includes(name)) {
+        if (callee.computed) {
+          if (
+            callee.property.type !== AST_NODE_TYPES.Literal ||
+            typeof callee.property.value !== 'string' ||
+            !aggregateFunctionNames.includes(callee.property.value)
+          ) {
+            return;
+          }
+        } else if (!aggregateFunctionNames.includes(callee.property.name)) {
           return;
         }
 
-        const { arguments: args } = node;
-        if (args.length !== 1) {
+        const callerType = services.getTypeAtLocation(callee.object);
+        if (!isPromiseLike(services.program, callerType)) {
+          return;
+        }
+
+        const args = node.arguments;
+        if (args.length < 1) {
           return;
         }
 
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -14,7 +14,7 @@ export default createRule({
     docs: {
       description:
         'Disallow passing non-Thenable values to promise aggregators',
-      recommended: 'recommended',
+      recommended: 'strict',
       requiresTypeChecking: true,
     },
     messages: {
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -168,7 +168,11 @@ export default createRule({
 
           for (const element of elements) {
             if (element == null) {
-              continue;
+              context.report({
+                messageId: 'inArray',
+                node: arg,
+              });
+              return;
             }
             const elementType = services.getTypeAtLocation(element);
             if (isTypeAnyType(elementType) || isTypeUnknownType(elementType)) {
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -1,0 +1,124 @@
+import * as tsutils from 'ts-api-utils';
+import {
+  isTypeAnyType,
+  isTypeUnknownType,
+} from '@typescript-eslint/type-utils';
+import { createRule, getParserServices } from '../util';
+
+export default createRule({
+  name: 'thenable-in-promise-aggregators',
+  meta: {
+    docs: {
+      description:
+        'Disallow passing non-Thenable values to promise aggregators',
+      recommended: 'recommended',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      inArray:
+        'Unexpected non-Thenable value in array passed to promise aggregator.',
+      arrayArg:
+        'Unexpected array of non-Thenable values passed to promise aggregator.',
+      nonArrayArg: 'Unexpected non-array passed to promise aggregator.',
+    },
+    schema: [],
+    type: 'problem',
+  },
+  defaultOptions: [],
+
+  create(context) {
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    return {
+      CallExpression(node): void {
+        if (node.callee.type !== 'MemberExpression') {
+          return;
+        }
+        if (node.callee.object.type !== 'Identifier') {
+          return;
+        }
+        if (node.callee.object.name !== 'Promise') {
+          return;
+        }
+        if (node.callee.property.type !== 'Identifier') {
+          return;
+        }
+
+        const { name } = node.callee.property;
+        if (!['race', 'all', 'allSettled'].includes(name)) {
+          return;
+        }
+
+        const { arguments: args } = node;
+        if (args.length !== 1) {
+          return;
+        }
+
+        const arg = args[0];
+        if (arg.type === 'ArrayExpression') {
+          const { elements } = arg;
+          if (elements.length === 0) {
+            return;
+          }
+
+          for (const element of elements) {
+            if (element === null) {
+              continue;
+            }
+            const elementType = services.getTypeAtLocation(element);
+            if (isTypeAnyType(elementType) || isTypeUnknownType(elementType)) {
+              continue;
+            }
+
+            const originalNode = services.esTreeNodeToTSNodeMap.get(element);
+            if (tsutils.isThenableType(checker, originalNode, elementType)) {
+              continue;
+            }
+
+            context.report({
+              messageId: 'inArray',
+              node: element,
+            });
+          }
+        } else {
+          const argType = services.getTypeAtLocation(arg);
+          if (isTypeAnyType(argType) || isTypeUnknownType(argType)) {
+            return;
+          }
+
+          if (!checker.isArrayType(argType)) {
+            context.report({
+              messageId: 'nonArrayArg',
+              node: arg,
+            });
+            return;
+          }
+
+          if (argType.typeArguments === undefined) {
+            return;
+          }
+
+          if (argType.typeArguments.length < 1) {
+            return;
+          }
+
+          const typeArg = argType.typeArguments[0];
+          if (isTypeAnyType(typeArg) || isTypeUnknownType(typeArg)) {
+            return;
+          }
+
+          const originalNode = services.esTreeNodeToTSNodeMap.get(arg);
+          if (tsutils.isThenableType(checker, originalNode, typeArg)) {
+            return;
+          }
+
+          context.report({
+            messageId: 'arrayArg',
+            node: arg,
+          });
+        }
+      },
+    };
+  },
+});
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -23,7 +23,6 @@ export default createRule({
         'Unexpected non-Thenable value in array passed to promise aggregator.',
       arrayArg:
         'Unexpected array of non-Thenable values passed to promise aggregator.',
-      nonArrayArg: 'Unexpected non-array passed to promise aggregator.',
     },
     schema: [],
     type: 'problem',
@@ -115,6 +114,52 @@ export default createRule({
       return false;
     }
 
+    function isPartiallyLikeType(
+      type: ts.Type,
+      predicate: (type: ts.Type) => boolean,
+    ): boolean {
+      if (isTypeAnyType(type) || isTypeUnknownType(type)) {
+        return true;
+      }
+      if (type.isIntersection() || type.isUnion()) {
+        return type.types.some(t => isPartiallyLikeType(t, predicate));
+      }
+      return predicate(type);
+    }
+
+    function isIndexableWithSomeElementsLike(
+      type: ts.Type,
+      predicate: (type: ts.Type) => boolean,
+    ): boolean {
+      if (isTypeAnyType(type) || isTypeUnknownType(type)) {
+        return true;
+      }
+
+      if (type.isIntersection() || type.isUnion()) {
+        return type.types.some(t =>
+          isIndexableWithSomeElementsLike(t, predicate),
+        );
+      }
+
+      if (!checker.isArrayType(type) && !checker.isTupleType(type)) {
+        const indexType = checker.getIndexTypeOfType(type, ts.IndexKind.Number);
+        if (indexType === undefined) {
+          return false;
+        }
+
+        return isPartiallyLikeType(indexType, predicate);
+      }
+
+      const typeArgs = type.typeArguments;
+      if (typeArgs === undefined) {
+        throw new Error(
+          'Expected to find type arguments for an array or tuple.',
+        );
+      }
+
+      return typeArgs.some(t => isPartiallyLikeType(t, predicate));
+    }
+
     return {
       CallExpression(node: TSESTree.CallExpression): void {
         const callee = skipChainExpression(node.callee);
@@ -181,64 +226,19 @@ export default createRule({
         }
 
         const argType = services.getTypeAtLocation(arg);
-        if (isTypeAnyType(argType) || isTypeUnknownType(argType)) {
-          return;
-        }
-
-        if (checker.isArrayType(argType)) {
-          if (
-            argType.typeArguments == null ||
-            argType.typeArguments.length < 1
-          ) {
-            throw new Error('Expected to find type arguments for an array.');
-          }
-
-          const typeArg = argType.typeArguments[0];
-          if (isTypeAnyType(typeArg) || isTypeUnknownType(typeArg)) {
-            return;
-          }
-
-          const originalNode = services.esTreeNodeToTSNodeMap.get(arg);
-          if (tsutils.isThenableType(checker, originalNode, typeArg)) {
-            return;
-          }
-
-          context.report({
-            messageId: 'arrayArg',
-            node: arg,
-          });
-          return;
-        }
-
-        if (checker.isTupleType(argType)) {
-          if (argType.typeArguments === undefined) {
-            return;
-          }
-
-          const originalNode = services.esTreeNodeToTSNodeMap.get(arg);
-          for (const typeArg of argType.typeArguments) {
-            if (isTypeAnyType(typeArg) || isTypeUnknownType(typeArg)) {
-              continue;
-            }
-
-            if (tsutils.isThenableType(checker, originalNode, typeArg)) {
-              continue;
-            }
-
-            context.report({
-              messageId: 'arrayArg',
-              node: arg,
-            });
-            return;
-          }
+        const originalNode = services.esTreeNodeToTSNodeMap.get(arg);
+        if (
+          isIndexableWithSomeElementsLike(argType, elementType => {
+            return tsutils.isThenableType(checker, originalNode, elementType);
+          })
+        ) {
           return;
         }
 
         context.report({
-          messageId: 'nonArrayArg',
+          messageId: 'arrayArg',
           node: arg,
         });
-        return;
       },
     };
   },
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -63,47 +63,31 @@ export default createRule({
       return false;
     }
 
-    // Is like Promise, or a class/interface extending from Promise
-    function isPromiseClassLike(program: ts.Program, type: ts.Type): boolean {
-      if (type.isIntersection()) {
-        return type.types.some(t => isPromiseLike(program, t));
-      }
-      if (type.isUnion()) {
-        return type.types.every(t => isPromiseLike(program, t));
-      }
-
-      const symbol = type.getSymbol();
-      if (!symbol) {
-        return false;
-      }
-
-      if (
-        symbol.getName() === 'Promise' &&
-        isSymbolFromDefaultLibrary(program, symbol)
-      ) {
-        return true;
-      }
-
-      if (symbol.flags & (ts.SymbolFlags.Class | ts.SymbolFlags.Interface)) {
-        const checker = program.getTypeChecker();
-
-        for (const baseType of checker.getBaseTypes(type as ts.InterfaceType)) {
-          if (isPromiseLike(program, baseType)) {
-            return true;
-          }
-        }
-      }
-
-      return false;
+    function isPromiseLike(program: ts.Program, type: ts.Type): boolean {
+      return isBuiltinSymbolLike(program, type, 'Promise');
     }
 
-    // Is like PromiseConstructor, or a class/interface extending from Promise
-    function isPromiseLike(program: ts.Program, type: ts.Type): boolean {
+    function isPromiseConstructorLike(
+      program: ts.Program,
+      type: ts.Type,
+    ): boolean {
+      return isBuiltinSymbolLike(program, type, 'PromiseConstructor');
+    }
+
+    function isBuiltinSymbolLike(
+      program: ts.Program,
+      type: ts.Type,
+      symbolName: string,
+    ): boolean {
       if (type.isIntersection()) {
-        return type.types.some(t => isPromiseLike(program, t));
+        return type.types.some(t =>
+          isBuiltinSymbolLike(program, t, symbolName),
+        );
       }
       if (type.isUnion()) {
-        return type.types.every(t => isPromiseLike(program, t));
+        return type.types.every(t =>
+          isBuiltinSymbolLike(program, t, symbolName),
+        );
       }
 
       const symbol = type.getSymbol();
@@ -112,7 +96,7 @@ export default createRule({
       }
 
       if (
-        symbol.getName() === 'PromiseConstructor' &&
+        symbol.getName() === symbolName &&
         isSymbolFromDefaultLibrary(program, symbol)
       ) {
         return true;
@@ -122,7 +106,7 @@ export default createRule({
         const checker = program.getTypeChecker();
 
         for (const baseType of checker.getBaseTypes(type as ts.InterfaceType)) {
-          if (isPromiseClassLike(program, baseType)) {
+          if (isBuiltinSymbolLike(program, baseType, symbolName)) {
             return true;
           }
         }
@@ -151,7 +135,10 @@ export default createRule({
         }
 
         const callerType = services.getTypeAtLocation(callee.object);
-        if (!isPromiseLike(services.program, callerType)) {
+        if (
+          !isPromiseConstructorLike(services.program, callerType) &&
+          !isPromiseLike(services.program, callerType)
+        ) {
           return;
         }
 
--- a/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
+++ b/packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
@@ -1,6 +1,5 @@
 import {
-  isPromiseConstructorLike,
-  isPromiseLike,
+  isBuiltinSymbolLike,
   isTypeAnyType,
   isTypeUnknownType,
 } from '@typescript-eslint/type-utils';
@@ -10,6 +9,8 @@ import * as tsutils from 'ts-api-utils';
 import * as ts from 'typescript';
 
 import { createRule, getParserServices } from '../util';
+
+const aggregateFunctionNames = new Set(['all', 'race', 'allSettled', 'any']);
 
 export default createRule({
   name: 'thenable-in-promise-aggregators',
@@ -25,6 +26,8 @@ export default createRule({
         'Unexpected non-Thenable value in array passed to promise aggregator.',
       arrayArg:
         'Unexpected array of non-Thenable values passed to promise aggregator.',
+      emptyArrayElement:
+        'Unexpected empty element in array passed to promise aggregator (do you have an extra comma?).',
     },
     schema: [],
     type: 'problem',
@@ -34,8 +37,6 @@ export default createRule({
   create(context) {
     const services = getParserServices(context);
     const checker = services.program.getTypeChecker();
-
-    const aggregateFunctionNames = ['all', 'race', 'allSettled', 'any'];
 
     function skipChainExpression<T extends TSESTree.Node>(
       node: T,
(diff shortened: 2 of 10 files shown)


packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
```diff
@@ -0,0 +1,358 @@
...
+async function test() {
+  class Foo extends Promise<number> {}
+  const foo: Foo = Foo.resolve(2);
+  await Promise.race([foo]);
+
+  class Bar extends Foo {}
+  const bar: Bar = Bar.resolve(2);
+  await Promise.race([bar]);
+}
+    `,
```
auvred, line 84: Ref: https://typescript-eslint.io/contributing/local-development#granular-unit-tests

We prefer test cases to be small and self-contained, rather than having multiple assertions in one case.
Could you please split this case?

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
```diff
@@ -0,0 +1,358 @@
...
+    `
+async function test() {
+  await Promise.race([Math.random() > 0.5 ? numberPromise : 0]);
+  await Promise.race([Math.random() > 0.5 ? foo : 0]);
+  await Promise.race([Math.random() > 0.5 ? bar : 0]);
+
+  const intersectionPromise: Promise<number> & number;
+  await Promise.race([intersectionPromise]);
+}
+    `,
```
auvred, line 94: Same as above ^

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  create(context) {
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    return {
+      CallExpression(node: TSESTree.CallExpression): void {
+        if (node.callee.type !== AST_NODE_TYPES.MemberExpression) {
+          return;
+        }
```
auvred, line 40: We should skip `ChainExpression`s here because
```ts
(Promise?.all)([1])
```

Won't be checked, but

```ts
Promise?.all([1])
```
will be

auvred: Btw, this logic is very similar to the logic from https://github.com/typescript-eslint/typescript-eslint/pull/8011

https://github.com/typescript-eslint/typescript-eslint/pull/8011/files#diff-801508874bd1d1cb9f122a9ca577fdf2350bfd756c5e63572487b65b859d6799R91-R102

Maybe we can reuse it between these two rules, what do you think?

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
```diff
@@ -0,0 +1,358 @@
...
+  const promiseArr: Promise<number>[];
+  await Promise.all(promiseArr);
+}
+    `,
+    `
+async function test() {
+  const intersectionArr: (Promise<number> & number)[];
+  await Promise.all(intersectionArr);
+}
+    `,
```
auvred, line 138: Maybe we should also add some cases with unions? Because it's not clear to me whether this rule should report on

```ts
declare const something: Promise<number> | number
Promise.all([something])
```

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,127 @@
...
+        if (node.callee.object.type !== AST_NODE_TYPES.Identifier) {
+          return;
+        }
+        if (node.callee.object.name !== 'Promise') {
+          return;
+        }
+        if (node.callee.property.type !== AST_NODE_TYPES.Identifier) {
+          return;
+        }
```
auvred, line 49: I've removed this condition and all tests passed. We should either remove this condition or add a test case proving its necessity.

I'd prefer the second option

---

Also we can check for cases such a `Promise['all'](...)`

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,127 @@
...
+        if (args.length !== 1) {
+          return;
+        }
+
+        const arg = args[0];
+        if (arg.type === AST_NODE_TYPES.ArrayExpression) {
+          const { elements } = arg;
+          if (elements.length === 0) {
+            return;
+          }
```
auvred, line 66: Can we add `Promise.all([])` test case to prove this condition?

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,127 @@
...
+
+          for (const element of elements) {
+            if (element == null) {
+              continue;
+            }
```
auvred, line 71: We don't have test case containing `Promise.all([,])`, let's add it?

Also not sure if this code should be reported or not? Technically `null` is not a thenable

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,127 @@
...
+            return;
+          }
+
+          if (argType.typeArguments === undefined) {
+            return;
+          }
+
+          if (argType.typeArguments.length < 1) {
+            return;
+          }
```
auvred, line 107: I've removed these two branches and all tests passed, so we should either assert that `typeArguments` is always defined array with known length, or add some test cases

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,127 @@
...
+
+          const typeArg = argType.typeArguments[0];
+          if (isTypeAnyType(typeArg) || isTypeUnknownType(typeArg)) {
+            return;
+          }
```
auvred, line 112: It'd be cool if we had the following test cases to fully test this condition
```ts
declare const arrOfAny: any[]
Promise.all(arrOfAny)
```
```ts
declare const arrOfUnknown: unknown[]
Promise.all(arrOfUnknown)
```

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
```diff
@@ -0,0 +1,358 @@
...
+    {
+      code: `
+class NonPromise extends Array {}
+await Promise.race([new NonPromise()]);
+      `,
+      errors: [
+        {
+          line: 3,
+          messageId: messageIdInArray,
+          suggestions: [],
```
auvred, line 217:
```suggestion
```

Do we really need `suggestions` here?

auvred (changes requested):
Nice start! Thanks for working on it!

Just requesting some changes:

packages/eslint-plugin/docs/rules/thenable-in-promise-aggregators.md
```diff
@@ -0,0 +1,49 @@
...
+>
+> See **https://typescript-eslint.io/rules/thenable-in-promise-aggregators** for documentation.
+
+A "Thenable" value is an object which has a `then` method, such as a Promise.
+The `await` keyword is generally used to retrieve the result of calling a Thenable's `then` method.
+
+When multiple Thenable's are running at the same time, it is sometimes desirable to wait until any one of them resolves (`Promise.race`), all of them resolve or any of them reject (`Promise.all`), or all of them resolve or reject (`Promise.allSettled`).
+
+Each of these functions accept an iterable of promises as input and return a single
+Promise.
```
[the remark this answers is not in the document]
Tjstretchalot: Oops, fixing!

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,127 @@
...
+
+import { createRule, getParserServices } from '../util';
+
+export default createRule({
+  name: 'thenable-in-promise-aggregators',
+  meta: {
+    docs: {
+      description:
+        'Disallow passing non-Thenable values to promise aggregators',
+      recommended: 'recommended',
```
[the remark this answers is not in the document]
Tjstretchalot: Ah yes! I got a little confused here. I'm updating to `recommended: 'strict'` since from some perusing I think that puts it in `strict-type-checked` since it's also marked `requiresTypeChecking`. Let me know if that's incorrect :)

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,127 @@
...
+      CallExpression(node: TSESTree.CallExpression): void {
+        if (node.callee.type !== AST_NODE_TYPES.MemberExpression) {
+          return;
+        }
+        if (node.callee.object.type !== AST_NODE_TYPES.Identifier) {
+          return;
+        }
+        if (node.callee.object.name !== 'Promise') {
+          return;
+        }
```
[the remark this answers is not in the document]
Tjstretchalot: Gotcha. Adding tests and using what the correct current way of doing it so it'll be easy to find for #8011, and if that gets merged first will refactor

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,127 @@
...
+              node: element,
+            });
+          }
+        } else {
+          const argType = services.getTypeAtLocation(arg);
+          if (isTypeAnyType(argType) || isTypeUnknownType(argType)) {
+            return;
+          }
+
+          if (!checker.isArrayType(argType)) {
```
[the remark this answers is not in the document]
Tjstretchalot: Good catch! Fixing

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: Ahh I wasn't really clear what this one was testing for I just saw it in several other tests. Thank you! Reporting & adding as a test

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: I added the test case I thought this might affect and changed to an assert

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: Added the tests. I'm noticing that you used the syntax I found for the invalid ones for a valid case; can/should I switch the valid tests from the current

```ts
async function test() {
  // ...
}
```

to this style, like in the invalid cases?

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
Tjstretchalot: Oops! Removed

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
```diff
@@ -0,0 +1,358 @@
+import { RuleTester } from '@typescript-eslint/rule-tester';
+
+import rule from '../../src/rules/thenable-in-promise-aggregators';
+import { getFixturesRootDir } from '../RuleTester';
+
+const rootDir = getFixturesRootDir();
+const messageIdInArray = 'inArray';
+const messageIdArrayArg = 'arrayArg';
+const messageIdNonArrayArg = 'nonArrayArg';
```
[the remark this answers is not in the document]
Tjstretchalot: Updating!

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
Tjstretchalot: Ahh! Copied these from `await-thenable` almost precisely, but happy to split

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: (Because of hidden comments I saw this in the wrong order)

It is very similar; I think they can be reused! I wrote it in a way so that it should be easy to adapt once one gets merged. Let me know if you'd rather break it out before one gets merged (and I guess rebase on the other one?)

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
Tjstretchalot: Fixing

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: Got refactored around a bit handling chain expressions, but added the tests!

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
Tjstretchalot: Adding explicit tests for unions. Personally I think it makes sense to let union types through, though I could see a case for a flag for it. I think unioning a promise with a non-promise is somewhat rare, but when it does happen it's intentional. For example, I do this regularly on buttons wrappers in react:

```tsx
type MyButtonProps = {
  /**
   * Handler for when the button is clicked. If the result is a promise, the button will be disabled
   * while the promise is running. If the promise takes a long time, the screen will be darkened slightly
   * and an animation will play
   */
  onClick: () => Promise<void> | undefined;
}
```

Then it's a question of if the caller should check for undefined first, or just run it through the promise aggregator and use the fact it'll resolve instantly. Here I'd be fine being explicit, but I wonder if there are some more complicated cases where that'd be tedious

Tjstretchalot:
I believe I covered all the comments. I left conversations open when there may be ambiguity, but since it's easy to miss (I missed 5 due to the hidden comments), I'll summarize still pending thoughts here that I'm seeking advice on the best way to resolve:

- This uses the exact same logic as #8011 for skipping chain express; should this be broken out? Should it be done in this PR or a separate one?
- This uses the exact same logic as #8011 for isSymbolFromDefaultLibrary, which is already broken out in #8011 - I copied it over here so that this PR doesn't conflict with that one and can be easily updated to use the imported function
- This and #8011 both have need to check for promises but they aren't implemented the same (mine is checking for subclasses of Promise as well as PromiseConstructor. I think this is a difference in what they are accomplishing since #8011 is specifically handling creating promises, however, worth getting another set of eyes on. If I copy over `isPromiseConstructorLike` and use that instead of `isPromiseLike`, the following test fails

```tsx
    {
      code: `
class Foo extends Promise<number> {}
await Foo.all([0]);
      `,
      errors: [
        {
          line: 3,
          messageId: 'inArray',
        },
      ],
    }
```

- I use two different styles for valid/invalid cases because I used `await-thenable` as a reference and it does that. Can/should I do something differently (now or if I make future PRs)?

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
auvred: It's not mandatory, but, personally, I prefer to avoid extra code in test cases. 
For example: there are two identical test cases, but the second one takes 5x more lines than the first one:

```ts
ruleTester.run('thenable-in-promise-aggregators', rule, {
  valid: [
    `await Promise.race([Promise.resolve(3)]);`,
    `
async function test() {
  await Promise.race([Promise.resolve(3)]);
}
    `,
  ]
})
```

(for multiline arranges + assertions it only adds two extra lines, but whatever)

We can wait for the maintainer to make a final decision, but in general I think it's fine to have cases with
```ts
async function test() {
  // ...
}
```

Tjstretchalot: Alright great! Wasn't sure if there was a particular reason for the function in valid cases; I'm updating to the shortened version

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
auvred: I moved both `isPromiseLike` and `isPromiseConstructorLike` to the `type-utils` in https://github.com/typescript-eslint/typescript-eslint/pull/8011/commits/a0699cfa5c5143a38c38f8d746a28688d736fd74#diff-c7675022367e4f5ec62101a590d4f77765532a6ec1a4e6e628b407a082bcf2fd, so we can easily reuse them later. Not sure about `skipChainExpression` though

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,258 @@
...
+            !aggregateFunctionNames.includes(callee.property.value)
+          ) {
+            return;
+          }
+        } else if (!aggregateFunctionNames.includes(callee.property.name)) {
+          return;
+        }
+
+        const callerType = services.getTypeAtLocation(callee.object);
+        if (!isPromiseLike(services.program, callerType)) {
```
auvred, line 154:
```suggestion
        const calleeType = services.getTypeAtLocation(callee.object);
        if (!isPromiseLike(services.program, calleeType)) {
```

[nit] let's name it `calleeType` instead of `callerType` since it is named `callee` elsewhere

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,258 @@
...
+            }
+            const elementType = services.getTypeAtLocation(element);
+            if (isTypeAnyType(elementType) || isTypeUnknownType(elementType)) {
+              continue;
+            }
+
+            const originalNode = services.esTreeNodeToTSNodeMap.get(element);
+            if (tsutils.isThenableType(checker, originalNode, elementType)) {
+              continue;
+            }
```
auvred, line 186: this logic (if `isTypeAnyType(sometype) || isTypeUnknownType(sometype)` or `tsutils.isThenableType(checker, originalNode, sometype)`) is used three times in this rule. Maybe we should move it to the function to avoid duplicated code?

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,258 @@
...
+          }
+          return;
+        }
+
+        const argType = services.getTypeAtLocation(arg);
+        if (isTypeAnyType(argType) || isTypeUnknownType(argType)) {
+          return;
+        }
+
+        if (checker.isArrayType(argType)) {
```
auvred, line 201: I noticed that we don't check for unions/intersections here, so the following code produces lint errors

```ts
declare const a: Promise<string>[] | Promise<number>[]
Promise.all(a)

declare const b: [Promise<number>] | [Promise<string>]
Promise.all(b)

declare const c: Promise<string>[] & Promise<number>[]
Promise.all(c)

declare const d: [Promise<number>] & [Promise<string>]
Promise.all(d)
```

[playground](https://deploy-preview-8094--typescript-eslint.netlify.app/play/#ts=5.3.2&showAST=es&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3lAXPACjEgLYCWaIAPJjKSgOYB8A2gLrwA%2BBRZFlKAV2IAjEDBasAUIRLkQAOigQIACigBKSZNCRYCZOizC8zGbyqCRYxuy6mec6hloMb0hxUXKVwzdvDQcIiomIh4Zo40dExs8ABk3LJ8lqLibO5JCkqqYH46gfohWMAmEclCqTbx8PaZTi5MUmVZ3sDqQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6fAC0SYEMAjJLWa1i0APYBbSskS0eAc3nRE8nvjHRUGSImjjokcGAC%2BIY0A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,258 @@
...
+          }
+
+          context.report({
+            messageId: 'arrayArg',
+            node: arg,
+          });
+          return;
+        }
+
+        if (checker.isTupleType(argType)) {
```
auvred, line 226: Same as above ^

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        const argType = services.getTypeAtLocation(arg);
+        if (isTypeAnyType(argType) || isTypeUnknownType(argType)) {
+          return;
+        }
+
+        if (checker.isArrayType(argType)) {
+          if (
+            argType.typeArguments == null ||
```
auvred, line 203: On line `227` it's compared against `undefined`
```ts
if (argType.typeArguments === undefined) {
  return;
}
```

Let's compare it against one them to be consistent (I'd prefer to compare it against `undefined` as IMO it's more explicit than comparing against `null`). What do you think?

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
```diff
@@ -0,0 +1,561 @@
...
+    },
+    {
+      code: 'await Promise.race(3);',
+      errors: [
+        {
+          line: 1,
+          messageId: 'nonArrayArg',
+        },
+      ],
+    },
```
auvred, line 427: Oh, I can't understand how I didn't notice it before! `Promise.race(3)` and others with error `nonArrayArg` are not valid TS code. `all`, `race` and `allSettled` have array/iterable in their signature, so the error will be thrown by `tsc` itself, I guess we shouldn't report on this

auvred (changes requested):

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,258 @@
...
+    schema: [],
+    type: 'problem',
+  },
+  defaultOptions: [],
+
+  create(context) {
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    const aggregateFunctionNames = ['all', 'race', 'allSettled'];
```
auvred, line 37: Btw, what do you think about `Promise.any`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any

I think we can also include it here

auvred (changes requested):

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: Hmm; I refactored around a bit, though I still have 3 calls like this. However, I don't think something like this would add anything besides some mental overhead:

```ts
function isTypeAnyOrUnknownType(type: ts.Type): boolean {
  return isTypeAnyType(type) || isTypeUnknownType(type);
}
```

For this to be useful from my current perspective there would need to be a more salient name for the method that is a more useful concept than `checking for any or unknown` and makes the implementation obvious (since it's so short). I looked elsewhere around the codebase and any types are not universally treated the same as unknown types, e.g., I could imagine a flag on this check later, perhaps two flags (one for any and one for unknown), and depending on exactly how those flags work it might make sense to break it out then, but more likely even if you made those flags it would still not be helpful to have a function like this given you would want different errors each which explanations of the flag to use to disable that feature, ect.

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: Added lots of tests around these, including absolutely psychotic stuff like the following:

```ts
type Foo = { new (): [number | Promise<number>] };
declare const foo: Foo;
class Baz extends foo {}
declare const baz: Baz;
await Promise.all(baz);
```

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
Tjstretchalot: That's a good point! It was useful in testing locally when I was doing an eslint plugin so I could understand why I wasn't triggering the warning, but theres no reason for it to be here

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: I didn't know about that one! Added with a couple tests

Josh-Cena (commented):
Was the decision to create a new rule discussed anywhere? I thought it would be a new option for `await-thenable`, like `promiseAggregators: "all" | "any" | "ignore"`.

Tjstretchalot:
Happy new year!

Rebased onto the current state of the main branch

@Josh-Cena There was no discussion AFAIK, I initially made this as a custom eslint rule for myself and so it was easier to translate it over as its own ESLint rule. I'm impartial to including it directly in await-thenable. The only main question I have is how to handle the defaults for the flags?

Right now it's in what I think is the ideal case: included when enabling `strict-typed`, not included in `recommended`, and it can be included in `recommended` on the next major version.

If we add it as configuration for `await-thenable`, since that is in `recommended`, the flags presumably need to default to off. But that would mean it's not included in `strict-typed` and doesn't have an obvious upgrade path to `recommended`. I suppose it might be possible to have the flags default change depending on whether you have `strict-typed` included or not, but that seems... weird?

I imagine this came up any other time typescript-eslint wanted to add flags, so I just need to be pointed in the right direction for how to handle them. It would be a little annoying for my use-case, though certainly not the end of the world, if I had to remember to specifically enable these flags on every project to make use of the rule, especially since I would likely only notice it's disabled after the issue it attempts to resolve bit me!

Josh-Cena:
Yeah, good point about different options. I think this fits in https://github.com/typescript-eslint/typescript-eslint/issues/7318.

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Yup, that's right! :)

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
JoshuaKGoldberg, line 1: Continuing the discussion of whether this should be a separate rule (https://github.com/typescript-eslint/typescript-eslint/pull/8094#issuecomment-1874127765): I think it makes sense as its own rule. `await-thenable` right now specifically targets `await`ing something. So adding Promise aggregators in there would be a nontrivial expansion of the rule's scope. And the rule's name would no longer really be precise... which IMO is a good indication of something that maybe should be a separate rule.

Plus, it's nice to finally have another rule that refers to "thenables" 😄 https://github.com/typescript-eslint/typescript-eslint/discussions/6012.

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,176 @@
...
+import * as ts from 'typescript';
+
+import { createRule, getParserServices } from '../util';
+
+export default createRule({
+  name: 'thenable-in-promise-aggregators',
+  meta: {
+    docs: {
+      description:
+        'Disallow passing non-Thenable values to promise aggregators',
```
JoshuaKGoldberg, line 19: [Docs] Interesting, we don't have a standard for whether we refer to Promises with an upper-case P or lower-case p... Outside the scope of this PR, maybe it'd be a good followup to file an issue about standardizing this in docs?

Not a blocker or request for changes 😄 just ruminating.

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,176 @@
...
+
+        if (callee.computed) {
+          if (
+            callee.property.type !== AST_NODE_TYPES.Literal ||
+            typeof callee.property.value !== 'string' ||
+            !aggregateFunctionNames.includes(callee.property.value)
+          ) {
+            return;
+          }
+        } else if (!aggregateFunctionNames.includes(callee.property.name)) {
```
JoshuaKGoldberg, line 109: [Refactor] This _"get the name one way if it's computed or another if it's static"_ need comes up a lot in rules. We should have a shared utility for it. `getStaticStringValue` should do the trick, or `getStaticValue`. One of those 😄.

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,176 @@
...
+    schema: [],
+    type: 'problem',
+  },
+  defaultOptions: [],
+
+  create(context) {
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    const aggregateFunctionNames = ['all', 'race', 'allSettled', 'any'];
```
JoshuaKGoldberg, line 38: [Style] Just nitpicking, but - what do you think about making this a `Set` and/or moving it outside of the rule's `create`? There's no need to remake it every time. And semantically, a `Set` is really more of what this is being used for (https://github.com/sindresorhus/eslint-plugin-unicorn/blob/bb9e7884c105f2c0f942153f46dcd3c4fe422482/docs/rules/prefer-set-has.md).

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,176 @@
...
+          !isPromiseConstructorLike(services.program, calleeType) &&
+          !isPromiseLike(services.program, calleeType)
+        ) {
+          return;
+        }
+
+        const args = node.arguments;
+        if (args.length < 1) {
+          return;
+        }
```
JoshuaKGoldberg, line 124: [Performance] Generally, asking for type information is the most performance-expensive part of rules. So if there's any not-type-informed _"bail early if ..."_ check that can happen first, that's better. Let's move this check to before the `services.getTypeAtLocation` call.

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,176 @@
...
+            return;
+          }
+        } else if (!aggregateFunctionNames.includes(callee.property.name)) {
+          return;
+        }
+
+        const calleeType = services.getTypeAtLocation(callee.object);
+        if (
+          !isPromiseConstructorLike(services.program, calleeType) &&
+          !isPromiseLike(services.program, calleeType)
```
JoshuaKGoldberg, line 116: [Performance] On the inside, both of these `is*Like` functions directly call to `isBuiltinSymbolLike`, which directly calls to `isBuiltinSymbolLikeRecurser`, which is a recursive function that roughly looks through all the node's types. So in the case of neither of these functions finding what they're looking for, they'll go through those steps twice.

This is also roughly the same code as `typeAtLocationIsLikePromise` from `prefer-promise-reject-errors`.

I think a good cleanup/followup/refactor would be to augment `isBuiltinSymbolLike` to take in an array or set of names that it checks for. That way we can do all that recursive stuff _once_ rather than _twice_ (one for each `symbolName`).

This is existing code (that I missed in a previous review!) so not a blocker at all IMO. Fine to file as an issue!

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,176 @@
...
+        if (arg.type === AST_NODE_TYPES.ArrayExpression) {
+          const { elements } = arg;
+          if (elements.length === 0) {
+            return;
+          }
+
+          for (const element of elements) {
+            if (element == null) {
+              context.report({
+                messageId: 'inArray',
```
JoshuaKGoldberg, line 136: [Design] 🤔 I don't love using this same general `'inArray'` message for the case of an empty array element/slot. Most of the time when there's an extra comma it's because the user has accidentally typed one - not intentionally.

I think this deserves its own special-case error message. WDYT?

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
JoshuaKGoldberg: Yeah I think unions should be allowed. It's a nice convenience and intentionally part of the Promise aggregator design to allow this.

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
```diff
@@ -0,0 +1,711 @@
...
+  parserOptions: {
+    ecmaVersion: 2018,
+    tsconfigRootDir: rootDir,
+    project: './tsconfig.json',
+  },
+  parser: '@typescript-eslint/parser',
+});
+
+ruleTester.run('thenable-in-promise-aggregators', rule, {
+  valid: [
```
JoshuaKGoldberg, line 18: [Testing] Some additional valid cases to put in...

* <code>Promise[\`all\`]</code>
* `Promise["all"]`
* `const key = "all"; Promise[key]`

JoshuaKGoldberg (changes requested):
This is looking great, thanks for working so much on this rule! I think @auvred reviewing earlier has made my job a lot easier 😄 thanks for that too!

I _think_ I responded to all the pending conversations but I'm not sure. There was a lot to go through 😅. If there's anything I missed, please do start a new comment thread in a file and I'll take a look. 

Requesting changes on some more testing and on reusing a shared utility for a static key value. The other points aren't blocking IMO as we can file them as followup issues.

Blue cartoonish character with a gray triangle hat dancing awkwardly on top of a yellow blanket saying "you're doing great!" on top of what looks like a potato with a smiley face.

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
JoshuaKGoldberg, line 1: [Praise] Super thorough tests by the way, nicely done!

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
JoshuaKGoldberg, line 1: [Praise] This was very clear to read through, nice! Thanks for making my life easier 😄

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: There is already a function `getStaticStringValue` in `eslint-plugin/src/util` but it's about finding what the source representation of a value would be, e.g., turning a regex into `/this/` or null into the string `null`, etc.

To avoid confusion with that I've broken out a `getMemberName` function you can look at and see if it might be worth breaking out into a shared utility

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: Makes sense to me! For breaking it out it's also more consistent with `space-infix-ops` after some digging

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: Sure, more clarity never hurts!

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,176 @@
...
+  name: 'thenable-in-promise-aggregators',
+  meta: {
+    docs: {
+      description:
+        'Disallow passing non-Thenable values to promise aggregators',
+      recommended: 'strict',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      inArray:
```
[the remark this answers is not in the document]
Tjstretchalot: This is the main case I'm specifically looking to lint, and is the most common case where mistakes are made. There is no advantage to doing this from the perspective of control flow (i.e., how long the returned promise takes to fulfill), however, it does change the return value. However, whenever you really do want to provide a non-Thenable to change the control flow, you can always wrap it in `Promise.resolve` to get exactly the same behavior while making it way more clear how the aggregator will behave. This is still true even when considering if the returned promise is fulfilled synchronously vs asynchronously: these promise aggregators only fulfill synchronously when passed an empty array.

I would argue especially for all of these promise aggregators, the most important thing is making the control flow clear. Any non-Thenable value passed to a promise aggregator will be treated as `Promise.resolve(value)` according to the spec; see e.g., [PerformPromiseAll](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-performpromiseall). Thus given that a value in the list is known to be non-Thenable, there is a static conversion available that will make the control flow easier to understand, or you can be explicit and wrap the value in Promise.resolve, which jumps out at the reader and thus also makes the control flow easier to understand.

Given:

```ts
declare const aPromise: Promise<number>;
declare const bPromise: Promise<number;
declare const cNumber: number;
```

The following are all equivalent:


```ts
// all

Promise.all([aPromise, bPromise, cNumber])
// equivalent to
Promise.all([aPromise, bPromise, Promise.resolve(bNumber)])
// which resolves in the same amount of time as
Promise.all([aPromise, bPromise])

// previously working, now causes lint error
const [a, b, c] = await Promise.all([aPromise, bNumber]);

// suggested change
const [a, b] = await Promise.all([aPromise, bPromise]);
const c = cNumber;
```

```ts
// allSettled

Promise.allSettled([aPromise, bPromise, cNumber]);
// equivalent to
Promise.allSettled([aPromise, bPromise, Promise.resolve(cNumber)]);
// resolves in the same time as
Promise.allSettled([aPromise, bPromise]);

// previously working, now causes lint error
const [aPromise2, bPromise2, cPromise] = await Promise.allSettled([aPromise, bPromise, cNumber]);

// suggested change
const [aPromise2, bPromise2] = await Promise.allSettled([aPromise, bPromise]);
const cPromise = Promise.resolve(cNumber);
```

```ts
// race

await Promise.race([aPromise, bPromise, cNumber]);
// equivalent to
await Promise.race([aPromise, bPromise, Promise.resolve(cNumber)]);
// always resolves immediately (but never synchronously, as Promise.race is never synchronous)

// previously working, now causes lint error
// this is a if a is immediately resolved, b if b is immediately resolved, otherwise c
const aOrBOrC = await Promise.race([aPromise, bPromise, cNumber]);

// suggested change if you really did want the above behavior (most of the time you do NOT)
const aOrBOrC = await Promise.race([aPromise, bPromise, Promise.resolve(cNumber)]);

// suggested change if you actually intended to wait for something to happen
const aOrB = await Promise.race([aPromise, bPromise])

// note that most of the time this comes up, this is whats actually happening:
declare const aPromise: Promise<number>;
declare const bPromise: Promise<number>;
declare const cWrappedPromise: { promise: Promise<number> };

// lint error that catches a real subtle bug
const aOrBOrC = await Promise.race([aPromise, bPromise, cWrappedPromise]);

// correct code
const aOrBOrC = await Promise.race([aPromise, bPromise, cWrappedPromise.promise]);
```

```ts
// any is the same as race for the above purposes, the difference is how rejected values are handled
```

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,176 @@
...
+      return node.type === AST_NODE_TYPES.ChainExpression
+        ? node.expression
+        : node;
+    }
+
+    function isPartiallyLikeType(
+      type: ts.Type,
+      predicate: (type: ts.Type) => boolean,
+    ): boolean {
+      if (isTypeAnyType(type) || isTypeUnknownType(type)) {
```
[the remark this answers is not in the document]
Tjstretchalot: The main goal for this lint rule for me is to catch cases where you can statically guarrantee there is a simpler way to call the aggregator that fulfills in the same amount of time, i.e., where it is very, very likely you accidentally provided the wrong value. For providing `unknown` to a promise aggregator, I might guard it with a type check, but I think it'd be perfectly reasonable not to:

```ts
// not looking to cause an error with something like this
async function doThingWithT(t: unknown) {
  let timedOut = false;
  const timeout = new Promise((resolve) => {
    timedOut = true;
    setTimeout(resolve, 100);
  });
  await Promise.race([t, timeout]);
  if (timedOut) {
    throw new Error('t is thenable and exceeded timeout');
  }
  const awaitedT = await t;
  if (typeof awaitedT === 'string') { 
    // etc etc
  }
}
```

```ts
// am looking to cause an error with something like this (type taken from production code)
type CancelablePromise<T> = { done: () => boolean, promise: Promise<T>, cancel: () => void };

const doThingWithGuard = async (a: Promise<number>, b: CancelablePromise<number>) => {
  try {
    await Promise.race([a, b]); // <--- error here, should be b.promise. What's written is useless
  } catch (e) {
    // this error handling isn't gonna work
  }

  if (b.done()) {
    // since the above is incorrect, this could still take time and could still error
    const aResult = await a;
    // ...
  }
  // ...
}
```

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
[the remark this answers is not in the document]
Tjstretchalot: Good catch! Added a test for `never` to fail as my argument for why unknown should be allowed doesn't apply to `never`

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
Tjstretchalot: I don't know how to test `Promise["all"]` without `yarn lint --fix` just swapping it to single quotes, assuming you mean for double quotes. I do have the single quotes case in there near the top of the valid list.

I added the template literal case for both valid/invalid. For invalid it actually failed that test case, I added the necessary code to handle it in the refactored `getMemberName` but I do wonder how bad the performance will be on that since I imagine member expression calls are quite common, though I don't know about indexed ones with non-literals. The only case where this type of thing might be convenient is if there were a parameter to switch between `any` vs `race` and they chose to do it like this:

```ts
async function waitWithTimeout(a: Promise<T>, aggregator: 'race' | 'any') {
  const timeout = new Promise((resolve) => setTimeout(resolve, 100));
  await Promise[aggregator]([a, timeout])
}
```

I added tests for this type of thing, but it meant that `getMemberName` had to become `isMemberName` instead to handle unions

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
Tjstretchalot: Easy enough! I kept with the predicate pattern it already had going, which also has the benefit of giving a little more freedom to the caller without being too tedious.

I'll note that this was reaaallly confusing until I realized that I had to do some combination of `yarn` / `yarn build` for changes to actually be applied!

Tjstretchalot:
(the force-pushes are just rebasing with main to keep it in sync)

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
JoshuaKGoldberg: Oh, I've been bit by that before. Do you think it'd be something useful to mention in docs? #8033 is an example of another recent augmentation in this line (not quite the same).

auvred: When I was working on the type-utils package, I put a little stub to the `packages/type-utils/dist/index.js`:

```js
module.exports = require('jiti')(null, { interopDefault: true })('/path/to/typescript-eslint/packages/type-utils/src/index.ts')
```

This way, I don't have to rebuild the whole package every time I change something inside it. `jiti` just reads actual code from disk each time and transpiles in on the fly!

Perhaps we can add a script (or even use postinstall) that places stubs instead of package build artifacts when developing locally. I think it should improve DX

Ref: https://antfu.me/posts/publish-esm-and-cjs#stubbing

Tjstretchalot: I'd definitely vote for some kind of automatic local development flag, even if it's as simple as changing the launch configurations within vscode so that the jest rebuilds / double checks. I would have checked the docs next, but I had to first have the insight that it was actually not running the new code rather than me just doing something really silly in order to know to check the docs, at which point I was 99% of the way there as I then started trying yarn / yarn build commands.

packages/eslint-plugin/docs/rules/thenable-in-promise-aggregators.md
```diff
@@ -0,0 +1,48 @@
...
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/thenable-in-promise-aggregators** for documentation.
+
+A "Thenable" value is an object which has a `then` method, such as a Promise.
+The `await` keyword is generally used to retrieve the result of calling a Thenable's `then` method.
+
+When multiple Thenable's are running at the same time, it is sometimes desirable to wait until any one of them resolves (`Promise.race`), all of them resolve or any of them reject (`Promise.all`), or all of them resolve or reject (`Promise.allSettled`).
```
JoshuaKGoldberg, line 12: [Docs] Nit: I know apostrophes to indicate plurals can sometimes be ok, but I've always found them to be ambiguous. Maybe this is just me 😄 

```suggestion
When multiple Thenables are running at the same time, it is sometimes desirable to wait until any one of them resolves (`Promise.race`), all of them resolve or any of them reject (`Promise.all`), or all of them resolve or reject (`Promise.allSettled`).
```

packages/eslint-plugin/docs/rules/thenable-in-promise-aggregators.md
```diff
@@ -0,0 +1,48 @@
...
+
+await Promise.race([
+  new Promise(resolve => setTimeout(resolve, 3000)),
+  new Promise(resolve => setTimeout(resolve, 6000)),
+]);
+```
+
+## When Not To Use It
+
+If you want to allow code to use `Promise.race`, `Promise.all`, or `Promise.allSettled` on arrays of non-promise values.
```
JoshuaKGoldberg, line 46:
```suggestion
If you want to allow code to use `Promise.race`, `Promise.all`, or `Promise.allSettled` on arrays of non-Thenable values.
```

packages/eslint-plugin/tests/rules/thenable-in-promise-aggregators.test.ts
JoshuaKGoldberg: > I don't know how to test `Promise["all"]` without `yarn lint --fix` just swapping it to single quotes, assuming you mean for double quotes

Oh yeah search `noFormat`, we've hit that annoyance before.

packages/type-utils/src/builtinSymbolLikes.ts
```diff
@@ -105,7 +117,7 @@ export function isBuiltinTypeAliasLike(
 export function isBuiltinSymbolLike(
   program: ts.Program,
   type: ts.Type,
-  symbolName: string,
+  predicate: (symbolName: string) => boolean,
```
JoshuaKGoldberg, line 120: [Refactor] Every one of the uses for this function does >=1 `===` check. How about using a `Set` instead to make it more clear what's happening?

```suggestion
  symbolNames: Set<string>
```

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: I realized a situation where this option would be helpful! [<a href="https://deploy-preview-8094--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=CYUwxgNghgTiAEAzArgOzAFwJYHtVJAzAAsAKZGCALngGcMYtUBzAShoAUYcBbLWkAB4cAIwBW4DAD4A3ACg5KdNjzwQADyg8ADhBClW8AN5z48OBgr4uvfiAB0UCBFIBtU2YJEyAIgD0PqwANB5mPkzwGMQIfMDAej4hnl4kpP6BHgC6rPIAvkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6fAC0SYEMAjJLWa1i0APYBbSskS0eAc3nRE8nvjHRUGSImjjokcGAC%2BIY0A&tsconfig=&tokens=false" title="Playground link with example function returning a Promise.all tuple">playground link</a>]

```ts
declare function fetch(url: string): Promise<object>;

function example() {
  return Promise.all([
    fetch("/"),
    "in the middle",
    fetch("/")
  ]);
}
```

If I was the developer here using `Promise.all` directly to avoid having to make & manage a `[object, string, object]` tuple, I'd be annoyed at the rule asking me to explicitly `Promise.resolve("in the middle")`. There's no benefit in this situation beyond consistency - at the cost of simplicity. 

But, I get what you're saying that this is really the main point of the rule. I think it's fine for now to leave as-is. Since it's only enabled in the strict config, it should just show up for users who have opted into the more strict opinions.

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: That first use case seems pretty niche to me. In that case I'd think a type guard _would_ be the right thing to use. Maybe I'm being overly prescriptive. 😄 

Either way, here's an example with an `unknown` that blatantly doesn't need to be passed in [<a href="https://deploy-preview-8094--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=IYZwngdgxgBAZgV2gFwJYHsIwDboOYBC6yAFgBQBQMMeApsgGKoBOIyAXDGQJQwC8APhgAFZugC2qELQA8bZqgh4BAGioxpUTABMAssDAAjWpyQBrCOgDuENbwDe6rRDYwA2nBZsVG2s%2B0AuvwwwFbAqMgiYpLSAHTA2NhkbnSMXsg8Ppo6%2Bka0AdwA3BROmCDo2LSxuHhk9vDpWX46MAC%2BRRStQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6fAC0SYEMAjJLWa1i0APYBbSskS0eAc3nRE8nvjHRUGSImjjokcGAC%2BIY0A&tsconfig=&tokens=false" title="Playground link with logBoth function passing unknown to Promise.all">playground link</a>]:

```ts
async function logBoth(
  getFirst: () => Promise<string>,
  secondMaybe: unknown,
) {
  const [first, second] = await Promise.all([getFirst(), secondMaybe]);

  console.log({ first, second });
}
```

I think the rule should complain on this by default and have an option to not.

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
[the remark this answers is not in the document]
Josh-Cena: Please do add an option that allows mixed arrays. I often parallel a bunch of tasks and sometimes tasks shift between being sync and async (for example, because it no longer dynamically imports a library, or it no longer reads a config file). I really don't want to break the symmetry in this case.

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,215 @@
...
+      }
+
+      return predicate(type.value);
+    }
+
+    function isMemberName(
+      node:
+        | TSESTree.MemberExpressionComputedName
+        | TSESTree.MemberExpressionNonComputedName,
+      predicate: (name: string) => boolean,
```
JoshuaKGoldberg, line 118: [Refactor] Similar to the `isBuiltinSymbolLike` comment, this is only ever used with a `Set<string>`. I think expanding the use case to predicates is just a bit more complexity than we need.

```suggestion
      memberNames: Set<string>
```

JoshuaKGoldberg (changes requested):
This is looking great, thanks! 🚀 

Just requesting changes on the use cases I commented in threads.

packages/type-utils/src/builtinSymbolLikes.ts
auvred: By the way, these functions are already included in several public releases. Shouldn't we keep the already published signatures to avoid breaking changes? Perhaps we can add this new functionality as overload/union? Although these packages are not documented yet, not sure, just asking :smile:

Tjstretchalot: Predicates are more composable than sets and offer easier ergonomics; for sets; `(v) => symbolNameSet.has(v)` is just as readable as `symbolNameSet`, and for looking up a single value `(symbolName) => symbolName === 'asdf'` is in my opinion much easier to interpret than `new Set(['foo'])` which offers no context for what the set is for, so you end up needing either a comment or breaking it out into a constant. Finally, predicates can be more consistently applied; e.g., we already have predicates for `isBuiltinSymbolLikeRecurser`, so switching between predicates and sets for arguments requires constantly deciding which ones "tend to be used like sets". Furthermore, predicates portray the intent of the argument more clearly: a set has many functions besides just contains-checking, including iteration. Providing a predicate makes it clear the only part the library function needs is the yes/no answer on each element.

For @auvred 's comment; happy to add the overload/union; adds a little clutter though, so will wait for more discussion since I'm not familiar enough with the project to have any opinion for backwards compatibility vs tidiness for those helpers

Tjstretchalot:
For some reason I can't respond to some these individually, at least as far as I can tell, only emote to them. Breaking out here instead

https://github.com/typescript-eslint/typescript-eslint/pull/8094#discussion_r1465013162

Sure, will add default blocking of unknown with an option to unblock. I agree that wouldn't use unknown without a type guard, but the use-case is logically consistent, as it is a type refinement, though one too subtle to be picked up by current typescript (i.e., `Awaited<unknown>` is just unknown now and for the foreseeable future, but there are some things that can be known about it).

However, I do think this is a bit inconsistent since [await-thenable allows unknown](https://deploy-preview-8094--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3hkmBc8ArigNYpIDuKA3AFAAUUaAnimPPQJTwC8AfPADetePCgUoASyw5MtAL6cu1IA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AQwHd3K78ALRE3YAjJOiiJo0APbRI4MAF8QSoA&tsconfig=&tokens=false) (first time trying to use the playground, might have made a mistake, but this is definitely clear in the source code)

https://github.com/typescript-eslint/typescript-eslint/pull/8094#discussion_r1465021421

Responded in https://github.com/typescript-eslint/typescript-eslint/pull/8094#discussion_r1465271687 - if my argument for predicates isn't convincing will switch it over with an unambigious blocking remark

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: cc @bradzacher per https://github.com/typescript-eslint/typescript-eslint/issues/8301#issuecomment-1910253980 - would you say the same logic applies here? That we should ignore `unknown`s?

Tjstretchalot:
Made suggested documentation changes and rebased to latest main; going to allow time for more discussion on predicates for that change and for a conclusion to #8301 before changing the behavior of unknown

Tjstretchalot:
Updated to latest main branch and ensured this doesn't change the library functions in a backwards incompatible way (cc: @auvred https://github.com/typescript-eslint/typescript-eslint/pull/8094#discussion_r1465034144). Initially I did it via overloads so I could mark the string one deprecated, but that fails linting with `These overloads can be combined into one signature`, so this leaves no deprecation path for the string variant.

As https://github.com/typescript-eslint/typescript-eslint/issues/8301 appears to be resolved with wontfix I'll leave unknown behavior as is.

cc: @JoshuaKGoldberg for updating labels, deciding if predicate->set change is blocking

JoshuaKGoldberg:
Super, thanks! Putting back on the queue to review :)

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: In that case, +1ing Josh's request 🙂

packages/eslint-plugin/src/rules/thenable-in-promise-aggregators.ts
```diff
@@ -0,0 +1,215 @@
...
+      if (type.isIntersection() || type.isUnion()) {
+        return type.types.some(t =>
+          isIndexableWithSomeElementsLike(t, predicate),
+        );
+      }
+
+      if (!checker.isArrayType(type) && !checker.isTupleType(type)) {
+        const indexType = checker.getIndexTypeOfType(type, ts.IndexKind.Number);
+        if (indexType === undefined) {
+          return false;
```
JoshuaKGoldberg, line 79: [Testing] Missing unit test coverage for this and a few other lines. If it's a case where we _know_ the value is definitely not nullish (e.g. an array type having a numeric index) then a `!` is reasonable.

packages/type-utils/src/builtinSymbolLikes.ts
JoshuaKGoldberg, line 1: [Non-Actionable] Every time I come back to this PR I get confused as to how such an important-looking file has such major gaps in unit test coverage... it's #6701, but for this package.

Filed #8358 to track adding in unit tests separately.

packages/type-utils/src/builtinSymbolLikes.ts
JoshuaKGoldberg: > backwards compatibility

Yup, this'll need to retain backwards compatibility. I haven't heard of anybody telling us they use it, but [a Sourcegraph query for TypeScript files referring to it](https://sourcegraph.com/search?q=context:global+%40typescript-eslint/type-utils+lang:TypeScript++-repo:%5Egithub%5C.com/typescript-eslint%28.*%29+-file:%28.*%29node_modules%28.*%29&patternType=standard&sm=1) shows a few public repos that do.

> `(v) => symbolNameSet.has(v)` is just as readable as `symbolNameSet`

I think this is subjective 😄 I find the latter more intuitive. In my mind I'm kind of implicitly applying the context from _"is built-in symbol like"_ to the last parameter. But that's just me.

> offers no context for what the set is for, so you end up needing either a comment or breaking it out into a constant

This is equally applicable to both a predicate and a set. For both, a descriptive parameter name like `allowedSymbolNames` or `isAllowedSymbolName` fixes the context issue.

> predicates portray the intent of the argument more clearly
> a set has many functions besides ...

I suppose `Pick<Set<String>, 'has'>` would be the most precise thing to take in. All we really want in any of this is to know whether a name is what we're looking for.

---


---

JoshuaKGoldberg: > backwards compatibility

Yup, this'll need to retain backwards compatibility. I haven't heard of anybody telling us they use it, but [a Sourcegraph query for TypeScript files referring to it](https://sourcegraph.com/search?q=context:global+%40typescript-eslint/type-utils+lang:TypeScript++-repo:%5Egithub%5C.com/typescript-eslint%28.*%29+-file:%28.*%29node_modules%28.*%29&patternType=standard&sm=1) shows a few public repos that do.

> `(v) => symbolNameSet.has(v)` is just as readable as `symbolNameSet`

I think this is subjective 😄 I find the latter more intuitive. In my mind I'm kind of implicitly applying the context from _"is built-in symbol like"_ to the last parameter. But that's just me.

> offers no context for what the set is for, so you end up needing either a comment or breaking it out into a constant

This is equally applicable to both a predicate and a set. For both, a descriptive parameter name like `allowedSymbolNames` or `isAllowedSymbolName` fixes the context issue.

> predicates portray the intent of the argument more clearly
> a set has many functions besides ...

I suppose `Pick<Set<String>, 'has'>` would be the most precise thing to take in. All we really want in any of this is to know whether a name is what we're looking for.

---

Anyway, the issue of whether we use a set or predicate is not particularly important to me. It's a small implementation detail and if you feel strongly about it I don't mind just shrugging and seeing what happens.

JoshuaKGoldberg: > backwards compatibility

Yup, this'll need to retain backwards compatibility. I haven't heard of anybody telling us they use it, but [a Sourcegraph query for TypeScript files referring to it](https://sourcegraph.com/search?q=context:global+%40typescript-eslint/type-utils+lang:TypeScript++-repo:%5Egithub%5C.com/typescript-eslint%28.*%29+-file:%28.*%29node_modules%28.*%29&patternType=standard&sm=1) shows a few public repos that do.

> `(v) => symbolNameSet.has(v)` is just as readable as `symbolNameSet`

I think this is subjective 😄 I find the latter more intuitive. In my mind I'm kind of implicitly applying the context from _"is built-in symbol like"_ to the last parameter. But that's just me.

> offers no context for what the set is for, so you end up needing either a comment or breaking it out into a constant

This is equally applicable to both a predicate and a set. For both, a descriptive parameter name like `allowedSymbolNames` or `isAllowedSymbolName` fixes the context issue.

> predicates portray the intent of the argument more clearly
> a set has many functions besides ...

I suppose `Pick<Set<String>, 'has'>` would be the most precise thing to take in. All we really want in any of this is to know whether a name is what we're looking for.

One flaw of functions is it leaks the implementation detail of _"this gets called up to once per symbol name"_, so folks can do weird things like:

```ts
isBuiltinSymbolLike(program, type, (name) => {
  someRegistry.add(name);
  if (someAttribute(name)) {
    context.report({ data: { name }, messageId: "someMessage", node: someNode });
  }
});
```

Just pointing this out as a counterpoint to the idea that a predicate adds more practical or semantic safety. I don't think we have to worry about anybody doing such an odd thing.

---

Anyway, the issue of whether we use a set or predicate is not particularly important to me. It's a small implementation detail and if you feel strongly about it I don't mind just shrugging and seeing what happens.

JoshuaKGoldberg (changes requested):
Just a unit test coverage request that's blocking from me right now 🙂 great stuff!

Edit: oh, and the request for an option around mixed arrays.

(I'll re-review more deeply once the comments are all resolved)

JoshuaKGoldberg:
👋 Hey @Tjstretchalot! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

Tjstretchalot:
Hey @JoshuaKGoldberg - apologies for this; work got very busy rather quickly and had to reduce some other stuff to make space. It'll hopefully get better in April, but if someone else wants to pick it up or you want to close until then let me know

JoshuaKGoldberg:
Thanks, good to know. I'll remove `stale` for now but if by the middle of the month it's still pending I'd say we should free it up then. Good luck with the work!

JoshuaKGoldberg:
Closing as it's been a bit - thanks for getting started on it! If you do end up having time to continue @Tjstretchalot feel free to reopen this PR if it's unlocked, or send a new one otherwise. 

In the meantime, if anybody else wants to work on the linked issue, go right ahead. Just be sure to add a co-author attribution if you use any code from this PR. Cheers!

PR #8108, opened 2023-12-21
fix(eslint-plugin): [no-restricted-imports] prevent crash when `patterns` or `paths` in options are empty

### LINKED ISSUE #8103: Bug: `no-restricted-imports` crashes when `paths` & `patterns` are both empty, but defined

### Playground Link

https://typescript-eslint.io/play/#ts=5.3.2&fileType=.tsx&code=JYWwDg9gTgLgBAKjgQwM5wCoFNXwGZQQhwBEMOMAdAFaokDcAUIwMYQB2qEANlpdxADmACmy4AlPSA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaFAF2gEsBjUxAE0IoFsAHfaU1DSfAM18gAacNkgABUgE8WKKpRalCKeBVykA9AWJlKNeo1btO6MAG0R2KImjR2Qi9iyXRLAIala0XFzMBdBxAAvg7%2BQSCBQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
import * as Test from "test.js";

console.log(Test);
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "rules": {
      "no-restricted-imports": "off",
      "@typescript-eslint/no-restricted-imports": [
        "error",
        {
          "patterns": [],
          // (alternatively)
          // "paths": []
        }
      ]
    }
  },
};
```

### tsconfig

_No response_

### Expected Result

- ESLint does not crash

### Actual Result

- ESLint crashes with the following error:  
  > TypeError: rules.ImportDeclaration is not a function
Occurred while linting ...filename...
Rule: "@typescript-eslint/no-restricted-imports"

### Additional Info

Looking at the code of both base rule & extension rule, it seems the base rule does the following check and bails out if there's nothing to lint (it returns nothing in this case):
https://github.com/eslint/eslint/blob/5aa9c499da48b2d3187270d5d8ece71ad7521f56/lib/rules/no-restricted-imports.js#L192-L195

However, the extension rule does not check for that, nor checks whether base rule did not return anything useful:
https://github.com/typescript-eslint/typescript-eslint/blob/6128a02cb15d500fe22fe265c83e4d7a73ae52c3/packages/eslint-plugin/src/rules/no-restricted-imports.ts#L228-L233

This is most likely not a usual case to handle (as "no paths and patterns" usually means we don't need the rule enabled), I've discovered it only because our shared eslint config has this rule configurable, and one of the projects did not need any exclude patterns. However, it would still be nice if this rule didn't crash the entire ESLint's execution, and handle it like the base rule does.

Tested with:
- ESLint: 8.56.0
- @typescript-eslint/*: 6.15.0

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8103

## Overview

Thanks to @mdziekon for investigation made in #8103!

Basically the extension rule didn't check if the base rule returned listeners or not (its typing was wrong, base rule may return empty object https://github.com/eslint/eslint/blob/5aa9c499da48b2d3187270d5d8ece71ad7521f56/lib/rules/no-restricted-imports.js#L192-L195)

### DIFF
--- a/packages/eslint-plugin/src/rules/no-restricted-imports.ts
+++ b/packages/eslint-plugin/src/rules/no-restricted-imports.ts
@@ -216,7 +216,7 @@ function shouldCreateRule(
   baseRules: RuleListener,
   options: Options,
 ): baseRules is Exclude<RuleListener, Record<string, never>> {
-  if (!Object.keys(baseRules).length || options.length === 0) {
+  if (Object.keys(baseRules).length === 0 || options.length === 0) {
     return false;
   }
 
--- a/packages/eslint-plugin/src/rules/no-restricted-imports.ts
+++ b/packages/eslint-plugin/src/rules/no-restricted-imports.ts
@@ -221,7 +221,7 @@ function shouldCreateRule(
   }
 
   if (!isOptionsArrayOfStringOrObject(options)) {
-    return !!(options[0].paths?.length || options[0].patterns?.length);
+    return options[0].paths?.length !== 0 || options[0].patterns?.length !== 0;
   }
 
   return true;
--- a/packages/eslint-plugin/src/rules/no-restricted-imports.ts
+++ b/packages/eslint-plugin/src/rules/no-restricted-imports.ts
@@ -8,6 +8,7 @@ import type {
 import type {
   ArrayOfStringOrObject,
   ArrayOfStringOrObjectPatterns,
+  RuleListener,
 } from 'eslint/lib/rules/no-restricted-imports';
 import type { Ignore } from 'ignore';
 import ignore from 'ignore';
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/no-restricted-imports.ts
```diff
@@ -211,6 +212,21 @@ function getRestrictedPatterns(
   return [];
 }
 
+function shouldCreateRule(
+  baseRules: RuleListener,
+  options: Options,
+): baseRules is Exclude<RuleListener, Record<string, never>> {
+  if (!Object.keys(baseRules).length || options.length === 0) {
```
StyleShit, line 219: [Nit]
Maybe use the same "is empty" logic for both? 😅 

```suggestion
  if (!Object.keys(baseRules).length || ! options.length) {
```

OR 

```suggestion
  if (Object.keys(baseRules).length === 0 || options.length === 0) {
```

StyleShit (changes requested):

packages/eslint-plugin/src/rules/no-restricted-imports.ts
auvred: Yup, you're right, it looks much cleaner, thx!

JoshuaKGoldberg (approved):
👍 thanks!

PR #8119, opened 2023-12-24
chore(website): [playground] add twoslash queries

### LINKED ISSUE #8055: Website: Add twoslash queries to the playground

### Suggested Changes

[TypeScript Twoslash](https://marketplace.visualstudio.com/items?itemName=Orta.vscode-twoslash-queries) is a lovely practice for using `// ^?` comments to visualize what type a particular expression is. [It's on the TypeScript playground (example)](https://www.typescriptlang.org/play?#code/DYUwLgBAtgngagQ2AVxBAvBAjAJgMwDcAUAPQkQQB6A-EUA). We don't have it on our playground though. Let's add it in?

### Affected URL(s)

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8055

## Overview

The issue is not marked as `accepting prs`, so I'm opening it as a draft for now

---

I took [this code](https://github.com/microsoft/TypeScript-Website/blob/c8b2ea8c8fc216c163fe293650b2666c8563a67d/packages/playground/src/twoslashInlays.ts) from `microsoft/Typescript-Website` and refactored it a bit

### DIFF
--- a/packages/website/src/components/editor/createProvideTwoslashInlay.ts
+++ b/packages/website/src/components/editor/createProvideTwoslashInlay.ts
@@ -1,0 +1,100 @@
+// The following code is adapted from the code in microsoft/TypeScript-Website.
+// Source: https://github.com/microsoft/TypeScript-Website/blob/c8b2ea8c8fc216c163fe293650b2666c8563a67d/packages/playground/src/twoslashInlays.ts
+// License: https://github.com/microsoft/TypeScript-Website/blob/c8b2ea8c8fc216c163fe293650b2666c8563a67d/LICENSE-CODE
+
+import type Monaco from 'monaco-editor';
+import type * as ts from 'typescript';
+
+import type { SandboxInstance } from './useSandboxServices';
+
+const twoslashQueryRegex = /^(\s*\/\/\s*\^\?)\s*$/gm;
+function findTwoshashQueries(code: string): RegExpExecArray[] {
+  let match: RegExpExecArray | null = null;
+  const matches: RegExpExecArray[] = [];
+  while ((match = twoslashQueryRegex.exec(code))) {
+    matches.push(match);
+  }
+  return matches;
+}
+
+export function createTwoslashInlayProvider(
+  sandbox: SandboxInstance,
+): Monaco.languages.InlayHintsProvider {
+  return {
+    provideInlayHints: async (
+      model,
+      _,
+      cancel,
+    ): Promise<Monaco.languages.InlayHintList> => {
+      const worker = await sandbox.getWorkerProcess();
+      if (model.isDisposed() || cancel.isCancellationRequested) {
+        return {
+          hints: [],
+          dispose(): void {
+            /* nop */
+          },
+        };
+      }
+
+      const queryMatches = findTwoshashQueries(model.getValue());
+
+      const results: Monaco.languages.InlayHint[] = [];
+
+      for (const result of await Promise.all(
+        queryMatches.map(q => resolveInlayHint(q)),
+      )) {
+        if (result) {
+          results.push(result);
+        }
+      }
+
+      return {
+        hints: results,
+        dispose(): void {
+          /* nop */
+        },
+      };
+
+      async function resolveInlayHint(
+        queryMatch: RegExpExecArray,
+      ): Promise<Monaco.languages.InlayHint | undefined> {
+        const end = queryMatch.index + queryMatch[1].length - 1;
+        const endPos = model.getPositionAt(end);
+        const inspectionPos = new sandbox.monaco.Position(
+          endPos.lineNumber - 1,
+          endPos.column,
+        );
+        const inspectionOff = model.getOffsetAt(inspectionPos);
+
+        const hint = await (worker.getQuickInfoAtPosition(
+          'file://' + model.uri.path,
+          inspectionOff,
+        ) as Promise<ts.QuickInfo | undefined>);
+        if (!hint?.displayParts) {
+          return;
+        }
+
+        // Make a one-liner
+        let text = hint.displayParts
+          .map(d => d.text)
+          .join('')
+          .replace(/\r?\n\s*/g, ' ');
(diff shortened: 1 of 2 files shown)


packages/website/src/components/editor/createProvideTwoslashInlay.ts
```diff
@@ -0,0 +1,100 @@
+// The following code is adapted from the code in microsoft/TypeScript-Website.
+// Source: https://github.com/microsoft/TypeScript-Website/blob/c8b2ea8c8fc216c163fe293650b2666c8563a67d/packages/playground/src/twoslashInlays.ts
+// License: https://github.com/microsoft/TypeScript-Website/blob/c8b2ea8c8fc216c163fe293650b2666c8563a67d/LICENSE-CODE
```
auvred, line 3: Not sure about licensing
https://github.com/microsoft/TypeScript-Website/tree/c8b2ea8c8fc216c163fe293650b2666c8563a67d?tab=readme-ov-file#legal-notices

packages/website/src/components/editor/useSandboxServices.ts
```diff
@@ -71,6 +72,13 @@ export const useSandboxServices = (
           colorMode === 'dark' ? 'vs-dark' : 'vs-light',
         );
 
+        // registerInlayHintsProvider may not be present for older TS versions
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        sandboxInstance.monaco.languages.registerInlayHintsProvider?.(
+          sandboxInstance.language,
+          createTwoslashInlayProvider(sandboxInstance),
+        );
```
auvred, line 80: Without optional chaining it fails on TS 4.3.5

error on TS 4.3.5

packages/website/src/components/editor/useSandboxServices.ts
```diff
@@ -71,6 +72,13 @@ export const useSandboxServices = (
           colorMode === 'dark' ? 'vs-dark' : 'vs-light',
         );
 
+        // registerInlayHintsProvider may not be present for older TS versions
```
JoshuaKGoldberg, line 75: Yeah "inlay hints" were added in TS 4.4: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#inlay-hints

This strategy makes sense - if TS doesn't support them, we just won't enable them.

One request, to make this easier to find when we grep for version things to drop upon losing TS 4.3 support:

```suggestion
        // registerInlayHintsProvider was added in TS 4.4 and isn't in TS <= 4.3.
```

packages/website/src/components/editor/createProvideTwoslashInlay.ts
JoshuaKGoldberg: +1 on think of the licensing!

This should be sufficient IMO. I'm not a lawyer so don't take this as legal advice 😄 but given that we're including a link to the original MIT license & source code, I'm happy.

packages/website/src/components/editor/createProvideTwoslashInlay.ts
```diff
@@ -0,0 +1,100 @@
...
+
+import type Monaco from 'monaco-editor';
+import type * as ts from 'typescript';
+
+import type { SandboxInstance } from './useSandboxServices';
+
+const twoslashQueryRegex = /^(\s*\/\/\s*\^\?)\s*$/gm;
+function findTwoshashQueries(code: string): RegExpExecArray[] {
```
JoshuaKGoldberg, line 11: [Refactor] This might be working now, but in general it's safer to recreate exec-ed regexes each time the execs are started.

```suggestion
function findTwoshashQueries(code: string): RegExpExecArray[] {
  const twoslashQueryRegex = /^(\s*\/\/\s*\^\?)\s*$/gm;
```

packages/website/src/components/editor/createProvideTwoslashInlay.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        const hint = await (worker.getQuickInfoAtPosition(
+          'file://' + model.uri.path,
+          inspectionOff,
+        ) as Promise<ts.QuickInfo | undefined>);
+        if (!hint?.displayParts) {
+          return;
+        }
+
+        // Make a one-liner
```
JoshuaKGoldberg, line 77: [Style] I don't know what this comment is trying to convey other than that it is a one-liner. Seems useless to me 😄 

```suggestion
```

(ACK that it comes from the original code)

packages/website/src/components/editor/createProvideTwoslashInlay.ts
```diff
@@ -0,0 +1,100 @@
...
+        let text = hint.displayParts
+          .map(d => d.text)
+          .join('')
+          .replace(/\r?\n\s*/g, ' ');
+        if (text.length > 120) {
+          text = text.slice(0, 119) + '...';
+        }
+
+        const inlay: Monaco.languages.InlayHint = {
```
JoshuaKGoldberg, line 86: [Style] Nit: the function already has a return type, so I don't think a separate variable declaration gives us much?

```suggestion
        return {
```

JoshuaKGoldberg (dismissed):
👏 This looks great, thanks for such a clean PR!

No blocking requests from me - just a comment change for a more searchable TS 4.3/4.4 reference, and a couple of small style touchups that we can ignore them if you hate them 😄. WDYT?

packages/website/src/components/editor/useSandboxServices.ts
auvred: Good point! Thanks

packages/website/src/components/editor/createProvideTwoslashInlay.ts
auvred: You're right, good catch! About a year ago I spent _two days_ catching a bug in `vuejs/core`. It turned out to be caused by one small stateful regex with the global flag......

PR #8142, opened 2023-12-27
test(eslint-plugin): assert that `ts`/`tsx` code blocks in docs are syntactically valid

### LINKED ISSUE #8134: Repo: add a test asserting that all code blocks in docs are syntactically valid
### Suggestion

Follow up to https://github.com/typescript-eslint/typescript-eslint/issues/8133

I propose adding a new `docs` test that will assert that the code blocks (with `Open in Playground` button) contain at least syntatically valid code

Perhaps we could also test that code under _**Incorrect**_ tab contains at least one lint error, and _**Correct**_ code contains no errors

I wrote a simple test that tries to parse content of all code blocks. It turns out that the following files contain code block with syntax error:
```
docs/rules/consistent-type-assertions.md
docs/rules/explicit-function-return-type.md
docs/rules/member-ordering.md
docs/rules/no-empty-function.md
docs/rules/no-unnecessary-type-constraint.md
docs/rules/prefer-enum-initializers.md
docs/rules/prefer-readonly-parameter-types.md
docs/rules/space-before-blocks.md
docs/rules/typedef.md
```

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8134

## Overview

For now, just trying to parse each code block with `ts` and `tsx` lang


packages/eslint-plugin/docs/rules/explicit-function-return-type.md
```diff
@@ -172,10 +172,6 @@ functionWithObjectArg({
     return 1;
   },
 });
-
-const Comp: FC = () => {
-  return <button onClick={() => {}} />;
-};
```
auvred, line 178: We can't mix angle bracket assertions and jsx syntax

packages/eslint-plugin/docs/rules/consistent-type-assertions.md
```diff
@@ -54,22 +54,22 @@ Examples of code for `{ assertionStyle: 'as', objectLiteralTypeAssertions: 'neve
 #### ❌ Incorrect
 
 ```ts option='{ "assertionStyle": "as", "objectLiteralTypeAssertions": "never" }'
-const x = { ... } as T;
+const x = { foo: 1 } as T;
 
 function foo() {
-  return { ... } as T;
+  return { bar: 1 } as T;
```
StyleShit, line 60: [Nit]
I'm wondering if a reader will get confused thinking `T` is the same type in both case

packages/eslint-plugin/tests/docs.test.ts
```diff
@@ -180,6 +181,32 @@ describe('Validating rule docs', () => {
...
       }
+
+      test('must include only valid code samples', () => {
+        for (const token of tokens) {
+          if (token.type !== 'code') {
+            continue;
+          }
+
+          const lang = token.lang?.trim();
+          if (!lang?.startsWith('ts')) {
```
StyleShit, line 192: in case you wanna be stricter, and because I just _love_ regex...

```suggestion
          if (!/^tsx?\b/i.test(lang))) {
```

Josh-Cena: Also do we have any `jsx?` blocks?

packages/eslint-plugin/docs/rules/consistent-type-assertions.md
auvred: Good point, thanks, I changed it https://github.com/typescript-eslint/typescript-eslint/pull/8142/commits/9b7a3f175414e943ad749d0877dee8bd1ef7e95d

packages/eslint-plugin/tests/docs.test.ts
auvred: > [Even Nitier] in case you wanna be stricter, and because I just _love_ regex...

IMO, `startsWith` looks more readable, especially for texts, although they are already generated automatically, which makes them more complex than the usual ones...

StyleShit: TBH, I think it's also more performant then regex 😅

auvred: > Also do we have any `jsx?` blocks?

Good catch, thanks, we don't have any `jsx` blocks, but we had a few `js` blocks (I changed them to `ts`)

packages/eslint-plugin/tests/docs.test.ts
```diff
@@ -180,6 +181,32 @@ describe('Validating rule docs', () => {
...
+            parseForESLint(token.text, {
+              ecmaFeatures: {
+                jsx: lang.startsWith('tsx'),
+              },
+              ecmaVersion: 'latest',
+              sourceType: 'module',
+              range: true,
+            });
+          } catch (e) {
+            throw new Error(`Parsing error:\n\n${token.text}`);
```
JoshuaKGoldberg, line 206: Nit: a bit more informative-ness in the error if we use `cause`, what do you think?

```suggestion
          } catch (e) {
            throw new Error(`Parsing error:\n\n${token.text}`, { cause: e });
```

Alternately my suggestion would be:

```suggestion
          } catch {
            throw new Error(`Parsing error:\n\n${token.text}`);
```

...but my guess is showing the caught error would be nicer.

JoshuaKGoldberg (approved):
LGTM, thanks!

I know @Josh-Cena has done a lot with Docusaurus things so leaving open in case they have a chance to review. Will come and merge back in a week otherwise.

packages/eslint-plugin/tests/docs.test.ts
auvred: What about this?

```ts
try {
  // ...
} catch (e) {
  if (
    e &&
    typeof e === 'object' &&
    'lineNumber' in e &&
    typeof e.lineNumber === 'number' &&
    'column' in e &&
    typeof e.column === 'number' &&
    'message' in e
  ) {
    const codeFrameLines = token.text.split('\n');
    codeFrameLines.splice(
      e.lineNumber,
      0,
      ' '.repeat(e.column) + '^: ' + e.message,
    );
    throw new Error(`Parsing error:\n\n${codeFrameLines.join('\n')}`);
  }
  throw new Error(`Parsing error:\n\n${token.text}`);
}
```

The code looks pretty scary, but what it does - it checks if `e` is `TSError`

https://github.com/typescript-eslint/typescript-eslint/blob/c61c3eedf5dec98c1548c8ab533dcbb756a9cd7b/packages/typescript-estree/src/node-utils.ts#L688

And if `e` looks like `TSError` it prints code frame with pointer on the location of parsing error: 

```
  ● Validating rule docs › consistent-type-exports.md › must include only valid code samples

    Parsing error:

    interface ButtonProps {
      onClick: () => vo id;
                        ^ ';' expected.
    }

    class Button implements ButtonProps {
      onClick = () => console.log('button!');
    }

    export { Button, ButtonProps };

      219 |                 ' '.repeat(e.column) + '^ ' + e.message,
      220 |               );
    > 221 |               throw new Error(`Parsing error:\n\n${codeFrameLines.join('\n')}`);
          |                     ^
      222 |             }
      223 |             throw new Error(`Parsing error:\n\n${token.text}`);
      224 |           }

      at Object.<anonymous> (tests/docs.test.ts:221:21)
```

(We can't import `TSError` from `@typescript-eslint/typescript-estree` because of `no-typescript-estree-import` internal rule)

JoshuaKGoldberg: That seems like a lot of effort for something that doesn't happen very often & has a pretty clear error by default 😄

auvred: Yup, you're right, but we have a little issue with `cause`:

It's defined in `ES2022`: https://github.com/microsoft/TypeScript/blob/01a51d2d017dd50d487835282b7e4a2a4f7300ab/src/lib/es2022.error.d.ts,
but we have `ES2021`:
https://github.com/typescript-eslint/typescript-eslint/blob/78db0add16e9d7ea23c5ffc13341cc13b16eec2d/tsconfig.base.json#L17

Soo..

```
$ yarn tsc --noEmit
tests/docs.test.ts:206:64 - error TS2554: Expected 0-1 arguments, but got 2.

206             throw new Error(`Parsing error:\n\n${token.text}`, { cause: e });
                                                                   ~~~~~~~~~~~~


Found 1 error in tests/docs.test.ts:206
```

I guess we stick with your second suggestion :slightly_smiling_face:

packages/eslint-plugin/tests/docs.test.ts
```diff
@@ -180,6 +181,32 @@ describe('Validating rule docs', () => {
...
+
+      test('must include only valid code samples', () => {
+        for (const token of tokens) {
+          if (token.type !== 'code') {
+            continue;
+          }
+
+          const lang = token.lang?.trim();
+          if (!lang?.startsWith('ts')) {
+            return;
```
Josh-Cena, line 193: Why is this `return` instead of `continue`?

packages/eslint-plugin/tests/docs.test.ts
```diff
@@ -180,6 +181,32 @@ describe('Validating rule docs', () => {
...
+          }
+
+          try {
+            parseForESLint(token.text, {
+              ecmaFeatures: {
+                jsx: lang.startsWith('tsx'),
```
Josh-Cena, line 199: I still think that instead of `startsWith` you should use regex. Unless you plan to support things like `tsx-noformat` (a trick we use in MDN to stop Prettier from formatting code). This way we can be future proof and avoid some confusing behaviors.

packages/eslint-plugin/tests/docs.test.ts
```diff
@@ -180,6 +181,32 @@ describe('Validating rule docs', () => {
...
+              },
+              ecmaVersion: 'latest',
+              sourceType: 'module',
+              range: true,
```
Josh-Cena, line 203: Why this?

Josh-Cena (changes requested):
Mostly LGTM

packages/eslint-plugin/tests/docs.test.ts
auvred: Thanks for catching this! It turns out `docs/rules/explicit-function-return-type.md` also had code blocks with syntax errors, but because of this eager return, the tests weren't catching them

packages/eslint-plugin/tests/docs.test.ts
auvred: Changed it

packages/eslint-plugin/tests/docs.test.ts
auvred: Without `range: true` parser throws `Cannot read properties of undefined (reading '0')` on valid code

Josh-Cena (approved):
LGTM

PR #8143, opened 2023-12-27
fix(eslint-plugin): [no-unnecessary-type-assertion] detect unnecessary non-null-assertion on a call expression

### LINKED ISSUE #8141: Bug: [no-unnecessary-type-assertion] does not detect unnecessary non-null-assertion on a call expression

https://typescript-eslint.io/rules/no-unnecessary-type-assertion

### Description

I propose that the `no-unnecessary-type-assertion` rule should also check for the return type of a function.

### Fail

```TypeScript
declare function foo(): number;
const a = foo()!; // no error report
const b = foo() as number; // report error but bad fixed: `const b = foo() ; `
const c = <number>foo(); // good now
```

### Pass

```TypeScript
declare function foo(): number;
const a = foo();
const b = foo();
const c = foo();
```

### Additional Info

Im ready to send a PR after the issue is marked with the `accepting prs` label as indicated on the PR Checklist. Welcome to provide more test cases.

lvjiaxuan:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8141

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
@@ -180,10 +180,7 @@ export default createRule<Options, MessageIds>({
             node,
             messageId: 'unnecessaryAssertion',
             fix(fixer) {
-              return fixer.removeRange([
-                node.expression.range[1],
-                node.range[1],
-              ]);
+              return fixer.removeRange([node.range[1] - 1, node.range[1]]);
             },
           });
         } else {
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
@@ -169,7 +169,10 @@ export default createRule<Options, MessageIds>({
         const type = getConstrainedTypeAtLocation(services, node.expression);
 
         if (!isNullableType(type)) {
-          if (isPossiblyUsedBeforeAssigned(node.expression)) {
+          if (
+            node.expression.type === AST_NODE_TYPES.Identifier &&
+            isPossiblyUsedBeforeAssigned(node.expression)
+          ) {
             return;
           }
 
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
JoshuaKGoldberg, line 1: [Cleanup] I can't comment on an unchanged file 😄 but it looks like this bugfix found an unnecessary `!` assertion in `packages/eslint-plugin/src/rules/no-unnecessary-condition.ts`! Great!

We ensure that lint rules pass on our current codebase as a form of testing. So that'll need to be fixed as part of this PR.

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
JoshuaKGoldberg, line 1: [Testing] Interestingly, the fact that this line in `no-unnecessary-condition.ts` wasn't caught by the rule:

```ts
const callback = node.arguments[0]!;
```

...indicates a gap in code coverage for the rule which this change also fixes. Could you please add test coverage that case too?

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -169,18 +169,18 @@ export default createRule<Options, MessageIds>({
...
 
           context.report({
             node,
             messageId: 'unnecessaryAssertion',
             fix(fixer) {
-              return fixer.removeRange([
-                node.expression.range[1],
-                node.range[1],
-              ]);
+              return fixer.removeRange([node.range[1] - 1, node.range[1]]);
```
JoshuaKGoldberg, line 183: [Praise] Nice call for the parenthesis!

JoshuaKGoldberg (changes requested):
Good stuff! Looks like it works well, so just requesting changes on test coverage and fixing the internal lint complaint.

Btw this is a bugfix rather than a feature so I've edited the PR title. They show up differently in the changelog and have different semver version bump implications.

Girl from the Muppet Christmas Carol saying 'wonderful' happily

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
lvjiaxuan: okay, I will do it later.

JoshuaKGoldberg (approved):
Swell, thanks! 🙌

PR #8146, opened 2023-12-27
fix(utils): improve error message on typed rule with invalid parser

### LINKED ISSUE #6814: Feature: getParserServices should throw informative error when used with non-typescript-eslint parser

### Relevant Package

utils

### Playground Link

_No response_

### Repro Code

```typescript
const parserServices = getParserServices(context, true);
```

### ESLint Config

```json
{
  "parser": "@babel/eslint-parser",
  "parserOptions": {
    "ecmaVersion": "latest",
    "requireConfigFile": false,
    "babelOptions": {
      "babelrc": false,
      "configFile": false
    }
  }
}
```

### tsconfig

```json
{
  "compilerOptions": {
    "lib": ["ESNext"],
    "module": "commonjs",
    "moduleResolution": "node",
    "skipLibCheck": true,
    "target": "ESNext"
  },
  "files": ["file.ts"]
}
```

### Expected Result

typeof `parserServices` should be `ParserServicesWithoutTypeInformation`

### Actual Result

An error is thrown:

`You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.`

### Additional Info

This check is testing for `null` or `undefined` but my `context.parserServices` is an empty object and thus the error is thrown.

https://github.com/typescript-eslint/typescript-eslint/blob/v6/packages/utils/src/eslint-utils/getParserServices.ts#L35-L39

Edit: Looks like `parserServices` isn't actually set directly on the `context` object but is instead being inherited through its prototype chain.

### Versions

| `@typescript-eslint/eslint-plugin`     | `6.0.0-alpha.106` |
| `@typescript-eslint/parser`            | `6.0.0-alpha.106` |
| `@typescript-eslint/scope-manager`     | `6.0.0-alpha.106` |
| `@typescript-eslint/typescript-estree` | `6.0.0-alpha.106` |
| `@typescript-eslint/type-utils`        | `6.0.0-alpha.106` |
| `@typescript-eslint/utils`             | `6.0.0-alpha.106` |
| `TypeScript`                           | `5.0.3` |
| `ESLint`                               | `8.36.0` |
| `node`                                 | `18.15.0` |

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6814

## Overview

Adds a small regex in `getParserServices` to catch common cases of incorrect parsers. It's not comprehensive (`../parser.js` unfortunately passes) but it does catch the common case of `node_modules/@babel/*`.

Note that this only augments the existing error message. It doesn't add any new error catching / situations where `getParserServices` would throw.


packages/utils/src/eslint-utils/parserPathSeemsToBeTSESLint.ts
```diff
@@ -0,0 +1,3 @@
+export function parserPathSeemsToBeTSESLint(parserPath: string): boolean {
```
JoshuaKGoldberg, line 1: [Bug] Hmm, I'm remembering now that `@typescript-eslint/parser` is sometimes wrapped by other parser:

* https://github.com/ota-meshi/eslint-plugin-astro/tree/417e6ae85d925f8b36a56a9d04b0db6a966cb5cc#configuration
* https://eslint.vuejs.org/user-guide/#how-to-use-a-custom-parser

So, perhaps checking based on the `parserPath` is not a feasible approach...

JoshuaKGoldberg: Ok, I think I've got a compromise: 

* Here: augmenting the existing error message instead of adding a new reason to throw
* Separating discussion for typed rules out in #8150

bradzacher (approved):

bradzacher:
Note: probably class this as a fix for semver purposes

PR #8152, opened 2023-12-28
feat!(utils): call getParserServices before create() in typed rules

### LINKED ISSUE #8150: Enhancement: Have RuleCreator rules throw an error if used with an incompatible parser

### Relevant Package

utils

### Description

#6814 -> #8146 and the discussion in #8149 are making me think that we really, _really_ don't want users to use a parser that isn't at the very least running `@typescript-eslint/parser` internally. This is for two reasons:

* typescript-eslint rules need to be able to assume the TSESTree shape for the AST
* Type-checked rules need our generated type information

#8146 tries enforcing the exact parser being used - but that won't work for all users because some languages require wrapping parsing with their own (https://github.com/typescript-eslint/typescript-eslint/pull/8146/files#r1437273011). So checking the exact parser isn't likely feasible.

Instead, what if we augment `RuleCreator` so that rules verify some basic assumptions about the parsed AST / context before `create(context)`?

* It can check the root nodes to make sure there's the expected TSESTree `Program` AST node
    * I haven't verified that this is true for other parsers - but likely an equivalent will hold up
* It can it can run the normal checks from `getParserServices` on the `esTreeNodeToTSNodeMap` and `tsNodeToESTreeNodeMap` node maps existing
* If the rule has type checking, it can run the normal check from `getParserServices` on the `program` existing if `!allowWithoutFullTypeInformation`

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8150

## Overview

Throwing up as a draft for reference in triaging #8150.

### DIFF
--- a/packages/utils/src/eslint-utils/RuleCreator.ts
+++ b/packages/utils/src/eslint-utils/RuleCreator.ts
@@ -6,6 +6,7 @@ import type {
   RuleModule,
 } from '../ts-eslint/Rule';
 import { applyDefault } from './applyDefault';
+import { getParserServices } from './getParserServices';
 
 export type { RuleListener, RuleModule };
 
@@ -73,6 +74,13 @@ export function RuleCreator(urlCreator: (ruleName: string) => string) {
         },
       },
       ...rule,
+      create(context, optionsWithDefault): RuleListener {
+        if (meta.docs.requiresTypeChecking) {
+          getParserServices(context);
+        }
+
+        return rule.create(context, optionsWithDefault);
+      },
     });
   };
 }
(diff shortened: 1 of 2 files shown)


packages/utils/src/eslint-utils/RuleCreator.ts
```diff
@@ -73,6 +74,13 @@ export function RuleCreator(urlCreator: (ruleName: string) => string) {
         },
       },
       ...rule,
+      create(context, optionsWithDefault): RuleListener {
+        if (meta.docs.requiresTypeChecking) {
```
bradzacher, line 78: This is essentially an internal-only property that we use for documentation generation purposes - so we can't really rely on it like this.

Additionally this breaks usecases where we want conditional type-awareness - eg `naming-convention` only uses types if and only if you set certain options. So *always* running the util would mean the rule *always* requires parser services - even if you don't use the type-aware options.

There are also going to be usecases externally where people go against our current recommendations and do progressive enhancement of rules wherein this logic would break their usecase entirely.

bradzacher (changes requested):

Issue #8164, opened 2024-01-01
Repo: lint fails on main branch after updating typescript to 5.3.3


auvred:
### Suggestion

https://github.com/typescript-eslint/typescript-eslint/actions/runs/7359445054/job/20034346745#step:5:76

After a little bisecting it turned out that it started failing after https://github.com/typescript-eslint/typescript-eslint/commit/eb9d28c9d5517d85c5263e76df153e5808082c8e

I can't figure out what's causing this, here's TypeScript [`v5.3.2...v.5.3.3` changes](https://github.com/microsoft/TypeScript/compare/v5.3.2...v5.3.3)

These lines were added in https://github.com/typescript-eslint/typescript-eslint/pull/7968

https://github.com/typescript-eslint/typescript-eslint/blob/675e987ca1d13244c03d7e09d4e42c6539689d9a/packages/typescript-estree/src/ts-estree/ts-nodes.ts#L28-L31

Issue #8131, opened 2023-12-25
Repo: Failure on main: Run Unit Tests with Experimental TSServer (eslint-plugin) 


JoshuaKGoldberg:
### Suggestion

It started after merging #7752. https://github.com/typescript-eslint/typescript-eslint/actions/runs/7317242945/job/19932592486:

```plaintext
Summary of all failing tests
 FAIL  tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts (17.679 s, 1025 MB heap size)
  ● no-unused-vars › invalid › 
/*global ����, ��*/
����;
      
    assert(received)
    Expected value to be equal to:
      true
    Received:
      false
    
    Message:
      A fatal parsing error occurred: Parsing error: Invalid character.

      356 |         }
      357 |         const fatalErrorMessage = messages.find(m => m.fatal);
    > 358 |         (0, node_assert_1.default)(!fatalErrorMessage, `A fatal parsing error occurred: ${fatalErrorMessage?.message}`);
          |                                 ^
      359 |         // Verify if autofix makes a syntax error or not.
      360 |         if (messages.some(m => m.fix)) {
      361 |             output = SourceCodeFixer.applyFixes(code, messages).output;

      at RuleTester.runRuleForItem (../rule-tester/dist/RuleTester.js:358:33)
      at RuleTester.runRuleForItem (../rule-tester/dist/RuleTester.js:503:25)
    Message:
      A fatal parsing error occurred: Parsing error: Invalid character.

      356 |         }
      357 |         const fatalErrorMessage = messages.find(m => m.fatal);
    > 358 |         (0, node_assert_1.default)(!fatalErrorMessage, `A fatal parsing error occurred: ${fatalErrorMessage?.message}`);
          |                                 ^
      359 |         // Verify if autofix makes a syntax error or not.
      360 |         if (messages.some(m => m.fix)) {
      361 |             output = SourceCodeFixer.applyFixes(code, messages).output;

      at RuleTester.runRuleForItem (../rule-tester/dist/RuleTester.js:358:33)
      at RuleTester.runRuleForItem (../rule-tester/dist/RuleTester.js:503:25)
      at Object.call (../rule-tester/dist/RuleTester.js:237:119)
 FAIL  tests/eslint-rules/no-undef.test.ts (13.747 s, 631 MB heap size)
  ● no-undef › valid › 
interface IteratorCallback<Subject, Key, Value> {
[...]

PR #8167, opened 2024-01-02
chore: unpin primary node version in ci

### LINKED ISSUE #8166: Repo: codecov coverage reports are not longer uploaded
### Suggestion

This PR https://github.com/typescript-eslint/typescript-eslint/pull/7567 pinned `PRIMARY_NODE_VERSION` variable in CI to `>=20.6.1`
https://github.com/typescript-eslint/typescript-eslint/blob/e566a5dda347470b8ced3cc301b7e4d3e7ed721b/.github/workflows/ci.yml#L16-L17

But this variable is also compared to `matrix.node-version` which van be `16` or `20`
https://github.com/typescript-eslint/typescript-eslint/blob/e566a5dda347470b8ced3cc301b7e4d3e7ed721b/.github/workflows/ci.yml#L194-L217

--- 

`'>=20.6.1' == '20'` -> `false`

Thus, the coverage isn't collected and uploaded

I think we can unpin it since the current version of Node.js is already `20.10.0`

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8166

## Overview

~Sendind as draft to see if this solves the issue~

~UPD: marking as ready for review (I hope issue will get `accepting prs` label soon)~

---

~Currently CI fails because of ~https://github.com/typescript-eslint/typescript-eslint/issues/8164~ and https://github.com/typescript-eslint/typescript-eslint/issues/8131, so we can't check if codecov report will be uploaded or not :(~

UPD: it works: https://github.com/typescript-eslint/typescript-eslint/actions/runs/7412569798/job/20170379090

Codecov bot is not sending a coverage report to this PR because it can't find previous report from `main` (I assume this is because the last report was uploaded in september):

https://app.codecov.io/gh/typescript-eslint/typescript-eslint/pull/8167

why codecov isn't uploading report

<sub>Ohh, 729 commits on main without coverage reports</sub>


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  PRIMARY_NODE_VERSION: '>=20.6.1'
+  PRIMARY_NODE_VERSION: 20
```
JoshuaKGoldberg, line 17: Hmm, can we go the other way? Changing `node-version` to be 16 and the pinned primary node version?

According to https://github.com/orgs/community/discussions/26388 and https://stackoverflow.com/questions/74072206/github-actions-use-variables-in-matrix-definition it's not a straightforward [`16, ${{ env.PRIMARY_NODE_VERSION }}`] 🙃 ... `[16, "$PRIMARY_NODE_VERSION"]` might not work either then. I haven't tried it.

So maybe a `startsWith` kind of check is necessary?

I'd rather not unpin a version and introduce unexpected version changes if we can avoid it.

JoshuaKGoldberg (changes requested):
Thanks for hopping onto this! I think I'd rather a more pin-oriented fix direction if we can make it?

.github/workflows/ci.yml
auvred: Yes, it looks like we can't use `env` context in matrix definition :(

> You can use the env context in any key in a workflow step except for the id and uses keys.

^ Ref: [GitHub docs](https://docs.github.com/en/actions/learn-github-actions/contexts#env-context)

---

> I'd rather not unpin a version and introduce unexpected version changes if we can avoid it.

On `main` branch (with pinned `PRIMARY_NODE_VERSION: '>=20.6.1'`) `actions/setup-node@v3` action downloads `v20.10.0`

```
Run actions/setup-node@v3
  with:
    node-version: >=20.6.1
    always-auth: false
    check-latest: false
    token: ***
  env:
    PRIMARY_NODE_VERSION: >=20.6.1
    NX_CLOUD_ACCESS_TOKEN: ***
Found in cache @ /opt/hostedtoolcache/node/20.10.0/x64
Environment details
  node: v20.10.0
  npm: 10.2.3
  yarn: 3.7.0
```

So I think it's safe to unpin it!

It was pinned in this commit https://github.com/typescript-eslint/typescript-eslint/pull/7567/commits/2a03e446e694be9721004f570689344516141659 only two days later when `v20.6.1` was released https://github.com/nodejs/node/releases/tag/v20.6.1. I guess it was the only way to force it to be higher than `v20.6.0` (For some reason website tests were failing without Node.js being pinned ([workflow runs](https://github.com/typescript-eslint/typescript-eslint/actions/runs/6136888430/job/16652678701)))

But now the latest `v20.x.x` version of Node.js is used, whether `PRIMARY_NODE_VERSION` is pinned or not

---

If we decide to leave it pinned and not touch it (I'd rather change it :smile:), then we should use something scary like

```
startsWith(env.PRIMARY_NODE_VERSION, format('>={0}', matrix.node-version))
```

JoshuaKGoldberg: Yeah that makes sense. My hesitation around unpinning is that I don't like "surprise" Node version bumps... ideally we'd have an exact version that gets updated over time the way Renovate bumps our versions of dependencies.

...but since we're already not doing that I think this is a straightforward bug fix. 😂

JoshuaKGoldberg (approved):
LGTM, thanks! 🚀

PR #8176, opened 2024-01-02
fix(eslint-plugin): [switch-exhaustiveness-check] fix new allowDefaultCaseForExhaustiveSwitch option

### LINKED ISSUE #8165: Bug: [switch-exhaustiveness-check] fails with option "allowDefaultCaseForExhaustiveSwitch": false

### Playground Link

https://typescript-eslint.io/play/#ts=5.3.3&fileType=.tsx&code=GYVwdgxgLglg9mABFApgZygCjALkWEAWwCMUAnASkQG8BYAKEUTQHcYoIALRbKuxpoggBDNCkQBGPGRRQQZJACYA3A0FDR4xdNnykAFlUCmAExTBhIADZQdchYgBsRpgF8G7%2BkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZAd0vzIAt6AHrwCGsZPkoA3RExTJafRGQDW6MAG1IiaNAD20SABpMUYfHi6OAEUQAzUfHwBhYckQAxfQFEho8VMQAZS4eXjV7eDcwAF8AXXAYkGigA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
function test(n: number) {
  switch (n) {
    case 1: return 2;
    case 2: return 4;
    default: return 6;
  }
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/switch-exhaustiveness-check": ["error", { "allowDefaultCaseForExhaustiveSwitch": false }]
  }
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

No error.

### Actual Result

The switch statement is exhaustive, so the default case is unnecessary. 5:5 - 5:23

### Additional Info

_No response_

Zamiell:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8165

## Overview

At first I thought this was a trivial fix to just check for `isUnion`, but then someone pointed out that it would fail for `string | number`, which is a union, but should actually have a default case.

Thus, I revert to creating a `doesTypeContainNonLiteralType` function.


ArnaudBarre:
I was just hit by the issue when enabling this new option that looks good!
Thanks for already fixing it, but I'm a bit afraid the `isUnion` check can lead to issues, for example does default will be allow with: `number | string` switch or with `"A" | "B" | string`? Ideally what you want is to be able to detect is the switch type is finite or not

Zamiell:
@ArnaudBarre You are right, I will add the following "allow" test to address this:
```ts
    // switch with default clause on primitive +
    // "allowDefaultCaseForExhaustiveSwitch" option
    {
      code: `
declare const value: number | string;
switch (value) {
  case 0:
    return 0;
  case 1:
    return 1;
  default:
    return -1;
}
      `,
      options: [
        {
          allowDefaultCaseForExhaustiveSwitch: false,
          requireDefaultForNonUnion: false,
        },
      ],
    },
```

Zamiell:
In order to address this, I added a new prop to the metadata called `containsNonLiteralType`, which uses the function called `doesTypeContainNonLiteralType`. Maybe there is a more performant way to do it?

JoshuaKGoldberg (approved):
Thanks!

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -17,12 +17,14 @@ interface SwitchMetadata {
...
   readonly isUnion: boolean;
+  readonly containsNonLiteralType: boolean;
 }
 
 type Options = [
   {
     /**
-     * If `true`, allow `default` cases on switch statements with exhaustive cases.
+     * If `true`, allow `default` cases on switch statements with exhaustive
+     * cases.
```
JoshuaKGoldberg, line 27: (this is fine I guess, no strong preference either way)

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -138,9 +144,29 @@ export default createRule<Options, MessageIds>({
...
         isUnion: true,
+        containsNonLiteralType,
       };
     }
 
+    /**
+     * For example:
+     *
+     * - `"foo" | "bar"` is a type with all literal types.
+     * - `"foo" | number` is a type that contain non-literal types.
```
JoshuaKGoldberg, line 155:
```suggestion
     * - `"foo" | number` is a type that contains non-literal types.
```

JoshuaKGoldberg (approved):

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
Zamiell: the comment was "illegal" such that it passed the length of the ruler

Zamiell: if you are interested in linting (and autofixing!) such comments, i wrote this: https://github.com/IsaacScript/isaacscript/blob/main/packages/eslint-plugin-isaacscript/docs/rules/format-jsdoc-comments.md

JoshuaKGoldberg: Nifty, might be good as a rule in eslint-plugin-jsdoc?

If you want to file an issue suggesting using it here I'd be interested in that too 😄. Need to think on it a bit (are there edge cases?) but in general I like the idea!

PR #8188, opened 2024-01-04
chore(typescript-estree): fix lint / CI

Zamiell:
Fixes: #8164

this just fixes CI on main, looks like it has been broken for several weeks?


auvred:
fixes https://github.com/typescript-eslint/typescript-eslint/issues/8164

Zamiell:
thanks, I edited that into the OP

JoshuaKGoldberg (approved):
Yeah I think this is the right fix. The nodes were previously being considered duplicate type constituents because they used to be `type` aliases. But then TypeScript changed them to interfaces so they can be merged: https://github.com/microsoft/TypeScript/issues/56477

PR #8190, opened 2024-01-04
docs: add `import/no-unresolved` to perf troubleshooting docs

Zamiell:
## PR Checklist

- [ ] Addresses an existing open issue: n/a

## Overview

I added 2 rules to the page, which are the following:

- `import/no-unresolved`
- `import/unambiguous`

I believe that these are correctly handled by the TypeScript compiler, correct me if I am wrong.

### DIFF
--- a/docs/linting/troubleshooting/Performance.mdx
+++ b/docs/linting/troubleshooting/Performance.mdx
@@ -81,7 +81,6 @@ There are many rules that do single file static analysis, but we provide the fol
 
 We recommend you do not use the following rules, as TypeScript provides the same checks as part of standard type checking:
 
-- `import/extensions`
 - `import/named`
 - `import/namespace`
 - `import/default`
--- a/docs/linting/troubleshooting/Performance.mdx
+++ b/docs/linting/troubleshooting/Performance.mdx
@@ -86,7 +86,6 @@ We recommend you do not use the following rules, as TypeScript provides the same
 - `import/default`
 - `import/no-named-as-default-member`
 - `import/no-unresolved`
-- `import/unambiguous`
 
 The following rules do not have equivalent checks in TypeScript, so we recommend that you only run them at CI/push time, to lessen the local performance burden.
 
--- a/docs/linting/troubleshooting/Performance.mdx
+++ b/docs/linting/troubleshooting/Performance.mdx
@@ -81,10 +81,13 @@ There are many rules that do single file static analysis, but we provide the fol
 
 We recommend you do not use the following rules, as TypeScript provides the same checks as part of standard type checking:
 
+- `import/extensions`
 - `import/named`
 - `import/namespace`
 - `import/default`
 - `import/no-named-as-default-member`
+- `import/no-unresolved`
+- `import/unambiguous`
 
 The following rules do not have equivalent checks in TypeScript, so we recommend that you only run them at CI/push time, to lessen the local performance burden.
 
--- a/docs/linting/troubleshooting/Performance.mdx
+++ b/docs/linting/troubleshooting/Performance.mdx
@@ -101,7 +101,7 @@ The following rules do not have equivalent checks in TypeScript, so we recommend
 
 If you want to enforce file extensions are always used and you're **NOT** using `moduleResolution` `node16` or `nodenext`, then there's not really a good alternative for you, and you should continue using the `import/extensions` lint rule.
 
-If you want to enforce file extensions are always used and you **ARE** using `moduleResolution` `node16` or `nodenext`, then you don't need to use the lint rule at all because TypeScript will automatically enforce that you include extensions! (However, the lint rule is still extremely useful as a one-time project-wide autofixer if you are switching from no file extensions --> file extensions, or vice versa.)
+If you want to enforce file extensions are always used and you **ARE** using `moduleResolution` `node16` or `nodenext`, then you don't need to use the lint rule at all because TypeScript will automatically enforce that you include extensions! (However, the lint rule is still extremely useful as a one-time project-wide fixer if you are switching from no file extensions --> file extensions, or vice versa.)
 
 #### Enforcing extensions are not used
 
--- a/docs/linting/troubleshooting/Performance.mdx
+++ b/docs/linting/troubleshooting/Performance.mdx
@@ -85,7 +85,7 @@ We recommend you do not use the following rules, as TypeScript provides the same
 - `import/namespace`
 - `import/default`
 - `import/no-named-as-default-member`
-- `import/no-unresolved`
+- `import/no-unresolved` (as long as you are using [`import` over `require`](https://typescript-eslint.io/rules/no-var-requires/))
 
 The following rules do not have equivalent checks in TypeScript, so we recommend that you only run them at CI/push time, to lessen the local performance burden.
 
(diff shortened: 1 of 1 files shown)


docs/linting/troubleshooting/Performance.mdx
JoshuaKGoldberg, line 1: [Build] This has build failures that will need to be addressed. From https://github.com/typescript-eslint/typescript-eslint/actions/runs/7415985668/job/20180149405?pr=8190:

```plaintext
./docs/linting/troubleshooting/Performance.mdx:104:321   - Unknown word (autofixer)  -- one-time project-wide autofixer if you are switching
	 Suggestions: [autofixes, autofixers, autofix, atomizer, autofill]
CSpell: Files checked: 13[8](https://github.com/typescript-eslint/typescript-eslint/actions/runs/7415985668/job/20180149405?pr=8190#step:4:9)5, Issues found: 1 in 1 files
```

Looks like the cspell dictionary could be informed of the word `autofixer`? Alternately, I think we use the word `fixer` instead most of the time, as that's what ESLint uses. So probably `fixer` is the better word here.

docs/linting/troubleshooting/Performance.mdx
```diff
@@ -85,6 +85,8 @@ We recommend you do not use the following rules, as TypeScript provides the same
 - `import/namespace`
 - `import/default`
 - `import/no-named-as-default-member`
+- `import/no-unresolved`
+- `import/unambiguous`
```
JoshuaKGoldberg, line 89: [Question] I don't think TypeScript always enforces this? There's no error reported locally when I write a new `.ts` file that doesn't `import` or `export`.

Unused code detection (e.g. [knip](https://knip.dev)) can, but that's separate.

docs/linting/troubleshooting/Performance.mdx
```diff
@@ -85,6 +85,8 @@ We recommend you do not use the following rules, as TypeScript provides the same
 - `import/namespace`
 - `import/default`
 - `import/no-named-as-default-member`
+- `import/no-unresolved`
```
JoshuaKGoldberg, line 88: [Question] I don't think TypeScript enforces this? I added a `require('./nope')` in a new `.ts` file and it didn't complain.

We do have the [`no-var-requires` rule](https://typescript-eslint.io/rules/no-var-requires) though, so maybe it can be noted that this rule is only unnecessary if you ban `require()`?

docs/linting/troubleshooting/Performance.mdx
```diff
@@ -99,7 +101,7 @@ The following rules do not have equivalent checks in TypeScript, so we recommend
 
 If you want to enforce file extensions are always used and you're **NOT** using `moduleResolution` `node16` or `nodenext`, then there's not really a good alternative for you, and you should continue using the `import/extensions` lint rule.
 
-If you want to enforce file extensions are always used and you **ARE** using `moduleResolution` `node16` or `nodenext`, then you don't need to use the lint rule at all because TypeScript will automatically enforce that you include extensions!
+If you want to enforce file extensions are always used and you **ARE** using `moduleResolution` `node16` or `nodenext`, then you don't need to use the lint rule at all because TypeScript will automatically enforce that you include extensions! (However, the lint rule is still extremely useful as a one-time project-wide autofixer if you are switching from no file extensions --> file extensions, or vice versa.)
```
JoshuaKGoldberg, line 104: [Docs] I don't think this note is necessary for this docs page. The page is intended to help people troubleshoot performance issues in day-to-day linting. Notes on using rules for other operations -especially ones that are ok to be slower- doesn't really fit in the narrative.

JoshuaKGoldberg (changes requested):
Thanks for sending! I don't _think_ any of these changes need to land, but maybe I'm wrong?

Also: please always write an unambiguous title for PRs. That helps both reviewers understand quickly what's going on, and helps the changelog be more informative.

docs/linting/troubleshooting/Performance.mdx
Zamiell: > I don't think TypeScript enforces this? I added a require('./nope') in a new .ts file and it didn't complain.

oh, then maybe it only enforces it for `import` statements and not `require` statements?
with that said, i doubt anyone reading these docs are using require statements in combination with typescript?

docs/linting/troubleshooting/Performance.mdx
Zamiell: > I don't think TypeScript always enforces this? There's no error reported locally when I write a new .ts file that doesn't import or export.

i am not an expert but i believe that `the [unambiguous](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/unambiguous.md) rule is unnecessary when creating a typescript program.
maybe someone else on the team can weigh in?

bradzacher: No - it is actually a necessity!

Without any import/exports TS does treat a file as a script - and in doing that the rules for global augmentation and such change. So including an empty export is still necessary in TS :)

docs/linting/troubleshooting/Performance.mdx
bradzacher: Correct - a plain require call is completely unchecked. TS only checks a require if it's in the form `import foo = require('foo');`

JoshuaKGoldberg: > i doubt anyone reading these docs is using require statements in combination with TypeScript

If that were the case, we wouldn't need the `no-var-requires` rule. But it still exists because it's a valid way to do things in TypeScript (with a caveat around the `any` type).

Zamiell: right, so maybe in the docs it should say:

```
- `import/no-unresolved` (as long as you are using `import` over `require`)
```

?

docs/linting/troubleshooting/Performance.mdx
Zamiell: hmm, ok.
see the other discussion.
if you accept that the other section should be listed as 
```
- `import/no-unresolved` (as long as you are using `import` over `require`)
```
then maybe this section should be:
```
- `import/unambiguous` (as long as you are using a dead code finder such as `knip` or `ts-prune`)
```
?

docs/linting/troubleshooting/Performance.mdx
JoshuaKGoldberg: 👍 sounds good to me. Linking to `no-var-requires` would be helpful too.

docs/linting/troubleshooting/Performance.mdx
JoshuaKGoldberg: Another use case for ambiguous scripts would be one that intentionally wants to keep the scripts as is.

Looking at the source for `import/unamiguous`, it's not a performance-critical rule: 
* Its `create`'s logic is mostly a call to an `isModule` method: https://github.com/import-js/eslint-plugin-import/blob/7a21f7e10f18c04473faadca94928af6b8e28009/src/rules/unambiguous.js#L28
* That `isModule` is a loop through nodes: https://github.com/import-js/eslint-plugin-import/blob/7a21f7e10f18c04473faadca94928af6b8e28009/utils/unambiguous.js#L29

Between TypeScript _not performing this check during standard linting_, there being legitimate use cases for code that violates the rule, and the rule being very cheap to run, there's not a need to mention it here.

bradzacher (approved):

PR #8136, opened 2023-12-26
fix(typescript-estree): only create project service from env setting if project is enabled

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7348; fixes #8131
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

I spent some time trying to investigate #8131's unit test failures. I made a bit of progress and then realized... wait, why is the project service enabled for either of these rules? We don't normally enable `parserOptions.project` for un-typed lint rules. Therefore, I don't think we'd want the presence of the environment variable to enable the project service unless `options.project` was also set.

Issue #7890, opened 2023-11-08
Repo: Integration tests failing on main from latest ESLint


JoshuaKGoldberg:
### Suggestion

See 75c128856b1ce05a4fec799bfa6de03b3dab03d0 > https://github.com/typescript-eslint/typescript-eslint/actions/runs/6774184010/job/18410868606 > _Run integration tests_ as an example:

```plaintext
 FAIL  tests/vue-sfc.test.ts (42.521 s)
  ● vue-sfc › should lint successfully
```

```plaintext
 FAIL  tests/recommended-does-not-require-program.test.ts (10.326 s)
  ● recommended-does-not-require-program › should lint successfully
```

```plaintext
Snapshot Summary
 › 2 snapshots failed from 2 test suites. Inspect your code changes or re-run jest with `-u` to update them.
Test Suites: 2 failed, 4 passed, 6 total
Tests:       2 failed, 4 passed, 6 total
Snapshots:   2 failed, 4 passed, 6 total
Time:        102.14 s
```

bradzacher:
The joys of testing the json output instead of the CLI output - we get little invisible changes like this!

Also we used a formatting rule cos it was easier to setup a simple repro for heh.

Issue #6761, opened 2023-03-24
Docs: Update the 'Major Release Steps' docs for what we did to release v6


JoshuaKGoldberg:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

We've done a few things to prep v6 that aren't in the docs yet. They should be documented.

At the very least:

* #6759
* #6760
* Writing & sharing out a blog post
* Creating a Discord channel

### Affected URL(s)

https://typescript-eslint.io/maintenance/releases#major-release-steps

JoshuaKGoldberg:
I think we'll need to update the list of expected status checks for the PRs whenever we update a Node version, too.

![Screenshot of several Run Unit Tests (18, ...) status checks still waiting to be reported in GitHub](https://github.com/typescript-eslint/typescript-eslint/assets/3335181/e861414c-e466-42a8-b4f9-231e8e191ed6)

bradzacher:
> I think we'll need to update the list of expected status checks for the PRs whenever we update a Node version, too.

I just did that yeah - that's not specifically a major release thing though - that's just a CI change step - but we happened to change the min AND max node versions in this major which meant that all the old required steps are now out-of-date.

I just updated it all.
Steps are:
1) https://github.com/typescript-eslint/typescript-eslint/settings/branches
2) edit the `main` branch protection
3) remove the now obsolete steps from "Status checks that are required."
4) add the new steps to "Status checks that are required."

NOTE: this sucks, the UI sucks, and it's a pain in the ass to do, sadly.

JamesHenry:
TODO for this is updating the text to reference `nx release` rather than `lerna` and tweak the instructions accordingly

PR #8194, opened 2024-01-05
chore: use nx release

JamesHenry:
For right now, a pinned header at the top of the changelog file is not supported, so I have removed this in every case to avoid confusion once we do the first release after this goes in:

```md
# Change Log

All notable changes to this project will be documented in this file.
See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
```

I'll add support for a pinned header like that into `nx release` after this and we can bring it back if we want


packages/ast-spec/project.json
```diff
@@ -16,7 +16,6 @@
       "executor": "@nx/eslint:lint",
       "outputs": ["{options.outputFile}"],
       "options": {
-        "lintFilePatterns": ["packages/ast-spec/**/*.{mts,cts,ts,tsx}"],
```
JamesHenry, line 19: The patterns are no longer required, the source files of the project are all eligible to be linted

JamesHenry:
@typescript-eslint/triage-team any idea why I am getting these integration-test failures?

CHANGELOG.md
```diff
@@ -1,8 +1,3 @@
-# Change Log
```
JoshuaKGoldberg, line 1: While we're here, can we change it to `# Changelog`? I don't know of any other project that uses two words 😄

package.json
```diff
@@ -23,13 +23,13 @@
     "url": "https://github.com/typescript-eslint/typescript-eslint/issues"
   },
   "scripts": {
-    "build": "npx nx run-many --target=build --parallel --exclude website --exclude website-eslint",
+    "build": "npx nx run-many -t build --parallel --exclude website --exclude website-eslint",
```
JoshuaKGoldberg, line 26: > ```diff
> - --target
> + -t
> ```

I have a very mild preference for the explicit syntax, to make it more understandable to not-frequent-Nx users like me. Not a blocker at all from me though.

package.json
```diff
@@ -74,14 +75,14 @@
     "@types/marked": "^5.0.0",
     "@types/natural-compare": "^1.4.1",
     "@types/ncp": "^2.0.5",
-    "@types/node": "^20.0.0",
+    "@types/node": "18.16.9",
```
JoshuaKGoldberg, line 78: Is this downgrade necessary? Merge conflict?

tools/release/release.ts
```diff
@@ -0,0 +1,61 @@
+import {
+  releaseChangelog,
+  releasePublish,
+  releaseVersion,
+} from 'nx/src/command-line/release';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import yargs from 'yargs';
```
JoshuaKGoldberg, line 7: [Question] Is there a reason to not add to `package.json` dependencies?

tools/release/release.ts
```diff
@@ -0,0 +1,61 @@
+import {
+  releaseChangelog,
+  releasePublish,
+  releaseVersion,
+} from 'nx/src/command-line/release';
```
JoshuaKGoldberg, line 5: [Product Feedback] Importing from a `src/` sub-path feels off to me. It'd be nice to have a more official/not-feeling-like-I'm-importing-from-internals way.

tools/release/release.ts
```diff
@@ -0,0 +1,61 @@
...
+      .version(false)
+      .option('version', {
+        description:
+          'Explicit version specifier to use, if overriding conventional commits',
+        type: 'string',
+      })
+      .option('dryRun', {
+        alias: 'd',
+        description:
+          'Whether or not to perform a dry-run of the release process, defaults to true',
```
JoshuaKGoldberg, line 21: [Proofreading] Super nitpicky I know, but the _"or not"_ here and below is redundant.

```suggestion
          'Whether to perform a dry-run of the release process, defaults to true',
```

package.json
JamesHenry: I just think it’s much cleaner, happy to revert though

package.json
JamesHenry: Probably an automated migration that was too eager, I’ll revert

CHANGELOG.md
JamesHenry: Ha yeah true

tools/release/release.ts
```diff
@@ -0,0 +1,61 @@
+import {
+  releaseChangelog,
+  releasePublish,
+  releaseVersion,
+} from 'nx/src/command-line/release';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import yargs from 'yargs';
+
+void (async () => {
+  try {
```
JoshuaKGoldberg, line 10: [Suggestion] We can avoid these wrappings by changing the file extension to `.mts` and adding `.mts` to `tsconfig.eslint.json`.

```suggestion
```

tools/release/release.ts
```diff
@@ -0,0 +1,61 @@
...
+    const options = await yargs
+      .version(false)
```
JoshuaKGoldberg, line 12: [Bug] On my machine's installed version at least, `yargs` needs to be provided with the argv now:

```suggestion
    const options = await yargs(process.argv.slice(2))
      .version(false)
```

.github/workflows/ci.yml
```diff
@@ -279,19 +279,25 @@ jobs:
...
-      # Fetch all history for all tags and branches in this job because lerna needs it
-      - run: |
-          git fetch --prune --unshallow
+      - name: Figure out and apply the next canary version
+        run: npx nx run repo-tools:apply-canary-version
 
-      - name: Publish all packages to npm
-        run: npx lerna publish --loglevel=verbose --canary --exact --force-publish --yes
+      - name: Publish all packages to npm with the canary tag
+        # NOTE: this needs to be npx, rather than yarn, to make sure the authenticated npm registry is used
```
JoshuaKGoldberg, line 298: [Question] It feels like we should be able to get Yarn to work with npm auth, no? `.yarnrc.yml` or some such?

Not a blocker IMO, just a curiosity.

packages/repo-tools/src/apply-canary-version.mts
JoshuaKGoldberg, line 1: [Product Feedback] This feels like something that would be nice to have something do for us 😄. IMO common versioning strategies such as alpha/canary releases make sense to standardize ... somewhere. Maybe in Nx core, maybe in an Nx package, maybe in a community package? Somewhere.

tools/release/changelog-renderer.js
```diff
@@ -0,0 +1,26 @@
+const { default: defaultChangelogRenderer } = require('nx/changelog-renderer');
```
JoshuaKGoldberg, line 1: [Question] Is there any reason to not put the scripts from `tools/` into `packages/repo-tools/src` instead? Having two folders with tools is a little confusing IMO. I'd personally rather put everything into this internal package.

This `const { default: ... }` pattern to me in particular is something I'd really like to avoid if possible. Even if it means having to build the `repo-tools` package before running certain commands.

package.json
JoshuaKGoldberg, line 1: [Docs] I did a search for `lerna` in the repo and found mentions in https://github.com/typescript-eslint/typescript-eslint/blob/7c79f54b9577ba99fbfcd925d0922fd4472d7ec2/docs/maintenance/Releases.mdx. Could you update those too?

If it's not a straightforward fix then I think it'd be reasonable to have a "todo" notice pending #6761.

(commenting here as I can't on `Release.mdx`)

JoshuaKGoldberg (changes requested):
Awesome to see the Lerna->Nx migration come to a head! 🚀 

Requesting changes on moving things to `repo-tools` and updating the last mention of Lerna from the docs. But I can aquiese if you don't think those should be blocking. 😄

tools/release/changelog-renderer.js
JamesHenry: Yes there is currently - npm publish hard errors when it is triggered from within a nested “npm workspace” which is what that package is . I had it there first because I forgot about that npm publish behaviour, I moved it up to tools out of necessity.

Having said that I seem to have the opposite intuition to you - having things which help manage the whole workspace live in a nested project is confusing to me /shrug

package.json
JamesHenry: Yes I’ll add a TODO

packages/repo-tools/src/apply-canary-version.mts
JamesHenry: Yes I agree just haven’t gotten to it yet

.github/workflows/ci.yml
JamesHenry: npx seems to be consistently used in the workflow already anyway, this comment was carried over from another repo where I had that experience but it was a while ago so fuzzy on the details

tools/release/release.ts
JamesHenry: How strange… I tested it multiple times on my machine without issue, and I also used it like this in other repos

tools/release/release.ts
JamesHenry: I don’t disagree! Hopefully it’s something we can tweak later in the year

JamesHenry:
@JoshuaKGoldberg any thoughts on the failures?

tools/release/changelog-renderer.js
JoshuaKGoldberg: Heh yeah now that you mention it... it is kind of weird to put non-package things in a package... and `tools/` is less overhead than a full package. For me 90% of happiness here comes from unifying. So if we go to just everything in `tools/` that works for me too!

tools/release/release.ts
JoshuaKGoldberg: For this and other product feedback points, is there a good link we can put in a todo comment?

JoshuaKGoldberg:
Oh right sorry - it looks like #7890 -> #7891... which is unfortunate because the `nx-release` branch is up to date. Odd. Maybe a caching issue?

package.json
```diff
@@ -74,14 +75,14 @@
...
     "@types/semver": "^7.5.0",
     "@types/tmp": "^0.2.3",
     "console-fail-test": "^0.2.3",
     "cross-fetch": "^4.0.0",
     "cspell": "^7.0.0",
     "downlevel-dts": ">=0.11.0",
-    "eslint": "^8.47.0",
+    "eslint": "8.48.0",
```
auvred, line 85:
```suggestion
    "eslint": "^8.48.0",
```

Integration tests're faililin because of pinned `eslint` version:
- Stylistic rules were deprecated in `v8.53.0` ([blog post](https://eslint.org/blog/2023/11/eslint-v8.53.0-released/))
- ESLint version for integration tests is taken from the root package.json https://github.com/typescript-eslint/typescript-eslint/blob/3d729303dd5cfd18ccb823e96f0e8761fda6adab/packages/integration-tests/tools/integration-test-base.ts#L27
- Since it is pinned, yarn resolves ESLint as `8.48.0` rather than `>=8.53.0`
- Tests fail because stylistic rules aren't marked as deprecated
```
FAIL  tests/vue-sfc.test.ts (21.592 s)
  ● vue-sfc › should lint successfully
    expect(received).toMatchSnapshot()
    Snapshot name: `vue-sfc should lint successfully 1`
    - Snapshot  - 18
    + Received  +  3
    @@ -53,[16](https://github.com/typescript-eslint/typescript-eslint/actions/runs/7426374087/job/20210070340?pr=8194#step:5:17) +53,11 @@
        },
      });
      </script>
      ",
          "suppressedMessages": [],
    -     "usedDeprecatedRules": [
    -       {
    -         "replacedBy": [],
    -         "ruleId": "semi-spacing",
    -       },
    -     ],
    +     "usedDeprecatedRules": [],
          "warningCount": 0,
        },
        {
          "errorCount": 1,
          "fatalErrorCount": 1,
    @@ -93,16 +88,11 @@
        }
      }
```

auvred (changes requested):

package.json
JamesHenry: Thanks @auvred!

tools/release/release.ts
JamesHenry: I've tracked it on our internal Linear, we don't tend to duplicate things like this into the OSS issue tracker. Feel free to open issues there any time or just ping me, either is fine

tools/release/release.ts
JamesHenry: I did initially and the lint errors went crazy to I punted on that 
image

Feel free to adjust in a small follow up if you know what the fix is

tools/release/release.ts
JamesHenry: Should be handled by the explicit yargs version being specified in the package.json now

tools/release/changelog-renderer.js
JamesHenry: Cool I'll revisit the thread we had on it in discord and do a proper reconciliation in a follow up

JamesHenry:
@JoshuaKGoldberg @auvred updated per the PR feedback and now everything is passing except for the TSServer tests.

I reran it in case it was flakiness, but the failure seems to be consistent. Any ideas please?

https://github.com/typescript-eslint/typescript-eslint/actions/runs/7431162445/job/20221866549?pr=8194

auvred:
Oh, this is a known issue, it's not related to this PR. It should be fixed in https://github.com/typescript-eslint/typescript-eslint/issues/8131 -> https://github.com/typescript-eslint/typescript-eslint/pull/8136

JamesHenry:
Ah cool thanks, pressing ahead with this then while I still have time to focus on it

tools/release/release.ts
JoshuaKGoldberg: It's `tsconfig.eslint.json` explicitly including only `.ts` files in this folder, not `.mts`.

tools/release/release.ts
JoshuaKGoldberg: 👍 https://github.com/nrwl/nx/issues/21029

PR #8216, opened 2024-01-08
feat(eslint-plugin): add rule prefer-find

### LINKED ISSUE #6886: Rule proposal: Preferring .find(...) over .filter(...)[0]

### Description

I've seen occasional code authors write code with a `.filter` returning element `[0]`, perhaps not knowing that `.find` does the same thing. Perhaps we could write a rule that flags for this, and uses the type system to know whether the collection being `.filter`ed is an array/array-like?

### Fail Cases

```typescript
const values = ['a', 'b ', 'c'];

const result = values.filter(value => value.includes(' '))[0];
```

### Pass Cases

```typescript
const values = ['a', 'b ', 'c'];

const result = values.find(value => value.includes(' '));
```

### Additional Info

_No response_

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6886 

## Overview

Adds a rule that prefers `arr.find(...)` over `arr.filter(...)[0]`. Also adds the corresponding autofix as suggestions.

### DIFF
--- a/packages/eslint-plugin/docs/rules/prefer-find.md
+++ b/packages/eslint-plugin/docs/rules/prefer-find.md
@@ -1,0 +1,29 @@
+---
+description: 'Enforce the use of Array.prototype.find() over Array.prototype.filter() followed by [0] when looking for a single result.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/prefer-find** for documentation.
+
+When searching for the first item in an array matching a condition, it may be tempting to use code like `arr.filter(x => x > 0)[0]`. However, it is simpler to use [Array.prototype.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead, `arr.find(x => x > 0)`, which also returns the first entry matching a condition.
+
+Beware that the two approaches are not quite identical, however! `.find()` will only execute the callback on array elements until it finds a match, whereas `.filter()` executes the callback for all array elements. Therefore, when fixing errors from this rule, be sure that your `.filter()` callbacks do not have side effects.
+
+<!--tabs-->
+
+### ❌ Incorrect
+
+```ts
+[1, 2, 3].filter(x => x > 1)[0];
+```
+
+### ✅ Correct
+
+```ts
+[1, 2, 3].find(x => x > 1);
+```
+
+## When Not To Use It
+
+If you don't mind `.filter(...)[0]`, you can avoid this rule.
--- a/packages/eslint-plugin/docs/rules/prefer-find.md
+++ b/packages/eslint-plugin/docs/rules/prefer-find.md
@@ -6,9 +6,17 @@ description: 'Enforce the use of Array.prototype.find() over Array.prototype.fil
 >
 > See **https://typescript-eslint.io/rules/prefer-find** for documentation.
 
-When searching for the first item in an array matching a condition, it may be tempting to use code like `arr.filter(x => x > 0)[0]`. However, it is simpler to use [Array.prototype.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead, `arr.find(x => x > 0)`, which also returns the first entry matching a condition.
+When searching for the first item in an array matching a condition, it may be tempting to use code like `arr.filter(x => x > 0)[0]`.
+However, it is simpler to use [Array.prototype.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead, `arr.find(x => x > 0)`, which also returns the first entry matching a condition.
+Because the `.find()` only needs to execute the callback until it finds a match, it's also more efficient.
 
-Beware that the two approaches are not quite identical, however! `.find()` will only execute the callback on array elements until it finds a match, whereas `.filter()` executes the callback for all array elements. Therefore, when fixing errors from this rule, be sure that your `.filter()` callbacks do not have side effects.
+:::info
+
+Beware the the difference in short-circuiting behavior between the approaches.
+`.find()` will only execute the callback on array elements until it finds a match, whereas `.filter()` executes the callback for all array elements.
+Therefore, when fixing errors from this rule, be sure that your `.filter()` callbacks do not have side effects.
+
+:::
 
 <!--tabs-->
 
@@ -16,6 +24,8 @@ Beware that the two approaches are not quite identical, however! `.find()` will 
 
 ```ts
 [1, 2, 3].filter(x => x > 1)[0];
+
+[1, 2, 3].filter(x => x > 1).at(0);
 ```
 
 ### ✅ Correct
@@ -26,4 +36,4 @@ Beware that the two approaches are not quite identical, however! `.find()` will 
 
 ## When Not To Use It
 
-If you don't mind `.filter(...)[0]`, you can avoid this rule.
+If you intentionally use patterns like `.filter(callback)[0]` to execute side effects in `callback` on all array elements, you will want to avoid this rule.
--- a/packages/eslint-plugin/docs/rules/prefer-find.md
+++ b/packages/eslint-plugin/docs/rules/prefer-find.md
@@ -12,7 +12,7 @@ Because the `.find()` only needs to execute the callback until it finds a match,
 
 :::info
 
-Beware the the difference in short-circuiting behavior between the approaches.
+Beware the difference in short-circuiting behavior between the approaches.
 `.find()` will only execute the callback on array elements until it finds a match, whereas `.filter()` executes the callback for all array elements.
 Therefore, when fixing errors from this rule, be sure that your `.filter()` callbacks do not have side effects.
 
--- a/packages/eslint-plugin/src/rules/prefer-find.ts
+++ b/packages/eslint-plugin/src/rules/prefer-find.ts
@@ -1,0 +1,110 @@
+import type { TSESLint } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+import * as tsutils from 'ts-api-utils';
+
+import { createRule, getParserServices } from '../util';
+
+export default createRule({
+  name: 'prefer-find',
+  meta: {
+    docs: {
+      description:
+        'Enforce the use of Array.prototype.find() over Array.prototype.filter() followed by [0] when looking for a single result',
+      requiresTypeChecking: true,
+      recommended: 'stylistic',
+    },
+    fixable: 'code',
+    messages: {
+      preferFind: 'Use .filter(...)[0] instead of .find(...)',
+      preferFindFix: 'Use .filter(...)[0] instead of .find(...)',
+    },
+    schema: [],
+    type: 'suggestion',
+    hasSuggestions: true,
+  },
+
+  defaultOptions: [],
+
+  create(context) {
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    return {
+      MemberExpression(node): void {
+        // Check if it looks like <<stuff>>[0] or <<stuff>>['0'], but not <<stuff>>?.[0]
+        if (
+          node.computed &&
+          !node.optional &&
+          node.property.type === AST_NODE_TYPES.Literal &&
+          (node.property.value === 0 ||
+            node.property.value === '0' ||
+            node.property.value === 0n)
+        ) {
+          // Check if it looks like <<stuff>>(...)[0], but not <<stuff>>?.(...)[0]
+          if (
+            node.object.type === AST_NODE_TYPES.CallExpression &&
+            !node.object.optional
+          ) {
+            const objectCallee = node.object.callee;
+            // Check if it looks like <<stuff>>.filter(...)[0] or <<stuff>>['filter'](...)[0],
+            // or the optional chaining variants.
+            if (objectCallee.type === AST_NODE_TYPES.MemberExpression) {
+              const isBracketSyntaxForFilter = objectCallee.computed;
+              if (
+                isBracketSyntaxForFilter
+                  ? objectCallee.property.type === AST_NODE_TYPES.Literal &&
+                    objectCallee.property.value === 'filter'
+                  : objectCallee.property.name === 'filter'
+              ) {
+                const isOptionalAccessOfFilter = objectCallee.optional;
+
+                const filteredObjectType = checker.getTypeAtLocation(
+                  services.esTreeNodeToTSNodeMap.get(objectCallee.object),
+                );
+
+                // We can now report if the object is an array
+                // or if it's an optional chain on a nullable array.
+                if (
+                  checker.isArrayType(filteredObjectType) ||
+                  (isOptionalAccessOfFilter &&
+                    tsutils
+                      .unionTypeParts(filteredObjectType)
+                      .every(
+                        unionPart =>
+                          checker.isArrayType(unionPart) ||
+                          tsutils.isIntrinsicNullType(unionPart) ||
+                          tsutils.isIntrinsicUndefinedType(unionPart),
+                      ))
+                ) {
+                  context.report({
+                    node,
+                    messageId: 'preferFind',
+                    suggest: [
+                      {
+                        messageId: 'preferFindFix',
+                        fix(fixer): TSESLint.RuleFix[] {
+                          return [
+                            // Replace .filter with .find
+                            fixer.replaceText(
+                              objectCallee.property,
+                              isBracketSyntaxForFilter ? '"find"' : 'find',
+                            ),
+                            // Get rid of the [0]
+                            fixer.removeRange([
+                              node.object.range[1],
+                              node.range[1],
+                            ]),
+                          ];
+                        },
+                      },
+                    ],
+                  });
+                }
+              }
+            }
+          }
+        }
+      },
+    };
+  },
+});
--- a/packages/eslint-plugin/src/rules/prefer-find.ts
+++ b/packages/eslint-plugin/src/rules/prefer-find.ts
@@ -3,6 +3,7 @@ import { AST_NODE_TYPES } from '@typescript-eslint/utils';
 import { getScope, getSourceCode } from '@typescript-eslint/utils/eslint-utils';
 import type { RuleFix, SourceCode } from '@typescript-eslint/utils/ts-eslint';
 import * as tsutils from 'ts-api-utils';
+import type { Type } from 'typescript';
 
 import {
   createRule,
@@ -70,7 +71,6 @@ export default createRule({
               : callee.property.name === 'filter'
           ) {
             const filterNode = callee.property;
-            const isOptionalAccessOfFilter = callee.optional;
 
             const filteredObjectType = getConstrainedTypeAtLocation(
               services,
@@ -79,20 +79,7 @@ export default createRule({
 
             // As long as the object is an array or an optional chain on a
             // nullable array, this is an Array.prototype.filter expression.
-            if (
-              checker.isArrayType(filteredObjectType) ||
-              checker.isTupleType(filteredObjectType) ||
-              (isOptionalAccessOfFilter &&
-                tsutils
-                  .unionTypeParts(filteredObjectType)
-                  .every(
-                    unionPart =>
-                      checker.isArrayType(unionPart) ||
-                      checker.isTupleType(unionPart) ||
-                      tsutils.isIntrinsicNullType(unionPart) ||
-                      tsutils.isIntrinsicUndefinedType(unionPart),
-                  ))
-            ) {
+            if (isArrayish(filteredObjectType)) {
               return {
                 isBracketSyntaxForFilter,
                 filterNode,
@@ -103,6 +90,30 @@ export default createRule({
       }
 
       return undefined;
+    }
+
+    /**
+     * Tells whether the type is a possibly nullable array/tuple or union thereof.
+     */
+    function isArrayish(type: Type): boolean {
+      let isAtLeastOneArrayishComponent = false;
+      for (const unionPart of tsutils.unionTypeParts(type)) {
+        if (
+          tsutils.isIntrinsicNullType(unionPart) ||
+          tsutils.isIntrinsicUndefinedType(unionPart)
+        ) {
+          // ignore
+        } else if (
+          checker.isArrayType(unionPart) ||
+          checker.isTupleType(unionPart)
+        ) {
+          isAtLeastOneArrayishComponent = true;
+        } else {
+          return false;
+        }
+      }
+
+      return isAtLeastOneArrayishComponent;
     }
 
     function getObjectIfArrayAtExpression(
--- a/packages/eslint-plugin/src/rules/prefer-find.ts
+++ b/packages/eslint-plugin/src/rules/prefer-find.ts
@@ -138,6 +138,10 @@ export default createRule({
       return undefined;
     }
 
+    /**
+     * Implements the algorithm for array indexing by `.at()` method.
+     * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#parameters
+     */
     function isTreatedAsZeroByArrayAt(value: unknown): boolean {
       const asNumber = Number(value);
 
@@ -160,6 +164,10 @@ export default createRule({
       );
     }
 
+    /**
+     * Implements the algorithm for array indexing by member operator.
+     * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#array_indices
+     */
     function isTreatedAsZeroByMemberAccess(value: unknown): boolean {
       return String(value) === '0';
     }
--- a/packages/eslint-plugin/src/rules/prefer-find.ts
+++ b/packages/eslint-plugin/src/rules/prefer-find.ts
@@ -9,10 +9,6 @@ import type {
 import * as tsutils from 'ts-api-utils';
 import type { Type } from 'typescript';
 
-import type {
-  MemberExpressionComputedName,
-  MemberExpressionNonComputedName,
-} from '../../../types/src/generated/ast-spec';
 import {
   createRule,
   getConstrainedTypeAtLocation,
@@ -84,8 +80,8 @@ export default createRule({
               callee.object,
             );
 
-            // As long as the object is an array or an optional chain on a
-            // nullable array, this is an Array.prototype.filter expression.
+            // As long as the object is a (possibly nullable) array,
+            // this is an Array.prototype.filter expression.
             if (isArrayish(filteredObjectType)) {
               return {
                 isBracketSyntaxForFilter,
@@ -297,8 +293,8 @@ export default createRule({
  */
 function isStaticMemberAccessOfValue(
   memberExpression:
-    | MemberExpressionComputedName
-    | MemberExpressionNonComputedName,
+    | TSESTree.MemberExpressionComputedName
+    | TSESTree.MemberExpressionNonComputedName,
   value: string,
   scope?: Scope.Scope | undefined,
 ): boolean {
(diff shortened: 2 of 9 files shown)


packages/eslint-plugin/docs/rules/prefer-find.md
```diff
@@ -0,0 +1,29 @@
...
+description: 'Enforce the use of Array.prototype.find() over Array.prototype.filter() followed by [0] when looking for a single result.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/prefer-find** for documentation.
+
+When searching for the first item in an array matching a condition, it may be tempting to use code like `arr.filter(x => x > 0)[0]`. However, it is simpler to use [Array.prototype.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead, `arr.find(x => x > 0)`, which also returns the first entry matching a condition.
+
+Beware that the two approaches are not quite identical, however! `.find()` will only execute the callback on array elements until it finds a match, whereas `.filter()` executes the callback for all array elements. Therefore, when fixing errors from this rule, be sure that your `.filter()` callbacks do not have side effects.
```
JoshuaKGoldberg, line 11: [Suggestion] What do you think about using a [Docusaurus admonition](https://docusaurus.io/docs/markdown-features/admonitions)? Maybe, `:::info` to not be as severe-looking as `:::warning`?

packages/eslint-plugin/docs/rules/prefer-find.md
```diff
@@ -0,0 +1,29 @@
+---
+description: 'Enforce the use of Array.prototype.find() over Array.prototype.filter() followed by [0] when looking for a single result.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/prefer-find** for documentation.
+
+When searching for the first item in an array matching a condition, it may be tempting to use code like `arr.filter(x => x > 0)[0]`. However, it is simpler to use [Array.prototype.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead, `arr.find(x => x > 0)`, which also returns the first entry matching a condition.
```
JoshuaKGoldberg, line 9: [Suggestion] You might want to also mention that it's faster to use `.find` because of it returning the first entry. ⚡

JoshuaKGoldberg, line 9: [Style] We'll lint for this eventually, but I personally like keeping to one sentence per line (https://github.com/JoshuaKGoldberg/sentences-per-line?tab=readme-ov-file#sentences-per-line). What do you think?

```suggestion
When searching for the first item in an array matching a condition, it may be tempting to use code like `arr.filter(x => x > 0)[0]`.
However, it is simpler to use [Array.prototype.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead, `arr.find(x => x > 0)`, which also returns the first entry matching a condition.
```

Not a blocker if you hate it 😄

packages/eslint-plugin/docs/rules/prefer-find.md
```diff
@@ -0,0 +1,29 @@
...
+
+### ✅ Correct
+
+```ts
+[1, 2, 3].find(x => x > 1);
+```
+
+## When Not To Use It
+
+If you don't mind `.filter(...)[0]`, you can avoid this rule.
```
JoshuaKGoldberg, line 29: [Docs] This should also mention that if you're intentionally writing `.filter`s with side effects, this rule might get in your way.

#7257 -> #7447 has a lot of examples of good _When Not To Use It_ docs.

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,110 @@
...
+import { createRule, getParserServices } from '../util';
+
+export default createRule({
+  name: 'prefer-find',
+  meta: {
+    docs: {
+      description:
+        'Enforce the use of Array.prototype.find() over Array.prototype.filter() followed by [0] when looking for a single result',
+      requiresTypeChecking: true,
+      recommended: 'stylistic',
```
JoshuaKGoldberg, line 14: [Bug] Yeah this would be a breaking change 😞 so we can't just yet. We don't have a "stylistic-strict" ruleset so there's really nowhere for this to go. 

The good news is, we're starting to ramp up on work for v7, so this won't be present-but-not-in-any-config for long!

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,110 @@
...
+    },
+    fixable: 'code',
+    messages: {
+      preferFind: 'Use .filter(...)[0] instead of .find(...)',
+      preferFindFix: 'Use .filter(...)[0] instead of .find(...)',
```
JoshuaKGoldberg, line 19: [Style] Just nitpicking, I assumed reading this at first to be an autofix/`--fix` rather than a suggestion. Maybe:

```suggestion
      preferFindSuggestion: 'Use .filter(...)[0] instead of .find(...)',
```

If you're going off existing precedent then totally ignore this. I do wonder if we should make a lint rule in [`eslint-plugin-eslint-plugin`](https://github.com/eslint-community/eslint-plugin-eslint-plugin). WDYT?

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,110 @@
...
+  defaultOptions: [],
+
+  create(context) {
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    return {
+      MemberExpression(node): void {
+        // Check if it looks like <<stuff>>[0] or <<stuff>>['0'], but not <<stuff>>?.[0]
+        if (
```
JoshuaKGoldberg, line 35: [Style] This is quite a lot of nesting for consecutive `if` statements. I'd suggest refactoring some of the logic to shared utility helpers. That way some of the nestings(?) can be reduced and/or removed.

One vague direction example:

```ts
if (!isZeroIndexOrSomeNameLikeThat(node)) {
  return;
}
```

Another vague example:

```ts
if (
  isZeroIndexOrSomeNameLikeThat(node) &&
  whateverTheOtherConditionIs(node.object)
)
```

[Praise] FWIW I found the comments very helpful 😄 so if you end up not doing this then +1 to them. But they are an indication IMO that a well-named function could clear up the code instead, too.

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,110 @@
...
+          !node.optional &&
+          node.property.type === AST_NODE_TYPES.Literal &&
+          (node.property.value === 0 ||
+            node.property.value === '0' ||
+            node.property.value === 0n)
+        ) {
+          // Check if it looks like <<stuff>>(...)[0], but not <<stuff>>?.(...)[0]
+          if (
+            node.object.type === AST_NODE_TYPES.CallExpression &&
+            !node.object.optional
```
JoshuaKGoldberg, line 46: [Question] Why not? This seems like it'd be a case folks would want to hit. We'd probably get a bug report from it not working.

E.g. for the `([]?.filter(f))[0];` case, I don't see why we wouldn't want to flag that.

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,110 @@
...
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    return {
+      MemberExpression(node): void {
+        // Check if it looks like <<stuff>>[0] or <<stuff>>['0'], but not <<stuff>>?.[0]
+        if (
+          node.computed &&
+          !node.optional &&
+          node.property.type === AST_NODE_TYPES.Literal &&
```
JoshuaKGoldberg, line 38: [Bug] We have at least one utility for this, `getStaticValue`. That way if someone does something like...

```ts
const value = 0;

values.filter(/* ... */)[value];
```

...then it'll be supported. Which is a good test case to add.

Here and the other `type === AST_NODE_TYPES.Literal`.

packages/eslint-plugin/tests/rules/prefer-find.test.ts
```diff
@@ -0,0 +1,176 @@
...
+const ruleTester = new RuleTester({
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    tsconfigRootDir: rootDir,
+    project: './tsconfig.json',
+  },
+});
+
+ruleTester.run('prefer-find', rule, {
+  valid: [
```
JoshuaKGoldberg, line 17: [Testing] A couple more test cases come to mind:

* Using `[1]` instead of `[0]`
* Using `.at(1)`
* Using `.filter(...)` without a `[0]`
* Putting something between the `.filter(...)` and the `[0]`
* Using another method name, like `.find(...)`

packages/eslint-plugin/tests/rules/prefer-find.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    '([]?.filter(f))[0];',
+    // Be sure that we don't try to mess with this case, since the member access
+    // should not need to be optional for the cases the rule is concerned with.
+    '[].filter(() => true)?.[0];',
+    // Be sure that we don't try to mess with this case, since the function call
+    // should not need to be optional for the cases the rule is concerned with.
+    '[].filter?.(() => true)[0];',
+  ],
+
+  invalid: [
```
JoshuaKGoldberg, line 43: [Testing] A few invalid cases come to mind...

* Using `.at(0)` should be flagged too, I think?
* Comments: please enjoy the obnoxious code that is https://github.com/typescript-eslint/typescript-eslint/pull/8067#discussion_r1438397989 😄 
* `0.0`
* Negatives for all the bigint/number cases: `-0`, `-0.0`, `-0n`, etc.
* <code>\`0\`</code>
* `Array<string>`
* `ReadonlyArray<string>`
* `readonly string[]`
* Tuples
* Constrained type parameters (`function withValues<T extends string[]>(values: T) {`)
    * Tip: `getConstrainedTypeAtLocation`

JoshuaKGoldberg (changes requested):
A solid start, thanks for sending this in! Left some requests around edge cases to work on. Have fun 😄 

Phoebe from Friends laughing, slightly pained maybe. Caption: 'SO MUCH FUN!'

packages/eslint-plugin/docs/rules/prefer-find.md
kirkwaiblinger: Changed

packages/eslint-plugin/docs/rules/prefer-find.md
kirkwaiblinger: Changed

kirkwaiblinger: Changed

packages/eslint-plugin/docs/rules/prefer-find.md
kirkwaiblinger: Changed

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: Removed

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: Changed the messageIds.

I do have a question though of whether the messages themselves are ok, or need to be more thought out. I guess at a minimum I should probably make it use `.at()`.... or, make the wording _less_ specific, i.e. `Use Array.prototype.find() rather than accessing the first element of Array.prototype.filter()`. 

In any case, please let me know what's expected here! Thanks :)

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: Significant changes have been made in this area.

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: Changed.... though, I also wonder whether we should try to account for this case (which doesn't presently work)
```ts
declare const filterLiteral: 'filter';
[1, 2, 3][filterLiteral](callback)[0];
```
This would require the type checker, of course. Not sure if that's a significant tradeoff or anything.

packages/eslint-plugin/tests/rules/prefer-find.test.ts
kirkwaiblinger: added coverage in these areas

packages/eslint-plugin/tests/rules/prefer-find.test.ts
kirkwaiblinger: added coverage in these areas... The `0n` tests do pass, but are a little suspect in that typescript considers it an error to index into an array with a bigint (surprisingly, tbh).

packages/eslint-plugin/src/rules/prefer-find.ts
bradzacher: nah we aren't looking to support weird shit like that. Variables as array indexers are normal - variables as indexers for method names are not.

packages/eslint-plugin/tests/rules/prefer-find.test.ts
bradzacher: > little suspect in that typescript considers it an error to index into an array with a bigint

`bigint` and `number` are two separate types entirely and the APIs are defined as accepting `number`. At runtime you'll find that there's an implicit runtime coercion to a string that occurs - which is why `'0'` is also a valid indexer.

kirkwaiblinger: ah, yes, understood. Perhaps I should have phrased my comment differently. I was trying to find a way to insinuate the question "Bigints are not valid as index values. Are you sure that we want to have tests for them in the repo, regardless of whether they can be made to pass?" without coming across as challenging what I was asked to do.

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,283 @@
...
+export default createRule({
+  name: 'prefer-find',
+  meta: {
+    docs: {
+      description:
+        'Enforce the use of Array.prototype.find() over Array.prototype.filter() followed by [0] when looking for a single result',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      preferFind: 'Use .find(...) instead of .filter(...)[0]',
```
JoshuaKGoldberg, line 25: [Phrasing] _"Use"_ is an action verb, which we'd normally put in the suggestion message. _"Prefer"_ is what we tend to put for the actual rule error complaint.

```suggestion
      preferFind: 'Prefer .find(...) instead of .filter(...)[0]',
```

Related: https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/368

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,283 @@
...
+        }
+      }
+
+      return isAtLeastOneArrayishComponent;
+    }
+
+    function getObjectIfArrayAtExpression(
+      node: TSESTree.CallExpression,
+    ): TSESTree.Expression | undefined {
+      if (node.arguments.length === 1) {
```
JoshuaKGoldberg, line 122: [Bug] Sorry, just realizing this now on deeper re-review: `filter` can take a second argument, `thisArg`. So this will need to allow that & have test cases.

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,283 @@
...
+            (callee.computed
+              ? getStaticValue(callee.property, globalScope)?.value === 'at'
+              : callee.property.name === 'at')
+          ) {
+            return callee.object;
+          }
+        }
+      }
+
+      return undefined;
```
JoshuaKGoldberg, line 138: [Style] What do you think about returning early to avoid more indentation?

```ts
if (node.arguments.length !== 1) {
  return undefined;
}

// ...
```

(here and elsewhere)

Tangent: I used to use the phrase _"invert if, reduce nesting"_ more in PR reviews. Now I'd rather make a lint rule for it: https://github.com/JoshuaKGoldberg/create-typescript-app/issues/1267

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,283 @@
...
+      if (node.arguments.length === 1) {
+        const atArgument = getStaticValue(node.arguments[0], globalScope);
+        if (atArgument != null && isTreatedAsZeroByArrayAt(atArgument.value)) {
+          const callee = node.callee;
+          if (
+            callee.type === AST_NODE_TYPES.MemberExpression &&
+            !callee.optional &&
+            (callee.computed
+              ? getStaticValue(callee.property, globalScope)?.value === 'at'
+              : callee.property.name === 'at')
```
JoshuaKGoldberg, line 131: [Non-Action] This code here tends to show up a lot: _"check if this call/field/property computes to or is statically a particular value"_. `getStaticValue` works only for the `.computed` side of that. 

Proposal:

* In this PR, can you add a utility in this file?
* Separate from PR, someone should file an issue suggesting using that kind of utility in more places

WDYT?

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,283 @@
...
+  name: 'prefer-find',
+  meta: {
+    docs: {
+      description:
+        'Enforce the use of Array.prototype.find() over Array.prototype.filter() followed by [0] when looking for a single result',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      preferFind: 'Use .find(...) instead of .filter(...)[0]',
+      preferFindSuggestion: 'Use .find(...) instead of .filter(...)[0]',
```
StyleShit, line 26: Seems like most of the messages have a period at the end, so maybe let's add them also here? just for the sake of consistency:

```suggestion
      preferFind: 'Use .find(...) instead of .filter(...)[0].',
      preferFindSuggestion: 'Use .find(...) instead of .filter(...)[0].',
```

StyleShit (changes requested):

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,283 @@
...
+    const globalScope = getScope(context);
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    function parseIfArrayFilterExpression(
+      expression: TSESTree.Expression,
+    ):
+      | { isBracketSyntaxForFilter: boolean; filterNode: TSESTree.Node }
+      | undefined {
+      if (expression.type === AST_NODE_TYPES.SequenceExpression) {
```
JoshuaKGoldberg, line 45: [Praise] Nice 😄 I probably would have forgotten about sequence expressions...

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,283 @@
...
+
+  create(context) {
+    const globalScope = getScope(context);
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    function parseIfArrayFilterExpression(
+      expression: TSESTree.Expression,
+    ):
+      | { isBracketSyntaxForFilter: boolean; filterNode: TSESTree.Node }
```
JoshuaKGoldberg, line 43: [Refactor] I just filed https://github.com/typescript-eslint/typescript-eslint/issues/8320 thinking on inline object literals. WDYT?

packages/eslint-plugin/tests/rules/prefer-find.test.ts
```diff
@@ -0,0 +1,532 @@
...
+ruleTester.run('prefer-find', rule, {
+  valid: [
+    `
+      interface JerkCode<T> {
+        filter(predicate: (item: T) => boolean): JerkCode<T>;
+      }
+
+      declare const jerkCode: JerkCode<string>;
+
+      jerkCode.filter(item => item === 'aha');
```
JoshuaKGoldberg, line 25: [Testing] This should have a `[0]` at the end, as the current code wouldn't violate the rule.

But otherwise +1 😄 love the copying of the code example from before!

packages/eslint-plugin/tests/rules/prefer-find.test.ts
JoshuaKGoldberg: I can see why it'd feel funky writing a test for a report in the presence of a type error. But that's ok: if the rule _should_ report in that case, a type error is kind of just a coincidental by-product.

Note that #8298 tracks testing that type-aware rules' test cases are error-free. I think this is another good example of when we'd want to add an _"intentionally allow these type errors"_ kind of feature.

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,283 @@
...
+                {
+                  messageId: 'preferFindSuggestion',
+                  fix: (fixer): TSESLint.RuleFix[] => {
+                    const sourceCode = getSourceCode(context);
+                    return [
+                      generateFixToReplaceFilterWithFind(
+                        fixer,
+                        filterExpression,
+                      ),
+                      // get rid of the .at(0) or ['at'](0)
```
JoshuaKGoldberg, line 228: [Nit] The later comment capitalizes the 'g'. This one doesn't. Up for switching to always the same style? 😄

JoshuaKGoldberg (changes requested):
This is looking fantastic! Left a few comments, but mostly just nitpicks. Really happy about the implementation. ☺️

I think the only remaining blocker is the `0n` bigint tests?

JoshuaKGoldberg (dismissed):
Wait sorry, (as you pointed out in Twitch chat), the `0n`s are there! I just missed them!

So I think it's just a few nitpicky comments from me left. I'll set a reminder to merge before our next release if you don't have time or interest for them.

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: changed

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: changed

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: Changed this specific instance.... I think it would be better if you'd be willing to chang the other instances that you're concerned with while merging, though. 

I actually don't subscribe to unqualified use of the guard-clause-if pattern (since I think nested ifs without elses make the happy path clearer), so I don't feel qualified to carry out the changes in a way that you find appealing 😅

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: yep. You make a fair point. changed!

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: Yep, changed!
And filed https://github.com/typescript-eslint/typescript-eslint/issues/8323

packages/eslint-plugin/tests/rules/prefer-find.test.ts
kirkwaiblinger: good catch. changed!

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: changed

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,313 @@
...
+  Scope,
+  SourceCode,
+} from '@typescript-eslint/utils/ts-eslint';
+import * as tsutils from 'ts-api-utils';
+import type { Type } from 'typescript';
+
+import type {
+  MemberExpressionComputedName,
+  MemberExpressionNonComputedName,
+} from '../../../types/src/generated/ast-spec';
```
auvred, line 15: [nit] Other rules don't import `@typescript-eslint/types` package using relative paths. + Additionally `TSESTree` is already imported, I think these types should be accessed from it.

Also, there are `TSESTree.MemberExpressionComputedName` and `MemberExpressionComputedName` in one file. Looks a bit inconsistent :smile: 

Let's refactor it? What do you think?

packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -0,0 +1,313 @@
...
+
+      return undefined;
+    }
+
+    /**
+     * Tells whether the type is a possibly nullable array/tuple or union thereof.
+     */
+    function isArrayish(type: Type): boolean {
+      let isAtLeastOneArrayishComponent = false;
+      for (const unionPart of tsutils.unionTypeParts(type)) {
```
auvred, line 107: [maybe bug] Perhaps we should check for intersections here too:

This code is not reported (but I think it should be!)
_[playground](https://deploy-preview-8216--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3rGAueA3tvgIzwC%2BA2gLrwBkh8ARvgEwU0BQnOAdAGYBLCBhAwAFOICU8ALwA%2BeBhgBXEFMoAGakA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wlq%2BYATdFF7QA9tEjgwAXxCygA&tsconfig=&tokens=false)_

```ts
declare const arr: { a: 1 }[] & { b: 2 }[]

arr.filter(() => true)[0]
```

+1 on https://github.com/typescript-eslint/typescript-eslint/pull/8250#discussion_r1471943792

auvred (changes requested):

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: Added support for this.

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: yep, cleaned that up!

auvred:
looks good to me

auvred (approved):

packages/eslint-plugin/src/rules/prefer-find.ts
JoshuaKGoldberg: Maybe a good lint rule internally, as a followup?

kirkwaiblinger:
Oh dear. I've been making commits under the wrong github account.... @JoshuaKGoldberg do you mind if I force push to correct the author info?

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: Filed https://github.com/typescript-eslint/typescript-eslint/pull/8216#discussion_r1472343770

kirkwaiblinger: Filed https://github.com/typescript-eslint/typescript-eslint/issues/8345

packages/eslint-plugin/src/rules/prefer-find.ts
JoshuaKGoldberg: Haha makes sense, I appreciate the honesty 😄. I'll probably lazy-out and just leave as-is.

Most of the time when I've put this feedback the author either didn't care one way or the other or didn't think of the reduced-nesting approach. Since you actually have an opinion I'm much happier just rolling with it.

kirkwaiblinger: Sounds good. Sorry for being difficult!

JoshuaKGoldberg: I wish _more_ people were _more_ "difficult" in this way. 😄

JoshuaKGoldberg (approved):
Aggressively enthusiastic guy on a sports field shouting 'let's goooo' in front of a crowd of people

packages/eslint-plugin/src/rules/prefer-find.ts
JoshuaKGoldberg: Heck yes!

PR #8217, opened 2024-01-08
docs: underline URLs

### LINKED ISSUE #8214: Website accessibility: links not distinguishable without relying on color

### Suggested Changes

I ran the [axe browser dev tools](https://www.deque.com/axe/browser-extensions) on https://typescript-eslint.io/getting-started. The report indicated inline links such as the _ESLint_ and _TypeScript_ ones under _Step 1: Installation_ aren't visually distinct enough from surrounding text to always be perceived as different. The color contrast of their current color in light mode (`#889dfa`) compared to surrounding text (`#e3e3e3`) is only color contrast 1.98:1 - lower than the AA minimum of 3:1.

Getting links distinguishable from other body text is hard by color alone. Especially when they have to _also_ have sufficient color contrast against the background of the page. It's generally easiest to add a non-color indicator such as a <ins>text underline</ins> the way GitHub did recently (https://github.blog/changelog/2023-10-18-new-default-underlined-links-for-improved-accessibility).

Accepting PRs to make the website's links underlined!

### Affected URL(s)

lucas-amberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8214 

## Overview

Added underlines to all URLs on the documentation and took steps to verify everything worked as usual.

### DIFF
--- a/docs/Contributing.mdx
+++ b/docs/Contributing.mdx
@@ -11,5 +11,5 @@ So, from the bottom of our hearts, _thank you_ 💕.
 
 ## Getting Started
 
-Please read our [Code of Conduct](https://github.com/typescript-eslint/typescript-eslint/blob/main/CODE_OF_CONDUCT.md) first.
-If you're new to open source, you may also find the [How to Contribute to Open Source](https://opensource.guide/how-to-contribute) guide from https://opensource.guide useful.
+Please read our <u>[Code of Conduct](https://github.com/typescript-eslint/typescript-eslint/blob/main/CODE_OF_CONDUCT.md)</u> first.
+If you're new to open source, you may also find the <u>[How to Contribute to Open Source](https://opensource.guide/how-to-contribute)</u> guide from https://opensource.guide useful.
--- a/docs/Contributing.mdx
+++ b/docs/Contributing.mdx
@@ -12,4 +12,4 @@ So, from the bottom of our hearts, _thank you_ 💕.
 ## Getting Started
 
 Please read our <u>[Code of Conduct](https://github.com/typescript-eslint/typescript-eslint/blob/main/CODE_OF_CONDUCT.md)</u> first.
-If you're new to open source, you may also find the <u>[How to Contribute to Open Source](https://opensource.guide/how-to-contribute)</u> guide from https://opensource.guide useful.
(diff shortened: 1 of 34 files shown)


docs/Contributing.mdx
```diff
@@ -11,5 +11,5 @@ So, from the bottom of our hearts, _thank you_ 💕.
 
 ## Getting Started
 
-Please read our [Code of Conduct](https://github.com/typescript-eslint/typescript-eslint/blob/main/CODE_OF_CONDUCT.md) first.
-If you're new to open source, you may also find the [How to Contribute to Open Source](https://opensource.guide/how-to-contribute) guide from https://opensource.guide useful.
+Please read our <u>[Code of Conduct](https://github.com/typescript-eslint/typescript-eslint/blob/main/CODE_OF_CONDUCT.md)</u> first.
+If you're new to open source, you may also find the <u>[How to Contribute to Open Source](https://opensource.guide/how-to-contribute)</u> guide from <u>https://opensource.guide</u> useful.
```
auvred, line 15: [just asking] Oh, this approach would require us to remember to always wrap any links in our docs in a `<u>` tag. Maybe we should underline them with css only? `text-decoration: underline` should do the trick

Moreover, I suppose we might want links in blog posts to be underlined too. I think it would be much easier to maintain if the underline was added automatically

So I guess we can just apply `text-decoration` to all links inside blog/docs container. Or vice-versa - apply underline to all `a` tags on the page and disable `text-decoration` in undesirable places like the side bar or table of content

What do you think?

auvred (changes requested):

docs/Contributing.mdx
auvred: Perhaps Docusaurus even has some functionality to customize markdown links (though I haven't searched)

lucas-amberg: I definitely agree, I'm new to the MDX format and I was unsure how styling was applied in these pages in the first place aside from general markdown syntax.

Should be a fairly easy fix, I'll get it done in the morning.

JoshuaKGoldberg: Yup, what @auvred said - we're definitely going to forget this if we need to do it manually. And it's a bit of a pain. There should be a straightforward path to get most or all of the links inside markdown content.

JoshuaKGoldberg (changes requested):
Thanks for starting! What auvred said on making this more automatic.

PR #8225, opened 2024-01-08
docs: underline URLs

### LINKED ISSUE #8214: Website accessibility: links not distinguishable without relying on color

lucas-amberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8214 

## Overview
Updated CSS to add underlines to URLs and added another rule to remove underlines from the table of contents, next page buttons, edit page button, and menu sidebars along with the social media links in the footer (as they can be distinguished with the external link icon).

### DIFF
--- a/packages/website/src/css/custom.css
+++ b/packages/website/src/css/custom.css
@@ -191,3 +191,17 @@ h6 {
 [data-theme='dark'] .code-block-added-line {
   background-color: #00ff9510;
 }
+
+a {
+  text-decoration: underline;
+}
+
+/*Removes underlines from table of contents, menu (left hand sidebar),
+edit this page button, and footer links (github, twitter, etc) respectfully*/
+a.table-of-contents__link,
+a.pagination-nav__link,
+a.menu__link,
+a.theme-edit-this-page,
+footer * {
+  text-decoration: none;
+}
--- a/packages/website/src/css/custom.css
+++ b/packages/website/src/css/custom.css
@@ -196,8 +196,9 @@ a {
   text-decoration: underline;
 }
 
-/*Removes underlines from table of contents, menu (left hand sidebar),
-edit this page button, and footer links (github, twitter, etc) respectfully*/
+/*Removes underlines from table of contents, next page button,
+menu (left hand sidebar), edit this page button,
+and footer links (github, twitter, etc) respectfully*/
 a.table-of-contents__link,
 a.pagination-nav__link,
 a.menu__link,


JoshuaKGoldberg:
Btw @lucas-amberg is there a reason you closed #8217 and sent this new one?

In case this is useful info: you don't have to delete a PR then re-make it every time you have a new commit 🙂. You can push commits to the same branch. 

I'd also recommend sending from a branch other than `main`. https://stackoverflow.com/questions/60631439/why-is-it-bad-practice-to-submit-pull-request-from-master-branch (replace "`master`" with "`main`")

lucas-amberg:
The reason I closed it was just so I could restart with a clean fork, I'm sorry for the confusion. 

Also thank you for the recommendation, I'll make sure to do that in the future. I appreciate the help as I'm really new to this stuff  😅.

JoshuaKGoldberg:
Ooh! Welcome to Git / GitHub / open-source / whatever-combination-of-those-is-new-for-you! 😄🙌 

You're actually doing quite well with adhering to standards (the PR is titled well, addresses an accepted+open issue, etc.). This is great!

packages/website/src/css/custom.css
```diff
@@ -191,3 +191,25 @@ h6 {
...
+a.menu__link,
+a.theme-edit-this-page,
+footer *,
+.button,
+.navbar__link,
+.navbar__brand,
+aside *,
+header * {
+  text-decoration: none;
+}
```
JoshuaKGoldberg, line 215: [Request] I think this approach works well for the how the site is structured now, but:

* `*`s are kind of hard to work with long-term. If we add in more cascading layers of styles they'll be hard to override.
* Similarly, this is a list based on the structure & stuff we use now. But if that changes this list will have to change.

If I understand it right, this approach is making all anchors underlined by default, then disabling that for certain others. How about going the opposite direction instead - targeting only the anchors we know we want to be underlined? I think that's doable by selecting all anchors inside markdown, like `.markdown a`. Or maybe even only the anchors inside paragraphs. Thoughts?

JoshuaKGoldberg (changes requested):
Progress! 🚀

packages/website/src/css/custom.css
lucas-amberg: I agree, thank you for the idea, that should work perfectly.

lucas-amberg:
This route would've been a lot easier

packages/website/src/css/prism.css
```diff
@@ -6,16 +6,16 @@
...
-  --token-color-function: #569cd6;
+  --token-color-function: #2b74b1;
   --token-color-function-variable: #000;
   --token-color-important: #e90;
-  --token-color-class-name: #2b91af;
+  --token-color-class-name: #237690;
   --token-color-selector: #800000;
   --token-color-regexp: #800000;
-  --token-color-property: #d00;
+  --token-color-property: #c70000;
```
StyleShit, line 18: [Curious]

How did you choose those specific colors?

StyleShit (approved):
Welcome to the open-source world!

Looks good to me :)

packages/website/src/css/prism.css
lucas-amberg: I ran axe devtools to see which colors on each page were below the 4.5:1 threshold and then I used [this website](https://webaim.org/resources/contrastchecker/) to check the contrast ratio and modify the current foreground color until it was over that amount. Then I re-ran the devtools to verify the changes were adequate 🙂.

packages/website/src/css/prism.css
auvred, line 1: [tip] Try to avoid fixing several unrelated issues in one PR, even if they are small. This is generally a good practice + it helps maintainers better focus on reviewing fixes for a particular issue.

If two issues are about the same bug/feature, then they should be merged into one issue. Thus, one issue = one PR

auvred (approved):
Looks cool! Thanks for working on this! :wink:

Josh-Cena (approved):

packages/website/src/css/prism.css
JoshuaKGoldberg: Big +1 on https://webaim.org/resources/contrastchecker, great tool!

JoshuaKGoldberg (approved):
Thanks for iterating with us on this!

Joe Pesci dressed as a cop in Home Alone 2 saying 'Excellent. Excellent.'

PR #8245, opened 2024-01-13
chore: make lint use eslint-plugin outputs as inputs

JamesHenry:
- Adds the `build` of `eslint-plugin` as a dependency of the lint target
- Configures the `inputs` of lint to take the `.js` files of that dependency into account when hashing for cache hits (https://nx.dev/reference/project-configuration#dependent-tasks-output)
- Also makes the `copy-ast-spec` target cacheable and makes its config more succinct

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -49,25 +49,17 @@
     "targets": {
       "copy-ast-spec": {
         "dependsOn": [
-          {
-            "target": "build",
-            "projects": "dependencies"
-          }
+          "^build"
         ],
         "outputs": [
           "{projectRoot}/src/generated"
-        ]
+        ],
+        "cache": true
       },
       "build": {
         "dependsOn": [
-          {
-            "target": "build",
-            "projects": "dependencies"
-          },
-          {
-            "target": "copy-ast-spec",
-            "projects": "self"
-          }
+          "^build",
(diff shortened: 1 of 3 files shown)


packages/types/package.json
```diff
@@ -49,25 +49,17 @@
     "targets": {
       "copy-ast-spec": {
         "dependsOn": [
-          {
-            "target": "build",
-            "projects": "dependencies"
-          }
+          "^build"
```
JamesHenry, line 52: This is an exact equivalent `^` means "of dependencies"

packages/types/package.json
```diff
@@ -49,25 +49,17 @@
...
         "dependsOn": [
-          {
-            "target": "build",
-            "projects": "dependencies"
-          },
-          {
-            "target": "copy-ast-spec",
-            "projects": "self"
-          }
+          "^build",
```
JamesHenry, line 61: This is an exact equivalent `^` means "of dependencies"

packages/types/package.json
```diff
@@ -49,25 +49,17 @@
...
+          "copy-ast-spec"
```
JamesHenry, line 62: This is an exact equivalent, no `^` means "self"/"from current project"

bradzacher (approved):

PR #8252, opened 2024-01-14
fix(rule-tester): fixes on the same line are not fixed properly

### LINKED ISSUE #8251: Bug: RuleTester doesn't fix properly when there are multiple fixes on the same line

### Relevant Package

rule-tester

### Repro Code
See: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/space-infix-ops.test.ts#L1021-L1049
```TypeScript
{
  invalid: [
    {
      code: `
        type Test=|string|(((() => void)))|string;
      `,
      output: `
        type Test = |string | (((() => void))) | string;
      `,
      errors: [
        {
          messageId: 'missingSpace',
          column: 18,
          line: 2,
        },
        {
          messageId: 'missingSpace',
          column: 19,
          line: 2,
        },
        {
          messageId: 'missingSpace',
          column: 26,
          line: 2,
        },
        {
          messageId: 'missingSpace',
          column: 43,
          line: 2,
        },
      ],
    },
  ],
}
```

### Description

Splits from this thread:
https://github.com/typescript-eslint/typescript-eslint/pull/8065#discussion_r1429216128

The code should be fixed properly when there are multiple fixes on the same line.
In the repro code, the fix should actually be
```TS
type Test = | string | (((() => void))) | string;
```

And in the thread, it should've been fixed regardless of me adding space as a workaround
(you can see that I've changed it from `${'a'}${'b'}` to `${'a'} ${'b'}`)

Opened a [draft PR](https://github.com/typescript-eslint/typescript-eslint/pull/8252) to start an investigation.
Seems like there are some actual wrong fixes in the tests.

StyleShit:
Closes #8251

### DIFF
--- a/packages/eslint-plugin/tests/rules/semi.test.ts
+++ b/packages/eslint-plugin/tests/rules/semi.test.ts
@@ -1069,7 +1069,7 @@ class PanCamera extends FreeCamera {
       },
       {
         code: 'a;\n++b;',
-        output: 'a\n++b;',
+        output: 'a\n++b',
         errors: [
           {
             line: 1,
--- a/packages/eslint-plugin/tests/rules/space-infix-ops.test.ts
+++ b/packages/eslint-plugin/tests/rules/space-infix-ops.test.ts
@@ -1022,7 +1022,7 @@ ruleTester.run('space-infix-ops', rule, {
         type Test=|string|(((() => void)))|string;
       `,
       output: `
-        type Test = |string | (((() => void))) | string;
+        type Test = | string | (((() => void))) | string;
       `,
       errors: [
         {
--- a/packages/rule-tester/src/utils/SourceCodeFixer.ts
+++ b/packages/rule-tester/src/utils/SourceCodeFixer.ts
@@ -63,7 +63,7 @@ export function applyFixes(
     const end = fix.range[1];
 
     // Remain it as a problem if it's overlapped or it's a negative range
-    if (lastPos >= start || start > end) {
+    if (lastPos > start || start > end) {
       remainingMessages.push(problem);
       return false;
     }


packages/eslint-plugin/tests/rules/semi.test.ts
```diff
@@ -1069,7 +1069,7 @@ class PanCamera extends FreeCamera {
       },
       {
         code: 'a;\n++b;',
-        output: 'a\n++b;',
```
StyleShit, line 1072: This was a bug in the tests. See other tests above.

packages/eslint-plugin/tests/rules/space-infix-ops.test.ts
```diff
@@ -1022,7 +1022,7 @@ ruleTester.run('space-infix-ops', rule, {
         type Test=|string|(((() => void)))|string;
       `,
       output: `
-        type Test = |string | (((() => void))) | string;
+        type Test = | string | (((() => void))) | string;
```
StyleShit, line 1025: This should be the real fix. See [this playground](https://typescript-eslint.io/play/#ts=5.3.3&fileType=.ts&code=C4TwDgpgBAKhDOwoF4A%2BiBOBLAdgc1QApjCBKFAPigDcB7LAE1OfWG3wG4g&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKWQBwEMBjRAWgEsA7AM3IA9SB7PVDSR66yAGnG0gACAFwCeeFEWjk8Q0inhUhAenzEyVWg2asoiaNEbRIfAL4gTQA&tsconfig=&tokens=false).

The test currently fails because it's being fixed to:
```TS
type Test =  | string | (((() => void))) | string;
```
(2 spaces after the `=`)
The reason is probably because there are 2 fixes that add spaces before and after the relevant characters (`=` & `|`), which then results in 2 spaces.

Not sure how to approach this. I have 2 options in mind:
1. Trimming the spaces after each fix (for sure a bad idea)
2. Re-running the rule tester on the same line after each fix (is that what happens in real-life? not sure if/how to do it though)


WDYT?

packages/rule-tester/src/utils/SourceCodeFixer.ts
```diff
@@ -63,7 +63,7 @@ export function applyFixes(
     const end = fix.range[1];
 
     // Remain it as a problem if it's overlapped or it's a negative range
-    if (lastPos >= start || start > end) {
+    if (lastPos > start || start > end) {
```
StyleShit, line 66: The actual fix that causes all of this mess 😅

bradzacher: https://github.com/eslint/eslint/blob/1f3744278433006042b8d5f4e9e1e488b2bbb011/lib/linter/source-code-fixer.js#L92

As per the comment at the top of the file - this code is directly forked from eslint core.

If it's broken here then this only changes tests run using our rule tester. Code run through eslint will have the old logic.

So either you've found a bug upstream or tests are running correctly, but correctly, but differently to how you expect.

If it's the former - the fix needs to land upstream before we land it here.

bradzacher (changes requested):
Above comment

bradzacher:
Closing for now as per above.
We can reopen if it lands upstream.

PR #8263, opened 2024-01-16
fix(eslint-plugin): [member-ordering] report alphabetical sorting for all groups instead of just the first failing group

### LINKED ISSUE #8100: [member-ordering] report alphabetical sorting for all groups instead of just the first failing group

### Playground Link

[link](https://typescript-eslint.io/play/#ts=5.3.2&fileType=.ts&code=MYGwhgzhAEBiD29oG8BQ1oDcwgK4FMAhaAXmgHJs99pDyBudLHAgQVIqoOlYdSYAOAJ3gAXfMHEATaBFFhRAS2C0OlFvj4ZhYidNnylK9mXXU%2BTCLgBGAW0WiAFAEoU0AL5MA5vlEARBTAXN08MYCF8BXxg5A8mCIhfGM9PIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AW0XYCNFpaA9tAAmfZgHN0YANrhsOPtCGQANHPlZ52SKIBmAQwT4pmrdqGjoUyPvjEAFvt75KZW-EKr1ZqJx58AFRIUKVkfLVZWMABJJlEADzBkSnEmfXxYaERvcMhk1PTMxC9w%2BUg3eHhafLSMrJLSiEiwADFKRHhhZByfSGIqADd0xGr8dNdaXXbOhsbIAGJ%2ByiH8EeQxlzJJ6eFZ0r6lVbJV4VHxramO3bVGnGJYbkYt9fPtq73cpZWR0TIhYdOlxmN1uBwERxOtF%2B-0hQOuPTMfQeTyhiD%2B0ABb2BCK0B2Ww1ozBeTDIIzhH16i0GBKJYxJZJ2FMR-XBaMhtP09Kx8NuUHujwmHK55JBcxZEMQpyc6wxx25TNx-JR0vwsroIpxZS%2BBI1vIW2tW8tF%2B3FbMlRs12iVE11oJemwteqFpMdoJVatdc2hGNhjONuVtjWaAEEyKTkMghN09Qa1hsJvowyhI1Z-ZTY2cHYnwymFVrDmbTvaE0mI8o08zkRNi1ts8ny5a7tTDd7MXWy6nG3yC8dza3Ie3cxXFVWtv3zYOGzHmyNnSNJ529VT8Ya57QF3mrT32UxiS6N8OtaPCbu6fvS0Ou2CJVLuDLE3QD1frbW76qH%2BuL1PQRmn0vf1%2Bi4-tuE6AZudzHn%2BdrxrWYGHtoa5QXM7ofkh%2BzjlKcFXmh4TNAA4og%2BCrNA0a3M0vIUfmK5xq84iEeBYDkZRlH6jOmYTHRxjwU0UTMSxpq9kWMG0JxDFMXxP7HjWIn0dxjGsF24kSSabEYTJXGKbxyliiBpxqaJclKdplYCmOaIwuaBmaZpxk6dRJ57iMVm8kZtnaMu3wOWeTmyTZbm9AJO6OepYlaf5R6mV5nIus5ZEKS5YXhU2rKCeub4erFQaJUlSKRShcqZaUkR%2BUlW72YVuHZeFrHlb5CWlfmKWQhVPiuf5uUoi1ZjFfVDU4NJXURFV7VroN8htW5NjpR%2BY3YD1cV9Tg%2Bl1XFJUNZAs08WAADKhHEaRWWLd29nScgK2HYtNWead51FcNk2BeaN0ab1fUddWwlnS9C2XRmalfaFR3XoWqLopiAOGfdtnvWZYOQhD1mvetGZrgjSOlVdNKntFay3ZVQOPacqN461UPGTDUVcmjq3ozlhNpfecrUxdb0vgz75MyT3XxT9rNsczd1Ax5BIC-jv26dUXNDQTUlS%2BNPMs%2Btz2A5dxPfYrGP5XQouk2tGP-XLc0K4Ll069zYAALKEfYAhdM%2BUnCZw%2BA2zywFNU9jvW7bDGY4a0lOy7PsZv7Xuu2Kx5qQH3tycDqWR6HQeqeZPrmlHYcmsea5p4n7tE9jXLZzHKP5y6hdXsLq4lyMZcxpB01yjXbs3uzHqN%2BHkVtypue0J3nxsb36b9wnMch870dXlnw-YfXdAD4i8dj%2BnuRz9L20wWAzCUC4tiUAAXuMAhMFe0mb9vjD7y4h%2BAwAwofMqwMcUbPmxfynnAj9ATp3ev-fH9B8eP9VQP3wN%2BOYgD34gKsJaAAujiAAvj0BB8hYEQAQXAoAA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

### Repro Code

```TypeScript
class Foo {
  valueB = 'value B';
  valueA = 'value A';

  protected static B = 'value';
  protected static A = 'value';

  submit() { }
  getData() { }
  create() { }
  reset() {}
}
```

### ESLint Config
<details>
<summary>Config</summary>

```javascript
{
  "rules": {
    "@typescript-eslint/member-ordering": [
      "error",
      {
        "default": {
          "order": "alphabetically",
          "memberTypes": [
            // Index signature
            "signature",
            "call-signature",
            // Fields
            "private-static-field",
            "#private-static-field",
            "protected-static-field",
            "public-static-field",
            "private-decorated-field",
            "protected-decorated-field",
            "public-decorated-field",
            "private-instance-field",
            "#private-instance-field",
            "protected-instance-field",
            "public-instance-field",
            "protected-abstract-field",
            "public-abstract-field",
            "private-field",
            "#private-field",
            "protected-field",
            "public-field",
            "static-field",
            "instance-field",
            "abstract-field",
            "decorated-field",
            "field",
            // Accessors
            "private-static-accessor",
            "#private-static-accessor",
            "protected-static-accessor",
            "public-static-accessor",
            "private-decorated-accessor",
            "protected-decorated-accessor",
            "public-decorated-accessor",
            "private-instance-accessor",
            "#private-instance-accessor",
            "protected-instance-accessor",
            "public-instance-accessor",
            "protected-abstract-accessor",
            "public-abstract-accessor",
            "private-accessor",
            "#private-accessor",
            "protected-accessor",
            "public-accessor",
            "static-accessor",
            "instance-accessor",
            "abstract-accessor",
            "decorated-accessor",
            "accessor",

            // Methods
            "public-static-method",
            "protected-static-method",
            "#private-static-method",
            "private-static-method",
            "public-decorated-method",
            "protected-decorated-method",
            "private-decorated-method",
            "public-instance-method",
            "protected-instance-method",
            "private-instance-method",
            "#private-instance-method",
            "public-abstract-method",
            "protected-abstract-method",
            "public-method",
            "protected-method",
            "private-method",
            "#private-method",
            "static-method",
            "instance-method",
            "abstract-method",
            "decorated-method",
            "method",
            // Static initialization
            "static-initialization",
            // Constructors
            "private-constructor",
            "protected-constructor",
            "public-constructor",
            "constructor"
          ]
        }
      }
    ]
  }
}
```

</details>

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

Get the complete list of all the errors at once.

If order of member types is not respected and alphabetical order is not respected within the member type we first get the error of the member type order then the error about the alphabetical order. Would be nice if we can have both errors at the same time.

### Actual Result

We have the first error on the property order.

Then start the alphabetical order errors on the protected static properties. When fixed we then have the error on the public properties. Plus we have one alphabetical order error at a time.

Same applies on the methods. When fixing the error on getData before submit we then have create must be before submit the create must be before getData then reset must be before submit.

### Additional Info

I hope it will be an easy enhancement. :slightly_smiling_face: 

Let me know if I need to provide more info.

Additionally, the error reported about where to place the node is not always the more logic to me. But it will require another issue and tied to the algorithm used to detect the errors which may be hard to change and we can deal with it. Let's improve one step at a time. :slightly_smiling_face:

arka1002:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8100 

## Overview

1. Previously, if a group of MemberType was unsorted, then it was reported ignoring the rest of the MemberTypes, now all of them are reported.
2. Previously, if the MemberTypes weren't in order wrt config, alphabetic sort was avoided, now both takes place.

### DIFF
--- a/packages/eslint-plugin/src/rules/member-ordering.ts
+++ b/packages/eslint-plugin/src/rules/member-ordering.ts
@@ -621,25 +621,25 @@ function getRank(
  * ```
  * interface Foo {
       [a: string]: number;
-
+ 
       a: x;
       B: x;
       c: x;
-
+ 
       c(): void;
       B(): void;
       a(): void;
-
+ 
       (): Baz;
-
+ 
       new (): Bar;
     }
  * ```
   The resulting array will look like this - [[a, B, c], [c, B, a]].
  * @param memberSet The members in the class of the source code on which the grouping operation will be performed.
  * @param memberType The member type order listed in the eslint config file.
- * @param supportsModifiers
- * @returns
+ * @param supportsModifiers it'll get passed to getRank()
+ * @returns the array of groups of members
  */
 function groupMembersByType(
   memberSet: Member[],
@@ -997,7 +997,7 @@ export default createRule<Options, MessageIds>({
 
       /**
        * It runs an alphabetic sort on the groups of the members of the class in the source code.
-       * @param memberSet
+       * @param memberSet The members in the class of the source code on which the grouping operation will be performed.
        */
       const checkAlphaSortForAllMembers = (memberSet: Member[]): undefined => {
         const hasAlphaSort = !!(order && order !== 'as-written');
--- a/packages/eslint-plugin/src/rules/member-ordering.ts
+++ b/packages/eslint-plugin/src/rules/member-ordering.ts
@@ -615,13 +615,38 @@ function getRank(
   return getRankOrder(memberGroups, orderConfig);
 }
 
-function changeToGroups(
+/**
+ * If there are more than 1 consecutive same ranked member in 1 class, those members are put in an array & an array of these members' array is returned as a result. If, for example, the memberSet parameter looks like the below example - (contd. after the example tag)
+ * @example
+ * ```
+ * interface Foo {
+      [a: string]: number;
+
+      a: x;
+      B: x;
+      c: x;
+
+      c(): void;
+      B(): void;
+      a(): void;
+
+      (): Baz;
+
+      new (): Bar;
+    }
+ * ```
+  The resulting array will look like this - [[a, B, c], [c, B, a]].
+ * @param memberSet The members in the class of the source code on which the grouping operation will be performed.
+ * @param memberType The member type order listed in the eslint config file.
+ * @param supportsModifiers
+ * @returns
+ */
+function groupMembersByType(
   memberSet: Member[],
   memberType: MemberType[],
   supportsModifiers: boolean,
 ): Member[][] {
   const groupedMembers: Member[][] = [];
-
   const getPreviousRank = (): number | undefined => {
     if (groupedMembers.at(-1)?.at(0) === undefined) {
       return undefined;
@@ -632,16 +657,19 @@ function changeToGroups(
       supportsModifiers,
     );
   };
-
   memberSet.forEach((member, index) => {
-    if (getRank(member, memberType, supportsModifiers) === getPreviousRank()) {
-      groupedMembers.at(-1)?.push(member);
-    } else if (index === memberSet.length - 1) {
+    const rankOfcurrentMember = getRank(member, memberType, supportsModifiers);
+    const rankOfNextMember = getRank(
+      memberSet[index + 1],
+      memberType,
+      supportsModifiers,
+    );
+    if (index === memberSet.length - 1) {
       return;
-    } else if (
-      getRank(member, memberType, supportsModifiers) ===
-      getRank(memberSet.at(index + 1)!, memberType, supportsModifiers)
-    ) {
+    }
+    if (rankOfcurrentMember === getPreviousRank()) {
+      groupedMembers.at(-1)?.push(member);
+    } else if (rankOfcurrentMember === rankOfNextMember) {
       groupedMembers.push([member]);
     }
   });
@@ -967,10 +995,14 @@ export default createRule<Options, MessageIds>({
       let memberTypes: MemberType[] | string | undefined;
       let optionalityOrder: OptionalityOrder | undefined;
 
-      const runSort = (memberSet: Member[]): undefined => {
+      /**
+       * It runs an alphabetic sort on the groups of the members of the class in the source code.
+       * @param memberSet
+       */
+      const checkAlphaSortForAllMembers = (memberSet: Member[]): undefined => {
         const hasAlphaSort = !!(order && order !== 'as-written');
         if (hasAlphaSort && Array.isArray(memberTypes)) {
-          changeToGroups(memberSet, memberTypes, supportsModifiers).forEach(
+          groupMembersByType(memberSet, memberTypes, supportsModifiers).forEach(
             groupMember => {
               checkAlphaSort(groupMember, order as AlphabeticalOrder);
             },
@@ -991,18 +1023,16 @@ export default createRule<Options, MessageIds>({
           );
 
           if (grouped == null) {
-            runSort(members);
+            checkAlphaSortForAllMembers(members);
             return false;
           }
 
           if (hasAlphaSort) {
-            const sortResults: boolean[] = [];
-            grouped.forEach(groupMember => {
-              sortResults.push(
+            return grouped
+              .map(groupMember =>
                 checkAlphaSort(groupMember, order as AlphabeticalOrder),
-              );
-            });
-            return sortResults.some(result => !result);
+              )
+              .some(result => !result);
           }
         } else if (hasAlphaSort) {
           return checkAlphaSort(memberSet, order as AlphabeticalOrder);
--- a/packages/eslint-plugin/src/rules/member-ordering.ts
+++ b/packages/eslint-plugin/src/rules/member-ordering.ts
@@ -643,27 +643,26 @@ function getRank(
  * @returns The array of groups of members.
  */
 function groupMembersByType(
-  memberSet: Member[],
-  memberType: MemberType[],
+  members: Member[],
+  memberTypes: MemberType[],
   supportsModifiers: boolean,
 ): Member[][] {
   const groupedMembers: Member[][] = [];
+  const memberRanks = members.map(member =>
+    getRank(member, memberTypes, supportsModifiers),
+  );
   let previousRank: number | undefined = undefined;
-  memberSet.forEach((member, index) => {
-    if (index === memberSet.length - 1) {
+  members.forEach((member, index) => {
+    if (index === members.length - 1) {
       return;
     }
-    const rankOfcurrentMember = getRank(member, memberType, supportsModifiers);
-    const rankOfNextMember = getRank(
-      memberSet[index + 1],
-      memberType,
-      supportsModifiers,
-    );
-    if (rankOfcurrentMember === previousRank) {
+    const rankOfCurrentMember = memberRanks[index];
+    const rankOfNextMember = memberRanks[index + 1];
+    if (rankOfCurrentMember === previousRank) {
       groupedMembers.at(-1)?.push(member);
-    } else if (rankOfcurrentMember === rankOfNextMember) {
+    } else if (rankOfCurrentMember === rankOfNextMember) {
       groupedMembers.push([member]);
-      previousRank = rankOfcurrentMember;
+      previousRank = rankOfCurrentMember;
     }
   });
   return groupedMembers;
@@ -996,8 +995,8 @@ export default createRule<Options, MessageIds>({
         const hasAlphaSort = !!(order && order !== 'as-written');
         if (hasAlphaSort && Array.isArray(memberTypes)) {
           groupMembersByType(memberSet, memberTypes, supportsModifiers).forEach(
-            groupMember => {
-              checkAlphaSort(groupMember, order as AlphabeticalOrder);
+            groupMembers => {
+              checkAlphaSort(groupMembers, order as AlphabeticalOrder);
             },
           );
         }
--- a/packages/eslint-plugin/src/rules/member-ordering.ts
+++ b/packages/eslint-plugin/src/rules/member-ordering.ts
@@ -621,17 +621,17 @@ function getRank(
  * ```
  * interface Foo {
       [a: string]: number;
- 
+
       a: x;
       B: x;
       c: x;
- 
+
       c(): void;
       B(): void;
       a(): void;
- 
+
       (): Baz;
- 
+
       new (): Bar;
     }
  * ```
@@ -647,16 +647,7 @@ function groupMembersByType(
   supportsModifiers: boolean,
 ): Member[][] {
   const groupedMembers: Member[][] = [];
-  const getPreviousRank = (): number | undefined => {
-    if (groupedMembers.at(-1)?.at(0) === undefined) {
-      return undefined;
-    }
-    return getRank(
-      groupedMembers[groupedMembers.length - 1][0],
-      memberType,
-      supportsModifiers,
-    );
-  };
+  let previousRank: number | undefined = undefined;
   memberSet.forEach((member, index) => {
     if (index === memberSet.length - 1) {
       return;
@@ -667,10 +658,11 @@ function groupMembersByType(
       memberType,
       supportsModifiers,
     );
-    if (rankOfcurrentMember === getPreviousRank()) {
+    if (rankOfcurrentMember === previousRank) {
       groupedMembers.at(-1)?.push(member);
     } else if (rankOfcurrentMember === rankOfNextMember) {
       groupedMembers.push([member]);
+      previousRank = rankOfcurrentMember;
     }
   });
   return groupedMembers;
--- a/packages/eslint-plugin/src/rules/member-ordering.ts
+++ b/packages/eslint-plugin/src/rules/member-ordering.ts
@@ -658,15 +658,15 @@ function groupMembersByType(
     );
   };
   memberSet.forEach((member, index) => {
+    if (index === memberSet.length - 1) {
+      return;
+    }
     const rankOfcurrentMember = getRank(member, memberType, supportsModifiers);
     const rankOfNextMember = getRank(
       memberSet[index + 1],
       memberType,
       supportsModifiers,
     );
-    if (index === memberSet.length - 1) {
-      return;
-    }
     if (rankOfcurrentMember === getPreviousRank()) {
       groupedMembers.at(-1)?.push(member);
     } else if (rankOfcurrentMember === rankOfNextMember) {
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -947,20 +991,24 @@ export default createRule<Options, MessageIds>({
...
-              groupMember =>
-                !checkAlphaSort(groupMember, order as AlphabeticalOrder),
-            );
+            const sortResults: boolean[] = [];
+            grouped.forEach(groupMember => {
+              sortResults.push(
+                checkAlphaSort(groupMember, order as AlphabeticalOrder),
+              );
+            });
+            return sortResults.some(result => !result);
```
JoshuaKGoldberg, line 1005: [Style] No need to create a separate array and then `.forEach`; `.map` works the same.

```suggestion
            return grouped
              .map(groupMember =>
                checkAlphaSort(groupMember, order as AlphabeticalOrder),
              )
              .some(result => !result);
```

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -615,6 +615,39 @@ function getRank(
   return getRankOrder(memberGroups, orderConfig);
 }
 
+function changeToGroups(
```
JoshuaKGoldberg, line 618: [Style] I hard a really hard time reading through this function and how it's used. Which is not really this PR's fault; `member-ordering` is an annoyingly complex rule by definition. But to make it easier for future us, I think this function would be more understandable with:

* A name like `groupMembers` or `groupMembersBy(something)`, where that _something_ is what they're actually being grouped on... `Type`?
* A JSDoc comment explaining what it does

WDYT?

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -615,6 +615,39 @@ function getRank(
...
+  memberSet: Member[],
+  memberType: MemberType[],
+  supportsModifiers: boolean,
+): Member[][] {
+  const groupedMembers: Member[][] = [];
+
+  const getPreviousRank = (): number | undefined => {
```
JoshuaKGoldberg, line 625: [Refactor] This function is constantly calling `getRank` and `.at(...)` to recompute things. But I think they can be cached in variables:

* Instead of calling `getRank(member, memberType, supportsModifiers)` twice, store in a variable
* Instead of a dynamic `getPreviousRank`, have a nullable state variable like `let previousGroupAndRank: { group: Member[]; rank: number } | undefined;`

Does that make sense?

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -615,6 +615,39 @@ function getRank(
...
+      supportsModifiers,
+    );
+  };
+
+  memberSet.forEach((member, index) => {
+    if (getRank(member, memberType, supportsModifiers) === getPreviousRank()) {
+      groupedMembers.at(-1)?.push(member);
+    } else if (index === memberSet.length - 1) {
+      return;
+    } else if (
```
JoshuaKGoldberg, line 641: [Refactor] Bailing out early is always nice to save time & make exit cases more clear. No unit tests fail when I try this:

```suggestion
    if (index === memberSet.length - 1) {
      return;
    }

    if (getRank(member, memberType, supportsModifiers) === getPreviousRank()) {
      groupedMembers.at(-1)?.push(member);
    } else if (
```

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -615,6 +615,39 @@ function getRank(
...
+      getRank(member, memberType, supportsModifiers) ===
+      getRank(memberSet.at(index + 1)!, memberType, supportsModifiers)
```
JoshuaKGoldberg, line 643: [Style] No need to `.at`. The types for it are always nullable, but since we know it's defined, we can just `[index + 1]`.

```suggestion
      getRank(memberSet[index + 1], memberType, supportsModifiers)
```

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -934,6 +967,17 @@ export default createRule<Options, MessageIds>({
       let memberTypes: MemberType[] | string | undefined;
       let optionalityOrder: OptionalityOrder | undefined;
 
+      const runSort = (memberSet: Member[]): undefined => {
```
JoshuaKGoldberg, line 970: [Style] Similarly, this name kind of confused me. Everything here can arguably be referred to as "run(ning a) sort". Maybe ... `checkAlphaSortForAllMembers`? `runSortForAlphaGroup`? What would you say this function is really trying to do?

JoshuaKGoldberg (changes requested):
Seems reasonable on a read-through, but I'm having a hard time understanding the added code. Which means future folks with less context would probably have a harder time. What do you think about the clarification + cleanup comments?

Btw, emphasizing from a comment: this rule is inherently pretty tricky and complex. So I think any changes will be that way too. Nice job getting it to work! 👏

JoshuaKGoldberg:
> fixes: #8100
> 
> ## PR Checklist
> ...

Btw, no need to put the `fixes: #...` on top of the PR description. It's in the checklist. 🙂

packages/eslint-plugin/src/rules/member-ordering.ts
arka1002: Thank you so much for the reviews.

But, can you explain to me the 2nd point a little bit more ? I can't seem to understand that.

JoshuaKGoldberg: Ah sorry for not being clear!

What I'm looking at: right now, `getPreviousRank()` is being called in each iteration of `memberSet.forEach(...)`. It's essentially recomputing a value each time it's called, right? Where that "value" is the previous rank.

Problem (I think, unless I've misinterpreted): Wasn't that rank computed for the previous member? If, say, we're at `forEach` index 3, then `getPreviousRank()` will try to return the rank for the member at index 2. But didn't we compute the rank for member at index 2 in the previous iteration of the `forEach` loop?

My proposal: instead of recomputing the previous rank, storing it in a variable. That way the code is a little more clear and the act of computing a member's rank isn't done so many times per member.

JoshuaKGoldberg: Ah sorry for not being clear!

What I'm looking at: right now, `getPreviousRank()` is being called in each iteration of `memberSet.forEach(...)`. It's essentially recomputing a value each time it's called, right? Where that "value" is the previous rank.

Problem (I think, unless I've misinterpreted): Wasn't that rank computed for the previous member? If, say, we're at `forEach` index 3, then `getPreviousRank()` will try to return the rank for the member at index 2. But didn't we compute the rank for member at index 2 in the previous iteration of the `forEach` loop?

My proposal: instead of recomputing the previous rank, storing it in a variable. That way the code is a little more clear and the act of computing a member's rank isn't done so many times per member.

Does that make sense?

arka1002: > My proposal: instead of recomputing the previous rank, storing it in a variable. That way the code is a little more clear and the act of computing a member's rank isn't done so many times per member.

Yes, thanks :raised_hands:

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -934,6 +987,21 @@ export default createRule<Options, MessageIds>({
...
 
+      /**
+       * It runs an alphabetic sort on the groups of the members of the class in the source code.
+       * @param memberSet The members in the class of the source code on which the grouping operation will be performed.
+       */
+      const checkAlphaSortForAllMembers = (memberSet: Member[]): undefined => {
+        const hasAlphaSort = !!(order && order !== 'as-written');
+        if (hasAlphaSort && Array.isArray(memberTypes)) {
+          groupMembersByType(memberSet, memberTypes, supportsModifiers).forEach(
+            groupMember => {
```
JoshuaKGoldberg, line 998: [Naming] "Group member" is singular. But then this is passed to `checkAlphaSort`, which calls it "members" (plural). It's a little confusing to have the style of name for something change when it's used.

How about, maybe?:

```suggestion
            members => {
```

...or failing that, `groupMembers`?

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -615,6 +615,59 @@ function getRank(
...
+    }
+ * ```
+  The resulting array will look like this - [[a, B, c], [c, B, a]].
+ * @param memberSet The members in the class of the source code on which the grouping operation will be performed.
+ * @param memberType The member type order listed in the eslint config file.
+ * @param supportsModifiers it'll get passed to getRank()
+ * @returns the array of groups of members
+ */
+function groupMembersByType(
+  memberSet: Member[],
```
JoshuaKGoldberg, line 645: [Naming] "Set" is normally used to refer to the actual `Set` object, or sometimes the general concept of a unique set of things. So it's a little confusing to see it written here.

The parameter passed to `checkAlphaSortForAllMembers` is called `members`. How about using that as the name? Or `allMembers` if you find `members` to be too broad? Something like those?

```suggestion
  members: Member[],
```

I do see that this is an existing name from the `checkOrder` function. Which came from a PR I approved. Now I know better. 😄

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -615,6 +615,59 @@ function getRank(
...
+      c(): void;
+      B(): void;
+      a(): void;
+ 
+      (): Baz;
+ 
+      new (): Bar;
+    }
+ * ```
+  The resulting array will look like this - [[a, B, c], [c, B, a]].
```
JoshuaKGoldberg, line 638: [Praise] This example & surrounding text was super helpful, thank you very much for writing it! 🙌

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -615,6 +615,59 @@ function getRank(
...
+ * @param memberSet The members in the class of the source code on which the grouping operation will be performed.
+ * @param memberType The member type order listed in the eslint config file.
+ * @param supportsModifiers it'll get passed to getRank()
+ * @returns the array of groups of members
+ */
```
JoshuaKGoldberg, line 643: [Docs] Applying a few touchups...

* Accuracy: "eslint config file" isn't accurate, because the rule might be configured in another way - such as an inline comment
* Accuracy: this might be in more than just classes
* Consistency in whether the `*` is used in the example's code block
* Consistency in whether sentences are capitalized & have ending punctuation
* Consistency in indentation with the rest of the file
* Trimming down the words a bit

```suggestion
/**
 * Groups members into arrays of consecutive members with the same rank.
 * If, for example, the memberSet parameter looks like the following...
 * @example
 * ```
 * interface Foo {
 *   [a: string]: number;
 *
 *   a: x;
 *   B: x;
 *   c: x;
 *
 *   c(): void;
 *   B(): void;
 *   a(): void;
 *
 *   (): Baz;
 *
 *   new (): Bar;
 * }
 * ```
 * ...the resulting array will look like: [[a, B, c], [c, B, a]].
 * @param memberSet The members to be grouped.
 * @param memberType The configured order of member types.
 * @param supportsModifiers It'll get passed to getRank().
 * @returns The array of groups of members.
 */
```

This set of suggestions applies to to the `checkAlphaSortForAllMembers` JSDoc tags too.

How does that make you feel?

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -615,6 +615,59 @@ function getRank(
...
+function groupMembersByType(
+  memberSet: Member[],
+  memberType: MemberType[],
```
JoshuaKGoldberg, line 646: [Naming] And for this one, it's passed in as `memberTypes: MemberType[]` (plural) but here is referred to as `memberType: MemberType[]` (singular). Is there a reason to change?

If not:

```suggestion
  memberTypes: MemberType[],
```

JoshuaKGoldberg (changes requested):
⚡ this is looking really snazzy! I'm happy with how it's coming along, and think I understand it a lot better now thanks to the new structures & comments.

Mostly requesting changes on some small touchups. I think this is getting pretty close to merge-ready!

Guy wearing a tie shooting blue lightning out of his hands, calmly

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -947,20 +1015,22 @@ export default createRule<Options, MessageIds>({
...
           if (hasAlphaSort) {
-            return !grouped.some(
-              groupMember =>
-                !checkAlphaSort(groupMember, order as AlphabeticalOrder),
-            );
+            return grouped
+              .map(groupMember =>
+                checkAlphaSort(groupMember, order as AlphabeticalOrder),
+              )
+              .some(result => !result);
```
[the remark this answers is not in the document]
arka1002: Yeah, using `.some` is a part of the problem of the issue. Currently before this pr, the rule only run alphabetical sort on a particular group, if that group was sorted, then only it checked the next group, else the next group remains unchecked.

An example test - https://github.com/typescript-eslint/typescript-eslint/pull/8263/files#diff-28e99227546efaad4ed38d47dcd8746943b60b09b2095a68ec4209856d709e1fR128-R129

But yeah, I removed `.some`, I think it isn't necessary.

JoshuaKGoldberg (approved):
🙌 Whoop! This looks wonderful - thanks for working with me on it. 

Since I've looked at it so much, I'll leave it open as `1 approval` in the hopes another team member can get to it.

auvred (approved):
The changes looks great! :rocket:

flohw:
Just updated and tested. This is a huge improvement to have all the errors at once.

Thank you again for the fix and reviews! :slightly_smiling_face:

PR #8282, opened 2024-01-21
fix(eslint-plugin): [no-redundant-type-constituents] incorrectly marks & string as redundant

### LINKED ISSUE #7580: Bug: `no-redundant-type-constituents` incorrectly marks `& string` as redundant

### Playground Link

https://typescript-eslint.io/play/#ts=5.2.2&fileType=.ts&code=LAKApgHgDg9gTgFwAQIJ5TEgggGxzAdwGcA5AVwFsAjMOAaTFSKQF4kBvUJJOMAQwAmMAHY5USANoBjIkQAKcGBkSoSfCpgCWwpAGtGMAGZIAwgGULaHGAAiYKTj5w%2BCTSIC6AfgBcpi2atbe0dnVxFpWQUlWjQ1DXcAblAAXySQUEhYRBR0TBtNIj48QlJKGnpGZjZOEG4AejqkABUACwKkdr4kKkFmswBRIhxtZFpFOCRAXg3AUp2uJAa5hqQASQQAcmYEFsxDIqJMWCJNVwA3TAAKXkFfKjIAcw6dLcxh4WQ4MmsASgAaRcaBGAKCIiAhQtoHi1CChtkhBAJjm4dEYkAADABkSFBcAhqI6zAAVMIYAgCTwwAIyMIBHw3gA6OZXISicQReSKZSxdRaHT6VAo8yWVDWOwOJwuJFITHYiFeXyCgLCoJi8HhGTs6IqOJgRIpUAZaDwZBSEHIFp8IhYcjUWgMcTVOYABl8a0mgAkdmHtPn4uFIYRlWhrP4gVL68CG7Im4SgpAETSAu1EXy4fDEa3lROsJDmy3p22MNKgJatdqdJB3GAwAR9JBjeBIQCg5EW6s3TCD47QIbGwGszrXoMMpMcxH6nIoCBSkEICE9YXzW2gMEgUaj8oVimmAxUmHiEDBlyy4VIpGAoMgZcI7kQGelw1ljabR3Bx4nfGuiqnSjbt1VsxarVudppEAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1ukQBNYmvAIYtaRUrTIcmyfJXyxELVBkiJo0DtEjgwAXxB6gA&tsconfig=&tokens=false

### Repro Code

```typescript
export type AllowsNumberKeys = {
  readonly [cssPropertyName in keyof CSSStyleDeclaration]?: CSSStyleDeclaration[cssPropertyName];
};

export type DisallowsNumberKeys = {
  // This is a bad TSEslint error 💥
  //
  // It's the false positive (read: bug in the lint rule),
  // demonstrating how the addition of `& string` is *not* redundant.
  readonly [cssPropertyName in keyof CSSStyleDeclaration & string]?: CSSStyleDeclaration[cssPropertyName];
}

export const hasANumberKey = {
  0: '👈 this key is a number',
};

export const wideKeys: AllowsNumberKeys = hasANumberKey;

// This is a good TS error ✅
//
// Considering we've explicitly narrowed down the key
// type of `DisallowsNumberKeys` to only accept strings.
export const narrowKeys: DisallowsNumberKeys = hasANumberKey;
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/no-redundant-type-constituents": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {}
}
```

### Expected Result

No lint error, considering that adding `& string` to the key type is *not* redundant.

### Actual Result

A false positive lint error claiming that the addition of `& string` is redundant, which it is *not*.

### Additional Info

Hello 👋

I searched and found no apparent overlap with any previous issue. Apologies if I missed anything and this turns out to be a duplicate after all.

arka1002:
fixes: #7580

## PR Checklist

- [x] Addresses an existing open issue: fixes #7580 


packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -323,8 +327,58 @@ export default createRule({
...
+          typePartFlags.length >= 2
+            ? seenUnionTypes.set(typeNode, typePartFlags)
+            : null;
+        }
+        const checkIfUnionsAreAssignable = (): undefined => {
+          let typeFlagsOfUnions: ts.TypeFlags[] = [];
+          let result = true;
+          let primitiveUnit: number;
+
+          seenUnionTypes.forEach((value, key) => {
```
JoshuaKGoldberg, line 340: [Style] This code is already kind of complex, let's stick with informative names.

```suggestion
          seenUnionTypes.forEach((unionTypeFlags, unionTypeNode) => {
```

(or some names with more meaning - I'm not sold on those)

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -323,8 +327,58 @@ export default createRule({
...
           }
-        }
 
+          typePartFlags.length >= 2
+            ? seenUnionTypes.set(typeNode, typePartFlags)
+            : null;
+        }
+        const checkIfUnionsAreAssignable = (): undefined => {
+          let typeFlagsOfUnions: ts.TypeFlags[] = [];
+          let result = true;
```
JoshuaKGoldberg, line 337: [Style] This code is already kind of complex, let's stick with informative names.

```suggestion
          let hadPrimitiveOverridden = true;
```

(note: a later comment also suggests removing this variable)

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -323,8 +327,58 @@ export default createRule({
...
+                  primitive:
+                    primitiveTypeFlagNames[
+                      primitiveUnit as keyof typeof primitiveTypeFlagNames
+                    ],
+                },
+                messageId: 'primitiveOverridden',
+                node: key,
+              });
+            }
+            typeFlagsOfUnions = [];
```
JoshuaKGoldberg, line 375: [Refactor] Unit tests still pass when this is removed. Any particular reason to keep it around?

```suggestion
```

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -323,8 +327,58 @@ export default createRule({
...
+        }
+        const checkIfUnionsAreAssignable = (): undefined => {
+          let typeFlagsOfUnions: ts.TypeFlags[] = [];
+          let result = true;
+          let primitiveUnit: number;
+
+          seenUnionTypes.forEach((value, key) => {
+            value.forEach(union => {
+              typeFlagsOfUnions.push(union.typeFlags);
+            });
```
JoshuaKGoldberg, line 343: [Style] Is there a need to have a shared variable outside the `forEach`? Switching it to being inside the `.forEach` and never modified doesn't fail any unit tests:

```suggestion
            const typeFlagsOfUnions = unionTypeFlags.flatMap(
              union =>
                union.typeFlags as keyof typeof literalToPrimitiveTypeFlags,
            );
```

That includes an extra tip: since we know the type flags are all the `keyof typeof literalToPrimitiveTypeFlags`, using the assertion early on means we don't need multiple assertions later.

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -323,8 +327,58 @@ export default createRule({
...
+                )
+              ) {
+                result = true;
+                primitiveUnit =
+                  literalToPrimitiveTypeFlags[
+                    iterator as keyof typeof literalToPrimitiveTypeFlags
+                  ];
+              } else {
+                result = false;
+                break;
```
JoshuaKGoldberg, line 359: [Testing] Removing this `break;` doesn't fail any unit tests. Either there's a gap in test coverage or this is dead code that can be deleted.

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -323,8 +327,58 @@ export default createRule({
...
-        }
 
+          typePartFlags.length >= 2
+            ? seenUnionTypes.set(typeNode, typePartFlags)
+            : null;
+        }
+        const checkIfUnionsAreAssignable = (): undefined => {
+          let typeFlagsOfUnions: ts.TypeFlags[] = [];
+          let result = true;
+          let primitiveUnit: number;
```
JoshuaKGoldberg, line 338: [Refactor] Right now this function stores two pieces of state: `result` (`boolean`) and `primitiveUnit` (`number`). If one is truthy, the other must be. That's a bit of a code smell for duplicate state.

Tests all still pass locally if I remove `result` altogether, set `primitiveUnit`'s type to `number | undefined`, and change from `if (result)` to `if (primitiveUnit)`.

Also, there's no need to declare `primitiveUnit` outside the `forEach`. Each iteration of the function has its own idea of what the primitive unit should be.

Proposal: delete `result` and move `primitiveUnit` inside the `forEach`?

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -323,8 +327,58 @@ export default createRule({
...
+                  literalToPrimitiveTypeFlags[
+                    iterator as keyof typeof literalToPrimitiveTypeFlags
+                  ],
+                )
+              ) {
+                result = true;
+                primitiveUnit =
+                  literalToPrimitiveTypeFlags[
+                    iterator as keyof typeof literalToPrimitiveTypeFlags
+                  ];
```
JoshuaKGoldberg, line 356: [Question] Here, `primitiveUnit` is set to its primitive type flags number. But then in the `context.report` call, it's used as a key for the type flag names. It seems a bit 'extra' to me to have two places where lookups are being done. How about unifying and renaming to a more succinct name like `primitive`?

```suggestion
                primitive = literalToPrimitiveTypeFlags[iterator];
```

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
JoshuaKGoldberg, line 340: [Style] The existing code uses `for...of` loops. This new code uses `.forEach`. Is there a strong reason to switch the style over?

If not, please stick with the existing style in a file/project. It's hard to mentally map between two. 😅

JoshuaKGoldberg (changes requested):
Ok! This rule is trickier than I remembered it being 😅 nice job adding to its logic.

I think I understand what the new code is doing, but it was a little hard to read through. A couple general suggestions:

* Try to stick with the existing conventions in code unless there's a strong reason to deviate.
* Try to use informative names when possible, not general ones like `key` or `iterator`

The core logic seems reasonable to me. Requesting changes on refactors. There might be a point where more tests would be reasonable depending on one or two of them?

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
arka1002: I removed the variable, I dont think we need it anymore.  
I might need some help with the type assertions though :sweat_smile:

bradzacher (approved):

PR #8287, opened 2024-01-22
chore(type-utils): reuse newly added "is builtin symbol like" logic

### LINKED ISSUE #8234: Repo: reuse newly added "is builtin symbol like" logic from `type-utils`
### Suggestion

Followup for https://github.com/typescript-eslint/typescript-eslint/pull/8011#discussion_r1417737352

We've added few utility functions to the `packages/type-utils/src/builtinSymbolLikes.ts`

Basically they contain the logic for determining whether a given type is a type from the TS default library or not

These files already contain similar logic, it'd be cool if we dedupe it!

https://github.com/typescript-eslint/typescript-eslint/blob/705370ac0d9c54081657b8855b398e57d6ea4ddb/packages/eslint-plugin/src/rules/no-implied-eval.ts#L82-L90 https://github.com/typescript-eslint/typescript-eslint/blob/705370ac0d9c54081657b8855b398e57d6ea4ddb/packages/eslint-plugin/src/rules/no-implied-eval.ts#L145-L153

Let's label this issue as `good first issue`? Ref: https://github.com/typescript-eslint/typescript-eslint/pull/8011#discussion_r1438375603

arka1002:
fixes: #8234

## PR Checklist

- [x] Addresses an existing open issue: fixes #8234 

### DIFF
--- a/packages/eslint-plugin/src/rules/no-implied-eval.ts
+++ b/packages/eslint-plugin/src/rules/no-implied-eval.ts
@@ -1,11 +1,10 @@
-import { isFunctionLike } from '@typescript-eslint/type-utils';
 import type { TSESTree } from '@typescript-eslint/utils';
 import { AST_NODE_TYPES } from '@typescript-eslint/utils';
 import { getScope } from '@typescript-eslint/utils/eslint-utils';
 import * as tsutils from 'ts-api-utils';
 import * as ts from 'typescript';
 
-import { createRule, getParserServices } from '../util';
+import { createRule, getParserServices, isFunctionSimilar } from '../util';
 
 const FUNCTION_CONSTRUCTOR = 'Function';
 const GLOBAL_CANDIDATES = new Set(['global', 'window', 'globalThis']);
@@ -79,9 +78,8 @@ export default createRule({
         return true;
       }
 
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
-      if (symbol && symbol.escapedName === FUNCTION_CONSTRUCTOR) {
-        return isFunctionLike(services.program, type);
+      if (isFunctionSimilar(services.program, type)) {
+        return true;
       }
 
       const signatures = checker.getSignaturesOfType(
@@ -138,7 +136,7 @@ export default createRule({
         const type = services.getTypeAtLocation(node.callee);
         const symbol = type.getSymbol();
         if (symbol) {
-          if (isFunctionLike(services.program, type)) {
+          if (isFunctionSimilar(services.program, type)) {
             context.report({ node, messageId: 'noFunctionConstructor' });
           }
         } else {
--- a/packages/eslint-plugin/src/rules/no-implied-eval.ts
+++ b/packages/eslint-plugin/src/rules/no-implied-eval.ts
@@ -1,6 +1,7 @@
 import type { TSESTree } from '@typescript-eslint/utils';
 import { AST_NODE_TYPES } from '@typescript-eslint/utils';
 import { getScope } from '@typescript-eslint/utils/eslint-utils';
+import { isFunctionLike } from '@typescript-eslint/type-utils';
 import * as tsutils from 'ts-api-utils';
 import * as ts from 'typescript';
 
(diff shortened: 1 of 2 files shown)


packages/type-utils/src/builtinSymbolLikes.ts
```diff
@@ -69,6 +69,18 @@ export function isReadonlyTypeLike(
     );
   });
 }
+/**
+ * let F = new Function("foo");
+ *      ^ FunctionLike
+ * let I = (callback: Function) => {}
+ *           ^ FunctionLike
+ */
+export function isFunctionLike(program: ts.Program, type: ts.Type): boolean {
```
auvred, line 78: `typescript` package already exports the function with same name `isFunctionLike`. Having two functions with the same name may cause confusion.

packages/type-utils/src/builtinSymbolLikes.ts
```diff
@@ -69,6 +69,18 @@ export function isReadonlyTypeLike(
...
+  return (
+    isBuiltinSymbolLike(program, type, 'Function') ||
+    isBuiltinSymbolLike(program, type, 'FunctionConstructor')
```
auvred, line 81: The original rule didn't check for symbols with name `FunctionConstructor`. Should we add tests that contain symbols with name `FunctionConstructor`? But that is probably not within the scope of this pr. IMO we shouldn't change the rule logic in this pr

packages/eslint-plugin/src/rules/no-implied-eval.ts
```diff
@@ -80,13 +81,7 @@ export default createRule({
...
       if (symbol && symbol.escapedName === FUNCTION_CONSTRUCTOR) {
-        const declarations = symbol.getDeclarations() ?? [];
-        for (const declaration of declarations) {
-          const sourceFile = declaration.getSourceFile();
-          if (services.program.isSourceFileDefaultLibrary(sourceFile)) {
-            return true;
-          }
-        }
+        return isFunctionLike(services.program, type);
       }
```
auvred, line 85: `isFunctionLike` internally already checks that the type has a symbol, and that symbol is named `Function`, so the condition above is redundant.

packages/eslint-plugin/src/rules/no-implied-eval.ts
```diff
@@ -1,3 +1,4 @@
+import { isFunctionLike } from '@typescript-eslint/type-utils';
```
auvred, line 1: `@typescript-eslint/type-utils` shouldn't be imported directly. This function should probably be imported from `src/utils`, as other rules do

https://github.com/typescript-eslint/typescript-eslint/blob/920f909debc3eb394f5aef4e36568bdf6641e304/packages/eslint-plugin/src/util/index.ts#L18-L19

auvred (changes requested):

packages/type-utils/src/builtinSymbolLikes.ts
auvred: You're right `Function` here has `FunctionConstructor` type. I missed this because in the source code of rule both symbol name and callee identifier's name are compared to `FUNCTION_CONSTRUCTOR`

Then +1 on renaming `isFunctionLike`... Not sure what's the best name for it. Maybe `isFunctionTypeLike` or something like this?

Also we should consider that #8094 most likely will introduce changes to `isBuiltinSymbolLike`.
As per https://github.com/typescript-eslint/typescript-eslint/pull/8094#discussion_r1449120581:
`isBuiltinSymbolLike(program, type, 'Function') || isBuiltinSymbolLike(program, type, 'FunctionConstructor')` recursively visits all subtypes twice, we can optimize to not do the same job twice. #8094 does the following:

```ts
isBuiltinSymbolLike(
  services.program,
  calleeType,
  name => name === 'PromiseConstructor' || name === 'Promise'
)
```

Perhaps we can wait until #8094 is merged and then do the same

arka1002: Yes, sure.

JoshuaKGoldberg: > But that is probably not within the scope of this pr. IMO we shouldn't change the rule logic in this pr

Yeah this is just a refactor (`chore`). If you want to change how it works that'd be a separate issue.

👍 I'll mark this as blocked on #8094. Thanks :)

JoshuaKGoldberg: #8094 is closed - so we can now pretend it doesn't exist. cc @arka1002, I'm un-blocking this PR. Sorry for the long wait!

JoshuaKGoldberg:
👋 ping @arka1002, just checking in - is this PR still something you have the time and energy for?

arka1002:
Closing it 👍

PR #8296, opened 2024-01-24
chore: add `typescript-eslint` package scaffold

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: #7694

## Overview

This PR just adds an private scaffold for the `typescript-eslint` package being added in #7935. I've split this out so that the automation will bump the versions with each release.

The most annoying part of the branch is every time I go back to it and rebase it on `main` I have to manually resolve and correct all of the versions or else things will use old releases from npm instead of the repo code.

### DIFF
--- a/packages/typescript-eslint/package.json
+++ b/packages/typescript-eslint/package.json
@@ -1,0 +1,83 @@
+{
+  "name": "typescript-eslint",
+  "version": "6.20.0",
+  "description": "Tooling which enables you to use TypeScript with ESLint",
+  "files": [
+    "dist",
+    "_ts4.3",
+    "README.md",
+    "LICENSE"
+  ],
+  "type": "commonjs",
+  "private": true,
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./package.json": "./package.json"
+  },
+  "engines": {
+    "node": "^16.0.0 || >=18.0.0"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/typescript-eslint/typescript-eslint.git",
+    "directory": "packages/typescript-eslint"
+  },
+  "bugs": {
+    "url": "https://github.com/typescript-eslint/typescript-eslint/issues"
+  },
+  "license": "MIT",
+  "keywords": [
+    "ast",
+    "ecmascript",
+    "javascript",
+    "typescript",
+    "parser",
+    "syntax",
+    "eslint",
+    "eslintplugin",
+    "eslint-plugin"
+  ],
+  "scripts": {
+    "build": "tsc -b tsconfig.build.json",
+    "postbuild": "downlevel-dts dist _ts4.3/dist --to=4.3",
+    "clean": "tsc -b tsconfig.build.json --clean",
+    "postclean": "rimraf dist && rimraf _ts4.3 && rimraf coverage",
+    "format": "prettier --write \"./**/*.{ts,mts,cts,tsx,js,mjs,cjs,jsx,json,md,css}\" --ignore-path ../../.prettierignore",
+    "lint": "nx lint",
+    "test": "jest --coverage --passWithNoTests",
+    "typecheck": "tsc -p tsconfig.json --noEmit"
+  },
+  "peerDependencies": {
+    "eslint": "^7.0.0 || ^8.0.0"
+  },
+  "dependencies": {
+    "@typescript-eslint/eslint-plugin": "6.20.0",
+    "@typescript-eslint/parser": "6.20.0"
+  },
+  "devDependencies": {
+    "downlevel-dts": "*",
+    "jest": "29.7.0",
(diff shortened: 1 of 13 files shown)


merceyz:
> I have to manually resolve and correct all of the versions or else things will use old releases from npm instead of the repo code.

You can replace the explicit versions with `workspace:*` and Yarn will replace them on publish.

bradzacher:
@merceyz oh that's interesting! I didn't know that existed.

I had noticed that there was the syntax when I looked at the lock file (you can see I used it in the root package.json) - but I didn't realise that it would automatically get replaced on publish! That's super cool!

Cc @JamesHenry - we use the nx packages for publishing - does nx use `yarn npm publish` under the hood or does it not? I.e. Could we use this syntax? Could nx support it?

And for an FYI @JoshuaKGoldberg

JamesHenry:
@bradzacher Right now there is a fork in the road when it comes to non semver versions.

Either:

- you apply your versions using valid semver version in your source files, tracked by git

OR

- you can leverage either semver versions of any valid workspace/file protocol specifiers if version your _dist_ files, not tracked by git

This second option is how the Nx repo itself does it. The source of truth for versions in this case is therefore not disk, but rather the npm registry. Nx release supports this case via config in nx.json.

---

## Background

This fundamental choice is down to the implementation of `nx release` being based on isolated version/changelog/publish steps, and the fact that we want to publish something tangible on disk.

Some release tools read your files but then perform all kinds of modifications in memory which are not reflected anywhere on your file system for you to see/play with and then publish that in memory object to npm. This can make troubleshooting bad publishes nearly impossible at times - with this approach it's trivial, because you can always `cd` somewhere and run `npm publish` directly with whatever options you want.

So when you perform your versioning step, `nx release` gets your files ready for your publishing step on disk, therefore it needs to replace those workspace/file references with the real numbers so that they work for your consumers.

Now, having said all that...

When we first created this hard line we only had the three subcommands available: `nx release version`, `nx release changelog`, `nx release publish` and therefore was not straightforward way to plumb things through from one to the other.

These days we have the programmatic API for `nx release` as a first class use-case - it's already what we use in typescript-eslint, and so it is therefore possible for me to imagine exposing a util function on the programmatic API to reset workspaces/file references after running `releasePublish`. The main complication though will be not committing those package.json dependency updates as part of the versioning or changelog step. I think it is doable though.

This is the issue to track for progress: https://github.com/nrwl/nx/issues/20978

packages/core/package.json
```diff
@@ -0,0 +1,83 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/typescript-eslint/typescript-eslint.git",
+    "directory": "packages/core"
+  },
+  "bugs": {
+    "url": "https://github.com/typescript-eslint/typescript-eslint/issues"
+  },
+  "license": "BSD-2-Clause",
```
JoshuaKGoldberg, line 31:
```suggestion
  "license": "MIT",
```

packages/core/README.md
```diff
@@ -0,0 +1,12 @@
+# `@typescript-eslint/core`
+
+> Tooling which enables you to use TypeScript with ESLint
+
+[![NPM Version](https://img.shields.io/npm/v/@typescript-eslint/core.svg?style=flat-square)](https://www.npmjs.com/package/@typescript-eslint/core)
+[![NPM Downloads](https://img.shields.io/npm/dm/@typescript-eslint/core.svg?style=flat-square)](https://www.npmjs.com/package/@typescript-eslint/core)
+
+👉 See **https://typescript-eslint.io/packages/core** for documentation on this package.
```
JoshuaKGoldberg, line 8: Missing?

JoshuaKGoldberg (changes requested):
LGTM to start, just requesting changes on the license and missing docs. 🚀

packages/core/README.md
bradzacher: Coming with the next PR.
The package is private for now so nobody shall see the broken links.

JoshuaKGoldberg (dismissed):
🚢

JamesHenry (dismissed):
Let’s finalize the discord discussion before merging

packages/typescript-eslint/package.json
```diff
@@ -0,0 +1,83 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/typescript-eslint/typescript-eslint.git",
+    "directory": "packages/typescript-eslint"
+  },
+  "bugs": {
+    "url": "https://github.com/typescript-eslint/typescript-eslint/issues"
+  },
+  "license": "BSD-2-Clause",
```
JoshuaKGoldberg, line 31:
```suggestion
  "license": "MIT",
```

bradzacher: arg that slipped back in

JoshuaKGoldberg (approved):
Cartoon happy yellow bird bouncing up and down above the word 'YASS'

PR #8297, opened 2024-01-24
feat(website): add ESQuery filter to URL state

### LINKED ISSUE #8174: Website: Also add ESQuery filter state to URL

### Suggested Changes

It would occasionally be useful to be able to send a playground share that includes the ESTree > ESQuery filter in its URL. For example, in https://github.com/typescript-eslint/typescript-eslint/issues/3729#issuecomment-1874432897 I would have liked to be able to send a playground link with the exact query for what was being asked. [This playground link includes the code snippet](https://typescript-eslint.io/play/#ts=5.0.4&showAST=es&fileType=.tsx&code=CYUwxgNghgTiAEkoGdnwMIB4BqA%2BeA3gFDzwC2IALgBYD2wmAKgDTwCquAFAA6xRkAueIwCUQtgG4S8bjFrchmAApcR8ALz4lUgL5A&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false) but doesn't preserve the `MethodDefinition[value.typeParameters]` ESQuery filter.

### Affected URL(s)

StyleShit:
Closes #8174

Here us [a working example](https://deploy-preview-8297--typescript-eslint.netlify.app/play/#ts=5.3.3&showAST=es&fileType=.tsx&esQuerySelector=N4IgLgngDgpiBcIDGB7AtlFBXAdgExABoQBnGAGxiTBQCcSEBtUSWBEASzxhzA4DMOMWkRAA3AIbkscRAFkYYABYo8AERiCcHPihwgAvoRbRZICWDC0OAIyxg4xHBLRnJ0mADpWMAAoTaF0VhBgMAXQMgA&code=FAEwpgxgNghgTmABNGBnViDCAeAagPkQG9hFEBbMAFwAsB7EbAFQBpEBVfACgAd4ZyALkRMAlMPYBuUoh5w6PYdgAK3UYgC8hZdIC%2BQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&esQueryFilter=MethodDefinition%5Bvalue.typeParameters%5D&tokens=false)

### DIFF
--- a/packages/website/src/components/options.ts
+++ b/packages/website/src/components/options.ts
@@ -44,6 +44,4 @@ export const defaultConfig: ConfigModel = {
   }),
   scroll: true,
   showTokens: false,
-  esQueryFilter: '',
-  esQuerySelector: undefined,
 };
--- a/packages/website/src/components/options.ts
+++ b/packages/website/src/components/options.ts
@@ -44,4 +44,6 @@ export const defaultConfig: ConfigModel = {
   }),
   scroll: true,
   showTokens: false,
+  esQueryFilter: '',
+  esQuerySelector: undefined,
 };
--- a/packages/website/src/components/ESQueryFilter.tsx
+++ b/packages/website/src/components/ESQueryFilter.tsx
@@ -5,20 +5,22 @@ import styles from './ESQueryFilter.module.css';
 import Text from './inputs/Text';
 
 export interface ESQueryFilterProps {
-  readonly onChange: (value: Selector) => void;
+  readonly onChange: (filter: string, selector: Selector) => void;
   readonly onError: (value?: Error) => void;
+  defaultValue?: string;
 }
 
 export function ESQueryFilter({
   onChange,
   onError,
+  defaultValue,
 }: ESQueryFilterProps): React.JSX.Element {
-  const [value, setValue] = useState('');
+  const [value, setValue] = useState(defaultValue ?? '');
   const changeEvent = (value: string): void => {
     setValue(value);
     try {
       const queryParsed = window.esquery.parse(value);
-      onChange(queryParsed);
+      onChange(value, queryParsed);
       onError(undefined);
     } catch (e: unknown) {
       onError(e instanceof Error ? e : new Error(String(e)));
--- a/packages/website/src/components/Playground.tsx
+++ b/packages/website/src/components/Playground.tsx
@@ -2,7 +2,6 @@
 /* eslint-disable @typescript-eslint/no-unnecessary-condition */
 import { useWindowSize } from '@docusaurus/theme-common';
 import clsx from 'clsx';
-import type * as ESQuery from 'esquery';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import {
   type ImperativePanelHandle,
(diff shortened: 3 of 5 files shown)


packages/website/src/components/options.ts
```diff
@@ -44,4 +44,6 @@ export const defaultConfig: ConfigModel = {
   }),
   scroll: true,
   showTokens: false,
+  esQueryFilter: '',
+  esQuerySelector: undefined,
```
StyleShit, line 48: I don't really like the fact that I have 2 variables that basically hold the same state (string vs selector object) instead of having one of them, and the other will be a derived state.
I wasn't able to find an ESQuery method that converts a selector to a string, seems to be a one-way ticket. It means that I can save only the string and then parse it to a selector. I tried 2 approaches:

1. Parse when reading the initial state from the URL - Didn't work because `window.esquery` isn't available yet
2. Passing the string as a default value to the `ESQueryFilter` input and parse when on mount (using the same logic of the `onChange` handler) - Looked ugly and felt wrong (`useEffect` for such cases always feels to me like a code smell)

Any idea?

JoshuaKGoldberg: Yeah +1 to not having two variables. Also +1 that `useEffect` feels like a code smell for this case. Tricky... 🤔 

The first thought that comes to my head is changing the state to be `string | undefined | ESQueryFilter`. It'd be `string | undefined` when first initialized, then turn into `ESQueryFilter` later... I don't love that. Seems kind of convoluted.

Second thought: how about ignoring the return of `window.esquery.parse(value)` inside `ESQueryFilter` and using a second `window.esquery.parse(filter)` inside `ASTViewer`'s `tryToApplyFilter`? It's a duplicate parse but removes a state...

Third thought: this also feels like a reasonable feature request on ESQuery. Up for filing an issue there? 😄 

Fourth thought: maybe instead of a duplicate parse, the state should be an object like `{ filter: string; selector?: ESQuery.Selector } | undefined`? In my mind that's a better representation of what the state truly is.

JoshuaKGoldberg (changes requested):
Looks great, works great - nice! 🙌 

Requesting changes on the refactor in thread. I like where you're going with this. Even if the result is just filing a feature request on ESQuery & adding a comment in code.

packages/website/src/components/options.ts
JoshuaKGoldberg: > Do you mean that instead of passing a selector we'll now pass a filter and the AST viewer will handle the parsing?

Exactly, yeah! Either this or option 4 make me less unhappy.

JoshuaKGoldberg (approved):
LGTM, thanks!

I'm not super picky about the website code. @armano2 and @Josh-Cena  know more about it than me anyway. Tagging them in case they have a chance to review in the next week, and setting a reminder to merge next Monday morning if not. Cheers!

PR #8312, opened 2024-01-29
chore(eslint-plugin): [no-invalid-void-type] fix `Options` typing to reflect `minItems: 1`

kazarmy:
## PR Checklist

- [ ] Addresses an existing open issue

## Overview

This PR fixes the `Options` typing for the `no-invalid-void-type` rule so that it reflects
https://github.com/typescript-eslint/typescript-eslint/blob/70505e45bee0c2b05777a3f9e5b886a1a607d349/packages/eslint-plugin/src/rules/no-invalid-void-type.ts#L51
and also [its documentation](https://typescript-eslint.io/rules/no-invalid-void-type/#options). I think this should be considered an extremely minor typo even though it doesn't involve documentation.

:sweat_smile:

### DIFF
--- a/packages/eslint-plugin/src/rules/no-invalid-void-type.ts
+++ b/packages/eslint-plugin/src/rules/no-invalid-void-type.ts
@@ -5,7 +5,7 @@ import { getSourceCode } from '@typescript-eslint/utils/eslint-utils';
 import { createRule } from '../util';
 
 interface Options {
-  allowInGenericTypeArguments?: string[] | boolean;
+  allowInGenericTypeArguments?: [string, ...string[]] | boolean;
   allowAsThisParameter?: boolean;
 }
 


packages/eslint-plugin/src/rules/no-invalid-void-type.ts
```diff
@@ -5,7 +5,7 @@ import { getSourceCode } from '@typescript-eslint/utils/eslint-utils';
 import { createRule } from '../util';
 
 interface Options {
-  allowInGenericTypeArguments?: string[] | boolean;
+  allowInGenericTypeArguments?: [string, ...string[]] | boolean;
```
JoshuaKGoldberg, line 8: I suppose this makes sense technically 😄. But working with types like this can be irksome since you have to do it consistently. Is there an end-user reason for this change?

I'm not opposed to it, just curious how it came up.

kazarmy: Oh hi 😄.

It's true that this isn't a random PR out of nowhere. I was reviewing my `typescript-eslint` rule typings (I have only 4 of them and they are for a no-deadline personal project, which means that there's room for everything + the kitchen sink in it) and originally I had something like the following:

```ts
'@typescript-eslint/no-invalid-void-type'?: Linter.RuleEntry<[{
  allowInGenericTypeArguments?: string[] | boolean;
  allowAsThisParameter?: boolean;
}]>;
```

I checked [`no-invalid-void-type`'s documentation](https://typescript-eslint.io/rules/no-invalid-void-type/#options) and noticed that the typing for `allowInGenericTypeArguments` had tightened up. For some reason (probably because I wanted to find out whether I can get rule typings directly from `typescript-eslint` source) I checked the rule source and noticed immediately that the `Options` typing for it was out-of-date, largely because it's at the top of the file I guess.

> Is there an end-user reason for this change?

I figured that the out-of-date typing would affect DX for `typescript-eslint` itself, but after creating a test script in `packages/eslint-plugin/src` (of the `typescript-eslint` repo) and doing:

```ts
import rules from './rules';
const aVar = rules['no-invalid-void-type'].create(/* ??? */);
```

I felt I was in a bit too deep to be comfortable, so instead of figuring out how the `create` method fits in the whole scheme of things, and then doing a "Complex Bug With a Reproduction" issue with my fork of `typescript-eslint` as the repo, I figured that I'd just take my chances and go straight to a PR.

JoshuaKGoldberg: Cool, thanks. It looks like this is the only `minItems: 1` in a non-deprecated rule, so I guess you're setting a good precedent. Thanks 😄 .

JoshuaKGoldberg: Got it, makes sense. It looks like this is the only `minItems: 1` in a non-deprecated rule, so I guess you're setting a good precedent. Nice 😄

JoshuaKGoldberg (approved):
> considered an extremely minor typo even though it doesn't involve documentation

Yeah, agreed. Since it's the only one of its kind I suppose it doesn't hurt to try it out.

Thanks for sending in!

😅

kazarmy:
Thanks for merging!

:tada:

Issue #4733, opened 2022-03-25
Stale version of interface from another file being used on second linter run with caching disabled


kowsen:
- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.
- [x] I have [read the FAQ](https://typescript-eslint.io/docs/linting/troubleshooting) and my problem is not listed.

**Repro**

If you run `node index.js` on this stackblitz and follow the instructions it prints, you'll run into the error: [https://stackblitz.com/edit/node-skp6gk?file=index.js](https://stackblitz.com/edit/node-skp6gk?file=index.js)

Generally, I run into this when I have one file containing an interface with a property and another file that references that property. Then following these steps gets me a "@typescript-eslint/no-unnecessary-type-assertion" error on step 3:

1. Run esLint.lintFiles(["code_file_path.ts", "interface_file_path.ts"])
2. Make the property on the interface optional in interface_file_path.ts and add a non-null assertion to where we reference it in code_file_path.ts
3. Run the same command from step 1

If I then kill the node process that I'm running the linter from and restart it, it doesn't throw the error on step 1. Additionally, if I swap the order that the files go into the linter that seems to fix it too.

**Expected Result**

I expect that without any caching enabled, there shouldn't be a case where running eslint.lintFiles on a file a second time would give me an error based on the state of the files when they were passed into eslint.lintFiles the first time.

**Actual Result**

Even after disabling any kind of ESLint caching, following the steps in the repro case causes a linter error even though the passed in files are correctly formatted.

**Additional Info**

Sorry if this is on the wrong repo, I don't understand enough about the inner workings to be sure this is the right place.

I ran into this using ESLintWebpackPlugin, which uses webpack's compilation hooks to determine what files have changed during a rebuild and passes an array of their absolute paths into eslint.lintFiles. Before opening a bug with them to see if there's a way to order this list to put dependencies before files they depend on, I want to make sure that the lintFiles function having a required order for files that are passed into it is deliberate.

**Versions**

| package                     | version |
| --------------------------- | ------- |
| `@typescript-eslint/parser` | `5.16.0` |
| `TypeScript`                | `4.6.3` |
| `ESLint`                    | `8.11.0` |
| `node`                      | `14.19.0` |

bradzacher:
> I expect that without any caching enabled

We actually do do caching! We have to because for any non-trivial sized project it will take way too long to recalculate the state for the entire project.

[...]

Issue #4694, opened 2022-03-17
Docs: explain whether type-aware linting is safe to use with `eslint --cache`


six-transit:
### Suggested Changes

I'd like to use ESLint's `--cache` functionality with typescript-eslint -- but [according to the docs](https://eslint.org/docs/user-guide/command-line-interface#caching), the latter performs caching on a per-file basis. That suggests to me that it's incompatible with type-aware linting, since I might make a change in one file that would create a new lint in a different file -- one that might not have changed since the previous ESLint run.

[typescript-eslint.io](https://typescript-eslint.io) doesn't have any info about caching that I've been able to find. So I guess my suggestion for improving the docs would be:

- if ESLint caching is incompatible with type-aware linting, that should be noted in the docs, since it's a fairly significant caveat.
- if ESLint caching _is_ compatible with type-aware linting, that seems like a pretty worthwhile thing to note in the docs as a way to speed up checks, maybe in the "Linting with Type Information" or FAQ page.

### Affected URL(s)

https://typescript-eslint.io/docs/linting/type-linting/ and/or https://typescript-eslint.io/docs/linting/troubleshooting/

bradzacher:
From experience - in general ESLint caching can be a real problem across the entire ecosystem.

Unless every single lint rule you use *only* depends on single-file analysis, you can quickly end up in an inconsistent lint state.
A few quick examples off the top of my head:
- there are a number of rules in `eslint-plugin-import` all do their own cross-file analysis (not using TS).
- `eslint-plugin-prettier` depends on your `.prettierrc` file.
- At Meta we also have a number of internal lint rules that depend on code loaded from source and also on file locations.

In general I would just warn against using the cache - you don't really know what external metadata your plugins may or may not depend on and the cache could cause undefined behaviour.

six-transit:
Thanks for the quick response! Yeah, I was wondering whether `eslint-plugin-import` would also be affected by issues.

Anyway, that's too bad -- maybe eventually we'll see an RFC in ESLint for more robust expression of dependencies that could be used for cache keying? (Poking around I see you've already commented in https://github.com/eslint/eslint/issues/14139 and https://github.com/eslint/eslint/issues/15583.)

bradzacher:
`--cache` is a pretty advanced usecase / optimisation anyway - most people wouldn't even know the flag existed!

TBH I don't think I've ever even seen the flag turned on in a codebase (eslint themselves don't even use it!).

[...]

PR #8322, opened 2024-01-30
feat: speed up non-type-aware linting with project service

### LINKED ISSUE #8427: ⚡️ Performance: Don't open client files unnecessarily for project service

### Relevant Package

typescript-estree

### Description

As reported by @yepitschunked in #8322 (thanks! ❤️‍🔥):

> The current implementation opens all files using the project service, even if the file is not opted into type-aware linting (with the project or programs options). 
> This makes linting slower than necessary, since project for every file will be loaded.

I like the idea of not calling [`service.openClientFile`](https://github.com/typescript-eslint/typescript-eslint/blob/5a1e85da65cb83bc4e02965f8eb8f1f51347e004/packages/typescript-estree/src/useProgramFromProjectService.ts#L19) unnecessarily. `useProgramFromProjectService` is indeed called always if `parseSettings.EXPERIMENTAL_useProjectService` is true:

https://github.com/typescript-eslint/typescript-eslint/blob/5a1e85da65cb83bc4e02965f8eb8f1f51347e004/packages/typescript-estree/src/parser.ts#L48-L53

Logs in #8424 show that opening files with a warmed-up project service to take a few ms per file. That can really add up in projects with many files.

https://github.com/typescript-eslint/typescript-eslint/pull/8322#issuecomment-1937075984 shows that avoiding calling `service.openClientFile` on files that don't ask for type checking can speed up service-type-checked linting to be on par with traditional type-checked linting. ⚡ 

### Additional Info

Note that I'm not yet convinced that #8322's starting approach is the correct one. That approach only calls `openClientFile` on files with type-checked linting enabled. But some users enable type information on all files despite only using it in some. And, heck, type-checked rules only use type checking conditionally - and try to avoid calling it if they can bail out early. So gating `openClientFile` on whether a _file_ wants type information can still allow files through the gate unnecessarily.

An alternative possibility might be to avoid opening the client file altogether until `getParserServices` is called within a lint rule. I think that's the best possible strategy for avoiding unnecessary `openClientFile` calls. But that might require moving all the type generation logic to `getParserServices`... which might be good in general? Investigation required.

yepitschunked:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8427

## Overview

This PR fixes two issues with the experimental project service option. 

**Faster non-type-aware linting**
The current implementation opens all files using the project service, even if the file is not opted into type-aware linting (with the `project` or `programs` options). This makes linting slower than necessary, since project for every file will be loaded. Of course, you could also just omit the `EXPERIMENTAL_useProjectService` option, but I think that:

- this option should eventually become the default, so it won't be configurable
- enabling project service on non-type-aware linting also helps interop with type-aware linting (see next section)

This PR skips opening files with the project service if `hasFullTypeInformation` is false, matching the existing behavior. 

**Refresh non-type-aware files with project service**
Due to the performance tradeoffs of type-aware linting, we currently only enable it on specific files/folders in our repo. However, this means that Typescript is never informed of changes to files outside of those locations (due to typescript-eslint hijacking the watch mechanism). We end up linting with stale type information if we depend on any types that are defined outside of these locations. 

When opening non-type-aware files, this PR checks if the project service is already aware of this file
- If so, this means this file is part of a program used in type-aware linting, and we should open it with the project service to update it
- If not, then we can fall back to the fast path (described in the previous section).  

This approach ensures that if you *only* have non-type-aware files opened in the IDE, linting never creates a program and remains fast. If you open a mixture of files, then the project service is still informed of changes to non-type-aware files. There's of course some reliance on ordering here, but this is already a known/wontfix problem (e.g https://github.com/typescript-eslint/typescript-eslint/issues/4733), and shouldn't affect IDE use cases.


packages/typescript-estree/src/parseSettings/createParseSettings.ts
```diff
@@ -127,6 +127,7 @@ export function createParseSettings(
           DEFAULT_TSCONFIG_CACHE_DURATION_SECONDS,
     )),
     tsconfigRootDir,
+    typeAware: !!options.programs || !!options.project,
```
yepitschunked, line 130: I noticed that `hasFullTypeInformation` was checking whether `programs` or `projects` were empty. However, this will always be the case when using the project service. The docs seem to point to using `programs` and `project` (singular) as the way to enable type-aware linting, so I decided to make this explicit here.

JoshuaKGoldberg:
👋 thanks for sending this over @yepitschunked! I see from your GitHub history that you have a good history of sending nice paper cut / performance fixes to projects. Exciting that you're looking at typescript-eslint. 😄 

But:

> - [] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

We ask that people file issues for non-trivial changes like this one for a few reasons:

* The templates ask reporters to provide info we'd need to triage requests, including a full reproduction - which is also useful if we later have to spelunk through the repository's history
* Sometimes the suggested changes are a duplicate of existing issues or PRs, and/or end up splitting into multiple issues
    * https://github.com/typescript-eslint/typescript-eslint/pull/8322/files#r1471992155 gives the impression there's a bug to be filed about always thinking `hasFullTypeInformation` is true?
* We want to raise visibility for the changes to let folks have a chance to discuss nuances before work is done
* Even if you're at the 'expert' level where specifically you know to check for duplicates _and_ understand the codebase well enough to make these changes, most users aren't - so it sets a negative precedent

Do you have the time to file an issue?

For visibility, https://antfu.me/posts/why-reproductions-are-required is a good read on why I'm asking for a reproduction in particular.

JoshuaKGoldberg:
Note that I tried this out on #8424 and it seemed to nicely fix Sentry's `time yarn lint:js`:

* Not using project service: `✨  Done in 35.34s. \n yarn lint:js  50.11s user 1.47s system 145% cpu 35.472 total`
* Without this change: `✨  Done in 57.70s. \n yarn lint:js  78.63s user 5.09s system 144% cpu 57.814 total`
* With this change: `✨  Done in 36.23s. \n yarn lint:js  50.69s user 1.53s system 143% cpu 36.354 total`

Very nicely done @yepitschunked! I need to do some more research to figure out whether there's a more root-level fix to be done, but my gut says the work here won't go to waste.

JoshuaKGoldberg:
> Refresh non-type-aware files with project service
> There's of course some reliance on ordering here, but this is already a known/wontfix problem (e.g https://github.com/typescript-eslint/typescript-eslint/issues/4733), and shouldn't affect IDE use cases.

For non-IDE use cases, we definitely would need a repro. I'm not familiar with Webpack plugins, etc. enough to evaluate this. Note that `eslint --cache` is not something we recommend: #4694. (it's not mentioned here, just posting as reference)

For IDE use cases: tracked for VS Code in https://github.com/microsoft/vscode-eslint/issues/1774. (again just posting as reference)

packages/typescript-estree/src/create-program/createProjectProgram.ts
```diff
@@ -52,10 +52,10 @@ function createProjectProgram(
   );
   const describedPrograms =
     relativeProjects.length === 1
-      ? relativeProjects[0]
+      ? ` ${relativeProjects[0]}`
       : `\n${relativeProjects.map(project => `- ${project}`).join('\n')}`;
   const errorLines = [
-    `ESLint was configured to run on \`${describedFilePath}\` using \`parserOptions.project\`: ${describedPrograms}`,
+    `ESLint was configured to run on \`${describedFilePath}\` using \`parserOptions.project\`:${describedPrograms}`,
```
JoshuaKGoldberg, line 58: Not strictly needed for this PR, but it was _so annoying_ having Prettier auto-save the snapshot to no longer have a trailing slash...

packages/typescript-estree/src/parser.ts
```diff
@@ -195,11 +195,10 @@ function parseAndGenerateServices<T extends TSESTreeOptions = TSESTreeOptions>(
...
   }
 
-  /**
-   * Generate a full ts.Program or offer provided instances in order to be able to provide parser services, such as type-checking
-   */
   const hasFullTypeInformation =
-    parseSettings.programs != null || parseSettings.projects.size > 0;
+    parseSettings.programs != null ||
+    parseSettings.projects.size > 0 ||
+    !!parseSettings.projectService;
```
JoshuaKGoldberg, line 201: @yepitschunked following up on `parserSettings.typeAware`: whether the file is being parsed with type information is still inferable from existing `parseSettings` properties. So rather than refactor to add a _new_ property, this just corrects the logic for `const hasFullTypeInformation`.

packages/typescript-estree/src/useProgramFromProjectService.ts
JoshuaKGoldberg, line 1: @yepitschunked I ended up refactoring this file a good bit to move secondary logic into functions. It got rather difficult to read through once I added in handling for the new "default project".

packages/typescript-estree/tests/lib/useProgramFromProjectService.test.ts
```diff
@@ -58,7 +68,44 @@ const createProjectServiceSettings = <
...
+
+// if hasFullTypeInformation is false and isDefaultProjectAllowed is true:
+//  don't openClientFileFromProjectService
+//  ✅ did use the isDefaultProjectAllowed boolean - in the if (!... && !...) { createNoProgramWithProjectService check
+//  do
+
+// if hasFullTypeInformation is false and isDefaultProjectAllowed is false:
+//  ✅ did use the isDefaultProjectAllowed boolean - in the if (!... && !...) { createNoProgramWithProjectService check
+// do createNoProgramWithProjectService
+//  - which still opens the file if service.getScriptInfo knows of it
```
JoshuaKGoldberg, line 83:
```suggestion
```

packages/typescript-estree/tests/lib/useProgramFromProjectService.test.ts
```diff
@@ -58,7 +68,44 @@ const createProjectServiceSettings = <
...
+
 describe('useProgramFromProjectService', () => {
+  it('creates a standalone AST with no program when hasFullTypeInformation is false', () => {
```
JoshuaKGoldberg, line 86:
```suggestion
  it('creates a standalone AST with no program when hasFullTypeInformation is false and allowDefaultProject is falsy', () => {
```

packages/typescript-estree/src/useProgramFromProjectService.ts
```diff
@@ -42,70 +47,107 @@ const updateExtraFileExtensions = (
...
-      }
-    } else if (!isDefaultProjectAllowedPath) {
       throw new Error(
-        `${parseSettings.filePath} was not found by the project service. Consider either including it in the tsconfig.json or including it in allowDefaultProject.`,
+        `Too many files (>${serviceSettings.maximumDefaultProjectFileMatchCount}) have matched the default project.${DEFAULT_PROJECT_FILES_ERROR_EXPLANATION}
+Matching files:
+${filesToPrint.map(file => `- ${file}`).join('\n')}
+${truncatedFileCount ? `...and ${truncatedFileCount} more files\n` : ''}
+If you absolutely need more files included, set parserOptions.projectService.maximumDefaultProjectFileMatchCount_THIS_WILL_SLOW_DOWN_LINTING to a larger value.
+`,
```
JoshuaKGoldberg, line 109: Note that I've moved this _"Too many files ..."_ notice to _inside_ the `if (opened.configFileName)` in this PR. That should stop the notice from getting logged on all files!

JoshuaKGoldberg:
cc @bradzacher - I think this is ready for re-review.

bradzacher (approved):

PR #8331, opened 2024-02-01
feat: export plugin metadata

bradzacher:
## Overview

Splitting out of #7935 to focus that PR and reduce its line count.

Adds plugin metadata which is a requirement for flat config files.
https://eslint.org/docs/latest/extend/plugin-migration-flat-config#adding-plugin-meta-information

I also added some `satisfies` around the place to help enforce that our previously anonymous objects actually match the shapes we have defined.


packages/utils/src/ts-eslint/Linter.ts
```diff
@@ -262,6 +269,13 @@ namespace Linter {
     parserOptions?: ParserOptions;
   }
 
+  // TODO - RuleCreateFunction is no longer supported in ESLint v9
```
JoshuaKGoldberg, line 272:
```suggestion
  // TODO - AnyRuleCreateFunction is no longer supported in ESLint v9
```

JoshuaKGoldberg (approved):
🚀

PR #8265, opened 2024-01-17
fix(eslint-plugin): [no-unused-vars] don't report on types referenced in export assignment expression

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6383
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Explicitly visit `TSExportAssignment`'s expression, so if it contains some type, that type gets a new reference

PR #8332, opened 2024-02-01
fix(eslint-plugin): [consistent-type-imports] dont report on types used in export assignment expressions

### LINKED ISSUE #8330: Bug: [consistent-type-imports] erroneous report on type used in an `export =` assertion

### Playground Link

n/a - the broken commit hasn't released yet so it doesn't repro on playground

### Repro Code

```TypeScript
import type Foo from 'foo';

export = {} as Foo;
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/consistent-type-imports": "error",
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

No error

### Actual Result

Type import "Foo" is used by decorator metadata

### Additional Info

Looks like it was introduced by #8265

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8330

## Overview

I missed that in https://github.com/typescript-eslint/typescript-eslint/pull/8265 :

In cases like `export = {} as T`, T was referenced as `isTypeReference: true` and `isValueReference: true`. Now it's only referenced as a type

### DIFF
--- a/packages/eslint-plugin/src/rules/consistent-type-imports.ts
+++ b/packages/eslint-plugin/src/rules/consistent-type-imports.ts
@@ -183,12 +183,15 @@ export default createRule<Options, MessageIds>({
                        * keep origin import kind when export
                        * export { Type }
                        * export default Type;
+                       * export = Type;
                        */
                       if (
                         ref.identifier.parent.type ===
                           AST_NODE_TYPES.ExportSpecifier ||
                         ref.identifier.parent.type ===
-                          AST_NODE_TYPES.ExportDefaultDeclaration
+                          AST_NODE_TYPES.ExportDefaultDeclaration ||
+                        ref.identifier.parent.type ===
+                          AST_NODE_TYPES.TSExportAssignment
                       ) {
                         if (ref.isValueReference && ref.isTypeReference) {
                           return node.importKind === 'type';
--- a/packages/eslint-plugin/tests/rules/consistent-type-imports.test.ts
+++ b/packages/eslint-plugin/tests/rules/consistent-type-imports.test.ts
@@ -566,6 +566,22 @@ export type Y = {
   [constants.X]: ReadonlyArray<string>;
 };
     `,
+    `
+      import A from 'foo';
+      export = A;
+    `,
+    `
+      import type A from 'foo';
+      export = A;
+    `,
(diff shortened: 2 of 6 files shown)


bradzacher (changes requested):
Thanks for hopping on this so fast!

I believe here should be no change required to the rule here - just a fix to the scope analysis. The tests are correct to add though!

Given the export equals declaration accepts an expression - we don't want to drop down into the export visitor. We just need a special case for the one special case - the identifier case. All other cases can be handled as a normal expression would.

In the identifier case we can immediately create a dual-reference. We know it's just this case because both `T[U]` and `T.U` are semantic errors from TS if `T` is a type, or `T.U` is a type (any syntactically those are the two valid ways to "chain" off a type reference).

packages/eslint-plugin/src/rules/consistent-type-imports.ts
```diff
@@ -183,12 +183,15 @@ export default createRule<Options, MessageIds>({
...
+                       * export = Type;
                        */
                       if (
                         ref.identifier.parent.type ===
                           AST_NODE_TYPES.ExportSpecifier ||
                         ref.identifier.parent.type ===
-                          AST_NODE_TYPES.ExportDefaultDeclaration
+                          AST_NODE_TYPES.ExportDefaultDeclaration ||
+                        ref.identifier.parent.type ===
+                          AST_NODE_TYPES.TSExportAssignment
```
auvred, line 194: Thanks for the review!

---

> I believe here should be no change required to the rule here - just a fix to the scope analysis. The tests are correct to add though!

If we don't check `TSExportAssignment` here, then cases such as

```ts
import type A from 'foo';
export = A;
```

will result in false positives. From my understanding, export assignment here behaves like default export (exports type/value as is).

Please correct me if I'm wrong

bradzacher: ah you're right.
I'd just assumed that the rule would handle these dual-type-value references and not enforce anything for the import - I didn't remember that it was special-cased.

bradzacher (approved):
brilliant - thanks!

PR #8334, opened 2024-02-01
chore(eslint-plugin): deprecate no-var-requires in favor of no-require-imports

### LINKED ISSUE #8092: Repo: deprecate no-var-requires in favor of no-require-imports
### Suggestion

I was discussing this with @bradzacher and @JoshuaKGoldberg.

`no-var-requires` is a strict subset of `no-require-imports`: the former only detects `require()` expressions in very limited contexts (including variable declarations, new expressions, call expressions, member expressions). The latter bans all `require` expressions and `import = require` declarations.

They were both imported from other plugins:

- `no-require-imports` comes from TSLint: https://github.com/typescript-eslint/typescript-eslint/pull/199
- `no-var-requires` comes from eslint-plugin-typescript: https://github.com/bradzacher/eslint-plugin-typescript/pull/114 => https://github.com/typescript-eslint/typescript-eslint/commit/470c2648ef861d36e081e4351c5281e951fe0e47 which in turn comes from TSLint!

Why TSLint has two rules remains unknown, but for the purpose of our maintenance I don't see why we need both. Let's implement `no-var-requires` as part of `no-require-imports` (an option called `allowAsImport`) instead.

arka1002:
BREAKING CHANGE:
Deprecates a previously-recommended rule.

## PR Checklist

- [x] Addresses an existing open issue: fixes #8092 


arka1002:
Hi, I just have one doubt, just to make sure I understood the problem correctly 

In `no-require-imports` if we set `allowAsImport` as `true`, then all the test cases in `no-var-requires` will pass right ??

example -

```ts
/* eslint "@typescript-eslint/no-require-imports": ["error", { "allowAsImport": true } ] */
import foo = require("foo"); // no error (??)
```

packages/eslint-plugin/tests/rules/no-require-imports.test.ts
JoshuaKGoldberg, line 1: [Testing] A couple of gaps:
* The rule's source includes logic around scope analysis (good!) but the tests don't. Could you add cases for user-defined `require` please? (`function require(...)`, etc.)
* The rule mentions Call, Member, New, and As Expressions, as well as type assertions and variable declarators. Each should have a valid & invalid case.

packages/eslint-plugin/docs/rules/no-require-imports.md
```diff
@@ -50,6 +50,30 @@ console.log(require('../data.json').version);
 console.log(require('../package.json').version);
 ```
 
+### `allowAsImport`
+
+A boolean. If it's set to `true`, the require statements with an `import` declaration won't be reported. The default value is `false`. If you want to apply the `no-var-requires` rule, then set it to `true`.
```
JoshuaKGoldberg, line 55: [Docs] Two thoughts here:

* We know it's a `boolean` with a default value of `false` from the auto-generated code block earlier in the file.
* Soon, `no-var-requires` will be a distant memory. And these rules have a lot of overlap anyway. So IMO there's no need to explicitly mention it.

```suggestion
If it's set to `true`, the require statements with an `import` declaration won't be reported.
This is useful if you're on an older CommonJS-style codebase and haven't been able to transition to ESM-style imports yet.
```

packages/eslint-plugin/docs/rules/no-require-imports.md
```diff
@@ -32,7 +32,7 @@ import * as lib3 from 'lib3';
 
 ### `allow`
 
-A array of strings. These strings will be compiled into regular expressions with the `u` flag and be used to test against the imported path. A common use case is to allow importing `package.json`. This is because `package.json` commonly lives outside of the TS root directory, so statically importing it would lead to root directory conflicts, especially with `resolveJsonModule` enabled. You can also use it to allow importing any JSON if your environment doesn't support JSON modules, or use it for other cases where `import` statements cannot work.
+An array of strings. These strings will be compiled into regular expressions with the `u` flag and be used to test against the imported path. A common use case is to allow importing `package.json`. This is because `package.json` commonly lives outside of the TS root directory, so statically importing it would lead to root directory conflicts, especially with `resolveJsonModule` enabled. You can also use it to allow importing any JSON if your environment doesn't support JSON modules, or use it for other cases where `import` statements cannot work.
```
JoshuaKGoldberg, line 35: [Praise] Nice catch 😄

packages/eslint-plugin/src/rules/no-var-requires.ts
```diff
@@ -14,6 +14,7 @@ type MessageIds = 'noVarReqs';
 export default createRule<Options, MessageIds>({
   name: 'no-var-requires',
   meta: {
+    deprecated: true,
```
JoshuaKGoldberg, line 17: [Bug] Deprecating `no-var-requires` should include removing it from `recommended`. We don't want to be in the business of recommending deprecated rules!

This PR should also add `no-require-imports` to `recommended`.

JoshuaKGoldberg (changes requested):
Mostly solid! A bit around config recommendations & unit tests, and this'll be good to go...

...once ~v7~ v8 is ready to go. Which will be in a few months. So this'll have to wait 😅. But thanks in the interim!

packages/eslint-plugin/tests/rules/no-require-imports.test.ts
arka1002: Thank you so much for the reviews !

I just had a small question - 
> should have a valid

Would a valid case look like - 
```ts
/* eslint "@typescript-eslint/no-require-imports": ["error", { "allowAsImport": false } ] */
const foo: Foo = require('./foo.json').default; // no error ( ?? )
```
This case normally is invalid in `no-require-imports` though.

arka1002: Thank you so much for the reviews!!

I just had a small doubt - 
> should have a valid

Will this be valid - 
```ts
/* eslint "@typescript-eslint/no-require-imports": ["error", { "allowAsImport": true } ] */
import foo: Foo = require('./foo.json').default; // no error ??

```

JoshuaKGoldberg: Oh, I was looking at these: https://github.com/typescript-eslint/typescript-eslint/blob/0ab1f789a155d2b94161b1e0ea9eb26e65b41cc6/packages/eslint-plugin/src/rules/no-require-imports.ts#L74-L79

For each: if they can't be hit by unit tests, then they don't need to be in the source code. In other words: each of those is either testable or dead code. Does that make sense?

Josh-Cena: @arka1002 You are referring to the `import x = require()` syntax which is not a function call; it's just a special import syntax. You want to test places where `require()` is an actual function.

arka1002: No no
Sorry, if i was unable to express correctly.

@JoshuaKGoldberg 
Those lines are getting used .... in the invalid tests. I was talking about the valid test cases. I just want to know - what valid test cases can I write for those lines.

JoshuaKGoldberg: Got it - a good strategy for mapping invalid tests to valid is to take the same syntax and make a modification that should turn it into something the rule allows. In this case, adding a `declare function require(path: string): unknown;` on top should make them pass.

JoshuaKGoldberg: (I haven't thought deeply on this - there might be other things to do)

JoshuaKGoldberg (changes requested):
Progress!

Note that I retargeted this against the `v8` branch.

packages/eslint-plugin/src/rules/no-var-requires.ts
```diff
@@ -13,10 +13,10 @@ type MessageIds = 'noVarReqs';
...
```
armano2, line 16: do you mind adding missing meta field, `replacedBy`
https://eslint.org/docs/latest/extend/custom-rules#rule-structure

packages/eslint-plugin/tests/rules/no-require-imports.test.ts
```diff
@@ -47,6 +47,62 @@ require('remark-preset-prettier');
       code: "import pkg = require('some-package');",
       options: [{ allow: ['^some-package$'] }],
     },
+    {
+      code: "import foo = require('foo');",
+      options: [{ allowAsImport: true }],
+    },
+    {
+      code: 'require(foo);',
```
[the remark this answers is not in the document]
arka1002: added as invalid, :+1:, https://github.com/typescript-eslint/typescript-eslint/pull/8334/files#diff-a5ad891733aae70415d2717222585dcee37d24848333afcc15a9d4e4d73b251eR354-R371

eslint.config.mjs
```diff
@@ -169,6 +169,12 @@ export default tseslint.config(
           ignorePrimitives: true,
         },
       ],
+      '@typescript-eslint/no-require-imports': [
+        'error',
+        {
+          allow: ['/package\\.json$'],
```
arka1002, line 175: @JoshuaKGoldberg I just wanted to ask a question.

The recent lint errors point to - 
1. https://github.com/typescript-eslint/typescript-eslint/blob/59bbf417702b53fe78d8a190da7054bc6c22b780/packages/website/sidebars/sidebar.rules.js#L1
2. https://github.com/typescript-eslint/typescript-eslint/blob/59bbf417702b53fe78d8a190da7054bc6c22b780/packages/website/src/clientModules.js#L4
  
& a few other places.

Should I add the regexes in this file 
or
should I add an eslint comment directive ?

JoshuaKGoldberg: I'd say the inline config comments strategy. Config files in `packages/website` are pretty different from the rest of the codebase and tend to do some wacky things.

packages/eslint-plugin/docs/rules/no-require-imports.mdx
```diff
@@ -59,6 +59,34 @@ console.log(require('../package.json').version);
...
+
+If it's set to `true`, the require statements with an `import` declaration won't be reported.
+This is useful if you're on an older CommonJS-style codebase and haven't been able to transition to ESM-style imports yet.
+
+With `{allowAsImport: true}`:
+
+<Tabs>
+<TabItem value="❌ Incorrect">
+
+```ts
```
auvred, line 72:
```suggestion
```ts option='{ "allowAsImport": true }'
```

It looks like `option` is missing here?

packages/eslint-plugin/docs/rules/no-require-imports.mdx
```diff
@@ -59,6 +59,34 @@ console.log(require('../package.json').version);
...
+var foo = require('foo');
+const foo = require('foo');
+let foo = require('foo');
+```
+
+</TabItem>
+<TabItem value="✅ Correct">
+
+```ts
```
auvred, line 81:
```suggestion
```ts option='{ "allowAsImport": true }'
```

And here as well

packages/eslint-plugin/docs/rules/no-var-requires.mdx
```diff
@@ -9,6 +9,12 @@ import TabItem from '@theme/TabItem';
 >
 > See **https://typescript-eslint.io/rules/no-var-requires** for documentation.
 
+:::danger Deprecated
+
+This rule has been deprecated in favour of the [`@typescript-eslint/no-require-imports`](./no-require-imports.md) rule.
```
auvred, line 14:
```suggestion
This rule has been deprecated in favour of the [`@typescript-eslint/no-require-imports`](./no-require-imports.mdx) rule.
```

auvred (changes requested):

packages/rule-tester/src/utils/dependencyConstraints.ts
```diff
@@ -21,6 +21,7 @@ function satisfiesDependencyConstraint(
       : constraintIn;
 
   return semver.satisfies(
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
```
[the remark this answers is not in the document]
arka1002: I think its a bug, https://github.com/typescript-eslint/typescript-eslint/issues/8994

Happening somewhere here I think https://github.com/typescript-eslint/typescript-eslint/blob/4bed24d8d37ab066b6f3b988fd1ca7accec931c0/packages/eslint-plugin/src/util/getStaticStringValue.ts#L39-L42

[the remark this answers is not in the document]
arka1002: I think the eslint comment is necessary in this case .... https://github.com/typescript-eslint/typescript-eslint/issues/8994#issuecomment-2078092697

JoshuaKGoldberg (approved):
Swell, thanks! Looks great to me. Thanks for making the changes 🙂 

Leaving with `1 approval`, since @auvred was reviewing too.

packages/eslint-plugin/docs/rules/no-require-imports.mdx
```diff
@@ -59,6 +59,33 @@ console.log(require('../package.json').version);
 </TabItem>
 </Tabs>
 
+### `allowAsImport`
+
+When set to `true`, require statements with an `import` declaration won't be reported.
+This is useful if you're on an older CommonJS-style codebase and haven't been able to transition to ESM-style imports yet.
```
Josh-Cena, line 65:
```suggestion
When set to `true`, the `import x = require(...)` declaration won't be reported.
This is useful if you use certain module options that require strict CommonJS interop semantics.
```

packages/scope-manager/tests/fixtures.test.ts
```diff
@@ -1,4 +1,5 @@
 import fs from 'fs';
+// eslint-disable-next-line @typescript-eslint/no-require-imports
 import glob = require('glob');
```
Josh-Cena, line 3: We can't use `import * as glob` here?

arka1002:
The tests which failed in this pr, failed in the v8 branch too ? https://github.com/typescript-eslint/typescript-eslint/commit/c0bf6fc21f82f9545f01a2606f522efa43fdd06c , see the CI check in this commit.

Josh-Cena (approved):

PR #8335, opened 2024-02-01
feat(eslint-plugin): [consistent-type-imports] ignore files with decorators, experimentalDecorators, and emitDecoratorMetadata

### LINKED ISSUE #5468: Bug: [consistent-type-imports] Using with isolatedModules and emitDecoratorMetadata

### Playground Link

https://typescript-eslint.io/play/#ts=4.7.4&sourceType=module&code=JYWwDg9gTgLgBDAnmApnA3nA0ixcC+cAZlBCHAORQoCGAxjBQNwBQLAJinQDY3XEBXAHYNgEIXAAiXaDRjQAFGFKpYiAHI0QKAPwAuOAGcYUYEIDmASgMAFPlpQwUUaXVnyorFigAekWHCcRDQC3PA8NIaGcAAqKMYYbHBwkYgigiIKLMnJAAKu7oqWcGD2IAY4iNlwxejVydQwAlASpVBarMn4bN1AA&eslintrc=N4KABGBECmAeAu0B2ATAzpAXGA2uCUADgDYCuA5gJZKYAC8AnodGgMYBOlh8AtC8dXgB6dtFYB7ALaTkKaCkj4AugBp8kQgEN2aaOyxR6TFhy69+goVp17IaiBu272AeW6VxSDNlAEi7cQArMXgDSAA6IXg2TwAzSnJwwLRPRQgAX3siMiovAzw-SCNmNk5uPjQBJFDlLMh2UmIWA18CIsYS03KLaqEJL0o0RGqeDugeSklCcXZo-PxCgHdtJDsFgla-B0JRWNtsSDGJqZm59Yz1pXx0kHSgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEk8sAJgK4YVTolHoAeekANOBqRkJWKgCGedAwCyjFhzZ5MTdIJrt4yPABF0cTFNiZZ6PBIZSJy1eqG0eiLMnjoiF1PsPHMZSipqQgC+IMFAA

### Repro Code

```typescript
import type { Key } from 'react';

declare function Decorator( propertyName?: string ): ParameterDecorator;

export default class Test {
	
	async func(
		@Decorator() param: Key
	) {
		return param;
	}
	
}
```

### ESLint Config

```javascript
{
	"parser"       : "@typescript-eslint/parser",
	"parserOptions": {
		"project": "./tsconfig.json"
	},
	"plugins"      : [ "@typescript-eslint" ],
	"rules"        : {
		"@typescript-eslint/consistent-type-imports": [ "warn", { "prefer": "type-imports" } ]
	}
}
```

### tsconfig

```jsonc
{
	"compilerOptions": {
		"module"                : "esnext",
		"isolatedModules"       : true,
		"emitDecoratorMetadata" : true,
	}
}
```

### Expected Result

No errors showing for both the decorator and the type import

### Actual Result

if you use import type, you get 
`ESLint: Type import "Key" is used by decorator metadata.(@typescript-eslint/consistent-type-imports)`
while if you use just import, you get 
`TS1272: A type referenced in a decorated signature must be imported with 'import type' or a namespace import when 'isolatedModules' and 'emitDecoratorMetadata' are enabled.`
Not sure if there's a better way to do this since isolatedModules must be set

### Additional Info

This doesn't need complex setup, literally any type import will throw an error for this so I'm not sure if this needs reproduction repository

### Versions

| `@typescript-eslint/eslint-plugin`     | `5.33.0` |
| `@typescript-eslint/parser`            | `5.33.0` |
| `TypeScript`                           | `4.7.4` |
| `ESLint`                               | `8.21.0` |
| `node`                                 | `16.5.0` |

### LINKED ISSUE #6393: Docs: consistent-type-imports vs importsNotUsedAsValues

### Suggested Changes

The `@typescript-eslint/consistent-type-imports` rule flags the following code:

```ts
import { MyType } from "./MyType";
```

And forces you to change it to:

```ts
import type { MyType } from "./MyType";
```

However, it seems that the [`importsNotUsedAsValues`](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues) compiler flag does the same thing (when set to the "error" value).

Thus, the documentation for the `@typescript-eslint/consistent-type-imports` rule should provide a short discussion of the differences between the two, if any, and why you would want to use one over the other.

In issue #4268, Brad explains that you should definitely not use both of them at the same time, but fails to mention anything further.

### Affected URL(s)

https://typescript-eslint.io/rules/consistent-type-imports/

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5468, fixes #6393

## Overview

Note: I've labelled this as a `feat` on purpose to mark it for a minor bump. It's technically a "bug fix" but it's also a change in behaviour. One might classify it as a breaking change because a specific usecase will require a config change - but given that usecases is also already broken, I didn't think that it truly qualified as breaking.

----

This PR implements a sledge-hammer approach to fixing the problem in #5468 -  I have opted to just completely ignore a file if it passes all three of the following conditions:
- there is a decorator in the file
- `experimentalDecorators: true`
- `emitDecoratorMetadata: true`

Originally in #5468 I suggested going with the the approach of specifically ignoring imports that are used in locations that might be passed to decorator metadata. However I ultimately chose not to go down this path because:
- As of TS 5.0 `experimentalDecorators` is considered legacy
- The stable TC39 decorator metadata proposal does not include type annotations
- `verbatimModuleSyntax` exists for the above usecase
- The logic required to concretely detect if a specifier is used within decorator metadata is complex and a maintenance burden and only specifically applies to `experimentalDecorators`

This PR:
- Removes the decorator metadata detection logic from `scope-manager`
- Adds a new `experimentalDecorators` parser option that is auto-populated from type info if available
- Exposes both `experimentalDecorators` and `emitDecoratorMetadata` on the parser services
	- I did this to make it easy for rules to derive this state without having to manually interact with the program to extract the compiler options.
- Updates `consistent-type-imports`:
   - so that it bails out of reporting on a file if the aforementioned 3 conditions are met.
   - removes the error messages relating to decorator metatadata
- Refactors the tests to validate the conditions are correctly applied.

NOTE:: I strongly suggest turning on the no-whitespace diff when reviewing this PR.


packages/eslint-plugin/docs/rules/consistent-type-imports.md
```diff
@@ -88,11 +88,70 @@ type T = import('Foo').Foo;
...
 
-If you are using [type-aware linting](https://typescript-eslint.io/linting/typed-linting), then you just need to ensure that the `tsconfig.json` you've configured for `parserOptions.project` has `emitDecoratorMetadata` turned on. Otherwise you can explicitly tell our tooling to analyze your code as if the compiler option was turned on [by setting `parserOptions.emitDecoratorMetadata` to `true`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/parser/README.md#parseroptionsemitdecoratormetadata).
+When both the compiler options `experimentalDecorators` and `emitDecoratorMetadata` are turned on and a class is annotated with decorators, TypeScript will emit runtime metadata for the class that captures the property types, method parameter types, and method return types. This runtime code is derived using type information - meaning that the runtime code emitted changes based on the cross-file type information that TypeScript has computed.
+
+This behavior is problematic for the style enforced by this lint rule because it also means that annotating an import with `type` can change the runtime metadata! For example in the following snippet TS:
```
JoshuaKGoldberg, line 97: [Proofreading] Nit, unnecessary word:

```suggestion
This behavior is problematic for the style enforced by this lint rule because it also means that annotating an import with `type` can change the runtime metadata! For example in the following snippet:
```

packages/eslint-plugin/docs/rules/consistent-type-imports.md
```diff
@@ -88,11 +88,70 @@ type T = import('Foo').Foo;
 const x: import('Bar') = 1;
 ```
 
-## Usage with `emitDecoratorMetadata`
+## Caveat: `@decorators` + `experimentalDecorators: true` + `emitDecoratorMetadata: true`
```
JoshuaKGoldberg, line 91: [Refactor] This is a pretty huge caveat that takes up multiple windows-worth of content for my desktop monitors. I'd expect that most users probably won't read it... even though the information is good & pertinent.

What do you think about moving roughly everything after the _TL;DR_ and before the next heading into a linked blog post? As in, have this rule's docs contain the high-level important conclusions, and any explanations >=1 paragraph long moved into a deeper read format?

I'm in favor to the point of requesting changes.

packages/parser/src/parser.ts
```diff
@@ -160,14 +160,11 @@ function parseForESLint(
...
-      emitDecoratorMetadata = true;
-    }
   }
 
-  if (emitDecoratorMetadata) {
-    analyzeOptions.emitDecoratorMetadata = true;
-  }
+  // if not defined - override from the parserOptions
+  services.emitDecoratorMetadata ??= options.emitDecoratorMetadata === true;
+  services.experimentalDecorators ??= options.experimentalDecorators === true;
```
JoshuaKGoldberg, line 167: [Refactor] `createParserServices.ts` also has defaulting logic around these two options, along with a now-contradictory comment:

https://github.com/typescript-eslint/typescript-eslint/blob/1c1f9f46b05ff8cf777c6555994b30c0e47da8ef/packages/typescript-estree/src/createParserServices.ts#L27-L29

How about passing through the `options` through `parseAndGenerateServices` to `createParserServices`? That way the defaulting logic can all be in one place.

```ts
  return {
    program,
    emitDecoratorMetadata:
      compilerOptions.emitDecoratorMetadata ??
      originalOptions.emitDecoratorMetadata === true,
    experimentalDecorators:
      compilerOptions.experimentalDecorators ??
      originalOptions.experimentalDecorators === true,
```

Aside: I found the naming of the options variables in `parser.ts` to be a little confusing to read through. There's an `options` variable of type `ParserOptions` and a `parserOptions` variable of type `TSESTreeOptions`... oy.

packages/scope-manager/src/analyze.ts
```diff
@@ -57,9 +57,9 @@ interface AnalyzeOptions {
    */
   sourceType?: SourceType;
 
+  // TODO - remove this in v7
   /**
-   * Emit design-type metadata for decorated declarations in source.
-   * Defaults to `false`.
+   * This option no longer does anything and will be removed in a future major release.
```
JoshuaKGoldberg, line 62:
```suggestion
   * @deprecated This option no longer does anything and will be removed in a future major release.
```

packages/eslint-plugin/src/rules/consistent-type-imports.ts
JoshuaKGoldberg, line 1: [Tests] I can't comment on the line because GitHub, but 350 -> 415 isn't missing unit test coverage on `main`, but is as of this PR:

```ts
                  *fix(fixer) {
                    // take all the valueSpecifiers and put them on a new line
                    yield* fixToValueImportDeclaration(
```

The other two GitHub-reported gaps are missing on `main` already.

docs/packages/Parser.mdx
```diff
@@ -128,6 +129,12 @@ Specifies the version of ECMAScript syntax you want to use. This is used by the
 
 This option allow you to tell parser to act as if `emitDecoratorMetadata: true` is set in `tsconfig.json`, but without [type-aware linting](../linting/Typed_Linting.mdx). In other words, you don't have to specify `parserOptions.project` in this case, making the linting process faster.
 
+### `experimentalDecorators`
+
+> Default `undefined`.
+
+This option allow you to tell parser to act as if `experimentalDecorators: true` is set in `tsconfig.json`, but without [type-aware linting](../linting/Typed_Linting.mdx). In other words, you don't have to specify `parserOptions.project` in this case, making the linting process faster.
```
JoshuaKGoldberg, line 136: [Docs] This paragraph is another motivator for why I think it'd be good to spin out a blog post explaining the changes. On its own / without seeing the `consistent-type-imports` docs, this doesn't explain itself very well. But having a deeper resource explaining the differences separate from a lint rule would be very helpful.

packages/scope-manager/src/analyze.ts
```diff
@@ -57,9 +57,9 @@ interface AnalyzeOptions {
    */
   sourceType?: SourceType;
 
+  // TODO - remove this in v7
```
JoshuaKGoldberg, line 60: Per the new issue:

```suggestion
  // TODO - remove this in v8
```

packages/scope-manager/src/analyze.ts
JoshuaKGoldberg, line 1: > One might classify it as a breaking change because a specific usecase will require a config change - but given that usecase is also already broken, I didn't think that it truly qualified as breaking.

[Meta] The fact that we're making an existing option do nothing makes me think this should really be treated as a major version breaking change. ACK that the use case is roughly broken, and I know the scope manager is rarely used outside of the plugin, but it's still a change to the semantics of scope analysis and therefore parsing as a whole. It's making me nervous. 😬

Beyond taking a bunch of months longer for the slim % of users who are broken right now, what are the main downsides to treating this as a non-breaking feature? Adding to the inevitable slog of merge conflicts from a `v8`->`main` PR?

JoshuaKGoldberg (commented):
Really nice solution, +1 from me on the direction!

Requesting changes mainly on extracting the deeper explanations out of the rule file and into a blog post. Also for at least resolving the breaking change discussion.

But the rule's new implementation looks great to me (thanks for the whitespace suggestion!). ✨

ZackMFleischman:
@bradzacher Thanks for building this solution! I'm eagerly awaiting it being merged so I can use it in my project where I've been forced to just turn this rule off entirely for the moment. 

Also hi @JoshuaKGoldberg! Always fun to see your name pop up. Thanks for fighting the good fight and keeping high-quality open source software available! 😄

JoshuaKGoldberg:
Haha hey @ZackMFleischman, great to see you pop up too! 😄

packages/parser/src/parser.ts
bradzacher: The reason I didn't do that is because I didn't want the ESLint parser options to bleed into the typescript-estree parser options.

Ultimately `parserOptions.emitDecoratorMetadata` and `parserOptions.experimentalDecorators` are not used at all within `typescript-estree` - so it seems wrong to pass them all the way through into that package.

I think it's better to keep the logic isolated to `parser` so it only impacts true eslint parser.

packages/scope-manager/src/analyze.ts
bradzacher: > is roughly broken

I wouldn't say roughly, I'd say completely.
Anyone who depends on scope analysis to return the correct state when `emitDecoratorMetadata` is going to get a completely broken result that does not reflect reality.

So all this PR is doing is saying "this is currently broken so we're unbreaking it by not returning the broken data any more"

---

The reality is that we're the only consumers of this broken data - the only consumer is specifically the `consistent-type-imports` rule.

We've removed the only usage internally and I'm highly confident that it's the only usage in the world.

TBH I'd be okay with removing this prop but that would explicitly be a breaking API surface change.

---

> Beyond taking a bunch of months longer for the slim % of users who are broken right now, what are the main downsides to treating this as a non-breaking feature? 

It's actually not just the decorator users that are broken - there are a number of cases where non-decorator users will see an error saying `Type import Foo is used by decorator metadata.`

This was actually the "last straw" motivation for me to fix this! I encountered this problem in a codebase that didn't use decorators and was like "ugh I should really fix that".
We've had a few issues filed about that and usually it occurs because there's some weird code the rule didn't really expect - but real users do run into it.

docs/packages/Parser.mdx
```diff
@@ -128,6 +129,12 @@ Specifies the version of ECMAScript syntax you want to use. This is used by the
 
 This option allow you to tell parser to act as if `emitDecoratorMetadata: true` is set in `tsconfig.json`, but without [type-aware linting](../getting-started/Typed_Linting.mdx). In other words, you don't have to specify `parserOptions.project` in this case, making the linting process faster.
 
+### `experimentalDecorators`
+
+> Default `undefined`.
+
+This option allow you to tell parser to act as if `experimentalDecorators: true` is set in `tsconfig.json`, but without [type-aware linting](../getting-started/Typed_Linting.mdx). In other words, you don't have to specify `parserOptions.project` in this case, making the linting process faster.
```
JoshuaKGoldberg, line 136: [Praise] 💯 excellent.

packages/scope-manager/src/analyze.ts
JoshuaKGoldberg: Makes sense, thanks. I think with the docs you've added this will be a lot more clear to end users too!

packages/typescript-estree/src/createParserServices.ts
```diff
@@ -8,16 +8,25 @@ export function createParserServices(
...
 ): ParserServices {
   if (!program) {
-    // we always return the node maps because
-    // (a) they don't require type info and
-    // (b) they can be useful when using some of TS's internal non-type-aware AST utils
-    return { program, ...astMaps };
+    return {
+      program,
+      emitDecoratorMetadata: undefined,
+      experimentalDecorators: undefined,
```
JoshuaKGoldberg, line 14: [Style] Is there a reason to not let these be `?:` optional? Just consistency of whether they exist?

Two cents: I'd personally prefer to keep them out of the objects & snapshots for the sake of keeping things lean. They're not used by most users, so it's a fair bit of extra noise.

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+
+However we decided that adding type-information to the rule to handle this edge case was not a positive change for users or the ecosystem for two main reasons:
+
+1. It requires a specific combinations of compiler options to trigger it means that not everyone is impacted by the problem - so we'd be preventing a lot of un-impacted users from using the rule.
+2. With the release of [TypeScript v5.0 and its stable decorators](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#decorators) `experimentalDecorators` are now the legacy syntax. Whilst [TypeScript v5.2 added support for the latest stable decorator metadata proposal](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#decorator-metadata) this proposal does not include type metadata - so it doesn't suffer the same drawbacks as its legacy counterpart.
+
+## The Final Solution
```
JoshuaKGoldberg, line 70: "Final" isn't a word that often holds up over time. We might end up changing this later.

There's also a pretty, ah, strong association with this specific phrase that I personally would prefer to avoid. 

How about...

```suggestion
## The Compromise
```

...or _"Today's Solution"_? _"Next"_? ...?

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+  [
+    consistent-type-imports,
+    experimentalDecorators,
+    emitDecoratorMetadata,
+    typescript-eslint,
+  ]
+title: Changes to consistent-type-imports when used with legacy decorators and decorator metadata
+---
+
+We've made some changes to `consistent-type-imports` to fix some long-standing issues when used alongside `experimentalDecorators: true` and `emitDecoratorMetadata: true`. These changes increase safety and prevent invalid fixes when using decorator metadata.
```
JoshuaKGoldberg, line 19: A little convenience for folks:

```suggestion
We've made some changes to the [`consistent-type-imports` rule](/rules/consistent-type-imports) to fix some long-standing issues when used alongside `experimentalDecorators: true` and `emitDecoratorMetadata: true`. These changes increase safety and prevent invalid fixes when using decorator metadata.
```

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+
+<!-- truncate -->
+
+## Experimental Decorator Metadata
+
+The `experimentalDecorators` compiler option (referred to as "legacy decorators" from here on) turns on support for an old version of the decorator proposal that was never standardized. Syntactically this proposal isn't too dissimilar to the current [Stage 3 proposal](https://github.com/tc39/proposal-decorators), however they differ in one major way - the use of [metadata reflection](https://rbuckton.github.io/reflect-metadata/) when `emitDecoratorMetadata` is turned on.
```
JoshuaKGoldberg, line 25: Some readers aren't likely to know what the compiler options are or even the difference between a compiler option and a TC39 proposal, I'd wager. Rewording a bit to introduce it first:

```suggestion
TypeScript's [`experimentalDecorators` compiler option](https://aka.ms/tsconfig#experimentalDecorators) (referred to as "legacy decorators" from here on) turns on support for an old version of the [JavaScript TC39 decorator proposal](https://github.com/tc39/proposal-decorators) that was never standardized. TypeScript's legacy decorators are similar to the current proposal, but differ in that they use [metadata reflection](https://rbuckton.github.io/reflect-metadata/) when TypeScript's [`emitDecoratorMetadata` compiler option](https://aka.ms/tsconfig#emitDecoratorMetadata) is turned on.
```

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+
+When using legacy decorators with decorator metadata and a class is annotated with decorators TypeScript will emit runtime metadata for the class that captures the property types, method parameter types, and method return types. This metadata is incredibly powerful and provides a bridge between the types (which are elided at compile time) and the runtime code. However this runtime code is generated at a cost as it is derived using type information - meaning that the runtime code emitted changes based on the cross-file type information that TypeScript has computed.
```
JoshuaKGoldberg, line 27: Suggestion, I'd reword a bit to split up sentences and the paragraphs into two discrete points, and be slightly less flattering for the decorators:

```suggestion
When using legacy decorators with decorator metadata and a class is annotated with decorators, TypeScript will emit runtime metadata for the class. That decorator metadata will capture property types, method parameter types, and method return types. Decorator metadata provides a bridge between the types (which are not available at compile time) and the runtime code. 

The downside of generating this runtime code that it is derived using type information: meaning that the runtime code emitted changes based on the cross-file type information that TypeScript has computed. Doing so violates a key [TypeScript design goal](https://github.com/microsoft/TypeScript/wiki/TypeScript-Design-Goals) of not changing runtime behavior based on type information.
```

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+If the imported name `Foo` resolves to...
+
+- a type then TS will emit metadata that annotates `arg` with one of `Function`, `Object`, `String`, `Number`, or `Boolean` depending on what that type resolves to.
+- an enum then TS will emit metadata that annotates `arg` with one of `String`, `Number`, or `Object` depending on the type of the enum's members.
+- a class declaration:
+  - and the import **_is NOT_** annotated as `import type` then TS will emit metadata that annotates the type of `arg` as `Foo`.
+  - and the import **_IS_** annotated as `import type` then TS will emit metadata that annotates the type of `arg` as `Function`.
+
```
JoshuaKGoldberg, line 48: I'm of the opinion it's good for narrative flow to end every section with a brief summation of the final point(s) that lead into the next one.

```suggestion

In addition to requiring runtime type information, those metadata emit rules are confusing for developers to reason about.
They necessitate understanding edge cases specific to TypeScript's handling of decorators and type information.

```

packages/scope-manager/src/analyze.ts
```diff
@@ -57,9 +57,9 @@ interface AnalyzeOptions {
    */
   sourceType?: SourceType;
 
+  // TODO - remove this in v8
   /**
-   * Emit design-type metadata for decorated declarations in source.
-   * Defaults to `false`.
+   * @deprecated This option no longer does anything and will be removed in a future major release.
```
JoshuaKGoldberg, line 62: I came into this confused at seeing as what looked like "breaking" change in a minor. Just making sure nobody else gets confused by that too:

```suggestion
   * @deprecated This option never did what it was intended for and will be removed in a future major release.
```

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
JoshuaKGoldberg, line 27: > runtime metadata

I think many developers won't know what this means, so it'd be nice to have a brief mention... but I can't think of a good way to describe it 😬. Maybe a one sentence description around here? Maybe a small sketch of emitted JS after the TS snippet? Thoughts?

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+## `consistent-type-imports` caused runtime breakage
+
+The important piece is that last dot point - the handling of imported names that resolve to class declarations. If the import is not annotated as `import type` then TS emits a runtime reference to the imported name. This runtime reference is implicit and requires type information to derive - you cannot derive its existence purely based on single-file AST analysis.
+
+The `consistent-type-imports` rule was introduced to allow users to enforce that any imported names are annotated as `import type` if they are not used in a value location. How the rule makes this decision is based on single-file AST analysis; it scans the code using a technique called scope analysis so that it can find all references to the imported names and determine if each reference is a value reference.
```
JoshuaKGoldberg, line 53: Devs reading this might not know what an AST is. And they don't need to in order to understand that it's a one-file-at-a-time thing. This is a good place to suggest they learn it though!

```suggestion
The `consistent-type-imports` rule was introduced to allow users to enforce that any imported names are annotated as `import type` if they are not used in a value location. How the rule makes this decision is based solely on the single file it's looking at.
It scans the code using a technique called "scope analysis" so that it can find all references to the imported names and determine if each reference is a value reference.

:::tip
See [ASTs and typescript-eslint](../asts-and-typescript-eslint) to understand how rules look at the syntax of files.
:::
```

JoshuaKGoldberg: Aside: I found it a little confusing to read through this. I think the point that `consistent-type-imports` tries to be an untyped rule tackling a typed problem isn't super clear immediately. This was a vague attempt at rewriting some of it:

> Where the `consistent-type-imports` rule used to trip up is around TypeScript's code emit including a runtime reference to any import not annotated as `import type`. That runtime reference is implicit and requires type information to derive. TypeScript needs to understand the type of the import to emit the metadata. But, `consistent-type-imports` was an untyped lint rule. It didn't have access to type information to know whether an import used as a decorator could be switched from `import` to `import type`!

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+
+However this solution had a hidden pitfall; if the user also used `isolatedModules: true` then TS will enforce that all imported types are explicitly marked as `import type` for compatibility with single-file build tools. This lead to an unresolvable situation where `consistent-type-imports` would enforce that an imported name _must not be_ marked with `import type` so that we could ensure we don't break decorator metadata, and simultaneously TS would enforce that that same imported name _must be_ marked with `import type`.
+
+There have been a few attempts to fix this issue but the resolution we came to was that the only solution was to add type information to the rule so that it could correctly understand all of the above type-aware constraints. Adding type information to an existing rule is something we try to avoid because it is a major breaking change that restricts the rule to just users that leverage [type-aware linting](/getting-started/typed-linting).
+
+However we decided that adding type-information to the rule to handle this edge case was not a positive change for users or the ecosystem for two main reasons:
```
JoshuaKGoldberg, line 65: Being a little more precise here: we didn't decide reality, we just determined it.

```suggestion
Adding type-information to the rule to handle this edge case would not be a positive change for users or the ecosystem:
```

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+
+1. It requires a specific combinations of compiler options to trigger it means that not everyone is impacted by the problem - so we'd be preventing a lot of un-impacted users from using the rule.
```
JoshuaKGoldberg, line 67: I don't think the downsides of typed linting are mentioned previously?

```suggestion
1. Many users are unable to configure typed linting and/or unwilling to take its performance hit. Requiring typed linting for this rule to serve a very small subset of impacted users would reduce the linting ability of many more un-impacted users.
```

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+2. With the release of [TypeScript v5.0 and its stable decorators](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#decorators) `experimentalDecorators` are now the legacy syntax. Whilst [TypeScript v5.2 added support for the latest stable decorator metadata proposal](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#decorator-metadata) this proposal does not include type metadata - so it doesn't suffer the same drawbacks as its legacy counterpart.
+
+## The Final Solution
+
+Ultimately we decided the best solution was to just opt-out of handling this use-case entirely. This means that we can avoid accidentally reporting the wrong thing and fixing to code that either fails to compile or alters the emitted runtime metadata. To be clear this means that if you have **both** `experimentalDecorators: true` and `emitDecoratorMetadata: true` then rule will **_not_** report any errors within any files _that contain decorators_.
```
JoshuaKGoldberg, line 72: Nit: I think emphasizing visually with some separation does a bit better job of making something clear, without adding more words to read.

```suggestion
Ultimately we determined the best solution was to just opt-out of handling this use-case entirely. This means that we can avoid accidentally reporting the wrong thing and fixing to code that either fails to compile or alters the emitted runtime metadata. 

Now, if you have **both** `experimentalDecorators: true` and `emitDecoratorMetadata: true`, then the `consistent-type-imports` rule will **_not_** report any errors within any files _that contain decorators_.
```

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+
+All files without decorators will continue to report as expected. Similarly all projects that use `experimentalDecorators: false` and/or `emitDecoratorMetadata: false` will continue to report as expected.
+
+### Configuring the Linter to expect `experimentalDecorators: true` and `emitDecoratorMetadata: true`
```
JoshuaKGoldberg, line 76: Nit: unnecessary capitalization?

```suggestion
### Configuring the linter to expect `experimentalDecorators: true` and `emitDecoratorMetadata: true`
```

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,101 @@
...
+description: Changes to consistent-type-imports when used with decorators, experimentalDecorators, and emitDecoratorMetadata
+slug: changes-to-consistent-type-imports-with-decorators
+tags:
+  [
+    consistent-type-imports,
+    experimentalDecorators,
+    emitDecoratorMetadata,
+    typescript-eslint,
+  ]
+title: Changes to consistent-type-imports when used with legacy decorators and decorator metadata
```
JoshuaKGoldberg, line 16: Nit: other blog post titles are all capitalized.

```suggestion
title: Changes to `consistent-type-imports` with Legacy Decorators and Decorator Metadata
```

JoshuaKGoldberg (changes requested):
Awesome! I'm super happy with the implementation, and think I understand this much better now thanks to the docs & comments. Thanks!

Left some proofreading thoughts on the blog post. That's the only real blocker from me - and there's nothing I likely can't be convinced otherwise on.

🔥

packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
```diff
@@ -93,11 +93,34 @@ type T = import('Foo').Foo;
...
+There are a few behavior differences:
+| Situation | `consistent-type-imports` (ESLint) | `verbatimModuleSyntax` (TypeScript) |
+| -------------------------------------------------------------- | --------------------------------------------------------- | ----------------------------------------------------------- |
+| Unused imports | Ignored (consider using [`@typescript-eslint/no-unused-vars`](/rules/no-unused-vars)) | Type error |
+| Usage with `emitDecoratorMetadata` & `experimentalDecorations` | Ignores files that contain decorators | Reports on files that contain decorators |
+| Failures detected | Does not fail `tsc` build; can be auto-fixed with `--fix` | Fails `tsc` build; cannot be auto-fixed on the command-line |
+| `import { type T } from 'T';` | TypeScript will emit nothing (it "elides" the import) | TypeScript emits `import {} from 'T'` |
+
+Because there are some differences, using both this rule and `verbatimModuleSyntax` at the same time can lead to conflicting errors.
+As such we recommend that you only ever use one _or_ the other -- never both.
```
JoshuaKGoldberg, line 123: [Praise] this is a very clear section, nice.

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
JoshuaKGoldberg, line 1: [Praise] Pretty much all my comments are on small narrative/proofreading things. The post overall is super clear and helped me come back into this. Nicely done!

packages/typescript-estree/src/createParserServices.ts
bradzacher: The reason I didn't was yeah so that things were consistent - they either exist or they don't. We ran into problems in the past with properties maybe existing and it complicating things.

Eg if someone uses a different parser or an older parser and the properties don't exist - is that exactly the same as them existing but being `undefined`?
By making them exist it's much easier should we need to check down the line.

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
bradzacher: Rewrote the end of the section to include the transpiled output.
I think it actually made things much clearer and more concrete! Before it was a little bit abstract.

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
```diff
@@ -0,0 +1,133 @@
...
+  @decorator
+  method(arg: Foo) {}
+}
+```
+
+TypeScript will transpile this code to the following:
+
+<!-- prettier-ignore -->
+```ts
+var _a;
```
JoshuaKGoldberg, line 47: Nit: this isn't used later.

```suggestion
```

If TypeScript is producing an `_a` despite never using it, then that's a bug we can ignore.

JoshuaKGoldberg (dismissed):
Whoop, this is great - useful userland fixing and really solid explanations @bradzacher! 👏

packages/website/blog/2024-03-25-changes-to-consistent-type-imports-with-decorators.md
bradzacher: Oh nah it was there cos the code has no types so this is an extra case I (purposely) didn't mention. In this case when TS can't determine the type but it can see its a value-import - instead TS emits  `[typeof _a = (typeof Foo !== "undefined" && Foo) === "function" ? _a : Object]`. Which is some funky code-golf heh.

Can remove it though yeah as it's not necessary for the examples I'm mentioning.

PR #8339, opened 2024-02-02
feat: allow `parserOptions.project: false`

bradzacher:
## Overview

More splitting out of #7935.
This change allows us to pass `false` to `parserOptions.project`.
This allows us to work-around eslint/eslint#17820 not being in ESLint v8


JoshuaKGoldberg (approved):
Swell! 🙌

PR #8344, opened 2024-02-02
feat(utils): improve eslint types

bradzacher:
## Overview

More splitting out of https://github.com/typescript-eslint/typescript-eslint/pull/7935.

- Loosen parser types in flat configs to allow for non-typescript-estree parsers
- Loosen plugin types to allow for not-yet-updated plugins to be passed to flat configs
- Add some missing features added in newer v8 versions


JoshuaKGoldberg (approved):
🚢

PR #8351, opened 2024-02-03
feat(rule-tester): check type-aware rules test cases for TS type errors

### LINKED ISSUE #8298: Make sure type-aware lint rules' test cases are type-error-free
### Suggestion

We should add an extra type-checking step to rule-tester when testing type-aware lint rules. Right now, we have some test cases that don't pass, most notably because they have undeclared variables (which become `any`). However we don't guarantee linting behavior when there are type errors (because the result is likely `any`, which either created new lint errors or masks actual lint errors). Such type errors may be hard to catch in more complex situations.

Brad mentioned there may be a perf penalty, but it would be useful to experiment nonetheless.

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8298

## Overview

wip: just experimenting for now


JoshuaKGoldberg:
Closing out as it's been a while. Cheers!

Issue #7297, opened 2023-07-22
Docs: restrict-plus-operands example is wrong for the default configuration


lgenzelis:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

[restrict-plus-operands docs](https://typescript-eslint.io/rules/restrict-plus-operands/) starts by giving these incorrect examples:

```typescript
let foo = '5.5' + 5;
```

I remember that was once the case, and I don't know when it changed, but now that's not an error anymore. The default configuration is:

```javascript
const defaultOptions: Options = [
  {
    allowAny: true,
    allowBoolean: true,
    allowNullish: true,
    allowNumberAndString: true,
    allowRegExp: true,
    skipCompoundAssignments: false,
  },
];
```

Since we have `allowNumberAndString: true`, it's ok to do `'5.5' + 5`. It can be verified [in the playground](https://typescript-eslint.io/play/#ts=5.1.6&fileType=.ts&code=DYUwLgBAZg9jEF4IHICsA6VyIGoKoG4AoIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6aFfKsu4%2BWZLQD2paAEMmAE1QZIiaNCHRI4MAF8QaoA&tsconfig=&tokens=false).

### Affected URL(s)

https://typescript-eslint.io/rules/restrict-plus-operands/

BrianLeishman:
Also running into this, and it's quite confusing. Shouldn't all of these really default to `false` instead of `true`?

The `Caution` note seems to imply that these *shouldn't* be set to true at all.

ad1992:
Looks like this issue was fixed by #8362  so can be closed?

auvred:
Yes, thanks!

PR #8362, opened 2024-02-04
docs: fix restrict-plus-operands example

nobuaki0331:
I have removed the example of restrict-plus-operands

## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/7297

## Overview

I have removed the example of restrict-plus-operands because it seems to be inaccurate and no longer relevant.

### DIFF
--- a/packages/eslint-plugin/docs/rules/restrict-plus-operands.md
+++ b/packages/eslint-plugin/docs/rules/restrict-plus-operands.md
@@ -19,13 +19,14 @@ This rule reports when a `+` operation combines two values of different types, o
 
 ```ts
 let foo = 1n + 1;
+let fn = (a: string, b: never) => a + b;
 ```
 
 ### ✅ Correct
 
 ```ts
-let foo = parseInt('5.5', 10) + 10;
 let foo = 1n + 1n;
+let fn = (a: string, b: string) => a + b;
 ```
 
 ## Options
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/docs/rules/restrict-plus-operands.md
```diff
@@ -18,7 +18,6 @@ This rule reports when a `+` operation combines two values of different types, o
 ### ❌ Incorrect
 
 ```ts
-let foo = '5.5' + 5;
```
JoshuaKGoldberg, line 21: Ah, but then the first line in _Correct_ no longer matches up. Instead of removing this, could you replace the two first lines with something that actually does error?

JoshuaKGoldberg (changes requested):
👏 Nice spot, thanks for sending! This must be a holdover from before the last time we modified the rule's default options.

packages/eslint-plugin/docs/rules/restrict-plus-operands.md
nobuaki0331: Thank you sincerely for your thoughtful review comments.

I have considered the example.I would be grateful to receive your valuable insights on this matter.

JoshuaKGoldberg (approved):
Beautiful, thanks!

Issue #6887, opened 2023-04-11
Docs: Mention preference for small, self-contained rule test cases


JoshuaKGoldberg:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

In general, in our `packages/eslint-plugin/tests/rules/**/*.test.ts` tests, I'm under the impression we generally ask for test cases that are single-purpose.

Good:

```ts
{
  code: `willCauseComplaint(1);`
  errors: [
    { /* ... */ }
  ]
}
```

Not so good, should be split up:

```ts
{
  code: `
    willCauseComplaint(1);
    willCauseComplaint(2);
  `
  errors: [
    { /* ... */ },
    { /* ... */ }
  ]
}
```

I see the latter form come up sometimes in PRs. Let's document our preference for the former?

### Affected URL(s)

https://typescript-eslint.io/maintenance/pull-requests

Or perhaps we'd want to split out a page dedicated to good practices for ESLint rule PRs? Since they're such a large percentage of our PRs & have specific practices associated.

Aside: if this is accepted by other maintainers we should file an issue to fix up old test files that violate this preference. E.g. `no-floating-promises.test.ts`.

PR #8374, opened 2024-02-04
feat(eslint-plugin): [prefer-string-starts-ends-with] add allowSingleElementEquality option

### LINKED ISSUE #7140: Enhancement: [prefer-string-starts-ends-with] Option to ignore [0] === cases?

https://typescript-eslint.io/play/#ts=5.0.4&sourceType=module&code=CYUwxgNghgTiAEYD2A7AzgF3gNyhAriAFzyYwCWKA5gNwBQd5AZvABS4EgDaADALrwAvMJx5CvPgDoMSAKoAHeSBgBhKGhCsAlFvgBvOvHgB6Y-EkW6AXwbM2HQpMywMaAOrkMAC3Zju-aTlFZTUNbR19QxMzC0lrOiA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WluXxUmAc374AhtHzJ6TACYyA7pXwALdFF7QA9tEjgwAXxCGgA

### Description

Coming over from https://github.com/typescript-eslint/typescript-eslint/pull/7138#discussion_r1241171363: when checking whether a string's `[0]` member equals a modification of `[0]`, the rule's preference ends up being even more code:

```ts
declare const value: string;

if (value[0] === value[0].toUpperCase()) { /* ... */ } // before

if (value.startsWith(value[0].toUpperCase())) { /* ... */ } // after
```

Agreeing with @bradzacher:

> I actually think that the previous form is clearer for this specific case.
> [It's also substantially faster (not that it matters at the this speed... but it is twice as fast - ~1.2B ops/s vs ~2.4B ops/s)](https://jsbench.me/buljbfvcn4/1). 
>
> I wonder if we should add an option to skip this case from the rule?

It's not yet clear to me how we would categorize and describe "this case". Maybe it's that we don't want to flag cases that are just checking whether `value[0] ===` something? In that case, maybe the option would be named something like `checkFirstElementEquality`?

### Fail

```typescript
n/a
```

### Pass

```typescript
n/a
```

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7140

## Overview

Adds a new `allowSingleElementEquality` option to the rule that's allowed to be `'never'` (default) or `'always'`.

Doesn't yet add a third option for case matching as described in the issue. Is that ok to leave as a followup?

Also removes the auto-generation of tests from `prefer-string-starts-ends-with.test.ts` per #6887. The test file predates that preference.


JoshuaKGoldberg:
On mobile, let's assume I'll update the snapshots before merge :)

packages/eslint-plugin/docs/rules/prefer-string-starts-ends-with.md
```diff
@@ -52,6 +52,26 @@ foo.startsWith('bar');
...
+## Options
+
+### `allowSingleElementEquality`
+
+If switched to `'always'`, the rule will allow equality checks against the first or last character in a string.
+This can be preferable in projects that don't deal with special character encodings and prefer a more succinct style.
+
+The following code is considered incorrect by default, but is allowed with `allowSingleElementEquality: 'always'`:
+
+```ts
```
auvred, line 66:
```suggestion
```ts option='{ "allowSingleElementEquality": "always" }' showPlaygroundButton
```

What do you think about adding "Open in Playground" button here? :slightly_smiling_face:

auvred (approved):
Looks great! :fire:

packages/eslint-plugin/docs/rules/prefer-string-starts-ends-with.md
JoshuaKGoldberg: Oh good point. I wonder if we should add in something to make sure we always remember to ask for that (I don't)... or make it more automatic. 🤔

bradzacher (approved):

PR #8375, opened 2024-02-04
chore: enable no-confusing-void-expression internally

JoshuaKGoldberg:
## PR Checklist

## Overview

Enables the rule internally and runs its `--fix` to correct all violations. The fixed style isn't always quite what I'd want, but I figured anything we definitively don't like should probably be enforced with separate lint rules:

* Using `for..of` instead of `.forEach`
* Not having `{}`s around a `switch` `case` unnecessarily


JoshuaKGoldberg:
I'll wait to merge these changes until the flat config work lands.

Issue #8346, opened 2024-02-02
📣 Upcoming v7 Major Release


bradzacher:
We're planning on releasing our next major version within the week with the following breaking changes:

- Drop support for Node 16.
- Bump the minimum required TypeScript version to v4.7.
- Bump the minimum required ESLint version to v8.56.0.

And that's it - it's just dependency bumps! In a nutshell we're looking to make our maintenance easier going forward by modernising our requirements:

- Node 16 is EOL and some of our dev dependencies have already dropped support, preventing us from upgrading.
- TypeScript <v4.7 are now more than 2 years old, beyond our [documented compatibility window](https://typescript-eslint.io/users/dependency-versions#typescript)

As for ESLint - with their upcoming v9 release they will be removing a number of APIs - all of which they added replacements for in v8.40.0. Our next major was going to drop support for ESLint v7 as its 3 years old now and we chose to require the latest ESLint v8 release to reduce variance in the ecosystem and enforce that the necessary APIs are available.

PR #8377, opened 2024-02-05
feat: bump ESLint, NodeJS, and TS minimum version requirements

bradzacher:
BREAKING CHANGE:
Bumps the minimum dependency requirements

## Overview

Fixes https://github.com/typescript-eslint/typescript-eslint/issues/8346

Breaking changes:
- Bump the minimum NodeJS version to v18.18.0 - which aligns with ESLint v9's minimum.
- Drop support for NodeJS 19 (same as ESLint v9)
- Bump the minimum TypeScript version to v4.7.5
- Bump the minimum ESLint version to v8.56.0
- Remove backwards-compatability helpers from `@typescript-eslint/utils/eslint-utils` (`getAncestors`, `getCwd`, `getDeclaredVariables`, `getFilename`, `getScope`, `getSourceCode`)

Other changes:
- Update ESLint types to match the new minimum version.
- Update the playground so it doesn't allow users to select an unsupported TS version.
- Update the docs so that dependency versions are pulled in automatically.


packages/eslint-plugin/tests/rules/consistent-type-imports.test.ts
JoshuaKGoldberg, line 1: [Testing] I did a full text search for `dependencyConstraints: {` in `eslint-plugin` test files and got 48 results in 9 files:

* `typescript: '3.8',`: 1 result, 1 file
* `typescript: '4.1',`: 35 results, 6 files
* `typescript: '4.3',`: 10 results, 2 files
* `typescript: '4.7',`: 2 results, 1 file

More removals to be done? 🔪

packages/typescript-estree/src/parseSettings/warnAboutTSVersion.ts
```diff
@@ -2,11 +2,11 @@ import semver from 'semver';
 import * as ts from 'typescript';
 
 import type { ParseSettings } from './index';
+
 /**
- * This needs to be kept in sync with /docs/users/Versioning.mdx
- * in the typescript-eslint monorepo
+ * This needs to be kept in sync with package.json in the typescript-eslint monorepo
```
JoshuaKGoldberg, line 7: [Praise] Heh, good fix.

packages/website/src/components/OptionsSelector.tsx
```diff
@@ -24,6 +25,8 @@ export interface OptionsSelectorParams {
   readonly tsVersions: readonly string[];
 }
 
+const MIN_TS_VERSION_SEMVER = '>=4.7.5';
```
JoshuaKGoldberg, line 28: [Bug] 4.7 isn't being given as an option in the playground, was this meant to be 4.7.4?

Playground TypeScript version selector showing minors from 4.8 up through 5.3

```suggestion
const MIN_TS_VERSION_SEMVER = '>=4.7.4';
```

JoshuaKGoldberg (dismissed):
LGTM! I trust you on the playground selector 🙂 . This is great - really excited to lose the burden of older ESLint support!

JoshuaKGoldberg:
I ran a few full text regex searches in `.ts` files and found some older ESLint compat support:

* `/eslint.*v7/`
* `/eslint.*v8/

Not blocking IMO, but at least something to note as a good followup.

JamesHenry:
🎉

package.json
```diff
@@ -63,7 +63,6 @@
     "@nx/eslint": "17.2.8",
     "@nx/jest": "17.2.8",
     "@nx/workspace": "17.2.8",
-    "@prettier/sync": "^0.5.0",
```
JoshuaKGoldberg, line 66: _heck yes_

JoshuaKGoldberg, line 66: _heck yes_ async everywhere

bradzacher: everywhere *except* `eslint-plugin-internal`. One day we'll have async plugins . . . .

JoshuaKGoldberg (approved):
A girl dressed in victorian clothing (in the Muppet Christmas Carol movie) happily saying 'Wonderful'

package.json
JoshuaKGoldberg: 🙏

PR #8389, opened 2024-02-06
feat(eslint-plugin): [restrict-template-expressions] add `allowArray` option

### LINKED ISSUE #6279: Enhancement: [restrict-template-expressions] More permissive type check

https://typescript-eslint.io/rules/restrict-template-expressions/

### Description

Related: #6203 #5325

I've originally opened #5600 because I'd like to leverage this rule to discover the *accidental misuse* of string template expression (e.g., the mostly occurring case is, interpolating a `() => string`, `Promise<string>` or `Promise<...>` instead of `string`). However, after I've opened `restrict-template-expressions`. there seems to be a lot of rule violations that cannot be silenced with existing options. For example,
* `string[]`
* `unknown`
* `never`
* `URL`
* `URLSearchParameters`

Instead of defining what types are *allowed* in the interpolate expression, is it possible to set up a list of blocked types?

### Fail

```typescript
const formatErrorMessage = (error: unknown) => String(error);
// `formatErrorMessage` is of function type. Converting it into string results in function body being printed.
const message = "Error is {formatErrorMessage}";
```

### Pass

```typescript
const formatErrorMessage = (error: unknown) => String(Error);
// Call the function to return the string result.
const message1 = "Error is {formatErrorMessage(err)}.";
// Or if you are converting function into string by design.
const message2 = "Error is {String(formatErrorMessage)}.";

// Stringify the URL
let url: URL | undefined;
const message3 = "URL is {url}.";
```

### Additional Info

_No response_

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: begins to fix #6279

## Overview
I've added a new `allowArray` option that applies the existing checks from `restrict-template-expressions` recursively (i.e. no `undefined`s, etc.).

I haven't added unit tests yet but I'd like to begin receiving feedback on this change.


packages/eslint-plugin/src/rules/restrict-template-expressions.ts
```diff
@@ -12,16 +12,35 @@ import {
...
-    allowNumber?: boolean;
-    allowRegExp?: boolean;
-    allowNever?: boolean;
+const optionsObj = {
+  Any: isTypeAnyType,
+  Array: (type, checker, rec): boolean => {
+    if (!checker.isArrayType(type)) {
+      return false;
+    }
+    const numberIndexType = type.getNumberIndexType();
```
JoshuaKGoldberg, line 21: [Style] Nit: a little more terse...

```suggestion
    const numberIndexType = checker.isArrayType(type) && type.getNumberIndexType();
```

packages/eslint-plugin/src/rules/restrict-template-expressions.ts
```diff
@@ -12,16 +12,35 @@ import {
...
+  RegExp: (type, checker): boolean => getTypeName(checker, type) === 'RegExp',
+  Never: isTypeNeverType,
+} satisfies Record<
+  string,
+  (
+    type: ts.Type,
+    checker: ts.TypeChecker,
+    rec: (type: ts.Type) => boolean,
+  ) => boolean
+>;
```
JoshuaKGoldberg, line 38: [Naming] `fn` and `rec` aren't super informative names. I had a hard time figuring out what this does without scanning back and forth through other code. Maybe... `recurser`? `typeTester`? Names like that?

Extracting out a `type OptionTypeTester` for the `(type, checker, rec)` functions also helped me read it a bit.

packages/eslint-plugin/src/rules/restrict-template-expressions.ts
```diff
@@ -12,16 +12,35 @@ import {
...
+const optionsArr = Object.entries(optionsObj).map(([_type, fn]) => {
+  const type = _type as keyof typeof optionsObj;
+  return { type, option: `allow${type}` as const, fn };
+});
```
JoshuaKGoldberg, line 42: [Refactor] This doesn't need both an `optionsObj` and an `optionsArr`. `optionsObj` is only ever used to create `optionsArr` - so we can inline them a bit.

```ts
const optionsTesters = [
  ["allowAny", isTypeAnyType],
  // ...
] as const satisfies [string, OptionTypeTester][];

type Options = [{ [Type in (typeof optionsTesters)[number][0]]?: boolean }];
```

...and similar changes lower down.

JoshuaKGoldberg (changes requested):
This is a clever way of automating the repeated parts 🙂. Nicely done.

...but like many clever code things, there was a difficulty cost of reading it at first. I played around with it locally and found it easier to read through with a few small structural changes. What do you think?

Aaron Ashmore in character in some sci fi show saying "Nifty!", impressed

packages/eslint-plugin/src/rules/restrict-template-expressions.ts
```diff
@@ -175,7 +122,12 @@ export default createRule<Options, MessageId>({
           return innerType.types.some(rec);
         }
 
-        return predicate(innerType);
+        return (
+          isTypeFlagSet(innerType, ts.TypeFlags.StringLike) ||
+          optionsArr.some(
+            ({ option, fn }) => options[option] && fn(innerType, checker, rec),
```
JoshuaKGoldberg, line 128: > `options[option]`

[Refactor] Maybe it'd make sense to generate this list within the rule, so a bunch of truthiness checks don't need to happen each time a template expression is looked at?

packages/eslint-plugin/src/rules/restrict-template-expressions.ts
abrahamguo: nice! not sure why I didn't think of this haha

JoshuaKGoldberg (dismissed):
Super, looks good to me! Thanks for all the code tinkering. 😄 

Leaving open with `1 approval` as I know at least @bradzacher has had well-informed thoughts on this family of rules in the past.

packages/eslint-plugin/src/rules/restrict-template-expressions.ts
```diff
@@ -12,15 +13,44 @@ import {
...
+  [
+    ['Any', isTypeAnyType],
+    [
+      'Array',
+      (type, checker, recursivelyCheckType): boolean =>
+        checker.isArrayType(type) &&
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        recursivelyCheckType(type.getNumberIndexType()!),
+    ],
+    // eslint-disable-next-line @typescript-eslint/internal/prefer-ast-types-enum
```
bradzacher, line 37: @JoshuaKGoldberg - Originally this rule's purpose was to prevent code like `node.type === 'string'`.
Given we now have the `no-unsafe-enum-comparison` rule - do you think this internal rule still has value? WDYT about deleting it? (we already have nuec turned on via the recommended config)

bradzacher (approved):

packages/eslint-plugin/src/rules/restrict-template-expressions.ts
JoshuaKGoldberg: Yeah I'd be in favor of that. It wouldn't be the first example of an internal lint rule ending up as a subset of a general good one.

PR #8421, opened 2024-02-09
fix(eslint-plugin): [prefer-find] support ternary branches in prefer-find

### LINKED ISSUE #8379: Bug: [prefer-find] Add support for checking ternaries

https://main--typescript-eslint.netlify.app/rules/prefer-find

### Description

This addresses a minor omission when I did the PR for this rule (https://github.com/typescript-eslint/typescript-eslint/pull/8216).

I forgot to add support for checking ternary branches. I've made a quick branch that would add support. See https://github.com/typescript-eslint/typescript-eslint/commit/aa43ec626a718867568f028e0e148f894d3da739

### Fail

```TypeScript
(Math.random < 0.5 ? [1, 2, 3].filter(x => x % 2 === 0) : [4, 5, 6].filter(x => x % 2 === 0))[0]
```

### Pass

```TypeScript
Math.random < 0.5 ? [1, 2, 3].find(x => x % 2 === 0) : [4, 5, 6].find(x => x % 2 === 0)
```

### Additional Info

_No response_

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8379 

## Overview

Adds support for recursively checking ternary branches.

### DIFF
--- a/packages/eslint-plugin/src/rules/prefer-find.ts
+++ b/packages/eslint-plugin/src/rules/prefer-find.ts
@@ -60,13 +60,6 @@ export default createRule({
       // This is the only reason we're returning a list rather than a single value.
       if (expression.type === AST_NODE_TYPES.ConditionalExpression) {
         // Both branches of the ternary _must_ return results.
-        const alternateResult = parseArrayFilterExpressions(
-          expression.alternate,
-        );
-        if (alternateResult.length === 0) {
-          return [];
-        }
-
         const consequentResult = parseArrayFilterExpressions(
           expression.consequent,
         );
@@ -74,8 +67,15 @@ export default createRule({
           return [];
         }
 
+        const alternateResult = parseArrayFilterExpressions(
+          expression.alternate,
+        );
+        if (alternateResult.length === 0) {
+          return [];
+        }
+
         // Accumulate the results from both sides and pass up the chain.
-        return [...alternateResult, ...consequentResult];
+        return [...consequentResult, ...alternateResult];
       }
 
       // Check if it looks like <<stuff>>(...), but not <<stuff>>?.(...)
--- a/packages/eslint-plugin/src/rules/prefer-find.ts
+++ b/packages/eslint-plugin/src/rules/prefer-find.ts
@@ -41,20 +41,41 @@ export default createRule({
       filterNode: TSESTree.Node;
     }
 
-    function parseIfArrayFilterExpression(
+    function parseArrayFilterExpressions(
       expression: TSESTree.Expression,
-    ): FilterExpressionData | undefined {
+    ): FilterExpressionData[] {
       if (expression.type === AST_NODE_TYPES.SequenceExpression) {
         // Only the last expression in (a, b, [1, 2, 3].filter(condition))[0] matters
         const lastExpression = nullThrows(
           expression.expressions.at(-1),
           'Expected to have more than zero expressions in a sequence expression',
         );
-        return parseIfArrayFilterExpression(lastExpression);
+        return parseArrayFilterExpressions(lastExpression);
       }
 
       if (expression.type === AST_NODE_TYPES.ChainExpression) {
-        return parseIfArrayFilterExpression(expression.expression);
+        return parseArrayFilterExpressions(expression.expression);
+      }
+
+      // This is the only reason we're returning a list rather than a single value.
+      if (expression.type === AST_NODE_TYPES.ConditionalExpression) {
+        // Both branches of the ternary _must_ return results.
+        const alternateResult = parseArrayFilterExpressions(
+          expression.alternate,
+        );
+        if (alternateResult.length === 0) {
+          return [];
+        }
+
+        const consequentResult = parseArrayFilterExpressions(
+          expression.consequent,
+        );
+        if (consequentResult.length === 0) {
+          return [];
+        }
+
+        // Accumulate the results from both sides and pass up the chain.
+        return [...alternateResult, ...consequentResult];
       }
 
       // Check if it looks like <<stuff>>(...), but not <<stuff>>?.(...)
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/prefer-find.ts
```diff
@@ -41,20 +41,41 @@ export default createRule({
...
-        return parseIfArrayFilterExpression(lastExpression);
+        return parseArrayFilterExpressions(lastExpression);
       }
 
       if (expression.type === AST_NODE_TYPES.ChainExpression) {
-        return parseIfArrayFilterExpression(expression.expression);
+        return parseArrayFilterExpressions(expression.expression);
+      }
+
+      // This is the only reason we're returning a list rather than a single value.
```
JoshuaKGoldberg, line 60: [Praise] I suspect from the code + comments + general flow of this rule that this was probably annoying to implement on your end. Thanks for going through that annoyance. 😄 

(I would be annoyed at it)

packages/eslint-plugin/tests/rules/prefer-find.test.ts
```diff
@@ -70,6 +70,12 @@ ruleTester.run('prefer-find', rule, {
     `,
     "[1, 2, 3].filter(x => x)[Symbol('0')];",
     "[1, 2, 3].filter(x => x)[Symbol.for('0')];",
+    '(Math.random < 0.5 ? [1, 2, 3].filter(x => true) : [1, 2, 3])[0];',
+    `
+      (Math.random < 0.5
```
JoshuaKGoldberg, line 75: [Testing] Typo:

```suggestion
    '(Math.random() < 0.5 ? [1, 2, 3].filter(x => true) : [1, 2, 3])[0];',
    `
      (Math.random() < 0.5
```

packages/eslint-plugin/tests/rules/prefer-find.test.ts
```diff
@@ -584,5 +590,72 @@ arr.find(f, thisArg);
         },
       ],
     },
+    {
```
JoshuaKGoldberg, line 593: [Testing] Just to be through, could you also add test for a desired report with and within a ternary? Like the `valid` case above, but either with two `.filter`s or with the `[0]` on the other side of the last `)`?

```ts
(Math.random() < 0.5
  ? [1, 2, 3].filter(x => false)
  : [1, 2, 3].filter(x => true))[0];
```

```ts
(Math.random() < 0.5
  ? [1, 2, 3].find(x => true)
  : [1, 2, 3].filter(x => true)[0]);
```

The cases lower down are a bit more complex. It's nice to have more straightforward ones to use as a reference too.

packages/eslint-plugin/tests/rules/prefer-find.test.ts
```diff
@@ -584,5 +590,72 @@ arr.find(f, thisArg);
...
+      code: `
+declare const f: any, g: any;
```
JoshuaKGoldberg, line 595: [Testing] Nit: here and below, if there's no need to use an `any`, might as well avoid it:

```suggestion
declare const f: unknown, g: unknown;
```

JoshuaKGoldberg (changes requested):
Code changes look great, nicely done! 👏 

Just requesting changes on a bit more testing.

A League of Legends character dabbing in front of purple smoke and flames. Another character is fly-kicking down towards them from the air.

packages/eslint-plugin/tests/rules/prefer-find.test.ts
kirkwaiblinger: it would have a type error without. I was just too lazy to write out the actual type of the filter callback (and the rule doesn't care about the type of the filter callback) so i went for the easiest way to not error :)

packages/eslint-plugin/src/rules/prefer-find.ts
kirkwaiblinger: haha actually i was pleasantly surprised to find that it was quite straightforward :) I was scared there would be complications with overlapping fixes or something, but everything just worked.

packages/eslint-plugin/tests/rules/prefer-find.test.ts
JoshuaKGoldberg: Ha, gotcha. #8298 -> #8351 tracks making tests error on type unhappiness. cc @auvred - do you think we should start asking folks to try to not have type errors?

JoshuaKGoldberg (approved):
LGTM, nice! Since it's tricky logic being touched, will leave open for a bit under `1 approval`.

packages/eslint-plugin/tests/rules/prefer-find.test.ts
auvred: Personally, I'm +1 to avoid semantic errors, at least in newly added tests :)

Especially when it's not that hard:
```diff
- declare const f: any, g: any;
+ declare const f: (x: unknown) => boolean, g: (x: unknown) => boolean;
```

auvred (dismissed):
LGTM :rocket:

packages/eslint-plugin/tests/rules/prefer-find.test.ts
kirkwaiblinger: Fair enough, changed

packages/eslint-plugin/tests/rules/prefer-find.test.ts
```diff
@@ -584,5 +590,142 @@ arr.find(f, thisArg);
...
+            },
+          ],
+        },
+      ],
+    },
+
+    {
+      code: `
+declare const spreadArgs: [(x: unknown) => boolean];
+[1, 2, 3].filter(...spreadArgs).at(0);
```
kirkwaiblinger, line 712: not related, just an extra test I wish I'd written on the initial implementation

auvred (approved):
LGTM! Thanks for updating tests!

Man claps his hands

PR #8443, opened 2024-02-12
feat(eslint-plugin): [require-types-exports] add new rule

### LINKED ISSUE #7670: Rule proposal: Export all types used in exports

### Description

If a file exports a construct that relies on some other types, but doesn't export those other types, it can be difficult for other files to declare types based on the construct.

For example, suppose a `./somewhere` file exports a `doSomething` function with a single parameter of type `SomeSettings`, but doesn't export `SomeSettings`. What if a consuming file wants to declare a local/parameter of that same `SomeSettings` type? There'd be no easy way to refer to `SomeSettings`.

```ts
import { doSomething } from "./somewhere";

export function callDoSomethingAndLog(settings: Parameters<typeof doSomething>[0]) {
  doSomething(settings);
  console.log("Done!");
}
```

I see this a lot in definitions for libraries. Example: https://github.com/saurabhnemade/react-twitter-embed/issues/132

### Fail Cases

```typescript
interface SomeSettings {
  something: string;
}

export function something(settings: SomeSettings) {
  // ...
}
```

### Pass Cases

```typescript
export interface SomeSettings {
  something: string;
}

export function something(settings: SomeSettings) {
  // ...
}
```

### Additional Info

...I could have sworn this was already filed somewhere, but I couldn't find it after a few minutes of searching. 🤷

StyleShit:
Closes #7670 

___
_Dear Reviewer,_
_Good luck on your journey!_

 **StyleShit.**


packages/eslint-plugin/tests/rules/require-types-exports.test.ts
```diff
@@ -0,0 +1,1217 @@
...
+          messageId: 'requireTypeExport',
+          line: 4,
+          column: 37,
+          endColumn: 39,
+          data: {
+            name: 'Arg',
+          },
+        },
+      ],
+    },
```
StyleShit, line 827: Help would be greatly appreciated here ❤️

JoshuaKGoldberg: Sorry, I'm not clear: what's the request?

JoshuaKGoldberg: Have you tried using the scope manager? That's our normal go-to for answering questions like _"where is this referenced?"_.

https://typescript-eslint.io/play#ts=5.3.3&showAST=scope&fileType=.tsx&code=C4TwDgpgBAggTgcygXigOwK4FsBGE4DcAUERAB5gD2cwUAZhmgMbACWla9APACpTnAIaACYBnWIgB8ACgCGALig8AlIoBulVsKgBvKAF8iQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

StyleShit: Scope manager isn't for variables?
I used it here for something else:

https://github.com/StyleShit/typescript-eslint/blob/b309b51e1ae3246441e31e77724a5046f850f2d7/packages/eslint-plugin/src/rules/require-types-exports.ts#L82-L105

(or we're talking about different things? can you direct me to the APIs you're referring to? or to a code example in the codebase?)

JoshuaKGoldberg: Yup exactly - it sees all (type and variables) and knows all (of their referenced scopes).

StyleShit: (It's ok, you've already helped me a lot here with those few words 🙏)

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
StyleShit, line 1: OK
so I got into a rabbit hole here...
to which level should we follow the exported functions?

should we support things like this?
```TS
function func1() { /* ... */ }
const func2 = () => { /* ... */ };

export default {
	func1,
	func2,
	func3: () => { /* ... */ },
	obj: {
		func4: () => { /* ... */ }
	}
};
```

or this?

```TS
function func1() { /* ... */ }
const func2 = () => { /* ... */ };

const functions = {
	func1,
	func2,
	func3: () => { /* ... */ },
	obj: {
		func4: () => { /* ... */ }
	}
};

export default functions;
```

and to which level should we follow the types? 
```TS
// (-- imagine the same nesting but with types --)
```

StyleShit:
No idea why the lint/tests are failing, seems to be fine... Am I missing something?

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit, line 1: OK
so I got into a rabbit hole here...
to which level should we follow the exported functions?

should we support things like this?
```TS
function func1() { /* ... */ }
const func2 = () => { /* ... */ };

export default {
	func1,
	func2,
	func3: () => { /* ... */ },
	obj: {
		func4: () => { /* ... */ }
	}
};
```

or this?

```TS
function func1() { /* ... */ }
const func2 = () => { /* ... */ };

const functions = {
	func1,
	func2,
	func3: () => { /* ... */ },
	obj: {
		func4: () => { /* ... */ }
	}
};

export default functions;
```

and to which level should we follow the types? 
```TS
// (-- imagine the same nesting but with types --)
```

packages/eslint-plugin/src/rules/require-types-exports.ts
```diff
@@ -0,0 +1,408 @@
...
+      requireTypeExport: 'Expected type "{{ name }}" to be exported',
+    },
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    const exportedTypes = new Set<string>();
+    const reported = new Set<string>();
+
+    function collectExportedTypes(program: TSESTree.Program): void {
```
Omerisra6, line 35: You also need to check for imported types:
```ts
import Arg from './types';

// This shouldn't report
export function f(arg: Arg): void {}
```

StyleShit: Good catch, fixed, thanks! 😄

packages/eslint-plugin/src/rules/require-types-exports.ts
```diff
@@ -0,0 +1,403 @@
...
+      if (typeReference.typeName.type === AST_NODE_TYPES.Identifier) {
+        return typeReference.typeName.name;
+      }
+
+      return '';
+    }
+
+    return {
+      'ImportDeclaration[importKind="type"] ImportSpecifier, ImportSpecifier[importKind="type"]':
+        collectImportedTypes,
```
StyleShit, line 379: Currently, it doesn't support this:

```ts
import { Type } from './types';

export function (arg: Type) { /* ... */ }
```

but only things like this:

```ts
import type { Type1 } from './types';
import { type Type2 } from './types';
import { type Type3 as Type4 } from './types';
```

because IDK if I can know for sure if the imported name is a type or not
any idea if this could be done?

packages/eslint-plugin/docs/rules/require-types-exports.md
```diff
@@ -0,0 +1,75 @@
+---
+description: 'Require exporting types that are used in exported functions declarations.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/require-types-exports** for documentation.
+
+When exporting functions from a module, it is recommended to export also all the
```
JoshuaKGoldberg, line 9: > functions

[Feature] It's not just functions though. It could also be, say, variables:

```ts
// source.ts
interface MyType {
  apple: boolean;
}

const value: MyType = { apple: true };

export default value;
```

```ts
import value from "./source.js";

let myValue; // What type is this?!
myValue = value;
```

packages/eslint-plugin/docs/rules/require-types-exports.md
```diff
@@ -0,0 +1,75 @@
...
+types that are used in the function declarations. This is useful for consumers of
+the module, as it allows them to use the types in their own code without having to
+use things like [`Parameters`](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype)
```
JoshuaKGoldberg, line 12: [Docs] Nit: a bit more precise...

```suggestion
use helper types like [`Parameters`](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype)
```

JoshuaKGoldberg (changes requested):
Lovely start! I've wanted this for ages, so very excited to see you working on it 😄.

Leaving a preliminary review as I think it'll need to be expanded to at least variables. Really, anything that can refer to a type: classes, interfaces / type aliases, namespaces, ...

Good luck on _your_ journey. 😜

packages/eslint-plugin/src/rules/require-types-exports.ts
JoshuaKGoldberg: The type checker can determine whether something is in type space and/or value space.

But for that case of `export function (arg: Type) { /* ... */ }`, do we need to know whether it's type and/or value? I'd say that should be a complaint in the rule no matter what, no?

JoshuaKGoldberg:
👋 just checking in @StyleShit, are you still able to put time towards this? And/or, did I miss something blocking it?

StyleShit:
Thanks for the ping, I'm hoping to be able to get to this in a couple of weeks 😄

JoshuaKGoldberg:
I'm going to go ahead and close this out as it's been about 4 weeks with no activity. Appreciate you getting started on it @StyleShit - if you end up having the time to continue, no worries, either ping us to reopen this PR or just send a new one.

In the meantime, if anybody else wants to work on the linked issue, go right ahead. Just make sure to add a co-author attribution if it reuses code from this PR.

Cheers!

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: Done

packages/eslint-plugin/docs/rules/require-types-exports.md
StyleShit: Nittier: "utility types" 😛

StyleShit:
Hey @JoshuaKGoldberg, I keep working on it, can you please re-open? 😄

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: OK,
After thinking about it for a little bit...

What should we do in cases where someone imports a type from another module and uses it in their exported function?

```TS
// my-package/src/types.ts
export type A = number;
```

```TS
// my-package/src/function.ts
import type { A } from './types';

export function f(): A {
	return 1;
}
```

```TS
// my-package/src/index.ts
export { f } from './function';
```

Then, in user-land:
```TS
// my-code.ts
import { f } from 'my-package';

let value; // What type is it?
value = f();
```

I mean... on the one hand, we can't know whether the type is exported somehow to the user.
On the other hand, (I think?) we don't want to enforce everyone to (re)export the types they imported 😓

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: Started working on this
There is this magical `Scope.through` thing that basically saved me hours of work! ✨ 
https://github.com/StyleShit/typescript-eslint/commit/1812e37f6501844fc72fc1d4ae006300464a8e34

Is there anything similar for variables, before I start going crazy with those infinite conditions again?

To make it clear - I'm looking for something that extracts the type references from this for example:
```TS
type T1 = number;
type T2 = boolean;

export const value: { a: { b: { c: T1 } } } | [string, T2 | null]= {
	a: {
		b: {
			c: 1,
		}
	}
}
```

StyleShit: In addition, is there any utility to run recursively on objects like the one you mentioned above? (`functions.obj.func4()`)

packages/eslint-plugin/docs/rules/require-types-exports.md
StyleShit: updated the docs

packages/eslint-plugin/src/rules/require-types-exports.ts
JoshuaKGoldberg: On a [pure scope analysis level](https://typescript-eslint.io/packages/scope-manager), we do know this. The `A` identifier in `function.ts` is part of what's exported with `f()`.

I think it'd be very preferable to stick with the pure scope analysis strategy. Dipping into types land is much harder to work with, since `type` aliases sometimes become new things and sometimes are just references to existing things.

Does that answer what you're looking for?

packages/eslint-plugin/src/rules/require-types-exports.ts
JoshuaKGoldberg: To be honest, I'm not super practiced with [`@typescript-eslint/scope-manager`](https://typescript-eslint.io/packages/scope-manager) myself 😅 and don't have a great answer for you. Sorry. I'd say your best bet is to play around with the APIs and the [scopes on the playground for your snippet](https://typescript-eslint.io/play/#ts=5.4.3&showAST=scope&fileType=.tsx&code=C4TwDgpgBAKgjFAvFAdgVwLYCMICcDcAUKJLAExJRYD21ANhAIYpGEQAeY1uwUAxtRQBnXgDdGdNBABcUAN5RGshVmX9Z8KAF9tugD5QA2iNwBLFAHMANOSgH0dOgF1KcwlEXL3Hql58%2B%2BWTgrbw8tb3DwoA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false).

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: I think the link here is wrong?

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: ugh, that's a bummer

(And I play with the scope manager playground, that's how I [mistakenly] figured out the current usage 😄)

packages/eslint-plugin/src/rules/require-types-exports.ts
JoshuaKGoldberg: I meant to link to https://typescript-eslint.io/packages/scope-manager/. As in, the thing you referenced using in another comment.

StyleShit: Oh,
so I don't understand what you meant in the previous comment 🫣

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: > Is there anything similar for variables, before I start going crazy with those infinite conditions again?

I have a stupid idea that works, not sure whether it's acceptable 😂 

will probably need a review on that before I use the same approach for the rest:
https://github.com/typescript-eslint/typescript-eslint/commit/e57985a6fdbd710c8deffd8ad34b52054e3b8422

packages/eslint-plugin/src/rules/require-types-exports.ts
JoshuaKGoldberg: Oh, sorry! I'm being unclear. Answering more specifically:

> I mean... on the one hand, we can't know whether the type is exported somehow to the user.

I believe we can. In the case of a function like `export function f(): A`, we know that it's exported and that within its return type exists an identifier `A`. The rule would need to check: for every identifier in the return type, is that also exported in the current file (module)?

> On the other hand, (I think?) we don't want to force everyone to (re)export the types they imported

I think we do, and that that's the goal of this rule! In the case of the `f(): A`, it'd be something like:

```ts
// my-package/src/function.ts
import type { A } from './types';

+ export { A };

export function f(): A {
	return 1;
}
```

That does leave a "hole" in the rule, that `my-package/src/index.ts` exports `f` but not `A`. In theory we could dip into the type checker to know this, and then report... but really that's (a) a lot of work and (b) dipping into TypeScript's ["inferred type cannot be named without a reference"](https://stackoverflow.com/questions/72041763/typescript-inferred-type-cannot-be-named-without-reference) territory. I'll defer to @typescript-eslint/triage-team on this one, but I'm ok with the rule not using type checking.

Something like [Knip](https://knip.dev) would IMO be better suited to cross-file type checking things.

Does that answer what you're looking for?

StyleShit: > The rule would need to check: for every identifier in the return type, is that also exported in the current file (module)?

My current logic is:
"Does this type exist in the outside world (whether it's imported/exported in the current file)?"
	- If so: Great!
	- Otherwise: Report

> I think we do, and that that's the goal of this rule! 

I don't fully agree
I mean... yeah, that's the goal of the rule, but the type might already be exported through the index file, for example, and we'll force developers to export it multiple times in their code.

Let's take my example above and modify it a little bit:

```TS
// my-package/src/types.ts
export type A = number;
```

```TS
// my-package/src/f1.ts
import type { A } from './types';

export function f1(): A {
	return 1;
}
```

```TS
// my-package/src/f2.ts
import type { A } from './types';

export function f2(): A {
	return 2;
}
```

```TS
// my-package/src/index.ts
export { f1 } from './f1';
export { f2 } from './f2';

export * from './types';
```

Then, in user-land:
```TS
// my-code.ts
import { f1 } from 'my-package';
import type { A } from 'my-package';

let value: A; // I know what the type is! YAY!
value = f1();
```

This code is totally fine. But with the changes you suggest, it'll have 2 more redundant exports (and it'll increase linearly based on the type usage inside other modules).
Yeah, it probably won't hurt anybody, but it will make the DX awful, don't you think?

JoshuaKGoldberg: If we don't enforce that a `export function fn(): A;` is in a file that also exports `A`, what is the point of this rule?

JoshuaKGoldberg: For lint rules, the line generally lies with the single file being linted. Because lint rules operate on a single file at a time, that naturally falls out of the architecture.

It's pretty common that lint rules will only be able to catch issues within the small scope of a single rule at a time. For example, `no-unused-vars` won't detect things that are exported from a file but never imported. Hence [Knip](https://knip.dev).

StyleShit: do you wanna loop in the triage team so we can agree on something together?

JoshuaKGoldberg: Already did 🙂 just waiting for them to respond.

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: bump?
will be glad to get some directions here, I'm lost...

I have an awful WIP code locally that collects all the type references from the Program, puts them in a Set, and tries to guess the connection between them and the functions' scopes they're related to
(hint: it's ugly and doesn't work 😓)

packages/eslint-plugin/src/rules/require-types-exports.ts
```diff
@@ -0,0 +1,241 @@
...
+
+      reportedTypes.add(name);
+    }
+
+    function getTypeName(typeReference: TSESTree.TSTypeReference): string {
+      if (typeReference.typeName.type === AST_NODE_TYPES.Identifier) {
+        return typeReference.typeName.name;
+      }
+
+      return '';
```
JoshuaKGoldberg, line 182: From the ast spec:

```ts
export declare type EntityName = Identifier | ThisExpression | TSQualifiedName;
```

This'll need to handle `this` and namespace types, no?

```ts
export class Wrapper {
  work(other: this) {}
}
```

packages/eslint-plugin/src/rules/require-types-exports.ts
JoshuaKGoldberg: Heya sorry, was out - when you say "doesn't work", what do you mean? 

Re https://github.com/typescript-eslint/typescript-eslint/commit/e57985a6fdbd710c8deffd8ad34b52054e3b8422#diff-c33471d1953c12297755643e4d460779c5a5323f4535a87cba96ff86abeb40d4R140-R142:

> ```ts
>      typeReferences.forEach(r => {
>         // TODO: Probably not the best way to do it...
>         if (isLocationOverlapping(r.identifier.loc, node.loc)) {
> ```

You might be able to check `ts.Node`s by referential equality. Like, if `r.identifier === node`, or some similar thing that maybe involves a `.parent`.

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: Yeah... IDK why I haven't looked at spec before writing this 😅 
Thanks!

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: by "not working" I mean "I can't make it guess properly" haha

regarding the referential equality check - I managed to make it work thanks to your direction, thanks!

anyway, I still can't find a way to get the types used in an exported object properly 😓 

```TS
import type { A } from './types';

interface Arg1 {
  key: number;
}

interface Arg2 {
  key: string;
}

interface Arg3 {
  key: boolean;
}

type Ret1 = number;
type Ret2 = string;
type Ret3 = boolean;

const func1 = (arg: Arg1): Ret1 => 1;

function func2(arg: Arg2): Ret2 {
  return 'apple';
}

type Apple = string;

const apple: Apple = 'apple';

export const functions = {
  func1,
  func2,
  path: {
    to: {
      func3: (arg: Arg3): Ret3 => true,
      apple,
    },
  },
};
```

I know I can get the full type as a string using `typeToString()`, but is there a way to get it as an object of nodes/types or something? because seems like this might help me figure out which types have been used

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
```diff
@@ -0,0 +1,1985 @@
...
+      export function f(arg: Record<PropertyKey, Promise<string>>) {
+        return arg;
+      }
+    `,
+
+    `
+      export function f<T extends Record<PropertyKey, Promise<string>>>(arg: T) {
+        return arg;
+      }
+    `,
```
StyleShit, line 365: Those 2 tests are failing for some reason. TBH, it was an accidental discovery.
Seems like `isSymbolFromDefaultLibrary(PropertyKey)` returns false for some reason. Am I doing something wrong?

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
```diff
@@ -0,0 +1,2007 @@
...
+
+    `
+      export function f<T extends ReturnType<() => string>>(arg: T) {
+        return arg;
+      }
+    `,
```
StyleShit, line 371: This test is failing for some reason. Seems like this code resolves to `undefined` for `ReturnType`:
```TS
const symbol = type.aliasSymbol ?? type.getSymbol();
```

Am I doing something wrong?

packages/eslint-plugin/src/rules/require-types-exports.ts
JoshuaKGoldberg: I think you'll have to do this manually: a function that takes in an AST node, descends through the AST node, and collects all the types (identifiers & object literals). I don't know of any existing utilities to do this. 😞

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
JoshuaKGoldberg: Putting a breakpoint on that line of code this in the debugger, I get:

```plaintext
> services.program.getTypeChecker().typeToString(type)
'string'
```

I'm inferring that the type is the result of evaluating `ReturnType<() => string>`, which reduces to `string`. Which actually makes sense to me!

...I don't have a clear thought in my mind as to how to solve for this though. If we skip reporting when `!symbol`, then test cases like this one fail:

```ts
type Arg = number;

const a = function (a: Arg): void {};

export default a;
```

I wonder if the conclusion we should make is to not use the type checker at all? As in, if a type isn't declared or visible from the current file, it's either...

* ...a built-in one
* ...an error type
* ...something globally augmented in

...and for all of those, I _think_ it's fine to not report?

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: Uh, that's a bummer...

Anyway, I'll need to find the time to write that function, sounds like a lot of work 🫣

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
StyleShit: well, that's an interesting take
sounds good to me (at least in theory)

StyleShit: Done

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: Never mind, found the time 😄 

I think the code is much more decent now, thanks!!

JoshuaKGoldberg: Good luck on your journey 😜

StyleShit: Haha
It's basically ready for review from my end, but I can't re-request review for some reason. Could you please take a look?

Disclaimer: pretty sure I've missed some cases in the new function, so I'll be glad to get some test cases you think I should add

StyleShit:
@JoshuaKGoldberg Question:

Should `B` be reported here?
```TS
type A = number;
type B = string;

export function func1<R extends A>(arg: R): R {
  doWork(String(arg));

  return arg;
}

declare function doWork(arg: B): void;
```

StyleShit:
OK, I've kept working on this and found some more cases, but I have absolutely no idea how to work with cases like [this](https://typescript-eslint.io/play/#ts=5.4.5&fileType=.tsx&esQuery=N4IgZglgNgLgpgJxALhCAvkA&code=C4TwDgpgBAglC8UB2BXAtgIwgJwNwChRIoAhBKAZ2GwEskBzAo6AYXIwHsOAbCAQyQF8AYw5IqUDhgBWALigBvKAGsIIeUuAd5MADRQBAE3lsAvlHOIF%2BKCrUabtqFvkBGXY9tH51FBA%2B2ph6mQqLiwJIyAEzk1rYY8gDkfBSGiQYUpMH4%2BBAAHmAc2BEAZihIwsA0YlBlFQAUAJSKjjQlUPUAsnzAABYAdNhGHGhNUAB8UAAM-QCszXFO2BDAKNhIkdL9qiAEgTm2bR3dfYPDo80APNNzC55Qy6vrm9tq-Vp7FgcPK2sbUlsdu8OFAANSbKL9DBgqDJVKJAimIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

It should report A & B, but not C. It _should<sup>TM</sup>_ be easy to just report all the types that are being used inside the objects (which means A, B & C), but that's probably not what we want.

I'm kinda re-writing the whole TS inferring system by myself 😅 

Do I need to? Should I stop already? 😂

JoshuaKGoldberg:
> Should `B` be reported here?

I don't think it should be. It's not visible in anything exported from the file. That file's `.d.ts` equivalent is roughly:

```ts
type A = number;

export function func1<R extends A>(arg: R): R;
```

...so only `A` should be reported by the rule.

JoshuaKGoldberg:
> cases like [this](https://typescript-eslint.io/play/#ts=5.4.5&fileType=.tsx&esQuery=N4IgZglgNgLgpgJxALhCAvkA&code=C4TwDgpgBAglC8UB2BXAtgIwgJwNwChRIoAhBKAZ2GwEskBzAo6AYXIwHsOAbCAQyQF8AYw5IqUDhgBWALigBvKAGsIIeUuAd5MADRQBAE3lsAvlHOIF%2BKCrUabtqFvkBGXY9tH51FBA%2B2ph6mQqLiwJIyAEzk1rYY8gDkfBSGiQYUpMH4%2BBAAHmAc2BEAZihIwsA0YlBlFQAUAJSKjjQlUPUAsnzAABYAdNhGHGhNUAB8UAAM-QCszXFO2BDAKNhIkdL9qiAEgTm2bR3dfYPDo80APNNzC55Qy6vrm9tq-Vp7FgcPK2sbUlsdu8OFAANSbKL9DBgqDJVKJAimIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

That file's `.d.ts` equivalent is:

```ts
type A = number;
type C = boolean;

export function func(): { to: A, and: C } | number | string;
```

...so it should only report `A` and `C` as needing to be exported, no? Am I missing something?

packages/eslint-plugin/docs/rules/require-types-exports.mdx
JoshuaKGoldberg, line 1: @StyleShit let's move to threaded conversations, the root level discussion on this PR is getting kind of messy...

From https://github.com/typescript-eslint/typescript-eslint/pull/8443#issuecomment-2198328764:

> Should `B` be reported here?

I don't think it should be. It's not visible in anything exported from the file. That file's `.d.ts` equivalent is roughly:

```ts
type A = number;

export function func1<R extends A>(arg: R): R;
```

...so only `A` should be reported by the rule. Does that seem right to you?

JoshuaKGoldberg, line 1: From https://github.com/typescript-eslint/typescript-eslint/pull/8443#issuecomment-2198588243:

> cases like [this](https://typescript-eslint.io/play/#ts=5.4.5&fileType=.tsx&esQuery=N4IgZglgNgLgpgJxALhCAvkA&code=C4TwDgpgBAglC8UB2BXAtgIwgJwNwChRIoAhBKAZ2GwEskBzAo6AYXIwHsOAbCAQyQF8AYw5IqUDhgBWALigBvKAGsIIeUuAd5MADRQBAE3lsAvlHOIF%2BKCrUabtqFvkBGXY9tH51FBA%2B2ph6mQqLiwJIyAEzk1rYY8gDkfBSGiQYUpMH4%2BBAAHmAc2BEAZihIwsA0YlBlFQAUAJSKjjQlUPUAsnzAABYAdNhGHGhNUAB8UAAM-QCszXFO2BDAKNhIkdL9qiAEgTm2bR3dfYPDo80APNNzC55Qy6vrm9tq-Vp7FgcPK2sbUlsdu8OFAANSbKL9DBgqDJVKJAimIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

That file's `.d.ts` equivalent is:

```ts
type A = number;
type C = boolean;

export function func(): { to: A, and: C } | number | string;
```

...so it should only report `A` and `C` as needing to be exported, no? Am I missing something?

packages/eslint-plugin/src/rules/require-types-exports.ts
JoshuaKGoldberg: You got it! I started a couple of threads in the `.mdx` file (arbitrarily, since it's the first one in the PR).

packages/eslint-plugin/src/rules/require-types-exports.ts
```diff
@@ -0,0 +1,386 @@
...
+    functionNode.type === AST_NODE_TYPES.ArrowFunctionExpression &&
+    functionNode.body.type === AST_NODE_TYPES.Identifier;
+
+  if (isArrowFunctionReturn) {
+    return new Set([functionNode.body]);
+  }
+
+  const returnStatements = new Set<TSESTree.Node>();
+
+  simpleTraverse(functionNode, {
```
JoshuaKGoldberg, line 358: [Refactor] I don't think `simpleTraverse` is what we should use here. It's tailored to the way it's currently used in `packages/typescript-estree`:

* There's no way to break, but it looks like this usage would want to halt traversal after finding a `ReturnStatement`
* It allows configurable selectors with `'enter'`, but this code just uses `ReturnStatement`

I think it'd be better to avoid `simpleTraverse` here, and instead do some or all of:
* Look into the existing `forEachReturnStatement`, maybe it's exactly what you need?
* Failing that, write a dedicated simpler traverser for just this rule that halts on `ReturnStatement`s, omitting anything around parent pointers or `'enter'`
* Reuse some of the design and logic from one of the rules that does similar things: `explicit-module-boundary-types`, `explicit-function-return-type`
  * This one I'm low confidence in, but figured I'd mention...

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
JoshuaKGoldberg, line 1: [Testing] I ran `yarn lint` in the root and looked at the reported violations. It was very informative, would recommend!

Some highlights of cases that aren't reported now but I think should be (am I wrong about any of these?)...

* Exported types that contain other types:
  
  ```ts
  // Reduction of member-ordering.ts:
  type BaseMemberType = 'abc';
  export type MemberType = BaseMemberType | BaseMemberType[];
  ```
  
  ```ts
  // Loosely inspired by member-ordering.ts:
  type Accessibility = 'abc';
  export type MemberType = `${Accessibility}`;
  ```

...and a false positive!

* References to functions declared later on:

  ```ts
  // Reduction of getOperatorPrecedence.ts:
  import type { TSESTree } from '@typescript-eslint/utils';
  import { SyntaxKind } from 'typescript';
  
  import type { ValueOf } from './types';
  
  export function getOperatorPrecedenceForNode(): number {
    return getBinaryOperatorPrecedence(SyntaxKind.AbstractKeyword);
  }
  
  export type TSESTreeOperatorKind =
    | ValueOf<TSESTree.BinaryOperatorToText>
    | ValueOf<TSESTree.PunctuatorTokenToText>;
  
  export function getBinaryOperatorPrecedence(
    kind: SyntaxKind | TSESTreeOperatorKind,
    //                 ~~~~~~~~~~~~~~~~~~~~
    // Expected type "TSESTreeOperatorKind" to be exported
  ): number {
    return kind ? -1 : 1;
  }
  ```

packages/eslint-plugin/src/rules/require-types-exports.ts
```diff
@@ -0,0 +1,386 @@
...
+export default createRule<[], MessageIds>({
+  name: 'require-types-exports',
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Require exporting types that are used in exported entities',
+      recommended: 'strict',
+    },
+    messages: {
+      requireTypeExport: 'Expected type "{{ name }}" to be exported',
```
JoshuaKGoldberg, line 22: [Docs] What do you think of this, for a more informative error?

```suggestion
      requireTypeExport: '"{{ name }}" is used in other exports from this file, so it should also be exported.',
```

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
JoshuaKGoldberg, line 1: [Testing] I'd think `keyof` and/or `typeof` types should be flagged too:

```ts
const fruits = { apple: true };

export function getFruit<Key extends keyof typeof fruits>(
  key: Key,
): (typeof fruits)[Key] {
  return fruits[key];
}
```

Thoughts?

JoshuaKGoldberg (changes requested):
Progress! 🚀 

I think this is making great headway through the slog of edge cases and tests that such a nuanced rule would need. Very nice to see.

Requesting changes on:

* Edge cases: left a few comments suggesting them, marked as `[Bug]` or `[Testing]`
* Structure: streamlining traversals a bit to not do extra work (avoiding `simpleTraverse`)

I also started a thread for any open questions I saw. If I'm missing one of yours, sorry, please yell at me. 🙂

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
JoshuaKGoldberg: Oh, there were more I just noticed.

* Globals shouldn't be reported on, I don't think:
  
  ```ts
  // Reduction of test-utils.ts:
  export function createSnapshotTestBlock(
    code: string,
    config: TSESTreeOptions,
    generateServices?: true,
  ): jest.ProvidesCallback {
  // ~~~~~~~~~~~~~~~~~~~~~
  // Expected type "jest" to be exported
  ```

  Also in this file, adding an `export` before `type UnknownObject` doesn't stop it from being reported on. Bug?

packages/eslint-plugin/docs/rules/require-types-exports.mdx
StyleShit: yeah, sorry, I've mistyped the type names there, it should be A & C as you said,

Anyway... I didn't explain myself properly (it was after a lot of work on this, my brain basically melted 🧠)


1. What I meant is that it (at least in my head and with my current implementation) becomes overly complicated to collect types from things like this:

```TS
type A = number;

const obj: { path: { to: { key: A } } } = {
  path: { to: { key: 1 } }
};

export function func() {
  return obj.path.to.key;
}
```

Because I need to somehow follow the whole path of the `MemberExpression` and try to infer its value from the root object. Maybe that's fine and easy and I'm just making it hard in my mind? 🤷 

2. Should A be reported in this case?
```TS
type A = number;

const obj = {
  value: 1 as A,
};

export function func() {
  return obj.value;
}
```

StyleShit: Yup, seems right to me, that's what I have in the current logic. Just wanted to make sure I don't miss anything 😄

JoshuaKGoldberg: > Because I need to somehow follow the whole path of the `MemberExpression`

Indeed. I think this is going to need recursive logic to traverse through object literals like that.

> `Should A be reported in this case?`

I'd think so. I don't think we should care about what a `type`'s actual type is - just that it's exported. We'd certainly want to report in this case:

```ts
type A = { inner: string };

const obj = {
  value: { inner: "" } as A,
};

// Inferred return type of func(): A
export function func() {
  return obj.value;
}
```

StyleShit: Not sure whether I understand the difference here

If the type is an alias to a built-in primitive type we shouldn't report, and if it's an object or something we should?

I mean... yeah, TS won't emit `A` in the `.d.ts` file in my example, so if I understand you correctly, I'll need to re-create TS's logic here about whether a type should appear in the function's declaration, right?

JoshuaKGoldberg: > I'll need to re-create TS's logic here about whether a type should appear in the function's declaration, right?

Not that, I'm suggesting the opposite. What I'm proposing is: _any_ `type` that's visible in other exports should be exported itself. That would hold true for `string`, `number | string`, `(number | string)[]`, `{ inner: string }`, etc.

I'm suggesting this in part to try to make this rule easier to implement and maintain 😄. Having to determine whether some `type` is emitted is more work.

But also, folks sometimes use `type`s for [opaque values](https://github.com/LearningTypeScript/site/pull/135):

```ts
// TODO: Switch to a branded type once we decide on a library
type Id = string;
export declare function getId(): Id;
``` 

In this case, I think it'd match the intent of the rule to say `Id` should be exported.

...but anyway, what I'm trying to suggest here is the simplest approach: if something in the AST is visible to an export, we go with it.

StyleShit: Ok, that's exactly what I thought! 

Now I realize that I'm stupid and misunderstood you
You said "I think so", while I read it (for some reason...) "I think no" - totally different meaning 😅

Yeah, that's _much_ easier

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
StyleShit: That's interesting. So we'll need to flag both types and values?

If so - I'll probably need to rework the reporting logic to also support values, and we should probably rename the rule (?)

JoshuaKGoldberg: In this case, it's really the `keyof typeof fruits` that needs to be exported. So I'd assume the squiggly would be on there:

```ts
export function getFruit<Key extends keyof typeof fruits>(
//                                   ~~~~~~~~~~~~~~~~~~~
```

This would certainly make it difficult to write an auto-fixer 😅 glad that that's out of scope for this PR.

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: Created a custom traverser, inspired by the original `forEachReturnStatement`

packages/eslint-plugin/src/rules/require-types-exports.ts
```diff
@@ -0,0 +1,386 @@
...
+
+      reportedTypes.add(name);
+    }
+
+    return {
+      'ImportDeclaration ImportSpecifier': collectImportedTypes,
+      'ImportDeclaration ImportNamespaceSpecifier': collectImportedTypes,
+      'ImportDeclaration ImportDefaultSpecifier': collectImportedTypes,
+
+      'Program > ExportNamedDeclaration > TSTypeAliasDeclaration':
```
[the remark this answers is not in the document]
StyleShit: Added missing tests

packages/eslint-plugin/src/rules/require-types-exports.ts
```diff
@@ -0,0 +1,386 @@
...
+  ScopeType,
+} from '@typescript-eslint/scope-manager';
+import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES, simpleTraverse } from '@typescript-eslint/utils';
+
+import { createRule, findVariable } from '../util';
+
+type MessageIds = 'requireTypeExport';
+
+export default createRule<[], MessageIds>({
```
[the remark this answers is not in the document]
StyleShit: had a wrong selector
fixed 😄

packages/eslint-plugin/src/rules/require-types-exports.ts
StyleShit: I like it
Done

packages/eslint-plugin/tests/rules/require-types-exports.test.ts
StyleShit: > Exported types that contain other types
Done

> References to functions declared later on
Done

> Globals shouldn't be reported on
TBD

StyleShit: I think we need to report only the `typeof` part, because that's what really blocks the developers from using it, no?

Do you think we should have a different message for this case? Maybe something like "typeof needs to be exported. please extract it to a type alias"?

JoshuaKGoldberg: > only the `typeof` part, because that's what really blocks the developers from using it, no?

Well, it's certainly _better_...

```ts
export type Fruits = typeof fruits;

export function getFruit<Key extends keyof Fruits>(
//                                   ~~~~~~~~~~~~
```

...but it strikes me as a kind of arbitrary exception to allow a `keyof ...` to be inline. What would make this so special?

> different message for this case

+1, yes!

StyleShit: > What would make this so special?

IDK, what's the difference between these 2?
```TS
export declare function(key: keyof Items);
export declare function({ items }: { items: Items });
```

In both cases you construct a type inline based on a type reference. So why wouldn't we force every inline type to be extracted to an alias?

What makes `keyof` special compared to a regular `TSTypeLiteral` in this case?

StyleShit: umm... great
now I'll need to redo a lot of my work 😂 

(I didn't understand this until now)

StyleShit: anyway, I _think_ it's gonna simplify _a lot_ of my code

StyleShit:
@JoshuaKGoldberg 

OK, so before I delete most of the code and start basically from scratch, let's go over some cases to make sure we're on the same page?

1. I'm assuming this is fine (easy):
```TS
type A = number;

export declare function func(a: A): void;
//                              ~
``` 

2. Also this one:
```TS
type A = number;

export declare function func<T extends A>(a: T): void;
//                                     ~
``` 

3. And as we said before, this one should also report:
```TS
type A = number;

export declare function func({ a }: { a: A }): void;
//                                  ~~~~~~~~
``` 


4. But what about this one?
```TS
type A = number;

export declare function func([ a ]: [ A ]): void;
//                                  ~~~~~
``` 

5. Or this one?
```TS
type A = number;

export declare function func([ a ]: A[]): void;
//                                  ~~~
``` 

6. You can also have something like this, should it be reported?
```TS
type A = number;

export declare function func(a: Record<string, A>): void;
//                              ~~~~~~~~~~~~~~~~~
``` 

I mean... you can still construct all of them (3-6) easily if you have `A`, so where do we draw the line where we say "ok, this one should be reported"?

Should we just say "everything has to be a type alias, including things like simple unions and arrays"?
(well... it's impractical and probably impossible [think about returning `T | null` for example], so probably no... but I'm talking hypothetically)

In addition, should we also force everyone to have explicit return types on their functions?
It'll be easier to implement & report, but again, will probably have its own issues...

Anyway, my point is - seems like there are a lot of open questions.
Should we go back to the issue and discuss this again? Should we start with a smaller scope that includes some specific cases and expand in the future?
I'm fine with any decision, I just think we need a better definition for this rule.

___
OK, that was a little bit longer than I expected 😅

JoshuaKGoldberg:
> open questions ... go back to the issue and discuss this again?

Yeah, agreed, you bring up a lot of important edge cases. I'll post there.

JoshuaKGoldberg:
👋 Just checking in @StyleShit, is this still something you have time & energy for?

StyleShit:
@JoshuaKGoldberg

Thanks for checking 😄
I have the energy, but unfortunately don't have the time right now due to being in reserve duty.
Anyway, I think I'll open a new PR anyway when I have the time to work on it, because this one is a total mess 😅

JoshuaKGoldberg:
Cool! I'd really like to get this one and #8509 in by the end of the year, they're great rules to have that I've wanted for a while. So let's timebox these two to, say, the end of the month. If you don't end up having time by then, no worries, one of us on the team can send in a PR with the same commit history & a co-author attribution.

Good luck + best wishes on reserve duty. ❤️

PR #7895, opened 2023-11-10
fix(eslint-plugin): [strict-boolean-expression] fix nullable object handling in LogicalExpression

barroudjo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7743
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

When a nullable object appears in a logical expression other than nullish coalescing (so `||` or `&&`), the suggested fix could be wrong and introduce a bug (as explained in the [issue](https://github.com/typescript-eslint/typescript-eslint/issues/7743)). This MR removes the suggested fix for these cases, and gives a more explicit message.

PR #8454, opened 2024-02-13
fix(eslint-plugin): [strict-boolean-expression] fix nullable object handling in LogicalExpression

### LINKED ISSUE #7743: Bug: [strict-boolean-expressions] autofix can be incorrect with allowNullableObject set to false

### Playground Link

https://typescript-eslint.io/play/#ts=5.2.2&fileType=.tsx&code=LAKAxg9gdgzgLgAgGYQgIwIYCcEF4EYwCeUYCAFEgK6kD8AXBQJR4B8CUVANly7uwG9QCBJFiIkASygYuAMRplcwkckUIAPhpUjy5Pu048mAbhVYApnCpYoyabIWl9ZkAF8TQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYkACAXAngA4oDG0AlobgLQrzkB2uA9MrhSTQEYD2PSAQwa0AHoWgpk5Hg1QYA2pETRoPaJAA0mcBCgD48HgHcAyu0YBzdGABm%2B5Ig06IkfYaMA5WAFsuy63bwDk66egbGXgYCXEgA8lwAVoicAfaOzmHukfDRSABCfIIMqUHpoa7hngg5MYhmFAxWGIHBGRVZ1bmIXr7%2BzWkhuu0RnbUAogw%2BJa3lbsYAggz402VDc0YASgiIACo8WwwA6uS4ABY8sLj15JzZAMKnyQDWyACSANIMxoenAriv828ABEeJYVhkIABfAC6kJAkKAA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```typescript
const foobar = async (func?: () => null) => {
  const finalFunc =
    func ||
    (() => null);
  return finalFunc();
};
```

### ESLint Config

```javascript
module.exports = {
  "rules": {"@typescript-eslint/strict-boolean-expressions": ["error", {
    "allowString": false,
    "allowNumber": false,
    "allowNullableObject": false,
    "allowNullableBoolean": false,
    "allowNullableString": false,
    "allowNullableNumber": false,
    "allowNullableEnum": false,
    "allowAny": false,
    "allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing": false
      }]}
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

The auto fix introduces a bug in this case, as it replaces `func || (() => null);` with `func != null  || (() => null);` which is now either a boolean or a function, when the original code was such that it always resulted in a function.

In this case there should be no auto fix, or the actual fix is to use nullish coalescing (`func ?? (() => null);`), but that seems a bit beyond the scope of this rule...

### Actual Result

The auto fix introduces a bug in this case, as it replaces `func || (() => null);` with `func != null  || (() => null);` which is now either a boolean or a function, when the original code was such that it always resulted in a function.

### Additional Info

_No response_

barroudjo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7743

## Overview

When a nullable object appears in a logical expression other than nullish coalescing (so || or &&), there is a suggested fix, and it could be wrong and introduce a bug (as explained in the https://github.com/typescript-eslint/typescript-eslint/issues/7743). This MR removes the suggested fix for these cases, and gives a more explicit message.

## Note

This is the follow-up to https://github.com/typescript-eslint/typescript-eslint/pull/7895, as I discovered that checking that the operator was not a nullish coalescing operator was not necessary in the first place because no errors are raised when using nullish coalescing.

### DIFF
--- a/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
+++ b/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
@@ -34,6 +34,7 @@ export type MessageId =
   | 'conditionErrorNullableEnum'
   | 'conditionErrorNullableNumber'
   | 'conditionErrorNullableObject'
+  | 'conditionErrorNullableObjectConvertToNullishCoalescing'
   | 'conditionErrorNullableString'
   | 'conditionErrorNullish'
   | 'conditionErrorNumber'
@@ -113,6 +114,9 @@ export default createRule<Options, MessageId>({
       conditionErrorNullableObject:
         'Unexpected nullable object value in conditional. ' +
         'An explicit null check is required.',
+      conditionErrorNullableObjectConvertToNullishCoalescing:
+        'Unexpected nullable object value in conditional with `||` or `&&`. ' +
+        'You should probably be using nullish coalescing instead.',
       conditionErrorNullableEnum:
         'Unexpected nullable enum value in conditional. ' +
         'Please handle the nullish/zero/NaN cases explicitly.',
@@ -721,6 +725,12 @@ export default createRule<Options, MessageId>({
                 },
               ],
             });
+          } else if (isLogicalExpressionNotNullishCoalescing(node.parent!)) {
+            context.report({
+              node,
+              messageId:
+                'conditionErrorNullableObjectConvertToNullishCoalescing',
+            });
           } else {
             // if (nullableObject)
             context.report({
@@ -935,6 +945,14 @@ function isLogicalNegationExpression(
   node: TSESTree.Node,
 ): node is TSESTree.UnaryExpression {
   return node.type === AST_NODE_TYPES.UnaryExpression && node.operator === '!';
+}
+
+function isLogicalExpressionNotNullishCoalescing(
+  node: TSESTree.Node,
+): node is TSESTree.LogicalExpression {
+  return (
+    node.type === AST_NODE_TYPES.LogicalExpression && node.operator !== '??'
+  );
 }
 
 function isArrayLengthExpression(
--- a/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
+++ b/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
@@ -725,7 +725,7 @@ export default createRule<Options, MessageId>({
                 },
               ],
             });
-          } else if (isLogicalExpressionNotNullishCoalescing(node.parent!)) {
+          } else if (isLogicalExpression(node.parent!)) {
             context.report({
               node,
               messageId:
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -113,6 +114,9 @@ export default createRule<Options, MessageId>({
       conditionErrorNullableObject:
         'Unexpected nullable object value in conditional. ' +
         'An explicit null check is required.',
+      conditionErrorNullableObjectConvertToNullishCoalescing:
+        'Unexpected nullable object value in conditional with `||` or `&&`. ' +
+        'You should probably be using nullish coalescing instead.',
```
auvred, line 119: If we move to this approach for objects, I see no reason to be inconsistent for the other types this rule works with: strings and numbers.

Currently, for both strings and numbers, the rule reports `conditionErrorNullableString` and `conditionErrorNullableNumber` respectively. These reports include both `value != null` and `value ?? 0` suggestions. So it's up to user to decide which suggestion should be applied in a particular situation. 

My take here is: what if we just add new suggestion with nullish coalescing to the existing `conditionErrorNullableObject` error instead of reporting new type of error without suggestions?

I see a several pros of this approach:

1) There are various cases where the logical expression can be used:

- Truth-testing context: `if (func || (() => null))`, ...
  Personally, I'd prefer to explicitly test for truth here (true/false), so if editor suggests me option of fixing to `!= null` and fixing to `?? ...`, I'd choose the former.

- Result-sensitive context: as in the original issue, where we care about the actual value of the expression result: `const foo = func || (() => null)`
  As mentioned in the original issue, fixing this to `!= null` will change the actual logic. So if editor suggests me to fix to `!= null` and fix to `?? ...`, I'd choose the latter.

  So unless rule can automatically determine which of the contexts mentioned above it operates in, I think it should provide both suggestions and leave the final decision up to the rule user.

2) If we decide to report error without providing suggestions for objects,  we should be consistent and do the same for numbers and strings (they actually provide the same `!= null` fix). I'm not sure this would be desirable behavior for all users.

What do you think?

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
```diff
@@ -1643,6 +1643,7 @@ if (x) {
         !obj
         obj || 0
         obj && 1 || 0
+        obj ?? 1
```
auvred, line 1646: [Testing] This case doesn't seem to be reported at all (which is great!), but it is currently placed in the `invalid` section of the rule tests. I think it would be better to put it in the `valid` section. What do you think?

auvred (changes requested):

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
barroudjo: I think I agree with you ! But this is a bit more work, I'll try to find the time to do it.

JoshuaKGoldberg:
👋 Hey @barroudjo! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

JoshuaKGoldberg:
Closing this PR as it's been stale for a while without activity. Feel free to reopen @barroudjo if you have time - but no worries if not! If anybody wants to drive it forward, please do post your own PR - and if you use this as a start, please add a co-author attribution. Cheers! 👋

Issue #7284, opened 2023-07-21
v6: Cannot find module '@typescript-eslint/*' or its corresponding type declarations


JoshuaKGoldberg:
## Overview

A handful of folks have mentioned that after upgrading to v6, trying to `import` from `@typescript-eslint/*` packages such as `@typescript-eslint/utils` causes a new TypeScript type error:

```ts
import { ESLintUtils } from '@typescript-eslint/utils';
```

```plaintext
Cannot find module '@typescript-eslint/utils' or its corresponding type declarations.
```

**This is not a bug.** We intentionally dropped support in v6 for Node <16 as [those versions of node are EOL](https://endoflife.date/nodejs). [v6 uses package.json exports](https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/#package-exports) - which requires Node 16+ and isn't supported by that older module resolution strategy.

**To fix the issue**, switch your TSConfig's module resolution to a newer version, most likely `"bundler"`, `"node16"`, or `"nodenext"`. See the [aka.ms/tsconfig `"moduleResolution"` docs](https://www.typescriptlang.org/tsconfig#moduleResolution) for more information.

```json
{
    "compilerOptions": {
        "moduleResolution": "bundler"
    }
}
```

See [TypeScript's new modules reference docs](https://www.typescriptlang.org/docs/handbook/modules/introduction.html) for more information.

Issues that are duplicates of this one:

* https://github.com/typescript-eslint/typescript-eslint/issues/7279
* https://github.com/typescript-eslint/typescript-eslint/issues/7213#issuecomment-1640401164
* https://github.com/typescript-eslint/typescript-eslint/issues/7185

https://arethetypeswrong.github.io/?p=%40typescript-eslint%2Futils%406.0.0 also shows a nice table of which `moduleResolution` values we support. It's all of them except `node10`.

Keeping this issue open so it shows up more in searches.

DanielDanielsson:
Sweet baby jesus, thank you so much! 🙏

Docs says it's "very likely" that you want to have these configurations for modern node projects: 

```
"compilerOptions": {
    "moduleResolution": "nodenext",
    "module": "nodenext",
    //...
  }
  ```

nbouvrette:
I spent a lot of time troubleshooting this so I am sharing this information if it can be useful for anyone else.

[...]

Issue #8206, opened 2024-01-06
Bug: Project service doesn't infer compilerOptions for files not included in a tsconfig.json


JoshuaKGoldberg:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Abug) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Issue Description

Moving https://github.com/typescript-eslint/typescript-eslint/discussions/8030#discussioncomment-8030825 into a standalone issue: users are going to want to be able to lint `.eslintrc.cjs` files without including them in the `tsconfig.json` `compilerOptions.allow` and `include`. But the project service as-is provides a "blank slate" set of default compiler options when a file isn't included in a loaded `tsconfig.json`.

Proposal: if a file is included in `allowDefaultProjectForFiles` (added in #7752) and isn't given compiler options from a loaded TSConfig, instead give it the compiler options for the nearest `tsconfig.json`?

### Reproduction Repository Link

https://github.com/JoshuaKGoldberg/repros/tree/ts-eslint-project-service-not-included

### Repro Steps

1) clone the repo
2) `npm install`
3) `npm lint`

### Versions

| package                                | version   |
| -------------------------------------- | --------- |
| `@typescript-eslint/parser`            | `6.18.0`  |
| `@typescript-eslint/typescript-estree` | `6.18.0`  |
| `TypeScript`                           | `5.3.3`   |

bradzacher:
I dunno if we should magically infer the options like that. I kinda hate black-box magic like that.

I'd be happy with an option to set the tsconfig used for extra files - so people could configure it from their eslint config. Then if they wanted to explicitly use a specific config then they could `require` it in place.

As an example the closest tsconfig might be one for Web files and include like the DOM libs or have a really restricted es version. An explicit config would let someone tweak things for their node env

mrazauskas:
Some `defaultTSConfig` option makes sense for me. It could take a path to `tsconfig.json` and use compiler options found there for inferred projects. Also the default value could be `'./tsconfig.json' | './jsconfig.json'`. This is some magic too, but with a possibility to opt out.

Issue #5871, opened 2022-10-24
Repo: Standardize how exports are used


JoshuaKGoldberg:
### Suggestion

https://github.com/typescript-eslint/typescript-eslint/blob/3ead14ebfea65058f530c0870a6c362ac042b355/packages/typescript-estree/src/create-program/createSourceFile.ts#L28

A lot of older source code files put all their `export`s at the bottom. I can see an argument in favor of it being nice to know where all the exports are for a file, but we've been slowly moving to smaller single-purpose files over time. And I personally find it a little irksome to have to edit two places to modify whether a function/etc is being exported.

Thoughts on removing all these `export { ... }`s and inline `export`ing everything?

bradzacher:
I've personally been in favour of adding exports at the bottom because it's clearer what the module signature encompasses.

In a world of "one file, ~one export" - it doesn't really matter for sure, but in the old world of "one file, many export", it creates a nice and easy-to-digest style.

I'm all for us moving towards "one file, ~one export".
While we're at it - I also tend to use named exports instead of `export default` just because it encourages you to keep things named the same - i.e. `export function foo` from file `foo.ts` instead of `export default function lolwut` from file `foo.ts`.

It also ensures your imported name isn't aliased eg `import { foo } from 'foo'` instead of `import lolwut from 'foo'`

But I think there's lint rules we can turn on to catch both of these cases and enforce naming matches.

JoshuaKGoldberg:
> adding exports at the bottom

Are there any other major repositories that do this? This is the only TypeScript codebase I can think of that does. 

In my experience it's a bit of a hassle to put things on the bottom. It's more code and an extra place to CRUD exports.

> use named exports

+1 from me! We already do in this repo, except for rule files:

https://github.com/typescript-eslint/typescript-eslint/blob/1302b30ecad9eb55aa9f1daa5068d9fb72c3688e/.eslintrc.js#L163-L164

Would you propose that we remove the override on that rule from `packages/eslint-plugin/src/rules/*.ts`? 

https://github.com/typescript-eslint/typescript-eslint/blob/1302b30ecad9eb55aa9f1daa5068d9fb72c3688e/.eslintrc.js#L265-L266

I would be in favor of that actually. It's made modifying those config files just a teeny bit more annoying. Though, it is a little nice how each rule file has the same `export`ed name...

bradzacher:
for rules it's really just a legacy thing.
[...]

PR #8460, opened 2024-02-14
fix: fix tsconfig-less check errors, fix `@types/eslint` incompatibilities, add tests

### LINKED ISSUE #8459: Docs: getting started guide causes type errors in eslint.config.js

### Suggested Changes

steps to reproduce:
```
mkdir test
cd test
npm init --yes
npm install --save-dev eslint typescript typescript-eslint
vim eslint.config.mjs
```
(paste in the code directly from the getting started guide)
```
code .
```
(open eslint.config.mjs inside vscode)
(observe the red squiggly)

Perhaps something is wrong with the published typescript definitions?

### Affected URL(s)

https://typescript-eslint.io/getting-started/

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8459, #8467

## Overview

This PR addresses a few problems:

1) Our flat config types are completely incompatible with `@types/eslint` - meaning plugins defined using those types cannot be used in `tseslint.config(..)` (#8467)
2) Our packages do not support `node10` (#8459)
3) We had no tests to validate how someone might use `typescript-eslint` IRL

---

For (3) I added an integration test which runs TS over a `eslint.config.js` file and validates the outputs so we can truely validate that `typescript-eslint` works as intended with no errors.

---

For (1) I went through and updated the types to align them or loosen them.
We've chatted about this before and I don't love loosening them but there's just severe incompatibilities between the two that means we have no choice really. We're a lot stricter than the DT types to enforce users do the right thing - but DT isn't about enforcing strictness; it's about correctly representing the state of the world.
i.e. the types are just fundamentally incompatible as they serve different use cases!

-----

For (2) ... Users right now are running into a problem:
Our getting started guide tells them to add `// @ts-check` to their `eslint.config.js` file.
When they copy+paste the example code into their file they see a type error from TS:
```
Cannot find module 'typescript-eslint' or its corresponding type declarations.ts(2307)
```

Playing around I believe the issue is that their `eslint.config.js` is not included in any `tsconfig.json`s. Without a `tsconfig.json` TS uses the "default" compiler options which sadly means the file is checked with `module: node10` 😢 (#7284).

So we have a three options really:
1) update the getting started guide to tell everyone to reconfigure their `tsconfig.json` so that they have their `eslint.config.js` included in it
2) make the package work under `node10`
3) remove `// @ts-check` and don't recommend typechecking of config files

Option (3) I hate. The entire point of `tseslint.config` was so that we could allow users to have typechecked config files. We have all the properties jsdoc'd so that users get autocomplete with all the documentation in their IDE.
So removing that from the guide is a MASSIVE step back.

Option (1) could work - however it's not going to be easy to write generic and simple docs to tell the user how to cover the `eslint.config.js` file with a tsconfig. For example if the user has just one `tsconfig.json` in their repo that they use for their build they need to separate their build config into a `tsconfig.build.json` so that they can add non-build files to `tsconfig.json`. Essentially we don't know the user's setup so it's hard to provide that succinct "silver bullet" advice on how you might do things.
Given that we want the getting started guide to be super simple - IMO this makes option (1) a non-starter.

So that means (2) is the best bet 😭. I hate it because as per #7284 we really don't want to do this hack for backwards compat.

### DIFF
--- a/packages/utils/src/ts-eslint/Parser.ts
+++ b/packages/utils/src/ts-eslint/Parser.ts
@@ -21,7 +21,7 @@ export namespace Parser {
    * This type intended to relax validation of configs so that parsers that have
    * different AST types or scope managers can still be passed to configs
    *
-   * See also LooseRuleDefinition, LooseProcessorModule
+   * @see {@link LooseRuleDefinition}, {@link LooseProcessorModule}
    */
   export type LooseParserModule =
     | {
--- a/packages/utils/src/ts-eslint/Parser.ts
+++ b/packages/utils/src/ts-eslint/Parser.ts
@@ -20,13 +20,15 @@ export namespace Parser {
    * A loose definition of the ParserModule type for use with configs
    * This type intended to relax validation of configs so that parsers that have
    * different AST types or scope managers can still be passed to configs
+   *
+   * See also LooseRuleDefinition, LooseProcessorModule
    */
   export type LooseParserModule =
     | {
         /**
          * Information about the parser to uniquely identify it when serializing.
          */
-        meta?: ParserMeta;
+        meta?: Partial<ParserMeta>;
         /**
          * Parses the given text into an ESTree AST
          */
@@ -36,7 +38,7 @@ export namespace Parser {
         /**
          * Information about the parser to uniquely identify it when serializing.
          */
-        meta?: ParserMeta;
+        meta?: Partial<ParserMeta>;
         /**
          * Parses the given text into an AST
          */
--- a/packages/utils/src/ts-eslint/Processor.ts
+++ b/packages/utils/src/ts-eslint/Processor.ts
@@ -51,7 +51,7 @@ export namespace Processor {
    * This type intended to relax validation of configs so that parsers that have
    * different AST types or scope managers can still be passed to configs
    *
-   * See also LooseRuleDefinition, LooseParserModule
+   * @see {@link LooseRuleDefinition}, {@link LooseParserModule}
    */
   export interface LooseProcessorModule {
     /**
--- a/packages/utils/src/ts-eslint/Processor.ts
+++ b/packages/utils/src/ts-eslint/Processor.ts
@@ -45,4 +45,43 @@ export namespace Processor {
      */
     supportsAutofix?: boolean;
   }
+
+  /**
+   * A loose definition of the ParserModule type for use with configs
+   * This type intended to relax validation of configs so that parsers that have
+   * different AST types or scope managers can still be passed to configs
+   *
+   * See also LooseRuleDefinition, LooseParserModule
+   */
+  export interface LooseProcessorModule {
+    /**
+     * Information about the processor to uniquely identify it when serializing.
+     */
+    meta?: Partial<ProcessorMeta>;
+
+    /**
+     * The function to extract code blocks.
+     */
+    /*
+    eslint-disable-next-line @typescript-eslint/no-explicit-any --
+    intentionally using `any` to allow bi-directional assignment (unknown and
+    never only allow unidirectional)
+    */
+    preprocess?: (text: string, filename: string) => any;
+
+    /**
+     * The function to merge messages.
+     */
+    /*
+    eslint-disable-next-line @typescript-eslint/no-explicit-any --
+    intentionally using `any` to allow bi-directional assignment (unknown and
+    never only allow unidirectional)
+    */
+    postprocess?: (messagesList: any, filename: string) => any;
+
+    /**
+     * If `true` then it means the processor supports autofix.
+     */
+    supportsAutofix?: boolean;
+  }
 }
--- a/packages/utils/src/ts-eslint/Rule.ts
+++ b/packages/utils/src/ts-eslint/Rule.ts
@@ -635,7 +635,7 @@ export interface RuleModule<
 export type AnyRuleModule = RuleModule<string, readonly unknown[]>;
 
 /**
- * A loose definition of the RuleModule type for use with configs. This type
+ * A loose definition of the RuleModule type for use with configs. This type is
  * intended to relax validation of types so that we can have basic validation
  * without being overly strict about nitty gritty details matching.
  *
@@ -644,7 +644,7 @@ export type AnyRuleModule = RuleModule<string, readonly unknown[]>;
  * super strict validation in a config - a loose shape match is "good enough" to
  * help validate the config is correct.
  *
- * See also LooseParserModule, LooseProcessorModule
+ * @see {@link LooseParserModule}, {@link LooseProcessorModule}
  */
 export type LooseRuleDefinition =
   // TODO - ESLint v9 will remove support for RuleCreateFunction
--- a/packages/utils/src/ts-eslint/Rule.ts
+++ b/packages/utils/src/ts-eslint/Rule.ts
@@ -8,7 +8,7 @@ import type { SourceCode } from './SourceCode';
 
 export type RuleRecommendation = 'recommended' | 'strict' | 'stylistic';
 
-interface RuleMetaDataDocs {
+export interface RuleMetaDataDocs {
   /**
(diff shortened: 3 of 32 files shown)


Zamiell:
Also note that in the ultra long term, TypeScript itself would update the default value from Node10 to the newer setting. In that world, this PR could be reverted. So maybe in 5 years or so?

bradzacher:
@Zamiell - https://github.com/microsoft/TypeScript/issues/54500 - TS 6.0 maybe?
we're on 5.3 - +0.1 per quarter so 7 quarters -- 2 years!?!?

htho:
Hi, I opened #8459 which is primarily a documentation issue, but it also shows, that type information for `tseslint.config()` is (partially) missing.

I manually applied ["main": "./dist/index.js", to my ./node_modules/typescript-eslint/package.json](https://github.com/typescript-eslint/typescript-eslint/pull/8460/commits/da23118848d17c7f33d4bbe9024f43806d4d1369)

The Problem is @typescript-eslint/utils/ts-eslint can not be resolved, and therefore FlatConfig can not be imported.

Maybe `"main": "./dist/index.js"` has to be applied to other `package.json` files too?

bradzacher:
Hmm. My little toy tester didn't actually paste the example in - I just had the import and fixed the error.

Will defs need to think on this more 🤔

bradzacher:
Cc @jakebailey do you know of a way we can help users solve this easily in a consistent, generic way? Or are we just stuck because of the `node10` default?

jakebailey:
Just to be super clear, are you worried about the editor case?

In the next version of TypeScript, there's a new mode that vscode will use as its default for the implicit project; module=preserve  + moduleResolution=bundler, which will likely make this all go away as that combo will be export map aware.

bradzacher:
@jakebailey yeah IMO the editor usecase is the primary concern. I think a lot of people won't want to strictly typecheck their config files via the CLI - if they do then this problem goes away!

Using `// @ts-check` means that they can have type errors and autocomplete in their IDE whilst authoring their config which should be "good enough" given eslint also checks the config at lint-time.

> default for the implicit project; module=preserve + moduleResolution=bundler, which will likely make this all go away as that combo will be export map aware

That's great to hear!
But that default will require people to upgrade to TS5.4, right? Or will it apply to vscode's setup by default?

JoshuaKGoldberg (approved):
👍 makes sense.

Once the experimental project service is stabilized and #8206 resolved, this should no longer be an issue. We then should be able to remove it for v8.

bradzacher:
> Once the experimental project service is stabilized and https://github.com/typescript-eslint/typescript-eslint/issues/8206 resolved, this should no longer be an issue

@JoshuaKGoldberg it won't fix anything! This isn't a problem with ESLint reporting on the file - it's a problem with TypeScript itself reporting on the file!

jakebailey:
> But that default will require people to upgrade to TS5.4, right? Or will it apply to vscode's setup by default?

Yes, people would need to update. Though, a lot of people use the built-in version of TS, which helps.

I'll note that the fact that vscode uses different defaults for the inferred (implicit? one of those) poses the question of what defaults you all should set too; I suspect that you may want to mirror them.

bradzacher:
The big two things we strongly disagree with is
- the default of `node10` module resolution -- we push everyone to `node16`
- the default of same strict settings like `strictNullChecks` -- we push everyone to turn them on.

But the former is the only requirement we've imposed (much to some people's chagrin #7284).

We honestly hate the fact that node10 resolution is the default and will likely remain the default till TS 6.0 because it's a land mine for a lot of usecases.

jakebailey:
In 5.4+, VS Code will default to module=preserve/moduleResolution=bundler, and it has defaulted to strictNullChecks and strictFunctionTypes for a year or two. I do feel like you'd stand to benefit from also setting those defaults if available; it'll only affect random files not in projects, and these are the settings I'd hazard to say that most people see anyway if they never configured TypeScript.

But, yes, I don't think these defaults are going to change until 6.0 unless we decide to change what we announced about setting stability previously. I'm not totally sure what a better default would be, though; the combo I described above is a perfect case for VS Code as it never emits, and arguably eslint, but it's not a great default for regular tsconfigs as its emit is probably not super helpful.

bradzacher:
@JoshuaKGoldberg this PR has been radically updated from the 1 line PR you stamped.
Please re-review 😄

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -632,33 +632,48 @@ interface RuleModule<
    */
   create(context: Readonly<RuleContext<TMessageIds, TOptions>>): TRuleListener;
 }
-type AnyRuleModule = RuleModule<string, readonly unknown[]>;
+export type AnyRuleModule = RuleModule<string, readonly unknown[]>;
 
-type RuleCreateFunction<
+/**
+ * A loose definition of the RuleModule type for use with configs. This type
```
JoshuaKGoldberg, line 638:
```suggestion
 * A loose definition of the RuleModule type for use with configs. This type is
```

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -632,33 +632,48 @@ interface RuleModule<
...
-  RuleFix,
-  RuleFixer,
-  RuleFunction,
-  RuleListener,
-  RuleListenerExtension,
-  RuleMetaData,
-  RuleMetaDataDocs,
-  RuleModule,
-  SharedConfigurationSettings,
-};
```
JoshuaKGoldberg, line 664: [Non-Actionable] That reminds me, I totally forgot about https://github.com/typescript-eslint/typescript-eslint/issues/5871 ...

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -632,33 +632,48 @@ interface RuleModule<
...
+ * A loose definition of the RuleModule type for use with configs. This type
+ * intended to relax validation of types so that we can have basic validation
+ * without being overly strict about nitty gritty details matching.
+ *
+ * For example the plugin might be declared using an old version of our types or
+ * they might use the DefinitelyTyped eslint types. Ultimately we don't need
+ * super strict validation in a config - a loose shape match is "good enough" to
+ * help validate the config is correct.
+ *
+ * See also LooseParserModule, LooseProcessorModule
```
JoshuaKGoldberg, line 647: [Docs] Nit: for nice JSDoc linkage

```suggestion
 * @see {@link LooseParserModule}, {@link LooseProcessorModule}
```

packages/utils/src/ts-eslint/Parser.ts
```diff
@@ -20,13 +20,15 @@ export namespace Parser {
    * A loose definition of the ParserModule type for use with configs
    * This type intended to relax validation of configs so that parsers that have
    * different AST types or scope managers can still be passed to configs
+   *
+   * See also LooseRuleDefinition, LooseProcessorModule
```
JoshuaKGoldberg, line 24:
```suggestion
   * @see {@link LooseRuleDefinition}, {@link LooseProcessorModule}
```

packages/utils/src/ts-eslint/Processor.ts
```diff
@@ -45,4 +45,43 @@ export namespace Processor {
      */
     supportsAutofix?: boolean;
   }
+
+  /**
+   * A loose definition of the ParserModule type for use with configs
+   * This type intended to relax validation of configs so that parsers that have
+   * different AST types or scope managers can still be passed to configs
+   *
+   * See also LooseRuleDefinition, LooseParserModule
```
JoshuaKGoldberg, line 54:
```suggestion
   * @see {@link LooseRuleDefinition}, {@link LooseParserModule}
```

JoshuaKGoldberg (dismissed):
Yeah 😬 I don't see a way around this. Good work!

JoshuaKGoldberg (approved):

Issue #8433, opened 2024-02-11
Enhancement: [no-floating-promises] Don't check coincidentally Promise-like (Thenable) type values by default


JoshuaKGoldberg:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### My proposal is suitable for this project

- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Link to the rule's documentation

https://typescript-eslint.io/rules/no-floating-promises

### Description

Forwarding out from conversations with @mcollina, especially https://github.com/nodejs/node/issues/51292#issuecomment-1937792288: right now `no-floating-promises` will consider anything that _looks_ like a Promise (i.e. is "Promise-like") to be a potential floating value:

https://github.com/typescript-eslint/typescript-eslint/blob/5857356962060b19aa792bee778f9167ee54154b/packages/eslint-plugin/src/rules/no-floating-promises.ts#L353

But to my knowledge this hasn't been particularly helpful. If something _coincidentally happens_ to match the [`Thenable` interface](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables) that doesn't necessarily mean it shouldn't float!

The historical context here is that it used to be common in JavaScript code to use alternative implementations of Promises, such as [Bluebird](https://github.com/petkaantonov/bluebird). Promises were only added to some browsers in ~2014 and Node & co in ~2015 or so. Many sites still had to support old versions of Internet Explorer. The `no-floating-promises` typescript-eslint rule was implemented in 2019 when it was still not-unheard-of to use alternative Promises or at least have Promise types provided by a polyfill: https://github.com/typescript-eslint/typescript-eslint/pull/495.

Now that the vast majority of projects are using built-in Promise values & types, I think the value of detecting those values automatically is no longer worth the pain of flagging general thenables. I propose we modify the rule to:

* _(non-breaking)_ Add in an option to specify types to be considered Promises - using the same standard format as #7008
[...]

Issue #8404, opened 2024-02-07
Enhancement: [no-floating-promises] add an 'allowForKnownSafeCalls' option


JoshuaKGoldberg:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### My proposal is suitable for this project

- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Link to the rule's documentation

https://typescript-eslint.io/rules/no-floating-promises

### Description

#7008's `allowForKnownSafePromises` is a _type_-oriented approach to allowlisting safe _types_ to be floating. For frameworks that have those Promises show up in many places (e.g. https://github.com/reduxjs/redux-toolkit/issues/4101), marking the _type_ as safe is a reasonable solution.

But (paraphrasing private remarks from @Josh-Cena): that requires framework authors to set things up in a very specific way that feels like jumping through hoops specific to another, tangential technology (us). For example, https://github.com/nodejs/node/issues/51292 tracks how `node:test`'s `it` & similar return a `Promise`. Even if the Node types get some fancy branded `SafePromise` type added (one inconvenience), users really just want a way to mark that `it()` calls are safe. Asking users to know the Node.js-specific _types_ is more learning curve for them (another inconvenience).

In other words, we think users would really want to mark specific _functions_ as safe to call. A kind of `allowForKnownSafePromiseReturns` option.

```js
"@typescript-eslint/no-floating-promises": ["error", {
  "allowForKnownSafePromiseReturns": [
    { "from": "package", "name": "it", "package": "node:test" }
  ]
}]
```

### Fail

```TypeScript
import { it } from "node:test";

it("...", () => { /* ... */ });
```

### Pass

```TypeScript
/* "allowForKnownSafePromiseReturns": [{ "from": "package", "name": "it", "package": "node:test" }] */

import { it } from "node:test";

it("...", () => { /* ... */ });
```

bradzacher:
```ts
// should be ignored
it('', () => {
  // can sometimes be ignored
[...]

PR #8502, opened 2024-02-18
feat(eslint-plugin): [no-floating-promises] add an 'allowForKnownSafePromises' option

### LINKED ISSUE #7008: Enhancement: [no-floating-promises] add an 'allowForKnownSafePromises' option

### Description

For years, various community members have asked to add a type allowlist/blocklist or other _"stop complaining about these specific areas"_ for `no-floating-promises`:

* #2640
* #3192 _(ish)_
* #4722
* #4791
* #5231
* #5844
* #6017
* #7006

<sub>_(including clearly duplicate issues to show how this is a relatively more requested option, and to surface the different use cases requested for)_</sub>

Out of those, #5844's `node:test` is most convincing - as `node:test` is stable as of Node 20.0.0 (https://github.com/nodejs/node/issues/46642 -> https://github.com/nodejs/node/pull/46983), and returns `Promise<void>` for tests: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8730660c37a33dc0768b3f1e36b679294acc4988/types/node/test.d.ts#L134.

I propose we adopt the format discussed in #6017 & implemented in #4436 ([see v6 docs](https://v6--typescript-eslint.netlify.app/rules/prefer-readonly-parameter-types/#allow)) as a new `allow` option for `no-floating-promises`. We'd want to add some very explicitly, strongly worded descriptions in the docs as to why it's not an option we recommend using lightly.

### Fail

```typescript
async function test(name: string, executor: () => Promise<void>) {
  // ...
}

test("example", () => {});
```

### Pass

```typescript
import test from 'node:test';

test("example", () => {});
```

### Additional Info

cc @armano2 @bradzacher as past people who, like me, have previously disagreed with the suggestion of adding this option.

Edit: to be clear, this is roughly what users would be able to specify in their ESLint config _(pending naming bikeshedding)_:

```js
"@typescript-eslint/no-floating-promises": ["error", {
  "allowForKnownSafePromises": [
    { "from": "package", "name": "SafePromise", "package": "some-fancy-package"
  ]
}]
```

arka1002:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7008

## Overview

~~Hi, I'm really curious, is there any way to test code which have `import` in them ? https://github.com/typescript-eslint/typescript-eslint/pull/4436 also didn't have them.~~
Nvm, found it.

Let's discuss.


packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -160,12 +160,19 @@ function typeDeclaredInPackage(
...
+  let matcher = new RegExp(`${packageName}|${typesPackageName}`);
+
+  if (packageName.startsWith('node:')) {
+    matcher = new RegExp(packageName.substring(5));
+  }
   return declarationFiles.some(declaration => {
-    const packageIdName = program.sourceFileToPackageName.get(declaration.path);
+    const packageIdName =
+      program.resolvedModules.has(declaration.path) ||
+      program.sourceFileToPackageName.has(declaration.path);
```
arka1002, line 171: I just want to add a small note:

In case of `@types/node`(the ts definitions for node), `sourceFileToPackageName` doesn't list the `.d.ts` files.

Example:
```ts
import url from "node:url";
```
`sourceFileToPackageName` won't list the file - `url.d.ts`

packages/type-utils/typings/typescript.d.ts
```diff
@@ -36,6 +36,8 @@ declare module 'typescript' {
      * Maps from a SourceFile's `.path` to the name of the package it was imported with.
      */
     readonly sourceFileToPackageName: ReadonlyMap<Path, string>;
+
+    readonly resolvedModules: ReadonlyMap<Path, unknown>;
```
arka1002, line 40: And this..... the typescript compiler source code has a very complex type for the value of the Map.

Thoughts about how to improve this ?

packages/eslint-plugin/docs/rules/no-floating-promises.md
```diff
@@ -109,6 +109,32 @@ await (async function () {
 })();
 ```
 
+### `allowForKnownSafePromises`
+
+This allows you to skip checking of the promise returning functions, which are supposed to be unhandled and unreturned, documented by an external package/module.
```
JoshuaKGoldberg, line 114: I think this conflates #7008 and #8404?

JoshuaKGoldberg (changes requested):
Haven't been able to look deeply yet, but - looks like there's a mismatch in naming?

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -160,12 +160,19 @@ function typeDeclaredInPackage(
   // Handle scoped packages - if the name starts with @, remove it and replace / with __
   const typesPackageName = packageName.replace(/^@([^/]+)\//, '$1__');
 
-  const matcher = new RegExp(`${packageName}|${typesPackageName}`);
+  let matcher = new RegExp(`${packageName}|${typesPackageName}`);
+
+  if (packageName.startsWith('node:')) {
+    matcher = new RegExp(packageName.substring(5));
+  }
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Yeah, I don't think we should hardcode `node:` or any other protocols.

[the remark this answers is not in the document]
arka1002: resolved here - https://github.com/typescript-eslint/typescript-eslint/issues/8572#issuecomment-1969886964  
> `node:foo` maps to `@types/node`

We should only allow those specifiers which has a corresponding `.d.ts` file in the node_modules dir. The latter part of the line is the implementation description, which I'll do in this pr.

JoshuaKGoldberg:
👋 @arka1002 adding the new `allow*` options to `no-floating-promises` is one of our priorities for the next few months. I'd like to make sure this is actively moving towards completion. Do you think you'll be able to make a lot of progress over the next week?

If not, no worries - I'd happily take this on myself. It's a nuanced rule with a lot of context around it. ❤️

arka1002:
Yes, I've almost done the tests, give me just 3 more days. I'll open it for review.

arka1002:
I think I might need some help with the lint ci failure, I have no idea why that line is reporting a prefer-nullish-coalescing error :smiling_face_with_tear:

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -115,6 +115,66 @@ await (async function () {
 })();
 ```
 
+### `allowForKnownSafePromises`
+
+This allows you to skip checking of those promises which are marked with some kind of branded type. The branding of a type can be done either by using a type alias or an interface.
```
JoshuaKGoldberg, line 120: > branded type

[Docs] No, branded types are just one coincidental example of something this option could be used with. A couple more examples:

* `Promise<number> & { hey?: string }`, as in the tests
* `class SafePromise extends Promise`
* `interface SecureThenableLike<T> { then(): T; }`

packages/eslint-plugin/src/rules/no-floating-promises.ts
JoshuaKGoldberg, line 1: [Bug] This PR right now adds the `allowForKnownSafePromises` check for a couple places, but not all of them. For example, the `isPromiseArray` check > `return { isUnhandled: true, promiseArray: true }` doesn't have the new option factored in. 

I tried out a test case like this that should be `valid` but still has `errors`:

```js
  valid: [
    {
      only: true,
      code: `
        declare const arrayOrPromiseTuple: Foo<unknown>[];
        arrayOrPromiseTuple;
        type Foo<T> = Promise<T> & { hey?: string };
      `,
      // errors: [{ line: 3, messageId: 'floatingPromiseArrayVoid' }],
      options: [{ allowForKnownSafePromises: [{ from: 'file', name: 'Foo' }] }],
    },
```

You'll want to do a search on `isUnhandled: true` to see all the places that will need to factor in the logic.

JoshuaKGoldberg, line 1: [Refactor] It looks like the approach taken is to add the `doesTypeMatchesSpecifier` check to all possible cases within `isUnhandledPromise`. That's a lot of potential cases! And it'd be easy to miss one. Or several, as is the case right now.

As an alternate approach, have you considered changing `isUnhandledPromise` to return the type(s) of floating Promise(-like)s/Thenables rather than a `boolean`? Then the one place that starts the root `isUnhandledPromise` call can run the `doesTypeMatchesSpecifier` logic on that result.

I haven't verified this locally, so no guarantees it'll work. 😄

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -323,6 +371,31 @@ export default createRule<Options, MessageId>({
...
+ * @param options The config object of `allowForKnownSafePromises`
+ * @returns `true` if the type matches, `false` if it isn't
+ */
+function doesTypeMatchesSpecifier(
+  services: ParserServicesWithTypeInformation,
+  node: TSESTree.Node,
+  options: TypeOrValueSpecifier[] | undefined,
+): boolean {
+  if (Array.isArray(options) && options.length > 0) {
+    const result = options.some(specifier =>
```
JoshuaKGoldberg, line 387: [Style] Small nit, you can make this function several lines smaller:

```ts
  return (
    !!options?.length &&
    options.some(specifier =>
```

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -323,6 +371,31 @@ export default createRule<Options, MessageId>({
...
 });
 
+/**
+ * It checks whether a node's type matches one of the types listed in the `allowForKnownSafePromises` config
+ * @param services services variable passed from context function to check the type of a node
+ * @param node the node whose type is to be calculated to know whether it is a safe promise or not
+ * @param options The config object of `allowForKnownSafePromises`
+ * @returns `true` if the type matches, `false` if it isn't
+ */
+function doesTypeMatchesSpecifier(
```
JoshuaKGoldberg, line 381: [Docs] A bit of naming grammar:

```suggestion
function doesTypeMatcheSpecifier(
```

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
JoshuaKGoldberg, line 1: [Testing] A lot of these cases have a lot of cases within a single code block. For `valid` ones, that's multiple calls to potentially floating methods; for `invalid` ones, that's multiple elements in their `errors`. Per https://typescript-eslint.io/contributing/local-development#granular-unit-tests we ask that unit tests be self-contained and granular. That makes them easier to read and debug. Could you please split them up?

packages/type-utils/src/TypeOrValueSpecifier.ts
JoshuaKGoldberg: Bump @arka1002 - I don't believe this logic should be changed in this PR. If there's an existing bug or usage blocker _specific to `no-floating-promises`_ then that'd make sense. But if it's some pre-existing issue, like `@types/node` not listing `.d.ts` files, then that's something separate. Especially if the proposed solution involves using a TypeScript internal like `resolvedModules` for the first time.

JoshuaKGoldberg (changes requested):
You're somewhat on the right track 🙂 but there are a few pieces here that need some refactoring & changing. My advice would be to really deeply read through the existing source code, tests, linked issue, and discussions - then consider all of that in the design you go with. Feel free to ask for help, or just throw this back at us if it's a bit more than you meant to take on. ❤️

JoshuaKGoldberg:
As an aside: we've been talking internally about getting more granular about labeling difficulty levels on issues. In retrospect, I wish we'd added a `non-trivial` label or some similar thing on this one to indicate it's not a `good first issue` or even `good second issue`. Sorry if it looked reasonable at first and is now showing itself to be a big thing.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -115,6 +115,66 @@ await (async function () {
...
+
+Each item must be one of :
+
+- A type defined in a file (`{from: "file", name: "Foo", path: "src/foo-file.ts"}` with `path` being an optional path relative to the project root directory)
+- A type from the default library (`{from: "lib", name: "PromiseLike"}`)
+- A type from a package (`{from: "package", name: "Foo", package: "foo-lib"}`, this also works for types defined in a typings package).
+
+Additionally, a type may be defined just as a simple string, which then matches the type independently of its origin.
```
JoshuaKGoldberg, line 128: [Docs] Let's, ah, not mention this. We really don't want people to use it. See linked issues' discussions around naming: catch-all names tend to be overly permissive.

Eventually/separately we'll set up unified docs for the format and mention the preference against strings.

packages/type-utils/typings/typescript.d.ts
JoshuaKGoldberg: https://github.com/typescript-eslint/typescript-eslint/pull/8502#discussion_r1541735957

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
arka1002: I don't think I understood this.
The linked issue says this right ?
> I propose we adopt the format discussed in https://github.com/typescript-eslint/typescript-eslint/discussions/6017 & implemented in https://github.com/typescript-eslint/typescript-eslint/pull/4436 ([see v6 docs](https://v6--typescript-eslint.netlify.app/rules/prefer-readonly-parameter-types/#allow)) as a new `allow` option for `no-floating-promises`.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
arka1002: Ah! I think I need some help on this.

What do you expect 1 test to look like ?
&, if we split the tests according to the type of ExpressionStatements, it'll result in a large number of context-less tests.

Isn't it more reasonable to keep the ExpressionStatements together ?

I'm re-requesting you for a review.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -486,11 +486,6 @@ declare const promiseArray: Array<Promise<unknown>>;
 void promiseArray;
       `,
     },
-    {
-      code: `
-[Promise.reject(), Promise.reject()].then(() => {});
-      `,
-    },
```
arka1002, line 493: Invalid js/ts code.
> Property 'then' does not exist on type 'Promise\<never\>[]'.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
JoshuaKGoldberg: Yup, and none of us on the team are enthused that the `string` type exists. At the time we'd added it for compatibility with the existing rule. Not because we particularly wanted it to exist.

Especially for this new rule, I don't think we want users to really be encouraged to use it.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
JoshuaKGoldberg: Instead of, for example, this:

```ts
[
  {
    code: `
type SafePromise = Promise<number> & { hey?: string };
declare const value: SafePromise;
value;
value.then(() => {});
    `,
    options: [{ allowForKnownSafePromises: [{ ... name: 'SafePromise' }] }],
  },
]
```

I'm suggesting this:

```ts
[
  {
    code: `
type SafePromise = Promise<number> & { hey?: string };
declare const value: SafePromise;
value;
    `,
    options: [{ allowForKnownSafePromises: [{ ... name: 'SafePromise' }] }],
  },
  {
    code: `
type SafePromise = Promise<number> & { hey?: string };
declare const value: SafePromise;
value.then(() => {});
    `,
    options: [{ allowForKnownSafePromises: [{ ... name: 'SafePromise' }] }],
  },
]
```

It's more code & annoying to write, but:
* Much clearer to work with when something fails, because it's clear exactly which part(s) are failing
* More precise - we end up with fewer _"oops we left this line in because we didn't notice it was unnecessary"_

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
JoshuaKGoldberg: This comment thread isn't resolved 🙂 maybe I wasn't being clear. I'm saying these docs shouldn't say that the `allowForKnownSafePromises` type is for branded types.

Vague phrasing starter:

> This option allows marking specific types as "safe" to be floating.
> For example, you may need to do this in the case of libraries whose APIs return Promises whose rejections are safely handled by the library.

arka1002: ye fixed, forgot to add the docs, in the prev commits they were added as tests, ptal.

JoshuaKGoldberg (changes requested):
Sorry, I'm a little rushed, haven't had time to do a deep review - but leaving some quick suggestions for now. Apologies if you're still working on this and that's why you haven't re-requested. 🙂

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
JoshuaKGoldberg, line 1: [Testing] Whoa, this is a _lot_ of added tests! All for just a few added pieces of logic in the rule. It looks like these tests are testing two things:

* That the rule factors in the `allowForKnownSafePromises` option
* That the existing `TypeOrValueSpecifier` logic handles all sorts of various types

The former is a good thing to test, +1 to that. But the latter doesn't generally need to be tested here. Only things specific to the rule should be tested.

In other words: each test should be targeted to new logic, not existing logic. It looks like there are a lot of tests that target much of the same logic. That makes them duplicate tests. Can you trim down the new tests?

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -325,20 +366,53 @@ export default createRule<Options, MessageId>({
...
   for (const ty of tsutils
     .unionTypeParts(type)
     .map(t => checker.getApparentType(t))) {
     if (checker.isArrayType(ty)) {
       const arrayType = checker.getTypeArguments(ty)[0];
+      if (Array.isArray(options) && options.length > 0) {
+        return !tsutils
+          .unionTypeParts(arrayType)
+          .some(type => doesTypeMatchSpecifier(services, options, type));
+      }
```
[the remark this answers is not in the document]
arka1002: > _otherwise, check if it's promise-like_

Ty [`aa94a28`](https://github.com/typescript-eslint/typescript-eslint/commit/aa94a28de74c225c8e7338c73f56330eb1015227)

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -325,20 +385,76 @@ export default createRule<Options, MessageId>({
...
+    typeOfNode = type;
+  }
+
+  if (Array.isArray(options) && options.length > 0) {
+    const result = options.some(specifier =>
+      typeMatchesSpecifier(typeOfNode, specifier, services.program),
+    );
+    return result;
+  }
+  return false;
```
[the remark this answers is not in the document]
arka1002: > by making the type system know

Ye, sorry :sweat:

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -115,6 +115,54 @@ await (async function () {
 })();
 ```
 
+### `allowForKnownSafePromises`
+
+This option allows marking specific types as "safe" to be floating. For example, you may need to do this in the case of libraries whose APIs return Promises whose rejections are safely handled by the library.
+
+Each item must be one of :
```
JoshuaKGoldberg, line 122: [Docs] Nit: I keep forgetting to mention this 😂 

```suggestion
Each item must be one of:
```

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
JoshuaKGoldberg, line 1: [Bug] [tagged template expression returning an allowed type](https://deploy-preview-8502--typescript-eslint.netlify.app/play/#ts=5.4.3&fileType=.ts&code=C4TwDgpgBAyghgMwgBQE4HsC2BLAztAXijSzwgB4A7AV0wCMJUA%2BKAMigG8oB9bgG2yVgjAEKo4lACa4A-AC4ouYKkEBzKAF8A3ACgdkiAGM%2BcVNEPpKSqJhAAVOKoUAKJSsqrcCuxExgTwjDKargAgqjiIACUUAQs8EgkOPi6tg6qAAZwdIYZOkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDN4OBDfMwDmtYtA4BbSshTowAbXDYciaBOiQANMpVYV2SP3h8A7gDEO0ANKdTTAMr9uiAAoTps1BiUGD%2BvxVIbg95YMokbV1AnCZ%2BSUQwpxd3KRlE6L8AXx0YiAC84NCMSEYAIyi8wziEsNTPRAAZSgBrDKqczP8uvyKpMOJ%2BMhb%2BYUTcqqgaxJKAIX5NCarIQeHRmagyhdpyyB7sLP2AXS7DlROIQ6ygA&tsconfig=&tokens=false)

JoshuaKGoldberg (changes requested):
🆒 thanks for trimming down the test cases! I think this is looking pretty great. Just one last bug from my end and I think I'll be done requesting changes (hopefully, probably).

I'll also leave this open to get at least one other review from @typescript-eslint/triage-team. I've looked at it enough that I don't trust myself to not be missing something. 😄

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
arka1002: [`ecd31e3`](https://github.com/typescript-eslint/typescript-eslint/pull/8502/commits/ecd31e3439fb3aad312f7216f6fb5bd44df65367) 👍 

& thanks for all the reviews :smiley:

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -115,6 +115,54 @@ await (async function () {
 })();
 ```
 
+### `allowForKnownSafePromises`
+
+This option allows marking specific types as "safe" to be floating. For example, you may need to do this in the case of libraries whose APIs return Promises whose rejections are safely handled by the library.
+
+Each item must be one of:
```
kirkwaiblinger, line 122: For a first-time reader, this needs a bit of clarification... Each item of what? 
Consider something like the following.

_This option takes an array of type specifier objects. There are three possible types of specifiers: "file", "lib", and "package"._

followed by the list of what each does

kirkwaiblinger (changes requested):

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -504,6 +499,209 @@ void promiseArray;
 ['I', 'am', 'just', 'an', 'array'];
       `,
     },
+    // Branded type annotations on variables containing promises
```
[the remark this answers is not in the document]
arka1002: Correct ... these are just left from the previous commits ...

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -89,12 +99,16 @@ export default createRule<Options, MessageId>({
...
+      allowForKnownSafePromises: readonlynessOptionsDefaults.allow,
     },
   ],
 
   create(context, [options]) {
     const services = getParserServices(context);
     const checker = services.program.getTypeChecker();
+    // TODO: #5439
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const allowForKnownSafePromises = options.allowForKnownSafePromises!;
```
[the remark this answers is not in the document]
JoshuaKGoldberg: I'm a little -1 on this change tbh, it's common for options to have the same name as their variables. Not a strong opinion from me though

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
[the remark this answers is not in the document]
arka1002: > there don't seem to be any tests that have `"from": "lib"`

it's present , https://github.com/typescript-eslint/typescript-eslint/pull/8502/files#diff-520e9e574b74ad1c0a275b21b5ad4f70f5cc8b7c8717a4884c4e3c4e8162ffe8R670-R676

> `"from": "package"`

Yes, it is absent & adding this test is a challenge too. Is there any package in typescript-eslint dependency list which exports a "branded type"(or something) of a Promise ?? If anyone knows please tell :sweat_smile: , only then it can be added to a test.

I think for exactly this reason `"from": "package"` tests weren't added in https://github.com/typescript-eslint/typescript-eslint/pull/4436 too

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
arka1002: I just copied what was written here ... https://typescript-eslint.io/rules/prefer-readonly-parameter-types#allow 
Should I change both or only this one ?

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -325,20 +373,46 @@ export default createRule<Options, MessageId>({
   },
 });
 
-function isPromiseArray(checker: ts.TypeChecker, node: ts.Node): boolean {
+function doesTypeMatchSpecifier(
```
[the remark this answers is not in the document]
arka1002: I dont really understand the point being made here though.

Are you saying the rule should report in this case ? [Playground link line 4](https://deploy-preview-8502--typescript-eslint.netlify.app/play#ts=5.4.3&fileType=.tsx&code=MYGwhgzhAEDKYDMCmAFATgewLYEsJIB4AVAPmiQA8AXJAOwBMZ1s9DToBvaAXwFgAoYBloQq0CIlSZc%2BAFxxJzGYVoBXLACMkaMgF5oS1gDo0SCBhAA3JAAoArAEoA3AID0AKnIQQOWmIBEAAJUAJ4ADmbAaDhhVAC0Zj5%2BrrQYcQggGGBUvgDmcWHSrBD%2B8gDa-tqYaP4ANJzQYCCZAO4AYhhoANKpLbTwyIb4EOVcCNLyAOQIOCBIk-W0YFhIUwNSLPiTPAC6PNB77q4CEoNF%2BEbA2cAAFjbOAkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

& the reason is - `catch()` returns a normal `Promise` ??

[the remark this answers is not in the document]
kirkwaiblinger: Hmmm. I think I see your point.... but let's drill down into the details a bit, here. There's a difference between `.catch()` and `.catch(() => {})`. 

for normal promises,
```ts
// doesn't flag, promise is handled
normalPromise.catch(() => {});
// does flag! catch requires a (function) argument to handle the promise
normalPromise.catch();
```

so, my thought would be
```ts
// shouldn't flag; since this would be considered a safe way to handle any thenable
// (hopefully won't be encountered in real world code, anyway, since the whole point
// is that you shouldn't need to .catch a safe promise)
safePromise.catch(() => {});
// likewise, shouldn't flag
safePromise.then(() => {}, () => 'has rejection handler');

// should flag - this wouldn't be a valid way of handling a rejection in a normal promise,
// and we don't know what the safePromise .catch does other than that it creates a normal promise 
// that's unhandled 🤷 
safePromise.catch()
```
But, I don't feel certain that my interpretation here is unambiguously correct, so feel free to make the case that it should go the other way 🙂 

Regardless of the above, I think that this test case unambiguously needs to flag (provided safePromise produces a normal Promise, not another SafePromise). 
```ts
safePromise.then(() => { throw new Error(); });
```

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
kirkwaiblinger: Oh, thanks for pointing that out! I would say, in that case, leave it as is 🙂 Better if they both get looked at in a separate issue, then, rather than being reexamined in this PR

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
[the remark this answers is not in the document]
kirkwaiblinger: > it's present , https://github.com/typescript-eslint/typescript-eslint/pull/8502/files#diff-520e9e574b74ad1c0a275b21b5ad4f70f5cc8b7c8717a4884c4e3c4e8162ffe8R670-R676

✅ 

> > `"from": "package"`
> 
> Yes, it is absent & adding this test is a challenge too. Is there any package in typescript-eslint dependency list which exports a "branded type"(or something) of a Promise ?? If anyone knows please tell 😅 , only then it can be added to a test.

Hmm - yeah, fair enough, makes sense to leave it as is based on the precedent 👍

packages/eslint-plugin/src/rules/no-floating-promises.ts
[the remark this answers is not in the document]
arka1002: > other than that it creates a normal promise

It isn't ? Lets take this example
```ts
class SafePromise<T> extends Promise<T> { }
const safePromise: SafePromise<number> = Promise.resolve(5);
safePromise.catch(); // the type of safePromise is `SafePromise<number>` hover on the variable name in vscode
```

> we don't know what the safePromise .catch does

Yeah, this is kindof what this PR is trying to do.

The `allowForKnownSafePromises` option is letting the rule know which types are safe to-not-handle-rejection.
To-not-handle-rejection means `.catch` can have 0 arguments, `.then` can have 1 arg etc ...

Lets take an example
```ts
class SafePromise<T> extends Promise<T> { }
const safePromise: SafePromise<number> = Promise.resolve(5);
/* eslint "@typescript-eslint/no-floating-promises": ["error", { allowForKnownSafePromises: [{ from: 'file', name: 'SafePromise' }] } ] */
safePromise.catch();
```
1. Here the rule finds out what the type of `safePromise` is and checks if it matches with any type present in the config file/comment. 
2. If it matches, the rule says it is safe (not a floating promise & doesn't report), else(if nothing is matched), it continues checking whether `.catch` has 1 arg or `.then` has 2 etc...

... since `safePromise.catch()` is a `CallExpression` & its callee's type is `MemberExpression`, thats why the `if` is located [here](https://github.com/typescript-eslint/typescript-eslint/pull/8502/files#diff-dec933bc2964d8570e62f3f2cd08e4237ee8e2dcfe894690afa5042095385089R285). & since there are so many `ExpressionStatement`s thats why the diff contains so many `if`s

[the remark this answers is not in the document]
kirkwaiblinger: I see what you're saying.... but I think we're talking past each other a tiny bit on the details.

Bringing it back to the simpler and more impactful corollary - shouldn't this flag? (current version of code doesn't)
```ts
class SafePromise<T> extends Promise<T> { }
const safePromise = new SafePromise<void>((resolve, reject) => { resolve() });
safePromise.then(() => { throw new Error(); });
```

`safePromise` is a `SafePromise`, of course, but the _whole_ expression is an _ordinary_ Promise, due to `.then` still returning a new ordinary Promise.

I get that `SafePromise` is "safe to-not-handle-rejection". But that wouldn't imply that it's safe to chain an unhandled ordinary `Promise` off of, right? From reading the issue and comments, I don't see somewhere that this part is spelled out explicitly (feel free to point me to it if it has already been discussed). But my current interpretation is that it should flag (unless, of course, `SafePromise` is modified to also override the `.then` to return another `SafePromise`).

Thoughts?

(The handlerless `.catch()` is a more subtle and less important version of the above, so my thought is to get to the same page about this first)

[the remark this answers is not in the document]
arka1002: Thanks for the comments, I did understood, tried to implement it locally & here are my few thoughts.

First, I dont think this rule ever checked whether the promises returned by `then`/`catch` are floating or not(with or without this PR)
[Minimum repro](https://typescript-eslint.io/play/#ts=5.4.3&fileType=.tsx&code=AoJw9gtglgzgpgOhHAVnAxgFwBQDYCUC6AhpugBbZwggAEAvAHy0DeAsAFC22bngDutAHZxBAURpgQVGvgDcnAL75aAelU9ycWgAdw0eLWSYAriBEATWgCMAnrQAGJMuQe1YtAOZxMmKEM8jVAxMOCtiISsQEwAbbQswOBghAHJMIJ0pdIjbXn9PTk5QSFhEZDQsPEJeOCFsAA8GZnqAGlpqOiZWTm5eAWFRWglwaQ75JRV1TW09EsNjM0sbewcaoTcPb198oIrQ8MijWPjE5LSMrNocvICgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDN4OBDfMwDmtYtA4BbSshTooiaBOiRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)
In both the cases, the promises returned by `then`/`catch` are getting rejected(run the code in browser console, it'll show `Uncaught ( in promise )`). So ... if the rule starts unflagging promises returned by `then` with 2 args(in your comment above), it might be a false positive(since they can also be floating).

Second, is it even possible for the rule to determine whether the promises returned `then`/`catch` are floating or not ? :sweat_smile:. It depends on purely on the `onRejected` function(not on the absence or presence of it). The 1st code snippet [here](https://github.com/typescript-eslint/typescript-eslint/pull/8502#discussion_r1586373572) is wrong ?

So, having a function as an argument in `catch`/`then` doesnt ensure that the promise returned by it will not get rejected(same as the 1st one). & I dont think this rule can ever decide whether the promise returned by it will get rejected or not(maybe :sweat_smile:) 

(Because its not only `throw` statements, anything can make it reject)
catch's promise rejection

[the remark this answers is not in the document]
kirkwaiblinger: I'm talking about `.then` _without_ a rejection handler.
The rule flags this:
```ts
promise.then(f);
```
_regardless_ of what `f` is. The question is whether it should flag this
```ts
safePromise.then(f);
```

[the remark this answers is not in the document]
arka1002: No, it shouldn't, 

Because unflagging can give a false positive.

We're talking about that promise which is returned by `then` right ? Not the promise on which `then` is attached.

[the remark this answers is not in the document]
kirkwaiblinger: > Because unflagging can give a false positive.

I'm not sure I follow... Could you explain what this means?

> We're talking about that promise which is returned by `then` right ? Not the promise on which `then` is attached.(both are different ones)

I'm saying this:
```ts
// (no error)
safePromise;

// error - expression evaluates to an ordinary unhandled promise
safePromise.then(f);
~~~~~~~~~~~~~~~~~~~
```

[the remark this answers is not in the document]
arka1002: This expression will still evaluate to an ordinary unhandled promise ( which will reject )
```ts
Promise.reject(1).then(x => x, err => {
  throw new Error(2);
});
```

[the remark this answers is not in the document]
kirkwaiblinger: That is a true statement (and if you're interested in the reasoning behind that design decision, you may be interested to read up starting from here: https://github.com/typescript-eslint/typescript-eslint/issues/5692#issuecomment-1504645034). 

However, that doesn't really impact this PR. A SafePromise should basically be granted exactly _one_ extra privilege over what the rule currently allows, and that is that a chain ending in a `SafePromise` (or, I suppose, `Array<SafePromise>`) should be ignored. Otherwise, a SafePromise should function the same as a Promise.

```ts
// unhandled
promise;
// handled!
safePromise;

// unhandled
[promise, promise, promise];
// handled
[safePromise, safePromise, safePromise];
// mixed should be unhandled
[safePromise, promise, safePromise];

// both unhandled (important!)
promise.then(f);
safePromise.then(f);

// both handled
promise.catch(f);
safePromise.catch(f);

// both handled, intentional design decision
promise.then(f, g);
safePromise.then(f, g);

// both unhandled 
promise.catch();
safePromise.catch();

// both unhandled (silly edge case)
promise.then(f, undefined);
safePromise.then(f, undefined);

// finally is basically chopped off the end for the purposes of this rule, so
// unhandled
promise.finally(f);
// handled
safePromise.finally(f);
```

Would you kindly update the PR with test cases for these (modifying the existing test cases that contradict these)? I think you'll find that it will also result in a much simpler implementation for the rule changes, as mentioned in the first post in this thread. 

Thanks!

arka1002:
Thanks for the reviews. I'm just writing down some notes for the future people who are going to stumble upon this pr.
1. `then` and `catch` functions with 1 and 0 arguments should be reported based on the above review comment. Here are the [tests](https://github.com/typescript-eslint/typescript-eslint/pull/8502/files#diff-520e9e574b74ad1c0a275b21b5ad4f70f5cc8b7c8717a4884c4e3c4e8162ffe8R2082-R2113).
2. About mixed arrays, its very interesting though, Typescript actually reduces the type of the array if it contains any branded type or anything else. Here is the [minimum repro](https://www.typescriptlang.org/play/?target=9#code/C4TwDgpgBAglC8UAKAnA9gWwJYGcIB4A7AVwwCMIUA+KAMigG8oALCEAfgC4odgUtCAcygBfANwAoAMYAbAIY4cUAEJQIAD2ARCAEyWpMuAiXKUaDERIFaUAMzlToAYTWbte5Omx4ipCtUZLCRkIYChbNDRuOEQDbwgAOhQIHDQZADcIAAoARgBKSRCwsjkUblVYryMklLTMrIAmAuDQqBKAL3buF0rDPBrUjOyAZmaWsIBHYikEKABtTz7E5MH6-IAacMioAF1JAHp9qGOAPXYJcagprABrWYW46pW67I220t2Do9Pzy5Q5LA6e6LeIDF65PKbDrtT4SQ7HKBnCRAA)(check the 2 slash comments). I have added [1 test](https://github.com/typescript-eslint/typescript-eslint/pull/8502/files#diff-520e9e574b74ad1c0a275b21b5ad4f70f5cc8b7c8717a4884c4e3c4e8162ffe8R2158-R2167). I think if in any future version typescript changes the type to `( Promise<number> | A )[]` ( which might be the correct one ), we have to change the logic in this rule.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -1855,5 +2053,136 @@ cursed();
...
+      `,
+      options: [{ allowForKnownSafePromises: [{ from: 'file', name: 'Bar' }] }],
+      errors: [{ line: 4, messageId: 'floatingPromiseArrayVoid' }],
+    },
+    {
+      code: `
+type SafePromise = Promise<number> & { hey?: string };
+let foo: SafePromise = Promise.resolve(1);
+let bar = [Promise.resolve(2), foo];
+bar;
```
kirkwaiblinger, line 2162: Thanks for the helpful explanation in https://github.com/typescript-eslint/typescript-eslint/pull/8502#issuecomment-2105734406!!

Consider leaving a little comment to that effect here with the test so that if this changes in a future TS update the author has a hint at why.

kirkwaiblinger (changes requested):

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -2079,19 +2079,19 @@ promise;
     },
     {
       code: `
-interface UnsafeThenable<T> {
+interface SafeThenable<T> {
```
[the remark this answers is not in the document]
arka1002: Yes, there's an issue which exactly describes `Thenable` in this rule - https://github.com/typescript-eslint/typescript-eslint/issues/8433.(It also says thenables should be valid)
It'll be solved there.

PR #8509, opened 2024-02-19
feat(eslint-plugin): [no-misused-spread] add new rule

### LINKED ISSUE #748: Rule proposal: prevent array, iterable and function spreads into objects

StyleShit:
Closes #748


packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
```diff
@@ -0,0 +1,735 @@
...
+      `,
+      errors: [
+        {
+          messageId: 'noFunctionSpreadInObject',
+          line: 4,
+          column: 21,
+          endColumn: 37,
+        },
+      ],
+    },
```
StyleShit, line 656: I can't figure out why the type of `getFunction()` doesn't have the `SymbolFlags.Function`flag set. 
Will be glad for some help here 😄

StyleShit: never mind, fixed

StyleShit:
Not sure why the tests are failing, seems to pass locally

packages/eslint-plugin/docs/rules/no-misused-spread.md
```diff
@@ -0,0 +1,57 @@
...
+
+declare function getObj(): { a: 1; b: 2 };
+const getObjSpread = { ...getObj() };
+```
+
+<!--/tabs-->
+
+## When Not To Use It
+
+When you want to allow using the spread operator on types that can't be spread.
```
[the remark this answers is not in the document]
StyleShit: Done (not sure it's the best one though...)

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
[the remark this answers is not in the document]
StyleShit: Ooh those are good ones!

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
```diff
@@ -0,0 +1,820 @@
...
+
+      f.prop = 1;
+
+      const o = { ...f };
+    `,
+
+    `
+      const f = () => {};
+
+      f.prop = 1;
```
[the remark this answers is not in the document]
StyleShit: Haha, it was in the issue. wouldn't have thought about it otherwise 😅

packages/eslint-plugin/docs/rules/no-misused-spread.md
```diff
@@ -0,0 +1,57 @@
+---
+description: "Disallow using the spread operator on types that can't be spread."
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-misused-spread** for documentation.
+
+The [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) (`...`) is a powerful feature in JavaScript, but it can only be used with types that support it. This rule disallows using the spread operator on types that can't be spread, or types that spreading them in specific cases can lead to unexpected behavior.
```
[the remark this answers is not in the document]
StyleShit: Sounds better :)

packages/eslint-plugin/docs/rules/no-misused-spread.md
```diff
@@ -0,0 +1,57 @@
+---
+description: "Disallow using the spread operator on types that can't be spread."
```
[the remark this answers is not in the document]
StyleShit: Yeah...
tried something else, WDYT?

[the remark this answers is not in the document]
JoshuaKGoldberg: I like it!

...and just noticed my suggestion had a blatant grammatical mistake. Oops. Heh.

packages/eslint-plugin/docs/rules/no-misused-spread.md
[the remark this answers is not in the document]
JoshuaKGoldberg: It's great!

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
```diff
@@ -0,0 +1,904 @@
...
+    `
+      function* generator() {}
+
+      generator.prop = 1;
+
+      const o = { ...generator };
+    `,
+
+    `
+      declare const promise: Promise<number>;
```
JoshuaKGoldberg, line 168: [Bug] Err, wouldn't we want the rule to flag on this case? And on the `{ ...new A() }` case? ...and now that we're looking at it, `{ ...(class A { static value = 1 }) }` and other `static`-y things?

As a prospective user of this rule, I would have expected instances of classes to not be allowed. Am I totally off base here?

My hunch is that this is a discovered edge case that should have an option. WDYT?

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,185 @@
...
+    type: 'problem',
+    docs: {
+      description:
+        'Disallow using the spread operator when it might cause unexpected behavior',
+      recommended: 'strict',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      noStringSpreadInArray:
+        "Using the spread operator on a string is not allowed in an array. Use `String.split('')` instead.",
```
JoshuaKGoldberg, line 30: [Docs] Something like this? _"is not allowed"_ is a bit too prescriptive.

```suggestion
        "Using the spread operator on a string can cause unexpected behavior. Prefer `String.split('')` instead.",
```

...and similar changes for the others?

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,185 @@
...
+
+      noSpreadInObject:
+        'Using the spread operator on `{{type}}` is not allowed in an object.',
+
+      noFunctionSpreadInObject:
+        'Using the spread operator on `Function` without properties is not allowed in an object. Did you forget to call the function?',
```
JoshuaKGoldberg, line 36: > without properties

Without ... _added_ properties? Some might point out that functions do have built-in properties.

JoshuaKGoldberg (changes requested):
Progress! 🚀 

Sorry, I'm only now realizing the class instance stuff might need an option. I think I've look at this PR too much - we should get someone else from @typescript-eslint/triage-team to look too. 😅

packages/eslint-plugin/src/rules/no-misused-spread.ts
StyleShit: haha sure, good idea

packages/eslint-plugin/src/rules/no-misused-spread.ts
StyleShit: yup, I like it

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
StyleShit: IMO, spreading a class instance is weird and probably an edge case as you're saying.

Not sure about spreading a class declaration though... it looks too obscure. 
Is there anyone who actually does that? 😅

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,255 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Disallow using the spread operator when it might cause unexpected behavior',
+      recommended: 'strict',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      noStringSpreadInArray:
```
JoshuaKGoldberg, line 37: [Naming] The rule also flags strings spread in objects:

```suggestion
      noStringSpread:
```

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,255 @@
...
+        "Using the spread operator on a string can cause unexpected behavior. Prefer `String.split('')` instead.",
+
+      noSpreadInObject:
+        'Using the spread operator on `{{type}}` can cause unexpected behavior.',
```
JoshuaKGoldberg, line 41: [Docs] I think we'll need to be precise about when the spread operators are or aren't allowed. Spreading an Iterable in an array is intentionally allowed.

```suggestion
        'Using the spread operator on `{{type}}` in an object can cause unexpected behavior.',
```

JoshuaKGoldberg (changes requested):
Ok, thanks for bearing with us here. Coming back to this PR I think I have a better feeling & understanding for it.

In general, I actually think it's trying a bit harder than it needs to to give the right errors. With a bit less logic & a bit more coaching in the messages, I think it'll be more accurate & precise for users. 

Great stuff so far! ⭐

JoshuaKGoldberg:
👋 Ping, just checking in @StyleShit - is this still something you have time to make progress on? Or is there anything you're waiting on us on?

packages/eslint-plugin/src/rules/no-misused-spread.ts
StyleShit: hmm, I think it's a bug, it shouldn't flag string spread in objects since it's being handled by TS already. WDYT?

[playground](https://deploy-preview-8509--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=MYewdgzgLgBAhjAvDA3jAdJg5HCATLGAXwG4g&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oFtLlZkiACa1kxaIgCGQ9FETRoHaJHBgAviDVA&tsconfig=&tokens=false)

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,255 @@
...
+
+      noFunctionSpreadInObject:
+        'Using the spread operator on `Function` without additional properties can cause unexpected behavior. Did you forget to call the function?',
```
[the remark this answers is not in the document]
StyleShit: Done

packages/eslint-plugin/docs/rules/no-misused-spread.md
```diff
@@ -0,0 +1,100 @@
+---
+description: 'Disallow using the spread operator when it might cause unexpected behavior.'
+---
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-misused-spread** for documentation.
+
+The [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) (`...`) is a powerful feature in JavaScript, but it can only be used with types that support it. This rule disallows using the spread operator on types where spreading can lead to unexpected behavior.
```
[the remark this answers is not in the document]
StyleShit: Done

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,255 @@
...
+
+      noClassSpreadInObject:
+        'Using the spread operator on class instances can cause unexpected behavior.',
```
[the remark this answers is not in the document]
StyleShit: I don't have any better phrasing either 😅

packages/eslint-plugin/src/rules/no-misused-spread.ts
StyleShit: Done

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,255 @@
...
+    return type.types.some(t => isIterable(t, checker));
+  }
+
+  const iterator = tsutils.getWellKnownSymbolPropertyOfType(
+    type,
+    'iterator',
+    checker,
+  );
+
+  return iterator !== undefined;
```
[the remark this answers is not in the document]
StyleShit: ooh, thanks!

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,255 @@
...
+        context.report({
+          node,
+          messageId: 'noSpreadInObject',
+          data: {
+            type: 'Promise',
+          },
+        });
+
+        return;
+      }
```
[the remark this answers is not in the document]
StyleShit: Removed the map/set checks. Those are iterable classes, so it's being caught in the `Iterable` check.

Regarding the promise - I'm not sure why, but calling `isClassInstance` doesn't flag it and I couldn't find a way to flag it as a general class instance (maybe `isSymbolFromDefaultLibrary` can help here? IDK)

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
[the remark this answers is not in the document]
StyleShit: Done :)

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
StyleShit: Done

StyleShit:
@JoshuaKGoldberg 

You pinged my right on the day I was able to work on it 😄

StyleShit:
The tests are passing locally, not sure why it fails here

packages/eslint-plugin/src/rules/no-misused-spread.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Maybe the tests don't have a high enough target for `Promise` to be a `class` rather than an `interface`?

packages/eslint-plugin/src/rules/no-misused-spread.ts
JoshuaKGoldberg: Ah, if TypeScript already handles something, then yeah, we don't need to.

StyleShit: Fixed

packages/eslint-plugin/docs/rules/no-misused-spread.md
auvred, line 1: [tests] After https://github.com/typescript-eslint/typescript-eslint/pull/8497 we render snapshots of ESLint output for each code sample in docs. Could you please merge `main` and update the snapshots with `yarn test docs -u`?

packages/eslint-plugin/docs/rules/no-misused-spread.md
```diff
@@ -0,0 +1,100 @@
...
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-misused-spread** for documentation.
+
+The [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) (`...`) is a powerful feature in JavaScript that can be misused in ways not always detectable by TypeScript. This rule disallows using the spread operator on types where spreading can lead to unexpected behavior.
+
+## Examples
+
+<!--tabs-->
```
auvred, line 13: [docs] The documentation is now being built with Docusaurus v3: https://github.com/typescript-eslint/typescript-eslint/pull/8209. Could you please update these `<!--tabs-->` comments to `<Tabs>` MDX tags?

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,201 @@
...
+    .some(
+      t => !!tsutils.getWellKnownSymbolPropertyOfType(t, 'iterator', checker),
+    );
+}
+
+function isString(type: ts.Type): boolean {
+  return tsutils
+    .typeParts(type)
+    .some(t => !!isTypeFlagSet(t, ts.TypeFlags.StringLike));
+}
```
auvred, line 185:
```suggestion
function isString(type: ts.Type): boolean {
  return tsutils
    .unionTypeParts(type)
    .some(unionPart => tsutils.intersectionTypeParts(unionPart)
      .some(intersectionPart => !!isTypeFlagSet(t, ts.TypeFlags.StringLike))
    );
}
```

`typeParts` iterates through the first level of `.types` property only. Looks like a bug:

```ts
declare const aa: (string & { foo: 1 }) | (string & { bar: 1 })
[...aa] // not reported - bug

declare const bb: (string & { foo: 1 })
[...bb] // reported - ok
```

[playground](https://deploy-preview-8509--typescript-eslint.netlify.app/play/#ts=5.4.3&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3lKAueAFJjAJYoDm8AZPAN7wBmSS%2BAjPAL4CU8APocTKUa9AEaw2nLgCgA2gDpFOALrwA9GvgokWOAAckMDCGDwAtPFEBXctOmhIsBMnRZRo-EQykK1Oo2ZJbjlFeXdVDXh9Q2NTCyQAayA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oFtLlZkiACa1kxaIgCGQ9FETRoHaJHBgAviDVA&tsconfig=&tokens=false)

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,201 @@
...
+  return tsutils
+    .typeParts(type)
+    .some(
+      t => t.getCallSignatures().length > 0 && t.getProperties().length === 0,
+    );
+}
+
+function isPromise(program: ts.Program, type: ts.Type): boolean {
+  return tsutils.typeParts(type).some(t => isPromiseLike(program, t));
+}
```
auvred, line 197:
```suggestion
```

`isPromiseLike` already contains the type part iteration logic :smile: I think there is no need to do `typeParts` here!

https://github.com/typescript-eslint/typescript-eslint/blob/f248e689b04bdfd81429977a731ba6b8ad82622e/packages/type-utils/src/builtinSymbolLikes.ts#L132-L141

auvred (changes requested):

packages/eslint-plugin/docs/rules/no-misused-spread.md
StyleShit: Done, thanks!

packages/eslint-plugin/docs/rules/no-misused-spread.md
StyleShit: Done

packages/eslint-plugin/src/rules/no-misused-spread.ts
StyleShit: Great catch!

packages/eslint-plugin/src/rules/no-misused-spread.ts
StyleShit: seems like it doesn't catch this case:

```TS
declare const maybePromise: Promise<number> | { a: number };
const o = { ...maybePromise };
```

and I think we should catch it, because it means we might spread a promise in runtime, right?

anyway, see this comment:
https://github.com/typescript-eslint/typescript-eslint/pull/8509#discussion_r1528451293

I'll be glad to get some help here with removing this check completely 😄 
Wasn't able to figure out why `isClassOrInterface` didn't work here

bradzacher (changes requested):
A few questions, one blocker.

---

Feel free to turn this rule on for this repo.
It'll be a good test case and we'll want this rule anyways.
Eventually (probably v9) I can see us turning this on as a recommended rule.

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,189 @@
...
+        "Using the spread operator on a string can cause unexpected behavior. Prefer `String.split('')` instead.",
+
+      noSpreadInObject:
+        'Using the spread operator on `{{type}}` in an object can cause unexpected behavior.',
+
+      noFunctionSpreadInObject:
+        'Using the spread operator on a function without additional properties can cause unexpected behavior. Did you forget to call the function?',
+
+      noClassInstanceSpreadInObject:
+        'Using the spread operator on class instances without `[Symbol.iterator]` can cause unexpected behavior.',
```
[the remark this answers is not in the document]
StyleShit: I... don't know why I wrote that 😂 
yup, that's wrong, I don't even check that in the rule...


changed the message

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,189 @@
...
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          allowClassInstances: {
+            description:
+              'Whether to allow spreading class instances in objects.',
+            type: 'boolean',
+          },
+        },
```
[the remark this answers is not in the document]
StyleShit: huh, I just proposed the same in [this comment](https://github.com/typescript-eslint/typescript-eslint/pull/8509#discussion_r1617448501) 😄 

will do

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,189 @@
...
+      }
+
+      if (node.argument.type === AST_NODE_TYPES.ClassExpression) {
+        context.report({
+          node,
+          messageId: 'noClassDeclarationSpreadInObject',
+        });
+
+        return;
+      }
```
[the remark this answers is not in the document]
bradzacher: RC = "request change"

----

Currently this check is a pure syntactic check. It is specifically only looking for this case: `({...(class C {})})`

This is, obviously, fairly limited and very specific. Instead we should be looking at the type of the thing being spread to determine if it is a class declaration or not so we can support cases like

```ts
class A {}
const B = class {};
const C = Set<string>;

({...A});
({...B});
({...C});
```

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,189 @@
...
+        context.report({
+          node,
+          messageId: 'noSpreadInObject',
+          data: {
+            type: 'Iterable',
+          },
+        });
+
+        return;
+      }
```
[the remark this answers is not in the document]
bradzacher: I'm just not familiar enough with custom iterables to know for sure.

Cc @typescript-eslint/triage-team

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,189 @@
...
+        context.report({
+          node,
+          messageId: 'noSpreadInObject',
+          data: {
+            type: 'Promise',
+          },
+        });
+
+        return;
+      }
```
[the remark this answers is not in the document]
bradzacher: My 2c:
I'm okay with having such a check here as it's possible to have one rule but not the other.

If we are going to special-case promise in this rule then the error should also be special-cased as well - eg by adding "did you forget to await the promise?"

packages/eslint-plugin/src/rules/no-misused-spread.ts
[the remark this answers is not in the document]
bradzacher: Note: It looks like the "array in object spread" case is handled via the "iterator" handling.

If we do remove this handling or put this handling behind an option then we should special-case arrays and tuples so they're always banned, given that that's a concrete, well-defined cases that you'd never ever want, and because it's easy to catch from the type system side of things (there's utils to detect the array type)

packages/eslint-plugin/src/rules/no-misused-spread.ts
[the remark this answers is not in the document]
StyleShit: Sounds good to me
done :)

packages/eslint-plugin/src/rules/no-misused-spread.ts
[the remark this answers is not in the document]
StyleShit: fixed

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -67,7 +115,13 @@ export default createRule<Options, MessageIds>({
 
   defaultOptions: [
     {
+      allowStrings: false,
+      allowPromises: false,
+      allowFunctions: false,
+      allowMaps: false,
```
bradzacher, line 121: My 2c: I think that we don't want `allowPromises` or `allowMaps`.
Like the Array case - these cases are special cases that are always going to be bad.

StyleShit: gotcha, done

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
```diff
@@ -95,6 +181,41 @@ If you want to allow this behavior, you can configure the rule with `"allowClass
 }
 ```
 
+### `allowClassDeclarations`
+
+By default, this rule disallows using the spread operator on class declarations:
```
omril1, line 186: I don't like the name of this option, as the class could also be passed as a reference or be generated by a class factory.
I asked ChatGPT for a better name and it suggested "class reference" as an alternative, WDYT?

StyleShit: I personally don't have any strong opinion about either of the names

Maybe let's wait for one of the maintainers to decide?

Josh-Cena: I would suggest `allowClasses`. Any concerns?

packages/eslint-plugin/src/rules/no-misused-spread.ts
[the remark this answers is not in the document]
Josh-Cena: Here's what I know: you would almost never have a custom iterable that actually spreads and remains iterable. This is because usually `[Symbol.iterator]()` is defined on the class (such as `Set.prototype[@@iterator]`), not as an own property of each instance, and it's usually bad practice to make methods enumerable in any case.

I would suggest forbidding all iterables from being spreaded as objects, with an option that allows certain types. There's probably no reason you want to spread maps or sets but there might be reasons to spread other iterables.

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
omril1: @Josh-Cena It's better than declaration, but `allowClasses ` can be understood (or confused) as class instances which is another option.

Josh-Cena: Right, but "class instances" and "classes" are obviously different things and the fact that we already have an "instances" option further mitigates that.

omril1: Well, yes, assuming you read about both in proximity, seeing an eslint-disable comment or config option can lose the context

Josh-Cena: Also classes are just sugared classes so I'm not sure if we want to treat them any more specially than functions or anything with a `new` signature...

bradzacher (approved):
this is looking good to me, great work!

I'm not fussed about `allowClassDeclarations` vs `allowClasses`, personally.

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,289 @@
...
+}
+
+function isMap(program: ts.Program, type: ts.Type): boolean {
+  return isTypeRecurser(
+    type,
+    t =>
+      isBuiltinSymbolLike(program, t, 'Map') ||
+      isBuiltinSymbolLike(program, t, 'ReadonlyMap'),
+  );
+}
```
auvred, line 278: Shouldn't it catch `WeakMap` as well?

This case is not reported

```ts
declare const map: WeakMap<{foo: string}, number>;
const mapSpread = { ...map };
```

[playground](https://deploy-preview-8509--typescript-eslint.netlify.app/play/#ts=5.4.5&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gWygBwC54B1EKAawFl8AeAbwDMkkjMYBLFAcwF8AaeCgCu2AEYgYAPgDcAKGTosuPAGU8cKMHgBeeHXgA6I8vg9pQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oFtLlZkiACa1kxaIgCGQ9FETRoHaJHBgAviDVA&tsconfig=&tokens=false)

auvred (changes requested):
One question about `WeakMaps`, otherwise LGTM!

Josh-Cena (commented):
If we have separate options for `allowStrings` and `allowIterables`, we should have one for `allowArrays` too. And TBH, I don't think `allowIterables` is useful as an option because iterables have drastically different shapes and functions. Some iterables can be perfectly used as plain objects while other like `Map` and `Set` are basically useless, so I personally would never turn it on, neither can I imagine it not being a footgun to anyone else.

Do `allowClassDeclarations` (which again I would strongly recommend renaming) and `allowClassInstances` have tests for built-in types such as `new RegExp` or `new HTMLElement` (which use a slightly different convention to allow declaration merging)? Are they intended to be supported?

StyleShit:
> If we have separate options for `allowStrings` and `allowIterables`, we should have one for `allowArrays` too. And TBH, I don't think `allowIterables` is useful as an option because iterables have drastically different shapes and functions. Some iterables can be perfectly used as plain objects while other like `Map` and `Set` are basically useless, so I personally would never turn it on, neither can I imagine it not being a footgun to anyone else.

Yeah, I get your point here.
So do you think we should stop reporting iterables unless they're of specific and known types (`Set`, `Array`, `String`, etc.)?

In contrast, you've said [here](https://github.com/typescript-eslint/typescript-eslint/pull/8509#discussion_r1623203725) that we should forbid spreading any iterables and allow certain types. So I'm not sure whether I understood you correctly 😅 


> Do `allowClassDeclarations` (which again I would strongly recommend renaming) and `allowClassInstances` have tests for built-in types such as `new RegExp` or `new HTMLElement` (which use a slightly different convention to allow declaration merging)? Are they intended to be supported?

There are no tests for these specific cases, but I've added tests for `new Date` IIRC.
Seems like the rule [reports these cases correctly](https://deploy-preview-8509--typescript-eslint.netlify.app/play/#ts=5.4.5&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3nA5iAHgFzwBKIOAovgA4DcAUKJLAsuliBCALYgobEAEgBUAsgBkKXXvwb02meFHgBeeAG94AOh24CAGm07OPPlgC%2BtIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oFtLlZkiACa1kxaIgCGQ9FETRoHaJHBgAviDVA&tsconfig=&tokens=false) BTW

Can you please explain (or point me to an explanation about) the difference in their declaration in the context of this rule? 
I'm not sure that I understand what you mean here, but it might explain why [this](https://github.com/typescript-eslint/typescript-eslint/pull/8509#discussion_r1658439063) happens(?)

Josh-Cena:
Sorry I missed the last reply.

My main point on `allowIterables` is this:

> so I personally would never turn it on, neither can I imagine it not being a footgun to anyone else.

i.e. it shouldn't exist as an option and iterables should be reported by default, with options to allow certain types.

> Can you please explain (or point me to an explanation about) the difference in their declaration in the context of this rule?

Built-in classes are not declared as classes. They look like this:

```ts
interface WeakMapConstructor {
  new (): WeakMap;
  // ...
}

interface WeakMap {
  // ...
}

var WeakMap: WeakMapConstructor;
```

This approach allows versioned declarations to be merged.

StyleShit:
> i.e. it shouldn't exist as an option and iterables should be reported by default, with options to allow certain types.

So you're proposing to change the `allowIterables` option to accept an array of types rather than a boolean?

> This approach allows versioned declarations to be merged.

Cool! thanks for the explanation 😄

packages/eslint-plugin/src/rules/no-misused-spread.ts
StyleShit: So after @Josh-Cena's [comment](https://github.com/typescript-eslint/typescript-eslint/pull/8509#issuecomment-2289631409), I think (?) that now I understand why this is happening.

Is there any similar check in the codebase that I can take as a reference to solve it?

Josh-Cena:
> So you're proposing to change the `allowIterables` option to accept an array of types rather than a boolean?

I think if we go down that path we should just have a generic `allow` array. Otherwise all the difference is that we also validate if the allowed type is also iterable, which doesn't appear necessary.

StyleShit:
> Otherwise all the difference is that we also validate if the allowed type is also iterable

Yeah, ok, makes sense!

@bradzacher @JoshuaKGoldberg, thoughts?

JoshuaKGoldberg:
I agree with Josh. 🙂

StyleShit:
@Josh-Cena @JoshuaKGoldberg 

Added an `allowForKnownSafeIterables` option to align with `allowForKnownSafePromises`, WDYT?

Josh-Cena:
Why `allowForKnownSafeIterables` instead of a generic `allow` array for any kind of type?

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
StyleShit, line 1: the docs are failing the tests and I can't figure out why. The only thing I get is `Validating rule docs › no-misused-spread.mdx › code examples ESLint output` which isn't very helpful considering all of my code examples seem to be working

StyleShit:
> Why `allowForKnownSafeIterables` instead of a generic `allow` array for any kind of type?

Well, I need a condition for reporting a type before allowing it, right?
which means something like:

```TS
if (shouldBeReported() && !isAllowed()) {
	report();
}
```

I mean... what will be the flags for reporting types if there is a generic "allow" option?
wouldn't it flag basically anything?

```TS
if (! isAllowed()) {
	report();
}
```

Unless I'm missing something here and/or not understanding you correctly

Josh-Cena:
I meant—currently, the rule looks like this:

```js
if (type.isString()) report();
else if (type.isFunction()) report();
else ...
else if (type.isIterable() && !allowForKnownSafeIterables.includes(type)) report();
```

I'm saying why can't it be:

```js
if (allow.includes(type)) return;
if (type.isString()) report();
else if (type.isFunction()) report();
...
```

This is not going to be much heavier since we have to query the type for every spread anyway. This is going to allow us to ignore only certain classes or certain class instances, in addition to certain iterable types.

Put another way, it could also look like this:

```js
function maybeReport(type) {
  if (allow.includes(type)) return;
  report();
}

if (type.isString()) maybeReport(type);
else if (type.isFunction()) maybeReport(type);
...
```

StyleShit:
Oh, I see, good point!

implemented, and it simplified the code a little bit

Josh-Cena (commented):
Now that we no longer have granular options (an idea I can fully get behind), shall we explain in prose what kind of entities are banned?

Also, sort of a general question, but how do I allow a specific _class_ (not its instances) to be spread? That would require `typeof SomeClass` right?

StyleShit:
> shall we explain in prose what kind of entities are banned?

Good idea, done.
Do you think we should also explain why?

> how do I allow a specific class (not its instances) to be spread

Do you mean something like this?

https://github.com/typescript-eslint/typescript-eslint/pull/8509/files#diff-ce49ac2672eac5b4d85a5c99fb18aaa335f867b71ab0e7a6e4dde6cbd476ec78R285-R294

Josh-Cena:
That looks odd to me. So `allow: ["A"]` would allow both `class A` and instances of this class to be spread? Looks like this is the way `typeMatchesSpecifier` works. Hmmm.

bradzacher (dismissed):
I'm personally happy with it how it is right now.
I think that the config surface looks good and the tests are looking good at a glance at least.

I'm okay if we land this to get it out there and we can iterate based on user feedback in the real world.

It's turned on in our repo now cos it's in the strict config and there are no false positives, at least.

Once we fix the tests I'm 👍 to land

JoshuaKGoldberg (commented):
LGTM! A nice and straightforward implementation. A few spot touchups here and there, and the test failures are blocking.

JoshuaKGoldberg (changes requested):
_(wrong button)_

packages/eslint-plugin/src/rules/no-misused-spread.ts
```diff
@@ -0,0 +1,278 @@
...
+
+function isMap(program: ts.Program, type: ts.Type): boolean {
+  return isTypeRecurser(
+    type,
+    t =>
+      isBuiltinSymbolLike(program, t, 'Map') ||
+      isBuiltinSymbolLike(program, t, 'ReadonlyMap') ||
+      isBuiltinSymbolLike(program, t, 'WeakMap'),
+  );
```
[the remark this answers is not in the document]
StyleShit: Ooh, thanks!

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
```diff
@@ -0,0 +1,140 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-misused-spread** for documentation.
+
+The [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) (`...`) is a powerful feature in JavaScript that can be misused in ways not always detectable by TypeScript. This rule disallows using the spread operator on types where spreading can lead to unexpected behavior.
```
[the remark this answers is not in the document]
StyleShit: haha, no idea if it was me or copilot, so I'll blame copilot here 😅 

removed

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+This option allows marking specific types as "safe" to be spread. It takes an
+array of type specifiers to consider safe.
+
+Each item in the array must have one of the following forms:
+
+- A type defined in the current file, or globally available (`"SafeType"`)
+- A type defined in a file (`{ from: "file", name: "SafeString", path: "src/safe-string.ts" }` with `path` being an optional path relative to the project root directory)
+- A type from the default library (`{ from: "lib", name: "BrandedArray" }`)
+- A type from a package (`{ from: "package", name: "ThisIsSafe", package: "safe-lib" }`, this also works for types defined in a typings package).
```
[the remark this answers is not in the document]
StyleShit: yay! great

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
```diff
@@ -0,0 +1,140 @@
...
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+If you intentionally want to use the spread operator in those cases, and expect
+the specific behavior that comes with it, you might not want this rule.
+For example, when you want to spread an array into an object and expect the
+result to be an object with the array elements as values and the array indices
+as keys.
```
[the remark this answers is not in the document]
StyleShit: Done

packages/eslint-plugin/src/rules/no-misused-spread.ts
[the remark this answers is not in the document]
StyleShit: Thank you! 🥰



(always trying to treat my code as art 😅)

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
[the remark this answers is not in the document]
StyleShit: ugh, I just can't figure it out!

The only thing I was able to figure out was that it probably happens because `getConstrainedTypeAtLocation` gives me `WeakRef`, which is a reference to `WeakRefConstructor`, and not the actual type itself:

https://github.com/microsoft/TypeScript/blob/48f0b3cc383bfd6337fc9b5a0afdc1642473a57d/src/lib/es2021.weakref.d.ts#L23

I thought about creating a `getActualType(type: ts.Type)` that will resolve the referenced type.

I was able to get the symbol of the `WeakRefConstructor` type, but then I got stuck basically:

```TS
function getActualType(type: ts.Type, checker: ts.TypeChecker): ts.Type {
  const valueDeclaration = type.getSymbol()?.valueDeclaration;

  if (
    isVariableDeclaration(valueDeclaration) &&
    valueDeclaration.type?.kind === ts.SyntaxKind.TypeReference
  ) {
    const allSymbols = checker.getSymbolsInScope(
      valueDeclaration,
      ts.SymbolFlags.Type,
    );

    const referencedTypeSymbol = allSymbols.find(
      symbol =>
        symbol.name ===
        (valueDeclaration.type as ts.TypeReferenceNode).typeName.getText(),
    );

    if (!referencedTypeSymbol) {
      return type;
    }

	// resolves to error type
    const referencedTypeNode = checker.getTypeOfSymbol(referencedTypeSymbol);
  }

  return type;
}

function isVariableDeclaration(
  type: ts.Node | undefined,
): type is ts.VariableDeclaration {
  return type?.kind === ts.SyntaxKind.VariableDeclaration;
}
```

packages/eslint-plugin/tests/fixtures/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "module": "commonjs",
     "strict": true,
     "esModuleInterop": true,
-    "lib": ["es2015", "es2017", "esnext"],
+    "lib": ["es2015", "es2017", "esnext", "DOM"],
```
StyleShit, line 8: this fixes the `HTMLElement` tests

packages/eslint-plugin/src/rules/no-misused-spread.ts
JoshuaKGoldberg, line 1: Looks like this caught an existing violation in our codebase:

```plaintext
/home/runner/work/typescript-eslint/typescript-eslint/packages/website/src/theme/NotFound/Content/index.tsx
  16:17  error  Using the spread operator on a string can cause unexpected behavior. Prefer `String.split('')` instead  @typescript-eslint/no-misused-spread
```

`yarn lint` on the repository will need to pass.

_(posting this comment here because GitHub won't let me post on an unchanged file)_

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-misused-spread** for documentation.
+
+The [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) (`...`) is a JavaScript
+feature that can be misused in ways not always reported by TypeScript. This rule disallows using the spread operator on types where
+spreading can lead to unexpected behavior.
+
+This rule disallows the spread operator in the following cases:
```
JoshuaKGoldberg, line 16: > Do you think we should also explain why each case is disallowed?

[Docs] Yes 🙂. But I'm ok with that being a followup to this PR, IME these kinds of detailed many-cases-explained docs discussions can take a while.

JoshuaKGoldberg (changes requested):
Awesome!! I think the only blocker from me is fixing existing violations from this repo.

I took the liberty of merging the latest `main` in to fix the unrelated CI failures.

StyleShit:
> Awesome!! I think the only blocker from me is fixing existing violations from this repo.
> 
> I took the liberty of merging the latest `main` in to fix the unrelated CI failures.

@JoshuaKGoldberg 

Seems like my most important comment here is hidden, so I'm assuming you haven't seen it 😅 
https://github.com/typescript-eslint/typescript-eslint/pull/8509#discussion_r1752527309

The tests will still fail 😞

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: See `builtinSymbolLikes.ts` and `isSymbolFromDefaultLibrary.ts`. You could roughly co-opt those, or failing that, the logic within them.

The worst case scenario would be an allowlist of known buitlins that includes `'WeakRef'`... which is fine IMO if all else doesn't work.

packages/eslint-plugin/src/rules/no-misused-spread.ts
StyleShit: haha, seems like we have conflicting rules here 😅 

image


image

JoshuaKGoldberg: lol. cc @abrahamguo @kirkwaiblinger from #9834 - looks like this just blatantly disagrees with https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1489 -> https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-spread.md.

Maybe we should just not go for strings at all to start? Seems like maybe it's use-case specific?

StyleShit: I don't have any strong opinion for either of the sides, so... 🤷‍♂️ 

(it's also more of a style thing rather than something that could cause errors, no?)

any objection to removing it @abrahamguo @kirkwaiblinger?

Josh-Cena: Here are my suggestions.

1. Never ever use `split("")`. It doesn't do what you think it does for many scripts.
2. `[...str]` is fine _by principle_ but may still be surprising and not have the intended behavior, especially for emojis. In fact TC39 now considers strings being iterable as a mistake and all future built-in APIs that accept iterables will reject strings by default. If you really want to spread strings, either disable this rule or use `Iterator.from`, or think about alternatives such as `Intl.Segmenter`.
3. An option to allow it is good enough. We can consider changing the default to `true`, but my vote is to be strict by default.

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+- Spreading a string into an array
```
Josh-Cena, line 18: As a quick remedy we can link to https://www.xanthir.com/b4wJ1, which I believe is often quoted in TC39 discussions.

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
```diff
@@ -0,0 +1,141 @@
...
+- Spreading a `Promise` into an object
+- Spreading a function without properties into an object
+- Spreading a class instance or declaration into an object
+- Spreading an iterable (`Map`, `Array`, etc.) into an object
```
Josh-Cena, line 22: I think we should have one sentence explanations for each, such as:

```suggestion
- Spreading a `Promise` into an object. You probably meant to `await` it.
- Spreading a function without properties into an object. You probably meant to call it.
- Spreading a class instance into an object. This does not faithfully copy the instance because only its own properties are copied, but the inheritance chain is lost, including all its methods.
- Spreading a class (defined with a `class` declaration or `class` expression) into an object. This copies all static own properties of the class, but none of the inheritance chain.
- Spreading an iterable (`Map`, `Array`, etc.) into an object. Iterable objects usually do not have meaningful enumerable properties and you probably meant to spread it into an array instead.
```

packages/eslint-plugin/tests/fixtures/tsconfig.json
abrahamguo: This seems like it needs further clarification from #9648

packages/eslint-plugin/src/rules/no-misused-spread.ts
abrahamguo: @Josh-Cena do you have a source for #2?

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
```diff
@@ -0,0 +1,141 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-misused-spread** for documentation.
+
+The [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) (`...`) is a JavaScript
```
abrahamguo, line 12: `Spread syntax` (spread is not an operator)

packages/eslint-plugin/src/rules/no-misused-spread.ts
Josh-Cena: Already mentioned below: https://www.xanthir.com/b4wJ1, which has been quoted by many other TC39 delegates too.

JoshuaKGoldberg: +1 from me to everything Josh has indicated. I think this is a really interesting case of the language going one direction (`...` iterable strings), lint rules following that direction, then the language realizing the direction being wrong. If other lint rules are asking to use `...` on strings then we should file an issue on them saying they should update.

abrahamguo: @JoshuaKGoldberg Sure. However, if you are choosing between `str.split('')` vs `[...str]`, don't you agree that `[...str]` is certainly more correct than `str.split('')`, since `str.split('')` will break surrogate pairs??

JoshuaKGoldberg: To be honest, I'm coming back from a couple of weeks of intense conference organizing, and don't have the headspace to re-think this deeply right now. I trust @Josh-Cena's opinion on things. 😄

Josh-Cena: Both `split("")` and `...` are only correct under certain assumptions, and practically, most of the time, the former is not much more wrong than the latter. It's fine for unicorn to recommend spreading since it's nominally safer, and if users accept safety at the level of spreading they are free to configure the allow config, but banning it by default exposes such unsafety.

Josh-Cena:
Hi @StyleShit do you think you could find some time recently to finish it? Otherwise we could also take it on since there isn't a lot left

StyleShit:
> Hi @StyleShit do you think you could find some time recently to finish it? Otherwise we could also take it on since there isn't a lot left

@Josh-Cena 

I'm currently on reserve duty so I don't know whether I'll have time in the near future 😓 
If you have some time to finish this PR, that would be great!
(honestly, I don't remember how much work is left here)

Thank you! 🙏

JoshuaKGoldberg:
From https://github.com/typescript-eslint/typescript-eslint/pull/8443#issuecomment-2466798013:

> Cool! I'd really like to get this one and #8509 in by the end of the year, they're great rules to have that I've wanted for a while. So let's timebox these two to, say, the end of the month. If you don't end up having time by then, no worries, one of us on the team can send in a PR with the same commit history & a co-author attribution.
> 
> Good luck + best wishes on reserve duty. ❤️

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
JoshuaKGoldberg: 👍 implemented in #10551.

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
JoshuaKGoldberg: 👍 added to #10551

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
JoshuaKGoldberg: Done in #10551

packages/eslint-plugin/docs/rules/no-misused-spread.mdx
JoshuaKGoldberg: Fixed in #10551 👍

packages/eslint-plugin/src/rules/no-misused-spread.ts
JoshuaKGoldberg: Filed https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2521.

packages/eslint-plugin/tests/rules/no-misused-spread.test.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Fixed in https://github.com/typescript-eslint/typescript-eslint/pull/10551/commits/39daadfb22ec75ddc69995b074925b287a393f98

Issue #8373, opened 2024-02-04
Repo: CI flake on lint task in NX: "Unexpected error."


JoshuaKGoldberg:
### Suggestion

Seeing this in a few PRs, including #8371, #8357, and #8354. Each has an error linting a file starting with `Occurred while linting /home/runner/work/typescript-eslint/typescript-eslint/packages/eslint-plugin/tests/rules/`.

* 8354: `no-throw-literal.test.ts:16`
* 8357: `no-require-imports.test.ts:12`
* 8371: `no-unsafe-member-access.test.ts:15`

These all work locally for me. Maybe it's an out-of-memory issue in CI or something along those lines? cc @JamesHenry

JamesHenry:
@JoshuaKGoldberg do we think things have improved in this area now? I do remember us having some issues with our cache inputs configuration in the repo that we addressed. Could this one be closed or is there more to dig into do you think? (Totally fine if so)

auvred:
I encountered this error again a few days ago while working on https://github.com/typescript-eslint/typescript-eslint/pull/9455

Logs:

https://github.com/typescript-eslint/typescript-eslint/actions/runs/9731567772/job/26856205197#step:5:497

https://cloud.nx.app/runs/hgAzIgJIn5/task/eslint-plugin-internal%3Alint

JamesHenry:
@auvred I'm sorry I was too late looking at this, as per my message on discord please ping me on there to ensure I see the messages my github/github email notifications are too full of noise

JoshuaKGoldberg:
It's been 13 months and I haven't seen this recently. Hooray!

PR #8539, opened 2024-02-23
chore: tmp skip cache for eslint-plugin lint

JamesHenry:
- Fixes an existing issue with stale rule configs
- Updates duplicate CI job names
- Disables the cache for linting eslint-plugin to see if the prettier sync flakiness goes away


JoshuaKGoldberg:
For tracking: #8373

JoshuaKGoldberg (approved):
💚

Issue #8382, opened 2024-02-05
Docs: render snapshots of all rule doc example eslint output


bradzacher:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

We've had a number of docs issues where the errors reported on an example block were not correct. The big reason this occurs is a few reasons:

1) we don't validate the examples at all
2) it's really hard for a reviewer to visualise the errors (open pr, click netlify preview link, navigate to docs, pick example, click open in playground, preview errors - PHEW that's a lot of steps)
3) we have no way to visualise changes to the errors over time - so changes to rule behaviour that may remove or add errors aren't tracked.

For each rule doc code block we should render a snapshot output for each of its code samples which is the result of running the rule+config on the snapshot.
We could even make this a test which enforces that "valid" examples have no errors.

This would go hand-in-hand with #8381 to allow us to pre-visualise the shiki output and would be a natural partner to #6994.

### Affected URL(s)

docs/rules/*

PR #8558, opened 2024-02-27
fix(eslint-plugin): [no-unnecessary-type-assertion] fix false negative for const variable declarations

### LINKED ISSUE #4485: [no-unnecessary-type-assertion] only checks non-null-assertion

**Repro**

```JSON
{
	"root": true,
	"parser": "@typescript-eslint/parser",
	"parserOptions": {
		"ecmaVersion": 2021,
		"project": "./tsconfig.json",
		"sourceType": "module"
	},
	"plugins": [
		"@typescript-eslint"
	],
	"rules": {
		"@typescript-eslint/no-unnecessary-type-assertion": "error"
	}
}
```

```TS
// code.ts
type Foo = 3

const foox: Foo = 3 as 3
const fooy = 3 as Foo
const fooz: Foo = 3

const fooa = <3>3
const foob = <Foo>3

const fooc = 3
const bar = fooc!

function foo (x: number): number {
	return x! // unnecessary non-null
}
```

```JSON
{
	"compilerOptions": { "strict": true },
	"include": [ "./code.ts" ],
}
```

**Expected Result**

```sh
3:14   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
4:14   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
7:14   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
8:14   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
11:13  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
14:9   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion

✖ 6 problems (6 errors, 0 warnings)
```
OR (see below)
```sh
3:14   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
4:14   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
5:7    error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
7:14   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
8:14   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
11:13  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
14:9   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion

✖ 7 problems (7 errors, 0 warnings)
```

**Actual Result**

```sh
  11:13  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
  14:9   error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion

✖ 2 problems (2 errors, 0 warnings)
```

**Additional Info**

The `typesToIgnore` option description implies also checking annotations. Since annotations and assertions are different things, I would expect them to be handled by separate rules. Depending on whether or not they are an intended target for this rule, the first or second expectation should be fulfilled.

**Versions**

| `@typescript-eslint/eslint-plugin` | `5.10.0` |
| `@typescript-eslint/parser`        | `5.10.0` |
| `TypeScript`                       | `4.5.5` |
| `ESLint`                           | `8.7.0` |
| `node`                             | `14.17.5`, `16.13.2` |

### LINKED ISSUE #8458: Docs: [no-unnecessary-type-assertion] incorrect examples don't report errors

### Suggested Changes

[Some incorrect examples](https://typescript-eslint.io/rules/no-unnecessary-type-assertion/#examples) for `no-unnecessary-type-assertion` don't report expected errors.

1. [reports an expected error](https://typescript-eslint.io/play#eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&code=MYewdgzgLgBAZiEMC8MDMBuAUKSsBGAhgE4ryICEGQA)
2. [doesn't report errors!](https://typescript-eslint.io/play#eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&code=MYewdgzgLgBAZiEMC8MA8BmAfBg3EA)
3. [doesn't report errors!](https://typescript-eslint.io/play#eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&code=C4TwDgpgBAYg9nKBeKBmA3AKAMZwHYDOwUAZgslADzxwB8GQA)
4. [doesn't report errors!](https://typescript-eslint.io/play#eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&code=C4TwDgpgBAYg9nKBeKBmA3AKAMZwHYDOwUAZgsmlAIYGwLpA)
5. [reports an expected error](https://typescript-eslint.io/play#eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&code=GYVwdgxgLglg9mABMOcAUAPAXIsIC2ARgKYBOAlDnkWYgN4BQiipxUIpSGAhANyIB6AYnBhiEYgGdJAQ1IBPXAgC0eADZqGAXyA)

Should they report errors as expected?

### Affected URL(s)

https://typescript-eslint.io/rules/no-unnecessary-type-assertion/#examples

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4485 and fixes #8458

## Overview
The `no-unnecessary-type-assertions` rule needs a bit more detail about when type assertions are truly useful or useless on literals.


packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
```diff
@@ -25,10 +25,21 @@ if (
...
+    "[1, 2, 3, 4, 5].map(x => [x, 'A' + x] as [number, string]);",
+    `
+      let x: Array<[number, string]> = [1, 2, 3, 4, 5].map(
+        x => [x, 'A' + x] as [number, string],
+      );
+    `,
     'const foo = 3 as number;',
     'const foo = <number>3;',
-    'const foo = <3>3;',
-    'const foo = 3 as 3;',
```
auvred, line 31: According to the original issue, only rule docs are invalid. If you believe that the current behavior of the rule is invalid, then you should open a separate issue so we can discuss this.

abrahamguo: Sorry about that — I didn't want to create noise by opening a second issue as I thought that issue was already raised in the first issue. However, I have done so ✅

auvred: > I didn't want to create noise by opening a second issue

Don't worry! It's always better to open a separate issue!

bradzacher (changes requested):
See the end of this comment  https://github.com/typescript-eslint/typescript-eslint/issues/4485#issuecomment-1027182184 for details of the changes we would accept to the rule.

> Furthermore, the TSLint PR above also introduced some special handling for tuple types, which appears to be unnecessary, so it has been removed. 

[playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZiEMC8MDaBGAugbgFAD0BMJJAegPx56iSwBGAhgE4rrYyMTu6HGkxKeIA) is this not the case that this assertion is required?

Or are you saying that the logic isn't required because it's never hit?

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -242,20 +211,16 @@ export default createRule<Options, MessageIds>({
...
+          // It's not safe to remove a cast to a literal type, unless we are in a `const` variable declaration, as that
+          // type would otherwise be widened without the cast.
+          ((grandparent.type === AST_NODE_TYPES.VariableDeclaration &&
+            grandparent.kind === 'const') ||
+            !castType.isLiteral()) &&
+          services.getTypeAtLocation(node.expression) === castType
         ) {
-          // It's not always safe to remove a cast to a literal type or tuple
-          // type, as those types are sometimes widened without the cast.
-          return;
```
bradzacher, line 253: [style] IMO having two ifs is much clearer than one.
Personally I find that it can be much harder to read a bunch of mixed logic that's also of mixed negation.

abrahamguo:
> is this not the case that this assertion is required?
> Or are you saying that the logic isn't required because it's never hit?
@bradzacher this assertion is indeed required; however, no special handling is required for it (or, for any tuple types, I believe) in this lint rule because `services.getTypeAtLocation(node.expression) === castType` returns `false`.

abrahamguo:
Alrighty — I still need to clean up the tests a little bit, but I've reviewed (and added tests for):

- the examples (some of which are currently broken) from the [current docs](https://typescript-eslint.io/rules/no-unnecessary-type-assertion/), discussed further in [#8458](https://github.com/typescript-eslint/typescript-eslint/issues/8458)
- the examples from [the original TSLint PR](https://github.com/palantir/tslint/pull/3120)
- and all examples both invalid and valid, from the open issue #4485 

and I am confident that my logic is correct for detecting useful vs useless literal type assertions:

- First, as discussed above, any special logic for tuple type assertions is unnecessary. The base condition of `castType === uncastType` always returns false for any non-primitive type (in other words, this rule seems to do nothing at all for non-primitive types. Something to investigate for the future?)
- The new logic is in the `wouldSameTypeBeInferred` variable. There are two cases to consider, depending on whether `castType.isLiteral()` returns `true` or `false`.
- Note that `castType.isLiteral()` only returns `true` if the type assertion is a _single_ literal. If it is a union of literals, for example, then `castType.isLiteral()` will return `false`.
- if `castType.isLiteral()` returns `true`, then we know that the same type will be inferred, and therefore we are OK to report, if and only if we are within a `const` variable declaration. This _includes_ reporting on an `as const` of a primitive within a `const` variable declaration.
- On the other hand, if `castType.isLiteral()` returns false, then we must be looking at a more complex type, and we _should_ report (assuming that `typeIsUnchanged` is also `true`) _unless_ this is a `const` assertion, in which case, do not report.

Things to investigate for the future:
- Make this rule work for contextual typing (already noted as a TODO)
- Make this rule work for non-primitives (did it ever in the past? unsure)

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
```diff
@@ -16,6 +16,27 @@ const ruleTester = new RuleTester({
...
   valid: [
     `
+      let x = 1; // typeof x = number
+      let y = 1 as 1; // typeof y = 1
+      const x = 1; // typeof x = 1
+      let y = x; // typeof y = number
+      let z2 = x as 1; // okay - typeof z2 = 1
+      const foo = 3 as number;
+      const foon = <number>3;
+      const foom = 3 as number;
```
auvred, line 26: [Testing] We generally prefer test cases to be small and self-contained - rather than having multiple errors / types of parameter properties in one case (https://typescript-eslint.io/contributing/local-development#granular-unit-tests)

Could you please split them up?

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
```diff
@@ -245,6 +271,98 @@ const item = <object>arr[0];
...
+      errors: [
+        { messageId: 'unnecessaryAssertion', line: 3 },
+        { messageId: 'unnecessaryAssertion', line: 4 },
+        { messageId: 'unnecessaryAssertion', line: 6 },
+        { messageId: 'unnecessaryAssertion', line: 7 },
+        { messageId: 'unnecessaryAssertion', line: 9 },
+        { messageId: 'unnecessaryAssertion', line: 11 },
+        { messageId: 'unnecessaryAssertion', line: 16 },
+      ],
+    },
```
auvred, line 333: [Testing] Same as above ^

abrahamguo: Makes sense! I removed a fair number of tests after confirming that they were already accounted for in the existing tests.

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
auvred, line 1: [Docs] +1 on https://github.com/typescript-eslint/typescript-eslint/issues/8382

I didn't noticed this before:

The `Correct` section of rule docs now contains incorrect example!

https://github.com/typescript-eslint/typescript-eslint/blob/3ac824bf3f7a87ce1c8cbf3316407213b6a8969b/packages/eslint-plugin/docs/rules/no-unnecessary-type-assertion.md?plain=1#L54-L56

([playground](https://deploy-preview-8558--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=MYewdgzgLgBAZiEMC8MDkCRpgQwjUSKAbiA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&tsconfig=&tokens=false))

I think we should update it to match the new behavior. What do you think about something like

```ts
let foo = 'foo' as const;
```

 _(I can't comment on `docs/rules/no-unnecessary-type-assertion.md` because of GitHub)_

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
```diff
@@ -25,10 +25,28 @@ if (
   const name = member.id as TSESTree.StringLiteral;
 }
     `,
+    `
+      const cx = 1;
+      let z0 = cx as number;
+      let z1 = cx as const;
+      let z2 = cx as 1;
+    `,
```
auvred, line 33:
```suggestion
    `
      const c = 1;
      let z = c as number;
    `,
    `
      const c = 1;
      let z = c as const;
    `,
    `
      const c = 1;
      let z = c as 1;
    `,
```

[Testing] _(nit pick)_ - https://github.com/typescript-eslint/typescript-eslint/pull/8558#discussion_r1510849264

Looks like this test case contains assertions for both literal and const assertions, what do you think about splitting them so that the assertions are more granular?

auvred (changes requested):
Getting closer! :raised_hands: 

I think the main blocker in my mind is updating correct example in the docs

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
auvred: Also this PR resolves this issue - https://github.com/typescript-eslint/typescript-eslint/issues/8458

Let's update the PR description to link it?

abrahamguo: Ooh, really good catch, thanks!

auvred (dismissed):
LGTM :fire:

Thanks for working on this!

guy applauds

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -137,6 +106,13 @@ export default createRule<Options, MessageIds>({
       );
     }
 
+    function isConstVariableDeclaration(node: TSESTree.Node): boolean {
+      return (
+        node.type === AST_NODE_TYPES.VariableDeclaration &&
+        node.kind === 'const'
+      );
+    }
+
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Yeah I think we can leave these for now. If someone files an issue showing them being buggy in isolation, then that'd indicate there's enough user appetite to start thinking about whether the complexity is worth it.

JoshuaKGoldberg (dismissed):
🙌 swell, thanks!

PR #8559, opened 2024-02-27
fix(eslint-plugin): [prefer-optional-chain] only look at left operand for `requireNullish`

### LINKED ISSUE #8382: Docs: render snapshots of all rule doc example eslint output

### LINKED ISSUE #8487: Bug: [prefer-optional-chain] doesn't report at all when `requireNullish` is `true`

### Playground Link

https://typescript-eslint.io/play/#ts=5.3.3&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gMyUgXPJjAJYoDm8APvCgK4QQDcAULkvAGSc54B0GJAGUMpCgAoAlEyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5ujABtcNhy9o-aJAA0ylVhXYYiAI6xKnAHIJGyKfPxxEu7AF9nAXWVuXQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
declare const foo: string | null;
foo && foo.toString();
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/prefer-optional-chain": [
      "error",
      {
        "requireNullish": true
      }
    ]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I believe that this case should be reported. (see _Additional Info_ below)

### Actual Result

This case isn't reported.

### Additional Info

I noticed this while working on https://github.com/typescript-eslint/typescript-eslint/issues/8382: there is an example of incorrect code with `requireNullish: true` - https://typescript-eslint.io/rules/prefer-optional-chain#requirenullish. But it doesn't actually contain any lint errors.

As I understand it, `requireNullish` just restricts the set of types of the left operand of a logical expression to types that are nullish. If we disable `requireNullish`, then the rule reports on the code provided above ([playground](https://typescript-eslint.io/play/#ts=5.3.3&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gMyUgXPJjAJYoDm8APvCgK4QQDcAULkvAGSc54B0GJAGUMpCgAoAlEyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5ujABtcNhy9o-aJAA0ylVhXYYiAI6xKnAHIJGyKfK7jkiXdgC%2BLgLrL3roA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)).

Therefore, I believe that this case should be reported.

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8487 

## Overview
As mentioned in the linked issue, `requireNullish: true` wrongly disables all checking for this rule. This is because it mistakenly applies the `requireNullish` logic to both the left and right operands, rather than only the left.


packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
```diff
@@ -94,7 +95,7 @@ function isValidFalseBooleanCheckType(
     }
   }
 
-  if (options.requireNullish === true) {
+  if (options.requireNullish === true && operatorNode.left === node) {
```
auvred, line 98: This won't work for logical expressions with >2 operators, like for example `foo && foo.bar && foo.bar.toString()`. I haven't dived into the code yet, but I don't think we can rely on `operatorNode.left === node`.

[playground for this pr](https://deploy-preview-8559--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gMyUgXPAN4BGshmMAligOYC%2BAUI7kvAGTs54B0ZMHLqz6weGJAGUM1OgAoAlEA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5ujABtcNhy9o-aJAA0ylVhXYYiAI6xKnAHIJGyKfPxxEu7AF9nAXWVuXQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYwkJemy0eAcgK6qoDCAFutANZlKdFugC%2BIUUA&tokens=false)

```ts
declare const foo: {bar: string}
foo && foo.bar && foo.bar.toString()
```

This code should not be reported when `requireNullish` option is enabled, because there are no nullish types here at all. But it's actually reported.

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1866,6 +1866,16 @@ describe('hand-crafted cases', () => {
...
+      // requireNullish
+      {
+        code: `
+          declare const thing1: string | null;
+          thing1 && thing1.toString();
+        `,
+        options: [{ requireNullish: false }],
+        errors: [{ messageId: `preferOptionalChain` }],
+      },
+
```
auvred, line 1878: [Testing] I think it would be nice to add both valid and invalid test cases.

It would also be nice to have a richer set of cases for `requireNullish: true` and `requireNullish: false`.

- different operators (`||`, `&&`, etc..)
- count of operands (2, 3, etc...)
- types of operands: nullish/non-nullish, objects with nested properties, maybe even functions...

auvred (changes requested):
I haven't looked deep into the source code of the rule yet, so I can't suggest any complete solutions. Requesting changes to prevent one potential bug and to improve the tests.

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
abrahamguo: good catch, fixed!

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
abrahamguo: @auvred do you think that in order to provide a thorough test suite, we should run all ~700 existing tests with `requireNullish` set to both `true` and `false`?
If so, do you have a preferred way of duplicating the existing tests? I wasn't sure if it would be a good idea to simply duplicate the contents of the whole test file.

auvred: > we should run all ~700 existing tests with `requireNullish` set to both `true` and `false`?

Ah, I don't think so :smile:!

The default value for `requireNullish` is `false`, so it turns out that it's pretty well covered, especially in the `packages/eslint-plugin/tests/rules/prefer-optional-chain/base-cases.ts`.

So we can focus on `requireNullish: true` mostly. Here's a list of cases that might be a good starting point I think:

- ```ts
  declare const foo: string | null
  foo && foo.toString()
  ```

- ```ts
  declare const foo: string
  foo && foo.toString()
  ```

- ```ts
  declare const foo: string | null
  foo && foo.toString() && foo.toString()
  ```

- ```ts
  declare const foo: string
  foo && foo.toString() && foo.toString()
  ```

- ```ts
  declare const foo: { bar: string | null | undefined } | null | undefined
  foo && foo.bar && foo.bar.toString()
  ```

- ```ts
  declare const foo: { bar: string }
  foo && foo.bar && foo.bar.toString()
  ```

- ```ts
  declare const foo: { bar: string | null | undefined } | null | undefined
  foo && foo.bar && foo.bar.toString() && foo.bar.toString()
  ```

- ```ts
  declare const foo: { bar: string }
  foo && foo.bar && foo.bar.toString() && foo.bar.toString()
  ```

... and so on! This rule is quite complex in the sense that it is expected to work on any arbitrary chain of logical expressions. So we need to make sure we cover cases where the correct/incorrect part of the logical chain is at the beginning/middle/end of the expression.

But as per https://github.com/typescript-eslint/typescript-eslint/issues/6887, I don't think we can use the automatic test generation approach...

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
```diff
@@ -95,7 +96,10 @@ function isValidFalseBooleanCheckType(
   }
 
   if (options.requireNullish === true) {
-    return types.some(t => isTypeFlagSet(t, NULLISH_FLAGS));
+    return (
+      operatorNode.right === node ||
```
auvred, line 100: Ah, I don't think this fully fix the issue mentioned in the https://github.com/typescript-eslint/typescript-eslint/pull/8559#discussion_r1504104043. We can't rely on `operatorNode.right === node` either.

The problem is that the chain of expressions that may be replaced by a single optional chain expression can be anywhere in a logical expression. 

Unfortunately, for some reason the Netlify preview deployment got stuck, so I can't verify this on the online playground, but I'm pretty sure that

```ts
declare const thing1: string | null;
thing1 && thing1.toString() && true
```

won't be reported, because the `thing1.toString()` is not the `right` of the `operatorNode`.

auvred (changes requested):

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
JoshuaKGoldberg: Yeah +1, there are >=2 separate things in discussion:

* Handling for different variations of logical expressions & other AST shapes
* Handling for different types

As long as both have sufficient coverage, it's fine that every cross-combination isn't checked. That'd be a lot of redundant checks!

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/checkNullishAndReport.ts
```diff
@@ -0,0 +1,38 @@
...
+  maybeNullishNodes: TSESTree.Expression[],
+  descriptor: ReportDescriptor<PreferOptionalChainMessageIds>,
+): void {
+  if (
+    !requireNullish ||
+    maybeNullishNodes.some(node =>
+      unionTypeParts(parserServices.getTypeAtLocation(node)).some(t =>
+        isTypeFlagSet(t, ts.TypeFlags.Null | ts.TypeFlags.Undefined),
+      ),
+    )
```
auvred, line 34: [Bug] I think this is the wrong behavior here:

```ts
declare const foo1: { bar: string | null };
foo1 && foo1.bar; // reported - not ok

declare const foo2: { bar: string };
foo2 && foo2.bar; // not reported - ok
```

[playground](https://deploy-preview-8559--typescript-eslint.netlify.app/play/#ts=5.4.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gMyUgjAFzwDe8ARrMZjAJYoDm8APvCgK4QTwC%2BA3AChcBeADJROPPgB0lGH3gB6RfDgAHJDAwhg8ALRskWJAGsBA0JFgJk6LMIBMxMnOoY6jXoMdiJj2bAKyoZY6praugamQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5ujABtcNhy9o-aJAA0ylVhXYYiAI6xKnAHIJGyKfPxxEu7AF9nAXWVuXQA&tsconfig=&tokens=false)

The rule shouldn't report in both cases, because the left part (`foo1 &&` and `foo2 &&`) is not nullish. The only difference is the type of the `bar` (nullish/non-nullish).

I think that the type of the last element of the chain should not be taken into account in this check.

auvred (changes requested):

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/checkNullishAndReport.ts
abrahamguo: Good catch!

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -141,9 +142,9 @@ export default createRule<
 
           return leftPrecedence < OperatorPrecedence.LeftHandSide;
         }
-        context.report({
-          node: parentNode,
+        checkNullishAndReport(context, parserServices, options, [leftNode], {
           messageId: 'preferOptionalChain',
+          node: parentNode,
```
auvred, line 147: [Testing] Nice to see that `requireNullish` is now supported for `(foo ?? {}).bar`-like code! Could you add few tests to assert this?

Something like...

- ```ts
  declare const foo: string // and string | null
  (foo || {}).toString()
  ```

- ```ts
  declare const foo: string | null
  (foo || 'a' || {}).toString()
  ```


- ```ts
  declare const foo: string // and string | null
  (foo || undefined || {}).toString()
  ```

auvred (changes requested):
The overall code changes look great to me! Although, I haven't gotten deep enough into the rule code to 100% understand the logic yet. I'll try to take the time to figure out how it works internally, and then I can do a more thorough review :slightly_smiling_face:

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
auvred, line 1: https://github.com/typescript-eslint/typescript-eslint/pull/8497 was merged before this PR, so we should remove `skipValidation` from this code example in rule docs:

https://github.com/typescript-eslint/typescript-eslint/blob/fde44275117b3a7bbdcaaf267c73986b6637c5f9/packages/eslint-plugin/docs/rules/prefer-optional-chain.mdx?plain=1#L272-L277

auvred (changes requested):
Mostly LGTM! 

But there is conflict with `main` branch in `packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts` file. Could you please resolve it?

auvred (approved):
LGTM! Thanks for working on this :rocket:

JoshuaKGoldberg (approved):
LGTM!

Since @bradzacher touched a ton of this stuff in #6397, will wait for a bit in case Brad has time to review.

PR #8564, opened 2024-02-27
Export types of ts rule option.

linlinyang:
Support eslint flat config

Example:
```typescript
import type { Linter } from 'eslint';
import type {ruleOptions} from '@typescript-eslint/eslint-plugin';

// And than define rules:
type TSRules = {
  [K in keyof ruleOptions]?: Linter.RuleEntry<ruleOptions[K]>;
};
const rules: TSRules = {
  'accessor-pairs': 'error',
  'array-type': ['error', {
     readonly: 'array'
   }]
};
```


bradzacher (changes requested):
as per our contributing guide we ask that you raise an issue to discuss the change before raising a PR.

You have also provided no real context in our PR description.

Please file an issue to describe your intended usecase.

PR #8567, opened 2024-02-27
fix(typescript-eslint): apply `ignores` to all extended configs passed to `config` helper function

### LINKED ISSUE #8511: Bug: config function from typescript-eslint doesn't use `ignores` when `extends` is provided.

### Relevant Package

typescript-eslint

### Playground Link

_No response_

### Repro Code

I want to apply type-checked rules to most of my codebase with some exceptions. This is the config I have:

### ESLint Config

```TypeScript
tseslint.config({
    files: ['**/*.{ts,tsx}'],
    ignores: ['libs/legacy-lib/**'],
    extends: [
      ...tseslint.configs.strictTypeChecked,
      ...tseslint.configs.stylisticTypeChecked,
    ],
  })
```

### tsconfig

_No response_

### Expected Result

I expect the final config to have `ignores` applied to each of the configs I'm extending from.

### Actual Result

Only `files` are applied, not `ignores`.

### Additional Info

I can see from [the code](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-eslint/src/config-helper.ts#L94-L100) that `ignores` is not supported. This is somewhat surprising, given that `files` and `ignores` are two sides of the same coin - the ability to provide a list of files which to apply the rules to.

However, adding this functionality will likely be considered a breaking change.

### Versions

| `typescript-eslint` | `7.0.1` |

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8511

### DIFF
--- a/packages/typescript-eslint/src/config-helper.ts
+++ b/packages/typescript-eslint/src/config-helper.ts
@@ -91,14 +91,17 @@ export function config(
       return config;
     }
 
-    if (config.files) {
-      const files = config.files;
-      return [
-        ...extendsArr.map(conf => ({ ...conf, files: [...files] })),
-        config,
-      ];
-    }
+    const extension = {
+      ...(config.files && { files: config.files }),
+      ...(config.ignores && { ignores: config.ignores }),
+    };
 
-    return [...extendsArr, config];
+    return [
+      ...extendsArr.map(conf => ({
+        ...conf,
+        ...extension,
+      })),
+      config,
+    ];
   });
 }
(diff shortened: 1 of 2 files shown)


packages/typescript-eslint/src/config-helper.ts
```diff
@@ -91,14 +91,17 @@ export function config(
...
-    if (config.files) {
-      const files = config.files;
-      return [
-        ...extendsArr.map(conf => ({ ...conf, files: [...files] })),
-        config,
-      ];
-    }
+    const extension = {
+      ...(config.files && { files: config.files }),
+      ...(config.ignores && { ignores: config.ignores }),
```
auvred, line 96: `ignores` will overwrite `ignores` from the extended config. I'm not sure if this is the desired behavior, but so far I've made it to match the behavior of `files`. Any thoughts on this?

bradzacher: I think overriding is correct.
Any local config should always override the extended config

bradzacher (approved):

PR #8617, opened 2024-03-07
feat(parser): always enable comment, loc, range, tokens

### LINKED ISSUE #8347: Enhancement: Differentiate AST node types without parent (for estree) and with parent (for eslint)

### Relevant Package

parser

### Description

Spinning out of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68232: right now, the types for ASTs generated by `@typescript-eslint/parser` have "one size fits all" types. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68232#issuecomment-1902807538:

> https://github.com/typescript-eslint has options (just like espree, et al) which let you change the shape of the AST. By default those options turn off things like the ranges.
>
> However eslint does pass in the options to turn on the things it needs. So our AST then matches their shape.
> https://github.com/eslint/eslint/blob/60b966b6861da11617ddc15487bd7a51c584c596/lib/linter/linter.js#L828-L841
>
> We could define this with conditional types - but then it still wouldn't match the types you've declared.

One tricky thing is that for the sake of simplicity I don't think we want to make custom lint rule users have to see type parameters & optionality pop up all over the place in their AST. I think we'll want to have _two_ ASTs that can be generated:

* [`parse`](https://github.com/typescript-eslint/typescript-eslint/blob/3cdf5c93163036fdc60325129d9746ff57574336/packages/parser/src/parser.ts#L80): Utilizes the fancy new conditional types to make the AST shapes more precise
* [`parseForESLint`](https://github.com/typescript-eslint/typescript-eslint/blob/3cdf5c93163036fdc60325129d9746ff57574336/packages/parser/src/parser.ts#L87): Avoids the fancy new conditional types and uses the same existing non-generic versions of AST nodes

I think we can get there with the following changes:

* Give [`NodeOrTokenData`](https://github.com/typescript-eslint/typescript-eslint/blob/dd140643b457aa515cc21fcda2b3cd4acc2a1c5c/packages/ast-spec/src/base/NodeOrTokenData.ts) an type parameter object for whether it has `loc` and/or `range`
* For the AST types that feed into [`types/src/generated/ast-spec.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/types/src/generated/ast-spec.ts), create a second version of the nodes with that same type parameter object - i.e. `ParsedProgram<* extends *> extends NodeOrTokenData<*>`
* Make `parse` configurable with that new type parameter object
* Have `parse`'s return type use a new `ParseResult` interface that forwards that type parameter object to its AST nodes
* Make no changes to [`ParseForESLintResult`](https://github.com/typescript-eslint/typescript-eslint/blob/3cdf5c93163036fdc60325129d9746ff57574336/packages/parser/src/parser.ts#L80) or other rule infrastructure: it returns the same nodes as before

I.e. we'd be making a new set of node types that users of `parse` now see, without changing `parseForESLint` or custom rules.

### Additional Info

_No response_

JoshuaKGoldberg:
BREAKING CHANGE: Changes opt-in options from `@typescript-eslint/parser` to always-on.

## PR Checklist

- [x] Addresses an existing open issue: fixes #8347

## Overview

Hardcodes those values to `true` in `parser`, and sets up the types so that generated ASTs always have the `comment`, `range`, and `tokens`.

Note that this doesn't change the values in `typescript-estree`. That package still has the generic `AST` type with conditional types on `'comment'` and `'token'`.

Adds some more unit tests while I'm in the area.

Co-authored-by: @bradzacher 

💖


packages/types/src/parser-options.ts
```diff
@@ -62,13 +61,10 @@ interface ParserOptions {
   extraFileExtensions?: string[];
   filePath?: string;
   jsDocParsingMode?: JSDocParsingMode;
-  loc?: boolean;
   programs?: Program[] | null;
   project?: string[] | string | boolean | null;
-  projectFolderIgnoreList?: (RegExp | string)[];
-  range?: boolean;
+  projectFolderIgnoreList?: string[];
```
JoshuaKGoldberg, line 66: I haven't verified yet (https://github.com/typescript-eslint/typescript-eslint-examples/issues/9), but it looks to me like we never actually supported the `RegExp` case?

https://github.com/typescript-eslint/typescript-eslint/blob/d8164960d21336d4a726d8179b314f05fe3aeb22/packages/typescript-estree/src/parseSettings/resolveProjectList.ts#L55-L63

JoshuaKGoldberg, line 66: I haven't verified yet (https://github.com/typescript-eslint/typescript-eslint-examples/issues/9), but it looks to me like we never actually supported the `RegExp` case?

https://github.com/typescript-eslint/typescript-eslint/blob/d8164960d21336d4a726d8179b314f05fe3aeb22/packages/typescript-estree/src/parseSettings/resolveProjectList.ts#L57-L63

JoshuaKGoldberg, line 66: I haven't verified yet (https://github.com/typescript-eslint/typescript-eslint-examples/issues/9), but it looks to me like we never actually supported the `RegExp` case?

https://github.com/typescript-eslint/typescript-eslint/blob/d8164960d21336d4a726d8179b314f05fe3aeb22/packages/typescript-estree/src/parseSettings/resolveProjectList.ts#L57-L65

bradzacher (approved):

PR #8624, opened 2024-03-08
docs: getting started guide should reference allowJs

joehoppe:
## PR Checklist

## Overview

Quickstart.mdx recommends using a JavaScript eslint.config file. The documentation should reference that a tsconfig.json needs to allow JS files for this setup to work

### DIFF
--- a/docs/getting-started/Quickstart.mdx
+++ b/docs/getting-started/Quickstart.mdx
@@ -30,7 +30,7 @@ npm install --save-dev eslint typescript typescript-eslint
 
 ### Step 2: Configuration
 
-Next, Assuming that your tsconfig.json has "allowJs" set to true, create an `eslint.config.js` config file in the root of your project, and populate it with the following:
+Next, assuming that your tsconfig.json has "allowJs" set to true, create an `eslint.config.js` config file in the root of your project, and populate it with the following:
 
 ```js title="eslint.config.js"
 // @ts-check
--- a/docs/getting-started/Quickstart.mdx
+++ b/docs/getting-started/Quickstart.mdx
@@ -30,7 +30,7 @@ npm install --save-dev eslint typescript typescript-eslint
 
 ### Step 2: Configuration
 
-Next, create an `eslint.config.js` config file in the root of your project, and populate it with the following:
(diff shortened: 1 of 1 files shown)


docs/getting-started/Quickstart.mdx
```diff
@@ -30,7 +30,7 @@ npm install --save-dev eslint typescript typescript-eslint
 
 ### Step 2: Configuration
 
-Next, create an `eslint.config.js` config file in the root of your project, and populate it with the following:
+Next, assuming that your tsconfig.json has "allowJs" set to true, create an `eslint.config.js` config file in the root of your project, and populate it with the following:
```
auvred, line 33: https://www.typescriptlang.org/tsconfig#allowJs

> Allow JavaScript files to be imported inside your project, instead of just .ts and .tsx files.

But we don't recommend to import `eslint.config.js` into `.ts` files.  Why do we need to enable this option?

auvred (changes requested):

docs/getting-started/Quickstart.mdx
bradzacher: Yeah thus is only required if you want to either import the file or typecheck the file via `tsc`.

For the purposes of this guide we assume neither. The latter has much larger implications for your build which has many variations of fixes that we cannot comment about in the simple getting started guide.

joehoppe: I'm having a bunch of issues with the getting started guide

Thought that this was my resolution. But turns out I was blocked until I renamed eslint.config.js to have a .mjs extension

I will investigate separately

My mistake!

bradzacher: Yeah when we wrote the guide you could only do `.js` so you had to use ESM or CJS - whatever was defined in your package.json

We didn't think they were going to backport `.mjs` support until v9 released but they did it much sooner than we thought - so you can now use `.mjs` for your config file in v8.57.

Havent gotten around to updating the extension listed in the docs yet.

More than happy if you want to update `eslint.config.js` to `eslint.config.mjs` throughout the docs!

PR #8626, opened 2024-03-08
Update no-var-requires.md

AaronJY:
Add example of where to add `{ allow: ... }` option, as there doesn't seem to be documentation of that yet.


AaronJY:
Closed because there's too much effort involved in making a markdown change :(

Issue #8314, opened 2024-01-29
TypeScript 5.4 Support


bradzacher:
https://devblogs.microsoft.com/typescript/announcing-typescript-5-4-beta/

This issue is just to track all of the new features and their implementation state in this project.
As with all releases, we will not necessarily to support all features until closer to the full release when everything the features are stabilised.

_Please be patient._

**Seems like a pretty cruisy release for us thus far**

# 🔲 lib.d.ts Updates

We will need to regenerate our types within scope-manager.

# Other changes with no impact to us 

- Preserved Narrowing in Closures Following Last Assignments
- The NoInfer Utility Type
- Object.groupBy and Map.groupBy
    - Need to double check if this needs to be listed in `unbound-methods`
- Support for require() calls in --moduleResolution bundler and --module preserve
- Checked Import Attributes and Assertions
- Quick Fix for Adding Missing Parameters
- Upcoming Changes from TypeScript 5.0 Deprecations
- Breaking Changes lib.d.ts Changes
- More Accurate Conditional Type Constraints
- More Aggressive Reduction of Intersections Between Type Variables and Primitive Types
- Errors When Type-Only Imports Conflict with Local Values
- Emit Changes

bradzacher:
TS 5.4 has released.
No changes require our action aside from lib updates.
So we should be able to land this for the next release.

melroy89:
> TS 5.4 has released. No changes require our action aside from lib updates. So we should be able to land this for the next release.

Great! good to know. I updated already in my project, I will ignore this warning from eslint for now, assuming everywhere will be solved soon.

trivikr:
> assuming everywhere will be solved soon.

The change to update typescript version to include 5.4.x is in PR https://github.com/typescript-eslint/typescript-eslint/pull/8630
Once it's merged and released, the warning should go away.

PR #8630, opened 2024-03-08
feat: support TS 5.4

strmer15:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/8314

## Overview

This adds support for Typescript 5.4, which was recently released as generally available - https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/

The only changes I made were to the version used by the root `package.json` and the supported TS versions in `typescript-estree`.


vacas (approved):

bradzacher (commented):
As per the issue https://github.com/typescript-eslint/typescript-eslint/issues/8314 - we will need to regenerate the lib.d.ts types.

There is also more to do than just those numbers.
Feel free to action it - here is the guide:
https://typescript-eslint.io/maintenance/pull-requests/dependency-version-upgrades#adding-support-for-a-new-typescript-version

Start at step 4

strmer15:
> As per the issue #8314 - we will need to regenerate the lib.d.ts types.
> 
> There is also more to do than just those numbers. Feel free to action it - here is the guide: https://typescript-eslint.io/maintenance/pull-requests/dependency-version-upgrades#adding-support-for-a-new-typescript-version
> 
> Start at step 4

Thanks @bradzacher !

I looked at step 4, but it's only about updating an RC version of Typescript, which isn't the case here - this is the GA version, so I skipped to step 6.

I updated the types and then updated the doc as well, it was asking to change `devDependencies` when the typescript version is actually in `resolutions` now.

Let me know if there's anything else to do to get TS 5.4 supported, thanks!

bradzacher (approved):
LGTM - thanks for doing this!

PR #8549, opened 2024-02-24
docs: fill out maintenance docs on PR review flow

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8440
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Fills out PR review flow docs.

I'm sure I missed some important things 😄. What did I miss? 

Co-authored-by: @auvred

PR #8632, opened 2024-03-09
feat(eslint-plugin): [no-confusing-void-expression] add ignoreVoidInVoid option to void in void situation

### LINKED ISSUE #8538: Enhancement: [no-confusing-void-expression] Add option to ignore void<->void

https://typescript-eslint.io/rules/no-confusing-void-expression

### Description

Following from https://github.com/typescript-eslint/typescript-eslint/pull/8375#pullrequestreview-1897027584: we tried enabling `@typescript-eslint/no-confusing-void-expression` internally and very quickly found it to be inconvenient:

> Looking at the changes - TBH I don't really like this rule. I don't think any of these changes improve clarity or readability of code.
> 
> The biggest issue I see is that if the function is implicitly typed or explicitly marked as returning `void` the rule still requires you to change things.
> 
> For example this code should pass the rule, IMO:
> 
> ```ts
> function returnsVoid(): void {}
> 
> function test1(): void {
>   return returnsVoid(); // should be fine
> }
> const test2 = (): void => returnsVoid(); // should be fine
> ```
> 
> Because there's nothing confusing about that code - the expressions are all typed as `void` and the functions are typed as `void`.
> 
> Also changes like `() => expr` => `() => { expr }` actively reduce readability, IMO, esp when considering the above.

Agreed. @bradzacher and I chatted briefly, and Brad suggested:

1. Adding an option to ignore the `void` <-> `void` return case
2. Making that option the default in v8

Agreed! Filing this issue to add that option. We can file a followup if & when it gets in.

### Fail

```TypeScript
declare function returnsVoid(): void;

function outerReturnsVoid(): void {
  return returnsVoid();
}
```

### Pass

```TypeScript
declare function returnsString(): string;

function outerReturnsString(): string {
  return returnsString();
}
```

### Additional Info

_No response_

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8538

## Overview

add ignoreVoidInVoid option in no-confusing-void-expression to resolve void in void situation. below example is avaliable
```
function returnsVoid(): void {}

// type is correct with void, so not evoke error
function test1(): void {
  return returnsVoid();
}

// type is correct with void, so not evoke error
const test2 = (): void => returnsVoid();

// type is correct with void, so not evoke error
const test3 = (): void => {
  return returnsVoid();
};
```


packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +184,26 @@ export default createRule<Options, MessageId>({
...
+              invalidAncestor.parent.parent?.type ===
+                AST_NODE_TYPES.FunctionDeclaration &&
+              invalidAncestor.parent.parent.returnType?.typeAnnotation.type ===
+                AST_NODE_TYPES.TSVoidKeyword
+            ) {
+              return;
+            }
+
+            if (
+              invalidAncestor.parent.parent?.type ===
```
developer-bandi, line 198: this line is not corvered with same reason in upper comment

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +184,26 @@ export default createRule<Options, MessageId>({
         if (invalidAncestor.type === AST_NODE_TYPES.ReturnStatement) {
           // handle return statement
 
+          if (options.ignoreVoidInVoid) {
+            if (
+              invalidAncestor.parent.parent?.type ===
```
developer-bandi, line 189: this line's ? operator is not covered with test code. but i don't know the case that invalidAncestor is ReturnStatement and invalidAncestor.parent.parent is undefind

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
auvred:
```ts
declare function returnsVoid(): void
return returnsVoid()
```

[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.3.3&showAST=es&fileType=.ts&code=GYVwdgxgLglg9mABAJwKZRMsBnAanGAEwAoBKALkQDcDDEBvRAXwCg0MsV1Md8iygA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiIT2AL7eAupp%2BPkA&tsconfig=&tokens=false)

Technically, this is incorrect TypeScript code, but AFAIK JavaScript allows top-level returns, so the rule shouldn't crash in this case (see [this commit](https://github.com/typescript-eslint/typescript-eslint/pull/8549/commits/b0382cd82505a722b92e0bf7fbdc025d604c2c71#diff-f974c8ecb4f89a6ea10ccd583c49bc3017a3465595bcd23ffb67a131506a8f0cR86) from the upcoming https://github.com/typescript-eslint/typescript-eslint/pull/8549)

Let's add a test to prove that the rule doesn't crash in similar situations! What do you think?

packages/eslint-plugin/docs/rules/no-confusing-void-expression.md
```diff
@@ -109,6 +109,29 @@ function doSomething() {
...
 ```
 
+### `ignoreVoidInVoid`
+
+Allow using `void` type expressions in return value of a function that specified as `void`
+
+Examples of additional **correct** code with this option enabled:
+
+```ts option='{ "ignoreVoidInVoid": true }' showPlaygroundButton
+function returnsVoid(): void {}
```
auvred, line 119: [docs] It seems that other examples in the rules documentation use `window.postMessage(value)` as an example of a function that returns `void`. What do you think about using this function instead of `returnsVoid`?

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
auvred, line 189: [bug] We can't assume that `return` statement is always in the root-level block scope of the function. It may be inside blocks, if-else statements, switch statements, loops, etc.

So the following example is false positive (the rule shouldn't complain):

```ts
declare function returnsVoid(): void

function test(): void {
  {
    return returnsVoid();
  }
}
```

([playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXzg2RlQGcA1HLYACgEoAueANyuACh2V1s94MQpDPSatq8AN7t4k6TIIgiJBUrKVq9ANxyAvux1A&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiIT2AL7eAupp%2BPkA&tsconfig=&tokens=false))

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +184,26 @@ export default createRule<Options, MessageId>({
...
+                AST_NODE_TYPES.FunctionDeclaration &&
+              invalidAncestor.parent.parent.returnType?.typeAnnotation.type ===
+                AST_NODE_TYPES.TSVoidKeyword
+            ) {
+              return;
+            }
+
+            if (
+              invalidAncestor.parent.parent?.type ===
+                AST_NODE_TYPES.ArrowFunctionExpression &&
```
auvred, line 199: [bug] We should also check function expressions, like:

```ts
declare function returnsVoid(): void

const foo = function (): void {
  return returnsVoid();
}

class Foo {
  foo(): void {
    return returnsVoid();
  }
}

const data = {
  foo(): void {
    return returnsVoid();
  }
}
```

(all the cases above are false positive)

([playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXzg2RlQGcA1HLYACgEoAueANyuACh2w9SMkcc8ALxI0mXPnpNW1eAG928AiCIllqspWr0A3OwC%2BnSFFKl4AMQHzF-HFJZtrSpYWL5XJCm103Dhrjx8wFAYUMJOtvYywBEuKm7qbl7adHpKfkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiIT2AL7eAupp%2BPkA&tsconfig=&tokens=false))

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
```diff
@@ -119,6 +119,34 @@ function cool(input: string) {
...
     },
+    {
+      options: [{ ignoreVoidInVoid: true }],
+      code: `
+function returnsVoid(): void {}
+
+function test1(): void {
+  return returnsVoid(); // should be fine
+}
+      `,
```
auvred, line 130: [testing] The rest test cases use `console.log` to produce the `void` type, let's use `return console.log('foo')` to be consistent with other test cases? Also, using `console.log` would shorten all cases by two lines, so I think the tests would be more readable. WDYT?

auvred (changes requested):
A nice start, thank you for sending it in!

packages/eslint-plugin/docs/rules/no-confusing-void-expression.md
developer-bandi: that's make sence. i edit the docs using `window.postMessage(value)`

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: it seems like a case I hadn’t thought of. Thank you.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: it seems like a case I hadn’t thought of. Thank you.

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
developer-bandi: i think that's good idea!

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: The code has been changed to reflect the comments left below, so i'm no longer need to cover that branch.

Therefore, I think there is no need to add the above test case. What do you think?

developer-bandi:
There were no objections to the comments you left, so i has reflected all of them.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -112,6 +120,15 @@ export default createRule<Options, MessageId>({
...
 
+          if (options.ignoreVoidInVoid) {
+            if (
+              invalidAncestor.returnType?.typeAnnotation.type ===
+              AST_NODE_TYPES.TSVoidKeyword
+            ) {
+              return;
+            }
+          }
+
```
auvred, line 131:
```suggestion
          if (options.ignoreVoidInVoid && invalidAncestor.returnType?.typeAnnotation.type === AST_NODE_TYPES.TSVoidKeyword) {
            return;
          }

```

[Style] Totally non-blocking, I just noticed that we can inline these two conditions. Feel free to ignore this, if you find the current implementation more readable.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +184,18 @@ export default createRule<Options, MessageId>({
         if (invalidAncestor.type === AST_NODE_TYPES.ReturnStatement) {
           // handle return statement
 
+          if (options.ignoreVoidInVoid) {
+            if (
+              targetNodeAncestorHasVoidReturnType(invalidAncestor, [
+                AST_NODE_TYPES.FunctionDeclaration,
+                AST_NODE_TYPES.ArrowFunctionExpression,
+                AST_NODE_TYPES.FunctionExpression,
+              ])
```
auvred, line 193: [Refactor] I see no reason why we would want to use `targetNodeAncestorHasVoidReturnType` with some other set of `targets` nodes. So I think we can just put those node types inside the function. WDYT?

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +405,36 @@ export default createRule<Options, MessageId>({
...
+        }
+        return false;
+      });
+
+      if (filter.length > 0) {
+        return true;
+      }
+
+      return targetNodeAncestorHasVoidReturnType(node.parent, targets);
+    }
```
auvred, line 438: [Bug] This function won't stop until it finds a matching parent, so the following case is false negative

```ts
function foo(): void {
  const bar = () => {
    return console.log()
  }
}
```

([playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=GYVwdgxgLglg9mABMOcAUBKAXIgbnGAE0QG8AoRRCBAZykQCMBDAJ0QF5FMOA%2BUiyohYBTKCBZJqYGnAA2wgHSy4Ac0wCAvmS1A&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiIT2AL7eAupp%2BPkA&tsconfig=&tokens=false))

I think that this function should look +- like this:

- if `!node.parent` -> return false
- if `node.parent` is one of the `FunctionDeclaration`, `ArrowFunctionExpression`, `FunctionExpression`:
  - if `node.parent` has `void` in type annotation -> return true
  - if `node.parent` hasn't `void` in type annotation -> return false
- return `targetNodeAncestorHasVoidReturnType(node.parent)`

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: thank you i agree this style

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: I agreed that there was no possibility of reuse, so i moved the targets inside the function and changed the function name to be more clear.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: in playground, this case seems false positive. anyway, this is bug, so i add test code and fix `targetNodeAncestorHasVoidReturnType` function

developer-bandi: I'm not familiar with the testing, so I think I used the terminology the other way around. thank you for telling me!

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +183,14 @@ export default createRule<Options, MessageId>({
         if (invalidAncestor.type === AST_NODE_TYPES.ReturnStatement) {
           // handle return statement
 
+          if (options.ignoreVoidInVoid) {
+            if (
+              targetNodeFunctionAncestorNodeHasVoidReturnType(invalidAncestor)
+            ) {
+              return;
+            }
+          }
```
auvred, line 192: [Style] Same as here: https://github.com/typescript-eslint/typescript-eslint/pull/8632#discussion_r1518572544

```suggestion
          if (options.ignoreVoidInVoid && targetNodeFunctionAncestorNodeHasVoidReturnType(invalidAncestor)) {
            return;
          }
```

WDYT about inlining this if statement?

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +400,39 @@ export default createRule<Options, MessageId>({
...
+
+      if (!node.parent) {
+        return false;
+      }
+
+      if (
+        targets[0] === node.parent.type ||
+        targets[1] === node.parent.type ||
+        targets[2] === node.parent.type
+      ) {
```
auvred, line 425:
```suggestion
      if (
        AST_NODE_TYPES.FunctionDeclaration === node.parent.type ||
        AST_NODE_TYPES.ArrowFunctionExpression === node.parent.type ||
        AST_NODE_TYPES.FunctionExpression === node.parent.type
      ) {
```

[Refactor] Why do we need the `targets` tuple if we don't iterate over it? IMO if we directly compare `node.parent.type` to AST node types, it will be more readable. What do you think?

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +400,39 @@ export default createRule<Options, MessageId>({
...
+        if (
+          node.parent.returnType?.typeAnnotation.type ===
+          AST_NODE_TYPES.TSVoidKeyword
+        ) {
+          return true;
+        }
+        return false;
```
auvred, line 432: [Bug] I just noticed that we determine whether function signature has `void` return type or not by doing pure syntax check. 

Thus, the following case is false positive (the rule shouldn't report error, since function returns `Foo` - an alias of `void` type):

```ts
type Foo = void

function test(): Foo {
  return console.log()
}
```

[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.3.3&showAST=types&fileType=.ts&code=C4TwDgpgBAYg9nKBeKA3OBLAJgKBwMwFcA7AY2AzmKmAgGdgAKASgC5YEoBvHKKAJwjBC-aqSp04AGwgA6KXADmLHAF8cQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiIT2AL7eAupp%2BPkA&tsconfig=&tokens=false)

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
auvred, line 1: [Testing] Some more cases to add:

- the case mentioned [here](https://github.com/typescript-eslint/typescript-eslint/pull/8632#discussion_r1518567970) - `return console.log('foo')` on the top-level
- function without `void` in signature contains in its body function with `void` in signature, and this nested function returns `return console.log('foo')`
- Type aliases, like in the comment above ^ (in the signature of the parent function, in the signature of the returned function, in both signatures at the same time)

auvred (changes requested):
Progress :rocket:

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: Comparing variables directly seems much more readable!

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
developer-bandi: I added a test case, but I may have misunderstood the last comment, so please check my new test code. thank you!

auvred: I think it's fine.

Also, I think it would be nice to add the following cases:

- ```ts
  function test(): void & void {
    return console.log('foo')
  }
  ```

- ```ts
  type Foo = void
  declare function foo(): Foo
  function test(): Foo {
    return foo()
  }
  ```

- ```ts
  type Foo = void
  const test: Foo = () => console.log('err')
  ```

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -112,6 +120,14 @@ export default createRule<Options, MessageId>({
...
           // handle arrow function shorthand
 
+          if (
+            options.ignoreVoidInVoid &&
+            invalidAncestor.returnType?.typeAnnotation.type ===
+              AST_NODE_TYPES.TSVoidKeyword
+          ) {
+            return;
+          }
+
```
auvred, line 130: [Bug] Same as here https://github.com/typescript-eslint/typescript-eslint/pull/8632#discussion_r1519205444...

```ts
type Foo = void
const test = (): Foo => console.log()
```

The error shouldn't be reported, since arrow function has the `void` return type

[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.3.3&showAST=types&fileType=.ts&code=C4TwDgpgBAYg9nKBeKA3OBLAJgKAMZwB2AzsFMBKclABQCUAXLAsgHxQElwA2EAdNzgBzekA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiIT2AL7eAupp%2BPkA&tsconfig=&tokens=false)

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +399,33 @@ export default createRule<Options, MessageId>({
...
+        if (
+          node.parent.returnType &&
+          !tsutils.isTypeFlagSet(
+            getConstrainedTypeAtLocation(services, node.parent.returnType),
+            ts.TypeFlags.VoidLike,
+          )
+        ) {
+          return true;
+        }
+        return false;
```
auvred, line 425:
```suggestion
        return (
          node.parent.returnType &&
          !tsutils.isTypeFlagSet(
            getConstrainedTypeAtLocation(services, node.parent.returnType),
            ts.TypeFlags.VoidLike,
          )
        )
```

[Refactor] This code may be converted into a single return statement. What do you think?

auvred (changes requested):

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
developer-bandi: thank you! add all test case and last case has ignoreArrowShorthand option to pass, so add  ignoreArrowShorthand option

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: i think this error is because of  ignoreArrowShorthand. so  ignoreArrowShorthand option to true or change code like:
```
type Foo = void
const test = (): Foo => { return console.log(); }
```
I add test case with ignoreArrowShorthand option true.

[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.3.3&showAST=types&fileType=.ts&code=C4TwDgpgBAYg9nKBeKA3OBLAJgKAMZwB2AzsFMBKclABQCUAXLAsgHxQElwA2EAdNzgBzekA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiJj5i1ZuIAQV0OAHcAZQALfXwIgEMmKXdPE2wAXxSAXU101KA&tsconfig=&tokens=false)

developer-bandi: thank you i understand it! and since it seems to be valid even when declaring a type in a variable of a function expression, I modified the code and test case. 

```
type Foo = () => void;
const test: Foo = function () {
  return console.log('err');
};
```

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -112,6 +121,25 @@ export default createRule<Options, MessageId>({
         if (invalidAncestor.type === AST_NODE_TYPES.ArrowFunctionExpression) {
           // handle arrow function shorthand
 
+          // console.log(invalidAncestor.parent.type === AST_NODE_TYPES.VariableDeclaration)
+
```
auvred, line 125: [Cleanup] Do we really need this comment?

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -112,6 +121,25 @@ export default createRule<Options, MessageId>({
...
+              return;
+            }
+            if (
+              !invalidAncestor.returnType &&
+              invalidAncestor.parent.type ===
+                AST_NODE_TYPES.VariableDeclarator &&
+              isVoidLikeType(invalidAncestor.parent.id.typeAnnotation)
+            ) {
+              return;
+            }
```
auvred, line 140: [Bug] Why do we have it here? We shouldn't take into account this situation (the function is being assigned to a variable with `void` type).

If you added that after my comment  (https://github.com/typescript-eslint/typescript-eslint/pull/8632#discussion_r1519929980), I apologize, it was a typo. 

I was thinking of the following code:

```ts
type Foo = void
const test = (): Foo => console.log('err')
```

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +411,39 @@ export default createRule<Options, MessageId>({
...
+        AST_NODE_TYPES.FunctionDeclaration === node.parent.type ||
+        AST_NODE_TYPES.ArrowFunctionExpression === node.parent.type ||
+        AST_NODE_TYPES.FunctionExpression === node.parent.type
+      ) {
+        return (
+          (node.parent.returnType && isVoidLikeType(node.parent.returnType)) ||
+          (!node.parent.returnType &&
+            node.parent.parent.type === AST_NODE_TYPES.VariableDeclarator &&
+            isVoidLikeType(node.parent.parent.id.typeAnnotation))
+        );
```
auvred, line 443: [Bug] Same as in the comment above ^

auvred (changes requested):

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: thank you! delete the code and test code.

packages/eslint-plugin/docs/rules/no-confusing-void-expression.md
```diff
@@ -109,6 +109,27 @@ function doSomething() {
...
 ```
 
+### `ignoreVoidInVoid`
+
+Allow using `void` type expressions in return value of a function that specified as `void`
+
+Examples of additional **correct** code with this option enabled:
+
+```ts option='{ "ignoreVoidInVoid": true }' showPlaygroundButton
+// type is correct with void, so not evoke error
```
auvred, line 119: [Docs] I'm not sure I got the meaning of those comments right :smile:

"type is correct with void" - the point of this option is to ignore void <-> void cases, type correctness is checked by `tsc`, so I think that we should rephrase these comments or remove them. Furthermore, I think this code is self-explanatory, so I think it would be easy to guess what this option does just by looking at the description and the code. Please correct me if I'm wrong

auvred (changes requested):
The code changes and tests look good to me! Requesting a small change/discussion on code example in the rule docs

packages/eslint-plugin/docs/rules/no-confusing-void-expression.md
developer-bandi: The reason for including the comment was because the code in the previous example included the comment, but there is no other reason. I will remove the comment!

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +401,36 @@ export default createRule<Options, MessageId>({
...
+        return false;
+      }
+
+      if (
+        AST_NODE_TYPES.FunctionDeclaration === node.parent.type ||
+        AST_NODE_TYPES.ArrowFunctionExpression === node.parent.type ||
+        AST_NODE_TYPES.FunctionExpression === node.parent.type
+      ) {
+        return (
+          !!node.parent.returnType && isVoidLikeType(node.parent.returnType)
```
yeonjuan, line 429: Considering the following case, perhaps we should use signature to check the return type?

[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.3.3&showAST=types&fileType=.tsx&code=C4TwDgpgBAYg9nKBeKAKAlMgfFAbnASwBMBuAKAGM4A7AZ2CmAnoC5YFk1Mkcq64ANhAB0AuAHNUAcgBmCKehJQA9MqgQAThrga0AfnSUa9Rs2AAmThjb5i2KH1qCRYybPmKVa6ok3aNqHAA1uhAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANqRE0aB2iQANJnDYclYZ0UA1SVICSTa9NX44iYxAC%2BAXWMeQHoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)
```ts
type Foo = () => void;
const test: Foo = () => console.log('foo'); // error (?)
const test2 = (): void => console.log('foo'); // no error(ok)
```

developer-bandi: If the function type that returns void is specified, it seems good to receive the effect of the new option!

I edit the code to pass the four test cases below, including the code provided as an example.

```javascript
type Foo = () => void;

// case 1 (ArrowFunction + Signature)
const test: Foo = () => console.log('foo'); // error

// case 2 (ArrowFunction + noSignature)
const test2: Foo = function () { return console.log('foo') } // error

// case 3 (functionExpression + Signature)
const test3: () => void = () => console.log('foo'); // error 

// case 4 (functionExpression + noSignature)
const test4: () => void = function () { return console.log('foo') } // error 
```

yeonjuan: Hi @auvred, :) 
I thought [the review mentioned in the issue](https://github.com/typescript-eslint/typescript-eslint/pull/8375#pullrequestreview-1897027584) meant [these kind of cases](https://github.com/typescript-eslint/typescript-eslint/pull/8375/files#diff-0acfb0bea43769266683f8645d9804098eeb04529992352922eb00ff5b130334R334).
I didn't just mean typed arrow, I thought it was meaningful to check if the function's signature returns void. Thanks 👍 . I agree to wait for a discussion.

```ts
interface Rule {
  SomeNode(): void;
}

declare const context: {
  report(): void
}

const rule: Rule = {
  SomeNode() {
    return context.report();
  }
}
```

developer-bandi: @auvred As discussed in the issue, it seems like a good idea for me to start working on this. 

I plan to write code that satisfies the test cases written in this comment, but are there any test cases you would like to add?

```javascript
type Foo = () => void;

// case 1 (ArrowFunction + Signature)
const test: Foo = () => console.log('foo'); // error

// case 2 (ArrowFunction + noSignature)
const test2: Foo = function () { return console.log('foo') } // error

// case 3 (functionExpression + Signature)
const test3: () => void = () => console.log('foo'); // error 

// case 4 (functionExpression + noSignature)
const test4: () => void = function () { return console.log('foo') } // error

// case 5 (functionParameter is Function)
declare function foo(arg: () => void): void;
foo(() => console.log());

// case 6 (HigherOrderFunction)
type HigherOrderType = () => () => () => void;
const x: HigherOrderType = () => () => () => console.log();

// case 7 (interface with Function)
interface Foo {
  foo: () => void
}
function bar(): Foo {
  return {
    foo: () => console.log(),
  };
}

// case 8 (objectMethod)
const q = {
  foo: { bar: () => console.log() },
} as {
  foo: { bar: () => void }
}

```

auvred: I took those examples from `explicit-function-return-type`'s tests: https://github.com/typescript-eslint/typescript-eslint/blob/ecae04f3800fae255db102055e49eef51e31e006/packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts

You can search for `allowTypedFunctionExpressions` here to find some other cases

> but are there any test cases to add?

After a quick look, I think it would be nice to add several tests with JSX (just like in `explicit-function-return-type.test.ts` mentioned above)

developer-bandi:
@auvred  It was written with reference to the test case of explicit-function-return-type, but if there is a case where more is needed, please let me know.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -112,6 +139,20 @@ export default createRule<Options, MessageId>({
         if (invalidAncestor.type === AST_NODE_TYPES.ArrowFunctionExpression) {
           // handle arrow function shorthand
 
+          if (options.ignoreVoidInVoid) {
+            if (
+              (invalidAncestor.returnType &&
+                isVoidLikeType(invalidAncestor.returnType)) ||
```
auvred, line 145:
```suggestion
              invalidAncestor.returnType ||
```

All tests still pass without this `isVoidLikeType` call. We should either add some tests to prove that this line is really needed, or remove it.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -112,6 +139,20 @@ export default createRule<Options, MessageId>({
...
+              isValidFunctionExpressionReturnType(invalidAncestor, {
+                allowDirectConstAssertionInArrowFunctions: true,
```
auvred, line 147:
```suggestion
```

[Refactor] Unit tests still pass when this line is removed. Do we really need it? As far as I know, TypeScript will complain if we do `console.log() as const`

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +208,27 @@ export default createRule<Options, MessageId>({
...
+            if (
+              targetNodeFunctionAncestorNodeHasVoidReturnType(invalidAncestor)
+            ) {
+              return;
+            }
+
+            if (
+              functionNode != null &&
+              (isValidFunctionExpressionReturnType(functionNode, {
+                allowDirectConstAssertionInArrowFunctions: true,
```
auvred, line 223:
```suggestion
```
[Refactor] Same as above ^

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +208,27 @@ export default createRule<Options, MessageId>({
         if (invalidAncestor.type === AST_NODE_TYPES.ReturnStatement) {
           // handle return statement
 
+          if (options.ignoreVoidInVoid) {
+            const functionNode = findFunction(invalidAncestor);
```
auvred, line 212: [Bug] The following code won't be reported, because `findFunction` doesn't stop on function declarations:

```ts
type Foo = () => void;

const test1: Foo = function () {
  function testtt() {
    return console.log(); // not reported - bug
  }
};


const test2 = function () {
  function testtt() {
    return console.log(); // reported - nice
  }
};

```

[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.ts&code=C4TwDgpgBAYg9nKBeKAKAlMgfFAbnASwBMBuAKAGM4A7AZ2CmAnoC5YFkoAzAV2ouAEaaTAG8yUbnwFDqjZsEUYo4yZIBOEYD3VyqdOABsIAOkNwA5hnKSAvmVvkgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiIT2AL7eAupp%2BPkA&tsconfig=&tokens=false)

auvred (changes requested):
I like the new added tests, but there is one bug!

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: If a returnType exists, but the returnType is not void, it is not a valid typescript syntax, so there seems to be no problem with removing it.

```typescript
const arrowFn = ():undefined => console.log('foo'); // type error
```

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +447,36 @@ export default createRule<Options, MessageId>({
       const type = getConstrainedTypeAtLocation(services, targetNode);
       return tsutils.isTypeFlagSet(type, ts.TypeFlags.VoidLike);
     }
+
+    function isVoidLikeType(typeNode?: TSESTree.TSTypeAnnotation): boolean {
+      const services = getParserServices(context);
+      return (
+        !!typeNode &&
+        !tsutils.isTypeFlagSet(
```
yeonjuan, line 455: Hi @developer-bandi 
It seems a typo. `isVoidLikeType` will return `false` when the given typeNode is void type.

```suggestion
        tsutils.isTypeFlagSet(
```

It can cause false positives such as the following case.

```ts
// ignoreVoidInVoid: true
 (): any => console.log('foo'); // no error
```

If there is no option, It's properly reported. [playground](https://typescript-eslint.io/play/#ts=5.4.2&fileType=.tsx&code=BQSgXABAhgdgnhAvAPggYwPYwM4YDYCmAdHhgObADkAZhhpSANxA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90URNGgdokcGAC%2BIPUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +211,33 @@ export default createRule<Options, MessageId>({
...
+              if (functionNode.type === AST_NODE_TYPES.FunctionDeclaration) {
+                if (functionNode.returnType) {
+                  return;
+                }
+              } else {
+                if (
+                  isValidFunctionExpressionReturnType(functionNode, {
+                    allowTypedFunctionExpressions: true,
+                  }) ||
+                  ancestorHasReturnType(functionNode)
```
auvred, line 233: [Testing] All tests still pass without this line. Could you add something like to prove that we need this line?

```ts
type HigherOrderType = () => () => () => void;
const x: HigherOrderType = () => () => () => { return console.log() };
```

```ts
type HigherOrderType = () => () => () => void;
const x: HigherOrderType = () => () => function () { return console.log() };
```

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +211,33 @@ export default createRule<Options, MessageId>({
...
+                ) {
+                  return;
+                }
+              }
+            }
```
auvred, line 238: `targetNodeFunctionAncestorNodeHasVoidReturnType` basically does the same that `findFunction` does + check if the found function has `returnType`. Couldn't we just add `|| functionNode.returnType` to the last if-statement? I think the rule logic won't be changed

```suggestion
            if (functionNode != null) {
              if (functionNode.type === AST_NODE_TYPES.FunctionDeclaration) {
                if (functionNode.returnType) {
                  return;
                }
              } else if (
                  isValidFunctionExpressionReturnType(functionNode, {
                    allowTypedFunctionExpressions: true,
                  }) ||
                  ancestorHasReturnType(functionNode)
                  || functionNode.returnType
                ) {
                  return;
                }
            }
```

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
```diff
@@ -473,5 +739,149 @@ function notcool(input: string) {
...
+{ console.log('foo'); return; }
+      `,
+    },
+    {
+      options: [{ ignoreVoidInVoid: true }],
+      code: `
+      type Foo = () => void;
+      const test1: Foo = function () {
+        function testtt() {
+          return console.log(); // not reported - bug
```
auvred, line 840: [Testing] I don't think we need these comments in tests because they might confuse people (the bug has already been fixed and this line is reported, but the comment says there is a bug)

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
```diff
@@ -473,5 +739,149 @@ function notcool(input: string) {
...
+          line: 5,
+          column: 18,
+          messageId: 'invalidVoidExprReturnLast',
+        },
+      ],
+      output: `
+      type Foo = () => void;
+      const test1: Foo = function () {
+        function testtt() {
+          console.log(); // not reported - bug
```
auvred, line 855: [Testing] Same as above ^

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
```diff
@@ -473,5 +739,149 @@ function notcool(input: string) {
...
+      };
+      `,
+    },
+    {
+      options: [{ ignoreVoidInVoid: true }],
+      code: `
+      type Foo = () => void;
+      const test2 = function () {
+        function testtt() {
+          return console.log(); // reported - nice
```
auvred, line 866: [Testing] Same as above ^

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
```diff
@@ -473,5 +739,149 @@ function notcool(input: string) {
...
+          line: 5,
+          column: 18,
+          messageId: 'invalidVoidExprReturnLast',
+        },
+      ],
+      output: `
+      type Foo = () => void;
+      const test2 = function () {
+        function testtt() {
+          console.log(); // reported - nice
```
auvred, line 881: [Testing] Same as above ^

auvred (changes requested):

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: I needed that line for the above case, but I missed adding the test case. thank you!

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: The code you presented was an obvious error, but it was naturally removed as the function was no longer used during the refactoring process below.

However, the exception case you presented was occurred because the return type did not expect a type other than void, and was resolved by adding logic to check whether the return type was void or unknown.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: To remove the function, check the returnType of all functionNodes and execute the `isValidFunctionExpressionReturnType` function only for `functionExpression` and `arrowFn` to avoid errors, so I slightly modified the suggested code to reflect it.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -167,6 +215,29 @@ export default createRule<Options, MessageId>({
...
+                    AST_NODE_TYPES.TSAnyKeyword &&
+                  functionNode.returnType.typeAnnotation.type !==
+                    AST_NODE_TYPES.TSUnknownKeyword) ||
+                ((functionNode.type === AST_NODE_TYPES.FunctionExpression ||
+                  functionNode.type ===
+                    AST_NODE_TYPES.ArrowFunctionExpression) &&
+                  isValidFunctionExpressionReturnType(functionNode, {
+                    allowTypedFunctionExpressions: true,
+                  })) ||
+                ancestorHasReturnType(functionNode)
```
auvred, line 234: [Refactor] What do you think about moving this condition to the separate function?

That way we can avoid code duplication:

https://github.com/typescript-eslint/typescript-eslint/blob/f392b3961fcb478d4f8f8d2df44301b9dfc0e56a/packages/eslint-plugin/src/rules/no-confusing-void-expression.ts#L148-L158

auvred (changes requested):

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +422,33 @@ export default createRule<Options, MessageId>({
       const type = getConstrainedTypeAtLocation(services, targetNode);
       return tsutils.isTypeFlagSet(type, ts.TypeFlags.VoidLike);
     }
+
+    function hasVoidReturnType(
```
yeonjuan, line 426: What do you think about renaming this function? maybe `hasValidReturnType`?

Because, this function doesn't actually check whether the given function's return type is specified void or not.

developer-bandi: i agree. that make sense!

auvred (changes requested):
The code changes and tests look good to me! But there is one issue with docs

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +422,33 @@ export default createRule<Options, MessageId>({
...
+        | TSESTree.FunctionDeclaration
+        | null,
+    ): boolean {
+      if (node != null) {
+        if (
+          (node.returnType &&
+            node.returnType.typeAnnotation.type !==
+              AST_NODE_TYPES.TSAnyKeyword &&
+            node.returnType.typeAnnotation.type !==
+              AST_NODE_TYPES.TSUnknownKeyword) ||
```
auvred, line 439: [Bug] I somehow missed this in previous reviews: we can't rely on syntactic check here

```ts
(): any => console.log(); // reported - nice!
(): unknown => console.log(); // reported - nice!

type Foo = any
(): Foo => console.log(); // not reported - bug!
type Bar = unknown
(): Bar => console.log(); // not reported - bug!
```

[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.4.3&fileType=.ts&code=BQSgXABAhgdgnhAvAPggYwPYwM4YDYCmAdHhgOagDcEA9DRAE4EAOGDALgQCYQC0EMAJZoCAQgBQoSAFcYAaxgYA7jCSpMOfMVIUQ1OoxZtOPfkJETx7OMwIQAYhgxJo8SeAdO16LLkIlyKlp6RXZDVg5uPggAI2kyCWtbCAAhKAYXWQVlGHdINIyUH01-HSCDUPDjKP44hKA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiIT2AL7eAupp%2BPkA&tsconfig=&tokens=false)

auvred (changes requested):

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: When specifying a type using a type alias, it is difficult to find a way to determine what type it is.

Could you please give me a hint?

```
type Bar = unknown
(): Bar => console.log();  // how to know Bar is unknown type in create fn
```

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
auvred, line 1: [Testing] I think it would be nice to add few tests after https://github.com/typescript-eslint/typescript-eslint/pull/8809 is merged. This PR fixes the `ancestorHasReturnType` behavior. 

We can add something like this:

```ts
function foo(): void {
  () => () => console.log();
}

function bar(): any {
  () => () => console.log();
}
```

(both of the above cases should be reported)

auvred (changes requested):

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: @auvred Can you help with the issue left in the comment above?

auvred: I'm not 100% sure about this, but I'd try playing around with function signatures. There is nice utility function in the `ts-api-utils` package - `getCallSignaturesOfType`.

See: https://github.com/JoshuaKGoldberg/ts-api-utils/blob/d808eba09baa3ba093cd31b85aad57bdcc539867/src/types/getters.ts#L23

Search with some examples of usage: https://github.com/search?q=repo%3Atypescript-eslint%2Ftypescript-eslint%20getCallSignaturesOfType&type=code

auvred (changes requested):
Requesting changes on https://github.com/typescript-eslint/typescript-eslint/pull/8632#discussion_r1567182022. I will re-review this PR more thoroughly after this comment is resolved.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: it works to me very well. thank you!

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +427,55 @@ export default createRule<Options, MessageId>({
       const type = getConstrainedTypeAtLocation(services, targetNode);
       return tsutils.isTypeFlagSet(type, ts.TypeFlags.VoidLike);
     }
+
+    function hasValidReturnType(
+      services: ParserServicesWithTypeInformation,
```
auvred, line 432: [Refactor] (optional) `hasValidReturnType` function is declared in block where `context` is visible. What do you think about calling `const services = getParserServices(context);` inside `hasValidReturnType` (like `canFix` function above does) instead of passing `services` as an argument?

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +427,55 @@ export default createRule<Options, MessageId>({
...
+    ): boolean {
+      if (node != null) {
+        const checker = services.program.getTypeChecker();
+        const functionTSNode = services.esTreeNodeToTSNodeMap.get(node);
+        const functionType =
+          (ts.isFunctionExpression(functionTSNode) ||
+          ts.isArrowFunction(functionTSNode)
+            ? getContextualType(checker, functionTSNode)
+            : services.getTypeAtLocation(node)) ??
+          services.getTypeAtLocation(node);
```
auvred, line 447: [Bug] Why do we get a contextual type for function expressions and arrow functions? 

The rule you took this code from has a detailed explanation of why it calls `getContextualType` for function expressions and arrow functions:

https://github.com/typescript-eslint/typescript-eslint/blob/f248e689b04bdfd81429977a731ba6b8ad82622e/packages/eslint-plugin/src/rules/no-unsafe-return.ts#L91-L99

This means that the following case is reported:

```ts
const test: () => any  = (): void => console.log()
```

(it should not be reported since `(): void => console.log()` is allowed with `ignoreVoidInVoid: true`)

[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.4.3&fileType=.ts&code=MYewdgzgLgBFCm0BcMAUBKGBeAfDAhmAJ4zZrooBuIAlgCbZ6iQgA28AdKyAOYZA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANqRE0aB2iQANJkiVhnRQDVJUgJJNL01fjiIAvgF1wYFyBdA&tsconfig=&tokens=false)

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +427,55 @@ export default createRule<Options, MessageId>({
...
+
+        if (functionTSNode.type) {
+          for (const signature of tsutils.getCallSignaturesOfType(
+            functionType,
+          )) {
+            return !(
```
auvred, line 453: [Bug] This code returns from `for` loop on the first iteration. So if a function has several signatures, the rule will check only the first signature:

```ts
function test1(): void
function test1(arg: string): any
function test1(arg?: string): any | void {
  if (arg) {
    return arg
  }

  return console.log() // not reported - ok
}

function test2(arg: string): any
function test2(): void
function test2(arg?: string): any | void {
  if (arg) {
    return arg
  }

  return console.log() // reported - bug
}
```

[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.4.3&fileType=.ts&code=GYVwdgxgLglg9mABFApgZygRgBQEoBciAbnDACYBQoksCy6W2AhgE4DmhGLMYbBiTMAE8q4aPCSoMOVmwD8nKN179BQxAB9ipMogDeFRIhjBEzdrn2GjiFiiggWSWdYC%2BFa3YdPEEBGjgAGxQAOkC4NjxEAHpoxDA4KFsUAAc4FlRdAFpEOABrCndRGgl6DAAmcw5ELh4%2BQjVi8TopKEr%2BEnIm2kkGStkFGqU61WFNbXIrIxMzWUsDG2TvZ3Y3DyMvRyQ-MADgsIio2OS0jJRsxAAjEDZCoA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANqRE0aB2iQANJkiVhnRQDVJUgJJNL01fjiIAvgF1wYFyBdA&tsconfig=&tokens=false)

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +427,55 @@ export default createRule<Options, MessageId>({
...
+              tsutils.isTypeFlagSet(
+                signature.getReturnType(),
+                ts.TypeFlags.Any,
+              ) ||
+              tsutils.isTypeFlagSet(
+                signature.getReturnType(),
+                ts.TypeFlags.Unknown,
+              )
```
auvred, line 461:
```suggestion
              tsutils.isTypeFlagSet(
                signature.getReturnType(),
                ts.TypeFlags.Any | ts.TypeFlags.Unknown,
              )
```

[Refactor] Let's do a bitwise OR of these flags instead of calling `isTypeFlagSet` twice?

auvred (changes requested):

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: I did not understand that the isTypeFlagSet function utilizes bitwise operators, so I called it twice. thank you for telling me!

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
developer-bandi: I too believe that using context is better than passing a service.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -376,5 +422,44 @@ export default createRule<Options, MessageId>({
...
+        if (
+          ((node.type === AST_NODE_TYPES.FunctionExpression ||
+            node.type === AST_NODE_TYPES.ArrowFunctionExpression) &&
+            isValidFunctionExpressionReturnType(node, {
+              allowTypedFunctionExpressions: true,
+            })) ||
+          ancestorHasReturnType(node)
+        ) {
+          return true;
+        }
```
auvred, line 459: If we check that function signature don't contain `any` and `unknown` in the return type, we should do the same for typed functions/variable declarations. The following cases are not reported:

```ts
type HigherOrderType = () => any;
(): HigherOrderType => () => console.log();

const x: HigherOrderType = () => console.log();

```
[playground](https://deploy-preview-8632--typescript-eslint.netlify.app/play/#ts=5.4.3&fileType=.ts&code=C4TwDgpgBAEglgcwBYQE4HlUBM0BVzQC8UAFAJRSEB8UAhgHYgDcAUOQFyyIobZ4GUa5QVADGAe3oBncQBsIAOlniE5ViwnTgUAB6d4yNJhyp8kSqQrUxkmfKUq1QA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1rI6YDNYyZgHNaANw6UAJvQAexaCiG90YANrhsORNGgdokADSatWLdkiVhnRQDVJUgJJN701fjiIT2AL7eAupp%2BPkA&tsconfig=&tokens=false)

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
```diff
@@ -119,6 +119,325 @@ function cool(input: string) {
...
+    {
+      code: `
+function test2(arg: string): any;
+function test2(): void;
+function test2(arg?: string): any | void {
+  if (arg) {
+    return arg;
+  }
+
+  return console.log(); // reported - bug
```
auvred, line 430: Please see my comment from the one of the previous reviews https://github.com/typescript-eslint/typescript-eslint/pull/8632#discussion_r1539679818. :slightly_smiling_face: 

The code samples I provide, which contain some error messages, are for reference only. They are not intended to be directly copied into real tests. I'm including comments on these code samples to better illustrate what the bug I found is. The same goes for variable names: `test1` and `test2` are named to avoid name collisions in the playground. Having the `test2` function in a separate test case without `test1` looks a bit confusing.

I ask you to pay attention to the reviews and not just blindly insert some stuff into the code, assuming that someone will notice it in one of the next reviews anyway.

auvred (changes requested):

packages/eslint-plugin/tests/rules/no-confusing-void-expression.test.ts
developer-bandi: I apologize for making the same mistake by not checking the code thoroughly after writing it.

In the future, I will look more closely before submitting a review.

developer-bandi:
I tried several methods to correct the error, but it seems that I need to check if the return type is void, rather than excluding the current any and unknown types. 

However, there is currently not much time to work on this, so only the test code in this PR has been modified and reflected. If there is anyone who wants to continue this work or it has been too long, you can close the pr or provide a new pr.

JoshuaKGoldberg:
👍 thanks for the heads up, and of course all your work on this @developer-bandi!

Closing the PR out now so if other folks want to tackle it, they can. If anybody wants to drive it forward, please do post your own PR - and if you use this as a start, consider adding `Co-authored-by: @developer-bandi` at the end of your PR description. Thanks! 😊

PR #8640, opened 2024-03-10
fix(eslint-plugin): [no-unused-vars] clear error report range

### LINKED ISSUE #8543: Enhancement: Improve lint error report location

### Relevant Package

typescript-eslint

### Description

Some rule's error report location seems too wide. I think it could be improved to report more accurate location where user should care about.For example,

- [x] no-unused-vars: #8640
  ```ts
  // AS-IS
  let unusedVar: { foo: number } | undefined = 1;
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  // TO-BE
  let unusedVar: { foo: number } | undefined = 1;
       ^^^^^^^^
  ```

- [x] explicit-member-accessibility : #8869
  ```ts
  // AS-IS
  class Animal {
     get name(): string {  return this.animalName;  }
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  }
  // TO-BE
   class Animal {
     get name(): string {  return this.animalName;  }
     ^^^^^^^^^
  }
  ```

- [x] no-for-in-array #8874

  ```ts
  // AS-IS
  for (const i in array) { console.log(array[i]); }
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  // TO-BE
  for (const i in array) { console.log(array[i]); }
  ^^^^^^^^^^^^^^^^^^^^^
  ```

- [x] init-declarations https://github.com/typescript-eslint/typescript-eslint/pull/8893
  ```ts
  // AS-IS
  let bar: number | undefined | string;
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  bar = 2;
  // TO-BE
  let bar: number | undefined | string;
     ^^^^^
  bar = 2;
  ```
- [ ] prefer-readonly https://github.com/typescript-eslint/typescript-eslint/pull/8894

  ```ts
  // AS-IS
  class Container {
    private onClick = () => { /* ... */ };
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  }
  // TO-BE
  class Container {
    private onClick = () => { /* ... */ };
    ^^^^^^^^^^^^^^^
  }
  ```
- ~~space-before-blocks~~ (formatting/layout rule)

... maybe more 

### Additional Info

_No response_

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes (partially) #8543

## Overview

clear error range in `no-unused-vars` rule


yeonjuan:
Hi @developer-bandi , I think it would be nice to explicitly specify `endColumn` in some invalid test cases to guarantee the behavior and to prevent regressions! 👍

- no-unused-vars/no-unused-vars.test.ts
```ts
{
          messageId: 'unusedVar',
          line: 3,
          column: 1,
          endColumn:  3, <<<<<<< add `endColumn` on some test cases
          data: {
            varName: 'foo',
            action: 'assigned a value',
            additional: '',
          },
        },
```

developer-bandi:
@yeonjuan  thank you, i add test case with `endline` and `endColumn` that cover shorten error range

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -419,10 +419,24 @@ export default createRule<Options, MessageIds>({
                 ref.from.variableScope === unusedVar.scope.variableScope,
             );
 
+            const node = writeReferences.length
+              ? writeReferences[writeReferences.length - 1].identifier
+              : unusedVar.identifiers[0];
+
+            const { start } = node.loc;
+            const nodeVariableLength = node.name.length;
```
yeonjuan, line 427: LGTM!  What do you think about changing the variable names a bit?

```suggestion
            const id = writeReferences.length
              ? writeReferences[writeReferences.length - 1].identifier
              : unusedVar.identifiers[0];

            const { start } = node.loc;
            const idLength = node.name.length;
```

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -419,10 +419,24 @@ export default createRule<Options, MessageIds>({
...
+                line: start.line,
+                column: start.column + nodeVariableLength,
+              },
+            };
+
             context.report({
-              node: writeReferences.length
-                ? writeReferences[writeReferences.length - 1].identifier
-                : unusedVar.identifiers[0],
+              node,
```
kirkwaiblinger, line 438: nit
```suggestion
```
Unless there is some other effect that `node` has that is not documented, supplying `loc` takes precedence over `node`, so I think it's misleading to still supply `node`

packages/eslint-plugin/src/rules/no-unused-vars.ts
developer-bandi: it make sense, so i change variable name. thank you!

packages/eslint-plugin/src/rules/no-unused-vars.ts
developer-bandi: i delete node. thank you!

kirkwaiblinger: Ah, looking closer at this... I may have led you astray, at least partially. Sorry about that! The test failures in CI are due to having removed the `node` from the report, since some of the tests specifically check for the type of the `node`
(the following function generates error assertions that demand an `Identifier` be reported on via the `type` field) 
https://github.com/typescript-eslint/typescript-eslint/blob/584db29ec44ce4e9cb71afac35d48994889168e6/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts#L56-L77

Now, it's still not clear to me that the value of the `node` is actually used in any way other than being tested, though, when the `loc` is also provided. I ran into the same problem in https://github.com/typescript-eslint/typescript-eslint/pull/8874 and just removed the `type` from the test cases, but I'm not totally sure if that is sound.

My thought is 
if it has no effect => we should remove the `type` from the test, since it's not testing anything at all, but it looks like it is
if it does have an effect => just put `node` back. 

Maybe @JoshuaKGoldberg or @bradzacher will know for sure? Would either of you know whether there's any reason to provide a report `node` if `loc` is being provided?

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -419,10 +419,23 @@ export default createRule<Options, MessageIds>({
                 ref.from.variableScope === unusedVar.scope.variableScope,
             );
 
+            const id = writeReferences.length
```
kirkwaiblinger, line 422: Even though these branches collapse to do the same thing, I think it could be good to test both branches, in case one day they get separated (for example, to report different error messages)

What do you think about adding test cases with all 4 report loc fields for some cases like the following?
```ts
let foo: number;
```
```ts
let foo: number = 1;
foo = 2;
foo = 3;
```

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
kirkwaiblinger, line 1: (testing) 
Could you also add a test case that uses the `declare` modifier with the full `loc` data?

FYI - here and in other comments, there's probably no need to add _new_ test cases; modifying existing ones to have the full `loc` data would work too 🙂

kirkwaiblinger (changes requested):
The code looks good to me! Just requesting a bit more test coverage of the edge cases.
fireworks

kirkwaiblinger (approved):

packages/eslint-plugin/src/rules/no-unused-vars.ts
JoshuaKGoldberg: Yeah you only need to provide one or the other.

armano2 (dismissed):

bradzacher (dismissed):

yeonjuan:
The [test in CI](https://github.com/typescript-eslint/typescript-eslint/actions/runs/8754864666/job/24027842994?pr=8640) is failing. This is likely due to the fact that we removed “node” from the report and changed it to use “loc”. 
Maybe we should remove `type` in the `no-unused-vars-eslint.test.ts`. (or revert the change). 

https://github.com/typescript-eslint/typescript-eslint/blob/219b841206656d2364b7f1a9ce61ceeee68bbbe6/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts#L1339-L1345

kirkwaiblinger (changes requested):
In great shape - only thing remaining blocking merge is resolution of https://github.com/typescript-eslint/typescript-eslint/pull/8640#discussion_r1571761505 (by removing type from test cases). Thanks!

kirkwaiblinger (approved):

cpojer:
I believe this change broke [`eslint-plugin-unused-imports`](https://github.com/sweepline/eslint-plugin-unused-imports), which relies on `no-unused-vars` from this plugin to pull the node and amend the error. See https://github.com/sweepline/eslint-plugin-unused-imports/blob/master/lib/rules/load-rule.js for the code that loads this rule.

It now crashes here because `node` is undefined: https://github.com/sweepline/eslint-plugin-unused-imports/blob/master/lib/rules/predicates.js#L28

I don't see a good reason why `node` has to be removed. Would you mind putting `node` back into the `context.report` call?

bradzacher:
It does seem a very brittle that that plugin is depending on our private, internal implementation details like that.

For example if we were to change the rule to report on a different node then the rule would similarly break.

From the looks of it that's a horridly inefficient plugin too.
It utilises `eslint-rule-composer`.
That util works by:
1) duplicating the rule by wrapping its create function in a new rule
2) monkey-patches the `context` object that ESLint passes so that it can intercept the reports generated by the wrapped rule.

(1) is especially bad because it means you have multiple copies of the rule. i.e. if you use both `unused-imports/no-unused-imports` and `unused-imports/no-unused-vars` then you're running the `no-unused-vars` rule TWICE.

This is a really unstable setup and I'm really not sure if we should be supporting usage of our private internals like that.

----

I'm not sure I understand the point of the plugin - what is the usecase for separating out the imports from the regular variables? What's the benefit from working this way?

cpojer:
Yup, I get your argument. In this case you could argue it keeps API shape compatibility with the corresponding rule in eslint, if you need a reason.

The plugin comes with a fixer, which I run when hitting `cmd+s` in VS Code, and it removes all unused imports automatically. If you hoisted this feature into your version of `no-unused-vars`, I can imagine dropping `eslint-plugin-unused-imports`.

Note that the ESLint plugin has almost 2m downloads per week, and I'm sure many are using it while also using typescript-eslint, so you'll likely hear from more people soon. I'm just the one who lives on the bleeding edge :D

cc @sweepline – would you be open to porting this rule into `eslint-plugin-unused-imports`? If yes, @bradzacher would you be open to making a patch release that adds back `node` until that happens?

timtucker-dte:
@bradzacher we're using the plugin for auto-removing unused imports on save / build as well

JacobZyy:
This plugin åy used in development, primarily for detecting and automatically removing unused imports (with an emphasis on automatic deletion). Additionally, I believe that the operation of deleting node attributes has altered the original data structure. Should this be considered a breaking change and thus be removed in the upcoming version 8.0?

> It does seem a very brittle that that plugin is depending on our private, internal implementation details like that.
> 
> For example if we were to change the rule to report on a different node then the rule would similarly break.
> 
> From the looks of it that's a horridly inefficient plugin too. It utilises `eslint-rule-composer`. That util works by:
> 
> 1. duplicating the rule by wrapping its create function in a new rule
> 2. monkey-patches the `context` object that ESLint passes so that it can intercept the reports generated by the wrapped rule.
> 
> (1) is especially bad because it means you have multiple copies of the rule. i.e. if you use both `unused-imports/no-unused-imports` and `unused-imports/no-unused-vars` then you're running the `no-unused-vars` rule TWICE.
> 
> This is a really unstable setup and I'm really not sure if we should be supporting usage of our private internals like that.
> 
> I'm not sure I understand the point of the plugin - what is the usecase for separating out the imports from the regular variables? What's the benefit from working this way?

This plugin is currently widely be used for detecting and automatically removing unused imports (the point is automatic deletion). Additionally, I believe that the operation of deleting `node` has altered the original data structure, so it be considered as a breaking update and published in v8.0 rather than v7.8?

dawn-ylcode:
I ran into the same issue, and now it's just messed up the whole plugin. I don't think they should be doing these destructive updates.

bradzacher:
Let me be very clear:
We changed a private, explicitly not exposed to the public, internal implementation detail.

The other plugin **monkey patches** eslint's API so it can hook into our _internals_.

There is no breaking change here.

If something builds on top of private implementation details and breaks when we change those details - that is not a breaking change.

**That fact is not up for debate**.

---

How to best remedy this is an open question that we shall discuss amongst the maintenance team and make a decision.

HitkoDev:
To be clear: the plugin in question is used by almost 10% of `typescript-eslint` userbase. The plugin itself adds ~100 lines of code to automatically remove unused variables and imports. 

Is there any good reason such functionality shouldn't be a part of `typescript-eslint`?

JacobZyy:
mabe typescript-eslint can set a new rule named `no-unused-imports` and make it fixable. the fact is, I have three plugin have a rule named `'no-unused-vars'`, and I also really need the auto-fixable rule ` unused-imports `
Besides, sometimes `no-unused-imports` is to noisy in editor, and that can refer to the solution of @antfu 's config, turn off the rule in editor and turn it on in the cli.
image

JoshuaKGoldberg:
> plugin in question is used by almost 10% of `typescript-eslint` userbase

No, it's very unlikely that 1 in 10 of our users are using `eslint-plugin-unused-imports`. npm download numbers are an inaccurate metric (reference: 2.3m for the plugin, 27m for us today). Another inaccurate measurement would be GitHub stars, 420 vs 14.6k today.

A more reliable metric would be something like Sourcegraph searches of open source repositories' `package.json` files today:
* [3.3k results for `eslint-plugin-unused-imports`](https://sourcegraph.com/search?q=context:global+eslint-plugin-unused-imports+file:package.json%24+-path:node_modules&patternType=keyword&sm=0)
* [111.3k results for `@typescript-eslint/eslint-plugin`](https://sourcegraph.com/search?q=context:global+%40typescript-eslint/eslint-plugin+file:package.json%24+-path:node_modules+count:200000&patternType=keyword&sm=0)
  * [841 results for the new `typescript-eslint`](https://sourcegraph.com/search?q=context:global+%22%5C%22typescript-eslint%5C%22%22+file:package.json%24+-path:node_modules&patternType=keyword&sm=0)

At least in open source land, it seems more likely that the `no-unused-imports` plugin is floating around 3% of our users.

JoshuaKGoldberg:
> Is there any good reason such functionality shouldn't be a part of `typescript-eslint`?

Two reasons. The first is that [@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars) acts as an "extension rule" on top of the base [eslint/no-unused-vars](https://eslint.org/docs/rules/no-unused-vars) rule. We don't add any major new functionality on top of base rules here. If you want that added, it'd be an issue for ESLint core.

The reason why ESLint core isn't enthusiastically adding in the automatic removal of values is that it's actually a tricky thing to get right. Seemingly unused imports and variables can actually have side effects that change your code's behavior - which is very dangerous to have in an automatic rule fixer!

Example of an import with a side effect:

```js
// a.js
let count = 0;
console.log("Gotcha!");
export const getCount = () => count;
```

```js
// b.js

// Unused. Should we auto fix to...
//  import {} from "./a.js" ?
//  import "./a.js" ?
//  nothing at all?
import { getCount } from "./a.js";

console.log("Hi!");
```

Example of a variable with a side effect:

```ts
let count = 0;

export function getNextCount() {
  return (count += 1);
}

// Unused. Should we auto-fix to...
//  getNextCount()
//  nothing at all?
let unusedIsh = getNextCount();
```

Example of a variable without:

```js
function hasNoSideEffect() {
  for (let i = 0; i < 9001 ** 7; i += 1) {}
}

// Unused. Should we auto-fix to...
//  hasNoSideEffect()
//  nothing at all?
let alsoUnusedIsh = hasNoSideEffect()
```

By using a plugin like [eslint-plugin-unused-imports](https://github.com/sweepline/eslint-plugin-unused-imports), you explicitly are taking on the risk of changing code behavior due to side effects like that. Which, hey, might be a great sign about how you've structured your code 😄 nice job avoiding side effects! But lint rules can't make that assumption for the general case.

One thing lint rules _can_ do is add _suggestions_: which are like _fixes_ but won't be applied automatically. **[Rule Change: Add support for suggestions for unused var](https://github.com/eslint/eslint/issues/17545)** tracks adding that on the ESLint side. That issue was accepted, assigned, and has an open pull request now! 🙌 

For more info on the difference between a _fix_ and a _suggestion_, see:

* https://eslint.org/docs/latest/use/core-concepts/#rule-fixes
* https://eslint.org/docs/latest/use/core-concepts/#rule-suggestions

JoshuaKGoldberg:
Given that:

* Reports from our rules are an implementation detail not covered by our public API contract (see [typescript-eslint.io > Versioning > `eslint-plugin`](https://typescript-eslint.io/users/versioning#eslint-plugin) and _(I'll file an issue in a moment)_ to make this more clear)
* `eslint-plugin-unused-imports` is stuck having to rely on the specifics of what the rule reports (there's nothing better it _could_ use)

...the "right" solution would be for `eslint-plugin-unused-imports` to work with what it's given: either the `loc` or `node` depending on the version of `@typescript-eslint/eslint-plugin`. That's being tracked by https://github.com/sweepline/eslint-plugin-unused-imports/issues/77, and has a PR open already in https://github.com/sweepline/eslint-plugin-unused-imports/pull/78.

JoshuaKGoldberg:
Anyway, per [our PR contributing guide](https://typescript-eslint.io/contributing/pull-requests), a previously-merged PR isn't the right place to discuss changes. We'd generally ask that bug reports and feature requests be reported as a standalone new issue so they're more discoverable. We get that this was a blocker for everyone using `eslint-plugin-unused-imports` so it's reasonable that the discussion ended up happening here. Thanks for the reports, all! ❤️ 

Locking this issue as this specific typescript-eslint PR doesn't need any more discussion. If more issues on the topic get filed I'll make sure they show up here. Cheers!

bradzacher:
The issue with an autofixer is that a lot of people run them on save.

So imagine this:
1) you add `import foo from "foo"` to your file
2) you save the file
3) you attempt to use the variable `foo`
4) you get an error because the import was deleted in (2)

There's many variations of this situation - the development loop is an inherently messy thing and there's no way for a static analysis tool to delineate between "unused and it should be deleted" and "unused but it should not be deleted".

There's also the issue of side-effects that Josh mentioned. If you autofix delete an import or a variable - you can change runtime behaviour because side-effects are lost.

There actually used to be a fixer for `no-unused-vars` a long, long time ago but it was removed because people found it both annoying and dangerous.

There's always scope to introduce an optional fixer for people to opt-in to the danger / annoyance. But as Josh mentioned the rule is intended to be a TS-compatible replacement for the base rule. Whilst it _is_ a complete rewrite - the options, features and reports of the rule are the same.
We don't like to break too far from the base rules for our extension rules because it places a heavy maintenance burden on us. The more forking we do; the harder it is to merge upstream changes.

PR #8643, opened 2024-03-11
docs(website): display count on error tab in playground 

### LINKED ISSUE #8641: Enhancement: Display error count on tabs in playground

### Relevant Package

website

### Description

In "Errors", "ESTree"... tabs in Playground, it would be nice to show the number of errors.
This way, even if we don't select the tab, we can aware that errors occurred.

| AS-IS | TO-BE |
| --- | --- |
| as-is  | to be preview |

### Additional Info

_No response_

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8641

## Overview
As if raising an issue, make error count view in playground 

In my opinion, there is no great meaning in indicating all the number of errors, and it is not neatly displayed in the UI, so errors with more than 9 are indicated as 9+.


yeonjuan:
Hi @developer-bandi, Looks like a good start! I haven't dug deep into the website code, so I'm commenting based on the behavior shown in the preview.

The intent of creating this issue was to recognize that an error occurred on a particular tab even when it was not selected.
However, with the current implementation, it seems that only the selected tab exposes the error count.

https://github.com/typescript-eslint/typescript-eslint/assets/41323220/70089375-9161-4661-bc77-3791001b3adf

packages/website/src/components/layout/EditorTabs.tsx
```diff
@@ -36,6 +38,13 @@ function EditorTabs<T extends boolean | string>({
             onClick={(): void => change(item.value)}
           >
             {item.label}
+            {additionalTabsInfo?.[item.label] ? (
+              <div>
+                {additionalTabsInfo[item.label] > 9
+                  ? '9+'
```
Josh-Cena, line 44: "9+" seems really low. Does it significantly impact layout? If not can we use "99+"?

developer-bandi: I adjusted the layout to make longer letters look clear, and then modified it to show up to 99+!

so layout is like below images.

스크린샷 2024-03-12 오후 8 56 35

스크린샷 2024-03-12 오후 8 56 49

스크린샷 2024-03-12 오후 8 56 58

developer-bandi:
I temporarily added code to reflect the [previous comment](https://github.com/typescript-eslint/typescript-eslint/pull/8643#issuecomment-1991827700). (There is an issue with the setMarkers type that could not be resolved, but it is not important for now.)

The current issue is that when entering the playground using a URL with a query containing code information, the warning display is only visible in the first code tab, and subsequent tabs must be activated to activate the warning display.

In order to accomplish this, the Monaco library needs a way to create markers in advance for the code, tsconfig, and eslintrc codes included in the query, but it is difficult to find a way because I am not familiar with the Monaco editor library. Could you please help me with this?

developer-bandi:
add fix commit. 

when go to playground with code, tsconfig, eslintrc code with query, error mark is evoke first time.

armano2:
> Hi @developer-bandi, Looks like a good start! I haven't dug deep into the website code, so I'm commenting based on the behavior shown in the preview.
> 
> > This way, even if we don't select the tab, we can aware that errors occurred.
> 
> As written in the issue, the intent of this enhancement was to recognize that an error occurred on a particular tab even when it was not selected. However, with the current implementation, it seems that only the selected tab exposes the error count. I think the error count should be exposed even when the tab is not selected.

currently sadly there is no way to retrieve errors from other "files", they are not flushed when you change tab

developer-bandi:
@yeonjuan  According to the answers above, it doesn't seem possible to expose an error when the tab is not selected.

Therefore, it seems better to close this mr or to expose the number of errors in the tab to the errros tab as was initially implemented. What do you think?

---
> we should definitly define some state manager for errors, files and all other stuff as amount of properties that we are passing back and forward is slowly getting oo hand
this may be a little oo context of this change, but we should def do it soon

@armano2 I'm not sure if my understanding of the last sentence is correct.

From what I understand, the state in the playground below is not managed, so state management is being introduced using state management libraries such as zustand and redux or context API. Is this correct?

https://github.com/typescript-eslint/typescript-eslint/blob/4d6d0d5950f587780dec998816d458ae4b27a513/packages/website/src/components/Playground.tsx#L27-L46

I understand that this work will not be carried out in this pr, but I am asking this question because I think it would be good to know properly!

yeonjuan:
> Therefore, it seems better to close this mr or to expose the number of errors in the tab to the errros tab as was initially implemented. What do you think?

IMO, It's good to just show the number of errors. But I'd love to hear from typescript-eslint members.

JoshuaKGoldberg:
Yeah 👍 I'd rather get _some_ improvement for now, even if it's not the ideal state. Either way we'll end up with followup issues filed for future refactors.

developer-bandi:
Based on the summarized opinions, I have modified it so that only the number of errors occurring in the current tab can be checked.

packages/website/src/components/layout/EditorTabs.module.css
```diff
@@ -12,12 +12,25 @@
...
 
   font-size: 0.75rem;
+  display: flex;
 }
 
 .tabStyle svg {
   margin-left: 0.3rem;
 }
 
+.tabStyle div {
```
JoshuaKGoldberg, line 22: Selectors that target all tags of a certain type can accidentally target nested children. If we were to, say, add a child component that happens to render `<div>`s, this would target that child's elements accidentally. Let's use a specific class name:

```suggestion
.tabErrors {
```

packages/website/src/components/layout/EditorTabs.tsx
```diff
@@ -36,6 +38,13 @@ function EditorTabs<T extends boolean | string>({
             onClick={(): void => change(item.value)}
           >
             {item.label}
+            {additionalTabsInfo?.[item.label] ? (
+              <div>
```
JoshuaKGoldberg, line 42:
```suggestion
              <div className={styles.tabErrors}>
```

JoshuaKGoldberg (dismissed):
Looks great to me, thanks! ✨ 

Just a small thing to touch up, nothing major on my end. Will re-request review from other Josh too.

packages/website/src/components/layout/EditorTabs.module.css
developer-bandi: i'm agree, so i change it to suggestion code

JoshuaKGoldberg (approved):
Thanks! ✨

PR #8653, opened 2024-03-12
chore: fix test cache inputs, clean up config

JamesHenry:
## Overview

Please see the notes inline for a description of the changes


.eslintrc.js
JamesHenry, line 1: No longer needed

tsconfig.json
```diff
@@ -9,7 +9,6 @@
     "typings",
     "tools/**/*.ts",
     "tools/**/*.mts",
-    ".eslintrc.js",
```
JamesHenry, line 12: File removed

nx.json
```diff
@@ -86,6 +80,12 @@
         "runtime": "yarn -v"
       }
     ],
-    "production": ["default", "!{projectRoot}/src/test-setup.[jt]s"]
+    "production": [
```
JamesHenry, line 83: We were missing a bit more granularity on our definition of "production" files, we can exclude spec files and test specific config files from consideration

nx.json
```diff
@@ -63,7 +58,6 @@
         "{workspaceRoot}/eslint.config.js",
         "{workspaceRoot}/eslint.config.mjs",
         "{workspaceRoot}/yarn.lock",
-        "{workspaceRoot}/.eslintignore",
```
JamesHenry, line 66: File was already removed in a previous PR

nx.json
```diff
@@ -40,6 +34,7 @@
     "@nx/jest:jest": {
       "inputs": [
         "default",
+        "^production",
```
JamesHenry, line 37: This was important - we want the test tasks of projects to be affected when we make changes to the production files of dependencies (which is what the `^` denotes,  "of dependencies")

nx.json
```diff
@@ -31,6 +24,7 @@
     "test": {
       "inputs": [
         "default",
+        "^production",
```
JamesHenry, line 27: This was important - we want the test tasks of projects to be affected when we make changes to the production files of dependencies (which is what the `^` denotes,  "of dependencies")

nx.json
```diff
@@ -16,10 +12,7 @@
       }
     },
     "version": {
-      "generatorOptions": {
```
JamesHenry, line 19: Applying our newer shorthand for using conventional commits for versioning

bradzacher (approved):
You can remove the `eslint.config.js` now too and we can just have the `eslint.config.mjs`

JamesHenry:
@bradzacher done

JamesHenry:
@bradzacher scratch that, it looks like we need to handle `.mjs` flat config files on the nx side for this to work without the `eslint.config.js` being present, reverted for now so that we can land this. I'll update the Nx side

Issue #8548, opened 2024-02-24
Docs: Add maintenance guide for renaming a rule


JoshuaKGoldberg:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

Coming over from https://github.com/typescript-eslint/typescript-eslint/issues/8544#issuecomment-1962365028: occasionally in the past we've felt a need to rename rules. For example, #5841 -> #5879.

This generally involves:
1. Copying the old name rule to a new name
2. Adding `replacedBy` to the meta of the old rule
3. Later, removing the old rule as a breaking change

Let's add docs around this.

### Affected URL(s)

https://typescript-eslint.io/maintenance/issues/rule-renames

JoshuaKGoldberg:
Note from #8656 (thanks for bearing with me @ericmatthys): we should also search open+accepted issues and rename them to the new rule name.

PR #8656, opened 2024-03-12
fix(eslint-plugin): [sort-type-constituents] prevent formatting differences in complex types from affecting sort order

### LINKED ISSUE #3639: [sort-type-union-intersection-members] Sort is formatting-dependent
I'm in a project that uses Redux, where our actions are defined as a discriminated union keyed by `type`. Some of these actions have long names, causing Prettier to format those union members onto multiple lines. I'd like order this union, but it seems like the ordering changes depending on the formatting.

**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/sort-type-union-intersection-members": ["error"]
  }
}
```

```TS
export type Actions1 =
  | { type: "a" }
  | { type: "b" }
  | { type: "c" }
  | { type: "d" };

export type Actions2 =
  | { type: "a" }
  | { type: "b" }
  | { type: "c" }
  | {
      type: "d";
    };
```

**Expected Result**

No changes to the ordering.

**Actual Result**

The `Actions1` type passes the ESLint check, but we receive a linter violation from the `Actions2` object at

```
8:3 - Union type Actions2 members must be sorted.(@typescript-eslint/sort-type-union-intersection-members)
```

If we try applying the autofix, `Actions2` gets sorted into:

```TS
export type Actions2 =
  | {
      type: "d";
    }
  | { type: "a" }
  | { type: "b" }
  | { type: "c" };
```

**Additional Info**

**Versions**

| `@typescript-eslint/eslint-plugin` | `4.28.3` |
| `@typescript-eslint/parser`        | `4.24.0` |
| `ESLint`                           | `7.17.0` |
| `node`                             | `12.13.1` |

ericmatthys:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3639

## Overview

Currently, `sort-type-constituents` compares all source text for a given "constituent". This means that whitespace can affect the sort order unexpectedly. Particularly when paired with a formatter like Prettier, the addition or removal of a line break can end up violating this lint rule.

Considering these two examples, the first results in a lint error and the second does not, even though you'd assume they'd be sorted the same.

```tsx
type T = { a: string } & {
  b: string
};
```

```tsx
type T = { a: string } & { b: string };
```

I've tried to address these differences without completely changing how the rule works by determining a separate text value on which the sort operates, and this value has normalization to better ignore formatting differences between constituents.

### DIFF
--- a/packages/eslint-plugin/src/rules/sort-type-constituents.ts
+++ b/packages/eslint-plugin/src/rules/sort-type-constituents.ts
@@ -170,6 +170,41 @@ export default createRule<Options, MessageIds>({
       numeric: true,
     });
 
+    // Normalize source text for complex types to ignore formatting differences.
+    function getSortText(
+      type: TSESTree.TypeNode | TSESTree.TypeElement,
+    ): string {
+      if (type.type === AST_NODE_TYPES.TSTypeLiteral) {
+        const properties = type.members
+          .map(member => getSortText(member))
+          .join('; ');
+
+        return `{ ${properties} }`;
+      }
+
+      if (type.type === AST_NODE_TYPES.TSMappedType) {
+        return `{ [${context.sourceCode.getText(type.typeParameter)}]: ${context.sourceCode.getText(type.typeAnnotation)}`;
+      }
+
+      if (type.type === AST_NODE_TYPES.TSTupleType) {
+        const properties = type.elementTypes
+          .map(elementType => getSortText(elementType))
+          .join(', ');
+
+        return `[${properties}]`;
+      }
+
+      if (type.type === AST_NODE_TYPES.TSFunctionType) {
+        const params = type.params
+          .map(param => context.sourceCode.getText(param))
+          .join(', ');
+
+        return `(${params}) => ${type.returnType ? getSortText(type.returnType.typeAnnotation) : 'void'}`;
+      }
+
+      return context.sourceCode.getText(type);
+    }
+
     function checkSorting(
       node: TSESTree.TSIntersectionType | TSESTree.TSUnionType,
     ): void {
@@ -178,6 +213,7 @@ export default createRule<Options, MessageIds>({
         return {
           group: group === -1 ? Number.MAX_SAFE_INTEGER : group,
           node: type,
+          sortText: getSortText(type),
           text: context.sourceCode.getText(type),
         };
       });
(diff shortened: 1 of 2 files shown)


ericmatthys:
The rule has since been renamed, but I did find this issue with the old name: https://github.com/typescript-eslint/typescript-eslint/issues/3639

JoshuaKGoldberg:
Aha, great, thanks for finding that! Re-opening this PR and renaming existing issues under the old name.

That brings up a good point that we should probably always rename old issues when we rename a rule. Linking: #8548

JoshuaKGoldberg:
Note that we're feature-freezing `sort-type-constituents` and other naming & sorting stylistic rules: #8792. Since this PR was sent before #8792 was filed, we'll still accept it. We'd hate to mark work as accepted, then pull it out from under you after you've done the work. 🙂

ericmatthys:
👍 I'll take a look at https://eslint-plugin-perfectionist.azat.io/ as well. I'm curious how it handles these cases. I won't be bothered if the PR is rejected either way, so no worries there.

packages/eslint-plugin/tests/rules/sort-type-constituents.test.ts
JoshuaKGoldberg, line 1: [Testing] Could you add in a few quick tests for `//` and `/*` style comments, just to make sure they're factored in appropriately?

packages/eslint-plugin/src/rules/sort-type-constituents.ts
```diff
@@ -170,6 +170,41 @@ export default createRule<Options, MessageIds>({
       numeric: true,
     });
 
+    // Normalize source text for complex types to ignore formatting differences.
+    function getSortText(
+      type: TSESTree.TypeNode | TSESTree.TypeElement,
+    ): string {
```
JoshuaKGoldberg, line 176: [Bug] As convenient as `sourceCode.getText` is, most cases where it's used are susceptible to bugs. Here's an example for type parameters in mapped types:

```ts
type _ = {
  [P in 'a' /* now gets included */ | 'b']: string
}
```

This function will need to account for all the AST types that might have comments / formatting differences, included `TSUnionType`, `TSIntersectionType`, and `TSTypePredicate`.

JoshuaKGoldberg (changes requested):
The approach here seems reasonable to me. But in order to avoid a slew of edge case bugs, it'll need to be more comprehensive in which types it accounts for. Have fun! ✨

JoshuaKGoldberg:
👋 ping @ericmatthys, ust checking in - is this still something you have time for?

JoshuaKGoldberg:
That makes it easier on our end. Either way, thanks for the work so far! ❤️

Issue #8652, opened 2024-03-12
Bug: [no-unnecessary-condition] False positive with nested optional chains if nullability is inferred via type parameters


cairomassimo:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gDwFzwAoAeAFQD4C94SBKeAXjOroB94UBXCCAbgCg%2B2AISUaAfgB0owZNGzsNPkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK%2BSpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false

### Repro Code

```TypeScript
declare const x: (<T>(x: T) => T) | null;

x!(x)?.(x)
x?.(x)?.(x)
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/no-unnecessary-condition": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true
  }
}
```

### Expected Result

_No errors should be reported._

According to the types specified:
- the expression `x!(x)` can be `null`,
- the expression `x?.(x)` can be `null` or `undefined` (in the latter case, causing optional chaining short-circuit).

In both `x!(x)?.(x)` and `x?.(x)?.(x)`, the first part of the expression can be `null` (and not trigger short-circuit), so the second optional chaining operator is not redundant in either case.

### Actual Result

A spurious error "Unnecessary optional chain on a non-nullish value." is reported on the second optional chain in the expression `x?.(x)?.(x)` (but not in `x!(x)?.(x)` - which should be indistinguishable).

[...]

PR #8662, opened 2024-03-13
fix(eslint-plugin): [no-unnecessary-condition] false positives with optional chaining

cairomassimo:
Partial mitigation for #8652.

Avoids some false positives with nested optional chaining, e.g.

```ts
declare const x: (<T>(x: T) => T) | null;
declare const y: (() => void) | null

x?.(y)?.()
```

## PR Checklist

- [x] Addresses an existing open issue: #8652

## Overview

This PR changes the way an expression `a` is tested for nullability, inside an optional chain such as `a?.b`, as follows:

- if TypeScript says `a` is either `any`, `unknown`,  `null` or `void`, then `a` is considered nullable,
- otherwise, the previous logic is used as-is.

The reason why we cannot test for `undefined` as well (possibly replacing the second step entirely), is that `undefined` is always present if the expression is part of a nested optional chain, such as `a?.b` in `a?.b?.c`.

Ideally, in this case, we would like to ask TypeScript the type of `a?.b` _after any narrowing resulting from optional chain short-circuit_, which would exclude `undefined` (unless the member `b` itself can be `undefined`), and solve all false positives. This is not (yet) covered by this PR.

Note that the following still results in a false positive (with a `null` replaced with `undefined`):

```ts
declare const x: (<T>(x: T) => T) | null;
declare const y: (() => void) | undefined

x?.(y)?.()
```

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -636,6 +636,23 @@ export default createRule<Options, MessageId>({
 
     function isOptionableExpression(node: TSESTree.Expression): boolean {
       const type = getConstrainedTypeAtLocation(services, node);
+
+      // We would like to test for `undefined` here as well,
+      // but it's always present in case of nested optional chaining, resulting in false positives.
+      // TODO(#8652): Is there a way to get the type for TS, after optional-chain short-circuiting?
+      if (
+        isTypeFlagSet(
+          type,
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
Since the linked behavior is still in discussion (and looks like it might not be something we can ?), marking this as a draft to keep off the review queue for now. Cheers!

JoshuaKGoldberg:
Closing as #8652 was closed out. Cheers!

PR #8671, opened 2024-03-14
fix: correct `engines.node` constraints in `package.json`

MichaelDeBoey:
These seem to be forgotten in #8377


bradzacher (approved):
what a silly mistake I made!
Thanks for catching this

Technically this is a breaking change - but given this breaking change was _supposed_ to land with v7 I'm okay with landing it now.
If I wasn't lazy I'd backport it as a patch to 7.0, 7.1, and 7.2 to make sure it's applied for all released minors... but I'm lazy so just the latest will do.

PR #8673, opened 2024-03-14
fix(eslint-plugin): [no-unnecessary-template-expression] add missing parentheses in autofix

### LINKED ISSUE #8669: Bug: [no-unnecessary-template-expression] Auto fix will change meaning of code

### Playground Link

https://typescript-eslint.io/play#ts=5.3.0-beta&fileType=.tsx&code=MYGwhgzhAECC0G8BQ1oQC5nQS2NAZgPYBOAtlgBQZYCuEAXItAHZikCmjGx2zA5tAA%2B0GswAm7fL3ZjoAXyEjxk6WICUXdD36LREqcxmIUqaMXboaxZmkyWIAfgB0rDtAfQABgBIE1ey5s7ELCAOShcp5OWtikFGrQjHoqhmIA3CZySHJAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYkACAXAngA4oDG0AlobgLQrzkB2uA9AwPbWzKJLLLW5EAW0LwAhoOr1B0MfFQZIiaNDbRIAXxAagA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYwAviNUA&tokens=false

### Repro Code

```TypeScript
class A {
  static format(status: { name: string | undefined } | undefined): string | undefined {
    return status?.name ? `${status.name || ''}`.trim() : undefined;
  }
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {"@typescript-eslint/no-useless-template-literals": "error"}
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {}
}
```

### Expected Result

```ts
class A {
  static format(status: { name: string | undefined } | undefined): string | undefined {
    return status?.name ? (status.name || '').trim() : undefined;
  }
}
```

I think the autofixer should have added parens when it removed the template literal. 

### Actual Result

```ts
class A {
  static format(status: { name: string | undefined } | undefined): string | undefined {
    return status?.name ? status.name || ''.trim() : undefined;
  }
}
```

### Additional Info

This seems to repro in Playground with or without `strictNullChecks`. I'm my project, I was not using `strictNullChecks`.

The autofix highlights that the original code is silly. 

I tested this also with minimal eslint config, and it still happens.

Edit: I changed Playground link to be most minimal rule set & tsConfig.

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8669

## Overview

Fix no consider precedence when fixer function is excution using util function.


JoshuaKGoldberg (changes requested):
I like the approach here of using `getWrappingFixer`! 👏 

I also think the code can be refactored a bit to remove the extra layer of indirection & the `as`. Requesting changes on that please. :)

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -92,21 +93,16 @@ export default createRule<[], MessageId>({
...
-                ]),
-              ];
+            fix(fixer): TSESLint.RuleFix {
+              const replaceResult = getWrappingFixer({
+                sourceCode: context.sourceCode,
+                node: node.expressions[0],
+                wrap: (...code: string[]) => {
+                  return code.join('');
+                },
+              })(fixer) as TSESLint.RuleFix;
```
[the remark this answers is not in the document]
developer-bandi: Based on what you mentioned, I separated it into a helper function.

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -7,12 +7,32 @@ import {
...
+  node: TSESTree.TemplateLiteral,
+  fixer: TSESLint.RuleFixer,
+): string | null {
+  const replaceResult = getWrappingFixer({
+    sourceCode: context.sourceCode,
+    node: node.expressions[0],
+    wrap: (...code: string[]) => {
+      return code.join('');
+    },
+  })(fixer);
```
auvred, line 28: [Refactor] Coming from https://github.com/typescript-eslint/typescript-eslint/pull/8673#discussion_r1537842742:

> Instead of adding the overhead of the rest of `getWrappingFixer()`, suggestion: the specific text generation stuff should be extracted out and used here. Like, you could move the text generation from the `getWrappingFixer` into a separate function, and use that separate helper here.

I think this means that we should extract codegen part **from** `getWrappingFixer` implementation to the separate utility function, not extract whole `getWrappingFixer` call to the separate function. WDYT?

auvred (changes requested):

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
developer-bandi: I think I misunderstood the explanation. thank you The code generation part that generates parentheses was separated and made into a function.

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
```diff
@@ -207,7 +207,7 @@ ruleTester.run('no-useless-template-literals', rule, {
 
     {
       code: noFormat`\`\${    'a' + 'b'    }\`;`,
-      output: `'a' + 'b';`,
+      output: `('a' + 'b');`,
```
kirkwaiblinger, line 210: How come this test case gains parens? Is that just the `getWrappingFixer` being overly cautions?

packages/eslint-plugin/src/util/getWrappingFixer.ts
```diff
@@ -27,51 +27,59 @@ interface WrappingFixerParams {
 
 /**
  * Wraps node with some code. Adds parenthesis as necessary.
- * @returns Fixer which adds the specified code and parens if necessary.
+ * @returns Code which adds the specified code and parens if necessary.
  */
-export function getWrappingFixer(
-  params: WrappingFixerParams,
-): TSESLint.ReportFixFunction {
+export function getWrappingCode(params: WrappingFixerParams): string {
```
kirkwaiblinger, line 32: One thing worth noting is that this uses the parent node to infer whether parens are needed (see `if (isWeakPrecedenceParent(node))`). So I'm wondering if that could give mistaken results for a scenario where we're reparenting code?

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
JoshuaKGoldberg: I also have this question. Feels like a bug to me?

packages/eslint-plugin/src/util/getWrappingFixer.ts
developer-bandi: I understand that you are using the above node to infer whether parentheses are needed, but I don't understand this part (So I'm wondering if that could give mistaken results for a scenario where we're reparenting code?). Can you tell me in more detail what the scenario is for reparenting the code?

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
developer-bandi: At first, I thought it would be good for the expression to be wrapped, but as you said, I think it's excessive and looks like a bug.

However, in order to restore the test code, the getWrappingFixer function must be modified. Would it be a good idea to modify this function?

packages/eslint-plugin/src/util/getWrappingFixer.ts
kirkwaiblinger: Sorry for taking a bit to get back to you - took me a while to figure out what I meant 😆 . Basically, `getWrappingFixer()` allows you to transform a node whose parent did not change.
```txt
// input
<stuff before> nodeToBeModified <stuff after>
// output
<stuff before> modifiedNode <stuff after>
```
But, _now that getWrappingCode has been extracted_, it can be used instead like 
```txt
// input
<stuff before> nodeToBeModified <stuff after>
// getWrappingCode output
modifiedNode
// used in fixer as
<DIFFERENT stuff before> modifiedNode <DIFFERENT stuff after>
```
In fact, that's the whole reason that you've extracted it in this PR.
```ts
// input
condition ? `${'left' || 'right'}`.trim() : 'lol';
// getWrappingCodeOutput
('left' || 'right')
// used in fixer as 
condition ? ('left' || 'right').trim() : 'lol';
```
So, `getWrappingCode` needs to deduce whether to wrap the expression in parentheses based on the _NEW_ parent, not the existing one.

Used within the `getWrappingFixer` those were the one and the same, so the line `if (isWeakPrecedenceParent(node))` was justified, but _now that getWrappingCode may be used indepdently_, we'd need something like `if (isWeakPrecedenceNewParent(node, newParent))`, in order for the function to give expected results _in general_. 

I'd have to have a longer think to come up with test cases where this discrepancy would manifest, but hopefully this gets you started? If it's still unclear let me know and I can think up some test cases. The bottom line is that we need to be aware of whether parens are needed for the _new_ parent, not the _existing_ parent, in order for `getWrappingCode` to be safe for general use outside of `getWrappingFixer`.

kirkwaiblinger (changes requested):

packages/eslint-plugin/src/util/getWrappingFixer.ts
kirkwaiblinger: this might well be related actually: https://github.com/typescript-eslint/typescript-eslint/pull/8673/files/5a9f63085e1a3b08ae1c63e29deb34ca93b5165d#r1573616588

developer-bandi: I apologize for not clear question. But the answers were the best!

If my understanding is correct, the parent that the `isWeakPrecedenceNewParent` function checks must be the new parent. In fact, the previous code was checking the parent of the expression, so it was always checking the templete literal node as the parent, so it was not getting normal results.

However, if i slightly modify the code to satisfy the above conditions, i will get the following results depending on the conditions of innerNode and parents condition.

```
// before
true ? `${'test' || ''}`.trim() : undefined;
// after
true ? (('test' || '')).trim() : undefined;
```
In my opinion, it seems better to use parentheses when both the logic of applying parentheses to innerNode and the logic of examining the state of the parent are satisfied. so i revert `getWrappingFixer` and make new `getWrappingCode` function 

I think we're just getting started, so i appreciate your feedback.

kirkwaiblinger:
FYI @developer-bandi - This will a semantic merge conflict with https://github.com/typescript-eslint/typescript-eslint/pull/8821, which has now been merged. When this is ready to go, you'll want to duplicate the fix to the `no-unnecessary-template-expression` copy of the rule

Also, I'm sorry for taking so long to do another review pass. I'm setting an intention for myself to get another pass in this week. Please bug me if I don't get to it within that time!

kirkwaiblinger (changes requested):
I spent a long time playing around with this on my machine.... this is a _difficult_ area of the code. Requesting a few changes for clarity, but I think the essential logic is in place! 🙂 

Also, would like to tag in another member of @typescript-eslint/triage-team to help make sure we're on the right track.

packages/eslint-plugin/src/util/getWrappingFixer.ts
```diff
@@ -75,6 +75,23 @@ export function getWrappingFixer(
   };
 }
 
+export function getWrappingCode(params: {
```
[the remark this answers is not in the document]
developer-bandi: It looks much more intuitive and good. We will reflect it appropriately

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -92,21 +93,14 @@ export default createRule<[], MessageId>({
...
-                  node.expressions[0].range[0],
-                ]),
-
-                fixer.removeRange([
-                  node.expressions[0].range[1],
-                  nextQuasi.range[0] + 2,
-                ]),
-              ];
+            fix(fixer): TSESLint.RuleFix | null {
+              const wrappingCode = getMovedNodeCode({
```
kirkwaiblinger, line 97: (nit) the name `wrappingCode` is slightly stale with the rename, but I don't have many good ideas for a better name 🙃. Any ideas? `codeToReplaceTemplate`? Up to you if you want to change this or leave it 🤷‍♂️

kirkwaiblinger (approved):
teamwork

kirkwaiblinger (changes requested):
Ah - reminder to duplicate these changes to the `no-unnecessary-template-expressions` copy of the rule. Otherwise I think I'm happy!

JoshuaKGoldberg:
👋 @developer-bandi just checking in, is this something you still have time and energy for?

developer-bandi:
@JoshuaKGoldberg  I haven't been able to proceed with the work recently due to lack of time. I plan to start working on it this weekend.

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
developer-bandi: Since the function name is getMovedNodeCode, how about changing the variable name to movedNodeCode? If you don't think it's appropriate, it might be a good idea to keep the current name.

developer-bandi:
copy work to [no-unnecessary-template-expression] and delete no-useless-template-literals to resolve conflict

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
kirkwaiblinger: Makes sense to me 👍

packages/eslint-plugin/src/util/getWrappingFixer.ts
```diff
@@ -74,6 +74,33 @@ export function getWrappingFixer(
...
+  destinationNode: TSESTree.Node;
+}): string {
+  const { sourceCode, nodeToMove: existingNode, destinationNode } = params;
+  const code = sourceCode.getText(existingNode);
+  if (isStrongPrecedenceNode(existingNode)) {
+    // Moved node never needs parens
+    return code;
+  }
+
+  if (!isWeakPrecedenceParent(destinationNode)) {
```
JoshuaKGoldberg, line 96: [Praise] Great reuse of the existing functions!

JoshuaKGoldberg (approved):
Nice!!

PR #8685, opened 2024-03-15
fix(eslint-plugin): [prefer-optional-chain] properly disambiguate between `boolean` and `false`

### LINKED ISSUE #8659: Docs: [prefer-optional-chain] improve the playground and examples for allowPotentiallyUnsafeFixesThatModifyTheReturnTypeIKnowWhatImDoing

### Suggested Changes

I have a hard time following all the words in the section on `allowPotentiallyUnsafeFixesThatModifyTheReturnTypeIKnowWhatImDoing`, so I tried using the "Open in Playground" button for this section. It... doesn't really work, though.
1. There are no type errors! The playground opens with an empty tsconfig, so we don't have "strict"/"strictNullChecks" specified to true.
2. I can't figure out how to trigger the report that's implied in the doc. Even after enabling strictNullChecks, there is no report in the snippet, so I copied the two code snippets I saw in writing (`!foo || foo.bar` and `foo?.bar`) into the playground and neither of them reports either.

### Affected URL(s)

https://typescript-eslint.io/rules/prefer-optional-chain/#allowpotentiallyunsafefixesthatmodifythereturntypeiknowwhatimdoing

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8659

## Overview
In #8659, we see that [the example in the docs](https://typescript-eslint.io/rules/prefer-optional-chain/#allowpotentiallyunsafefixesthatmodifythereturntypeiknowwhatimdoing) for `prefer-optional-chain`: `allowPotentiallyUnsafeFixesThatModifyTheReturnTypeIKnowWhatImDoing` is not reporting an error.

The root cause is actually _unrelated_ to `allowPotentiallyUnsafeFixesThatModifyTheReturnTypeIKnowWhatImDoing` — it is because `foo.bar` is of type `boolean`.

In the rule, there is a bypass to avoid flagging unions with falsy members, such as
```
declare const foo: { bar: boolean } | boolean | null | undefined;
foo && foo.bar;
```
since changing that to an optional chain would cause a TS error.

However, if the falsy literal is the last operand, then we _should_ report it. This is the change made by this PR.
```
declare const foo: { bar: boolean } | null | undefined;
foo && foo.bar;
```

### DIFF
--- a/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
+++ b/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
@@ -84,8 +84,10 @@ function isValidFalseBooleanCheckType(
     narrows out the non-nullish falsy cases - so converting the chain to `x?.a`
     would introduce a build error
     */
+    const booleanLiteralTypes = types.filter(isBooleanLiteralType);
     if (
-      types.some(t => isBooleanLiteralType(t) && t.intrinsicName === 'false') ||
+      (booleanLiteralTypes.length === 1 &&
+        booleanLiteralTypes[0].intrinsicName === 'false') ||
       types.some(t => isStringLiteralType(t) && t.value === '') ||
       types.some(t => isNumberLiteralType(t) && t.value === 0) ||
       types.some(t => isBigIntLiteralType(t) && t.value.base10Value === '0')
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (approved):
Clever - I like the changes. Very minimal and clean. Nicely done! 🙌

PR #6929, opened 2023-04-18
chore(website): improve error handling

armano2:
## PR Checklist

- [x] Addresses an existing open issue: fixes #000
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Add missing error handling for unexpected errors during parsing

see:
- v6: https://v6--typescript-eslint.netlify.app/play/#ts=4.9.3&sourceType=module&code=OQMw9mwHQLYIYBcDGALAFMA2vANuATjAKYAmwAlANwBQQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wls4DmiAB7F6wxGXRRe0APbRI4MAF8QKoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

- main: https://typescript-eslint.io/play/#ts=4.9.3&sourceType=module&code=OQMw9mwHQLYIYBcDGALAFMA2vANuATjAKYAmwAlANwBQQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wls4DmiAB7F6wxGXRRe0APbRI4MAF8QKoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA
[...]

PR #8686, opened 2024-03-16
chore(website): simplify playground logic

### LINKED ISSUE #7630: Repo: Playground code refactor: Playground.tsx state switches

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7630

## Overview

I took a stab at simplifying the playground code a bit...open to feedback and further discussion 👍🏻


packages/website/src/components/Playground.tsx
```diff
@@ -168,57 +155,56 @@ function Playground(): React.JSX.Element {
...
+                    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                    astModel.storedTsAST!
+                  }
                   onHoverNode={setSelectedRange}
                   cursorPosition={position}
                 />
-              )) ||
-              (state.showAST && astModel && (
+              ) : (
                 <ASTViewer
```
armano2, line 189: that looks ok to me.

-----

i was meaning to rewrite this at some point to use provider, passing events back and forward make all this code so unreadable and requires us to use stuff like this

packages/website/src/components/Playground.tsx
```diff
@@ -168,57 +155,56 @@ function Playground(): React.JSX.Element {
...
-            )) ||
-              (state.showAST === 'types' && astModel?.storedTsAST && (
+            ) : state.showAST && astModel ? (
+              state.showAST === 'types' ? (
                 <TypesDetails
                   typeChecker={astModel.typeChecker}
-                  value={astModel.storedTsAST}
+                  value={
+                    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                    astModel.storedTsAST!
```
armano2, line 183: this is not correct, when eslint crashes without reporting ast, this value may not be defined,

eg, null pointer, invalid syntax in rule
ref: https://github.com/typescript-eslint/typescript-eslint/pull/6929

abrahamguo: thanks for the clarification!

packages/website/src/components/Playground.tsx
```diff
@@ -115,46 +105,43 @@ function Playground(): React.JSX.Element {
...
-            tsconfig={state.tsconfig}
-            eslintrc={state.eslintrc}
-            sourceType={state.sourceType}
-            showAST={state.showAST}
             onASTChange={setAstModel}
             onMarkersChange={setMarkers}
             selectedRange={selectedRange}
             onChange={setState}
-            onLoaded={onLoaded}
+            onLoaded={useCallback((ruleNames, tsVersions) => {
```
JoshuaKGoldberg, line 140: [Refactor] Err, let's not put `use*` hooks inside components. The [React rules of hooks](https://react.dev/reference/rules/rules-of-hooks) get very hard to keep track of when hooks are called inline.

JoshuaKGoldberg (changes requested):
This looks a bit cleaner to me 🙂 thanks!

I think the root intent behind #7630 was to make the `Playground.tsx` component a bit more split up even than this. But there's not much velocity in this area of code so I don't think it's super important that we keep iterating. This code reduction makes me happy.

Just requesting the one change on not calling the `useCallback` hook inline.

Laila James opening a door and saying "this is nice"

JoshuaKGoldberg (approved):
🧹 thanks!

Issue #8554, opened 2024-02-26
[rule-tester] support multipass fixes


bradzacher:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Relevant Package

utils

### My proposal is suitable for this project

- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Description

https://github.com/eslint/eslint/issues/18007 was rejected as a breaking change enhancement.

Given this impacts us directly (eg https://github.com/typescript-eslint/typescript-eslint/issues/8251) we should build this into our fork.

I propose the following changes:
1) if the test case has `output` and the `output` value is a string and the test case requires more than 1 pass to fix then the test case fails with a message like "test requires multiple fixes to resolve due to overlapping fix ranges - please use the array form of `output` to declare all of the expected fix passes".
2) accept an array for `output` which declares the expected fix passes for a test case.
    - An array of size 0 is disallowed.
    - An array of size 1 is equivalent to the string form.
    - An array of size `n` asserts that there are exactly `n` fix passes
    - The rule tester asserts that after each  fix pass the output matches the `n`th element of the array.

bradzacher:
For me it's a debugging thing.
I remember we had some bugs with one of our rules (array-type) around how multipass fixes worked and we had to setup separate tests to help us debug and validate it.

It's about being explicit and acknowledging that the user will have two fixes applied because your rule defines overlapping ranges. This may be an important thing for perf reasons.

For example imagine switching out `Array<Array<T>>` for `T[][]`. One fix solution would be:
```ts
// fix one
  Array<Array<T>>
//^^^^^^^^^^^^^^^ replace with Array<T>[]

// fix two
  Array<Array<T>>
//      ^^^^^^^^ replace with T[]
```

But that solution is heavy-handed and has overlapping fix ranges so will always require two fix passes.

One might not consider this problem if they don't think about the fact that the test has overlapping fixes.

bradzacher:
[...]

PR #8688, opened 2024-03-16
fix(eslint-plugin): [no-unnecessary-template-expression] do not render escaped strings in autofixes

### LINKED ISSUE #8677: Bug: [no-unnecessary-template-expression] Inconsistent handling of escapes and quotes in autofixes.

### Playground Link

[click here](https://typescript-eslint.io/play/#ts=5.4.2&fileType=.ts&code=AYEg3gRAOgrgDHAogVggX2AbgFAHpcAEAhjAC4D2AZgJYAeApgM4EUF6HTxKo7bAHhOCFOgJZ2xMlTpMW5NvjEEAp2MxA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tmUSWWS18iALbF4AQ2G1Gw6BPioMkRNGgdokADThskTt179B5CaXRRV6zboC%2BIW0A&tsconfig=&tokens=false)

### Repro Code

```TypeScript
`${"\u00E5"}`;
// autofixes to 
// "\u00E5";

` ${"\u00E5"} `;
// autofixes to 
// ` å `;
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-useless-template-literals": "error",
    "no-useless-escape": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

I expect that the autofixed strings should preserve the string type and the escaped characters as the author wrote them.
So, we should have 
```typescript
`${"\u00E5"}`;
// autofixes to 
// `\u00E5`;

` ${"\u00E5"} `;
// autofixes to 
// ` \u00E5 `;
```

### Actual Result

actual autofix results change user's input regarding template and and escaped characters.

### Additional Info

This can also result in multiline template literals instead of single line literals that include a `\n`. 

It doesn't impact correctness at runtime, but it isn't the purview of this rule to make these stylistic changes (with the caveat, of course, that obviously `$`s and backticks still need to be escaped in order to incorporate a string literal into a template string)

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8677

## Overview

Addresses many issues with escaping of `` ` `` and `${`. To better understand what's going on here, I'd suggest reading the comments in the code.

**NOTE:** I changed the base branch for this PR to `v8` since this PR uses multipass fixes - https://github.com/typescript-eslint/typescript-eslint/pull/8883


packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -144,10 +145,11 @@ export default createRule<[], MessageId>({
                 ]),
               ];
 
-              const stringValue = getStaticStringValue(expression);
-
-              if (stringValue != null) {
-                const escapedValue = stringValue.replace(/([`$\\])/g, '\\$1');
+              if (isLiteral(expression)) {
+                const escapedValue =
```
kirkwaiblinger, line 149: I think you may want to do the escaping in both branches of the ternary, since non-strings may still stringify to something with characters needing escapes. Consider these sneaky test cases:
```typescript
`the nonsense regex ${/${}/} matches the end of a line followed by a pair of braces`;
`the regex ${/\${}/} matches a dollar sign followed by a pair of braces`;
```

kirkwaiblinger:
good stuff here! 

In addition to the comment, one thing though that I want to flag is that in the issue it was also intended that 
```typescript
`'a'`; 
// should autofix to 
`a`;
// rather than 
'a';
```
though it's possible that that flew under the radar on the issue itself as well when it was marked accepting PRs.

I also note that https://github.com/typescript-eslint/typescript-eslint/issues/8669 somewhat conflicts with that change (in that it would be made redundant).

Could be worth flagging there, and/or dropping the "a" -> `a` change from https://github.com/typescript-eslint/typescript-eslint/issues/8677 entirely and mentioning that explicitly on the issue. 

Not strongly opinionated to which decision is made but just think we should document it.

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
auvred: What a nice catch! I overlooked the fact that we can stringify regexes... Thanks!

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
auvred, line 1: > In addition to the comment, one thing though that I want to flag is that in the issue it was also intended that
> 
> ```ts
> `${'a'}`; 
> // should autofix to 
> `a`;
> // rather than 
> 'a';
> ```

@kirkwaiblinger, Ohh, I missed that in the original issue!

I'm 50/50 on this!

> though it's possible that that flew under the radar on the issue itself as well when it was marked accepting PRs.

Maybe..

Fixing to string literal vs template literal sounds a bit stylistic to me, I think it would be great to hear more opinions on this!

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -144,10 +145,11 @@ export default createRule<[], MessageId>({
...
+                  typeof expression.value === 'string'
+                    ? expression.raw.slice(1, -1).replace(/([`$])/g, '\\$1')
```
kirkwaiblinger, line 151: (nit, optional) - why are the escaping patterns different for each branch? Maybe worth a comment?

kirkwaiblinger (dismissed):

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
kirkwaiblinger: Yeah, agreed. I will say, at the time, I didn't realize it was just removing the wrapping template syntax, as opposed to overriding whatever the user had typed with one particular quote style. So my motivation for that change has softened a lot as well. It still feels a little surprising to me, but it does have a very justifiable internal logic as-is. 

So unless someone comes in and has a strong opinion, I'd say, just leave it as-is :)

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
auvred: Yeah, it does seem a bit confusing

Does https://github.com/typescript-eslint/typescript-eslint/pull/8688/commits/e07e1460765fe2f28fe26fbf6bde967412d84759 clear it up?

kirkwaiblinger: oof, i think i get it, but i still struggle tbh 😆 It's just really tricky to me to figure out which things are in escaped-land and which things are non-escaped in the code comment there (it's even hard for me to type out the special characters in this comment 😆 ). Is the idea 
raw string &rArr; 
   1. Removing enclosing quotation characters
   1. just add escapes to characters that need escape in template string, but didn't as a string literal (` and $)
   
cooked string &rArr;
   1. escape everything that needs escapes (`, $, and \\)
   
I really think a few English words would be helpful in the code comments (to go along with the examples), since my brain is not big enough to make meaning of all the special characters at a glance 😄

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
```diff
@@ -637,5 +691,79 @@ declare const nested: string, interpolation: string;
...
+    {
+      code: "` ${'👨‍👩‍👧‍👦'} `;",
+      output: '` 👨‍👩‍👧‍👦 `;',
+      errors: [{ messageId: 'noUselessTemplateLiteral' }],
+    },
+    {
+      code: "` ${'\\ud83d\\udc68'} `;",
+      output: '` \\ud83d\\udc68 `;',
+      errors: [{ messageId: 'noUselessTemplateLiteral' }],
+    },
```
kirkwaiblinger, line 767: I've thought of a few cruel test inputs that generate syntax errors on autofix. 

https://deploy-preview-8688--typescript-eslint.netlify.app/play#ts=5.4.2&fileType=.tsx&code=PTAEFMDcEsBtQK4DtwGcDGBDADuAJqHgPayyYBOoq0A5kqAPpJEAuDiq%2Bo09L4AttjJ8IAD2zk01IkgBQAAwAkigN4ByFQF81m%2BQG5ZskKAngAtGiy4CAI0zoA1i2iOFoVWoA68naH1A&eslintrc=N4KABGBEBOCuA2BTAzpAXGUFIAEAuAngA4oDG0AlkXgLQrwUB2eA9IwPY2zKJLLI08iALZF4AQyE0GQ6OPioMkRNGjtokAL4hNQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

kirkwaiblinger: note - For the second case, i think i've seen in other contexts that a strategy for this is to only add an escape if the character needing escape is preceeded by an even number (possibly 0) of backslashes. less sure about the '$' shenanigans.

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -144,10 +145,19 @@ export default createRule<[], MessageId>({
...
-              if (stringValue != null) {
-                const escapedValue = stringValue.replace(/([`$\\])/g, '\\$1');
+              if (isLiteral(expression)) {
+                const escapedValue =
+                  typeof expression.value === 'string'
+                    ? // '1'   -> 1
+                      // '`'   -> \`
+                      // '${}' -> \${}
+                      // '\\'  -> \\
+                      expression.raw.slice(1, -1).replace(/([`$])/g, '\\$1')
```
Josh-Cena, line 155: Should we only escape `${` instead of all `$`? Also what if these characters are already escaped?

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -144,10 +145,19 @@ export default createRule<[], MessageId>({
...
+                    : // 1     -> 1
+                      // /`/   -> /\`/
+                      // /${}/ -> /\${}/
+                      // /\\/  -> /\\\\/
+                      String(expression.value).replace(/([`$\\])/g, '\\$1');
```
Josh-Cena, line 160: Same here. I also wonder if we shouldn't do the template literal syntax escaping separately from this ternary to avoid some duplication.

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
Josh-Cena: Oops, I see @kirkwaiblinger has come up with the exact same ideas! My suggestion for the "already-escaped" case is to do something like `(?<=(\\\\)*)`.

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
auvred: What a nice catch! Thanks!

auvred:
Thanks for the reviews! I tried to cover all the possible edge cases (although I've likely missed something).

Currently this PR is blocked by https://github.com/typescript-eslint/typescript-eslint/issues/8554, because some test cases, like this one below, require multiple fixer passes:

```ts
` ${'$'}${''}{} `;
```

So I'm going to convert this PR do a draft until https://github.com/typescript-eslint/typescript-eslint/issues/8554 issue is resolved.

auvred:
https://github.com/typescript-eslint/typescript-eslint/issues/8554 solved! I will revisit and mark this PR as ready for review in a little while! Marking it as `awaiting response` for myself

kirkwaiblinger:
FYI @auvred - This will a semantic merge conflict with https://github.com/typescript-eslint/typescript-eslint/pull/8821, which has now been merged. When this is ready to go, you'll want to duplicate the fix to the `no-unnecessary-template-expression` copy of the rule

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
JoshuaKGoldberg: Yeah +1, I think _ideally_ it wouldn't change the syntax, but that's a small implementation detail and not worth doing a bunch of work for. If folks care they should lint for it separately.

packages/eslint-plugin/tests/rules/no-unnecessary-template-expression.test.ts
```diff
@@ -162,6 +162,48 @@ ruleTester.run('no-unnecessary-template-expression', rule, {
...
+          messageId: 'noUnnecessaryTemplateExpression',
+          line: 1,
+          column: 4,
+          endColumn: 8,
+        },
+      ],
+    },
+    {
+      code: '`${0b1010} ${0b1111}`;',
+      output: '`10 15`;',
```
bradzacher, line 179: part of me wonders if we should not handle these weird number representation cases and just leave them as is.

eg I could see more value in writing code with a bunch of variations on the binary string to ensure completeness vs having the converted int val.

for example
```ts
`${0b00} ${0b01} ${0b10} ${0b11}` 
```
encodes information better (i.e. that all binary cases are represented) rather than
```ts
`0 1 2 3`
```
Obviously this is a simplified case but one could imagine some complex binary sequence represented (eg power of two or something).
Same reason that it's not uncommon for people to define flag enums as
```
enum Flag {
  A = 1 << 0,
  B = 1 << 1,
  C = 1 << 2,
  D = 1 << 3,
  E = 1 << 4,
}

// rather than

enum Flag {
  A = 1,
  B = 2,
  C = 4,
  D = 8,
  E = 16,
}
```

probably just put a pin in it and leave it as a potential future exercise

bradzacher (approved):
I wonder - do you think it'd be worth having a 2nd pass test here which does like
```ts
cases.invalid.forEach(
	testCase =>
		expect(eval(testCase.code)).toEqual(eval(testCase.output))
)
```

to  really validate that we don't regress the value transform?
probably not worth the effort but definitely an interesting thought...

auvred:
> I wonder - do you think it'd be worth having a 2nd pass test here which does like
> 
> ```ts
> cases.invalid.forEach(
> 	testCase =>
> 		expect(eval(testCase.code)).toEqual(eval(testCase.output))
> )
> ```
> 
> to really validate that we don't regress the value transform? probably not worth the effort but definitely an interesting thought...

Wow, I was waiting for someone to mention this! That's exactly what I did locally, and it helped catch a lot of errors, but I didn't see this being common practice in typescript-eslint, so I decided not to commit it! I'll be sure to add those tests!

packages/eslint-plugin/tests/rules/no-unnecessary-template-expression.test.ts
auvred: Interesting point, yea, I can imagine someone would want their binary/hex numbers to remain untouched. But this PR already has a lot of changes beyond the original issue. So I'd say we should file a follow-up issue or wait for someone to report it.

> probably just put a pin in it and leave it as a potential future exercise

:+1: +1 here

bradzacher (approved):

kirkwaiblinger (approved):
What a complicated problem this turned out to be!

I've spent some amount time in the playground testing out confusing test cases...
1. They all passed
1.  When I looked closer, you already had an equivalent for each of them in your added test cases. Great job on test thoroughness!

My only complaint is that the big `fixableExpressions.forEach()` callback has gotten huge (~100 lines - in fairness, many of which are comments) and _really_ complicated (order-dependent control flow, `context.report()` side effects, local functions, and several different behaviors based on node type). 

Now, to be sure, a fair amount of that complexity may be irreducible, since... this is just a complex problem. So, I'm ok with merging as is (especially in light of your excellent test coverage 🙂). But if you think that there is room for some simplification, I would certainly welcome it as a reader. It does look like at a minimum the regex `evenNumOfBackslashesRegExp` and local function `endsWithUnescapedDollarSign` could be taken out of the callback, since don't use any variables or state from the closure.

Most importantly, though, once again great job in tackling this really complex bug!

kirkwaiblinger:
Oh, FYI - please, before merging, double check whether this will cause heinous merge conflicts with #8673 or not. If so, maybe wait for that to go through, since it's targeted to main rather than v8.

kirkwaiblinger:
Actually - there is one _oddity_ stemming from the fact that the fixes are mutually dependent, though, I don't think I'd consider it a bug. If you're using the editor, you can apply fixes separately, rather than all at once... So, given
```ts
`${'$'}${'{'}`
   ~~~   ~~~
```
if we fix _the first error only_ in the editor, we'll get
```ts
`\$${'{'}`
     ~~~
```
even though 
```ts
`$${'{'}`
    ~~~
```
would have been fine. 

Note that 
```ts
`${'$'}${''}`
   ~~~   ~~
```
fixes the first error to
```ts
`$${''}`
    ~~
```

I think that's an acceptable quirk that will most likely go completely unnoticed by users. Just noting it here to leave a bradcrumb unless some _really_ obscure bug somehow arises in relation to this.

auvred:
> My only complaint is that the big `fixableExpressions.forEach()` callback has gotten huge (~100 lines - in fairness, many of which are comments) and _really_ complicated (order-dependent control flow, `context.report()` side effects, local functions, and several different behaviors based on node type).

Yes, I can imagine how difficult this is to read for someone who hasn't struggled with this code for a decent amount of time, but I don't see any way to simplify this code :(

---

> Oh, FYI - please, before merging, double check whether this will cause heinous merge conflicts with #8673 or not. If so, maybe wait for that to go through, since it's targeted to main rather than v8.

Fortunately, #8673 changes a part of the rule that wasn't touched in this PR, so I think there should be no problem with the merge!

JoshuaKGoldberg (approved):
:shipit:

PR #2813, opened 2020-11-25
feat(eslint-plugin): [naming-convention] add `requiresQuotes` modifier

bradzacher:
Fixes #2761
Fixes #1483

This modifier simply matches any member name that requires quotes.
To clarify: this does not match names that have quotes - only names that ***require*** quotes.

PR #2207, opened 2020-06-11
fix(eslint-plugin): [switch-exhaustiveness-check] handle special characters in enum keys

karishnu:
Fixes #2181

PR #8691, opened 2024-03-16
test(type-utils): add unit tests for requiresQuoting

### LINKED ISSUE #8358: Repo: Add unit tests inside type-utils
### Suggestion

`packages/type-utils` provides a suite of types/TypeScript-related functions. They're used in many rules. But because we don't merge coverage reports across packages (#6701), it shows up as having surprisingly low unit test coverage. E.g. [`builtinSymbolLikes.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/1aa8664bab1f7ea77539ec83f65123e8ebe34b39/packages/type-utils/src/builtinSymbolLikes.ts) on 1d7447f4fd5e2fbae4211eb5fc07e93cf570b6b3 right now has 0%. https://app.codecov.io/gh/typescript-eslint/typescript-eslint/blob/1d7447f4fd5e2fbae4211eb5fc07e93cf570b6b3/packages%2Ftype-utils%2Fsrc%2FbuiltinSymbolLikes.ts

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes (partially) #8358

## Overview

add Testcase in requiresQuoting Function

### DIFF
--- a/packages/type-utils/tests/requiresQuoting.test.ts
+++ b/packages/type-utils/tests/requiresQuoting.test.ts
@@ -17,23 +17,23 @@ describe('getDeclaration', () => {
     it('start with dollorSign', () => {
       const name = '$bar';
       const result = requiresQuoting(name);
-      expect(result).toBe(true);
+      expect(result).toBe(false);
     });
     it('include dollorSign not start position', () => {
       const name = 'foo$bar';
       const result = requiresQuoting(name);
-      expect(result).toBe(true);
+      expect(result).toBe(false);
     });
 
     it('start with underScore', () => {
       const name = '_bar';
       const result = requiresQuoting(name);
-      expect(result).toBe(true);
+      expect(result).toBe(false);
     });
     it('include underScore not start position', () => {
       const name = 'foo_bar';
       const result = requiresQuoting(name);
-      expect(result).toBe(true);
+      expect(result).toBe(false);
     });
 
     it('non-alphabet characters', () => {
--- a/packages/type-utils/tests/requiresQuoting.test.ts
+++ b/packages/type-utils/tests/requiresQuoting.test.ts
@@ -1,0 +1,84 @@
+import { requiresQuoting } from '../src';
+
+describe('getDeclaration', () => {
+  describe('valid identifier', () => {
+    it('upper and lower case alphabet', () => {
+      const name = 'fooBar';
+      const result = requiresQuoting(name);
+      expect(result).toBe(false);
+    });
+
+    it('upper and lower case alphabet with number not first', () => {
+      const name = 'foo1234Bar';
+      const result = requiresQuoting(name);
+      expect(result).toBe(false);
+    });
+
+    it('start with dollorSign', () => {
+      const name = '$bar';
+      const result = requiresQuoting(name);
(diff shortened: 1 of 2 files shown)


packages/type-utils/tests/requiresQuoting.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      const result = requiresQuoting(name);
+      expect(result).toBe(false);
+    });
+  });
+
+  describe('invalid identifier', () => {
+    it('only number', () => {
+      const name = '12345';
+      const result = requiresQuoting(name);
+      expect(result).toBe(true);
```
developer-bandi, line 50: I think this test case return true is weird. 

because I don't think quotation marks are necessary if the name of the property variable consists entirely of numbers.

If the function is incorrect in returning true for this test case, we will create a separate issue and remove the test case.

kirkwaiblinger: Ooh, good eyes 👀. Yeah I guess it depends how `requiresQuoting` is intended to be used.... 

You can _define_ a property starting with a number, but you can't _access_ it that way using dot notation, whereas you can with the other special characters here. 
```ts
const objectWithCrapPropertyNames = {
  12345: "legal syntax!",
  $foo: "foo",
  _bar: "bar",
}

// syntax error
objectWithCrapPropertyNames.12345
// works
objectWithCrapPropertyNames.$foo
// works
objectWithCrapPropertyNames._bar
```

This was extracted to a util to support the naming-convention requiresQuotes option (https://github.com/typescript-eslint/typescript-eslint/pull/2813)... but existed before that as a utility in switch-exhaustiveness check for autofixing with code that performs member access, in which case the quoting _is_ required on the number (https://github.com/typescript-eslint/typescript-eslint/pull/2207).

That it can handle numbers in this way is actually explicitly tested for:
https://github.com/typescript-eslint/typescript-eslint/blob/82cb9dd580f62644ed988fd2bf27f519177a60bd/packages/eslint-plugin/tests/rules/switch-exhaustiveness-check.test.ts#L1861-L1897

So I'd say it probably needs to work the way it currently does for its intended usages to work - but maybe you can leave a comment to clarify that, either on the test cases or the function itself?

kirkwaiblinger (approved):
Very neat! Thanks for doing this.

FYI - Since this only addresses one of the files within the folder mentioned in the issue, I'm changing the description to say "fixes (partially) <ticket-number>", otherwise [this will autoclose the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). 🙂

packages/type-utils/tests/requiresQuoting.test.ts
developer-bandi: Thank you for telling us about the history. I completely understand why the test case is needed.

Regarding comments, I added them because I thought it would be good to have appropriate comments for both functions and test code. It would be a good idea to check them out.

kirkwaiblinger (approved):

JoshuaKGoldberg (approved):
🆒

PR #8742, opened 2024-03-21
docs: add homepage

moskalakamil:
## Overview

Added homepage to `package.json`.

After the merge, the homepage on [npmjs.org](https://www.npmjs.com/package/@typescript-eslint/eslint-plugin) will be replaced with the official documentation.

screen_of_npmjs.org


Josh-Cena (approved):

package.json
```diff
@@ -22,6 +22,7 @@
   "bugs": {
     "url": "https://github.com/typescript-eslint/typescript-eslint/issues"
   },
+  "homepage": "https://typescript-eslint.io/",
```
JoshuaKGoldberg, line 25: This is super nitpicky, but - there's no need for the trailing slash 😄. Most existing files don't use it.  

```suggestion
  "homepage": "https://typescript-eslint.io",
```

JoshuaKGoldberg (changes requested):
I ... don't think that this will change the actual `package.json`s that get published with packages? `@typescript-eslint/eslint-plugin`, for example, comes from `packages/eslint-plugin/package.json`. I believe you'll need to make this change in all those nested `package.json`s instead.

Someone please yell at me if there's actually Nx magic under the hood I don't know of.

moskalakamil:
@JoshuaKGoldberg done

JoshuaKGoldberg (approved):
LGTM, thanks!

I don't have a personal preference of which ones should point to the root website vs. a `/packages/*`. If anybody else does then I suppose they can request changes. 😄

PR #8752, opened 2024-03-22
feat(eslint-plugin): [array-type] detect `Readonly<string[]>` case

### LINKED ISSUE #7755: Enhancement: [array-type] Detect `Readonly<string[]>`

https://typescript-eslint.io/rules/array-type

### Description

The rule should also turn `Readonly<string[]>` into `readonly string[]`.

### Fail

```typescript
const y: Readonly<string[]> = ['a', 'b'];
```

### Pass

```typescript
const y: readonly string[] = ['a', 'b'];
```

### Additional Info

_No response_

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7755

## Overview

In the case of `Readonly<string[]>`, the same as `Readonly Array<string>`, an error is displayed and then changed to `readonly string[]`.

However, on the contrary, `readonly string[]` is converted to `ReadonlyArray<string>` as before.


packages/eslint-plugin/tests/rules/array-type.test.ts
auvred, line 1: [Testing] Could you revert the commented test cases?

auvred (changes requested):

packages/eslint-plugin/tests/rules/array-type.test.ts
```diff
@@ -254,18 +254,18 @@ ruleTester.run('array-type', rule, {
     },
     {
       code: `
-function fooFunction(foo: Array<ArrayClass<string>>) {
-  return foo.map(e => e.foo);
-}
-      `,
```
JoshuaKGoldberg, line 260: [Style] Sorry if I missed this before - could you please revert any unrelated changes, such as differences in indentation levels? It's harder to review the PR when there's a lot of irrelevant noise in the diff.

JoshuaKGoldberg (dismissed):
I'll defer to @auvred as they already started reviewing. Just fly-by leaving a couple of requests. Cheers!

packages/eslint-plugin/src/rules/array-type.ts
```diff
@@ -199,13 +202,20 @@ export default createRule<Options, MessageIds>({
...
+            node.typeName.name === 'Readonly'
           )
         ) {
           return;
         }
 
-        const isReadonlyArrayType = node.typeName.name === 'ReadonlyArray';
+        const isReadonlyWithGenericArrayType =
+          node.typeName.name === 'Readonly' &&
+          node.typeArguments?.params[0].type === AST_NODE_TYPES.TSArrayType;
```
auvred, line 214: [Bug] If the type name is `Readonly` and first type argument is not an array, we should return from the function. Because otherwise any `Readonly<...>` type will be reported.

```ts
const x: Readonly<string> = 'a'
```

[playground](https://deploy-preview-8752--typescript-eslint.netlify.app/play/#ts=5.4.2&fileType=.ts&code=MYewdgzgLgBAHgLhgJQKYEMAm4A2BPAHmgCcBLMAcwD4YBeGAcnQaA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AQ2mncNqNPRREXAPbRI4MAF8QUoA&tsconfig=&tokens=false)

packages/eslint-plugin/tests/rules/array-type.test.ts
```diff
@@ -1378,7 +1382,7 @@ interface ArrayClass<T> {
           messageId: 'errorStringArraySimple',
           data: { className: 'Array', readonlyPrefix: '', type: 'T' },
           line: 3,
-          column: 8,
+          column: 12,
         },
```
auvred, line 1386: [Tests] Could you revert these changed `column`s? These changes don't seem to be related to this PR.

auvred (dismissed):

developer-bandi:
Some of the current CI test cases are failing.

Since it is not failing locally, could you please help me resolve it?

packages/eslint-plugin/docs/rules/array-type.mdx
```diff
@@ -58,6 +59,7 @@ const y: readonly string[] = ['a', 'b'];
 ```ts option='{ "default": "generic" }'
 const x: Array<string> = ['a', 'b'];
 const y: ReadonlyArray<string> = ['a', 'b'];
+const z: Readonly<string[]> = ['a', 'b'];
```
kirkwaiblinger, line 62: re https://github.com/typescript-eslint/typescript-eslint/pull/8752#issuecomment-2093144419
(One of) the CI failure(s) is due to this line being flagged, despite being in the "Correct" tab. (you can try it for yourself in the deploy preview by clicking the "Open in playground" button)
```
 FAIL  tests/docs.test.ts (25.146 s, 827 MB heap size)
  ● Validating rule docs › array-type.mdx › code examples ESLint output
    assert(received)
    Expected value to be equal to:
      true
    Received:
      false
    
    Message:
      Expected not to contain lint errors:
    const x: Array<string> = ['a', 'b'];
    const y: ReadonlyArray<string> = ['a', 'b'];
    const z: Readonly<string[]> = ['a', 'b'];
```
Shoutout to @auvred's for his work in https://github.com/typescript-eslint/typescript-eslint/issues/8382 / https://github.com/typescript-eslint/typescript-eslint/pull/8497/files#diff-62842d0740c234c02c26a4ccd3b3fe920b872cff1479f0703c6ec44f9cfb2f3dR407 that caught this!

developer-bandi: thank you for your advise! 
스크린샷 2024-05-11 오후 11 22 13

but there is one problem. i use incorrect example`const z: Readonly<string[]> = ['a', 'b'];
` but lint cannot use this type. Is there no way?

kirkwaiblinger: Hi, sorry for the slow response.

I'm not really familiar with this change, and, I apologize, but I'm having a bit of a hard time understanding what you mean in the question... 

I'm wondering if the issue is that, now that the snapshot has been set, the snapshot tests are failing due to the examples being changed? If so, you may just need to run `cd eslint-plugin; yarn test schemas -u` (or similar); see https://github.com/typescript-eslint/typescript-eslint/issues/8997.

Let me know if that resolves your issue or if the problem is something else! 🙂

kirkwaiblinger (commented):
ignore this comment; just publishing a review to make our "awaiting response" label automation work right

packages/eslint-plugin/docs/rules/array-type.mdx
kirkwaiblinger: BTW - looking closer at the screenshot, i think you might also need to run `yarn lint-fix` to fix the existing violations throughout the repo of the new cases that this rule reports.

developer-bandi: The problem was that I had to apply the changed lint rule(array-type) to another file. Thank you!

bradzacher (approved):
This will sadly false-negative on a case like this:
```ts
type T = string[];
type U = Readonly<T>;
```

However that seems like it is okay for two reasons:
1) you can't convert `Readonly<T>` to `readonly T` - you would need to do `readonly T[number][]` which is just ew.
2) we would need type info to even begin to catch it - defs not worth adding that to the rule.

----

This LGTM - thanks for adding this!
This has been a pet peeve of mine for a while.

kirkwaiblinger (approved):
reading

PR #8754, opened 2024-03-22
docs: fix example of binding unbound method

victma:
## PR Checklist

## Overview

This fixes the  "Correct usage" example of the `unbound-method` rule.

My understanding is that the example tries to show how to bind the correct scope to the method. So the method that should be used is `logUnbound`.

Using `.bind` or an arrow function on `logBound` is useless, as it is already bound with the correct scope.

:sauropod:


JoshuaKGoldberg (approved):
Nice catch, thanks!

PR #8758, opened 2024-03-24
fix(eslint-plugin): [no-floating-promises] handle TaggedTemplateExpression

naruaway:
## PR Checklist

## Overview

This PR makes "no-floating-promises" rule to raise error when floating promise is created from a tagged templates, which is essentially a function call in this context.
The motivating example is with [dax](https://github.com/dsherret/dax), which is a popular library to write shell script in TypeScript.
Its `$` API is a tagged template and it returns a Promise (with other methods on it) and we need to `await` on that.
If we forget to `await`, execution orders will be messed up and the main script might exit too early.

After this PR, "no-floating-promises" will be able to capture the mistake in the following snippet:

```typescript
import $ from 'dax-sh'

// Here we are forgetting `await` and , "no-floating-promises" cannot detect it without the PR
$`echo hello`
$`echo world`
```

I added the corresponding test cases in the PR and confirmed that the test fails before the implementation change in this PR.

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
@@ -504,6 +504,18 @@ void promiseArray;
 ['I', 'am', 'just', 'an', 'array'];
       `,
     },
+    {
+      code: `
+declare const myTag: (strings: TemplateStringsArray) => Promise<void>;
+myTag\`abc\`.catch(() => {});
+      `,
+    },
+    {
+      code: `
+declare const myTag: (strings: TemplateStringsArray) => string;
+myTag\`abc\`;
+      `,
+    },
   ],
 
   invalid: [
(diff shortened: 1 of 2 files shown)


Josh-Cena (commented):
I don't know if they work differently but could you add tests for ``tag`aaa`.then``, ``tag`aaa`.catch``, and ``tag`aaa`.finally?``

naruaway:
@Josh-Cena thanks for the review, I just added test cases for then/catch/finally for sure. I think adding these makes sense to make sure the logic is the same as the function call

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
JoshuaKGoldberg, line 1: [Testing] Could you add a `valid` case too? That a tagged template expression which doesn't create a floating Promise is allowed?

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -593,6 +593,30 @@ doSomething();
...
       ],
     },
+    {
+      code: `
+declare const myTag: (strings: TemplateStringsArray) => Promise<void>;
+
+myTag\`abc\`;
+myTag\`abc\`.then(() => {});
+myTag\`abc\`.catch(() => {});
+myTag\`abc\`.finally(() => {});
```
JoshuaKGoldberg, line 603: [Testing] Per https://typescript-eslint.io/contributing/local-development#granular-unit-tests, we generally ask that tests are granular and single-purpose. Could you split this up into one test per error please?

JoshuaKGoldberg (changes requested):
We normally ask that PRs (other than small docs fixes) have a backing issue for discussion. But this one is pretty small and straightforward. Just requesting changes on tests. Thanks! 🏷️

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
naruaway: Sure, I can separate cases but I just thought this is granular enough. For example, existing test case is for function call is [this](https://github.com/naruaway/typescript-eslint/blob/d5f77d4dd299cac41cb50f3952a0a06b1f8262ea/packages/eslint-plugin/tests/rules/no-floating-promises.test.ts#L509-L537):
```typescript
async function test() {
  Promise.resolve('value');
  Promise.resolve('value').then(() => {});
  Promise.resolve('value').catch();
  Promise.resolve('value').finally();
}
```

naruaway: Now I separated the test cases in 7c8c9dda34b16e98ecfb0adf2561fc4defed6a48. Not bad as I thought 👍

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
naruaway: Thanks for the review, now I added two valid cases in 7c8c9dda34b16e98ecfb0adf2561fc4defed6a48

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
JoshuaKGoldberg: Yeah a lot of not-new rules have tests that aren't quite the same as our standards 😅. Thanks!

JoshuaKGoldberg (approved):
Older woman on Antiques Roadshow happily clapping and saying 'this is wonderful'

auvred (approved):

PR #8765, opened 2024-03-25
feat(eslint-plugin): [no-misused-promises] check subtype methods against heritage type methods

### LINKED ISSUE #8739: Enhancement: [no-misused-promises] Check when an abstract method implementation returns a promise instead of void

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.4&showAST=es&fileType=.tsx&code=FBA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
abstract class A {
  thing: number | null = null;

  abstract setThing(): void;

  setAndPrintThing(): void {
    this.setThing();
    console.log(this.thing);
  }
}

class B extends A {
  // Expected no-misused-promises error here
  async setThing(): Promise<void> {
    const thing = await fetchThing();
    this.thing = thing;
  }
}

async function fetchThing(): Promise<number> {
  return await new Promise<number>((resolve) => {
    setTimeout(() => {
      resolve(5);
    }, 1000);
  });
}

const b = new B();

// Prints null instead of 5
b.setAndPrintThing();

function _setAndPrintThing(setThing: () => void): void {
  setThing();
  console.log(b.thing);
}

// Correctly triggers no-misused-promises error:
//  "Promise returned in function argument where a void return was expected"
_setAndPrintThing(b.setThing.bind(this));
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/no-misused-promises": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

Line 14 (where `B.setThing` is implemented) should show an error like "Promise returned in abstract method implementation where a void return was expected"

### Actual Result

There's no error on line 14

### Additional Info

I'm not sure why the playground link doesn't seem to be working, but hopefully you can see my code sample anyway and it makes sense.

Based on the [`no-misused-promises`](https://typescript-eslint.io/rules/no-misused-promises/) page it looks like maybe this should be another category under [`checksVoidReturn`](https://typescript-eslint.io/rules/no-misused-promises/#checksvoidreturn)?

alythobani:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8739

## Overview

This PR adds a new `heritageTypes` option (under `checksVoidReturn`) to the [`no-misused-promises`](https://typescript-eslint.io/rules/no-misused-promises/) rule, defaulted to `true`. If enabled, it checks `ClassDeclaration` / `ClassExpression` / `InterfaceDeclaration` nodes for any methods that return `Promise`s, and checks those methods against any matching methods in their declared `extends ...` / `implements ...` heritage types. Error outputted when a heritage type has the given method declared/implemented as returning `void` instead of a `Promise`.

(The issue #8739  I filed only mentioned abstract classes/methods, but I realized the topic is more generalizable to heritage types in general.)

Note that this is my first PR here and first time diving into this repo / the Typescript source code. So hopefully I didn't miss any cases to test / handle, but it's possible I did! There did end up being small edge cases I realized weren't handled as I iterated and had to tweak things. E.g., calling `checker.getTypeAtLocation(heritageTypeExpression)` ended up being necessary in order to properly check against a `ClassExpression` heritage type's members; `checker.getTypeFromTypeNode(heritageTypeExpression)` worked for all other heritage types, just not class expressions.


packages/eslint-plugin/docs/rules/no-misused-promises.mdx
```diff
@@ -102,6 +102,7 @@ The following options are supported:
 - `attributes`: Disables checking an asynchronous function passed as a JSX attribute expected to be a function that returns `void`
 - `properties`: Disables checking an asynchronous function passed as an object property expected to be a function that returns `void`
 - `returns`: Disables checking an asynchronous function returned in a function whose return type is a function that returns `void`
+- `subtypes`: Disables checking an asynchronous method in a type whose extended/implemented heritage type expects that method to return `void`
```
alythobani, line 105: At first called this `methods` but figured `subtypes` might be a bit more specific to the case we're handling. But maybe there's a better name we could use for this

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -73,14 +77,16 @@ export default createRule<Options, MessageId>({
...
       voidReturnProperty:
         'Promise-returning function provided to property where a void return was expected.',
       voidReturnReturnValue:
         'Promise-returning function provided to return value where a void return was expected.',
-      voidReturnAttribute:
-        'Promise-returning function provided to attribute where a void return was expected.',
+      voidReturnSubtype:
+        "Promise-returning method provided where a void return was expected by heritage type '{{ heritageTypeName }}'.",
+      voidReturnVariable:
+        'Promise-returning function provided to variable where a void return was expected.',
```
alythobani, line 89: Other enumerations over the `checksVoidReturn` options were sorted alphabetically, so just went ahead and did a swap here (attribute <=> variable) to maintain that consistent order

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
```diff
@@ -500,6 +500,462 @@ foo(bar);
       },
       options: [{ checksVoidReturn: { attributes: true } }],
     },
+    // #region checksVoidReturn.subtypes
```
alythobani, line 503: Hope the amount of test cases here wasn't too overkill. I actually did find them useful for catching random cases (e.g. type intersections, conditional types, class expressions) that weren't properly handled by my original implementation(s).

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
```diff
@@ -1275,5 +1731,389 @@ consume(...cbs);
...
+
+interface MyAsyncInterface extends MyGenericType<false> {
+  setThing(): Promise<void>;
+}
+      `,
+      errors: [
+        {
+          line: 7,
+          messageId: 'voidReturnHeritageType',
+          data: { heritageTypeName: '{ setThing(): void; }' },
```
alythobani, line 1949: I wasn't fully sure about including the `data` in this test case, since it depends on the specific styling produced by `checker.typeToString(heritageType)`, which actually happens to differ from eslint styling (the semicolon is added).

But if it's unlikely to change then it's maybe fine / nice to keep in as a check that this case is formatted nicely (vs e.g. one of my early iterations had `heritageTypeName` outputted as `"__type"` for a case like this).

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -369,6 +381,66 @@ export default createRule<Options, MessageId>({
...
+      const heritageTypes = getHeritageTypes(checker, tsNode);
+      if (heritageTypes === undefined || heritageTypes.length === 0) {
+        return;
+      }
+
+      for (const nodeMember of tsNode.members) {
+        const memberName = nodeMember.name?.getText();
+        if (memberName === undefined) {
+          continue;
+        }
```
JoshuaKGoldberg, line 401: [Testing] Either...

* this is a case that can be hit, and so should be covered by tests
* this is a case that can't be hit, and the type system should be told as such so the `if` can be removed

I haven't dug in deep enough to know which it is.

(here and elsewhere)

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
JoshuaKGoldberg: IME as long as the tests are hand-written and not auto-generated, they're almost always not overkill. Nice coverage!

JoshuaKGoldberg (changes requested):
Looks like a great start, nice! I haven't reviewed to deeply though, just looking at missing test coverage in the file. LMK if I'm off base though!

JoshuaKGoldberg:
> I realized the topic is more generalizable to heritage types in general

+1, this is a nice spot!

packages/eslint-plugin/src/rules/no-misused-promises.ts
alythobani: Hey Josh, thanks for the review :)

[Sorry for the long writeup incoming, but here are my thoughts/findings]

I've been looking into this and as far as I can tell, `nodeMember.name` will only ever be `undefined` for call signatures, construct signatures, and index signatures.

### Call signatures
Functions can have multiple call signatures, and types seem to be allowed in TS to have conflicting (overriding?) and duplicate call signatures. Although instantiating the type without TS erroring may require an `any` assertion?
```ts
interface A {
  (): number;
  (): Promise<number>;  // no error
  (): number;  // no error
}
interface B extends A {
  (): boolean; // no error
}
const b: B = () => true;  // Type '() => boolean' is not assignable to type 'B'
const b2: B = (): any => "true";  // no error (other than @typescript-eslint/no-explicit-any)
```
In the context of void- / promise-returning call signatures, I can in theory imagine something like:
```ts
type MySyncCallable = (arg: number) => void;
interface MySometimesAsyncCallable extends MySyncCallable {
  (arg: boolean): Promise<void>;
}
const myCallable: MySometimesAsyncCallable = (arg: number | boolean): any => {
  if (typeof arg === 'number') {
    return;
  }
  return Promise.resolve();
};

myCallable(3); // void
myCallable(true); // Promise<void>
```
Although I'm not sure this is the type of design decision we should be supporting lol / if anyone would even want to do something like this, with the `extends` clause and all. In any case it's a bit unclear to me what the best path is here, and maybe the case of Promise-void-mismatched subtype / heritage type call signatures is niche/unlikely enough that we can just leave it? But please let me know if you have a better grasp of this case and what the best approach is.

### Construct signatures

Constructors can't be async or return void AFAIK so we can just continue to the next `member` if the current one doesn't have a `name` due to being a construct signature.

### Index signatures

Unlike call signatures, TS doesn't seem to allow conflicting index signatures. So this seems more clear to me in terms of expected behavior, and relevant to `no-misused-promises` / `checksVoidReturn` since TS will complain about mismatched index signatures _unless the heritage type index signature is a void-returning function_.

```ts
interface ThisIsNotAllowed {
  [key: string]: () => string;
  [key: `key${string}`]: () => Promise<void>; // () => Promise<void> is not assignable to () => string
}

interface ThisIsAllowedButShouldntBe {
  [key: string]: () => void;
  [key: `key${string}`]: () => Promise<void>;
}
```
For a more full example we probably don't want to allow this:
```ts
type MySyncRecord = Record<`set${string}`, () => void>;
interface MyAsyncRecord extends MySyncRecord {
  someField?: number;
  [key: `set${string}`]: () => Promise<void>;  // this shouldn't be allowed
}

const myAsyncRecord: MyAsyncRecord = {
  setFoo: () => Promise.resolve(),
};

function setFooSync(mySyncRecord: MySyncRecord): void {
  mySyncRecord.setFoo();
}
setFooSync(myAsyncRecord);
```

but currently no errors are shown for the above code (TS or eslint errors).

In the context of `checkClassLikeOrInterfaceNode` when `nodeMember.name === undefined` I haven't actually been able to find a good way to parse the `nodeMember` to get its `keyType` and compare it against the heritage types' index signatures compatible with `keyType`. (I feel like it should be possible in theory, but just not sure what API functions to be using to accomplish that.) But if we wanted we could also just check index signatures all at once separately, something like:

```ts
function checkHeritageTypeIndexSignatures(
  tsNode: ts.ClassLikeDeclaration | ts.InterfaceDeclaration,
  heritageTypes: ts.Type[],
): void {
  const tsNodeType = checker.getTypeAtLocation(tsNode);
  const nodeIndexInfos = checker.getIndexInfosOfType(tsNodeType);
  for (const indexInfo of nodeIndexInfos) {
    if (!anySignatureIsThenableType(checker, tsNode, indexInfo.type)) {
      continue;
    }
    for (const heritageType of heritageTypes) {
      const heritageIndexInfos = checker.getIndexInfosOfType(heritageType);
      for (const heritageIndexInfo of heritageIndexInfos) {
        if (!checker.isTypeAssignableTo(indexInfo.keyType, heritageIndexInfo.keyType)) {
          continue;
        }
        if (!isVoidReturningFunctionType(checker, tsNode, heritageIndexInfo.type)) {
          continue;
        }
        context.report({
          node: services.tsNodeToESTreeNodeMap.get(tsNode),
          messageId: 'voidReturnHeritageTypeMethod',
          data: {
            heritageTypeName: checker.typeToString(heritageType),
          },
        });
      }
    }
  }
}
```

JoshuaKGoldberg: > ```ts
> [key: `set${string}`]: () => Promise<void>;
> ```
> ...
> the type of design decision we should be supporting lol

Haha yeah that code is funky. But it _is_ something we need to support. Lint rules generally don't impart judgement on code practices unrelated to their specific purpose. `no-misused-promises` isn't about funky index signatures and someone might legitimately be writing that code. So ideally this rule would support that.

...that being said, I don't know of how to support mapping from `setFoo` to <code>[key: \`set${string}\`]</code> without https://github.com/typescript-eslint/typescript-eslint/discussions/7936. So this might be blocked.

I'll defer to someone else from @typescript-eslint/triage-team if they have time to dig in.

alythobani: > > ```ts
> > [key: `set${string}`]: () => Promise<void>;
> > ```
> > 
> > ...
> > the type of design decision we should be supporting lol
> 
> Haha yeah that code is funky. But it is something we need to support. Lint rules generally don't impart judgement on code practices unrelated to their specific purpose. no-misused-promises isn't about funky index signatures and someone might legitimately be writing that code. So ideally this rule would support that.

Right, yeah good point on not going beyond the specific purpose of the lint rule.

Just to clarify a bit here, it was only the funky call signatures (not index signatures) I wasn't sure we should support—i.e. "should we throw an error if the subtype has any Promise-returning call signatures, when the heritage type only specifies void-returning call signatures?". Technically it could be valid (as in `MySometimesAsyncCallable`), so I think/agree we should probably allow it, and not raise any errors. But the couple potential reasons I was thinking about for raising an error would be:

* I wasn't sure if it was possible to actually use `MySometimesAsyncCallable` without casting to `any` somewhere anyway (going against `@typescript-eslint/no-explicit-any`). But as you point out, our lint rule should probably stay in scope, so that may be irrelevant. (Also I experimented more and it can be done with an `as unknown as _` instead of `any`)
* When we check functions in other contexts for `no-misused-promises` (e.g. [property assignments](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-misused-promises.ts#L276-L291)), we already iterate through call signatures in a way that would disallow this funky `MySometimesAsyncCallable` stuff:
  * If the [node has at least one Promise-returning call signature](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-misused-promises.ts#L674C1-L682C2), and the [contextual type has at least one void-returning call signature](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-misused-promises.ts#L647-L672) and no Promise-returning ones, then we throw an error. So by our current logic, you already wouldn't be able to pass in a `MySometimesAsyncCallable` to a property declared as a `MySyncCallable`, even though it could technically be legal (it could still have the void-returning signatures on it).

Another example showing how we already disallow the funky call signatures:

```ts
type A = () => void;
interface B extends A {
  (asyncArg?: number): Promise<void>;
}
function myWeirdFunction(asyncArg?: number): Promise<void> | void {
  if (asyncArg) {
    return Promise.resolve();
  }
  return;
}

// No TS error, but we report the `no-misused-promises` `voidReturnProperty` error message
const myWeirdA: A = myWeirdFunction;  
```

I think mainly because of this second reason (`no-misused-promises` already disallows the assignment of `MySometimesAsyncCallable` as a `MySyncCallable`) I'm still a bit split on what to do:
1. Ignore call signatures like in `MySometimesAsyncCallable` since "incompatible" call signatures are valid in TS, at least at the type-level before they're actually instantiated
2. Report call signatures like in `MySometimesAsyncCallable` since we already report `no-misused-promises` when a `MySometimesAsyncCallable` is assigned to a `MySyncCallable`

Or a third option, given that we may be blocked on checking index signatures without #7936 (thanks for noticing + calling that out):

3. Ignore all signatures (call or index) and have the main purpose of the new `heritageTypes` option explicitly be to check over named methods only
    * We can just add in a couple test cases with call/index signatures that we expect no errors on, for the code coverage

I think I lean towards 3 on the whole but very open to thoughts. I think we'd already get most of the benefit of the `heritageTypes` option by checking over named members only.

alythobani:
Pushed some small changes to add test coverage for unnamed methods (signatures) and add comments in the spot where we ignore them, in case that's the approach we want to take :). Also not sure why GitHub's diff for the test file got all messed up, but the changes in that file should all just be additions (which is what's shown locally in my git).

packages/eslint-plugin/src/rules/no-misused-promises.ts
JoshuaKGoldberg: Look, I'll be honest with you, we're pretty busy this month (ESLint 9, our v8, other things) and I don't have the time to read through two >500 words on this subject. This might sit for a while.

If you can simplify your findings to something much more scannable that would help, but I understand if you don't have time for that either. 

> “I didn't have time to write a short letter, so I wrote a long one instead.”
> ― Mark Twain

alythobani: Hey Josh, no worries, I understand. First time contributing here so I wasn't sure about the balance between sharing too much/little.

The edge case, and my suggestion, can be simplified to:
* `memberName === undefined` for unnamed methods, which are either: **(i) Call signatures**, **(ii) Index signatures**, or **(iii) Construct signatures**
  1.  **Call signatures**: TypeScript allows conflicting overloaded function signatures (e.g. `() => number` and `() => Promise<number>`), so we shouldn't care about checking mismatches here either (`continue`)
  2. **Construct signatures**: It's not possible for a constructor to be async, so we can also safely ignore these (`continue`)
  3. **Index signatures**: These would be meaningful to check. But to do that properly, we'd want to use `isTypeAssignableTo` (to know which heritage index signatures to check the `member` against). So for this PR, we can just ignore index signatures (`continue`) and leave a comment about implementing index-signature checking once `isTypeAssignableTo` is available.

TLDR suggestion: `continue` past all unnamed methods, specify in the docs that `heritageTypes` only checks named methods, and leave a code comment that we can check index signatures in the future once `isTypeAssignableTo` is available

JoshuaKGoldberg: ❤️ that sounds good to me, thanks for the summary!

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
JoshuaKGoldberg, line 1: [Docs] I like what you suggested in https://github.com/typescript-eslint/typescript-eslint/pull/8765#discussion_r1556259908 😄. Let's go with that. Requesting the addition of the limitations mention.

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
```diff
@@ -500,780 +500,1804 @@ foo(bar);
       },
       options: [{ checksVoidReturn: { attributes: true } }],
     },
-  ],
+    // #region checksVoidReturn.heritageTypes
```
JoshuaKGoldberg, line 503: Heh, we don't usually use `#region` tags. IMO the cases should be clear enough from their contents that we shouldn't need them. Regions also tend to get murky and poorly defined over time.

JoshuaKGoldberg (commented):
I had some time to do a deeper dive - and what you're saying makes sense. Thanks! I'm a bit heads-down on ESLint 9 work so removing my request & pinging @typescript-eslint/triage-team on this one. Cheers!

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
alythobani: Ok nice :)

> Requesting the addition of the limitations mention.

Oh I'd snuck that in there by specifying `asynchronous named method` rather than just `asynchronous method`. Do you think we should be more explicit here that we don't check index/construct/call signatures?

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
alythobani: Ah, yeah fair enough. I figured `#region`s aren't ideal but could be useful for such a huge list of cases like this, but yeah happy to remove the regions

alythobani:
@JoshuaKGoldberg  Nice thanks for finding the time! Good to hear the approach makes sense 

Is there anything else we need to do to tag the triage team? Your comment shows the tag in plain text, so I'm not sure if it worked or not.

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
JoshuaKGoldberg: Yeah. It's a surprising limitation if you don't have a good familiarity with the type checking APIs.

alythobani: Okay sounds good, I've made the doc more explicit now on this

packages/eslint-plugin/tests/docs-eslint-output-snapshots/no-misused-promises.shot
```diff
@@ -66,6 +66,18 @@ eventEmitter.on('some-event', async () => {
...
+  setThing(): void;
+}
+class MyClass implements MySyncInterface {
+  async setThing(): Promise<void> {
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Promise-returning method provided where a void return was expected by heritage type 'MySyncInterface'.
+    this.thing = await fetchThing();
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  }
+~~~
+}
```
alythobani, line 80: Couple small notes here:

1) It looks like the rule snapshot above for `EventEmitter` doesn't include an error message because it relies on `EventEmitter` being imported (which isn't included in the code sample). Should I go ahead and add the `import` to the code example and re-generate the snapshot to include the error? (I was able to do this locally, so it's just a matter of pushing that commit)
2) Took me a minute to realize what was going on with CI failing (find / look into #8497) and learn how to update this snapshot. Do we want to add a short subsection to [Local Development > Validating Changes > Tests](https://typescript-eslint.io/contributing/local-development#validating-changes) on generating snapshots with `yarn test -u` as needed?

I realize these may be out of scope so let me know what's best here :)

JoshuaKGoldberg:
> Your comment shows the tag in plain text

Heh, maybe a GitHub permissions oddity?

Screenshot of hovering of my triage team tag giving a GitHub hover card for the team

packages/eslint-plugin/tests/docs-eslint-output-snapshots/no-misused-promises.shot
JoshuaKGoldberg: For 1, #8942 - would be great if you / anybody had time to look into it!

For 2, I'm not sure 🤔. In general it's pretty standard Jest practice that snapshots get updated with `-u`.

auvred: > 2\. Do we want to add a short subsection to [Local Development > Validating Changes > Tests](https://typescript-eslint.io/contributing/local-development#tests) on generating snapshots with `yarn test -u` as needed?

Here it is - https://github.com/typescript-eslint/typescript-eslint/issues/8997

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -73,14 +77,16 @@ export default createRule<Options, MessageId>({
...
+      voidReturnHeritageType:
+        "Promise-returning method provided where a void return was expected by heritage type '{{ heritageTypeName }}'.",
       voidReturnProperty:
         'Promise-returning function provided to property where a void return was expected.',
       voidReturnReturnValue:
         'Promise-returning function provided to return value where a void return was expected.',
-      voidReturnAttribute:
-        'Promise-returning function provided to attribute where a void return was expected.',
+      voidReturnVariable:
+        'Promise-returning function provided to variable where a void return was expected.',
```
JoshuaKGoldberg, line 89: [Praise] I appreciate the alphabetization 😄

packages/eslint-plugin/src/rules/no-misused-promises.ts
JoshuaKGoldberg: Ok yes coming back to this! Confirmed, yes, agreed with your writeups. Let's mark the **iii. Index signatures** case as a followup for post-v8-launch. It'll be a great showcase of the type assignability API!

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -680,3 +757,27 @@ function returnsThenable(checker: ts.TypeChecker, node: ts.Node): boolean {
...
     .some(t => anySignatureIsThenableType(checker, node, t));
 }
+
+function getHeritageTypes(
+  checker: ts.TypeChecker,
+  tsNode: ts.ClassDeclaration | ts.ClassExpression | ts.InterfaceDeclaration,
+): ts.Type[] | undefined {
+  return tsNode.heritageClauses
+    ?.map(clause => clause.types)
+    .flat()
```
JoshuaKGoldberg, line 767: Nit:

```suggestion
    ?.flatMap(clause => clause.types)
```

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
```diff
@@ -100,6 +100,8 @@ The following options are supported:
 
 - `arguments`: Disables checking an asynchronous function passed as argument where the parameter type expects a function that returns `void`
 - `attributes`: Disables checking an asynchronous function passed as a JSX attribute expected to be a function that returns `void`
+- `heritageTypes`: Disables checking an asynchronous method in a type whose extended/implemented heritage type expects that method to return `void`
+  - **Limitation:** For now, this option only checks named methods: that is, call/construct/index signatures are ignored. Call signatures are not required in TypeScript to be consistent with one another, and construct signatures cannot be `async` in the first place. Index signature checking may be implemented in the future.
```
JoshuaKGoldberg, line 104: [Docs] This is a lot of seemingly-required reading for a limitation that most projects won't hit. It'd be better to move it later down into a `heritageTypes`-specific section.

This rule's docs are a little older than our standard and don't have a section for each of the sub-options... but this is a good opportunity to start one! Request: could you make a `checksVoidReturn` section that at mentions the sub-option names and explains `heritageTypes`?

I think explaining the other sub-options would be best as a follow-up issue. That we we don't bog this PR down with docs nits on `arguments`, `attributes`, etc.

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
```diff
@@ -500,780 +500,1768 @@ foo(bar);
       },
       options: [{ checksVoidReturn: { attributes: true } }],
     },
-  ],
+    {
+      // Valid void-returning class
```
JoshuaKGoldberg, line 504: Nit: rule tests have a bad habit of drifting away from their commented intent over time. The tests you're adding are pretty clear & apparent. I'd suggest removing these comments.

```suggestion
```

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
```diff
@@ -100,6 +100,8 @@ The following options are supported:
 
 - `arguments`: Disables checking an asynchronous function passed as argument where the parameter type expects a function that returns `void`
 - `attributes`: Disables checking an asynchronous function passed as a JSX attribute expected to be a function that returns `void`
+- `heritageTypes`: Disables checking an asynchronous method in a type whose extended/implemented heritage type expects that method to return `void`
```
JoshuaKGoldberg, line 103: [Naming] I don't love how this is the only option that mentions _"Types"_ in its name. It's redundant. But just _"heritage"_ on its own seems imprecise... hmm. Maybe, `heritageSignatures`? That's kind of wordy but 🤷 I can't think of anything more succinct. What do you think?

JoshuaKGoldberg (changes requested):
Ok! Sorry again for the long wait - this is a very good PR that I wanted to make sure we had the time to think through. The implementation is nice and clean and the tests are nice and thorough.

Just requesting changes on the `[Docs]` and `[Naming]` notes. 

SpongeBob happily sticking a many thumbs ups out of suddenly appearing hands out of his body. Caption: "Many thumbs up."

alythobani:
Aw thanks @JoshuaKGoldberg  I'm glad things look good! And thanks for the careful review, glad I have your eyes here to make sure things are implemented properly and cleanly.

I'm a bit swamped these days but will get back to addressing your comments and wrapping this up when I get the chance, hopefully soon!

packages/eslint-plugin/src/rules/no-misused-promises.ts
alythobani: Oh nice, thanks for the tip

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
alythobani: Yeah that makes sense, it'd be nice to have separate sections for each suboption.

It looks like the explanations for `checksVoidReturn` and `checksSpreads` were accidentally placed under the `## Examples` section instead of `## Options`. So I've just gone ahead and moved them up in the doc under `## Options` (and turned each of these `checksVoidReturn.<suboption>` bullet points into their own subsections). But let me know if you had/have something else in mind here!

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
alythobani: Ah ok, the massive list of test cases without sections/etc still scares me but I trust you 😁  removed!

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
alythobani: Ah I didn't necessarily see it as redundant since I'm using "heritage types" to refer specifically to the super-types aka heritage/base types that are extended/implemented. I.e. I meant "this option checks a subtype's async methods against its heritage types". But I'm also happy to go with another name, and maybe `heritageTypes` is inconsistent anyway if the other options refer to what we're checking (not what we're checking *against*).

Maybe `subtypeMethods`?

E.g.

- `subtypeMethods`: Disables checking an asynchronous method in a subtype whose base type expects that method to return `void`

For reference here's the [TS Handbook section on Class Heritage](https://www.typescriptlang.org/docs/handbook/2/classes.html#class-heritage) in case that helps spark any better alternatives for you (`subtypeMethods` is the most precise I can think of right now).

Re: `heritageSignatures` I wonder if "signatures" is more specific than we intend, depending on what we mean by signature? And maybe it's better to focus on the `subtype` since that's what we're checking

packages/eslint-plugin/tests/docs-eslint-output-snapshots/no-misused-promises.shot
alythobani: > For 1, https://github.com/typescript-eslint/typescript-eslint/issues/8942 - would be great if you / anybody had time to look into it!

Ah I think #8942 is a separate issue actually, about tests triggering runtime errors in rule-code and failing with unhelpful output. Vs there's no rule-code runtime error occurring here (and no tests fail); there's just no ESLint output in the snapshot for the incorrect `EventEmitter` example. But also I just saw now that you've already fixed the  example a couple days ago (#9251) so we should be all good anyway now :) (other than the broader issue of false negatives still slipping through, but that's even more outside the scope of this PR)

> Here it is - https://github.com/typescript-eslint/typescript-eslint/issues/8997

Ah nice haha. Ok I went ahead and drafted up a PR for this, let me know how it looks to you two :) #9302

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
alythobani: Nice yeah I like it, although do you happen to know if *derived* also applies when a subtype implements an interface?

JoshuaKGoldberg: Ah, you're right, _derived_ just refers to classes _deriving_ methods from their base. Good call! 

How about `inheritedMethods`?

alythobani: Ah okay thanks for confirming that. Hmm wouldn't `inheritedMethods` have the same issue as `derived`? A class doesn't inherit an interface method, it just implements it, as I understand it?

JoshuaKGoldberg: Whoops, sorry, I missed this till now!

My gut instinct tells me that difficulties in clearly naming something generally come from that thing being too broad. Which makes sense here! It's one option trying to do two things: `extends` and `inherits`. The two are similar on the surface but are actually two different things.

 How about splitting it into two options? Maybe, `extensions` and `implementations`?

JoshuaKGoldberg (changes requested):
Sorry for the delay! I commented on the last remaining thread.

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
alythobani: No prob! That's an interesting point. You're right we are checking two different things (extended types vs implemented types). Although under the surface we are actually just doing one thing, i.e. a single pass through `tsNode.heritageClauses` and running the same code on each heritage clause regardless of its type.

So if we wanted to split it into two options, we'd need to change up that singular code to, instead, first filter `heritageClauses` based on the user's choices (only check `extends` clauses, only check `implements` clauses, or check both); and then still run the same code on the filtered clauses. 

I wonder if this would provide extra value, or potentially just be more confusing for a user vs having a single option to check heritage methods. I think it's likely a similar intention either way: a subtype should be interchangable with its heritage type, e.g. `Dog.bark` should have the same return value as `Animal.bark` whether `Dog extends Animal` or `Dog implements Animal`. And the split options + filtering logic would add extra complexity to the rule implementation (especially e.g. where we [map node types to handlers based on options](https://github.com/typescript-eslint/typescript-eslint/pull/8765/files#diff-dd469c96cb4a7fd7e266d9957762126ca0e669b65eb00173edcf40d1dcdced6fR154-R179)). Let me know you can think of a use case here though, for wanting to check one but not the other.

At the moment I still lean towards `heritageMethods` being the most descriptive and simplest / least confusing. And a good thing is that the word "Heritage" does come straight from both the [TS docs](https://www.typescriptlang.org/docs/handbook/2/classes.html#class-heritage) and TS source code, and is used to cover both `extends` and `implements` clauses there. But if you still think splitting it into two options is more ideal let me know and I can make those changes. Otherwise if you think my logic here makes sense I can just change the name to `heritageMethods` and we'll go with that. Let me know either way, happy to do whatever you think is best :)

JoshuaKGoldberg: cc @typescript-eslint/triage-team - let's see if anybody else on the team has input.

kirkwaiblinger: My 2c:

- I prefer sticking to one option, unless someone complains with a concrete use case of why they'd want to have them specified separately

- I think we need a quick code example in the docs in order to make clear what this option means.

- I did not know the specific meaning of the word "heritage" in this context, and I'm still intimidated by it 😅 . I think it would be acceptable, but not preferable to use, provided there is explanation and example. I would prefer we use the word "inherited" instead. I personally like "inheritedMethodSignatures". or "inheritedMethods". Not super opinionated on the naming though.

alythobani: Hey @kirkwaiblinger  thanks for the thoughts :)

> * I think we need a quick code example in the docs in order to make clear what this option means.

I'd actually already added [an incorrect example](https://github.com/typescript-eslint/typescript-eslint/pull/8765/files#diff-64a7137a0db4419afd7a036a5fa31d285bb7891363e29db109766e41e28141f1R184-R192) and [a correct example](https://github.com/typescript-eslint/typescript-eslint/pull/8765/files#diff-64a7137a0db4419afd7a036a5fa31d285bb7891363e29db109766e41e28141f1R236-R244) to the `checksVoidReturn: true` examples section of the doc:

```ts
interface MySyncInterface {
  setThing(): void;
}
class MyClass implements MySyncInterface {
  async setThing(): Promise<void> {
    this.thing = await fetchThing();
  }
}

...

interface MyAsyncInterface {
  setThing(): Promise<void>;
}
class MyClass implements MyAsyncInterface {
  async setThing(): Promise<void> {
    this.thing = await fetchThing();
  }
}
```

 Do you think those are sufficient? Or is there anything that would help the clarity here?

> I did not know the specific meaning of the word "heritage" in this context, and I'm still intimidated by it 😅 . I think it would be acceptable, but not preferable to use, provided there is explanation and example. I would prefer we use the word "inherited" instead. I personally like "inheritedMethodSignatures". or "inheritedMethods". Not super opinionated on the naming though.

Haha, yeah that's fair enough. I think potentially the upside of `heritageMethods` is that it's more accurate since it refers to both `extends` and `implements` (vs `inheritedMethods` technically doesn't apply to e.g. `MyClass implements MySyncInterface`). But if it would be less intimidating/clearer to use `inheritedMethods` I could see that being worth it, and we could clarify in the option's description that it applies to both implemented and inherited/overridden methods. I can also see an argument that `MyClass` does inherit method signatures from `MySyncInterface`, so maybe `inheritedMethodSignatures` is fully accurate (albeit longer and maybe still ambiguous - is it inherited signatures, or signatures of inherited methods?).

**TLDR** I can see any of `heritageMethods`, `inheritedMethods`, `inheritedMethodSignatures` being good enough with small tradeoffs, and probably lean towards `heritageMethods` or `inheritedMethods` for brevity

kirkwaiblinger: It seems "inheritedMethods" is the unique element of the intersection of our preferences 🚀

JoshuaKGoldberg: I still wish we could find a single-word name 😞 but this compromise is fine with me.

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -369,6 +381,71 @@ export default createRule<Options, MessageId>({
...
+    function checkClassLikeOrInterfaceNode(
+      node:
+        | TSESTree.ClassDeclaration
+        | TSESTree.ClassExpression
+        | TSESTree.TSInterfaceDeclaration,
+    ): void {
+      const tsNode = services.esTreeNodeToTSNodeMap.get(node);
+
+      const heritageTypes = getHeritageTypes(checker, tsNode);
+      if (heritageTypes === undefined || heritageTypes.length === 0) {
```
JoshuaKGoldberg, line 393: [Style] Nitty nit ✨ 

```suggestion
      if (!heritageTypes?.length) {
```

JoshuaKGoldberg (dismissed):
I have been mollified. 😄 

This really is a lovely PR with a super interesting bug case. Thanks for all your work on it @alythobani! 🙌

packages/eslint-plugin/src/rules/no-misused-promises.ts
alythobani: Haha I've gotten mini-traumatized before from a bug related to `!undefined` and `!0` both resolving to `true` (when we just meant to check for `undefined`), so I've tended to make these checks more explicit since then. Which I guess is in line with the [strict-boolean-expressions](https://typescript-eslint.io/rules/strict-boolean-expressions/) TS-Eslint rule? But no strong feelings if you prefer the more concise version, happy to change it :)

alythobani:
😁 Sweet, and thanks for such a detailed review @JoshuaKGoldberg! Really appreciate all the catches and suggestions and all the time you put into it! And for putting up with my overly long explanations 😇

packages/eslint-plugin/src/rules/no-misused-promises.ts
JoshuaKGoldberg: No strong preference from me either. IMO if it was important we'd get around to writing a lint rule for it. 😄

alythobani: Ok cool committed :)

JoshuaKGoldberg (dismissed):
Looks great to me, thanks! Since @kirkwaiblinger was also reviewing, will wait a bit for Kirk to review.

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -369,6 +381,71 @@ export default createRule<Options, MessageId>({
...
+      if (!heritageTypes?.length) {
+        return;
+      }
+
+      for (const nodeMember of tsNode.members) {
+        const memberName = nodeMember.name?.getText();
+        if (memberName === undefined) {
+          // Call/construct/index signatures don't have names. TS allows call signatures to mismatch,
+          // and construct signatures can't be async.
+          // TODO - Once we're able to use `checker.isTypeAssignableTo` (v8), we can check an index
```
kirkwaiblinger, line 402: v8 is upon us! Though of course this PR has been in the works since long before that. 

@alythobani, let's merge as-is, but would you mind filing a followup (that we now know will not be blocked 🙂)

kirkwaiblinger (approved):
🚀 😁

JoshuaKGoldberg (approved):
All right!~ This is looking great! Thanks so much for working with us a bunch on it @alythobani. Your explanations were really helpful and I'm super happy where this went.

Merging now to be released later today in our weekly release! 🙌 

Homer Simpson raising his fists up and yelling "Whoo-hoo!"

alythobani:
Sorry again for the late replies here, life is a little hectic these days! But amaaazing, so good to see this merged! Thanks again so much for all the collaboration here @JoshuaKGoldberg  and @kirkwaiblinger !!

Obviously please keep me updated on any issues that arise, but hopefully there won't be any :)

packages/eslint-plugin/src/rules/no-misused-promises.ts
alythobani: Sweet, yeah definitely! Do you mean filing an issue for this? And I can start looking into a proper implementation too if you'd like

Issue #8597, opened 2024-03-05
Bug: [no-unnecessary-template-expression] Report squigglies should underline template syntax, not just expression inside


kirkwaiblinger:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.3.3&showAST=es&fileType=.tsx&code=AYJwpgDg9iAuAEBnAjgVwJYHNMBt1kXgBIBvAKHngHor4ATKeAOygVSbrBDybHlgAW6QgGMoAW3FgmsADTwY-AX1IKmSvhHAA3dFFSEeYAHQV4AIgAqygJ5qcd9p27peS4UlghXmeQDMQCXg0Vj5YRhDYE3MzGnhrPgBDVHC-dAAPeAB3dBwcT28IBRSNRD4xSWlYQkSODWISeFq6eABfeURGAEl3JgBreHQERAF9HBbY2icuIw1xJsIssDz5cPgRRLym2CiZPXU1wT4s0Zw%2BVyiQaBxE2H34ACMwH0n6rSgHs-F5FgQjwaYvBA8DA6S0BEQ9yGZRwflMrWAQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tmUSWWS18iALbF4AQ2G1Gw6BPioMkRNGgdokcGAC%2BIHUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
`report squigglies ${
  // do not underline this comment, or the ${ on the previous line.
  "They only underline this string, from quote to quote."
  // The autofix will strip out these comments and the ${ and }, so I think it should 
  // underline them as well, to call attention to the whole interpolation being
  // the problem, not the inner expression itself.
}`
```

[...]

PR #8821, opened 2024-04-02
feat(eslint-plugin): [no-useless-template-literals] rename to `no-useless-template-expression` (deprecate `no-useless-template-literals`)

### LINKED ISSUE #8544: Docs: Clarify purpose of no-useless-template-literal (or rename?)

### Suggested Changes

Reading the docs at face value, I would consider it a bug that the example
```ts
const useless = `I am a useless template literal that could be converted to an ordinary string`;
```
is not flagged by no-useless-template-literal. (If it actually should be flagged, feel free to convert this to a bug!)

The way I understand it though, the rule as-is is really going after useless template _interpolations_. I think this strongly needs clarification in the docs.

Furthermore, I think that the quotes rule (https://typescript-eslint.io/rules/quotes/), which _can_ be used to prohibit useless template literals, and the no-useless-template-literal rule docs ought to contain links to each other to help a user figure out which one to use, depending on what the goal is. (I'm currently setting this up at my day job and getting confused over which one to use 😆)

### Affected URL(s)

https://typescript-eslint.io/rules/no-useless-template-literals/
https://typescript-eslint.io/rules/quotes/

cc @StyleShit

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8544 

## Overview

Copy old rule to new rule.
Deprecate rule under old name.
Lots of documentation updating.


packages/eslint-plugin/docs/rules/no-useless-template-literals.mdx
```diff
@@ -9,53 +9,15 @@ import TabItem from '@theme/TabItem';
...
 
-This rule reports template literals that can be simplified to a normal string literal.
+This rule reports template literals that contain substitution expressions (also variously referred to as embedded expressions or string interpolations) that are unnecessary and can be simplified.
 
-## Examples
+:::warning
+This rule is being renamed to [`no-useless-template-expression`](./no-useless-template-expression.mdx).
+After the creation of this rule, it was realized that the name `no-useless-template-literals` could be misleading, seeing as this rule only targets template literals with substitution expressions.
+In particular, it does _not_ aim to flag useless template literals that look like `` `this` `` and could be simplified to `"this"`.
+If you are looking for such a rule, you can configure the [`quotes`](./quotes.mdx) rule to do this.
```
JoshuaKGoldberg, line 18: Heh, and `quotes` has the big _🔥 Deprecated_ notice that points to [@stylistic/ts/quotes](https://eslint.style/rules/ts/quotes). Let's point to that instead.

```suggestion
If you are looking for such a rule, you can configure the [`@stylistic/ts/quotes`](https://eslint.style/rules/ts/quotes) rule to do this.
```

packages/eslint-plugin/src/rules/no-useless-template-expression.ts
```diff
@@ -0,0 +1,194 @@
...
+export type NoUselessTemplateExpressionMessageId =
+  'noUselessTemplateExpression';
+
+export type NoUselessTemplateExpressionRuleObject = Parameters<
+  typeof createRule<[], NoUselessTemplateExpressionMessageId>
+>[0];
+
+export function makeNoUselessTemplateExpressionRuleObject(): NoUselessTemplateExpressionRuleObject {
+  // For some reason, TS isn't totally happy if you return the object directly,
+  // but it doesn't mind if you assign it to a variable first.
```
JoshuaKGoldberg, line 23: [Refactor] This isn't the case on my end 🤔. I just get a lint complaint about `create` not having a return type. What errors were you seeing?

As a general trend, comments like _"For some reason, ..."_ are things we should try to hammer out before merge IMO.

packages/eslint-plugin/src/rules/no-useless-template-expression.ts
```diff
@@ -0,0 +1,194 @@
...
+    },
+  };
+
+  return ruleObject;
+}
+
+// The rule object factory is to support code sharing between the current rule,
+// and the deprecated version of this rule under its previous name
+// no-useless-template-literals.
+// See https://github.com/typescript-eslint/typescript-eslint/issues/8544
```
JoshuaKGoldberg, line 191: [Docs] So this gets picked up in searches for todos...

```suggestion
// See https://github.com/typescript-eslint/typescript-eslint/issues/8544
TODO: Remove in v8
```

...which reminds me, I should note this in #8913. We have a [mention of todo searches for ESLint versions](https://github.com/typescript-eslint/typescript-eslint/blob/edb3f68beec5e13bdafc453afedc6d31f7ecfe06/docs/maintenance/pull-requests/Dependency_Version_Upgrades.mdx#L88) but not for our majors.

https://github.com/typescript-eslint/typescript-eslint/pull/8913/commits/f262b165ec58c3b85a32ca307bcfa8bee3a96522

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
JoshuaKGoldberg, line 1: Heh, I get why you did this object function creation stuff, but IMO a copy & paste would have been easier. We're going to get rid of the code within the next few months anyway. If we need to make changes to just one of the versions then it'd be easier that way.

🤷 I'm not picky though, there is a nice cleanliness to not duplicating the code. Not a blocker from me.

packages/eslint-plugin/docs/rules/no-useless-template-literals.mdx
```diff
@@ -9,53 +9,15 @@ import TabItem from '@theme/TabItem';
 >
 > See **https://typescript-eslint.io/rules/no-useless-template-literals** for documentation.
 
-This rule reports template literals that can be simplified to a normal string literal.
+This rule reports template literals that contain substitution expressions (also variously referred to as embedded expressions or string interpolations) that are unnecessary and can be simplified.
 
-## Examples
+:::warning
+This rule is being renamed to [`no-useless-template-expression`](./no-useless-template-expression.mdx).
```
JoshuaKGoldberg, line 15: [Docs] I think this admonition orders things a little differently from how users would expect to read it. Maybe instead the order could be:

1. This rule is being renamed (as-is)
2. The current name will be removed
3. (blank line)
4. Deeper explanation

That way we get the pertinent info first, then the explanation later.

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
```diff
@@ -1,175 +1,20 @@
...
+import type { NoUselessTemplateExpressionMessageId } from './no-useless-template-expression';
+import { makeNoUselessTemplateExpressionRuleObject } from './no-useless-template-expression';
+
+// This rule was renamed to `no-useless-template-expression`.
+// This module's purpose is just to import the code from the new implementation
+// and adjust its metadata to account for the renaming.
+// See https://github.com/typescript-eslint/typescript-eslint/issues/8544
+
+const ruleObject = makeNoUselessTemplateExpressionRuleObject();
+// @ts-expect-error: easier than figuring out how to make this mutable.
```
JoshuaKGoldberg, line 11: [Docs] Nit: we know how, it's just annoying 😄 

```suggestion
// @ts-expect-error: easier than asserting this as mutable or deep object spreads.
```

JoshuaKGoldberg (changes requested):
LGTM in general! Just requesting changes on some docs touchups (the suggestions). Nothing functionally blocking from me. Thanks! ✨

packages/eslint-plugin/docs/rules/no-useless-template-expression.mdx
JoshuaKGoldberg, line 1: [Docs] Maybe mention somewhere in here that it used to be called `no-useless-template-literals`?

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
kirkwaiblinger: Yeah, I was sort of on the fence. The factory is pretty cumbersome and circumvents some linting (in the case of the tests). But I was a little concerned about the potential for drift since there are other changes to the rule in progress, which would need to be duplicated if they are merged afterwards, or re-copied if merged before this PR.
https://github.com/typescript-eslint/typescript-eslint/pull/8688
https://github.com/typescript-eslint/typescript-eslint/pull/8673

Unsure what's best here.

packages/eslint-plugin/docs/rules/no-useless-template-literals.mdx
kirkwaiblinger: I wrote this to parallel [the deprecation notice for no-useless-template-literal](https://typescript-eslint.io/rules/no-throw-literal/)

https://github.com/typescript-eslint/typescript-eslint/blob/0bd126fb44810b731eb8f336019c40a07964a0e1/packages/eslint-plugin/docs/rules/no-useless-template-literals.mdx#L14-L21


Should we update that too?

packages/eslint-plugin/src/rules/no-useless-template-expression.ts
kirkwaiblinger: oh, yeah, I guess that lint warning is what i was referring to. Which is maybe a bug tbh? I don't get why I should have to put the return type there if I assign the object, but not if I return it.

kirkwaiblinger: filed https://github.com/typescript-eslint/typescript-eslint/issues/8918

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
kirkwaiblinger: I've put up https://github.com/typescript-eslint/typescript-eslint/pull/8919 as an alternate version of this PR that does the copypasta approach.

I guess one idea for mitigating drift could be to make sure that the `git diff` between the corresponding rules files (and also the corresponding test files) stays stable. Wonder if we could write a little test for that.

JoshuaKGoldberg: We could... but 🤷 if it's worth it heh.

packages/eslint-plugin/docs/rules/no-useless-template-literals.mdx
JoshuaKGoldberg: Yeah! Good call :smil

packages/eslint-plugin/docs/rules/no-useless-template-expression.mdx
kirkwaiblinger: I've done this with a big `:::info[Migration]` box, and copied that approach to `only-throw-literal` as well. Lmk if that looks ok.

packages/eslint-plugin/src/rules/no-useless-template-literals.ts
kirkwaiblinger: fair enough. Updated this PR to copypasta approach for simplicity, and addressed remaining comments. Will file followup to float the drift testing idea, but this PR should be good to go, I think

kirkwaiblinger:
@JoshuaKGoldberg 
This is a super last-minute thought, but I wonder whether we should soften it to "no-unnecessary-template-expression" instead of "no-useless-template-expression". The other no-useless-* rules are true noops that can be removed entirely, whereas this is just code that could be simplified. More in line with the no-unnecessary-* rules. We did plenty of bikeshedding on the "template-expression"/"template-interpolation"/"template-substitution" part of the rule in #8544, but didn't give the first half any thought. Not that it was the part of the name that was motivating to change, of course.

JoshuaKGoldberg:
> no-unnecessary-template-expression

I'm up for the `no-unnecesssary` preference, that's a good distinction. Thoughts @Josh-Cena ?

packages/eslint-plugin/docs/rules/no-throw-literal.mdx
```diff
@@ -16,10 +16,10 @@ This rule restricts what can be thrown as an exception.
 
 :::warning
 This rule is being renamed to [`only-throw-error`](./only-throw-error.mdx).
-When it was first created, it only prevented literals from being thrown (hence the name), but it has now been expanded to only allow expressions which have a possibility of being an `Error` object.
-With the `allowThrowingAny` and `allowThrowingUnknown` options, it can be configured to only allow throwing values which are guaranteed to be an instance of `Error`.
+The current name, `no-throw-literal`, will be removed in a future major version of typescript-eslint.
```
StyleShit, line 19: since this PR changes multiple rules, maybe we should also change its title?
(so it'll also be squashed with the proper commit message)

or maybe split tp multiple PRs?

packages/eslint-plugin/docs/rules/no-useless-template-expression.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+:::info
+This rule does not aim to flag template literals without substitution expressions that could have been written as an ordinary string.
+That is to say, this rule will not help you turn `` `this` `` into `"this"`.
+If you are looking for such a rule, you can configure the [`@stylistic/ts/quotes`](https://eslint.style/rules/ts/quotes) rule to do this.
+:::
+
+## When Not To Use It
+
+When you want to allow string expressions inside template literals.
```
StyleShit, line 82: IIRC, it's not only for strings, it's also for other literals such as numbers, booleans, null, etc.

packages/eslint-plugin/docs/rules/no-useless-template-expression.mdx
```diff
@@ -0,0 +1,87 @@
...
+```
+
+</TabItem>
+</Tabs>
+
+:::info
+This rule does not aim to flag template literals without substitution expressions that could have been written as an ordinary string.
+That is to say, this rule will not help you turn `` `this` `` into `"this"`.
+If you are looking for such a rule, you can configure the [`@stylistic/ts/quotes`](https://eslint.style/rules/ts/quotes) rule to do this.
+:::
```
StyleShit, line 78: Like it!

packages/eslint-plugin/tests/docs.test.ts
```diff
@@ -448,7 +448,7 @@ describe('Validating rule metadata', () => {
     describe(ruleName, () => {
       it('`name` field in rule must match the filename', () => {
         // validate if rule name is same as url
-        // there is no way to access this field but its used only in generation of docs url
+        // there is no way to access this field but it's used only in generation of docs url
```
StyleShit, line 451: hmm
should be in this PR?

packages/eslint-plugin/docs/rules/no-throw-literal.mdx
kirkwaiblinger: This comes out of https://github.com/typescript-eslint/typescript-eslint/pull/8821#discussion_r1566177427. 
I think it makes sense to include in this change set since it's such a small (and related) change.

packages/eslint-plugin/docs/rules/no-useless-template-expression.mdx
kirkwaiblinger: Yeah, true. That part of the PR is just unchanged code moved from [the existing page](https://typescript-eslint.io/rules/no-useless-template-literals/), though. See on main:
https://github.com/typescript-eslint/typescript-eslint/blob/f248e689b04bdfd81429977a731ba6b8ad82622e/packages/eslint-plugin/docs/rules/no-useless-template-literals.mdx#L55-L58
  
I don't feel strongly about it, so I would prefer not to bother changing it in this PR 🙃. Feel free to submit a separate PR to change that line if you like! 🙂

packages/eslint-plugin/tests/docs.test.ts
kirkwaiblinger: A previous version of this changeset included some actual code changes here, and this was leftover after reverting those haha. I can remove it 👍

StyleShit (approved):

JoshuaKGoldberg (approved):
🚀 sorry for the delay, this is great!

I'm happy for you to merge whenever you're comfortable with it.

Eugeno:
"no-useless-template-expression" was renamed to "no-unnecessary-template-expression", but in the release note:
> eslint-plugin: [no-useless-template-literals] rename to no-useless-template-expression (deprecate no-useless-template-literals) (https://github.com/typescript-eslint/typescript-eslint/pull/8821)

same for issues [8597](https://github.com/typescript-eslint/typescript-eslint/issues/8597), [8669](https://github.com/typescript-eslint/typescript-eslint/issues/8669), [8677](https://github.com/typescript-eslint/typescript-eslint/issues/8677).

kirkwaiblinger:
@Eugeno 

Yes, the premature merge was a tad unfortunate. Was amended in https://github.com/typescript-eslint/typescript-eslint/pull/9174, which went in before the release, though, so the only impact should be a small handful of slightly confusing release notes and issue/pr titles, but nothing overt to the user

Issue #6128, opened 2022-11-29
Repo: Flag instances of noFormat that don't do anything


JoshuaKGoldberg:
### Suggestion

Some test suites, such as https://github.com/typescript-eslint/typescript-eslint/blob/027ce98b8af1a901ee4362cc252295bc03e6ca31/packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts, use `noFormat` because they need some code blocks to preserve formatting. But many of those `noFormat`s end up being copy & pasted across multiple tests, some of which end up not needing it.

Proposal: let's add a lint rule that flags uses of `noFormat` that don't need to be used?

:x: Incorrect:

```ts
noFormat`a.b;`
```

✅ Correct:

```ts
`a.b;`
```

bradzacher:
we could probably include this directly in the [`plugin-test-formatting`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts) rule as a check.

I.e. something like "if code has `noFormat`, format it, and if the output matches the original code - flag the unnecessary `noFormat`"

Jolg42:
Note: I'm preparing a PR for this ✨

Jolg42:
@bradzacher here it is! https://github.com/typescript-eslint/typescript-eslint/pull/6853
It's my first PR here so feel free to add suggestions, this is a bit unfamiliar to me 😄

PR #8826, opened 2024-04-03
fix(eslint-plugin): [consistent-type-assertions] prevent syntax errors on arrow functions

### LINKED ISSUE #8824: Bug: [consistent-type-assertions] broken autofix for arrow functions

### Playground Link

https://typescript-eslint.io/play#ts=4.7.4&showAST=ts&fileType=.ts&code=MYewdgzgLgBAZjAvDAPAQzATwHwApcCUS2MA3jAL4EDcAUAPT0xoCuUIcAlgB4CmEMdjAZNQkWAmSFiZSswEZMdRjAB6qmBExgoabjF4AnQyENA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6MgeyeUuX0Ra0ipWgENkyRNHyUuqDAG1IU6B2iQANJijjJ02UwDKRJOh2oAvgF1wYCyAtA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
const f = <any>(() => { });
// autofixes to 
// const f = () => { } as any;
// ^^ syntax error
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/consistent-type-assertions": ["error", { "assertionStyle": "as"}]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expect that it should autofix to 
```ts
const f = (() => {}) as any;
```

### Actual Result

autofixes to 
```ts
const f = () => {} as any;
```

### Additional Info

Note that the input is only legal syntax in an ordinary .ts file, not one with jsx/tsx.

related: https://github.com/typescript-eslint/typescript-eslint/issues/6884

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8824 

## Overview

Upon looking into this bug, I learned that there are a whole host of circumstances where the `<T>(expr)` => `(expr) as T` fix failed to put necessary parens around expr (since it was coded to _never_ put any parens around expr). A generic fix has been added, and test cases from lots of precedence categories.

Also, had to change the precedence pertaining to the `=>` operator to address the specific issue in the bug, bringing the `=>` precedence into agreement with its ranking on MDN. Feels suspicious, but didn't seem to break any test cases.

### DIFF
--- a/packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
+++ b/packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
@@ -1,6 +1,6 @@
 /* eslint-disable deprecation/deprecation -- TODO - migrate this test away from `batchedSingleLineTests` */
 
-import { RuleTester } from '@typescript-eslint/rule-tester';
+import { noFormat, RuleTester } from '@typescript-eslint/rule-tester';
 
 import type {
   MessageIds,
@@ -41,8 +41,8 @@ const x = !'string' as A;
 const x = (a as A) + b;
 const x = (a as A) + (b);
 const x = new Generic<string>() as Foo;
-const x = new (Generic<string> as Foo)();
-const x = new (Generic<string> as Foo)('string');
+const x = new ((Generic<string>) as Foo)();
+const x = new ((Generic<string>) as Foo)('string');
 const x = () => ({ bar: 5 } as Foo);
 const x = () => ({ bar: 5 } as Foo);
 const x = () => (bar as Foo);`;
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
```diff
@@ -41,8 +41,8 @@ const x = !'string' as A;
 const x = (a as A) + b;
 const x = (a as A) + (b);
 const x = new Generic<string>() as Foo;
-const x = new (Generic<string> as Foo)();
-const x = new (Generic<string> as Foo)('string');
+const x = new ((Generic<string>) as Foo)();
+const x = new ((Generic<string>) as Foo)('string');
```
kirkwaiblinger, line 45: From playing around with the AST viewer, I think this test case was simply incorrect (though tbh this syntax is bonkers and I struggle to follow it so I'm not totally certain.)

For reference, the inputs are 
```typescript
const x = new (<Foo>Generic<string>)();
const x = new (<Foo>Generic<string>)('string');
```
(located above in the ANGLE_BRACKET_TESTS_EXCEPT_CONST_CASE string.)

packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
```diff
@@ -790,5 +790,126 @@ ruleTester.run('consistent-type-assertions', rule, {
...
+      ],
+      errors: [
+        {
+          messageId: 'as',
+          line: 1,
+        },
+      ],
+    },
+    {
+      code: 'const f = <any>function () {};',
```
kirkwaiblinger, line 824: note that non-arrow functions behave differently from arrow functions

packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
```diff
@@ -790,5 +790,126 @@ ruleTester.run('consistent-type-assertions', rule, {
...
+      ],
+      errors: [
+        {
+          messageId: 'as',
+          line: 1,
+        },
+      ],
+    },
+    {
+      // prettier wants to remove the parens around the yield expression,
```
kirkwaiblinger, line 854: I _think_ this is a prettier bug but tbh I'm not 100% certain. Filed https://github.com/prettier/prettier/issues/16192, we'll see what they say.

kirkwaiblinger: was able to get this bug fixed. This will play nicely with https://github.com/typescript-eslint/typescript-eslint/issues/6128

JoshuaKGoldberg: Nice!! Love to see an upstream bug discovered as part of our testing 😄

Josh-Cena (commented):
Great improvements 👍 Do you know why we have so many precedence levels defined for what MDN defines as "precedence 2"?

kirkwaiblinger:
> Great improvements 👍 

❤️ 

> Do you know why we have so many precedence levels defined for what MDN defines as "precedence 2"?

I personally do not know for sure. But looking at our precedences, eslint's precedences, and prettier's precedences, none of them seems to match 1:1 with the table on MDN. And, for TS syntax, part of that is explainable by there being additional operators not present in raw JS.... If I had to really speculate, though, I'd guess most of it comes down to the fact that we use the precedence switch for _printing_ not for _parsing_, and 
1. It doesn't look like we're tooooo careful about right/left associativity checks in the printing code
1. We don't always want to print code with the minimum of technically required parens (eg mixing `&&` and `||`), so this may have crept into precedence values themselves rather than special-casing in the printing logic 
so this may just be the organic result of getting test cases to pass.

To get more concrete info, we'd need to file a new issue to investigate, I think.

Josh-Cena:
Yeah, we should probably put more thought into this. I personally wrote that MDN table and towards the bottom things definitely get much murkier.

JoshuaKGoldberg (approved):
🔥

PR #8828, opened 2024-04-03
chore: disable cancelling concurrent workflow runs on main

### LINKED ISSUE #8773: Repo: CI runs on `main` commits are being cancelled if a new commit lands before CI completes
### Suggestion

https://github.com/typescript-eslint/typescript-eslint/pull/7819

This PR added concurrency config for our actions. This is great for branch (PR) CI runs as it ensures we don't keep CI runs on old commits if a new commit comes in. So if a user spans a bunch of pushes then we'll only have one run per PR.

However the config also applies to `main` - which is a problem. We rely on the `main` CI to validate each commit is correct. We also rely on it to publish a canary version per commit.

This has lead us into this state:

Screenshot of man's history

We currently have a breakage in the website caused by a semantic merge conflict.
However we have no idea what commit caused the problem because so many commits have had their CI cancelled.

We need to exclude `main` from this concurrency config.

tarunrajput:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8773

## Overview

This PR disables cancelling in progress workflow runs on `main` branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 
 concurrency:
   group: '${{ github.workflow }} - ${{ github.head_ref || github.ref }}'
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 env:
   PRIMARY_NODE_VERSION: 20


bradzacher (approved):
Amazing - thanks for doing this.
This should be a big help in making sure that our main branch is correct!

PR #8829, opened 2024-04-03
chore(typescript-estree): do not pin `minimatch`

SuperchupuDev:
## PR Checklist

## Overview
#7752 installed `minimatch` on the estree package but pinned it with no explanation. This results in duplicated dependencies in some projects now that minimatch 9.0.4 is out. This PR fixes it by using `^` in its version, just like every other package.

Note that #7752 was merged in December, and 9.0.3 was the latest minimatch version back then until 9.0.4 was released 6 days ago, so the PR author probably couldn't have found an incompatibility between different patch `minimatch` versions


bradzacher (approved):
thanks!

Issue #8695, opened 2024-03-17
Bug: RuleCreator leaks internal meta.docs, meta.type types


JoshuaKGoldberg:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Abug) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Relevant Package

utils

### Playground Link

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgUQMoBlgDsYFUbAA2AznAL5wBmUEIcA5AAIwCeYApsQMZTBgwBaToWwwA9AFcCJegG4AUPPYAPSLDhcIWYvCgTC7OAF4UGUfiLEAdACV97AMJR2AQxjQrAd2AwAFhCkAEQguYgAKBHk4DWc3djDNHBUYAEpEKOi4ZxgJKCx0zMyASQATdhxgSmB2KDCsCDK0yMLCxJhkq2c1GAi4EE5iFwBzdlKALjgAIhUXcANJgBo4erLyFIUW8gzosgyyBYyyyhd9GAB5fmAtYgmAbQBdA+j+mBcJ5sySkJuClrLuXiXLQTej0J4tZyaED9LBlEogyG0GFwsHbTIQPw1CYwPTscE7fF9AbDTjvNHRGZzdgTSZnXzLCAAQkmaP2aNYHBBhBcLACMFRhW4vnYIDecAeewOZHWiiAA

### Repro Code

```TypeScript
import { ESLintUtils } from '@typescript-eslint/utils';

export const rule = ESLintUtils.RuleCreator.withoutDocs({
  create(context) {
    return {
      Identifier(node) {
        context.report({ messageId: "example", node });
      }
    }
  },
  defaultOptions: [],
  meta: {
    docs: {
      description: '',
      recommended: 'recommended',
      other: true,
    },
    messages: {
      example: "Oh no!"
    },
    type: 'layout',
    schema: []
  },
});
```

### ESLint Config

```javascript
n/a
```

### tsconfig

```jsonc
n/a
```

### Expected Result

[...]

PR #8833, opened 2024-04-04
feat(eslint-plugin): remove formatting/layout rules

### LINKED ISSUE #8074: 💀 Rule deletions: formatting (layout) rules
## Overview

#8072 tracks _deprecating_ formatting rules as a non-breaking change. As a followup, this issue tracks _removing_ those formatting rules from typescript-eslint altogether. Those rules now live in [eslint-stylistic](https://eslint.style) and as such don't need to exist in this plugin anymore.

As a breaking change for our next major version, let's:

* Delete the files! 🔪 
* Look into removing the `'layout'` property from our internal rule `meta.type` type union
    * Or at least that this should only impact _internal_ rules. We shouldn't negatively impact `eslint-stylistic` or other consumers of our custom rule APIs

JoshuaKGoldberg:
BREAKING CHANGE: removes many rules.

## PR Checklist

- [x] Addresses an existing open issue: fixes #8074

## Overview

Deletes the rules marked as type `'layout'`, as well as handling for them in the config and website tooling.

I confirmed locally that https://github.com/eslint-stylistic/eslint-stylistic _does_ rely on the `'layout'` constituent for `meta.type` (#8695). But they only ever set `type: 'layout'` and later use [`rule as unknown as`](https://github.com/eslint-stylistic/eslint-stylistic/blob/4d0e78ecc222d7af39c01fdd95d08d5af7491754/packages/eslint-plugin-js/utils/createRule.ts#L7). So this can be worked around in userland.

💖


packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -75,7 +75,7 @@ export interface RuleMetaData<
    * - `"suggestion"` means the rule is identifying something that could be done in a better way but no errors will occur if the code isn’t changed.
    * - `"layout"` means the rule cares primarily about whitespace, semicolons, commas, and parentheses, all the parts of the program that determine how the code looks rather than how it executes. These rules work on parts of the code that aren’t specified in the AST.
    */
-  type: 'layout' | 'problem' | 'suggestion';
+  type: 'problem' | 'suggestion';
```
armano2, line 78: we should either keep type or remove documentation above

armano2:
seems good to me

armano2 (approved):

bradzacher (commented):
Delete it all. Yes. Die.

The only suggestion I would have:
like [camelcase](https://typescript-eslint.io/rules/camelcase/) - I think it would be best to leave tombstones behind to forward people to the new rules.

Why? Because people running on older versions will "click to open docs" in their IDE and it will open our site to https://typescript-eslint.io/rules/block-spacing/. They'll get a 404 page and be completely confused and probably file an issue.

Or they'll be using ESLint v8 and notice `comma-dangle` doesn't work as expected with TS constructs and they'll search our repo for a replacement (as is the standard when looking for extension rules) and find no existing rule -then request it cos they didn't read any other docs.

It'll cost us nothing to keep tombstones around but it'll help orient people and cut off useless support requests.

bradzacher (approved):

armano2:
> Delete it all. Yes. Die.
> 
> The only suggestion I would have: like [camelcase](https://typescript-eslint.io/rules/camelcase/) - I think it would be best to leave tombstones behind to forward people to the new rules.
> 
> Why? Because people running on older versions will "click to open docs" in their IDE and it will open our site to https://typescript-eslint.io/rules/block-spacing/. They'll get a 404 page and be completely confused and probably file an issue.
> 
> Or they'll be using ESLint v8 and notice `comma-dangle` doesn't work as expected with TS constructs and they'll search our repo for a replacement (as is the standard when looking for extension rules) and find no existing rule -then request it cos they didn't read any other docs.
> 
> It'll cost us nothing to keep tombstones around but it'll help orient people and cut off useless support requests.

what do you think about releasing old version of docs at https://v7.typescript-eslint.io/, that would let ppl who still use old version have access to it

bradzacher:
@armano2 we could do that. The problem there is discoverability - the rules are setup to link directly to `typescript-eslint.io/rules/rule-name` so if someone uses the "open docs" command then they're not gonna see the legacy docs.

I'm personally not too miffed about older versions losing docs - it's going to be a small fraction of users that actively need the old docs.
Each time this has been brought up in the past our collective response has been "it's just not worth the effort".

PR #8834, opened 2024-04-04
feat(typescript-estree): remove slow deprecated and isolated programs

### LINKED ISSUE #5855: Refactor: Investigate removing Program from isolated programs
## Overview

Right now, `createIsolatedProgram` creates a `ts.Program` with minimal type info behind the scenes:

https://github.com/typescript-eslint/typescript-eslint/blob/09c04fb0d5a5dee0dc266abb2e21aa9a1c489712/packages/typescript-estree/src/create-program/createIsolatedProgram.ts#L63

In theory, the services without a specified project path don’t need to make their own Program. That program was used for the experimental unused variables rule, to get program diagnostics such as unused var diagnostics... but this can take dozens of ms a file, and is rarely used by anybody.

Creating a standalone program that way isn't _very_ expensive, but it would be nice to avoid. We should investigate whether we can get rid of it altogether.

The tricky thing here is parser types: some internal code expects `ASTAndProgram` instances from the parsing results. In this new world, there may not be the `program` from there.

Marking as accepting PRs with the caveat that everything still has to work as expected. Which may or may not be doable. Good luck! 😄 

This issue is from a pairing I had with @bradzacher going over the parser internals.

### LINKED ISSUE #5856: Refactor: Remove createIsolatedProgram call from getProgramAndAST
## Overview

https://github.com/typescript-eslint/typescript-eslint/blob/09c04fb0d5a5dee0dc266abb2e21aa9a1c489712/packages/typescript-estree/src/parser.ts#L90-L91

That last `|| createIsolatedProgram(code, extra)` _should_ never be hit. We believe it was a just-in-case fallback added long ago that is no longer (and maybe never was) needed.

Removing that case may make #5855 easier. 

Marking as accepting PRs with the caveat that everything still has to work as expected. Which may or may not be doable. Good luck! 😄

This issue is from a pairing I had with @bradzacher going over the parser internals.

JoshuaKGoldberg:
BREAKING CHANGE: Removes public API option around default programs.

## PR Checklist

- [x] Addresses an existing open issue: fixes #5855; fixes #5856

## Overview

Removes program creation specific to the two linked issues. I found that `createProjectProgram` was only returning `undefined` when `parseSettings.DEPRECATED__createDefaultProgram` was truthy, which was how `getProgramAndAST` would fall back to an isolated program. Hence tackling both in the same PR.

💖


armano2 (approved):

packages/typescript-estree/src/create-program/createProjectProgram.ts
```diff
@@ -37,7 +37,7 @@ function createProjectProgram(
 
   // The file was either matched within the tsconfig, or we allow creating a default program
   // eslint-disable-next-line deprecation/deprecation -- will be cleaned up with the next major
```
bradzacher, line 39:
```suggestion
  // The file was matched within the tsconfig
```

bradzacher (approved):

PR #8872, opened 2024-04-07
feat(eslint-plugin): [prefer-nullish-coalescing] change ignoreConditionalTests default to true

### LINKED ISSUE #8243: Configs: [prefer-nullish-coalescing] change default for `ignoreConditionalTests` to `true`

### Description

Ref: [#7955](https://github.com/typescript-eslint/typescript-eslint/pull/7955#pullrequestreview-1819739415)

IMO most codebases would prefer to avoid nullish coalescing in conditional tests.

Consider the downlevelled output:
```ts
if (x || y) { }

// vs

if (x ?? y) { } 
// downlevelled to
if (x != null ? x : y) { }
```
There's an extra check there that doesn't do anything.

The logic behind the latter is much harder to grasp exactly, IMO:
- If x is truthy, then true,
- If x is falsey and not nullish, then false,
- Else the boolean coercion of y 

vs the former:
- If x is falsey then false
- Else the boolean coercion of y

It's just A lot simpler to read and understand the non-nullish-coalescing form, IMO.

### Impacted Configurations

_No response_

### Additional Info

_No response_

JoshuaKGoldberg:
BREAKING CHANGE:
Changes the default value of a rule's option.

## PR Checklist

- [x] Addresses an existing open issue: fixes #8243

## Overview

Straightforward change of the default value from `false` to `true`.

I'd considered renaming the option to `checkConditionalTests` so we could avoid having a truthy default... but figured that was a bit more user effort than folks are likely to want. I'd be happy to make that change if requested though.

💖

### DIFF
--- a/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.mdx
+++ b/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.mdx
@@ -65,7 +65,7 @@ foo ?? 'a string';
 
 ### `ignoreConditionalTests`
 
-Setting this option to `true` will cause the rule to ignore any cases that are located within a conditional test. This is set to `false` by default.
+Setting this option to `false` will cause the rule to also check cases that are located within a conditional test. This is set to `true` by default.
 
 Generally expressions within conditional tests intentionally use the falsy fallthrough behavior of the logical or operator, meaning that fixing the operator to the nullish coalesce operator could cause bugs.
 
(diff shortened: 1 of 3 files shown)


kirkwaiblinger (approved):
Lgtm. Yay!

Regarding the description, I don't see a reason to change the name of the option.... People that have already specified the option manually will get to keep their configuration, rather than being punished with a broken lint run just because we changed our minds about a default. And people who have not specified it will be impacted identically independently of whether the name is changed.

PR #8882, opened 2024-04-08
docs: fix link to renovate config

FloEdelmann:
*PR template intentionally not followed, because it is such a small documentation fix.*


bradzacher (approved):
oops good catch! thanks!

PR #8883, opened 2024-04-08
feat(rule-tester): support multipass fixes

### LINKED ISSUE #8554: [rule-tester] support multipass fixes

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8554


packages/rule-tester/src/RuleTester.ts
```diff
@@ -986,20 +997,43 @@ export class RuleTester extends TestFramework {
...
-      assert.strictEqual(
-        result.output,
-        item.code,
-        "The rule fixed the code. Please add 'output' property.",
-      );
+      if (result.outputs.length) {
+        assert.strictEqual(
+          result.outputs[0],
+          item.code,
+          "The rule fixed the code. Please add 'output' property.",
```
JoshuaKGoldberg, line 1034:
```suggestion
          "The rule fixed the code. Please add 'output' or 'outputs' property.",
```

packages/rule-tester/src/RuleTester.ts
JoshuaKGoldberg, line 1: [Docs] https://typescript-eslint.io/packages/rule-tester will need to have a mention of the difference. It's nice that the `raw-loader` usage already pulls in the interface. But I think calling out explicitly that we add the allowance of an array will help people who get confused when ESLint core doesn't have it. WDYT?

JoshuaKGoldberg (approved):
Looks lovely to me - and great test cases too! 

I think @bradzacher will want to take a look too if he has time, given that he wrote a bunch on the feature requests.

Big Boss from Metal Gear nodding goofily at a computer, then turning to the camera and giving a thumbs up

packages/rule-tester/src/RuleTester.ts
auvred: There is only a publicly visible property `output?: string | string[] | null` :smile: so we can't suggest adding `outputs` property. I think the main confusion here is that `runRuleForItem` now returns an array of `outputs` rather than just the string `output`

packages/rule-tester/src/RuleTester.ts
auvred: Ohh, nice catch! I always forget to update the docs!

packages/rule-tester/src/RuleTester.ts
JoshuaKGoldberg: Oh right sorry 🤦 I meant to delete this comment. Ignore me!

JoshuaKGoldberg (approved):
Just one small suggestion, looks great to me! 🔥

packages/eslint-plugin/tests/util/getWrappingFixer.test.ts
```diff
@@ -36,7 +36,7 @@ const voidEverythingRule = createRule({
         fix: getWrappingFixer({
           sourceCode: context.sourceCode,
           node,
-          wrap: code => `void ${code}`,
+          wrap: code => `void ${code.replaceAll('wrap', 'wrapped')}`,
```
kirkwaiblinger, line 39: not necessarily a problem, but how is this change related? Seems to just add noise to this PR?

packages/eslint-plugin/tests/rules/no-useless-template-literals.test.ts
```diff
@@ -409,11 +409,16 @@ declare const nested: string, interpolation: string;
   \`le\${  \`ss\`  }\`
 }\`;
       `,
-      output: `
+      output: [
```
kirkwaiblinger, line 412: yeah, as the author of this test case it really bugged me that I had to write this with the first pass only. Very excited to see this feature land!

packages/rule-tester/src/RuleTester.ts
```diff
@@ -573,29 +572,43 @@ export class RuleTester extends TestFramework {
...
+
+      const fatalErrorMessage = messages.find(m => m.fatal);
+      assert(
+        !fatalErrorMessage,
+        `A fatal parsing error occurred: ${fatalErrorMessage?.message}`,
+      );
+
+      fixedResult = SourceCodeFixer.applyFixes(code, messages);
+      if (!messages.length || fixedResult.output === code) {
+        continue;
```
kirkwaiblinger, line 605: Question - i'm confused why this line would be a `continue` rather than a `break`. Could you clarify?

packages/rule-tester/src/RuleTester.ts
```diff
@@ -573,29 +572,43 @@ export class RuleTester extends TestFramework {
...
+      }
+      outputs.push((code = fixedResult.output));
```
kirkwaiblinger, line 607: (Style preference, optional) 
```suggestion
      code = fixedResult.output;
      outputs.push(code);
```

kirkwaiblinger (approved):
I _LOVE_ this feature! ❤️ 

Approved pending resolution of comments, which may not require any changes at all. If you do make changes, I think they're simple enough that you have carte blanche from me to merge without an explicit reapproval from me.

packages/eslint-plugin/tests/util/getWrappingFixer.test.ts
auvred: Without this replacement, this rule produces infinite fixers :smile:, because the rule listeners, such as `Identifier[name="wrapMe"]`, still match the `wrap*` node.

packages/rule-tester/src/RuleTester.ts
auvred: The while loop will break because `fixedResult.fixed` will be `false` on the next iteration. But yes, I agree, using `break` shows a clearer intent. Thx!

kirkwaiblinger: TIL that the `continue` in a `do...while` still checks the condition 😆 I thought that 
```ts
do {
    continue;
} while(false);
``` 
would be an infinite loop lol

kirkwaiblinger (approved):

PR #5327, opened 2022-07-09
feat(eslint-plugin): [lines-around-comment] add extension rule

chbdetta:
## PR Checklist

-   [x] Addresses an existing open issue: fixes #1150
-   [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
-   [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

## Overview

port the `eslint/lines-around-comment` rule and support interface and type literals

### Options
in addition to the base rule options, this adds:
```typescript
allowInterfaceStart?: boolean;
allowInterfaceEnd?: boolean;
allowTypeStart?: boolean;
allowTypeEnd?: boolean;
allowEnumStart?: boolean;
allowEnumEnd?: boolean;
allowModuleStart?: boolean;
allowModuleEnd?: boolean;
```

PR #8893, opened 2024-04-10
fix(eslint-plugin): [init-declarations] refine report locations

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes (partially) #8543 

## Overview

```ts
/* eslint @typescript-eslint/init-declarations: ["error", "always"] */
// before
let bar: number | undefined | string;
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
// after
let bar: number | undefined | string;
    ^^^
```
Note that the following example is unchanged
```ts
/* eslint @typescript-eslint/init-declarations: ["error", "never"] */
let bar: number | undefined | string = 3;
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

### CODE REVIEW

File: packages/eslint-plugin/src/rules/init-declarations.ts [hunk 1]
```diff
@@ -28,7 +28,51 @@ export default createRule<Options, MessageIds>({
...
+      //
+      // So, we'll just use Proxy on a dummy object and use the `get` trap to
+      // proxy `context`'s properties.
+      return new Proxy({} as typeof context, {
+        get: (target, prop, receiver): unknown =>
+          prop === 'report'
+            ? reportOverride
+            : Reflect.get(context, prop, receiver),
+      });
+    }
```
JoshuaKGoldberg, line 73: 😬 It's a little awkward having to `Proxy` the base rule. Is there an issue that was ever filed on ESLint about better handling of report locations in this case? I couldn't find one. But even if it's declined as out-of-scope, it'd be good for reference.

JoshuaKGoldberg (approved):
LGTM, sorry for missing this till now! 🚀 

It'd be nice to try to get this fixed more cleanly in ESLint core... but my hunch is we don't have a super high chance of that getting accepted.

Hades from the Disney Hercules movie giving two flaming thumbs up

packages/eslint-plugin/src/rules/init-declarations.ts
kirkwaiblinger: I'm not sure... I basically copied this pattern from 

https://github.com/typescript-eslint/typescript-eslint/blob/21375f65c734afc4545c0c3e7a15f72d3959c660/packages/eslint-plugin/src/rules/lines-around-comment.ts#L417-L430

and 

https://github.com/typescript-eslint/typescript-eslint/blob/3c6379b7678bcb190ae70d211cb3930c942d17a0/packages/eslint-plugin/src/rules/prefer-destructuring.ts#L196-L206

And modified it with stylistic preferences. 

Following the trail through the git history, it doesn't look like any investigation into alternatives or issues with eslint were documented in those.

(breadcrumbs)
  - https://github.com/typescript-eslint/typescript-eslint/pull/5327: first appearance of proxy pattern, not remarked upon.
  - https://github.com/typescript-eslint/typescript-eslint/pull/7117: pattern copied, remark in https://github.com/typescript-eslint/typescript-eslint/pull/7117#discussion_r1285109350 addresses caching of the proxy, but not the proxy itself.

Will plan to check for eslint issues on the topic before this PR gets merged

JoshuaKGoldberg: > before this PR gets merged

Oh @kirkwaiblinger I was going to merge today for the Monday release and consider this a followup. But in that case I'll just let you merge whenever you want.

kirkwaiblinger: Followup makes sense to me too! 👍
Happy for us to merge for the release

kirkwaiblinger: Opened https://github.com/eslint/eslint/discussions/18557

PR #8920, opened 2024-04-14
fix(typescript-estree): add TSEnumBody node for TSEnumDeclaration body

### LINKED ISSUE #1305: Align AST for TSEnumDeclaration with TSInterfaceDeclaration, TSModuleDeclaration, etc
For each `TSEnumDeclaration`, we place its members directly on this AST node.
```ts
export interface TSEnumDeclaration extends BaseNode {
  type: AST_NODE_TYPES.TSEnumDeclaration;
  id: Identifier;
  members: TSEnumMember[];
  const?: boolean;
  declare?: boolean;
  modifiers?: Modifier[];
  decorators?: Decorator[];
}
```

For each `TSInterfaceDeclaration`, we have a separate `TSInterfaceBody` node, upon which we place the members.
```ts
export interface TSInterfaceDeclaration extends BaseNode {
  type: AST_NODE_TYPES.TSInterfaceDeclaration;
  body: TSInterfaceBody;
  id: Identifier;
  typeParameters?: TSTypeParameterDeclaration;
  extends?: ExpressionWithTypeArguments[];
  implements?: ExpressionWithTypeArguments[];
  decorators?: Decorator[];
  abstract?: boolean;
  declare?: boolean;
}

export interface TSInterfaceBody extends BaseNode {
  type: AST_NODE_TYPES.TSInterfaceBody;
  body: TypeElement[];
}
```

I'm not sure why there's this inconsistency between the two, but I think we should consider aligning the two.

I'm not sure why `TSEnumDeclaration` was created without a body, but I think that creating a `TSEnumBody` node seems like the correct thing to do.
Doing so would align it with not only `TSInterfaceDeclaration`, but also `TSModuleDeclaration`, `ClassDeclaration` and `FunctionDeclaration`.

JoshuaKGoldberg:
BREAKING CHANGE:
Deprecates an existing property on the AST

## PR Checklist

- [x] Addresses an existing open issue: fixes #1305

## Overview

Marks `TSEnumDeclaration`'s `members` property as `@deprecated` in favor of a new `body` property of a new type `TSEnumBody`. Adds a getter with similar `process.emitWarning` logic to our existing deprecated getters.


packages/eslint-plugin/tests/rules/lines-around-comment.test.ts
```diff
@@ -791,31 +761,6 @@ enum A {
...
-`,
-      options: [
-        {
-          beforeBlockComment: false,
-          afterBlockComment: true,
-          allowEnumEnd: false,
-        },
-      ],
-      errors: [{ messageId: 'after', type: AST_TOKEN_TYPES.Block, line: 3 }],
-    },
```
JoshuaKGoldberg, line 818: This rule is being removed in #8833 anyway. I didn't feel like fixing up bugs in it. 🔪

packages/ast-spec/src/special/TSEnumBody/spec.ts
```diff
@@ -0,0 +1,8 @@
+import type { AST_NODE_TYPES } from '../../ast-node-types';
+import type { BaseNode } from '../../base/BaseNode';
+import type { TSEnumMember } from '../../element/TSEnumMember/spec';
+
+export interface TSEnumBody extends BaseNode {
+  type: AST_NODE_TYPES.TSEnumBody;
+  members: TSEnumMember[];
```
bradzacher, line 7:
```suggestion
  members: TSEnumMember[];
  parent: TSEnumDeclaration;
```

This holds true, right?

packages/eslint-plugin-internal/src/rules/prefer-ast-types-enum.ts
```diff
@@ -41,9 +41,9 @@ export default createRule({
       Literal(node: TSESTree.Literal): void {
         if (
           node.parent.type === AST_NODE_TYPES.TSEnumMember &&
-          node.parent.parent.type === AST_NODE_TYPES.TSEnumDeclaration &&
+          node.parent.parent.parent.type === AST_NODE_TYPES.TSEnumDeclaration &&
```
bradzacher, line 44: with my comment above - I think this condition would fall away as unnecessary

packages/eslint-plugin/src/rules/indent.ts
bradzacher, line 1: note: we're deleting this in v8

bradzacher (approved):

packages/visitor-keys/src/visitor-keys.ts
```diff
@@ -206,7 +206,8 @@ const additionalKeys: AdditionalKeys = {
   TSDeclareFunction: SharedVisitorKeys.Function,
   TSDeclareKeyword: [],
   TSEmptyBodyFunctionExpression: ['id', ...SharedVisitorKeys.FunctionType],
-  TSEnumDeclaration: ['id', 'members'],
+  TSEnumBody: ['members'],
+  TSEnumDeclaration: ['body', 'id'],
```
bradzacher, line 210:
```suggestion
  TSEnumDeclaration: ['id', 'body'],
```

These need to be placed in the order that they should be visited during a traversal.
EG the order they appear in source

packages/ast-spec/src/special/TSEnumBody/spec.ts
JoshuaKGoldberg: Yeah!!

Issue #7350, opened 2023-07-28
Docs: Add blog post for EXPERIMENTAL_useProjectService


JoshuaKGoldberg:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

Following #7349: once `EXPERIMENTAL_useProjectService` is known to be not-terrible in v6, we can add a blog post explaining it. The post should mention why it exists, its pros (performance! simpler .js file configuration!), its cons (?), and that it's experimental & likely to change in v7.

I'll also set up a GitHub Discussion for folks to give feedback. The blog post can link to it.

Blocked on:

* #7349
* #6520: It would feel weird to me to publish this one first...
* #7348: or at least better understanding it

### Affected URL(s)

https://typescript-eslint.io/blog/experimental-project

NextThread:
Hey @JoshuaKGoldberg are you working on this ?

JoshuaKGoldberg:
👋 see the OP - it's blocked on #6520 and #7348. I'll work on it once those are done.

PR #8925, opened 2024-04-15
feat(typescript-estree): add maximumDefaultProjectFileMatchCount and wide allowDefaultProjectForFiles glob restrictions

### LINKED ISSUE #8923: Enhancement: ban wide globs in `parserOptions.useProjectService.allowDefaultProjectForFiles`

### Relevant Package

typescript-estree

### Description

We should lock down this API to start with and widen it as needed.
In particular we know that the default project can and will cause perf issues if used incorrectly.
EG if it matches too many files it'll cause dramatic slowdowns.

Based on what @JoshuaKGoldberg has found - it'll create one new `ts.Program` per file. This process is slow - but given the number of files is small then it shouldn't impact things too badly. If someone does something dumb like pass a glob like `**/*.ts` to match their entire project then their run will be stupidly slow for no good reason.

In a nutshell - the motivation against this was the entire reason we moved away from "default programs" back in [v2.0.0](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v2.0.0). Because people accidentally fell into the default program and their lint runs slowed to a crawl because of them.

We should do the following:

(1) ban obviously dumb globs like `*`, `**/*`, `**/*.*`, etc. Essentially any glob that doesn't match the regex `\w` should be banned with an explicit failure like

> Unnecessarily wide globs are disallowed. If you have a usecase for them please file an issue in the typescript-eslint project.

(2) ban too many files matching the glob. We should pick a sensible default that we think won't noticably impact the run performance (say, 5 files) and if the lint run matches against more than that many files - we should explicitly fail like

> `{count}` files have matched the default project. Having many files run with the default project is known to cause performance issues and slow down your lint run.
> Consider creating a separate `tsconfig.json` to match these files `https://typescript-eslint.io/docs/link/to/guide/about/this/optimisation`.
> If you're ok with slower lint runs then set `parserOptions.iLikeSlowLintRunsAndICannotLie_DO_NOT_USE_THIS_OR_YOU_WILL_BE_FIRED` to an integer greater than `{count}`

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8923

## Overview

Adds the two safety checks for `allowDefaultProjectForFiles` as suggested in the issue:
* A new `maximumDefaultProjectFileMatchCount_THIS_WILL_SLOW_DOWN_LINTING` option with a default value of `8`
* A check that `allowDefaultProjectForFiles` doesn't include `**`
   * Intentionally allows slightly more permissive matches like `./*.*s`, as that's what I'd expect to be convenient & reasonable alongside the count maximum

Adds an FAQs blog post to explain them.

💖

### DIFF
--- a/docs/troubleshooting/FAQ.mdx
+++ b/docs/troubleshooting/FAQ.mdx
@@ -32,6 +32,40 @@ If you don't find an existing extension rule, or the extension rule doesn't work
 > We release a new version our tooling every week.
 > _Please_ ensure that you [check our the latest list of "extension" rules](/rules/#extension-rules) **_before_** filing an issue.
 
+<HiddenHeading id="allowdefaultprojectforfiles-glob-too-wide" />
+
+## I get errors telling me "Having many files run with the default project is known to cause performance issues and slow down linting."
+
+These errors are caused by using the [`EXPERIMENTAL_useProjectService`](../packages/Parser.mdx#experimental_useprojectservice) `allowDefaultProjectForFiles` with an excessively wide glob.
+`allowDefaultProjectForFiles` causes a new TypeScript "program" to be built for each "out of project" file it includes, which incurs a performance overhead for each file.
+
+To resolve this error, narrow the glob(s) used for `allowDefaultProjectForFiles` to include fewer files.
+For example:
+
+```diff title="eslint.config.js"
+parserOptions: {
+  EXPERIMENTAL_useProjectService: {
+    allowDefaultProjectForFiles: [
+-      "**/*.js",
++      "./*.js"
+    ]
+  }
+}
+```
+
+You may also need to include more files in your `tsconfig.json`.
+For example:
+
+```diff title="tsconfig.json"
+"include": [
+  "src",
++ "*.js"
+]
+```
+
+If you cannot do this, please [file an issue on typescript-eslint's typescript-estree package](https://github.com/typescript-eslint/typescript-eslint/issues/new?assignees=&labels=enhancement%2Ctriage&projects=&template=07-enhancement-other.yaml&title=Enhancement%3A+%3Ca+short+description+of+my+proposal%3E) telling us your use case and why you need more out-of-project files linted.
+Be sure to include a minimal reproduction we can work with to understand your use case!
+
 ## I get errors telling me "ESLint was configured to run ... However, that TSConfig does not / none of those TSConfigs include this file"
 
 These errors are caused by an ESLint config requesting type information be generated for a file that isn't included in the TypeScript configuration.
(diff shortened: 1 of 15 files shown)


docs/troubleshooting/FAQ.mdx
```diff
@@ -499,6 +533,6 @@ If you think you're having issues with performance, see our [Performance Trouble
 
 ## Are TypeScript project references supported?
 
-No, TypeScript project references are not yet supported.
+Yes, but only with [`EXPERIMENTAL_useProjectService`](../packages/Parser.mdx#experimental_useprojectservice).
```
JoshuaKGoldberg, line 536: Aside: we'll definitely want more explanation here 😄 but quickly noted this in the meantime. Once this & a few more v8 milestone PRs are in, #7350 tracks a full blog post / explainer / revamp.

bradzacher (approved):

me4502:
Is this meant to impact situations where `EXPERIMENTAL_useProjectService` is set to `true`? Not sure if it's a bug (will make an issue if it is), but I'm now seeing `Parsing error: Too many files (>8) have matched the default project.` for every single file in each workspaces' lint run as of 7.8.0.

From my understanding the project service should be loading the closest `tsconfig.json` in the file tree, and using the default if it doesn't match an included file, correct? It seems to be either ignoring the tsconfig.json file per workspace or ignoring the `includes` values in each one.

JoshuaKGoldberg:
Hmm, every single file hitting this does sound like a bug. I'd appreciate an issue - yes please!

me4502:
Will do, thanks. I haven't managed to reproduce this outside of one specific large & complex monorepo, so I'll see if I can narrow it down to reproduce.

ehoogeveen-medweb:
I'm also seeing this match every file when linting a directory containing a project config, even after setting `allowDefaultProjectForFiles: []`. I'll try to get a minimal reproduction.

stepankuzmin:
Having the same issue with `Too many files (>8) have matched the default project` after upgrading to `7.8.0`.

JoshuaKGoldberg:
Heyall, per our [PR contributing guide](https://typescript-eslint.io/contributing/pull-requests) please don't post on resolved PRs. The comments tend to get lost and this isn't an effective way to discuss issues. Please [file a new issue](https://github.com/typescript-eslint/typescript-eslint/issues/new/choose) with the appropriate template if you've got something to report that hasn't yet been reported.

#9032 tracks the general case of this error. Nobody's investigated to see what the actual issue is (if you have time, please do!). That would be the issue to subscribe to for updates. If you've got any _new information_ to add such as investigation results or a new kind of reproduction, please add it there. Saying _"I also have this issue!"_ doesn't give us any information and just gives an unnecessary notification to everyone in the thread.

`maximumDefaultProjectFileMatchCount_THIS_WILL_SLOW_DOWN_LINTING` is appropriately obnoxiously named. It's a "bandaid" solution to files being unmatched by the project service. A better solution is to figure out why they're unmatched and fix that bug - which is either in the ESLint config being used or in typescript-eslint.

PR #8944, opened 2024-04-18
chore: new grapheme library

cometkim:
## PR Checklist

## Overview

Replace [graphemer] library to [unicode-segmenter] library which is ligher and faster. See [benchmark](https://github.com/cometkim/unicode-segmenter#grapheme-clusters)

The Unicode grapheme splitter library, [graphemer] that the TypeScript-ESLint core currently relies on was last updated two years ago and does not include the Unicode 15.1.0 specification.

I created an alternative, the [unicode-segmeneter] library, has a lighter bundle and better performance.

unicode-segmeneter is based on the latest(v15.1.0) Unicode data, is smaller than graphemer, and has over 6x faster count performance.

Specification compliance is tested through fast-check property assertion with V8's built-in [Intl.Segmenter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) which is based on ICU.

See also https://github.com/eslint/eslint/pull/18359

[graphemer]: https://github.com/flmnt/graphemer
[unicode-segmenter]: https://github.com/cometkim/unicode-segmenter

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -67,11 +67,11 @@
     "@typescript-eslint/utils": "7.7.0",
     "@typescript-eslint/visitor-keys": "7.7.0",
     "debug": "^4.3.4",
-    "graphemer": "^1.4.0",
     "ignore": "^5.3.1",
     "natural-compare": "^1.4.0",
     "semver": "^7.6.0",
-    "ts-api-utils": "^1.3.0"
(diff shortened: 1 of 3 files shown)


bradzacher:
We are not looking to diverge from ESLint core here - we use the same dependencies to ensure that things are in sync.
If they accept and merge your PR then we can do the same.

Issue #3503, opened 2021-06-08
[unbound-method] Should prevent conforming to an bound-style interface method with an unbound method


jtbandes:
- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.
- [x] I have [read the FAQ](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/FAQ.md) and my problem is not listed.

**Repro**

```TS
interface Foo {
  unbound(): number;
  unboundOk(this: void): number;
  bound: () => number;
}

function what(f: Foo) {
  const unbound = f.unbound;      // lint error 👍
  const unboundOk = f.unboundOk;  // no error 👍
  const bound = f.bound;          // no error 👍

  // console.log(unbound());  // runtime error (expected)
  console.log(unboundOk()); // runtime error ❌
  console.log(bound());     // runtime error ❌
}

class Example implements Foo {
  x = 3;
  unbound() { return this.x; }
  unboundOk() { return this.x; }   // no error 👎 (TypeScript issue rather than typescript-eslint issue?)
  bound() { return this.x; }      // no error 👎 (this issue)
}
what(new Example());
```

eslint config: https://github.com/foxglove/eslint-plugin/blob/main/configs/typescript.js
tsconfig: https://github.com/foxglove/studio/blob/main/packages/tsconfig/tsconfig.base.json

**Expected Result**

`class Example implements Foo` is problematic. There should be an error on `bound() { return this.x }` when is used to satisfy the interface requirement `bound: () => number;`.

**Actual Result**

The lint rule does not reject this incorrect method implementation, which leads to a runtime error.

**Versions**

| package                            | version |
| ---------------------------------- | ------- |
| `@typescript-eslint/eslint-plugin` | `4.26.0` |
| `@typescript-eslint/parser`        | `4.26.0` |
| `TypeScript`                       | `4.3.2` |
| `ESLint`                           | `7.27.0` |
| `node`                             | `15.9.0` |

bradzacher:
It's important to note that you are misunderstanding what TS syntax means.
```ts
interface Foo {
  prop: () => number;
}
```

This does not mean or enforce that `prop` is an arrow function.
Whilst it looks similar to an arrow function property - **the two are not at all related**!

[...]

PR #8952, opened 2024-04-19
fix(eslint-plugin): [unbound-method] report on destructuring in function parameters

### LINKED ISSUE #8947: Bug: [unbound-method] Does not fail for destructured parameters

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.3&fileType=.ts&code=MYGwhgzhAEAKD2ECWAXJ8B20DeBYAUNNAMQAe0AvNAAwDcBRxAnpTffg9KQBQCUO0TkQBOAUxQBXYVhQALJBAB0ZdkQC%2BBTkz44h0MZOnQ5C5U1XQN%2BKwQBmEjMDSZot%2BPG4AHAIwAuOIio6BgANAKk-qQATGFM-kxRlv4IyM4Y-HiE0MCYECjhkd6x8d6WrD4WBlJYANqk3joA1FxRfLEN-M0JfAC67FZAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJgIwHsOATWgFtE%2BABbc%2B6KImjRu0SODABfEMqA&tsconfig=&tokens=false

### Repro Code

```TypeScript
class Position {
  #x = 0;
  #y = 0;

  x() { 
    return this.#x;
  }

  y() {
    return this.#y;
  }
}

function foo(p1: Position, { x: x2, y: y2 }: Position) {
  const { x: x1, y: y1 } = p1;
  return [x1() + x2(), y1() + y2()];
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/unbound-method": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

I expect the rule to fail when destructuring `x2` and `y2` in the second function parameter.

### Actual Result

There are errors when destructuring `x1` and `y1` in the function body as expected, but no error for the equivalent code within the function parameter.

### Additional Info

_No response_

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8947

## Overview

Old approach: query for `VariableDeclarator, AssignmentExpression` and check if they have `ObjectPattern`

New approach: query for `ObjectPattern`, check if `VariableDeclarator, AssignmentExpression, AssignmentPattern` its parent

### DIFF
--- a/packages/eslint-plugin/src/rules/unbound-method.ts
+++ b/packages/eslint-plugin/src/rules/unbound-method.ts
@@ -135,9 +135,9 @@ export default createRule<Options, MessageIds>({
     function checkIfMethodAndReport(
       node: TSESTree.Node,
       symbol: ts.Symbol | undefined,
-    ): void {
+    ): boolean {
       if (!symbol) {
-        return;
+        return false;
       }
 
       const { dangerous, firstParamIsThis } = checkIfMethod(
@@ -152,7 +152,9 @@ export default createRule<Options, MessageIds>({
               : 'unbound',
           node,
         });
+        return true;
       }
+      return false;
     }
 
     return {
@@ -173,22 +175,25 @@ export default createRule<Options, MessageIds>({
 
         checkIfMethodAndReport(node, services.getSymbolAtLocation(node));
       },
-      'VariableDeclarator, AssignmentExpression'(
-        node: TSESTree.AssignmentExpression | TSESTree.VariableDeclarator,
-      ): void {
-        const [idNode, initNode] =
-          node.type === AST_NODE_TYPES.VariableDeclarator
-            ? [node.id, node.init]
-            : [node.left, node.right];
+      ObjectPattern(node): void {
+        let initNode: TSESTree.Node | null = null;
+        if (node.parent.type === AST_NODE_TYPES.VariableDeclarator) {
+          initNode = node.parent.init;
+        } else if (
+          node.parent.type === AST_NODE_TYPES.AssignmentPattern ||
+          node.parent.type === AST_NODE_TYPES.AssignmentExpression
+        ) {
+          initNode = node.parent.right;
+        }
 
-        if (initNode && idNode.type === AST_NODE_TYPES.ObjectPattern) {
-          const rightSymbol = services.getSymbolAtLocation(initNode);
+        if (initNode) {
+          const initSymbol = services.getSymbolAtLocation(initNode);
           const initTypes = services.getTypeAtLocation(initNode);
 
           const notImported =
-            rightSymbol && isNotImported(rightSymbol, currentSourceFile);
+            initSymbol && isNotImported(initSymbol, currentSourceFile);
 
-          idNode.properties.forEach(property => {
+          node.properties.forEach(property => {
             if (
               property.type === AST_NODE_TYPES.Property &&
               property.key.type === AST_NODE_TYPES.Identifier
@@ -209,6 +214,42 @@ export default createRule<Options, MessageIds>({
               );
             }
           });
+          return;
+        }
+
+        if (node.typeAnnotation) {
+          const subtypes = tsutils.unionTypeParts(
+            services.getTypeAtLocation(node.typeAnnotation.typeAnnotation),
+          );
+          for (const property of node.properties) {
+            for (const subtype of subtypes) {
+              if (
+                property.type !== AST_NODE_TYPES.Property ||
+                property.key.type !== AST_NODE_TYPES.Identifier
+              ) {
+                continue;
+              }
+
+              const propertyKey = property.key;
+              if (
+                checkIfMethodAndReport(
+                  propertyKey,
+                  subtype.getProperty(propertyKey.name),
+                ) ||
+                (tsutils.isIntersectionType(subtype) &&
+                  tsutils
+                    .intersectionTypeParts(subtype)
+                    .some(intersectionSubtype =>
+                      checkIfMethodAndReport(
+                        propertyKey,
+                        intersectionSubtype.getProperty(propertyKey.name),
+                      ),
+                    ))
+              ) {
+                break;
+              }
+            }
+          }
         }
       },
     };
--- a/packages/eslint-plugin/tests/rules/unbound-method.test.ts
+++ b/packages/eslint-plugin/tests/rules/unbound-method.test.ts
@@ -284,6 +284,43 @@ class Foo {
 }
 const { bound } = new Foo();
     `,
+    `
+class Foo {
+  bound = () => 'foo';
+}
+function foo({ bound } = new Foo()) {}
+    `,
+    `
+class Foo {
+  bound = () => 'foo';
+}
+declare const bar: Foo;
+function foo({ bound }: Foo) {}
+    `,
+    `
+class Foo {
+  bound = () => 'foo';
+}
+class Bar {
+  bound = () => 'bar';
+}
+function foo({ bound }: Foo | Bar) {}
+    `,
+    `
+class Foo {
+  bound = () => 'foo';
+}
+type foo = ({ bound }: Foo) => void;
+    `,
+    `
+class Foo {
+  bound = () => 'foo';
+}
+class Bar {
+  bound = () => 'bar';
+}
+function foo({ bound }: Foo & Bar) {}
+    `,
     // https://github.com/typescript-eslint/typescript-eslint/issues/1866
     `
 class BaseClass {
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin/src/rules/unbound-method.ts
```diff
@@ -209,6 +214,42 @@ export default createRule<Options, MessageIds>({
               );
             }
           });
+          return;
```
kirkwaiblinger, line 217: The control flow here has a couple gotchas surrounding 
1. variables with type annotations
2. declarations (via `declare` modifier or `FunctionDeclaration`)

```ts
class Foo {
  unbound = function () { };
}

// not sure why this is even legal syntax tbh, but this is a regression (shouldn't flag):
declare const { unboundMethod }: Foo; // ditto for var/let
                ^^^^^^^^^^^^^

// also should not flag
declare const { split } = "wtf";
                ^^^^^

interface HasMethod {
   method(): void;
}

// bug with current impl; rule optimistically uses the `initNode` but should still check the type annotation.
function shouldFlag({ method }: HasMethod = { method: () => {} }) {};
                      ^^^^^^

// declarations strike again. (this is just a signature, not runtime code)
function shouldNotFlag({ method }: HasMethod);
                         ^^^^^^

// Should this flag? I don't think so... but I could be convinced either way. 
const { method }: HasMethod = { method: () => {} };
```

packages/eslint-plugin/src/rules/unbound-method.ts
```diff
@@ -209,6 +214,42 @@ export default createRule<Options, MessageIds>({
...
+              if (
+                property.type !== AST_NODE_TYPES.Property ||
+                property.key.type !== AST_NODE_TYPES.Identifier
+              ) {
+                continue;
+              }
+
+              const propertyKey = property.key;
+              if (
+                checkIfMethodAndReport(
```
kirkwaiblinger, line 235: (style) Consider assigning the result of `checkIfMethodAndReport` to a variable to indicate clearly that it has side effects.

packages/eslint-plugin/src/rules/unbound-method.ts
```diff
@@ -209,6 +214,42 @@ export default createRule<Options, MessageIds>({
...
+                  propertyKey,
+                  subtype.getProperty(propertyKey.name),
+                ) ||
+                (tsutils.isIntersectionType(subtype) &&
+                  tsutils
+                    .intersectionTypeParts(subtype)
+                    .some(intersectionSubtype =>
+                      checkIfMethodAndReport(
```
kirkwaiblinger, line 243: Array callbacks should be pure functions. Please use explicit looping for control flow that includes `checkIfMethodAndReport`.

kirkwaiblinger (changes requested):
I really like these changes! The algorithm change makes sense to me, the code is easy to follow, and the new code for type inspection is very thorough. 

Added a few questions and some test cases to flesh out edge cases.

packages/eslint-plugin/src/rules/unbound-method.ts
```diff
@@ -209,6 +214,42 @@ export default createRule<Options, MessageIds>({
               );
             }
           });
+          return;
+        }
+
+        if (node.typeAnnotation) {
```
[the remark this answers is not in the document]
auvred: Yea, you are right! It would be 10x better if we relied on the actual type rather than the syntax around the destructuring. That way we can actually handle `VariableDeclarator`s without `init`:

```ts
class Unbound {
  method() { }
}

declare const foo: Unbound[]
for (const { method } of foo) {}
```

packages/eslint-plugin/src/rules/unbound-method.ts
auvred: A lot of nice catches, thanks!

packages/eslint-plugin/src/rules/unbound-method.ts
```diff
@@ -152,69 +147,162 @@ export default createRule<Options, MessageIds>({
...
+        const objectSymbol = services.getSymbolAtLocation(object);
+        const notImported =
+          objectSymbol && isNotImported(objectSymbol, currentSourceFile);
 
         if (
-          objectSymbol &&
-          nativelyBoundMembers.has(getMemberFullName(node)) &&
-          isNotImported(objectSymbol, currentSourceFile)
+          notImported &&
+          nativelyBoundMembers.has(`${object.name}.${property.name}`)
```
auvred, line 169:
```ts
const { log } = console;
```
In our current configuration, the `console.log` declarations are under `@types/node`. So if we remove this check, the rule will report on this case. Not sure what's the best solution there..

kirkwaiblinger (commented):
A couple questions/commenting requests, but no significant concrete changes being requested at this time.

packages/eslint-plugin/src/rules/unbound-method.ts
auvred: > you'd like to get rid of the first half of this function and only have the part after the `return` statement, but that that would cause bugs to do so?

Yup, exactly!

I'd prefer to use only type checking here for obvious reasons, as it is more flexible and robust than just checking identifier names.

But this type checking-based approach works by checking whether a method is defined in the default library. This can be a problem for us, especially in the case of `console` methods.

Imagine the following situation:

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "lib": ["ESNext"],
    "types": ["node"]
  }
}
```

```ts
// index.ts
const { log } = console
```

The `log` method of the `Console` interface is defined in the `node_modules/@types/node/console.d.ts`, not in the default TypeScript library. Therefore, the rule will false-positively report on this destructuring because it knows that this method is defined outside the default lib.

packages/eslint-plugin/src/rules/unbound-method.ts
```diff
@@ -152,69 +147,145 @@ export default createRule<Options, MessageIds>({
...
         ) {
+          return true;
+        }
+      }
+
+      return (
+        isBuiltinSymbolLike(
+          services.program,
+          services.getTypeAtLocation(object),
+          [
```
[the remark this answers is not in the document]
auvred: > Do we need both?

I really don't know :slightly_frowning_face: 

For more reasoning on this, see https://github.com/typescript-eslint/typescript-eslint/pull/8952#discussion_r1597644799


> Perhaps giving it a good name will make it clear?

+1 :+1:  Will rename it

packages/eslint-plugin/src/rules/unbound-method.ts
kirkwaiblinger: Hmm. I'm not sure I have enough knowledge to know what to do here. Maybe someone from @typescript-eslint/triage-team will have a suggestion?

JoshuaKGoldberg: Just to make sure I understand: we're saying that:

1. `console`'s `log` is a known exception defined in `lib.dom.d.ts`
2. A `log` override is defined in `@types/node`'
3. We'd like to still allow the exception for `console`'s `log`, even though type info thinks it's a `@types/node` thing

...is that right?

JoshuaKGoldberg: If it's intentional that you can `const { log } = console; log("...");`, then `log` being defined as a _method_ rather than a _property_ is a typings bug. Both in browsers and Node. We can verify that the unbound call is allowed in both - so both have a bug.

The `"DOM"` types have exceptions for built-in bugs like this because:

* The TypeScript team is unwilling to fix many of the dom type issues because that risks unexpected changes
    * I can't find a reference for one we've reported quickly, but #18286 is a general example
* They're built-in -> so we know they're tied to TypeScript releases

But, `@types/` packages from DefinitelyTyped I'm under the impression we've traditionally leaned towards modifying:

* It's much easier to change them ad-hoc: e.g. https://github.com/DefinitelyTyped/DefinitelyTyped/issues/24193
* They're a bit closer to a third-party extension in `node_modules/` (albeit one with built-in handling) than true built-in types

The fact that [`@typescript-eslint/unbound-method` doesn't report on `@types/node` with `lib: ["esnext"]`(https://typescript-eslint.io/play/#ts=5.2.2&fileType=.tsx&code=JYWwDg9gTgLgBAbzgGwgczgXzgMyhEOAIgDsIATAUwC4BjCEgZwmUqIG4AoT1NACgCU7IA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJgIwHsOATWgFtE%2BABbc%2B6KImjRu0SODABfEMqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkqyARhWANqTolHoAeekAugBpwNSCTyZk0PADkArqlQBhABbpoAazKUJc9CIC%2BIA0A&tokens=false) I would even say is a _bug_ we should follow up on:

```ts
import { log } from "node:console";

log();
```

Thoughts?

auvred: Yeah, that makes sense. So, I guess the plan would be as follows:

- Make exceptions only for known methods from the default TS library.
- Remove purely syntactic checks so that we only rely on type-level checks
- For now, treat types coming from `@types/*` as invalid and let the rule report them
- Report invalid types on the DT side

Am i right?

---

I wonder if this would be too breaking a change for people, though. We'll make them update `@types/node` to get rid of lint errors in Node.js projects.

JoshuaKGoldberg: I would say that _Report invalid types on the DT side_ should be done ASAP, so folks keep stay up-to-date-ish dev dependencies are less likely to be impacted. And maybe delaying merging the change on our end till a couple of months after that? Would that be overkill?

kirkwaiblinger: My opinion - keep the first check so we can merge without regression (even though it's "not ideal"). File DT issue, and put a link to it in code with an explanation that the first check is necessary due that issue.

auvred: Oops, I just realized that `Console` is an interface, and we can't rely on function signatures declared in interfaces, because classes implementing those interfaces can still access `this` even if `this: void` is declared in an interface:

```ts
interface Aaa {
  unbound: (this: void) => void
}

class Foo implements Aaa {
  data = 'string'
  unbound() {
    console.log(this.data) // runtime error!
  }
}

// not reported by unbound-method rule
const { unbound } = {} as Aaa
unbound()
```

[playground](https://typescript-eslint.io/play/#ts=5.5.2&fileType=.ts&code=JYOwLgpgTgZghgYwgAgIJzsg3gKGcgVxACMB7IgEwC5kAKMAC2AGcaA3U4CgSmQF4AfMg5ccAXxw4EAGzjNmyAGKlSyYAFsADtIjqI4Beky58FOGEx9kAcmZgooAObW8hEuRAVavE-mQJSEGZSHQA6aVJHeiZmULMLXgB6ROQoIjANFGgoUigAQlcJCRxk5BBSMFSITVzICmRiAE83MkoAWj1GUnq0nSlAu2wWj3qxfmwxuTQMHCJWz28gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJgIwHsOATWgFtE%2BABbc%2B6KImjRu0SODABfEMqA&tsconfig=&tokens=false)

This is a bug in a rule. See https://github.com/typescript-eslint/typescript-eslint/issues/3503. The rule should rely on function signatures only when dealing with classes, not with interfaces.

So even if we ask DT to change the types on their side, we have to make exceptions for those types because they do not guarantee runtime safety.

---

So I think now we have two options from https://github.com/typescript-eslint/typescript-eslint/pull/8952#discussion_r1657523233:

> 1. We can use the current method: pure AST node name comparison - the `log` method of `console` variable was called
> 
> 2. We can try to do some heuristics based on the `.d.ts` file path: for example, if it includes `node_modules` or something like that

Personally, I'm in favor of the second option, because it covers more cases and does so more reliably. On the other hand, pure AST checks have been around for a long time and no one has complained about them yet.

What do you think of that?

kirkwaiblinger: Similarly to my opinion in https://github.com/typescript-eslint/typescript-eslint/pull/8952#discussion_r1667775812, I think that Option 2 here is just out of scope for _this_ issue, granted that existing checks are based on Option 1 -like behavior. So my preference would be to go with Option 1 and discuss Option 2 in a followup.

kirkwaiblinger (approved):
I think we're there! Nice work!
engage

JoshuaKGoldberg (approved):
So thorough!! 🙌

PR #8953, opened 2024-04-19
docs: fix bad glob in Shared_Configurations.mdx doc

Faithfinder:
Hey, so, I noticed that docs mention a wrong glob in Shared Configs - `*.js` would only match files in root directory, it should instead be `**/*.js` if we want to match all JS files


JoshuaKGoldberg (approved):
Thanks!

PR #8954, opened 2024-04-19
chore: [disable-type-checked-config] add rule

### LINKED ISSUE #8955: `@typescript-eslint/explicit-function-return-type` does not work in .js files, need preset config for JS files

### Description

`@typescript-eslint/explicit-function-return-type` is not a "type-checked" rule but _could_ belong to the disableTypeChecked config since it does not make sense to require a return type in a context where types are not checked / mandatory, e.g. in `.js` files.

### Impacted Configurations

disableTypeChecked config

### Additional Info

_No response_

OlivierZal:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8955

## Overview

`@typescript-eslint/explicit-function-return-type` is a typed rule and then should belong to the `disableTypeChecked` config.

In particular this rule does not make sense for `.js` files.

### DIFF
--- a/packages/eslint-plugin/src/rules/explicit-function-return-type.ts
+++ b/packages/eslint-plugin/src/rules/explicit-function-return-type.ts
@@ -35,6 +35,7 @@ export default createRule<Options, MessageIds>({
     docs: {
       description:
         'Require explicit return types on functions and class methods',
+      requiresTypeChecking: true,
     },
     messages: {
       missingReturnType: 'Missing return type on function.',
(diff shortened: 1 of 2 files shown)


bradzacher:
> is a typed rule

No, it isn't. It is a pure syntactic rule.
It only applies to files that support TS syntax, yes, but it's not a type-aware rule.

---

Additionally as per the big comment at the top of the file - that file is auto-generated - so any manual changes like this one will be lost.

---

Finally we ask that contributors file issues before making such changes so that they can be discussed first.

PR #8972, opened 2024-04-22
feat(utils): swap LegacyESLint out for FlatESLint as ESLint export

### LINKED ISSUE #8970: Enhancement(utils): Swap FlatESLint-ESLint, FlatRuleTester-RuleTester in API

### Relevant Package

utils

### Description

From #8211: we have references to what `FlatESLint` and `FlatRuleTester`, which are the old terms for what ESLint now just calls `ESLint` and `RuleTester`, respectively. We'll need to:

1. Make sure any export like `LegacyESLint as ESLint` is exported as `Legacy*`
1. Make sure any export like `FlatESLint` is renamed to `FlatESLint as ESLint` (or just `ESLint`)

I suppose we could also remove the `Legacy*` classes too? Though ESLint v9 still does support the legacy configs if you opt into them...

### Additional Info

This is not a `good first issue`. Someone who's already familiar with the classes & utilities in question should tackle this.

💖

JoshuaKGoldberg:
BREAKING:
Changes the names of exports.

## PR Checklist

- [x] Addresses an existing open issue: fixes #8970

## Overview

Applies just a quick export name change so importing `ESLint` gives you `FlatESLint`, not `LegacyESLint`.

Also updates the _TODO_ comment to reflect keeping the legacy class around.

💖

### DIFF
--- a/packages/utils/src/ts-eslint/ESLint.ts
+++ b/packages/utils/src/ts-eslint/ESLint.ts
@@ -1,11 +1,11 @@
 /* eslint-disable @typescript-eslint/no-namespace */
 
 export {
-  // TODO - remove this in the next major
+  // TODO(v9) - remove this in the next major
   /**
-   * @deprecated - use FlatESLint or LegacyESLint instead
+   * @deprecated - use ESLint instead
    */
-  LegacyESLint as ESLint,
+  LegacyESLint,
 } from './eslint/LegacyESLint';
 export { FlatESLint } from './eslint/FlatESLint';
-export { LegacyESLint } from './eslint/LegacyESLint';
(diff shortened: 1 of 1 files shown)


packages/utils/src/ts-eslint/ESLint.ts
```diff
@@ -1,11 +1,11 @@
 /* eslint-disable @typescript-eslint/no-namespace */
 
 export {
-  // TODO - remove this in the next major
+  // TODO(v9) - remove this in the next major
```
bradzacher, line 4: better comment would be to remove this when eslint no longer exports it (which will likely be their next major)

bradzacher (approved):
I think

PR #8977, opened 2024-04-23
feat(eslint-plugin): split no-empty-object-type out from ban-types and no-empty-interfaces

### LINKED ISSUE #8700: Enhancement: [ban-types] Split the {} ban into a separate, better-phrased rule

https://typescript-eslint.io/rules/ban-types/

### Description

Filing this issue as a followup/superset of the more targeted #8697. As of https://github.com/microsoft/TypeScript/pull/49119, TypeScript has much better handling of `{}` semantics than it did when the `ban-types` default options were written. @RyanCavanaugh -the dev lead for TypeScript- called out in https://github.com/microsoft/TypeScript/issues/57735#issuecomment-2000069787 that as it stands today, `{ }` is a valid type with a valid meaning in TypeScript.

Proposal: let's re-think the default options in `ban-types` to agree with the way TypeScript now works?

My first proposal would be, roughly...

* Remove any bans of `{}` altogether, including in `recommended` and `strict`
* Add a ban on `NonNullable<unknown>` that suggests switching to `{}`

...but I'd want to hear from @bradzacher and @RyanCavanaugh on whether that satisfies the intents of both TypeScript and `ban-types`.

### Fail

```TypeScript
let nonNullish: Object;
```

### Pass

```TypeScript
let nonNullish: {};
```

### Additional Info

For clarity, the intent of the two issues I'm filing are:
* This one: holistically rethinking `ban-types`, which might take some time
* #8697: spot improving the `recommended` experience, which can land first

If this issue is resolved before #8697, that's great too.

JoshuaKGoldberg:
BREAKING CHANGE:
Changes the default options of a lint rule

## PR Checklist

- [x] Addresses an existing open issue: fixes #8700

## Overview

Per https://github.com/typescript-eslint/typescript-eslint/issues/8700#issuecomment-2073061412, splits out the banning of `{}` from the `ban-types` rule into a new `no-empty-object-type` rule.

Screenshot of the rule's new report message on a {} type in a VS Code hover card. Quick fixes are available.


packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg, line 1: I'd originally intended to add a lot more content here below the fold... but ended up not finding any of it particularly useful. The core issue of _"it's non-nullish value, not object value"_ is pretty straightforward to explain. Is there anything else we should definitely mention?

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,65 @@
...
+Often, developers writing `{}` actually mean either:
+
+- `object`: representing any _object_ value
+- `unknown`: representing any value _other than `null` and `undefined`_
+
+To avoid confusion around the `{}` type allowing non-object values, this rule bans usage of the `{}` type.
+
+:::tip
+If you do have a use case for an API allowing any _non-nullish value_, you can always use an [ESLint disable comment](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) or [disable the rule in your ESLint config](https://eslint.org/docs/latest/use/configure/rules#using-configuration-files-1).
+:::
```
JoshuaKGoldberg, line 28: We don't normally add this tip so prominently to rule docs (it's true for many of them!). But I figured this rule is a special case.

packages/eslint-plugin/src/rules/ban-types.ts
```diff
@@ -82,32 +85,6 @@ const defaultTypes: Types = {
       'If you are expecting the function to accept certain arguments, you should explicitly define the function shape.',
     ].join('\n'),
   },
-
-  // object typing
-  Object: {
```
JoshuaKGoldberg, line 87: We didn't explicitly discuss `Object`, but I think it makes sense in the spirit of this change to switch it to being treated as another banned uppercase alias. I've never seen people confuse `Object` for `{}`.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,65 @@
...
+`{}` represents any _non-nullish value_, including literals like `0` and `""`:
+
+```ts
+let anyNonNullishValue: {} = 'Intentionally allowed by TypeScript.';
+```
+
+Often, developers writing `{}` actually mean either:
+
+- `object`: representing any _object_ value
+- `unknown`: representing any value _other than `null` and `undefined`_
```
bradzacher, line 22: IMO this should include a note about "an object with no properties" and how there isn't a good representation/type for it in TS due to structural typing.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,66 @@
...
```
kirkwaiblinger, line 22:
```suggestion
- `NonNullable<unknown>`: representing any value _other than `null` and `undefined`_
```

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
kirkwaiblinger: If I were writing this, I would add a few things. Make of these what you will in light of my bias against `{}` and tendency for undue verbosity 😄 

- I would mention that any non-nullish value is what `{}` _presently_ means, and reference or allude to the fact that it is now checked much more safely than it once was, due to changes in TS. 

- For the line on why you might think that `{}` is an `object`, I would consider using this as a teachable moment for what `{ <etc> }` syntax in general means and give intuition for why `{}` means what it does... Eg, 
   	- `object`: representing any _object_ value. This might seem plausible since we usually use the `{ property: string }` syntax for things that are always `object`s, but, actually, primitives are assignable to these types as well, so long as they have all the declared properties! For example, `const notAnObject: { toFixed(): string } = 12.345;` is completely valid. Therefore, _anything_ (other than `null` and `undefined`) satisfies `{}`.

 - Similarly, consider adding something like "by analogy to the empty tuple type `[]`" for why you might think that `{}` is a closed/literal object.

- Somewhere also mention that this can also occur in the form `() => {}` when `() => void` is meant?

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,66 @@
...
+`{}` represents any _non-nullish value_, including literals like `0` and `""`:
+
+```ts
+let anyNonNullishValue: {} = 'Intentionally allowed by TypeScript.';
+```
+
+Often, developers writing `{}` actually mean either:
+
+- `object`: representing any _object_ value
+- `NonNullable<unknown>`: representing any value _other than `null` and `undefined`_
```
kirkwaiblinger, line 22: Sorry, my previous suggestion was maybe a little incomplete. 
```suggestion
- `NonNullable<unknown>`: representing any value _other than `null` and `undefined`_. In most contexts this is the same as `{}`, but more explicit
- `unknown`: representing any value at all, including `null` and `undefined`
```

bradzacher:
Mildly tangential - but I can envision the following options:

- `allowInTypeAliasWithName` - for react codebases to allow `type MyProps = {}`
- `allowInIntersection` - to allow `type T = A & {}` - wherein the `{}` falls away.
    - does this even need an option...? Probably should be the default?

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
kirkwaiblinger: The reason I think it helps to explain well the rationale behind the 4 plausible meanings of `{}` is that I can easily imagine disabling the rule, because _I know_ what `{}` means, and _I_ would never misuse it, and then autopilot write `() => {}` in a type context without my brain registering I wrote the `{}` type at all. (Not that I've ever done that before 🤣 )

packages/eslint-plugin/src/rules/no-empty-object-type.ts
```diff
@@ -0,0 +1,45 @@
...
+    hasSuggestions: true,
+    messages: {
+      banEmptyObjectType: [
+        'The `{}` ("empty object") type allows any non-nullish value, including literals like `0` and `""`.',
+        "- If that's what you want, disable this lint rule with an inline comment or in your ESLint config.",
+        '- If you want a type meaning "any object", you probably want `object` instead.',
+        '- If you want a type meaning "any value", you probably want `unknown` instead.',
+        '- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.',
+      ].join('\n'),
+      replaceEmptyObjectType: 'Replace `{}` with `{{replacement}}`.',
```
bradzacher, line 22: Do we want to use the dot points above as the suggestion fixer messages for more clarity?

kirkwaiblinger:
Isn't no-empty-object-type a bit of a misnomer since the whole point of the rule that it isn't necessarily an object? Maybe no-empty-structured-type? no-empty-curly-brace-type?

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
kirkwaiblinger: Agreed! Though maybe "If you do have a use case for the `{}` type" would be preferable, since we also already suggest `NonNullable<unknown>` as a way to specify "any nonnulish value"? Your call!

JoshuaKGoldberg: Hmm yeah I'm on the fence, leaning as-is... we do mention the `{}` in the previous line. I like having it different here both for more pleasant (less repetitive) reading and to enforce that they mean roughly the same thing.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg, line 1: @bradzacher threading https://github.com/typescript-eslint/typescript-eslint/pull/8977#issuecomment-2073549022's:

> `allowInTypeAliasWithName` - for react codebases to allow `type MyProps = {}`

Isn't that a code smell? Why would they declare an empty props type? The only reason I can think of is copypasta / using a template, but it's "dead" code at that point.

packages/eslint-plugin/src/rules/no-empty-object-type.ts
JoshuaKGoldberg: Ehhh not really IMO. They'll have to go through the big report message anyway to get here.

packages/eslint-plugin/docs/rules/no-empty-interface.mdx
JoshuaKGoldberg, line 1: @kirkwaiblinger threading https://github.com/typescript-eslint/typescript-eslint/pull/8977#issuecomment-2073563749:

> Isn't no-empty-object-type a bit of a misnomer since the whole point of the rule that it isn't necessarily an object? Maybe no-empty-structured-type? no-empty-curly-brace-type?

The community generally calls it the "empty object type" I'm afraid. Maybe that's a byproduct of early folks not understanding the structural type system. But I can't think of another clear, succinct name that is in common usage. 😬

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
Josh-Cena: @JoshuaKGoldberg It happens if you are using `forwardRef` and your component takes a second `ref` arg. Of course `props: unknown` would also work if you are not using it anyway...

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+To avoid confusion around the `{}` type allowing any _non-nullish value_, this rule bans usage of the `{}` type.
+
+:::tip
+If you do have a use case for an API allowing `{}`, you can always use an [ESLint disable comment](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) or [disable the rule in your ESLint config](https://eslint.org/docs/latest/use/configure/rules#using-configuration-files-1).
+:::
+
+Note that this rule does not report:
+
+- `{}` as a type constituent in an intersection type (e.g. `type NonNullable<T> = T & unknown`), as this is useful in type system operations.
```
Josh-Cena, line 36:
```suggestion
- `{}` as a type constituent in an intersection type (e.g. `type NonNullable<T> = T & {}`), as this is useful in type system operations.
```

?

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,74 @@
...
+- `{}` as an empty function body with a `void` return type (e.g. `() => {}`), as this is a function body, not a type.
```
Josh-Cena, line 37: This feels weird... Why specifically this? Then what about empty object literals, empty blocks, etc.?

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,74 @@
...
+Projects that extensively use type operations such as conditional types and mapped types oftentimes benefit from disabling this rule.
+
+## Further Reading
+
+- [Enhancement: [ban-types] Split the {} ban into a separate, better-phrased rule](https://github.com/typescript-eslint/typescript-eslint/issues/8700)
+- [The Empty Object Type in TypeScript](https://www.totaltypescript.com/the-empty-object-type-in-typescript)
+
+## Related To
+
+- [`no-empty-interface`](./no-empty-interface.mdx)
```
Josh-Cena, line 74: I feel like these two rules are really one, maybe as an option `allowInterface` for declaration merging

packages/eslint-plugin/docs/rules/ban-types.mdx
```diff
@@ -74,9 +68,6 @@ The default options provide a set of "best practices", intended to provide safet
 - Avoid the `Function` type, as it provides little safety for the following reasons:
   - It provides no type safety when calling the value, which means it's easy to provide the wrong arguments.
   - It accepts class declarations, which will fail when called, as they are called without the `new` keyword.
-- Avoid the `Object` and `{}` types, as they mean "any non-nullish value".
```
Josh-Cena, line 77: We still need to discuss `Object` separately because it's not an "uppercase primitive type".

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
kirkwaiblinger: I'm perplexed by the wording, too...

I'm guessing this came in part due to comments I've made about being annoyed that `const foo = () => {}` is not assignable to `type Foo = () => {}`. I think to make sense, the type vs value thing would have to be spelled out a little more.

However, in the context of the whole doc, I think it might flow better to just omit this line tbh.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
bradzacher: You have two choices - declare nothing at all `function Component() {}` or declare an empty something `function Component(_: Props) {}`

A lot of people like doing the latter as it gives them an extension point to build from - eg if they know for sure they are going to add props later.

Forward refs is another good example of when it's hard required.

packages/eslint-plugin/docs/rules/ban-types.mdx
kirkwaiblinger: i think the thought here was to keep it simple and sweep that caveat under the rug a bit: https://github.com/typescript-eslint/typescript-eslint/pull/8977/files#r1576801720

packages/eslint-plugin/docs/rules/ban-types.mdx
kirkwaiblinger, line 1: should a "Related To" link to no-empty-object-type get added to this too? (since users may remember that that used to be part of this rule and end up looking through this doc page for `{}`)

kirkwaiblinger (commented):
Big picture here looks good to me. Added a few more questions and suggestion, but basically consider this a thumbs up from me 👍  

I want to say that I really like how you managed to keep the doc page concise on this nuanced topic, and encouraging about disabling if it helps more than hurts.

packages/eslint-plugin/docs/rules/ban-types.mdx
JoshuaKGoldberg: Yeah, given how rare `Object` is in the wild I think we're safe to sweep?

packages/eslint-plugin/src/rules/no-empty-object-type.ts
```diff
@@ -0,0 +1,53 @@
...
+      recommended: 'recommended',
+    },
+    hasSuggestions: true,
+    messages: {
+      banEmptyObjectType: [
+        'The `{}` ("empty object") type allows any non-nullish value, including literals like `0` and `""`.',
+        "- If that's what you want, disable this lint rule with an inline comment or in your ESLint config.",
+        '- If you want a type meaning "any object", you probably want `object` instead.',
+        '- If you want a type meaning "any value", you probably want `unknown` instead.',
+        '- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.',
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Hmm yeah I like keeping to the plain text. I don't think I've ever seen a correct+legitimate use case for `Record<string, never>` from someone who didn't deeply understand the type system. +1

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg: Hmm, I see them as separate rules. They're definitely related but I think the justifications for disabling either are very different.

packages/eslint-plugin/docs/rules/ban-types.mdx
JoshuaKGoldberg: Like, @Josh-Cena I actually thought you were in favor of sweeping the `Object` stuff under the rug - maybe I misinterpreted your messages in our private channel. For this discussion's visibility, what do you see as the reason why we wouldn't want to do this?

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg: Why wouldn't you either omit the props type for `forwardRef` or use `object`?

packages/eslint-plugin/docs/rules/ban-types.mdx
Josh-Cena: I'm not sure what you mean by "sweeping under the rug". Like, we still have logic in place to handle it, and we have error messages for it. All I'm saying is, if you remove this sentence, then it reads to me as "`Object` will not be handled", because the statement "Don't use the upper-case primitive types, you should use the lower-case types for consistency." does not encompass `Object`: `Object` is not an "upper-case primitive type". You can say "Don't use the upper-case primitive types or `Object`", maybe.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg: Interesting. So just confirming, you're suggesting we should merge `no-empty-interface` into this rule for v8, too?

Josh-Cena: Yes, exactly.

packages/eslint-plugin/src/rules/no-empty-object-type.ts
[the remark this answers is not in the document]
kirkwaiblinger: Ok, so granted it's been removed, you'll want to remove 
1. The suggested fix and associated tests: https://github.com/typescript-eslint/typescript-eslint/blob/c01f10e15eaa4bdc9b91a2f50d4e1b1c92eba8f6/packages/eslint-plugin/src/rules/no-empty-object-type.ts#L40
2. The exmple on the doc page: https://github.com/typescript-eslint/typescript-eslint/blob/c01f10e15eaa4bdc9b91a2f50d4e1b1c92eba8f6/packages/eslint-plugin/docs/rules/no-empty-object-type.mdx#L53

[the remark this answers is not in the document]
kirkwaiblinger: (unresolving for the followup cleanup)

packages/eslint-plugin/docs/rules/ban-types.mdx
JoshuaKGoldberg: Oh! Ok phew yes 😄 I think I was tired and getting confused. Yes, that makes a lot of sense. 👍 will do!

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg: Slept on it and ruminated - I'm ever so slightly leaning away from wanting to merge. But can't really make up my mind. 🫠

On one side, I do like that they're essentially targeting the same motivation, as you noted: that any non-nullable value is assignable to the `{}` shape. So it does make sense to me to put them together.

On the other hand, I think specifically the object type alias form is kind of controversial, and a lot of users will want to be able to disable `no-empty-object-type`. Maybe less so now that we're fixing the docs but I'm still guessing a non-trivial portion of users. I'm also thinking that those users likely won't want to disable `no-empty-interface`.

We could solve that by adding an `allowTypeAlias` option to the rule... but then we'd also have to mention that in the report message. Which IMO is getting to the point of a little more complication than is ideal. This is a big part of why I've started leaning towards smaller, less configurable lint rules. There might be more of them to lean but each gets much simpler to understand.

So my current proposal would be: leave separate in v8 for ideation, then maybe merge in v9 if we think it'd be valuable?

bradzacher: the difference between the two is that `no-empty-interface` allows you to ban `interface Foo extends Bar {}` in favour of `type Foo = Bar`.
However a similar rule doesn't make sense for type aliases.

But there's no reason we couldn't merge them and keep the `allowSingleExtends` option just for interfaces.

but you're JoshG isn't wrong - the scope is different because the current form bans `{}` anywhere - not just in `type` aliases.

I personally can't think of a reason you'd ever want to have one banned, but not the other.

bradzacher:
Also worth noting that we could land this as a standalone rule for v7 and then in v8 we change the defaults and make this new rule recommended.

JoshuaKGoldberg:
> standalone rule for v7

I like that a lot! Regardless of what we choose for `no-empty-interface`, this lets us get it released faster. Excellent. I'll get that PR up now.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg: > best describes the intent (an object with more properties to be added).

But it doesn't describe just objects, right? It's an object _or_ non-nullish primitive? I think in this case what we're seeing is the rule acting as it should: telling folks not to use `{}` because it also allows those primitives.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
kirkwaiblinger: no strong opinions on this!

kirkwaiblinger: I will say, though, it does feel like yet another breaking change for users just to preserve the existing behavior. So, I guess "no strong opinion" defaults to not merging in this case 🤷

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
bradzacher: In the case of a react component - the type of the first argument is intersected with the standard framework provided JSX props by TS at the JSX "callsite". This is done automatically by TS - it intersects with `global.JSX.IntrinsicAttributes` IIRC.

EG if you defined `function C(props: P) {}` then render `<C />` - the props you're allowed to pass to `C` are `P & global.JSX.IntrinsicAttributes`.

AFAIK all "JSX" frameworks define at least `{ key?: string }` so the resulting type will **_ALWAYS_** be an object and not a primitive.

Additionally it's JSX - you can pass nothing but an object. JSX attributes are syntactic sugar for an object and a JSX spread attribute is enforced by TS to be an object.

So yeah you *could* do `function C(_: unknown, ref: ForwardedRef<x>) {}` - but that would be more verbose and more annoying than using `_: {}` or `_: Props`.

I'm all for us continuing to report on `_: {}` - but it's very common to do
```ts
type CompProps = {};
function Comp(_: Props, ref: X) {}
```

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
bradzacher: > it does feel like yet another breaking change for users just to preserve the existing behavior

I would have said the opposite - it's a breaking change TO merge.
Cos if you don't merge - people are just switching from `ban-types` to this new rule.
But if you merge they ALSO need to switch from `no-empty-interface` to this new rule.

packages/eslint-plugin/src/rules/no-empty-object-type.ts
```diff
@@ -0,0 +1,166 @@
...
+            .get(node.id.name)
+            ?.defs.some(
+              def => def.node.type === AST_NODE_TYPES.ClassDeclaration,
+            );
+
+          if (extend.length === 0) {
+            context.report({
+              data: { option: 'allowInterfaces' },
+              node: node.id,
+              messageId: 'noEmpty',
```
bradzacher, line 96: do we want a different message for interfaces?
otherwise the report can be a little confusing:
```ts
interface Foo {}
^^^^^^^^^^^^^^^^ The `{}` ("empty object") type allows any non-nullish value, including literals like `0` and `""`.
```

"But I'm not using the empty object type - I'm declaring an interface! What gives?!?"

Might be good to parameterise this for the two usecases eg
``The `{}` ("empty object") type ...``
and
``An empty interface declaration ...``

bradzacher (approved):
general logic LGTM

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg: I'm still unconvinced 😄. The `CompProps` example is IMO even more in-violation because it (a) suggests doing that kind of type (`= {}`) is ok and (b) is something that can be reused outside the component. The fact that the framework / its types make the types safe is nice, but doesn't stop folks from reusing the dangerous type aliases outside the framework.

```ts
let myPropsToBeUsedLater: CompProps;

myPropsToBeUsedLater = "oops";
```

Is there maybe a more targeted option we could go for than the broad `allowInTypeAliasWithName`?

If someone explicitly says _"Josh, we're in majority, we'd like to override your veto, please implement this"_ then I won't be upset at all. Nor would I be upset at all if you just push to this branch 🙂. Either way I'm expecting this rule to be one of the ones that gets a lot of scrutiny in the testing period for v8.

kirkwaiblinger (approved):

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+## When Not To Use It
+
+If your code commonly needs to represent the _"any non-nullish value"_ type, this rule may not be for you.
+Projects that extensively use type operations such as conditional types and mapped types oftentimes benefit from disabling this rule.
+
+## Further Reading
+
+- [Enhancement: [ban-types] Split the {} ban into a separate, better-phrased rule](https://github.com/typescript-eslint/typescript-eslint/issues/8700)
+- [The Empty Object Type in TypeScript](https://www.totaltypescript.com/the-empty-object-type-in-typescript)
```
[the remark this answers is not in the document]
mattpocock: Thanks pal, much appreciated!

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
bradzacher: > Josh, we're in majority, we'd like to override your veto, please implement this

Declaring a stand-alone, props type with an empty object:
[`/type \w+Props = {}/ or /interface \w+Props {}/`](https://sourcegraph.com/search?q=context:global+/type+%5Cw%2BProps+%3D+%7B%7D/+or+/interface+%5Cw%2BProps+%7B%7D/+count:all&patternType=keyword&sm=0) -> Found more than 99382 results from 4074 repositories.

Declaring an empty props type inline:
[`/function [A-Z]\w+\((props|_\w*): {}/ or /(const|let|var) [A-Z]\w+ = \((props|_\w*): {}/`](https://sourcegraph.com/search?q=context:global+/function+%5BA-Z%5D%5Cw%2B%5C%28%28props%7C_%5Cw*%29:+%7B%7D/+or+/%28const%7Clet%7Cvar%29+%5BA-Z%5D%5Cw%2B+%3D+%5C%28%28props%7C_%5Cw*%29:+%7B%7D/+count:all&patternType=keyword&case=yes&sm=0) -> Found more than 1272 results from 164 repositories.

Declaring props with `unknown`:
[`/function [A-Z]\w+\((props|_\w*): unknown/ or /(const|let|var) [A-Z]\w+ = \((props|_\w*): unknown/`](https://sourcegraph.com/search?q=context:global+/function+%5BA-Z%5D%5Cw%2B%5C%28%28props%7C_%5Cw*%29:+unknown/+or+/%28const%7Clet%7Cvar%29+%5BA-Z%5D%5Cw%2B+%3D+%5C%28%28props%7C_%5Cw*%29:+unknown/+count:all&patternType=keyword&case=yes&sm=0) -> Found more than 184 results from 91 repositories.

Some hard data to convince you that this is a very, very common pattern.
It's 10x more common to use an inline empty object.
It's 1000x more common to use a separate, empty prop type.

> but doesn't stop folks from reusing the dangerous type aliases outside the framework.

In react it's just rare that you use the prop types like this.
When the prop type is exported it's most often used like this:
```ts
import { type OtherProps, Other } from './other';

type Props = { ... } & OtherProps;
function Component(props: Props) {
  return (
    <div>
      <Other {...props} />
    </div>
  );
}
```

It's the very rare case that you would do:
```ts
const customProps: OtherProps = <a value>;
<Other {...customProps} />
```
And instead much more common to do:
```ts
<Other prop1={value} ... />
```

You're not wrong that there is *some* unsafety here - but it's rare that the types are used this way.

---

I don't love it either!
You know how much I **_hate_** fuzzy options with vague name-based matches.
But sadly there's no enforced standard for prop type names - so we can't really enforce anything ourselves. EG we can't do something like add `ignoreReactPropTypes` that makes the rule ignore all `*Props` named types.

The issue is that right now people opt react codebases out of the entire check with `{ extendsDefault: true, types: { '{}': false } }` because there's no configurability or way to say "hey yo react prop types are safe please ignore them".

I think that an option sucks hard - but I can only see two options:
1) add the option to allow people to opt-out specific cases but have the rule on in general
2) do nothing and have people turn the entire rule off

I think there's massive value in generally banning `{}` (which is why I've pushed back so hard over the years to keep `{}` in the `ban-types` defaults) - so I think that creating a world where more people can have the rule on is worth the loose bullshittery.

JoshuaKGoldberg: Okie dokie. I added it as an option to `allowObjectTypes`.

It sounds like we're on the same page on this being "loose bullshittery" (😄). I'm happy to make the compromise in the spirit of getting option 1, not option 2.

bradzacher: >  I added it as an option to allowObjectTypes.

Could we make it a regex check to narrow it?
EG `allowedTypeNames: /.*Props?$/` would be much less loose than just "always in type aliases".

I think we also would want to do it for interfaces as some codebases like using type aliases and others interfaces - so allowing both `type FooProps = {}` and `interface FooProps {}` would be valuable.

JoshuaKGoldberg: Heh, yeah, I was thinking of filing that as a followup issue. +1 to at least thinking it through here. Added a general `allowWithName` to align in naming with the other two options.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,151 @@
...
+interface Base {
+  value: boolean;
+}
+
+interface Derived extends Base {}
+```
+
+### `allowObjectTypes`
+
+Whether to allow empty object type literals, as one of:
```
kirkwaiblinger, line 110: I think this list is stale now, due to the new option being added, right? Or, I guess, the way it's written, it doesn't seem to me like `allowWithName` is part of the list.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+</TabItem>
+</Tabs>
+
+## Options
+
+By default, this rule flags both interfaces and object types.
+
+:::warning
+We recommend keeping all options to the default and not allowing exceptions to this rule.
```
kirkwaiblinger, line 85: (optional if you want to do anything with this)
It feels like this is overstating it a bit granted we're not all in agreement on this, and the options are specifically added in order to make it more likely for users to keep it at least partially enabled.

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,151 @@
...
+To avoid confusion around the `{}` type allowing any _non-nullish value_, this rule bans usage of the `{}` type.
+That includes interfaces and object type aliases with no fields.
+
+:::tip
+If you do have a use case for an API allowing `{}`, you can always configure the [rule's options](#options), use an [ESLint disable comment](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1), or [disable the rule in your ESLint config](https://eslint.org/docs/latest/use/configure/rules#using-configuration-files-1).
+:::
+
+Note that this rule does not report on:
+
+- `{}` as a type constituent in an intersection type (e.g. `type NonNullable<T> = T & {}`), as this can be useful in type system operations.
```
kirkwaiblinger, line 36: (optional)
Would be cool if there's a convenient way to mention that `NonNullable` is a built-in here. I personally learned about the existence of `NonNullable` from the old ban-types message, but that part of it has been removed now

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+Whether to allow empty object type literals, as one of:
+
+- `'always'`: to always allow object type literals with no fields
+- `'never'` _(default)_: to never allow object type literals with no fields
+
+### `allowWithName`
+
+A stringified regular expression to allow interfaces and object type aliases with the configured name.
+This can be useful if your existing code style still includes a pattern of declaring empty types with `{}` instead of `object`.
```
kirkwaiblinger, line 118: Maybe soften to
```suggestion
This can be useful if your existing code style includes a pattern of declaring empty types with `{}` instead of `object`.
```
?

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg: I'm having a hard time thinking of phrasing for this 🤔 do you have any suggestions @kirkwaiblinger?

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg: Hmm yeah I'll just remove it for now. The first few paragraphs of the page are pretty extensive.

packages/eslint-plugin/src/rules/no-empty-interface.ts
```diff
@@ -17,8 +17,8 @@ export default createRule<Options, MessageIds>({
     type: 'suggestion',
     docs: {
       description: 'Disallow the declaration of empty interfaces',
-      recommended: 'stylistic',
     },
+    deprecated: true,
```
bradzacher, line 21: Don't forget `replacedBy`

bradzacher (approved):

packages/eslint-plugin/docs/rules/no-empty-object-type.mdx
JoshuaKGoldberg: Merged so that I can get other PRa going, but happy to take a followup issue or PR!

kirkwaiblinger: see https://github.com/typescript-eslint/typescript-eslint/pull/9362

JoshuaKGoldberg: oop

Issue #7889, opened 2023-11-07
Bug: [return-await] "in-try-catch" not respected with explicit resource management


andrew0:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.2.2&fileType=.tsx&code=IYZwngdgxgBAZgV2gFwJYHsI1JKAxCACgEoYBvAKBmwHdhVkYIBTGmABQCd0BbVEZgB4AbulQATAHyFCnZiHQAbYc1IBeSTAHIAKqh7N0CZLPlKVAGhgBGAAy3ixANxUYUTAsXMAdIvQBzQgByHGgCGHFMZiDnVzlkBE4sIIUDZAALVAh-IJcAXwoKUNhEFAwseMSIAEE6BhJyVwQQLP8YAH0YNUbqagBtAGUwHgAjJW9xfgAHdAEAXQAuGAaNHt7qdwhPHz9AoMmQGYEYl3WC6gK45gSk2npGYoISfMLgOsZKpNr758LmAA8ZpxGGQYHkXEA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6aRfWaJ2gQwDufSvnRRE0aAHtokcGAC%2BIBUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```typescript
async function asyncFn() {
  await new Promise<void>((resolve) => setTimeout(resolve, 50));
  return 'something';
}

async function returnAwait() {
  using _ = {
    [Symbol.dispose]: () => {
      console.log('dispose');
    }
  }

  return await asyncFn();
}
```

### ESLint Config

```javascript
module.exports = {
  parser: "@typescript-eslint/parser",
  rules: {
    "@typescript-eslint/return-await": "error",
[...]

PR #9031, opened 2024-04-30
fix(eslint-plugin): [return-await] clean up in-try-catch detection and make autofixes safe

### LINKED ISSUE #8661: Bug: [return-await] unsafe autofixes in error handling code

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gcxBgMkmANYBc8AFAJTwC8AfPFCgJ4DcAUB1GiymPABmAV34YAlqhx4AqmhAwAkigAOwjNXIAFGEgC24%2BQB5MMcSmyMA3h3jwA9Pfhp9CZKHgYAFlCwgAHuDqIGjwAPq8-F66KEjCaBAsYfDAwmYWnl4IguIwmEyRAoIwUHogtg5OzMDwevFYPioqICjwAEYggkhw8OIYAOShaFCCCBhI8ABEohBExJOZCLMkAHQVyOhY8QrKali0TADuUH3wOvqGICtwLhAAbiAUkwDKrvAqugbymb6IzPCoRLwG5Ie4IEYYBSLeDLYi9UL9GZzEDAfqTKicOxwDBpVrbJSqdScAC%2BXB4fCKojAEikmGEKnEwGeOIZwAAwkhQJozp9LiYMOlLPAbHYIHgYXM6NICHNqJjPDAWMKKnZHCkJi4ys4cYJBD8sHooErvLpDms7Fi8LjpXIdoSNBiKsShOYoBAgSKLbDriAxTxHo67KTiUA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaRAF1ml0IEMB3agS1PSnwDM3IAacbSAAVIBPAA4oAxtAYjShFPAa5SAehLlKNekxaRE0aPmiReAXxAmgA&tsconfig=&tokens=false

### Repro Code

```TypeScript
declare const getLock: () => any;

async function getUserInput(): Promise<string> {
  // some code that executes _synchronously_ during the first async frame
  // and must happen before it's safe to "unlock" the lock.
  const userInput = await Promise.resolve("Some promise that can only resolve after the lock is 'unlocked'");
  return userInput;
}

async function stupidStupidCode(): Promise<string> {
  let lock = getLock();
  try {
    // do some stuff that may throw.
    return getUserInput();
  } finally {
    lock.release();
  }
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "no-return-await": "off",
    "@typescript-eslint/return-await": "error"
  }
}
```

### tsconfig

```jsonc
undefined
```

### Expected Result

I expect that the line `return getUserInput();` should flag and give me a _suggested fix_ of `return await getUserInput();`

### Actual Result

autofixes, rather than suggestion.

### Additional Info

The programmatically fixed code results in deadlock :( Adding the `await` in that context makes the unlocking wait for the user input, but the user input is waiting for the unlocking.

I realize that the interaction between the two functions in the example is very crappy and brittle due to depending on code in an async function to execute synchronously, but... it's 100% real code that I encountered at work, and, even though we only had return-await enabled as a warning, my "autofix on save" editor settings led to this getting autofixed in a file I was working on, and I figured, eh, autofixes are safe, i don't need to think too hard about what that function does, and we released it, causing a very difficult-to-debug bug in production (total showstopper too, lol 😆).

This simply isn't safe, in my opinion, to provide as an autofix rather than a suggestion which needs to be reviewed by a human. 

Note that there _are_ things that I think are totally safe to autofix, though! For example, depending on your config for this rule, you may switch either direction between these code examples safely
```typescript
async function safeToAutofix(): Promise<void> {
    return asyncFunctionCall()
}

async function safeToAutofix(): Promise<void> {
   return await asyncFunctionCall()
}
```
It's just when error handling stuff comes in that the control flow gets messed up and can lead to very very subtle bugs.

### LINKED ISSUE #8663: Bug: [return-await] awaited promise allowed in terminal catch block

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.2&fileType=.tsx&code=IYZwngdgxgBAZgV2gFwJYHsLwBQEoYDeAUDDMgE5iEkwC%2BMUwyUAFtaaeQKbILlbAA7sFTIYABXLoAtqhBcAdNwBWXKMjwBuGrSK0gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6aRfWaJ2gQwDufSvnRgA2pETRoAe2iQANFGa180QrTJ98ZABaQAuuDABfEKaA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
async function f() {
  try {
  } catch {
    return await Promise.reject();
  }
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/return-await": ["error", "in-try-catch"]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

Expected rule to flag on awaited promise in terminal catch block. From the docs: 
> - if you `return` a promise within a `catch`, and there **_is no_** `finally`, then it **_must not_** be `await`ed.

see https://typescript-eslint.io/rules/return-await/#in-try-catch

### Actual Result

no error reported

### Additional Info
 
Strangely, there is actually a _valid_ test case right now that is asserting this incorrect behavior 
https://github.com/typescript-eslint/typescript-eslint/blob/fc08f969e255b55ea35091b297b358bc434a76eb/packages/eslint-plugin/tests/rules/return-await.test.ts#L238-L253

### LINKED ISSUE #8907: Bug: [return-await] should handle nested try-catch-finally blocks with "in-try-catch" option

### Playground Link

https://typescript-eslint.io/play/#ts=5.2.2&showAST=es&fileType=.tsx&code=IYZwngdgxgBAZgV2gFwJYHsLwBQEoYDeAUDDMgE5iEmlmXW0wC%2B8qEwANh1cY6QPT8YwOHACmUZCBhjy5dORgALYBAAmHNgHMY6OLoTJZMKMGRQlNRuTHIE5LAAV5AW1QgxAOhsArCcjwAbitmGhZTcyVqMKImIiJQSFhEFAwsOAAmPAY6HjCTMwsc0go8xhY4Nk5uYtpBYVF-aVl5RRV1TQgdPQMjRUr2LjAQ0hs7BxhndDcPbzE-SSCQuNIKqqHamKYgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6aRfWaJ2gQwDufSvnRgA2pETRoAe2iQANFGa180QrTJ98ZABaQAuuDABfEKaA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
async function f() {
  try {
    try {
    } finally {
      // affects error handling of outer catch
      return Promise.reject();
    }
  } catch {

  }
}

// or,
async function f2() {
  try {
  } catch {
    try {
    } finally {
      // affects error handling of outer finally
      return Promise.reject();
    }
  } finally {
    
  }
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/return-await": ["error", "in-try-catch"]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expect that the commented promises should flag as requiring await.

### Actual Result

no errors

### Additional Info

The `in-try-catch` deductions are generally buggy. Note that putting an `await` doesn't cause the rule to flag that it's unnecessary either, despite [the rule's documentation implying that there will only be one correct choice in each scenario](https://typescript-eslint.io/rules/return-await/#in-try-catch).

 See also https://github.com/typescript-eslint/typescript-eslint/issues/8663

Note that putting a `return` in a `catch` suffers the same defect and is probably a more realistic example for this issue.

kirkwaiblinger:
## PR Checklist

  1. fixes #8661
  1. fixes #8663
  1. fixes #8907

## Overview

These 3 bugs are closely related, and the code is largely interdependent, so I figured it would be easiest to group into 1 PR. We can do sequential PRs instead if need be. 

The diff looks like there's a lot going on, but I think it's actually not so bad, if we go step-by-step. (Then again, this is all swimming around in my head so it probably makes more sense to me than to a reader):
1. docs changes. Not tied to any specific code change, just tried to improve them somewhat.
1. Refactoring/Logic simplification - when determining whether a node is inside of a `try`, `catch`, or `finally`, we now just find the nearest `try-catch-finally` statement and work from there, rather than checking each ancestor separately. (bye-bye `inTry`, `inCatch`, `hasFinallyBlock`, `isReturnPromiseInFinally`, all replaced by `findEnclosingTryStatement`)
1. Logic change - extract error-handling logic from `in-try-catch` into a helper function that now also checks _nested_ `try-catch-finally` statements, rather than just the closest one found (`affectsExplicitErrorHandling`), fixing #8907.
1. Trivially fix the control flow in the `in-try-catch` thanks to the above helper, fixing #8663.
1. Logic change - Use the above utilities outside of the `in-try-catch` option as well, as the condition for whether to autofix or suggest a fix, fixing #8661.
1. Lots of accompanying test changes. With the exception of one _valid_ test change for #8663, all the test _changes_ are just changing autofixes to suggestions (#8661). The test _additions_ are to handle the nested `try-catch-finally`s (#8907)


packages/eslint-plugin/tests/rules/return-await.test.ts
```diff
@@ -947,76 +1096,118 @@ async function test<T>(): Promise<T> {
...
+}
+      `,
+            },
+          ],
         },
       ],
     },
     {
       code: `
-        async function bar() {}
```
kirkwaiblinger, line 967: This just tests 3 totally separable errors, so I broke it up into 3 separate tests.

packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -237,38 +281,63 @@ export default createRule({
...
+                    {
+                      messageId: 'disallowedPromiseAwaitSuggestion',
+                      fix,
+                    },
+                  ],
+                }),
           });
-        } else if (!isAwait && isInTryCatch) {
-          if (inCatch(expression) && !hasFinallyBlock(expression)) {
-            return;
```
kirkwaiblinger, line 260: The return statements were just wrong. Directly caused #8663 and played into #8907 as well.

With the `inCatch`/`hasFinallyBlock` etc logic extracted to the `affectsErrorHandling` variable, they now correctly are included in the preceeding branch (enforcing removing the await).

packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -222,13 +250,29 @@ export default createRule({
         return;
       }
 
+      const affectsErrorHandling = affectsExplicitErrorHandling(expression);
```
kirkwaiblinger, line 253: The thought is that one day there will be a `|| affectsExplicitResourceManagement(expression)` 🙏 https://github.com/typescript-eslint/typescript-eslint/issues/7889

packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -68,64 +72,88 @@ export default createRule({
...
+          if (child === ancestor.tryBlock) {
+            block = 'try';
+          } else if (child === ancestor.catchClause) {
+            block = 'catch';
+          } else if (child === ancestor.finallyBlock) {
+            block = 'finally';
+          } else {
+            throw new Error(
+              'Child of a try statement must be a try block, catch clause, or finally block',
+            );
```
JoshuaKGoldberg, line 146: [Testing] If this case can be hit, it should be unit tested and the rule shouldn't crash. But if it can't be hit then it should be removed.

packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -321,3 +355,12 @@ export default createRule({
...
 });
+
+function fixOrSuggest<MessageId extends string>(
+  useFix: boolean,
+  suggestion: TSESLint.SuggestionReportDescriptor<MessageId>,
+):
+  | { fix: TSESLint.ReportFixFunction }
+  | { suggest: TSESLint.SuggestionReportDescriptor<MessageId>[] } {
+  return useFix ? { fix: suggestion.fix } : { suggest: [suggestion] };
+}
```
JoshuaKGoldberg, line 366: [Praise] I like this :)

JoshuaKGoldberg (changes requested):
This is excellent work, nicely done @kirkwaiblinger. I tried a few times to find something major to complain about but couldn't find anything. 👏 

From me, there's just the one nit on a thrown error. Since this is a pretty tricky area I'll slap on the `1 approval` label in the hopes someone else will be able to review too.

✨

JoshuaKGoldberg (dismissed):
_wrong button_

kirkwaiblinger:
@JoshuaKGoldberg 

> _wrong button_

WRONG LEVER

packages/eslint-plugin/src/rules/return-await.ts
kirkwaiblinger: Is there a good pattern for how to handle this sort of thing? 

One of the three conditions _must_ be true, or else we have an invalid AST. But there isn't any way to indicate that with the type system that I can think of. 

We _should_ be able to replace the last branch's check with an unconditional `else`, but that hampers readability and maintainability in multiple ways IMO.

We _could_ fail gracefully and `return undefined` in the unreachable case, but that's semantically incorrect and still untestable as far as I can tell. (or even `block = 'UNKNOWN'`, but that would _still_ be untestable and then, what's the caller supposed to do with it?)

JoshuaKGoldberg: Yeah it's not ideal. I'd say using a `!` on the variable name is probably fine, or some similar assertion.

JoshuaKGoldberg (approved):
Kuzco from The Emperor's New Groove dancing

packages/eslint-plugin/src/rules/return-await.ts
bradzacher: We have the nullthrows util to add clear runtime assertions to avoid `!`

bradzacher (approved):

PR #9038, opened 2024-05-02
chore: refactor tsconfigs to use project references

JamesHenry:


tsconfig.base.json
```diff
@@ -1,21 +1,54 @@
...
+    "incremental": true,
+    "isolatedModules": true,
+    "lib": ["es2022"],
+    "module": "NodeNext",
+    "noEmitOnError": true,
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitOverride": true,
     "noImplicitReturns": true,
-    "paths": {},
+    "noUnusedLocals": true,
```
bradzacher, line 21: can we turn this off?
`no-unused-vars` does this without causing TS errors (so we can remove the `@ts-expect-error`s)
it also means the build won't error out due to a measly unused variable.

packages/visitor-keys/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export { getKeys } from './get-keys';
-export { visitorKeys, VisitorKeys } from './visitor-keys';
+export { visitorKeys } from './visitor-keys';
+export type { VisitorKeys } from './visitor-keys';
```
bradzacher, line 3: we should probably turn on our `consistent-type-exports` rule

tsconfig.base.json
```diff
@@ -1,21 +1,54 @@
...
-    "module": "Node16",
-    "moduleResolution": "Node16",
+    "experimentalDecorators": false,
+    "forceConsistentCasingInFileNames": true,
+    "incremental": true,
+    "isolatedModules": true,
+    "lib": ["es2022"],
+    "module": "NodeNext",
+    "noEmitOnError": true,
+    "noFallthroughCasesInSwitch": true,
```
bradzacher, line 18: ditto here - there's an eslint rule for this
i'd rather not use TS for linting if we can avoid it

tsconfig.base.json
JamesHenry: Cool yeah I just took a relatively strong starting point inspired by other projects, we can change any of it to whatever's best. It isn't something that is generated or controlled by Nx in any way, it's completely up to us

packages/visitor-keys/src/index.ts
JamesHenry: Yeah all of these `type` changes allowed compilation `isolatedModules` to work

tsconfig.base.json
JamesHenry: Yeah same comment

jakebailey:
Tested this out locally and it does seem to work; `tsc -b` at the root does the right thing (awesome), and opening up `typescript-estree/src/index.ts` and doing a find-all-refs on `getCanonicalFileName` finds the other projects, including the use in `semanticInfo-singleRun.test.ts` which it previously couldn't find until that was opened.

packages/ast-spec/tsconfig.json
```diff
@@ -1,10 +1,15 @@
...
-  "references": [{ "path": "../typescript-estree/tsconfig.build.json" }]
+  "files": [],
+  "references": [
+    {
+      "path": "./tsconfig.build.json"
+    },
+    {
+      "path": "./tsconfig.spec.json"
+    }
+  ]
```
jakebailey, line 14: This is interesting; I assume these are the main code and the tests, in which case I would have actually expected you to have `tsconfig.json` here be the main code, then have a `tsconfig.test.json` (or `spec`, that's bikeshedding) separate and then referenced from the repo root

packages/ast-spec/tsconfig.json
```diff
@@ -1,10 +1,15 @@
 {
-  "extends": "./tsconfig.build.json",
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "composite": false,
-    "rootDir": "."
+    "outDir": "../../dist/out-tsc/ast-spec"
```
jakebailey, line 4: Since this is a "solution" file, there's actually no reason to extend anything here; these options are all noops.

But, I'm not sure that this layout is quite right, depending on other tooling which may assume that the tsconfig contains the options for the code. Maybe this extension is why you dont' have problems?

packages/ast-spec/tsconfig.build.json
```diff
@@ -1,10 +1,9 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "composite": true,
-    "jsx": "preserve",
-    "outDir": "./dist",
-    "rootDir": "./src",
+    "rootDir": "src",
+    "outDir": "dist",
+    "emitDeclarationOnly": false,
```
jakebailey, line 6: Oh, hm, you're using `emitDeclarationOnly`, but only sometimes. It seems simpler to only set noEmit for test projects, then leave all emit on in the base config, then only have `tsconfig.json` and `tsconfig.spec.json` per-package if you want to split apart tests.

tsconfig.json
```diff
@@ -1,18 +1,58 @@
...
-    "jest.preset.js"
+  "compileOnSave": false,
+  "files": [],
+  "references": [
+    {
+      "path": "./packages/ast-spec"
+    },
+    {
+      "path": "./packages/eslint-plugin"
+    },
```
jakebailey, line 11: One gotcha here may be that TS won't do rebuilds in some situations if _all_ projects are not listed here. Right now, you're using this top-level "solution" that points to other "solutions". Related is: https://github.com/microsoft/TypeScript/issues/30608 / https://github.com/microsoft/TypeScript/pull/55339

tsconfig.base.json
```diff
@@ -1,21 +1,34 @@
...
     "declaration": true,
     "declarationMap": true,
+    "emitDeclarationOnly": true,
+    "emitDecoratorMetadata": false,
     "esModuleInterop": true,
-    "lib": ["ES2021"],
-    "module": "Node16",
-    "moduleResolution": "Node16",
+    "experimentalDecorators": false,
+    "forceConsistentCasingInFileNames": true,
```
jakebailey, line 14: A lot of these settings are already their defaults and could be removed for clarity, e.g. allowJs, decorators, pretty, and so on.

jakebailey (commented):
Re-looking at the config layout, I do think you can get away with fewer configs, actually, via some restructuring. But it technically already functions now, of course.

packages/ast-spec/tsconfig.json
JamesHenry: The idea behind this approach/config structure was as follows:

- A `tsconfig.json` for a project will power its type-checking. This is is not expected to produce any .js etc artifacts. It is logically expected to type-check all files, both production and test files.
- An optional `tsconfig.lib.json` (customizable name, in this workspace the convention is `tsconfig.build.json`) would be used to control a `tsc` powered build, in which .js etc artifacts _are_ produced. It should naturally not include any test files as part of its build process.
  - The presence of this file is also how Nx, via the new plugin, can infer that a build "target" should exist for this project (meaning you can run `nx build ast-spec` and it just works, including distributable caching and other core Nx features, without any specific Nx config at the project level. 
- An optional `tsconfig.spec.json` seemed to make sense as a corollary of the `tsconfig.lib.json`/`tsconfig.build.json`, but note that its presence is not used to infer any targets. It is only there to pick up all the files that the build config would exclude and possible be passed to a test runner like jest with ts-jest as config.

Does that all make sense? Is there a different/more optimal way to achieve the above objectives around having typecheck and build be separate actions/targets?

packages/ast-spec/tsconfig.build.json
JamesHenry: Let me know if my comment here https://github.com/typescript-eslint/typescript-eslint/pull/9038#discussion_r1608042136 alters your view on this please. Having type-checking not produce .js etc artifacts would seem to be desirable, and therefore we would need a way to distinguish between type-checking vs building

tsconfig.base.json
JamesHenry: Yeah I think that's probably fair... I was shooting for maximum clarity, but probably the fewer things for the average user to think about the better. Y'all are striving for the most logical/useful default values and on the whole users don't need to care about many of the possible options

packages/ast-spec/tsconfig.json
jakebailey: That does make some sense, I was jus tlooking for ways to simplify things. But this is not dissimilar to how vite does things where it makes an extra tsconfig that's referenced just for config files.

JamesHenry:
Closing in favour o #10361

PR #9042, opened 2024-05-03
fix: do not pass tsconfig canonical file name to typescript API to get program details for config file

sheetalkamat:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/1192

## Overview

Without this it generates root files with incorrect casing and results in errors with casing with forceConsistentCasingInFileNames which is true by default in newer typescript versions

This came up while i was working on PR https://github.com/microsoft/TypeScript/pull/58398

cc @jakebailey i was able to turn the tsconfig cache into Map and find all places where it needed to be string and was passing canonical file path

### DIFF
--- a/packages/typescript-estree/src/create-program/createDefaultProgram.ts
+++ b/packages/typescript-estree/src/create-program/createDefaultProgram.ts
@@ -23,11 +23,11 @@ function createDefaultProgram(
     parseSettings.filePath || 'unnamed file',
   );
 
-  if (parseSettings.projects.length !== 1) {
+  if (parseSettings.projects.size !== 1) {
     return undefined;
   }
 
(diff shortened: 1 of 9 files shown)


JamesHenry (approved):
Thanks so much @sheetalkamat! So even without the change on the TS side this should massively help the linked OOM issue right?

sheetalkamat:
That’s right .. this should fix OOM with force consistent file name errors without typescript fix .. not sure if there are more places in which canonical path is passed in to API though ..

JamesHenry:
I'm going to go ahead and merge this, thanks again @sheetalkamat

JoshuaKGoldberg:
Woohoo, thanks for this !!

PR #9043, opened 2024-05-04
fix(eslint-plugin): [no-floating-promises] check top-level type assertions (and more)

### LINKED ISSUE #8692: Bug: [no-floating-promises] support TsAsExpression

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.2&fileType=.tsx&code=C4TwDgpgBAYg9nKBeKAFATnAtgSwM4QA8AdgK5YBGE6AfFAGRQDeUAFhCAPwBcUew6HMQDmUAL4BuALAAoADYRgUCgEMAXmuRQVeEMQDGUABQBKZHQBM0mao2nteWAglA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDN4OBDfMwDmtYtA4BbSshTooiaBOiRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
type Foo = Promise<number> & { hey?: string };
let bazz = async () => 2;
bazz() as Foo;
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-floating-promises": ["error"],
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expected that the 3rd line should report the absence of handling the rejection of the promise.

### Actual Result

There was no error on 3rd line

### Additional Info

_No response_

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8692

## Overview

Changes implementation from an unhandled promise syntax allowlist to a syntax denylist.


packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -140,9 +140,8 @@ async function test() {
 }
     `,
     `
+declare const promiseValue: Promise<number>;
 async function test() {
-  declare const promiseValue: Promise<number>;
```
kirkwaiblinger, line 144: These changes are just because declare is not syntactically allowed in that position. Just some hygiene, not relevant to the changes.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -516,6 +513,29 @@ declare const myTag: (strings: TemplateStringsArray) => string;
 myTag\`abc\`;
       `,
     },
+    {
+      code: `
+declare let x: any;
+declare const promiseArray: Array<Promise<unknown>>;
+x = promiseArray;
```
kirkwaiblinger, line 520: surprisingly, this was not tested for.

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -257,6 +261,16 @@ export default createRule<Options, MessageId>({
         return { isUnhandled: true, promiseArray: true };
       }
 
+      // await expression addresses promises, but not promise arrays.
+      if (node.type === AST_NODE_TYPES.AwaitExpression) {
+        // you would think this wouldn't be strictly necessary, since we're
+        // anyway checking the type of the expression, but, unfortunately TS
+        // reports the result of `await (promise as Promise<number> & number)`
```
arka1002, line 268: Typescript gives a warning
> 'await' has no effect on the type of this expression.

Lets not include example of those cases for which the ts compiler gives us a warning/error (?)

[Minimum repro for the warning](https://www.typescriptlang.org/play/?target=9#code/IYZwngdgxgBAZgV2gFwJYHsIwELAF54AUAlDAN4CwAUDLTADYCmy866MAvDMAO7CotCABQBO6ALaoQjAHQjGIdPQBujQgFZSoGKIlTGAHggJxAI0YiAfDABkMY2YsxiAbmp0YAek90AegH5qAF8gA) , hover over `await` keyword in line 2.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -1855,5 +1898,47 @@ cursed();
...
+declare const x: Generator<number, Promise<number>, void>;
+function* generator(): Generator<number, void, void> {
+  yield* x;
+}
+      `,
+      errors: [{ messageId: 'floatingVoid' }],
+    },
+    {
+      code: `
+3 as Promise<number>;
```
arka1002, line 1921: I'm not a community maintainer, but I'm very much against these tests.
Even, after checking off all the options in the type-checking section, typescript is still giving error on this statement.

[Playground link](https://www.typescriptlang.org/play/?strict=false&noImplicitAny=false&strictNullChecks=false&strictFunctionTypes=false&strictPropertyInitialization=false&strictBindCallApply=false&noImplicitThis=false&noImplicitReturns=false&alwaysStrict=false#code/MwAghgziAKBOD2BbAlhApgHgHYFdECM1YA+AbiA)

Either these tests be removed or should be replaced with valid ts code.

arka1002 (changes requested):

packages/eslint-plugin/src/rules/no-floating-promises.ts
JoshuaKGoldberg: > Lets not include example of those cases for which the ts compiler gives us a warning/error (?)

No, we'll still have to handle them correctly. Especially when it's just a warning like this one.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
JoshuaKGoldberg: Relevant: #8298

+1 that if we can get away with no type error, it'd be better to.

```ts
const value = {};

value as Promise<number>;
```

JoshuaKGoldberg (changes requested):
Syntax looks great to me! ✨ 🧹 

Just waiting on the test discussion.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
kirkwaiblinger: Makes sense. Also, very clever with suggesting the castable `{}` 👍

JoshuaKGoldberg: The empty object type has been on my mind a lot lately... 😂

JoshuaKGoldberg (dismissed):
Amusingly crude animation of a photo of a cat on a PB&J sandwich, floating up and down above the cosmos

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -327,10 +329,8 @@ export default createRule<Options, MessageId>({
         return isUnhandledPromise(checker, node.right);
       }
 
-      // We conservatively return false for all other types of expressions because
-      // we don't want to accidentally fail if the promise is handled internally but
-      // we just can't tell.
-      return { isUnhandled: false };
+      // Anything else is unhandled.
+      return { isUnhandled: true };
```
bradzacher, line 333: In your summary you wrote

> Changes implementation from an unhandled promise syntax allowlist to a syntax denylist.

But I don't understand what we gain from that switch

This seems... dangerous - switching to default assuming it's unhandled is probably going to lead to false positives at scale.

kirkwaiblinger: > I don't understand what we gain from that switch
 
implementation simplicity and future proofing 🤷  It seems more straightforward, and more correct (IMO), to opt _out_ of flagging behaviors indicated by the docs than it does to opt _in_ to the ones we do want to flag...

Nodes that are handled
- `await` expression
- assignment expression (x = y)

Nodes that require special handling to determine whether they're handled or unhandled (and therefore need to be specified individually no matter what)
- CallExpression
- LogicalExpression
- SequenceExpression
- ConditionalExpression
- `void` expression
- optional chaining expression
- iifes

Nodes that are always unhandled (if they result in a promise)
- NewExpression
- Identifier
- MemberExpression
- TaggedTemplateExpression (previously forgotten, recently fixed in https://github.com/typescript-eslint/typescript-eslint/pull/8758)
- TSAsExpression (previously forgotten, the point of this PR)
- TSTypeAssertion (previously forgotten, the point of this PR)
- TSNonNullExpression (previously forgotten, basically equivalent to previous two)
- ... probably more - this feels like the appropriate default behavior for a rule that doesn't fall into one of the other two buckets

> This seems... dangerous - switching to default assuming it's unhandled is probably going to lead to false positives at scale.

To be sure, I gave this some thought before making this change, and I just couldn't think of any scenarios where this would really be problematic... Every case I thought of seeed to be beneficial. Just to be sure we're on the same page, that code is only reached _if the expression results in a promise_. 

That does mean that this code will start flagging now
```ts
({
  then(resolve: Function, reject: Function) {
    return this;
  }
});
```
and I think that makes sense, given 
```ts
const unhandled = {
  then(resolve: Function, reject: Function) {
    return this;
  }
};
unhandled;
```
already flags (of course, this touches upon https://github.com/typescript-eslint/typescript-eslint/issues/8433). But it's not like we're suddenly going to flag every object literal; just thenable ones.

I think the `yield` cases added to the tests make sense to flag as well....

Note that we dogfood `no-floating-promises` in this repo and this change didn't result in _any_ new reports. 

My conclusion - false positives are a very valid concern to explore, but for now I do think that this change is safe/beneficial. Would definitely be open to being proven wrong, though, especially if a good counterexample comes to mind! 🙂

bradzacher (approved):

JoshuaKGoldberg (approved):
🚀 yay!

PR #9051, opened 2024-05-06
fix(typescript-estree): pass extraFileExtensions to projectService

### LINKED ISSUE #8899: Bug: parserOptions.projectService does not respect extraFileExtensions

### Issue Description

When linting a project containing e.g. svelte files, you need to provide `extraFileExtensions: ['.svelte']` to make the parser use the correct tsconfig for the specific file. This works well with `project: true`, since that path ends up in `createWatchProgram` where we provide the `extraFileExtensions` option to ts.

However, if we switch `EXPERIMENTAL_useProjectService` on, we end up in a code path where the extraFileExtensions are not passed to ts. In my reproduction case this is surfaced by the fact that ts will then use the default compiler options, which transpiles the code back to ES5, instead of using the correct tsconfig.

While this reproduction example is using Svelte, the same issue should apply to other cases where `extraFileExtensions` would be used (such as when using Vite).

https://github.com/alfredringstad/typescript-eslint-svelte-use-project-service-bug-reproduction

### Repro Steps

1) clone the repo
2) `npm i`
3) `npm run lint`

Switch `EXPERIMENTAL_useProjectService` to `false` in `eslint.config.js` and run `npm run lint` to see how it works well without the project service.

### Versions

| `@typescript-eslint/parser`            | `7.6.0` |
| `@typescript-eslint/typescript-estree`            | `7.6.0` |
| `TypeScript`                           | `5.4.5` |
| `ESLint`                               | `8.57.0` |

alfredringstad:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8899 

## Overview

Uses `setHostConfiguration` to pass `extraFileExtensions` to project service, when using `EXPERIMENTAL_useProjectService`. This fixes an issue where project service would not detect the appropriate tsconfig file when using non-default file extensions.

The tests are checking that the configFilePath is set in the compiler options. If it's not, it's an indication that the default inferred project is used. I duplicated the test cases used for the non-experimental projects option.

### DIFF
--- a/packages/typescript-estree/src/parser-options.ts
+++ b/packages/typescript-estree/src/parser-options.ts
@@ -124,6 +124,11 @@ export interface ProjectServiceOptions {
    * @default 8
    */
   maximumDefaultProjectFileMatchCount_THIS_WILL_SLOW_DOWN_LINTING?: number;
+
+  /**
+   * Additional file extensions which should be considered in the TypeScript Program compilation.
+   */
+  extraFileExtensions?: string[];
 }
 
 interface ParseAndGenerateServicesOptions extends ParseOptions {
--- a/packages/typescript-estree/src/parser-options.ts
+++ b/packages/typescript-estree/src/parser-options.ts
@@ -124,11 +124,6 @@ export interface ProjectServiceOptions {
    * @default 8
    */
   maximumDefaultProjectFileMatchCount_THIS_WILL_SLOW_DOWN_LINTING?: number;
-
-  /**
-   * Additional file extensions which should be considered in the TypeScript Program compilation.
-   */
-  extraFileExtensions?: string[];
 }
 
 interface ParseAndGenerateServicesOptions extends ParseOptions {
--- a/packages/typescript-estree/tests/lib/parse.test.ts
+++ b/packages/typescript-estree/tests/lib/parse.test.ts
@@ -352,7 +352,6 @@ describe('parseAndGenerateServices', () => {
         range: true,
         loc: true,
         tsconfigRootDir: PROJECT_DIR,
-        project: './tsconfig.json',
       };
       const testParse =
         (filePath: string, extraFileExtensions: string[] = ['.vue']) =>
@@ -362,12 +361,13 @@ describe('parseAndGenerateServices', () => {
               ...config,
               extraFileExtensions,
               filePath: join(PROJECT_DIR, filePath),
+              project: './tsconfig.json',
             });
           } catch (error) {
             alignErrorPath(error as Error);
           }
         };
-      const testParseExperimentalProjectService =
+      const testExtraFileExtensions =
         (filePath: string, extraFileExtensions: string[] = ['.vue']) =>
         (): void => {
           const result = parser.parseAndGenerateServices(code, {
@@ -499,14 +499,12 @@ describe('parseAndGenerateServices', () => {
       describe('"parserOptions.extraFileExtensions" is non-empty and EXPERIMENTAL_useProjectService is true', () => {
         describe('the extension matches', () => {
           it('the file is included', () => {
-            expect(
-              testParseExperimentalProjectService('other/included.vue'),
-            ).not.toThrow();
+            expect(testExtraFileExtensions('other/included.vue')).not.toThrow();
           });
 
           it("the file isn't included", () => {
             expect(
-              testParseExperimentalProjectService('other/notIncluded.vue'),
+              testExtraFileExtensions('other/notIncluded.vue'),
             ).toThrowErrorMatchingInlineSnapshot(
               `"No config file found, using inferred project"`,
             );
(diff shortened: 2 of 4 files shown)


packages/typescript-estree/src/parser-options.ts
```diff
@@ -123,6 +123,11 @@ export interface ProjectServiceOptions {
    * @default 8
    */
   maximumDefaultProjectFileMatchCount_THIS_WILL_SLOW_DOWN_LINTING?: number;
+
+  /**
+   * Additional file extensions which should be considered in the TypeScript Program compilation.
+   */
+  extraFileExtensions?: string[];
```
JoshuaKGoldberg, line 130: [Docs] Nit: would you mind putting these options in alphabetical order please? It helps keep the page well structured & easy to navigate.

packages/typescript-estree/src/parser-options.ts
JoshuaKGoldberg, line 1: [Docs] `docs/packages/Parser.mdx` will need to be updated for this too.

packages/typescript-estree/tests/lib/useProgramFromProjectService.test.ts
JoshuaKGoldberg, line 1: [Testing] What do you think about adding a more direct unit test to this file too? IMO it's nice to have tests that directly check the API boundaries of where we talk to TypeScript too.

packages/typescript-estree/tests/lib/parse.test.ts
```diff
@@ -367,6 +367,22 @@ describe('parseAndGenerateServices', () => {
             alignErrorPath(error as Error);
           }
         };
+      const testParseExperimentalProjectService =
+        (filePath: string, extraFileExtensions: string[] = ['.vue']) =>
```
JoshuaKGoldberg, line 371: [Testing] A couple of thoughts:

```suggestion
      const testExtraFileExtensions =
        (filePath: string, extraFileExtensions: string[]) =>
```

* This test is basically only used for `extraFileExtensions`, so let's name it like that
* IMO it's nice to be a bit more explicit on what settings are actually being tested

JoshuaKGoldberg (changes requested):
Source code changes look great to me! And checking that the config file path is found as a form of testing is very snazzy. Nicely done. 😄

Just requesting changes on a few small things that I'm open to being pushed back on - what do you think?

packages/typescript-estree/src/parser-options.ts
alfredringstad: I had another look at this, and realised this must've been a left-over from a previous iteration, this change is actually not needed, since we rely on the extraFileExtensions of the parent object (same as the non-project-service setup).

packages/typescript-estree/src/parser-options.ts
alfredringstad: See comment above, the change was reverted.

packages/typescript-estree/tests/lib/useProgramFromProjectService.test.ts
alfredringstad: Good idea! I added two tests to be extra specific, to verify that we only call setHostConfiguration when extraFileExtensions has been provided.

packages/typescript-estree/tests/lib/parse.test.ts
alfredringstad: Followed your suggestion. Just want to highlight that this is specifically testing extraFileExtensions when projectService is true, and that there are other tests to cover extraFileExtensions when projectService is not used.

alfredringstad:
Thanks for the feedback @JoshuaKGoldberg! I've implemented your suggestions. Seems like CI failure is due to issues in the main v8 branch, the typescript-estree tests seem to work fine.

packages/typescript-estree/tests/lib/useProgramFromProjectService.test.ts
JoshuaKGoldberg, line 1: [Praise] 💯

packages/typescript-estree/tests/lib/parse.test.ts
```diff
@@ -479,6 +494,48 @@ describe('parseAndGenerateServices', () => {
           `);
         });
       });
+
+      describe('"parserOptions.extraFileExtensions" is non-empty and EXPERIMENTAL_useProjectService is true', () => {
```
JoshuaKGoldberg, line 498: We _just_ applied this rename recently ✨ 

```suggestion
      describe('"parserOptions.extraFileExtensions" is non-empty and projectService is true', () => {
```

packages/typescript-estree/tests/lib/parse.test.ts
```diff
@@ -362,11 +361,27 @@ describe('parseAndGenerateServices', () => {
...
               extraFileExtensions,
               filePath: join(PROJECT_DIR, filePath),
+              project: './tsconfig.json',
             });
           } catch (error) {
             alignErrorPath(error as Error);
           }
         };
+      const testExtraFileExtensions =
+        (filePath: string, extraFileExtensions: string[] = ['.vue']) =>
```
JoshuaKGoldberg, line 371: [Testing] I think this got missed in the refactors?

```suggestion
        (filePath: string, extraFileExtensions: string[]) =>
```

I can just apply it myself, about to merge.

JoshuaKGoldberg (approved):
This is great, thanks! :shipit: 

Normally I'd wait for a bit to let someone else review but I would like to unblock folks who need the extra file extensions from using the project.

PR #9059, opened 2024-05-07
fix(eslint-plugin): [use-unknown-in-catch-callback-variable] flag second argument of `.then`

### LINKED ISSUE #9056: Bug: [use-unknown-in-catch-callback-variable] Does not flag second argument of `.then`

### Playground Link

https://typescript-eslint.io/play/#ts=5.3.3&fileType=.tsx&code=AoJw9gtglgzgpgOhHGYA2A3OAKAlAgFwAs4A7bPAAgF4A%2BSgb0oF8AaS7OEcEXG%2Brjz4B6YZUFgQlWJQIBPAA5wAJpQCGMdaTlA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WZRW2JgayYD2Adya1mtMgEN8ZABYTJ8eACNJZXrQBukqpOVJ0URNGgDokcGAC%2BIK0A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
Promise.resolve().then(() => { }, (error) => { })
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/use-unknown-in-catch-callback-variable": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expected an error as `error` is not explicitly typed as `unknown` and instead falls back to `any`.

### Actual Result

No error.

### Additional Info

_No response_

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9056


JoshuaKGoldberg:
👋 @abrahamguo did you mean to un-draft this?

abrahamguo:
Not it's not finished yet, I'll try to finish it up in the next few days!

abrahamguo:
Work done:

- [x] Update docs to mention `.then()` in addition to `.catch()` throughout.
- [x] Update rule report wording to be more specific, and rule suggestion wording to be more generic
- [x] Check simple form of `.then()`
- [x] Check `.then()` spreading in a 2-element tuple

kirkwaiblinger (approved):

kirkwaiblinger (changes requested):
Several comments, and questions. Main thing to resolve is whether to bother with the spread arg counting or not 🤷

packages/eslint-plugin/docs/rules/use-unknown-in-catch-callback-variable.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-description: 'Enforce typing arguments in `.catch()` callbacks as `unknown`.'
+description: 'Enforce typing arguments in `.then()` and `.catch()` rejection callbacks as `unknown`.'
```
Josh-Cena, line 2:
```suggestion
description: 'Enforce typing arguments in promise rejection callbacks as `unknown`.'
```

(Same for other places)

packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
```diff
@@ -290,40 +309,79 @@ export default createRule<[], MessageIds>({
...
+        if (!promiseMethodInfo) {
           return;
         }
+        const { argumentIndexToCheck, messageData } = promiseMethodInfo;
+
+        const report = (
+          descriptor: TSESLint.ReportDescriptor<MessageIds>,
+        ): void => context.report({ ...descriptor, data: messageData });
+
+        const firstArgument = args.at(0);
```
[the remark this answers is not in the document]
abrahamguo: I agree that it's probably not worth it, and provides very little user value. I was indeed just trying to match the standard you originally set with this rule, so I'm happy to go ahead and remove all spread-arg checking.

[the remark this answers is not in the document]
kirkwaiblinger: Yeah, setting that precedent was perhaps not one of my better ideas 😅

packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
```diff
@@ -69,22 +69,38 @@ export default createRule<[], MessageIds>({
...
+        }
+      | false {
+      if (node.type !== AST_NODE_TYPES.MemberExpression) {
+        return false;
+      }
+      const methodsToCheck = [
+        { method: 'catch', append: '' },
+        { method: 'then', append: ' rejection' },
+      ] as const;
+      const argumentIndexToCheck = methodsToCheck.findIndex(({ method }) =>
```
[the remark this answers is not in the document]
abrahamguo: I agree that the `-1` and the type assertion were a little bit clunky!
I preferred to have all of the options for `then` and `catch` consolidated rather than divided into an array and a `switch` like you had it. Based on this and a couple other of your comments (narrowing and return type), I felt that it might make most sense to inline the function. However, let me know if you don't like it and I'm happy to do something more like what you're recommending here.

packages/eslint-plugin/docs/rules/use-unknown-in-catch-callback-variable.mdx
abrahamguo: love it!

packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
```diff
@@ -289,44 +234,46 @@ export default createRule<[], MessageIds>({
     }
 
     return {
-      CallExpression(node): void {
-        if (node.arguments.length === 0 || !isPromiseCatchAccess(node.callee)) {
+      CallExpression({ arguments: args, callee }): void {
```
kirkwaiblinger, line 237: I was playing around with this locally and I sketched out something like the following: 
<details>
<summary>Expand to see</summary> 

```ts
CallExpression(node): void {
  const { arguments: args, callee } = node;

  if (callee.type !== AST_NODE_TYPES.MemberExpression) {
    return;
  }

  const memberAccessKey = getStaticMemberAccessKey(callee);
  if (memberAccessKey == null) {
    return;
  }

  let promiseMethodInfo: {
    method: string;
    append: string;
    argToCheck: number;
  };

  if (memberAccessKey.value === 'catch') {
    promiseMethodInfo = { method: 'catch', append: '', argToCheck: 0 };
  } else if (memberAccessKey.value === 'then') {
    promiseMethodInfo = {
      method: 'then',
      append: ' rejection',
      argToCheck: 1,
    };
  } else {
    return;
  }

  // need to be enough args to check.
  if (args.length < promiseMethodInfo.argToCheck + 1) {
    return;
  }

  // Argument to check, and all arguments before it, are an "ordinary" argument (i.e. not a spread argument)
  // promise.catch(f), promise.catch(() => {}), promise.catch(<expression>, <<other-args>>)
  const argsToCheck = args.slice(0, promiseMethodInfo.argToCheck + 1);
  if (
    argsToCheck.some(arg => arg.type === AST_NODE_TYPES.SpreadElement)
  ) {
    return;
  }

  if (
    !tsutils.isThenableType(
      checker,
      services.esTreeNodeToTSNodeMap.get(callee),
      checker.getTypeAtLocation(
        services.esTreeNodeToTSNodeMap.get(callee.object),
      ),
    )
  ) {
    return;
  }

  const argToCheck = args[promiseMethodInfo.argToCheck] as Exclude<
    (typeof argsToCheck)[number],
    TSESTree.SpreadElement
  >;

  if (shouldFlagArgument(argToCheck)) {
    // We are now guaranteed to report, but we have a bit of work to do
    // to determine exactly where, and whether we can fix it.
    const overrides = refineReportForNormalArgumentIfPossible(argToCheck);
    context.report({
      node,
      messageId: 'useUnknown',
      data: {
        method: promiseMethodInfo.method,
        append: promiseMethodInfo.append,
      },
      ...overrides,
    });
  }
},

```
(with helper changed to)
```ts
/**
 * `x.memberName` => 'memberKey'
 *
 * `const mk = 'memberKey'; x[mk]` => 'memberKey'
 *
 * `const mk = 1234; x[mk]` => 1234
 */
function getStaticMemberAccessKey(
  memberExpression:
    | TSESTree.MemberExpressionComputedName
    | TSESTree.MemberExpressionNonComputedName,
  scope?: Scope.Scope | undefined,
): { value: unknown } | null {
  if (!memberExpression.computed) {
    // x.memberName case.
    return { value: memberExpression.property.name };
  }

  // x['memberName'] cases.
  return getStaticValue(memberExpression.property, scope);
}
```
</details>

By no means do I mean to say that this code is correct or fully polished, but I wanted to point out a few improvements that I think we should make:

1. Let's do as many syntactic checks as possible/reasonable before invoking the type checker (performance reasons). _The current changeset invokes the type checker on every method call before checking whether it's a `then` or `catch` access - we don't need to do that._
2. Let's really try to avoid doing overly complicated `.find()` stuff for what is really just two cases (`'then'` and `'catch'`). _Note - my code is sort of guilty of this too with the `.slice()`. Feel free to find a better way_.
3. No need to keep looking after having checked the rejection handler's position! 😁 
4. Eliminate the need to deal with the confusing `false | undefined | [number, { method: string, append: string }]` type on `promiserMethodInfo`.

Feel free to make this your own, and definitely I'm not implying you should just copy the code I wrote, but I'd like to clean up the impl a bit on these points.

kirkwaiblinger (changes requested):
Logic is correct, and I agree with all the test cases. Left a comment on some style requests, but otherwise in great shape!

packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
```diff
@@ -235,69 +237,65 @@ export default createRule<[], MessageIds>({
...
-          ].find(([, { method }]) =>
-            isStaticMemberAccessOfValue(callee, method),
-          );
+        if (callee.type !== AST_NODE_TYPES.MemberExpression) {
+          return;
+        }
+
+        const staticMemberAccessKey = getStaticMemberAccessKey(callee);
+        if (!staticMemberAccessKey) {
+          return;
```
kirkwaiblinger, line 246: Let's get a test case to hit this branch... guessing this will work?
```ts
x[Math.random()];
```
Or anything not statically analyzable 🤷

packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
```diff
@@ -235,69 +237,65 @@ export default createRule<[], MessageIds>({
...
+        }
+
+        const promiseMethodInfo = [
```
kirkwaiblinger, line 249: make a type for this array - the trouble with inferring it is that one can omit `method` or `append`, or add extraneous properties, and the type checker won't complain. 

maybe like so?
```ts
type PromiseMethodInfo = { reportMessageData: { method: string, append: string }, argIndexToCheck: string }
```

kirkwaiblinger (changes requested):

packages/eslint-plugin/tests/rules/use-unknown-in-catch-callback-variable.test.ts
```diff
@@ -104,6 +112,57 @@ ruleTester.run('use-unknown-in-catch-callback-variable', rule, {
...
+declare const x: any;
+Promise.resolve().catch(...x);
+    `,
+    `
+declare const thenArgs: [() => {}, (err: any) => {}];
+Promise.resolve().then(...thenArgs);
+    `,
+    `
+      let method = 'catch';
+      Promise.resolve()[method]((error: Error) => {});
```
abrahamguo, line 164: This test covers the case when `staticMemberAccessKey` is `null`, getting us to 100% test coverage 👍 

However, while writing this test, I discovered that this rule, both in this branch and on `main`, does not report on
```
const x = 'catch';
Promise.resolve()[x]((err: any) => { });
```
even though some of the code comments in this file make it sound like it should.
Since that is unrelated to this change, I'm going to open that up as a separate issue and PR.

packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
```diff
@@ -288,68 +237,75 @@ export default createRule<[], MessageIds>({
...
-        if (node.arguments.length === 0 || !isPromiseCatchAccess(node.callee)) {
+      CallExpression({ arguments: args, callee }): void {
+        if (callee.type !== AST_NODE_TYPES.MemberExpression) {
           return;
         }
 
-        const firstArgument = node.arguments[0];
+        const staticMemberAccessKey = getStaticMemberAccessKey(
+          callee,
+          context.sourceCode.getScope(callee),
```
abrahamguo, line 247: `scope` was always missing, meaning that `const mk = 'catch'; promise[mk]` was not being detected properly ([playground](https://typescript-eslint.io/play/#ts=5.5.2&fileType=.tsx&code=MYewdgzgLgBAHjAvDA5MAhlYALFBuAKAAUAnEAWwEsIBTAOhJohABsA3GgCgEoBtOALqdONEiQBcMdGACe3JAD4YAbxgBfboVIVq9Rs3Zc%2BaTDhRCRYydLmKV6zUA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYkACAXAngA4oDG0AlobgLQrzkB2uA9LMotbAwNYMD2AdwbVG1EgENcJABZjx8eACNxJbtQBu4iuMVJ0URNGh9okAL4gzQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false))

packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
```diff
@@ -288,68 +237,75 @@ export default createRule<[], MessageIds>({
...
-          }
+        const promiseMethodInfo = (
+          [
+            { method: 'catch', append: '', argIndexToCheck: 0 },
+            { method: 'then', append: ' rejection', argIndexToCheck: 1 },
+          ] satisfies {
+            method: string;
+            append: string;
+            argIndexToCheck: number;
+          }[]
```
abrahamguo, line 261: > make a type for this array - the trouble with inferring it is that one can omit method or append, or add extraneous properties, and the type checker won't complain.

`satisfies` is the only way to get the type checker to complain about extraneous properties — any type annotation or type parameter anywhere else and the type checker will still be OK with extraneous properties 👍

packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
abrahamguo: 100% coverage 🎉

kirkwaiblinger (approved):
Thanks for going through this! Glad to see this changeset progress 🙂

JoshuaKGoldberg (approved):
🚀

PR #9074, opened 2024-05-11
fix(eslint-plugin): [strict-boolean-expressions] consider assertion function argument a boolean context

### LINKED ISSUE #9009: Enhancement: [strict-boolean-expressions] Check truthiness assertion functions

https://typescript-eslint.io/rules/strict-boolean-expressions/

### Description

For example, `ok` from `node:assert` has the following shape:

```ts
function ok(value: unknown, message?: string | Error): asserts value;
```

It appears that currently the rule just looks at `value`, sees `unknown`, and moves on. So it's possible to write this:

```ts
const foo: string | undefined = 'bar';
if (foo) ... // Error
ok(foo); // Fine
```

Instead, it would be nice if it also looked at the return type and took `asserts value` into account, treating it as `if (!value) throw`. There's more to `asserts` than just checking truthiness, but handling the simplest case (no `is`) would already go a long way, as that covers all `ok`-like assertion functions out there.

### Fail

```TypeScript
const foo: string | undefined = 'bar';
ok(foo);
```

### Pass

```TypeScript
const foo: string | undefined = 'bar';
ok(foo !== undefined);
```

### Additional Info

_No response_

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9009 

## Overview

Query for call expressions, check if the function being called is an assertion function, and line up the asserted parameter with the arguments that we see.


packages/eslint-plugin/docs/rules/strict-boolean-expressions.mdx
```diff
@@ -55,6 +56,11 @@ let obj = {};
 while (obj) {
   obj = getObj();
 }
+
+// assertion functions without an `is` are boolean contexts.
+declare function myAssert(value: unknown): asserts value;
+let maybeString = Math.random() > 0.5 ? '' : undefined;
+myAssert(maybeString);
```
JoshuaKGoldberg, line 63: [Docs] Nit: everywhere else generally refers to `assert()`, and that's a pretty commonly used name. I think it'd be nice to use it here for consistency.

```suggestion
declare function assert(value: unknown): asserts value;
let maybeString = Math.random() > 0.5 ? '' : undefined;
assert(maybeString);
```

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -232,10 +233,125 @@ export default createRule<Options, MessageId>({
...
+        assertedParameterIndex = assertedParameterIndexForThisSignature;
+      }
+
+      // Didn't find a unique assertion index.
+      if (assertedParameterIndex == null) {
+        return;
+      }
+
+      const assertedParameter = checkableArguments[assertedParameterIndex];
+      traverseNode(assertedParameter, true);
```
JoshuaKGoldberg, line 352: [Refactor] Totally optional, but I had a bit of a hard time figuring out the intended flow of all the added code. There's a lot. Extracting out the _"what is the asserted parameter index?"_ logic into a separate might be helpful in making that more clear. What do you think?

```suggestion
      const assertedParameter = getAssertedParameter(checkableArguments, node);

      if (assertedParameter) {
        traverseNode(assertedParameter, true);
      }
```

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
JoshuaKGoldberg, line 1: [Testing] Some more test cases I think would be good to have:

* Combining `this: ` and function overloads
* Three very different overloads
* `...` params, including in combination with the other two things

JoshuaKGoldberg (changes requested):
I squinted at the logic a bunch and it all seems pretty reasonable. Nice stuff. Requesting changes to add some tests just to be sure though. ✨

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -232,10 +233,125 @@ export default createRule<Options, MessageId>({
...
+        const assertionTarget = returnTypeAnnotation.parameterName;
+        if (assertionTarget.kind !== ts.SyntaxKind.Identifier) {
+          // This can happen when asserting on `this`. Ignore!
+          return;
+        }
+
+        const firstParameter = declaration.parameters.at(0);
+        const nonThisParameters =
+          firstParameter != null &&
+          firstParameter.name.kind === ts.SyntaxKind.Identifier &&
```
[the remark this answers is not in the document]
kirkwaiblinger: lol i actually hate this style, but i will comply 🤣

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kirkwaiblinger: Very much agree!

packages/eslint-plugin/docs/rules/strict-boolean-expressions.mdx
kirkwaiblinger: hmm, this intentional. I was hoping to emphasize that this applies to any user-defined assertion functions, not just special-casing a built-in (or some yucky function name heuristic).

Will change though if you think sticking to `assert()` is better

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
kirkwaiblinger: added some tests

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kirkwaiblinger: Extracted to a local function for now.... We'll probably want to extract this function to a separate file whenever we add this feature to no-unnecessary-condition as well, see https://github.com/typescript-eslint/typescript-eslint/issues/9076

packages/eslint-plugin/docs/rules/strict-boolean-expressions.mdx
JoshuaKGoldberg: I think the fact that it explicitly shows `function assert` (even after a `declare`) is enough to indicate it's any user-declared one. I think anybody who's understanding docs about assertion functions should understand this.

kirkwaiblinger: sounds good 👍 Changed

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -232,10 +233,139 @@ export default createRule<Options, MessageId>({
...
+        );
+
+        let assertedParameterIndexForThisSignature: number | undefined;
+        for (const [index, parameter] of checkableNonThisParameters.entries()) {
+          if (parameter.dotDotDotToken != null) {
+            // Cannot assert a rest parameter, and can't have a rest parameter
+            // before the asserted parameter. It's not only a TS error, it's
+            // not something we can logically make sense of, so give up here.
+            return undefined;
+          }
```
auvred, line 327: [probably dead code] The rest parameter can only be the last parameter, so the asserted parameter *should* be before the rest parameter (otherwise `asserts` must assert the name of the rest parameter, but that's a TS error). I can't think of a single where this condition would be true in the semantically valid TS code.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -232,10 +233,139 @@ export default createRule<Options, MessageId>({
...
+
+          if (parameter.name.kind !== ts.SyntaxKind.Identifier) {
+            // Only identifiers are valid for assertion targets, so skip over
+            // anything like `{ destructuring: parameter }: T`
+            continue;
+          }
```
auvred, line 333: [Testing] What do you think about adding a test case that proves that proves we only work with identifiers? Currently this branch is not covered:

the identifier checking branch is not covered by tests

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -232,10 +233,139 @@ export default createRule<Options, MessageId>({
...
+        }
+
+        if (
+          assertedParameterIndex != null &&
+          assertedParameterIndex !== assertedParameterIndexForThisSignature
+        ) {
+          // The asserted parameter we found for this signature didn't match
+          // previous signatures.
+          return undefined;
+        }
```
auvred, line 356: [Refactor] Nit: the if statement below contains `assertedParameterIndex != null && assertedParameterIndex !== assertedParameterIndexForThisSignature`. This means that `assertedParameterIndexForThisSignature` is non-nullable, since  we require it to be equal to the numeric `assertedParameterIndex`.

The first if statements seems redundant:

```suggestion
        if (
          assertedParameterIndex != null &&
          assertedParameterIndex !== assertedParameterIndexForThisSignature
        ) {
          // The asserted parameter we found for this signature didn't match
          // previous signatures.
          return undefined;
        }
```

auvred (changes requested):
Overall looks great! Just requesting a few touchups.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kirkwaiblinger: Thanks for pointing this out! Added

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kirkwaiblinger: > I can't think of a single where this condition would be true in the semantically valid TS code.

Agreed, that's what I tried to express in the comment. But I still think it's worth ensuring the rule doesn't crash. Added some test cases that hit this.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kirkwaiblinger: Hmmmmm. You're right that the runtime behavior is the same with or without the branch. However, it's a conceptually different reason for terminating, and, therefore, preserving it ensures code coverage for the different possibilities.

I think I have a slight preference for it to stay the way it is, but I will change if you prefer for the checks to be combined. 

*** 

Another example of something related is here:
https://github.com/typescript-eslint/typescript-eslint/blob/6d680202c7495359561f31273bd4e33e65c17324/packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts#L106-L122
While the branches are identical, the are identical for very different reasons (in the first case, we want the `T` in `Array<T>`, which coincidentally is the 0th type argument... note - it would probably be better written as `firstParamType.getNumberIndexType()`), whereas, in the case of the tuple type, we want the type specifically at the 0th position, i.e. `[string, number]`.

auvred (approved):
This looks excellent! I like the thoroughness of the added tests

JoshuaKGoldberg (approved):
🚢

alecmev:
Thank you! 👍

PR #9102, opened 2024-05-15
feat(eslint-plugin): replace ban-types with no-restricted-types, no-unsafe-function-type, no-wrapper-object-types

### LINKED ISSUE #8978: Enhancement: [ban-types] Split into default-less no-restricted-types and more targeted type ban rule(s)
## Overview

Splitting out of #8700: `ban-types` has long been a kind of mirror to the core ESLint rule [`no-restricted-syntax`](https://eslint.org/docs/latest/rules/no-restricted-syntax). But, `ban-types` _also_ comes with some default settings. That means the rule really has two areas of responsibility:

* Banning confusing built-in types
* Giving users the ability to ban types of their choosing

That dual-responsibility has led to the rule being more convoluted to configure than average. It's the only rule of ours right now that has to include an [`extendDefaults`](https://typescript-eslint.io/rules/ban-types/#extenddefaults) option.

https://github.com/typescript-eslint/typescript-eslint/blob/db4b5e07f5d7ab67e905517a52ab3974077038d3/packages/website/plugins/generated-rule-docs/insertions/insertNewRuleReferences.ts#L22-L28

Once #8977 is in, the defaults for `ban-types` will exclusively target the uppercase aliases of primitive types (`Boolean`, `Number`, ...) and the general `Function` and `Object` types. I think the v8 breaking changes are a good opportunity to further simplify the rule. Proposal: let's...

* Remove the remaining default options for `ban-types`, rename it to `no-restricted-types` (to mirror `no-restricted-syntax`), and remove it from `recommended`
* Add 1-3 new rule(s) in `recommended` that ban those previous default types
    * Starting proposal: a single `no-uppercase-alias-types` rule?

This would be a breaking change in that it would change the defaults for `ban-types`. Note that the stuff linted against by default in `ban-types` would still be linted against - just by different rule(s).

JoshuaKGoldberg:
BREAKING CHANGE:
Replaces rules in the recommended configs.

## PR Checklist

- [x] Addresses an existing open issue: fixes #8978

## Overview

Finally deprecates the venerable `ban-types` rule and adds three new rules:

- 🆕 `no-restricted-types`: allowing banning a configurable list of type names
- 🆕 `no-unsafe-function-type`: banning the built-in `Function`
- 🆕 `no-wrapper-object-types`: banning `Object` and built-in class wrappers such as `Number`

💖


packages/eslint-plugin/docs/rules/no-uppercase-wrapper-types.mdx
JoshuaKGoldberg, line 1: I don't like this name 😦 but couldn't think of one I really do like... `no-primitive-class-wrappers`? `no-class-wrapper-types`? So wordy, but I also like ending with `-types`... Someone please help 🙏

kirkwaiblinger: TL;DR `ban-bad-built-in-class-types` or even just `ban-bad-class-types`

*** 

wordy? you must hate `use-unknown-in-catch-callback-variable` 😆 

I guess the commonalities in this rule are, for type X flagged by the rule
   - X is built-in ("built-in", "intrinsic")
   - the types means "something which has X.prototype in its prototype chain". (gives us "class-type"/"object-type"/"prototype-type"/etc) <-- This makes the rule hard to name
   - that type is bad for some reason, but not necessarily the same reason...
       - in the case of `BigInt`, `Boolean`, `Number`, `String`, `Symbol`, because they refer to the boxed primitive, but the unboxed primitive is also assignable
       - `Object`, because it's terrible; any primitive is assignable, similar to previous
       - `Function`, because it's unsound <-- This makes the rule hard to name
       - _Technically_ we could argue all of these are unsound, but `Function` is really a difference in kind compared to the others, so it's hard to justify using "unsound"/"unsafe" in the name
   - X is uppercase - _technically true but not really the point_ 😄 

So that would lead me to a rough template of `{no/ban}-bad-{built-in/intrinsic}-{instance/prototype/class/object}-types` of which my tentative favorite is `ban-bad-built-in-class-types`.

***

I'm not a fan of "wrapper" since `Object` and `Function` aren't wrappers. And same problem arises with using the word "primitive". If `Function` weren't part of the rule though I think `no-{boxed/wrapped}-primitive-type` would work nicely 🤷

bachmacintosh: 👋 What about `no-ts-wrapper-types` ?

Pro: These types are all defined by TypeScript itself, and the list of banned types not only includes primitive wrappers, but also wrappers for `object` and a wide, any-ish `function` which are technically not primitives [per MDN](https://developer.mozilla.org/en-US/docs/Glossary/Primitive). The name indicates it's related to a TypeScript feature itself like `ban-ts-comment` does.

Con: Might be a little more confusing / need further explanation in the docs.

packages/eslint-plugin/docs/rules/no-uppercase-wrapper-types.mdx
```diff
@@ -0,0 +1,70 @@
...
+> See **https://typescript-eslint.io/rules/no-uppercase-wrapper-types** for documentation.
+
+The JavaScript language includes a set of primitives such as `boolean` and `number`s.
+JavaScript also contains a "wrapper" class object for each of those primitives, such as [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) and [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number).
+TypeScript represents the primitive in lower-case types and the wrapper classes with upper-case
+
+Using the primitives like `0` instead of class constructors like `Number(0)` is generally considered a JavaScript best practice.
+As a result, using the lower-case primitive type names like `number` instead of class names like `Number` is generally considered a TypeScript best practice.
+
+Similarly, TypeScript's `Function` type allows any function-like value.
```
kirkwaiblinger, line 19: Eh, I get we want to Keep It Stupid Simple I feel like this is just isn't quite right. 

Two ideas

***

Option 1. Try to preserve "Similarly" in order to attempt to tie these things together... 

`Similarly, "Function" refers to objects created with the "Function" constructor, i.e. "new Function("eval'd code with implicit args")", rather than ordinary function definitions. This is unsafe because yada yada yada`

***

Option 2. Just pivot.

`The Function type is unsound and permits being called with any number of arguments, yada yada yada`

packages/eslint-plugin/docs/rules/no-uppercase-wrapper-types.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+If your project is a rare one that intentionally deals with the class equivalents of primitives, it might not be worthwhile this rule.
+You might consider using [ESLint disable comments](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) for those specific situations instead of completely disabling this rule.
+
+## Further Reading
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Ehh I'm -1 on both of these. We already have a lot of links and the TypeScript ones aren't particularly informative. I don't think they add anything not mentioned in the MDN ones.

packages/eslint-plugin/docs/rules/no-uppercase-wrapper-types.mdx
JoshuaKGoldberg: Some more suggestions from across the internet:

* [`no-boxed-primitive-types`](https://bsky.app/profile/molisani.bsky.social/post/3ksl4ghxre32b) (@molisani)
* [`no-class-wraps`](https://x.com/alayfalupe/status/1791516887219568835) (@bumpylumps)
* [`no-class-wrapper-types`](https://x.com/bennycode/status/1791011736771793277) (@BennyCode)
* [`no-object-instances-as-types`](https://x.com/paul_melero/status/1790936516958573047) (@gangsthub)
* [`no-primitive-constructors`](https://x.com/niklas_wortmann/status/1790895990460879031) (@niklas-wortmann)
* [`no-primitive-wrappers`](https://x.com/davidlj95/status/1790912764887580767) (@davidlj95)
* [`no-primordial-constructors`](https://x.com/b0neskull/status/1790933347696509297) (@boneskull)
* [`no-upper-case-primitive-type`](https://x.com/bradzacher/status/1790905653613207792) (@bradzacher)
* [`no-wrapper-types`](https://x.com/andhaveaniceday/status/1790934705602531491) (@jakebailey)
* [`no-wrapper-types` or `no-primitive-wrapper-types`](https://elk.zone/fosstodon.org/@shivelysanders/112482007343550012) (@sandersn) 
* [`prefer-lowercase-intrinsic-type` or `prefer-intrinsic-type`](https://x.com/SidaChen63/status/1790928785245950044) (@Josh-Cena)
* [`prefer-primitive-type` (-thingy)](https://x.com/the_ult_dev/status/1790992356495249520) (@the-ult)

Some general thoughts...

* _"Primitive"_ _would_ be something nice to include in there, but as @Josh-Cena pointed out, `object` isn't a primitive and so that's inaccurate
* Moving `Function` bans into a separate rule (https://github.com/typescript-eslint/typescript-eslint/issues/9108#issuecomment-2128385390) would help narrow down the naming
    * It'd be nice for naming to also move out `Object` bans into its own rule so we could then use _"primitive"_ after all, but I don't know of any justifications on that one 😕 
* It's hard to justify using a term that isn't in common usage already. Words like _"boxed"_ and _"primordial"_ might (previously or presently) be good descriptors, but if they don't match MDN docs or common StackOverflow nomenclature, users will have a hard time with them
* We don't generally put abbreviations in rule names anymore. It's hard enough for users to understand lint rules, let alone have to remember random abbreviations in there too.
* I think it'd be nice to include some indication in the name that it's about _types_ / _type annotations_, not generally using those things. E.g. ending with `-types`. That'd help disambiguate from other rules that do ban _usage_ of some types (e.g. `no-unsafe-*` for `any`s).

That leaves a few finalists IMO:

* `no-class-wrapper-types`
* `no-intrinsic-wrapper-types`
* `no-intrinsic-class-wrapper-types`
* `no-primitive-class-wrapper-types`
* `prefer-lowercase-intrinsic-types`: _if_ we split out `Function` into its own rule, since lower-case `function` isn't a type

I'll continue to ruminate...

JoshuaKGoldberg: @kirkwaiblinger what do you mean about `Object` not being a wrapper? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#using_object_constructor_to_turn_primitives_into_an_object_of_their_respective_type refers to the wrapping, though admittedly that's pretty low down on the page for a first mention?

If we want to avoid _"wrapper"_ then that does narrow down the finalists to just one...

jakebailey: I personally find the mention of "class" to be very odd, especially given pages like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/Number do not contain that string.

bennycode: I can see how the term "class" could be confusing as it could be associated with limiting the behaviour in a real `class`. While reading through the MDN docs, I came across this:

> When properties are accessed on primitives, JavaScript **auto-boxes** the value into a **wrapper object** and accesses the property on that object instead. For example, "foo".includes("f") implicitly creates a [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) **wrapper object** and calls `String.prototype.includes()` on that object. This **auto-boxing** behavior is not observable in JavaScript code but is a good mental model of various behaviors

Source: https://developer.mozilla.org/en-US/docs/Glossary/Primitive

Stating the above, would these names be acceptable?

- `no-wrapper-object-types`
- `no-boxing-types`

kirkwaiblinger: > @kirkwaiblinger what do you mean about `Object` not being a wrapper? 

Oh, well, I think I meant the Object type doesn't conceptually refer to a wrapper around a primitive, though as you pointed out (and I wasn't aware) that is a subset of its possible uses. So, feel free to disregard my comment on that, my mind has been changed 🙂

packages/eslint-plugin/docs/rules/no-uppercase-wrapper-types.mdx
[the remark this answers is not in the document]
kirkwaiblinger: Very true. 

If anything, I would say your writeup is significantly more informative about what the TypeScript types mean than anything TypeScript docs!

packages/eslint-plugin/docs/rules/no-uppercase-wrapper-types.mdx
JoshuaKGoldberg: I like `no-wrapper-object-types` a lot! _"Wrapper"_ might be a bit general on its own, but the built-in upper-case types  are the only ones I can think of that that jump to mind as _"wrapper object types"_. Nice @bennycode! 

Renaming now, but if anybody doesn't like this then please do continue the conversation. 🙁

kirkwaiblinger (approved):
It's all coming together

packages/eslint-plugin/docs/rules/ban-types.mdx
```diff
@@ -15,6 +15,16 @@ It's often a good idea to ban certain types to help with consistency and safety.
 This rule bans specific types and can suggest alternatives.
 Note that it does not ban the corresponding runtime objects from being used.
 
+:::danger Deprecated
+This rule is deprecated and now split across several rules:
```
bradzacher, line 19: Why are we deprecating the rule?
There's still value in the rule - just that the default config can be empty now.

People use this rule all the time to ban various types across their codebase. It's the only way to reliably target type names only without hitting value names.

packages/eslint-plugin/src/rules/ban-types.ts
```diff
@@ -107,9 +107,14 @@ export default createRule<Options, MessageIds>({
     type: 'suggestion',
     docs: {
       description: 'Disallow certain types',
-      recommended: 'recommended',
     },
     fixable: 'code',
+    deprecated: true,
```
bradzacher, line 112: As above - why is it deprecated? It is still a useful utility rule even if we've emptied the default config.

bradzacher (changes requested):

packages/eslint-plugin/docs/rules/ban-types.mdx
JoshuaKGoldberg: @bradzacher because it's essentially being renamed to `no-restricted-types`, with the default values split into more targeted rules. I'll update the deprecation notice to make it more clear.

bradzacher: Given this is a major - should we just remove the old rule name entirely?
Is there a benefit to keeping it if we're adding the new rules to recommended and the old rule should almost be a 1:1 name swap in config?

JoshuaKGoldberg: I was thinking this is a pretty user-known rule so leaving it in would help with migration. The name swap might be 1:1 but the behavioral means figuring out which of the three new rules individual config entries / inline comments map to.

kirkwaiblinger: Makes sense to me too...
If we do that for this rule, should we just go ahead and do that for https://github.com/typescript-eslint/typescript-eslint/pull/8821, too, though?

packages/eslint-plugin/src/rules/no-wrapper-object-types.ts
```diff
@@ -0,0 +1,82 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow using confusing built-in primitive class wrappers',
+      recommended: 'recommended',
+    },
+    fixable: 'code',
+    messages: {
+      bannedClassType:
+        'Prefer using the primitive `{{preferred}}` as a type name, rather than the class wrapper `{{typeName}}`.',
```
bradzacher, line 29:
```suggestion
        'Prefer using the primitive `{{preferred}}` as a type name, rather than the upper-cased `{{typeName}}`.',
```

Opinion: These types aren't really class wrappers. They are interfaces that are technically a super-type of the primitive.

I think we should be less specific with the message here.

packages/eslint-plugin/src/rules/no-wrapper-object-types.ts
```diff
@@ -0,0 +1,82 @@
...
+      bannedFunctionType: [
+        'The `Function` type accepts any function-like value.',
+        'It provides no type safety when calling the function, which can be a common source of bugs.',
+        'It also accepts things like class declarations, which will throw at runtime as they will not be called with `new`.',
+        'If you are expecting the function to accept certain arguments, you should explicitly define the function shape.',
```
bradzacher, line 34: I wonder - is there a good doc on options here? Eg defining `(...args: any) => any` or something?

bradzacher (approved):

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-wrapper-object-types** for documentation.
+
+The JavaScript language includes a set of primitives such as `boolean` and `number`s.
+JavaScript also contains a "wrapper" class object for each of those primitives, such as [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) and [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number).
+TypeScript represents the primitive in lower-case types and instances of the wrapper classes with upper-case.
+
+Using the primitives like `0` instead of class constructors like `Number(0)` is generally considered a JavaScript best practice.
```
Josh-Cena, line 16:
```suggestion
Using the primitives like `0` instead of class constructors like `new Number(0)` is generally considered a JavaScript best practice.
```

`Number()` is the correct way to do coercion.

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -0,0 +1,74 @@
...
+As a result, using the lower-case primitive type names like `number` instead of class names like `Number` is generally considered a TypeScript best practice.
+
+Similarly, TypeScript's `Function` type refers to objects created with the `Function` constructor, i.e. `new Function("eval'd code with implicit args")`.
```
Josh-Cena, line 19: I think this part implies that uppercase types and wrapper constructor calls are equivalent, which is technically imprecise. The `Function` type denotes anything that _looks like_ a function (i.e. structurally equivalent to its lib interface). This includes classes, etc., which you can't create with the `Function` constructor. Similarly, `number` is a subtype of `Number` basically.

Josh-Cena (changes requested):
I think @bradzacher has given similar opinions on the code

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
bradzacher: for reference:
```ts
typeof Number('1') === 'number'
typeof new Number('1') === 'object'
```

packages/eslint-plugin/docs/rules/ban-types.mdx
bradzacher: @kirkwaiblinger we can land #8821 against main and then remove it in v8
similarly we could technically land this PR in main and then remove `ban-types` in v8 (which is what I originally suggested https://github.com/typescript-eslint/typescript-eslint/pull/8977#issuecomment-2079376537).

I.e. we could backport the new rules to v7 so they're available right now, deprecate `ban-types`, and then do the actual breaking change (removing `ban-types`) in v8.

JoshuaKGoldberg: I like it. Once this PR goes in I can send a followup to backport the changes to v7 / `main`, replacing the deletion with a deprecation.

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
JoshuaKGoldberg: Hmm yeah I see what you mean. But I'm having a hard time thinking of better phrasing 😬 do you have any suggestions @Josh-Cena?

packages/eslint-plugin/src/rules/no-wrapper-object-types.ts
JoshuaKGoldberg: I'm worried about users seeing it and thinking it's tailored to the specific function type.

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -0,0 +1,74 @@
...
+The JavaScript language includes a set of primitives such as `boolean` and `number`s.
+JavaScript also contains a "wrapper" class object for each of those primitives, such as [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) and [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number).
+TypeScript represents the primitive in lower-case types and instances of the wrapper classes with upper-case.
+
+Using the primitives like `0` instead of class constructors like `new Number(0)` is generally considered a JavaScript best practice.
+As a result, using the lower-case primitive type names like `number` instead of class names like `Number` is generally considered a TypeScript best practice.
+
+Similarly, TypeScript's `Function` type refers to objects created with the `Function` constructor, i.e. `new Function("eval'd code with implicit args")`.
+`Function` allows being called with any number of arguments and returns type `any`.
+It's generally better to specify function parameter and return types with function type syntax.
```
Josh-Cena, line 21:
```suggestion
The JavaScript language has a set of language types, but some of them correspond to two TypeScript types, which look similar: `boolean`/`Boolean`, `number`/`Number`, `string`/`String`, `bigint`/`BigInt`, `symbol`/`Symbol`, `object`/Object`.
The difference is that the lowercase variants are compiler intrinsics and specify the actual _runtime types_ (that is, the return value when you use the `typeof` operator), while the uppercase variants are _structural types_ defined in the library that can be satisfied by any user-defined object with the right properties, not just the real primitives.
JavaScript also has a "wrapper" class object for each of those primitives, such as [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) and [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number).
These wrapper objects are assignable to the uppercase types, but not to the lowercase types.

Using the primitives like `0` instead of object wrappers like `new Number(0)` is generally considered a JavaScript best practice.
In general, your functions want to work with real numbers, instead of objects that "look like" numbers.
As a result, using the lower-case primitive type names like `number` instead of class names like `Number` is generally considered a TypeScript best practice.

TypeScript's `Function` type suffers from the same issue of structural typing.
In addition, `Function` allows being called with any number of arguments and returns type `any`, while in reality, it may be a class which is never callable, or it may be a plain object that happens to possess all properties of the `Function` class!
It's generally better to specify function parameters and return types with the function type syntax. If you want a generic "callable" type, use `(...args: never) => unknown`.
```

Josh-Cena:
FWIW I still don't fully entertain the rule name😢 I know I'm being super pedantic, but `Function` and `Object` aren't really "wrapper objects". Anything that involves "wrapper", "object", and "primitive" would sound off to me. I would only prefer names that involve "lower/uppercase", "intrinsic", and "structural".

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -0,0 +1,76 @@
...
+These wrapper objects are assignable to the uppercase types, but not to the lowercase types.
+
+Using the primitives like `0` instead of object wrappers like `new Number(0)` is generally considered a JavaScript best practice.
+JavaScript programs typically work with the real number primitives, rather than objects that "look like" numbers.
+As a result, using the lowercase type names like `number` instead of the uppercase names like `Number` helps make your code behave more reliably.
+
+TypeScript's `Function` type suffers from the same issue of structural typing.
+In addition, `Function` allows being called with any number of arguments and returns type `any`, while in reality, it may be a class which is never callable, or it may be a plain object that happens to possess all properties of the `Function` class!
+It's generally better to specify function parameters and return types with the function type syntax.
```
JoshuaKGoldberg, line 23: @Josh-Cena btw I removed this from the suggestion:

> `If you want a generic "callable" type, use `(...args: never) => unknown`.`

I'm really scared of giving people a single syntax reference. The generic "callable" type one would want changes based on the scenario. And since this is early on in the doc, folks are likely to mistreat it as a true canonical reference.

Josh-Cena: Hmmm, ok. No strong opinions but still think it would be helpful to have some words about alternatives.

JoshuaKGoldberg:
_**OK THAT DOES IT!**_ I'm splitting out a dedicated `no-unsafe-function-type` rule. 😂 

Having an absurdly difficult time naming a rule is a strong indicator that the rule's trying to do too much. Even if the two areas (functions and objects) are conceptually very close, and/or implemented very similarly, having a clear user-friendly name is very valuable.

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
JoshuaKGoldberg: One advantage of having a dedicated rule (https://github.com/typescript-eslint/typescript-eslint/pull/9102#issuecomment-2142874497 -> https://github.com/typescript-eslint/typescript-eslint/pull/9102#issuecomment-2143965277) is we can get a bit more dedicated space in the docs. So, what are the different function forms we might want to mention?

Maybe:

* `() => void`: a function that has no parameters and whose return is ignored
* `(...args: never) => unknown`: a general "callable" type
* `(...args: unknown[]) => unknown`: a general "callable" type, also?

...I personally don't know of general practice for choosing between the `: never` and the `unknown[]`.

cc @bradzacher from https://github.com/typescript-eslint/typescript-eslint/pull/9102#discussion_r1616432707

packages/eslint-plugin/src/rules/no-wrapper-object-types.ts
JoshuaKGoldberg: https://github.com/typescript-eslint/typescript-eslint/pull/9102#discussion_r1623618173

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
Josh-Cena: `(...args: never) => unknown` is the top type for callables. Every function is assignable to it but itself is not callable because you don't know what to call it with. `(...args: unknown[]) => unknown` is actually callable with anything, but only things that accept `...args: unknown[]` (like `console.log`) are assignable to it, so it's pretty useless in practice.

I think when people use `Function` to mean "any function" they mean `(...args: never) => unknown`. It soundly forbids you from calling it.

JoshuaKGoldberg: Thanks, that's a great explanation 😄. 2508d846f

packages/eslint-plugin/docs/rules/no-uppercase-wrapper-types.mdx
JoshuaKGoldberg: Update: following https://github.com/typescript-eslint/typescript-eslint/pull/9102#issuecomment-2142874497, I split out a `no-unsafe-function-type`. @Josh-Cena makes a good point that the function type isn't a wrapper.

That leaves the question of whether the `no-wrapper-object-types` name is ok to include `Object`. I.e. what @kirkwaiblinger mentioned: that it's not technically a wrapper itself. I'm in favor of allowing this: more convenience at the cost of _slightly_ less pedantically correct rule name. But I could be convinced otherwise if we have a good rule name alternative and/or have justification to move `Object` into its own rule (do we?).

packages/website/plugins/generated-rule-docs/insertions/insertNewRuleReferences.ts
```diff
@@ -19,14 +19,6 @@ const COMPLICATED_RULE_OPTIONS = new Set([
   'naming-convention',
 ]);
 
-/**
- * Rules that do funky things with their defaults and require special code
- * rather than just JSON.stringify-ing their defaults blob
- */
-const SPECIAL_CASE_DEFAULTS = new Map([
-  ['ban-types', '[{ /* See below for default options */ }]'],
-]);
```
JoshuaKGoldberg, line 28: _hell yeah_ 🔪🔪🔪

packages/eslint-plugin/src/rules/no-unsafe-function-type.ts
```diff
@@ -0,0 +1,46 @@
...
+      if (node.type === AST_NODE_TYPES.Identifier && node.name === 'Function') {
+        context.report({
+          node,
+          messageId: 'bannedFunctionType',
+        });
+      }
+    }
+
+    return {
+      TSClassImplements(node): void {
```
Josh-Cena, line 35: Thought: should the two new rules check `implements` and `extends`? You can't extend `number` or `() => {}` so the use of `Number` must either be intentional or refer to another type (since we still don't have scope analysis for this rule)

JoshuaKGoldberg: Someone could, in theory, be doing:
* `extends`: `interface MyFunction extends Function { __brand: 'my-function'; }`
* `implements`: `class Weird implements Function {}`. 

...so for completeness, I think we need to?

Josh-Cena: Is that a bad idea? If they do it there's no reason they are doing it by accident or ignorance.

JoshuaKGoldberg: I'm not strongly opposed to removing this for simplicity. It does feel a little weird to me to leave in a known thing that nobody should really ever do - even if it's only doable with great intent.

Josh-Cena: Let's leave them in for now, but at least `extends Function` feels fine to me.

packages/eslint-plugin/docs/rules/no-unsafe-function-type.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-unsafe-function-type** for documentation.
+
+TypeScript's built-in `Function` type allows being called with any number of arguments and returns type `any`.
+`Function` also allows classes or plain objects object that happens to possess all properties of the `Function` class.
```
kirkwaiblinger, line 13:
```suggestion
`Function` also allows classes or plain objects that happen to possess all properties of the `Function` class.
```

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -9,16 +9,14 @@ import TabItem from '@theme/TabItem';
...
+JavaScript also has a "wrapper" class object for each of those primitives, such as [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) and [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number).
+These wrapper objects are assignable to the uppercase types, but not to the lowercase types.
 
-Using the primitives like `0` instead of class constructors like `Number(0)` is generally considered a JavaScript best practice.
-As a result, using the lower-case primitive type names like `number` instead of class names like `Number` is generally considered a TypeScript best practice.
-
-Similarly, TypeScript's `Function` type refers to objects created with the `Function` constructor, i.e. `new Function("eval'd code with implicit args")`.
-`Function` allows being called with any number of arguments and returns type `any`.
-It's generally better to specify function parameter and return types with function type syntax.
+Using the primitives like `0` instead of object wrappers like `new Number(0)` is generally considered a JavaScript best practice.
```
kirkwaiblinger, line 17: For flow, you might link this and the next sentence with a conjunction such as "..., since ..." or something like that.

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -9,16 +9,14 @@ import TabItem from '@theme/TabItem';
...
+JavaScript programs typically work with the real number primitives, rather than objects that "look like" numbers.
```
kirkwaiblinger, line 18: might find a way to mention that this is _required_ for equality to work correctly, (as well as truthiness/falsiness but that's less important)

packages/website/blog/2024-05-27-announcing-typescript-eslint-v8-beta.mdx
```diff
@@ -187,11 +187,45 @@ Several rules are changed in significant enough ways to be considered breaking c
...
+#### Replacement of `ban-types`
+
+[`@typescript-eslint/ban-types`](https://typescript-eslint.io/rules/ban-types) has long been one of the more controversial rules in typescript-eslint.
+It served two purposes:
+
+- Allowing users to ban a configurable list of types from being used in type annotations
+- Banning confusing or dangerous built-in types such as `Function` and `Number`
+
+Notably, `ban-types` banned the built-in `{}` ("empty object") type in TypeScript.
+The `{}` type is a common source of confusion for TypeScript developers because it matches _any non-nullable_ object, including primitives like `""`.
```
kirkwaiblinger, line 200:
```suggestion
The `{}` type is a common source of confusion for TypeScript developers because it matches _any non-nullable_ value, including primitives like `""`.
```

kirkwaiblinger (commented):
Some nits and stuff but looks good to me 👍

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
JoshuaKGoldberg: I like the idea but can't think of phrasing I like here to go with it. Do you have a suggestion?

kirkwaiblinger (approved):
🔥🔥

packages/eslint-plugin/docs/rules/ban-types.md
```diff
@@ -0,0 +1,22 @@
+:::danger Deprecated
+
+The old `ban-types` rule encompassed multiple areas of functionality, and so has been split into several rules.
+
+**[`no-restricted-types`](./no-restricted-types.mdx)** is the new rule for banning a configurable list of type names.
+It has no options enabled by default.
```
bradzacher, line 6: > And is akin to rules like `no-restricted-syntax`, `no-restricted-globals`, `no-restricted-exports`, `no-restricted-imports`, `no-restricted-properties`, `id-denylist`.

Idk if it's worth mentioning these or some of these to help contextualise the rule? (ditto for adding this to the new rule's docs).

bradzacher (approved):
My one feedback:
These two rules must use scope analysis to ensure the reference is to the global type, nor a local one.

Eg this code should be fine:
```ts
type T<Symbol> = Symbol;
type U<UU> = UU extends T<infer Function>
  ? Function : never;
```

Without scope analysis I believe the first case above will autofix to `type T<Symbol> = symbol;` and break the code!

I say this because we specifically received this feedback in the past from users but had rejected it because "`ban-types` is intentionally dumb and simple and just bans the name".
Given these rules are now targeted and toight - there's no reason we can't now add the scope analysis to be safe.

---

Otherwise LGTM!




---

Is this all documented in the v8 blog post?

packages/eslint-plugin/docs/rules/ban-types.md
JoshuaKGoldberg: Good call, I added a few in.

bradzacher (approved):

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
JoshuaKGoldberg: @kirkwaiblinger I really want to get this in to unblock testing on community repos, so I'm merging now - but let's definitely revisit the docs before v8 goes stable!

JoshuaKGoldberg:
Merging this now with _roughly_ resolved conversations (except some more docs work) to unblock #9141.

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
kirkwaiblinger: see https://github.com/typescript-eslint/typescript-eslint/pull/9363

FloEdelmann:
https://typescript-eslint.io/blog/announcing-typescript-eslint-v8-beta/ still mentions

> [#9102](https://github.com/typescript-eslint/typescript-eslint/pull/9102) is still in review; we'll update this post when the migration path is settled

Issue #9028, opened 2024-04-29
Repo: Raise TSConfig lib/target support range to match ESLint's


JoshuaKGoldberg:
### Suggestion

ESLint uses [`Object.hasOwn()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) (e.g. in [`lib/rules/prefer-object-has-own`](https://github.com/eslint/eslint/blob/0f5df509a4bc00cff2c62b90fab184bdf0231322/lib/rules/prefer-object-has-own.js#L2)). That's a good sign that we should be able to use it too!

But, our TSConfig lib range is only up to `"ES2021"` right now. That doesn't allow `Object.hasOwn`. Which makes me think that we should upgrade our `lib` and/or `target` in our TSConfigs to match ESLint's to at least... `"ES2022"`? Investigation required.

Note: please target this against the `v8` branch. Just to be safe, let's treat it as a followup to #8929 -> #8973.

💖

JoshuaKGoldberg:
#9113 was merged into the `v8` branch. ✅

PR #9113, opened 2024-05-17
chore: use recommended config for Node.js 18 in TSConfig

lesha1201:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/9028

## Overview

Changed TSConfig to target Node.js 18 since this is the minimal version we support. According to https://node.green/, Node.js 18 fully supports ES2022 and ES2023:

Table that shows Node.js 18 supports ES2022 and ES2023 fully

There is a recommended TSConfig for Node.js 18 - https://github.com/tsconfig/bases/blob/main/bases/node18.json and I followed it as a reference.

In another PR, I'm planning to enable ESLint https://eslint.org/docs/latest/rules/prefer-object-has-own rule and fix all related errors.

### DIFF
--- a/packages/eslint-plugin-internal/tsconfig.json
+++ b/packages/eslint-plugin-internal/tsconfig.json
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
     "composite": false,
-    "target": "ES2022",
     "rootDir": "."
   },
   "include": ["src", "typings", "tests", "index.d.ts"],
--- a/packages/scope-manager/tsconfig.spec.json
+++ b/packages/scope-manager/tsconfig.spec.json
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "outDir": "../../dist/out-tsc",
-    "module": "Node16",
     "types": ["jest", "node"]
   },
   "exclude": ["tests/fixtures"],
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -6,7 +6,7 @@
     "declaration": true,
     "declarationMap": true,
     "esModuleInterop": true,
-    "lib": ["ES2021"],
+    "lib": ["ES2023"],
     "module": "Node16",
     "moduleResolution": "Node16",
     "noImplicitReturns": true,
@@ -16,6 +16,6 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "ES2021"
+    "target": "ES2022"
   }
 }


packages/eslint-plugin-internal/tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
     "composite": false,
-    "target": "ES2022",
```
lesha1201, line 5: It extends `tsconfig.base.json` so no need to specify it anymore.

packages/scope-manager/tsconfig.spec.json
```diff
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "outDir": "../../dist/out-tsc",
-    "module": "Node16",
```
lesha1201, line 5: It extends `tsconfig.base.json` so no need to specify it.

auvred:
Note: it looks like https://github.com/typescript-eslint/typescript-eslint/pull/9038 already contains some changes in this area

tsconfig.base.json
```diff
@@ -6,7 +6,7 @@
     "declaration": true,
     "declarationMap": true,
     "esModuleInterop": true,
-    "lib": ["ES2021"],
+    "lib": ["ES2023"],
```
jakebailey, line 9: I was going to ask why ES2023 was legal, but node.green does in fact show that all versions of Node 18 support all runtime features of ES2023. I've fixed https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping to reflect this (which matches the bases too).

JoshuaKGoldberg: Nice!!

tsconfig.base.json
```diff
@@ -16,6 +16,6 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "ES2021"
+    "target": "ES2022"
```
JoshuaKGoldberg, line 19: Aside: I'm surprised TS 5.4 doesn't support an ES2023 target. I tried that locally and got:

```plaintext

   ✖  nx run ast-spec:build


      ../../tsconfig.base.json(19,15): error TS6046: Argument for '--target' option must be: 'es5', 'es6', 'es2015', 'es2016', 'es2017', 'es2018', 'es2019', 'es2020', 'es2021', 'es2022', 'esnext'.
      Warning: command "yarn build" exited with non-zero status code
```

https://github.com/microsoft/TypeScript/issues/56303 - turns out es2023 === es2022 for target so they omitted it. Huh.

JoshuaKGoldberg (approved):
Super, thanks! 🚀

tsconfig.base.json
jakebailey: Yes, this was rectified in https://github.com/microsoft/TypeScript/pull/58140 and we plan to not skip again.

PR #3737, opened 2021-08-15
feat: support ESLint v8

ota-meshi:
This PR updates this plugin to support ESLint v8.

However, currently eslint is still v8.0.0-beta-0.

**Summary of changes**

- ~~Change to get ESLint core rules from Linter instance.~~ See https://github.com/typescript-eslint/typescript-eslint/pull/3737#issuecomment-899159390.
- Adde `meta.hasSuggestions` to the rules that provide suggestions.
- Fix `no-dupe-class-members` and `no-extra-semi` to work with v8.0.0-beta.0.
- Change `CLIEngine` to be `undefined` in eslint v8.
- Add `jest-resolver.js`
  jest doesn't seem to support the exports map in package.json at this time.
  For this reason, jest-resolver.js is added as a workaround.
  See https://github.com/facebook/jest/issues/9771.

**About Lint errors:**
Since eslint-plugin-import and eslint-plugin-jest do not yet support eslint v8, still get a run error.

**Test failed on Node 10.x:**
Test fails because eslint v8 doesn't support Node 10.

Should I change the problematic workflow and use eslint v7 only for Node 10.x testing and linting? Or should I delete the workflow for Node 10.x?

PR #9119, opened 2024-05-18
chore: use eslint 9 internally

### LINKED ISSUE #9082: Repo: Use ESLint v9 for our internal linting
### Suggestion

Now that #9002 is merged, our `v8` branch supports ESLint v9. Yay! But, #9002 didn't actually bump our _internal_ `devDependency`. https://github.com/typescript-eslint/typescript-eslint/pull/9002#discussion_r1591704965 mentions the three plugins whose lack of ESLint v9 support were blocking us.

But, now that https://eslint.org/blog/2024/05/eslint-compatibility-utilities exists, we should be able to move to ESLint v9 without losing our preferred linting. Which would be great. Accepting PRs to our `v8` branch to bring back the bump of our `"eslint"` `devDependency` from 8 to 9.

💖

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9082 

## Overview
Use eslint@9 internally.


bradzacher:
We should delete this CI job
https://github.com/typescript-eslint/typescript-eslint/blob/87a0a484233ba76391be2d3a1588302051888485/.github/workflows/ci.yml#L137-L154

have we got an existing integration test to check for regressions on eslint v8?

abrahamguo:
@bradzacher there are currently no integration or CI tests for eslint 8. I can add one if you want, but do you feel that's necessary since we ask for eslint 9 in both our peerDependencies and devDependencies?

Doesn't seem like anything similar was added for eslint 7 when we moved from eslint 7=>8 (#3737) 

Also, seems unrelated to this PR since this is about using eslint 9 _internally_

bradzacher:
Looks like there's already an integration test for v8 - https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/integration-tests/fixtures/eslint-v8
So nothing to add!

> Also, seems unrelated to this PR since this is about using eslint 9 internally

By moving to v9 internally our CI is no longer running against v8 - so we lose the test coverage asserting that we don't regress v8 compat.

But given we have the above integration test - we still have enough coverage.
It's not a super exhaustive test but it'll do (it's the same integration test coverage we maintained when we migrated to v8 internally)

mattbrannon:
Just wanted to point out that one of your Netlify jobs has been running for nearly a week.
https://github.com/typescript-eslint/typescript-eslint/pull/9119/checks?check_run_id=25428243220

abrahamguo:
@mattbrannon that's nothing compared to #8673, whose job has been running for 3 weeks 😉

It does seem like that specific job sometimes hangs indefinitely. Not sure why

eslint.config.mjs
```diff
@@ -29,15 +30,20 @@ export default tseslint.config(
     plugins: {
       ['@typescript-eslint']: tseslint.plugin,
       ['@typescript-eslint/internal']: tseslintInternalPlugin,
-      ['deprecation']: deprecationPlugin,
+      // https://github.com/gund/eslint-plugin-deprecation/issues/85
+      // https://github.com/typescript-eslint/typescript-eslint/issues/8988
+      ['deprecation']: fixupPluginRules(deprecationPlugin),
```
abrahamguo, line 35: TODO: looks like this got fixed — try it out

eslint.config.mjs
```diff
@@ -29,15 +30,20 @@ export default tseslint.config(
...
       ['@typescript-eslint/internal']: tseslintInternalPlugin,
-      ['deprecation']: deprecationPlugin,
+      // https://github.com/gund/eslint-plugin-deprecation/issues/78
+      // https://github.com/typescript-eslint/typescript-eslint/issues/8988
+      ['deprecation']: fixupPluginRules(deprecationPlugin),
       ['eslint-comments']: eslintCommentsPlugin,
       ['eslint-plugin']: eslintPluginPlugin,
-      ['import']: importPlugin,
+      // https://github.com/import-js/eslint-plugin-import/issues/2948
+      ['import']: fixupPluginRules(importPlugin),
```
bradzacher, line 39: we might want to consider switching to `eslint-plugin-import-x`?

eslint.config.mjs
```diff
@@ -511,8 +517,8 @@ export default tseslint.config(
     files: ['packages/website/**/*.{ts,tsx,mts,cts,js,jsx}'],
     extends: [
       ...compat.config(jsxA11yPlugin.configs.recommended),
-      ...compat.config(reactPlugin.configs.recommended),
-      ...compat.config(reactHooksPlugin.configs.recommended),
+      ...fixupConfigRules(compat.config(reactPlugin.configs.recommended)),
+      ...fixupConfigRules(compat.config(reactHooksPlugin.configs.recommended)),
```
bradzacher, line 521: it's so weird that we need to fixup the plugin both here and in the plugin def.
seems like quite the flaw in the design of the compat tooling tbh

packages/website-eslint/build.ts
```diff
@@ -31,9 +31,7 @@ function createResolve(
   join: string,
 ): esbuild.OnResolveResult {
   const resolvedPackage = requireResolved(targetPath + '/package.json');
-  return {
-    path: path.join(resolvedPackage, '../src/', join),
-  };
+  return { path: path.join(resolvedPackage, '../src/', join) };
```
JoshuaKGoldberg, line 34: [Style] This looks like an unrelated styling change. Could you please revert it and any others? It's harder to review PRs that have unnecessary whitespace changes - especially these larger, more intricate PRs.

.github/workflows/ci.yml
```diff
@@ -134,25 +134,6 @@ jobs:
         run: yarn stylelint
         working-directory: packages/website
 
-  eslint_v9_tests:
```
JoshuaKGoldberg, line 137: [Testing] We'll still want to test against ESLint 8. I don't think anything would now do that? Unless I'm missing something, this should be switched to `eslint_v8_tests`, not removed.

packages/website/src/theme/MDXComponents/RuleAttributes.tsx
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
JoshuaKGoldberg, line 1: [Non-Actionable] Heh, I wonder how much this will come into conflict with #9339...

packages/website/src/hooks/useClipboard.ts
```diff
@@ -7,12 +7,13 @@ export type useClipboardResult = [copied: boolean, copy: () => void];
...
-      setCopied(true);
-    });
-  }, [setCopied, code]);
+  const copy = useCallback(
+    () =>
+      void navigator.clipboard.writeText(code()).then(() => {
+        setCopied(true);
+      }),
+    [setCopied, code],
+  );
```
JoshuaKGoldberg, line 16: [Question] Why this change? Seems unrelated & something to be reverted to me?

JoshuaKGoldberg (changes requested):
I think we'll need to keep ESLint 8 testing in there for a while - and there is some noise making it harder to review. But encouraging as a start! 🌞

packages/website/src/hooks/useClipboard.ts
abrahamguo: Nope — with ESLint 9, it now reports unused `eslint-disable` comments, like this one 🎉

packages/website-eslint/build.ts
abrahamguo: sorry about that! prettier is annoyingly un-opinionated about objects on one line vs multiple lines and was conflicting with my editor

eslint.config.mjs
abrahamguo: @bradzacher that's actually not true.

This can be removed _if_ you are OK with me editing the section at the top of this file labeled `registering all the plugins up-front`, and removing any plugin for which we use a shared config (for example, `reactPlugin` and `reactHooksPlugin`).
Shared configs automatically register their corresponding plugin, so it is unnecessary to manually register the plugin.

The flip side, of course, would be if you felt that it's important to have a complete up-front list of all plugins used in any part of the config.

If we choose to keep the complete up-front list of plugins, then the reason why we have to use the compat utilities twice, is because the plugins are defined twice (again, redundantly so), and we need to make sure that in _both_ places that each plugin is registered, it is the fixed-up version.

.github/workflows/ci.yml
abrahamguo: We still do have [this integration test for ESLint 8](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/integration-tests/fixtures/eslint-v8).

@bradzacher's [thoughts from above](https://github.com/typescript-eslint/typescript-eslint/pull/9119#issuecomment-2124749156)

and looking at the history of this file, it seems like we have not had similar tests in the past for other versions of ESLint

bradzacher: I don't think we need to worry too much.
There aren't any new APIs we need to worry about for plugins at least.
Given we have extension rules we need to be careful testing across majors. I'm surprised none of our tests broke!

When we had v7+v8 support we only relied on the integration test and not a full CI.
I think this might just be good enough.

eslint.config.mjs
JoshuaKGoldberg: I'd be in support of this as a separate followup yeah.

.github/workflows/ci.yml
JoshuaKGoldberg: Ah, great, this is enough for me. Thanks! 🎉

JoshuaKGoldberg (approved):
Whoohoo! Thanks!! 🚀

packages/eslint-plugin/rules.d.ts
```diff
@@ -75,5 +75,4 @@ export type TypeScriptESLintRules = Record<
 >;
 
 declare const rules: TypeScriptESLintRules;
-// eslint-disable-next-line import/no-default-export
```
abrahamguo, line 78: this is an unnecessary suppression reported by ESLint 9. It is unnecessary because this rule is already disabled on all `d.ts` files in our eslint config file

abrahamguo:
ESLint 9 displays unused lint suppressions as _warnings_, not errors, so there were 3 unused lint suppressions that were displayed in the pipeline but did not block it. I've removed all of those, and provided an explanation for each one.

eslint.config.mjs
auvred, line 1: > ESLint 9 displays unused lint suppressions as _warnings_, not errors, so there were 3 unused lint suppressions that were displayed in the pipeline but did not block it. I've removed all of those, and provided an explanation for each one.

Hmmm, turns out our `eslint.config.mjs` doesn't have  `linterOptions: { reportUnusedDisableDirectives: 'error' }`. Maybe this is a good chance to turn it on?

---

This seems a bit suspicious to me though: this option has not been enabled for this repository until now. Maybe there are some considerations in favor of leaving this option disabled?

abrahamguo: Just want to clarify that it was the upgrade from ESLint 8 to ESLint 9 that caused it to begin reporting unnecessary lint suppressions.
I feel like it makes sense to keep it at the default value, or change the level from `warning` to `error`, but I'm open to further discussion.

JoshuaKGoldberg: Yeah I'd +1 switching to `error` as a followup. Great find!

auvred (approved):
:rocket:

PR #9121, opened 2024-05-19
fix(typescript-estree): change mjs/mts files to always be parsed by ESM

### LINKED ISSUE #9101: Bug(typescript-estree): always parse mts/mjs as ESM for non-type-aware parsing

### Relevant Package

typescript-estree

### Playground Link

_No response_

### Repro Code

```TypeScript
import { parse } from "@typescript-eslint/typescript-estree"

const ast = parse('await(1)', {
  project: false,
  filePath: 'foo.mts'
})

console.log(ast.body[0].expression)
```

### ESLint Config

```javascript
N/A
```

### tsconfig

```jsonc
None
```

### Expected Result

The expression should be `AwaitExpression` instead of `CallExpression`.

### Actual Result

As described above.

### Additional Info

_No response_

### Versions

| `@typescript-eslint/typescript-estree` | `7.8.0` |
| `TypeScript`                           | `5.4.5` |
| `node`                                 | `20` |

goldentrash:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9101

## Overview

Change `mjs`/`mts` files to **always** be parsed by ESM

### How I solved this problem

- If the extension of the file is `mjs`/`mts` via `ParseOptions.filePath`, add an empty export at the end of the source code: (`export {};`)
- This method was inspired by a long-time favorite method for resolving `isolatedModules` errors that originate from tsconfig settings.
- According to [MDN documentation](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#description), this method does not affect functionality or logic.
  - **MDN**: Note that `export {}` does not export an empty object — **it's a no-op declaration that exports nothing** (an empty name list).
- I have confirmed that this fixes the issue mentioned in the #9101, see the issue for details 

### Reasons not to modify `convertNode` and `createSourceFile`.

- In `/tyscript-estree/src/convert.ts/Converter/convertNode`, modifying `sourceType` when it is `case SyntaxKind.SourceFile` does not make sense, I have left the details in the issue.
- Because `createSourceFile` relies on an external module called typescript (`ts.createSourceFile`), tweaking the parameters of this module is the only way to keep it out of scope of typescript-eslint.
  - Adjusting `setExternalModuleIndicator` doesn't make sense, I've left the details in an issue.
  - Adjusting `indicator` doesn't make sense, I've left the details in an issue.

### Why I thought it was in-appropriate to look for another API instead of `ts.createSourceFile`.

- Currently, `typescript-estree` relies **heavily** on the return value of `ts.createSourceFile`.
- Therefore, changing the API requires a change large enough **to be categorized as a `breaking-change`**.

---

If you have any feedback on function/variable naming or code block repositioning, I'll take it on board and fix the PR.

### DIFF
--- a/packages/typescript-estree/src/parser.ts
+++ b/packages/typescript-estree/src/parser.ts
@@ -1,5 +1,6 @@
 import debug from 'debug';
-import type * as ts from 'typescript';
+import path from 'path';
+import * as ts from 'typescript';
 
 import { astConverter } from './ast-converter';
 import { convertError } from './convert';
@@ -125,11 +126,28 @@ function parse<T extends TSESTreeOptions = TSESTreeOptions>(
   return ast;
 }
 
+function isESM<T extends TSESTreeOptions = TSESTreeOptions>(
+  options: T | undefined,
+): boolean {
+  const extension = path
+    .extname(options?.filePath ?? '')
+    .toLowerCase() as ts.Extension;
+
+  return extension === ts.Extension.Mjs || extension === ts.Extension.Mts;
+}
+
 function parseWithNodeMapsInternal<T extends TSESTreeOptions = TSESTreeOptions>(
   code: ts.SourceFile | string,
   options: T | undefined,
   shouldPreserveNodeMaps: boolean,
 ): ParseWithNodeMapsResult<T> {
+  /**
+   * Ensure that `mjs`/`mts` files are always treated as ESM
+   */
+  if (typeof code === 'string' && isESM(options)) {
+    code += '\n' + 'export {}';
+  }
+
   /**
    * Reset the parse configuration
    */


packages/typescript-estree/src/parser.ts
```diff
@@ -125,11 +126,28 @@ function parse<T extends TSESTreeOptions = TSESTreeOptions>(
...
 function parseWithNodeMapsInternal<T extends TSESTreeOptions = TSESTreeOptions>(
   code: ts.SourceFile | string,
   options: T | undefined,
   shouldPreserveNodeMaps: boolean,
 ): ParseWithNodeMapsResult<T> {
+  /**
+   * Ensure that `mjs`/`mts` files are always treated as ESM
+   */
+  if (typeof code === 'string' && isESM(options)) {
+    code += '\n' + 'export {}';
```
bradzacher, line 148: We cannot change the code at all.
Doing so changes the resulting AST and its ranges.
This may also cause lint rules to report on code that doesn't exist.

We need to accomplish this without mutating the passed code at all.

bradzacher (changes requested):

packages/typescript-estree/src/parser.ts
goldentrash: Ah! I hadn't thought of that, you're right.

I'll think of another way to do it.

Thanks for the review :)

goldentrash:
Hi @bradzacher , I've spent the last few days thinking of other ways to achieve our goal **without modifying the `code`**.
From the typescript playground, I knew that ts can properly recognize `mjs`/`mts` files as ESM, so I tried to analyze the Typescript code a bit more.

I found the [code](https://github.com/microsoft/TypeScript/blob/9370347f5b6bc72902c8d71ac80d20738849dabd/src/compiler/utilities.ts#L8750) below and it seems to be acceptable.

```ts

/** @internal */
export function getSetExternalModuleIndicator(options: CompilerOptions): (file: SourceFile) => void {
    // TODO: Should this callback be cached?
    switch (getEmitModuleDetectionKind(options)) {
        case ModuleDetectionKind.Force:
            // All non-declaration files are modules, declaration files still do the usual isFileProbablyExternalModule
            return (file: SourceFile) => {
                file.externalModuleIndicator = isFileProbablyExternalModule(file) || !file.isDeclarationFile || undefined;
            };
        case ModuleDetectionKind.Legacy:
            // Files are modules if they have imports, exports, or import.meta
            return (file: SourceFile) => {
                file.externalModuleIndicator = isFileProbablyExternalModule(file);
            };
        case ModuleDetectionKind.Auto:
            // If module is nodenext or node16, all esm format files are modules
            // If jsx is react-jsx or react-jsxdev then jsx tags force module-ness
            // otherwise, the presence of import or export statments (or import.meta) implies module-ness
            const checks: ((file: SourceFile, options: CompilerOptions) => Node | true | undefined)[] = [isFileProbablyExternalModule];
            if (options.jsx === JsxEmit.ReactJSX || options.jsx === JsxEmit.ReactJSXDev) {
                checks.push(isFileModuleFromUsingJSXTag);
            }
            checks.push(isFileForcedToBeModuleByFormat);
            const combined = or(...checks);
            const callback = (file: SourceFile) => void (file.externalModuleIndicator = combined(file, options));
            return callback;
    }
}
```

Would this be acceptable?

goldentrash:
Hi! **I'm always ready to update the PR** if I need to change my solution to fix this bug!
I'll keep an email alarm on for reviews :)

andrewbranch:
Does the detection need to agree with what the TypeScript compiler would do given the file extension and tsconfig options? If so, you do need to implement something like that snippet you found in TS that considers the computed value of `--moduleDetection`.

goldentrash:
> Does the detection need to agree with what the TypeScript compiler would do given the file extension and tsconfig options? If so, you do need to implement something like that snippet you found in TS that considers the computed value of `--moduleDetection`.

I think it needs to reflect what TSC sees as you said, and I'll try to modify the PR to take into account things like what you mentioned(`-- moduleDetection`). I'll have to analyze the TS repo and the typescript-eslint repo a bit more.

Thanks for the review! :)

andrewbranch:
Unfortunately, I think you’re essentially going to have to choose between porting some TS implementation details (i.e. what options imply what other options) and accessing internal functions 😕

goldentrash:
Thanks for the advice @andrewbranch!

I think we need to decide something else first to solve this problem: "Which should be **dominant** in the settings for the module, [`eslint.config.js`](https://eslint.org/docs/latest/use/configure/language-options#specifying-javascript-options) or [`tsconfig.json`](https://www.typescriptlang.org/tsconfig/#moduleDetection)?"

My suggestion would be to give `eslint.config.js` precedence. This is because in this case we can utilize [`ParserOptions.sourceType`](https://github.com/typescript-eslint/typescript-eslint/blob/ceb9b880b70d5b55e691a51c60d1b12612fbafd2/packages/types/src/parser-options.ts#L75).

1. In [`packages/parser/src/parser.ts/parseForESLint`](https://github.com/typescript-eslint/typescript-eslint/blob/ceb9b880b70d5b55e691a51c60d1b12612fbafd2/packages/parser/src/parser.ts#L99), add `sourceType` to  `parserOptions`
    ```ts
    function parseForESLint(
      code: ts.SourceFile | string,
      options?: ParserOptions | null,
    ): ParseForESLintResult {
      // https://eslint.org/docs/user-guide/configuring#specifying-parser-options
      // if sourceType is not provided by default eslint expect that it will be set to "script"
      if (options.sourceType !== 'module' && options.sourceType !== 'script') {
        options.sourceType = 'script';
      }
 
      const parserOptions: TSESTreeOptions = {};
      parserOptions.sourceType = options.sourceType;  // like this!
    }
    ```
2. In [`packages/typescript-estree/src/parseSettings/createParseSettings.ts/createParseSettings`](https://github.com/typescript-eslint/typescript-eslint/blob/ceb9b880b70d5b55e691a51c60d1b12612fbafd2/packages/typescript-estree/src/parseSettings/createParseSettings.ts#L99), set `setExternalModuleIndicator` based on `options.sourceType`
    ```ts
    export function createParseSettings(
      code: ts.SourceFile | string,
      options: Partial<TSESTreeOptions> = {},
    ): MutableParseSettings {
        setExternalModuleIndicator:
          options.sourceType === 'module' ||
          extension === ts.Extension.Mjs ||
          extension === ts.Extension.Mts
            ? (file): void => {
                file.externalModuleIndicator = true;
              }
            : undefined,
        ...
    }
    ```

Incidentally, if we do this, how about adding a rule to "warn if the module settings in `eslint.config.js` and `tsconfig.json` are different"?

JoshuaKGoldberg:
👋 Ping @goldentrash - is this still something you have time & energy for?

goldentrash:
> 👋 Ping @goldentrash - is this still something you have time & energy for?

I'm so sorry for the late response. I'm interested in solving this problem! However, I've been so busy with personal matters over the past three weeks that I haven't been able to give it much attention. I'm really sorry for not responding to this, but the good news is that as of today, I'm finally able to wrap up all of my pressing matters, so I can continue my challenge to solve this problem this weekend. I'm still interested in this issue, and I'm willing to continue the challenge to resolve it. Thank you.

goldentrash:
As discussed above, I modified the code to determine the `sourceType` based on the settings in `eslint.config.js` rather than `tsconfig.json`, and wrote the associated tests. There are a few things I need to ask, so I'll leave a comment.

1. There are three tests that were requested. I first wrote a test under `typescript-estree/tests/lib/parse.test.ts` that checks for TLA-allowed parsing **based on sourceType and extension**. (I thought this was the **type-aware** thing you were referring to.) However, I didn't quite understand your intent when you mentioned **Project Service**, so I set the `project` attribute to `/tsconfig.json`. Also, I checked the file `typescript-estree/tests/lib/parse.project-true.test.ts` and it seems to validate parent & sibling project detection regardless of ESM parsing, so I didn't write any additional tests. If you can be a bit more specific, I'll write those tests as well.
2. I've added a new property (`sourceType`) to `typescript-estree/src/parser-options.ts`. Based on the comments, it looks like we need to modify the documentation accordingly. Do I need to modify the relevant pages for this as well, and if so, which files?
3. If you look at `parser/src/parser.ts`, you'll see code like this
    ```ts
    const parserOptions: TSESTreeOptions = {};
    Object.assign(parserOptions, options, {
      sourceType: options.sourceType,     // Here's some code that didn't exist before.
    });

    ...

    const { ast, services } = parseAndGenerateServices(code, parserOptions);
    ast.sourceType = options.sourceType;  // Is this the right place for this code?
     ```
    I was thinking about moving the last line, `ast.sourceType = options.sourceType;`, inside `parseAndGenerateServices`, but I was worried about **side effects** I wasn't aware of, so I left it unmodified for now and kept the existing structure. Should I move that code inside `parseAndGenerateServices()` and `parse()`? (The two functions don't reference each other, so if I move it, I'd have to put it inside both of them.) If not, should I modify `astConverter` in `typescript-estree/src/ast-converter.ts` to fulfill that role together?
4. Finally, my branch now has a **conflict** with `main`. Should I resolve that conflict myself? If so, which method should I use? `Rebase` for linear history or `merge`?

goldentrash:
> However, I didn't quite understand your intent when you mentioned **Project Service**, so I tested `parseAndGenerateServices` using the `project` attribute in `/tsconfig.json`.

I modified the `parseAndGenerateServices` test to not use the `project` attribute. It made sense to test the same thing as the `parse` test, and when I checked the `testParse` helper in the **"isolated parsing"** part of the same test file, the attribute didn't seem important to verify the behavior.

First and foremost, the test you mentioned was written with the intent of **Project Service** in mind, so I thought it was more important to accurately reflect the intent based on your feedback.

goldentrash:
I'm sorry. I accidentally closed this PR, reopened it, but that proved impossible, so I opened a new one. Please see the following PR (#10011)

PR #9167, opened 2024-05-28
fix(eslint-plugin): [no-magic-numbers] fix implementation of the `ignore` option

### LINKED ISSUE #9161: Bug: no-magic-numbers Doesn't apply ignore to type definitions

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.5&fileType=.tsx&code=KYDwDg9gTgLgBDAnmYcDyA7YaoBUDuEcAvHAIxwA%2BcATANxA&eslintrc=N4KABGBEAOA2CuBzAlgO0gLjAbXBKAAgC4Ce0ApgM4DGATstEQLRWxpGR4C6ANHpLXiwqmMKHyFSFGvUYtKbVEQD0qAPZMAtgEMU1JqniaARuVqVR2SAHdttdDzGRkidbXKWAjI4BMXAL5ceP4g-kA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
export type OneOrTwo = 1 | 2;
```

### ESLint Config

```javascript
{
  "plugin": [
    "@typescript-eslint"
  ],
  "rules": {
    "@typescript-eslint/no-magic-numbers": ["warn", {"ignore": [1, 2]}]
  }
}
```

### tsconfig

_No response_

### Expected Result

I expected that because the numeric literals 1 and 2 are in the "ignore" setting of the "no-magic-numbers" rule, they would not be reported as magic numbers.

### Actual Result

The numeric literals 1 and 2 are reported as magic numbers, despite the rule being configured to ignore them.

### Additional Info

Here's a patch:

```
diff --git a/node_modules/@typescript-eslint/eslint-plugin/dist/rules/no-magic-numbers.js b/node_modules/@typescript-eslint/eslint-plugin/dist/rules/no-magic-numbers.js
index 07b7055..c5b6014 100644
--- a/node_modules/@typescript-eslint/eslint-plugin/dist/rules/no-magic-numbers.js
+++ b/node_modules/@typescript-eslint/eslint-plugin/dist/rules/no-magic-numbers.js
@@ -25,6 +25,17 @@ Array.isArray(baseRule.meta.schema)
         },
     },
 });
+/**
+ * Convert the value to bigint if it's a string. Otherwise return the value as-is.
+ * @param {bigint|number|string} x The value to normalize.
+ * @returns {bigint|number} The normalized value.
+ */
+function normalizeIgnoreValue(x) {
+    if (typeof x === "string") {
+        return BigInt(x.slice(0, -1));
+    }
+    return x;
+}
 exports.default = (0, util_1.createRule)({
     name: 'no-magic-numbers',
     meta: {
@@ -49,6 +60,7 @@ exports.default = (0, util_1.createRule)({
     ],
     create(context, [options]) {
         const rules = baseRule.create(context);
+        const ignore = new Set((options.ignore || []).map(normalizeIgnoreValue));
         return {
             Literal(node) {
                 // If it’s not a numeric literal we’re not interested
@@ -76,6 +88,10 @@ exports.default = (0, util_1.createRule)({
                 else if (isParentTSReadonlyPropertyDefinition(node)) {
                     isAllowed = options.ignoreReadonlyClassProperties === true;
                 }
+                // Check if the value is allowed
+                if (ignore.has(node.value)) {
+                    isAllowed = true;
+                }
                 // If we’ve hit a case where the ignore option is true we can return now
                 if (isAllowed === true) {
                     return;
```

Vinccool96:
## PR Checklist

- [X] Addresses an existing open issue: fixes #9161 

## Overview

As discussed in #9161, the `ignore` option wasn't correctly implemented.

`ignoreTypeIndexes` was also added to `defaultOptions` since it was apparently missing.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-magic-numbers.ts
+++ b/packages/eslint-plugin/src/rules/no-magic-numbers.ts
@@ -58,10 +58,13 @@ export default createRule<Options, MessageIds>({
       ignoreNumericLiteralTypes: false,
       ignoreEnums: false,
       ignoreReadonlyClassProperties: false,
+      ignoreTypeIndexes: false,
     },
   ],
   create(context, [options]) {
     const rules = baseRule.create(context);
+
+    const ignored = new Set((options.ignore ?? []).map(normalizeIgnoreValue));
 
     return {
       Literal(node): void {
@@ -76,8 +79,12 @@ export default createRule<Options, MessageIds>({
         // not one of our exception cases, and we’ll fall back to the base rule.
         let isAllowed: boolean | undefined;
 
+        // Check if the node is ignored
+        if (ignored.has(normalizeLiteralValue(node))) {
+          isAllowed = true;
+        }
         // Check if the node is a TypeScript enum declaration
-        if (isParentTSEnumDeclaration(node)) {
+        else if (isParentTSEnumDeclaration(node)) {
           isAllowed = options.ignoreEnums === true;
         }
         // Check TypeScript specific nodes for Numeric Literal
@@ -128,6 +135,40 @@ export default createRule<Options, MessageIds>({
     };
   },
 });
+
+/**
+ * Convert the value to bigint if it's a string. Otherwise, return the value as-is.
+ * @param value The value to normalize.
+ * @returns The normalized value.
+ */
+function normalizeIgnoreValue(
+  value: bigint | number | string,
+): bigint | number {
+  if (typeof value === 'string') {
+    return BigInt(value.slice(0, -1));
+  }
+
+  return value;
+}
+
+/**
+ * Converts the node to its numeric value, handling prefixed numbers (-1 / +1)
+ * @param node the node to normalize.
+ */
+function normalizeLiteralValue(
+  node: TSESTree.BigIntLiteral | TSESTree.NumberLiteral,
+): bigint | number {
+  if (
+    node.parent.type === AST_NODE_TYPES.UnaryExpression &&
+    ['-', '+'].includes(node.parent.operator)
+  ) {
+    if (node.parent.operator === '-') {
+      return -(node.value ?? BigInt(node.raw));
+    }
+  }
+
+  return node.value ?? BigInt(node.raw);
+}
 
 /**
  * Gets the true parent of the literal, handling prefixed numbers (-1 / +1)
--- a/packages/eslint-plugin/src/rules/no-magic-numbers.ts
+++ b/packages/eslint-plugin/src/rules/no-magic-numbers.ts
@@ -80,7 +80,7 @@ export default createRule<Options, MessageIds>({
         let isAllowed: boolean | undefined;
 
         // Check if the node is ignored
-        if (ignored.has(normalizeLiteralValue(node))) {
+        if (ignored.has(normalizeLiteralValue(node, node.value))) {
           isAllowed = true;
         }
         // Check if the node is a TypeScript enum declaration
@@ -154,20 +154,22 @@ function normalizeIgnoreValue(
 /**
  * Converts the node to its numeric value, handling prefixed numbers (-1 / +1)
  * @param node the node to normalize.
+ * @param value the node's value.
  */
 function normalizeLiteralValue(
   node: TSESTree.BigIntLiteral | TSESTree.NumberLiteral,
+  value: number | bigint,
 ): bigint | number {
   if (
     node.parent.type === AST_NODE_TYPES.UnaryExpression &&
     ['-', '+'].includes(node.parent.operator)
   ) {
     if (node.parent.operator === '-') {
-      return -(node.value ?? BigInt(node.raw));
+      return -value;
     }
   }
 
-  return node.value ?? BigInt(node.raw);
+  return value;
 }
 
 /**
--- a/packages/eslint-plugin/tests/rules/no-magic-numbers.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-magic-numbers.test.ts
@@ -118,47 +118,68 @@ type Foo = {
       options: [{ ignoreTypeIndexes: true }],
     },
     {
-      code: `
-type Foo =
-  | 1
-  | -2
-  | 3n
-  | -4n
-  | 5.6
-  | -7.8
-  | 0x0a
-  | -0xbc
-  | 1e2
-  | -3e4
-  | 5e-6
-  | -7e-8
-  | 1.1e2
-  | -3.1e4
-  | 5.1e-6
-  | -7.1e-8;
-      `,
-      options: [
-        {
-          ignore: [
-            1,
-            -2,
-            '3n',
-            '-4n',
-            5.6,
-            -7.8,
-            0x0a,
-            -0xbc,
-            1e2,
-            -3e4,
-            5e-6,
-            -7e-8,
-            1.1e2,
-            -3.1e4,
-            5.1e-6,
-            -7.1e-8,
-          ],
-        },
-      ],
+      code: 'type Foo = 1;',
+      options: [{ ignore: [1] }],
+    },
+    {
+      code: 'type Foo = -2;',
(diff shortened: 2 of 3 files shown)


packages/eslint-plugin/tests/rules/no-magic-numbers.test.ts
JoshuaKGoldberg, line 1: [Testing] A lot of the tests in this file have many cases being checked. The `invalid` ones have a lot of `error` entries. Could you please split them up a bit to be more clear?

If it helps, each test really only needs to test one aspect/behavior - and each one aspect/behavior doesn't need to be repeatedly tested. For example, a test checking that types (`type Foo = ...`) doesn't need to _also_ check that various number forms (`3n`, `-7.8`, ...) are each validated.

Docs on granular unit tests: https://typescript-eslint.io/contributing/local-development/#granular-unit-tests, with a bit more context in #9210.

packages/eslint-plugin/src/rules/no-magic-numbers.ts
```diff
@@ -129,6 +136,40 @@ export default createRule<Options, MessageIds>({
...
+  if (
+    node.parent.type === AST_NODE_TYPES.UnaryExpression &&
+    ['-', '+'].includes(node.parent.operator)
+  ) {
+    if (node.parent.operator === '-') {
+      return -(node.value ?? BigInt(node.raw));
+    }
+  }
+
+  return node.value ?? BigInt(node.raw);
```
JoshuaKGoldberg, line 170: [Testing] This is reported as only partially covered in the Codecov report: https://app.codecov.io/gh/typescript-eslint/typescript-eslint/pull/9167?src=pr&el=tree&filepath=packages%2Feslint-plugin%2Fsrc%2Frules%2Fno-magic-numbers.ts&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=typescript-eslint#diff-cGFja2FnZXMvZXNsaW50LXBsdWdpbi9zcmMvcnVsZXMvbm8tbWFnaWMtbnVtYmVycy50cw==. Same with the `return -(...);` a few lines earlier.

In fact, removing the two `?? BigInt(node.raw)`s results in no test failures:

```suggestion
  return node.value;
```

...and this function is only ever called on `node`s that have a defined` node.value`, per the `if (typeof ...)` checks in the `Literal(node): void { ... }`. My advice would be to either pass in `node.value` directly _or_ change the type of the `node` parameter to mention that it has a non-`null` `value`.

JoshuaKGoldberg (changes requested):
Implementation looks great, thanks! I like the splitting out of the helpful functions. ✨ 

Requesting changes on the `node.value` type shenanigans and a bit more granular testing please.

packages/eslint-plugin/tests/rules/no-magic-numbers.test.ts
Vinccool96: The tests are mostly to make sure that the numbers ignored are properly caught by the `ignore` option. Should I remove the tests I added in `invalid`?

JoshuaKGoldberg: I think it'd be good to still have some testing to make sure the option doesn't exclude _all_ numbers - just the listed ones.

Vinccool96: I think I will remove the tests I added in `ignore`, except for the first one, that I will split in multiple tests, that will each contain one error. Would that fix the issue?

JoshuaKGoldberg: I think that's a reasonable direction, but I'm not 100% sure what you're considering an error / what'll be deduplicated. Could you push it up?

Vinccool96: Yes, I just pushed it. It's to make sure that the negative catches them correctly. Well, for the invalid cases, to make sure that they're not caught as invalid.

packages/eslint-plugin/tests/rules/no-magic-numbers.test.ts
```diff
@@ -117,6 +117,104 @@ type Foo = {
...
+  | 0x0a
+  | -0xbc
+  | 1e2
+  | -3e4
+  | 5e-6
+  | -7e-8
+  | 1.1e2
+  | -3.1e4
+  | 5.1e-6
+  | -7.1e-8;
```
JoshuaKGoldberg, line 138: [Testing] Continuing https://github.com/typescript-eslint/typescript-eslint/pull/9167#discussion_r1623415921: what I mean is, this test case right now does two things:

* 🍏 Check that a `type`'s union type's constituents are ignored properly
* 🍌 Check that all sorts of numeric literals (bigints, `0x`, `e`, ...) are ignored properly

...so, really, there should be a standalone test for 🍏 union type constituents. The 🍌 literal parsing doesn't need to be in that test.

```suggestion
```

From another perspective: this PR is only meant to fix 🍏 what the `ignore` option looks at. It's not changing 🍌 how the rule parses literals. So there's no need to add 🍌 literal parsing tests as part of this PR.

If you _really want_ to add more testing for 🍌 literal parsing (which is actually a good idea IMO, kind of a gap in testing at the moment!)_ then I'd be in favor of those. They'd be separate tests from the 🍏 type ones.

packages/eslint-plugin/tests/rules/no-magic-numbers.test.ts
```diff
@@ -117,6 +117,104 @@ type Foo = {
...
+      ],
+    },
+    {
+      code: `
+interface Foo {
+  bar: 1;
+}
+      `,
+      options: [{ ignoreNumericLiteralTypes: true, ignore: [1] }],
+    },
```
JoshuaKGoldberg, line 170: [Praise] This is a perfect test 😁 tests one thing and is very clear. I like it!

packages/eslint-plugin/tests/rules/no-magic-numbers.test.ts
JoshuaKGoldberg: Continuing: https://github.com/typescript-eslint/typescript-eslint/pull/9167#discussion_r1623586724

packages/eslint-plugin/tests/rules/no-magic-numbers.test.ts
Vinccool96: So this one too should be split in multiple tests?

JoshuaKGoldberg: Yes please 👍

packages/eslint-plugin/tests/rules/no-magic-numbers.test.ts
```diff
@@ -599,5 +718,229 @@ type Foo = {
...
+        {
+          messageId: 'noMagic',
+          data: {
+            raw: '-7.1e-8',
+          },
+          line: 1,
+          column: 12,
+        },
+      ],
+    },
```
JoshuaKGoldberg, line 944: [Testing] Getting there! This is what I was referring to by _"each test really only needs to test one aspect/behavior - and each one aspect/behavior doesn't need to be repeatedly tested"_. Each of these highlighted tests checks that a particular kind of numeric literal is properly given an error on. But, I think the only kind of "edge cases" that need to be checked here are around negative (`-`) vs. positive, numbers, no? Am I misinterpreting some edge cases that should be explicitly checked?

```suggestion
```

JoshuaKGoldberg (approved):
💯 this is great! Thanks for massaging the tests a bunch!

Tom Hiddleston as Loki saying "Glorious"

Vinccool96:
LET'S GOOOOOOO
image

PR #9201, opened 2024-06-01
docs: improve return-await description about motivation

### LINKED ISSUE #8763: Improve return-await rule description, and better explain stack trace benefits

### Suggested Changes

From https://github.com/typescript-eslint/typescript-eslint/issues/8651 (cc @devuxer):

> What is says it does:
> 
> > Enforce consistent returning of awaited values.
> 
> What it actually does:
> 
> > Enforce consistent awaiting of returned values.
> 
> Wordier but more complete version:
> 
> > Enforce whether a returned promise is always awaited, never awaited, or awaited when inside a try/catch.
> 
> The reason I thought I wanted it to actually return the awaited value is to enable better stack traces, but I acknowledge I don't quite get the exact circumstances under which it's necessary to return the awaited expression to attain that benefit. It would be nice if the documentation made the stack trace behavior for the different options of this rule crystal clear, as it's difficult to really make an intelligent decision about this rule's options without this knowledge.

I think the stack trace of async code is not well understood by many people and the rule options' motivations remain unclear. Here is what I wrote for MDN, for reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#improving_stack_trace

### Affected URL(s)

https://typescript-eslint.io/rules/return-await/

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8763

### DIFF
--- a/packages/eslint-plugin/docs/rules/return-await.mdx
+++ b/packages/eslint-plugin/docs/rules/return-await.mdx
@@ -1,5 +1,5 @@
 ---
-description: 'Enforce consistent returning of awaited values.'
+description: 'Enforce consistent awaiting of returned promises.'
 ---
 
 import Tabs from '@theme/Tabs';
@@ -9,12 +9,14 @@ import TabItem from '@theme/TabItem';
 >
 > See **https://typescript-eslint.io/rules/return-await** for documentation.
 
-Returning an awaited promise can make sense for better stack trace information as well as for consistent error handling (returned promises will not be caught in an async function try/catch).
-
 This rule builds on top of the [`eslint/no-return-await`](https://eslint.org/docs/rules/no-return-await) rule.
 It expands upon the base rule to add support for optionally requiring `return await` in certain cases.
 
-The extended rule is named `return-await` instead of `no-return-await` because the extended rule can enforce the positive or the negative. Additionally, while the core rule is now deprecated, the extended rule is still useful in many contexts.
+The extended rule is named `return-await` instead of `no-return-await` because the extended rule can enforce the positive or the negative. Additionally, while the core rule is now deprecated, the extended rule is still useful in many contexts:
+
+- Returning an awaited promise [improves stack trace information](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#improving_stack_trace).
+- When the `return` statement is in `try...catch`, awaiting the promise also allows the promise's rejection to be caught instead of leaving the error to the caller.
+- Contrary to popular belief, `return await promise;` is [at least as fast as directly returning the promise](https://github.com/tc39/proposal-faster-promise-adoption).
 
 ## Options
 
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -28,7 +28,7 @@ export default createRule({
   name: 'return-await',
   meta: {
     docs: {
-      description: 'Enforce consistent returning of awaited values',
+      description: 'Enforce consistent awaiting of returned promises',
```
kirkwaiblinger, line 31: (praise) _much_ better than the old text!

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -9,12 +9,14 @@ import TabItem from '@theme/TabItem';
 >
 > See **https://typescript-eslint.io/rules/return-await** for documentation.
 
-Returning an awaited promise can make sense for better stack trace information as well as for consistent error handling (returned promises will not be caught in an async function try/catch).
-
 This rule builds on top of the [`eslint/no-return-await`](https://eslint.org/docs/rules/no-return-await) rule.
```
kirkwaiblinger, line 12: i kinda think we should drop this, as well as the comments around the naming. the base rule has been deprecated for some time, and this rule has an independent life of its own now. 

I think, instead, it would be better to relegate the reference to the base rule to a little info box saying (approximaely) "this was originally conceived as an extension rule associated with the eslint rule `no-return-await`, but that's no longer the case. The base rule became deprecated because it was realized that it did more harm than good, due to limitations of not having access to type information. If you're interested in reading on the history of that decision, see \<insert link\>"

What do you think?

Josh-Cena: FWIW, the base rule was deprecated because it was based on the false premise that `return promise;` is faster than `return await promise;`, which may be true for some time but not anymore. `return await promise;` is desirable in all other regards. It's totally possible to build a type-unaware rule that changes `return await something;` to `return something;`, but not in the other direction.

I think we should keep it at least for another ESLint major version, as people are still familiar with this rule.

kirkwaiblinger: > FWIW, the base rule was deprecated because it was based on the false premise that `return promise;` is faster than `return await promise;`

Right, performance was the once-believed-but-now-debunked good.

> `return await promise;` is desirable in all other regards. It's totally possible to build a type-unaware rule that changes `return await something;` to `return something;`, but not in the other direction.

harm, due to limitations of not having access to type information.

***

That's where I get my summary that it was doing more harm than good due to the limitations of not having access to type information 🙂 But no need to use this wording if you feel otherwise!

bradzacher (approved):

JoshuaKGoldberg (approved):
Despite only being a ~dozen added lines, this somehow makes the rule's doc page _much_ more clear & informative. Very nicely done @Josh-Cena! 👏 😍 

Shawshank Redemption scene where Andy is smiling and raising his arms to the rain. Caption: "AND IT'S GLORIOUS"

Issue #4513, opened 2022-02-04
[no-unnecessary-condition] False error on uninitialized variable


Retro64:
- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.
- [x] I have [read the FAQ](https://typescript-eslint.io/docs/linting/troubleshooting) and my problem is not listed.

[Playground](https://typescript-eslint.io/play#ts=5.3.2&fileType=.tsx&code=DYUwLgBAZg9jBcEDOYBOBLAdgcwNwCh8BjGTFZcAMTggF4IAKASkQDcZ0ATOgPggG98ECOiiNYMAPwA6UDjAALCAB4IARiYChw6DXoByCfoLCAvvlMEgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK%2BSpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

```JSON
{
  "rules": {
      "@typescript-eslint/no-unnecessary-condition": [
      "error"
    ],
  }
}
```

```TS
let foo: string;

const setFoo = (): void => {
  if (foo?.length < 1) {
    foo = 'foo';
  }
};

```

tsconfig
```
{
  "compilerOptions": {
    "outDir": "./lib",
    "module": "commonjs",
    "target": "es2015",
    "lib": ["es6", "es7"],
    "sourceMap": true,
    "allowJs": false,
    "moduleResolution": "node",
    "forceConsistentCasingInFileNames": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
    "noImplicitAny": true,
    "strict": true,
    "strictNullChecks": true,
    "suppressImplicitAnyIndexErrors": true,
    "noUnusedLocals": true,
    "noErrorTruncation": true,
    "pretty": true,
    "declaration": true
  },
  "include": [
    "src/**/*"
  ]
}
```

**Expected Result**

Existence check on a possibly uninitialized variable should not trigger no-unnecessary-condition

For sure, the type might be extended to `let foo = string | undefined;` as workaround, but that actually looks a bit uncommon for a let definition, where not initializing might be intended.

[...]

PR #9209, opened 2024-06-02
fix(eslint-plugin): [no-unnecessary-type-assertion] conflict with TS for variables used before assignment

### LINKED ISSUE #9206: Bug: [no-unnecessary-type-assertion] Conflict with TS for variables used before assignment

### Playground Link

https://typescript-eslint.io/play/#ts=4.6.2&fileType=.ts&code=JYMwBAFAsghgLgCwHQCcYDsAmB7AthASgLAG8BYAKDDADcYUwAPMAXjAEYBuSgX0se4VGAQkEBTRgAdsKOKR5A&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0kOdA7RI4MAF8QOoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkmsseAIsphVAHSQA04NkeAIaYA5ujztIAUQDKAOXQAPCbxpR4sACYBXDJIAKmdCSwA3dDz61UyAEbsA2lf6yFyyM4C6q-pvTRUYUECYnYAM0FUEx9aIlgpeGQJSjxMbXQY9S1ddAAlY1hUbRCiSQAhbSJNDDZMyGMAWWyMAEkiPCxYRHZU9LrkEkLg9E0mnQwyFLSM50hI1FgAdxakWEwCIhEAFRIpZXRSQlIe6bqSVORoZLBe9D4AXxB7oA&tokens=false

### Repro Code

```TypeScript
if (Math.random()) {
  var x = 1;
}
x;
x!;
export {}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-unnecessary-type-assertion":"error"
  }
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "rootDir": ".",
    "target": "ESNext",
    "module": "Preserve",
    "lib": [
      "ESNext"
    ],
    "declaration": false,
    "noEmit": true,
    "moduleResolution": "Bundler",
    "esModuleInterop": true,
    "isolatedModules": true,
    "allowImportingTsExtensions": true,
    "strict": true
  }
}
```

### Expected Result

`x!;` should not be reported.

### Actual Result

`x!;` was reported. This conflicts with TS.

### Additional Info

_No response_

liuxingbaoyu:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9206

## Overview

This PR makes the rule ignore var declared in child scopes.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
@@ -96,7 +96,10 @@ export default createRule<Options, MessageIds>({
         if (
           ts.isVariableDeclarationList(declaration.parent) &&
           // var
-          declaration.parent.flags === ts.NodeFlags.None
+          declaration.parent.flags === ts.NodeFlags.None &&
+          // If they are not in the same file it will not exist.
+          // This situation must not occur using before defined.
+          services.tsNodeToESTreeNodeMap.has(declaration)
         ) {
           const declaratorNode: TSESTree.VariableDeclaration =
             services.tsNodeToESTreeNodeMap.get(declaration);
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
@@ -83,12 +83,22 @@ export default createRule<Options, MessageIds>({
         // also ignore function arguments as they can't be used before defined
         ts.isVariableDeclaration(declaration)
       ) {
+        // For var declarations, we need to check whether the node
+        // is actually in a descendant of its declaration or not. If not,
+        // it may be used before defined.
+
+        // eg
+        // if (Math.random() < 0.5) {
+        //     var x: number  = 2;
+        // } else {
+        //     x!.toFixed();
+        // }
         if (
           ts.isVariableDeclarationList(declaration.parent) &&
           // var
           declaration.parent.flags === ts.NodeFlags.None
         ) {
-          const declaratorNode: TSESTree.VariableDeclarator =
+          const declaratorNode: TSESTree.VariableDeclaration =
             services.tsNodeToESTreeNodeMap.get(declaration);
           const scope = context.sourceCode.getScope(node);
           const declaratorScope = context.sourceCode.getScope(declaratorNode);
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
@@ -1,3 +1,4 @@
+import type { Scope } from '@typescript-eslint/scope-manager';
 import type { TSESTree } from '@typescript-eslint/utils';
 import { AST_NODE_TYPES, AST_TOKEN_TYPES } from '@typescript-eslint/utils';
 import * as tsutils from 'ts-api-utils';
@@ -80,33 +81,53 @@ export default createRule<Options, MessageIds>({
         ) &&
         // ignore class properties as they are compile time guarded
         // also ignore function arguments as they can't be used before defined
-        ts.isVariableDeclaration(declaration) &&
-        // is it `const x!: number`
-        declaration.initializer === undefined &&
-        declaration.exclamationToken === undefined &&
-        declaration.type !== undefined
+        ts.isVariableDeclaration(declaration)
       ) {
-        // check if the defined variable type has changed since assignment
-        const declarationType = checker.getTypeFromTypeNode(declaration.type);
-        const type = getConstrainedTypeAtLocation(services, node);
         if (
-          declarationType === type &&
-          // `declare`s are never narrowed, so never skip them
-          !(
-            ts.isVariableDeclarationList(declaration.parent) &&
-            ts.isVariableStatement(declaration.parent.parent) &&
-            tsutils.includesModifier(
-              getModifiers(declaration.parent.parent),
-              ts.SyntaxKind.DeclareKeyword,
-            )
-          )
+          ts.isVariableDeclarationList(declaration.parent) &&
+          // var
+          declaration.parent.flags === ts.NodeFlags.None
         ) {
-          // possibly used before assigned, so just skip it
-          // better to false negative and skip it, than false positive and fix to compile erroring code
-          //
-          // no better way to figure this out right now
-          // https://github.com/Microsoft/TypeScript/issues/31124
-          return true;
+          const declaratorNode: TSESTree.VariableDeclarator =
+            services.tsNodeToESTreeNodeMap.get(declaration);
+          const scope = context.sourceCode.getScope(node);
+          const declaratorScope = context.sourceCode.getScope(declaratorNode);
+          let parentScope: Scope | null = declaratorScope;
+          while ((parentScope = parentScope.upper)) {
+            if (parentScope === scope) {
+              return true;
+            }
+          }
+        }
+
+        if (
+          // is it `const x!: number`
+          declaration.initializer === undefined &&
+          declaration.exclamationToken === undefined &&
+          declaration.type !== undefined
+        ) {
+          // check if the defined variable type has changed since assignment
+          const declarationType = checker.getTypeFromTypeNode(declaration.type);
+          const type = getConstrainedTypeAtLocation(services, node);
+          if (
+            declarationType === type &&
+            // `declare`s are never narrowed, so never skip them
+            !(
+              ts.isVariableDeclarationList(declaration.parent) &&
+              ts.isVariableStatement(declaration.parent.parent) &&
+              tsutils.includesModifier(
+                getModifiers(declaration.parent.parent),
+                ts.SyntaxKind.DeclareKeyword,
+              )
+            )
+          ) {
+            // possibly used before assigned, so just skip it
+            // better to false negative and skip it, than false positive and fix to compile erroring code
+            //
+            // no better way to figure this out right now
+            // https://github.com/Microsoft/TypeScript/issues/31124
+            return true;
+          }
         }
       }
       return false;
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
@@ -1102,5 +1102,25 @@ x;
         },
       ],
     },
+    {
+      code: `
+var x = 1;
+{
+  x!;
+}
+      `,
+      output: `
+var x = 1;
+{
+  x;
+}
+      `,
+      errors: [
+        {
+          messageId: 'unnecessaryAssertion',
+          line: 4,
+        },
+      ],
+    },
   ],
 });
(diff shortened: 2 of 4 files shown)


JoshuaKGoldberg:
> - [ ] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)

This is a good example of the risks around sending a PR before the intended issue goes through triage. There might be a bigger issue that the intended issue is just one case of! 🙂

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
JoshuaKGoldberg, line 1: [Testing] Per #4513, we'll also want tests that exercise `let`.

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
```diff
@@ -346,6 +346,16 @@ const bar = foo.a as string | undefined | bigint;
...
+    {
+      code: `
+if (Math.random()) {
+  {
+    var x = 1;
+  }
+}
+x!;
+      `,
+    },
```
JoshuaKGoldberg, line 358: [Testing] Could you also add an `invalid` case that shows a `var` inside an `if`? Generally when testing rules, if there's a case that's _valid_, it's common to want to have a corresponding _invalid_ case.

JoshuaKGoldberg (changes requested):
Thanks for starting on this! I'm marking this as a draft because the root accepted issue is #4513, which adds on the idea of `let`s. Feel free to un-draft and re-request review once it's ready to address those too. Cheers! ✨ 

_(I haven't looked at the source yet - in case it gets refactored to handle `let`s)_

JoshuaKGoldberg (changes requested):
Sorry for the noise - still requesting changes on a bit more testing. Still holding off on reviewing the source code a bit - will do soon!

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
kirkwaiblinger, line 1: [Testing] Coming over here from https://github.com/typescript-eslint/typescript-eslint/issues/9206#issuecomment-2144541064, please add some test coverage around `let`... I'm not sure whether we really care about defining or testing the behavior of the use-before-define `as T` behavior, though, since it's a TS error anyway, unlike the postfix `!`.

liuxingbaoyu: https://github.com/liuxingbaoyu/typescript-eslint/blob/ea318d9492828f60dc923c9ede31472affa32af6/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts#L125-L126

The tests seem to have it covered. :)

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
liuxingbaoyu: The current implementation ignores any child scoped `var` to avoid false positives, and I haven't figured out a corresponding invalidation test yet.🤔

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
kirkwaiblinger: awesome! 🙂

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
JoshuaKGoldberg: Right, great 😄

JoshuaKGoldberg (dismissed):
This looks great to me, thanks! 🧹 ✨

packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
kirkwaiblinger, line 1: Could you add one more test, where the unnecessary type assertion is in a scope that is in a _child_ of its declaration?

(something like this)
```ts
    {
      code: `
var x = 1;
{
  x!;
}
      `,
      output: `
var x = 1;
{
  x;
}
      `,
      errors: [
        {
          messageId: 'unnecessaryAssertion',
          line: 4,
        },
      ],
    },

```

(reasoning) I was playing around locally and wondered why we wouldn't change the impl to the following, which passed all the existing tests, but realized my change would fail the child scope logic.
```ts
        if (
          ts.isVariableDeclarationList(declaration.parent) &&
          // var
          declaration.parent.flags === ts.NodeFlags.None
        ) {
          const declaratorNode =
            services.tsNodeToESTreeNodeMap.get(declaration);
          const declaratorScope = context.sourceCode.getScope(declaratorNode);
          if (
            scope !== declaratorScope &&
            scope.variableScope === declaratorScope.variableScope
          ) {
            return true;
          } 
        }

```

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -79,29 +80,49 @@ export default createRule<Options, MessageIds>({
...
-        // is it `const x!: number`
-        declaration.initializer === undefined &&
-        declaration.exclamationToken === undefined &&
-        declaration.type !== undefined
+        ts.isVariableDeclaration(declaration)
       ) {
-        // check if the defined variable type has changed since assignment
-        const declarationType = checker.getTypeFromTypeNode(declaration.type);
-        const type = getConstrainedTypeAtLocation(services, node);
         if (
```
kirkwaiblinger, line 85: (nit, optional) i think a comment would be helpful here, since it's pretty confusing territory. Maybe something like this?

```text
For var declarations, we need to check whether the node 
is actually in a descendant of its declaration or not. If not, 
it may be used before defined.

eg 
if (Math.random() < 0.5) {
    var x: number  = 2;
} else {
    x!.toFixed();
}
```

kirkwaiblinger (changes requested):
Just some very minor nitpicks, but this looks great to me! 🙂

kirkwaiblinger:
👋 Hey @liuxingbaoyu ! Just checking in, is this still something you have time for? No worries if not - you got us 98% of the way there, so we can take care of the last few chores and merge it in for you if you like. But if you have a few minutes to take care of the last few chores.... then it saves us a few minutes ❤️

liuxingbaoyu:
@kirkwaiblinger Sorry for forgetting about it. Thanks! I will finish it soon!

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -80,33 +81,63 @@ export default createRule<Options, MessageIds>({
...
+          declaration.parent.flags === ts.NodeFlags.None
         ) {
-          // possibly used before assigned, so just skip it
-          // better to false negative and skip it, than false positive and fix to compile erroring code
-          //
-          // no better way to figure this out right now
-          // https://github.com/Microsoft/TypeScript/issues/31124
-          return true;
+          const declaratorNode: TSESTree.VariableDeclaration =
+            services.tsNodeToESTreeNodeMap.get(declaration);
```
liuxingbaoyu, line 102: I noticed it when rebasing, I'm not sure if this will be affected.🤔
https://github.com/typescript-eslint/typescript-eslint/pull/9660

kirkwaiblinger: @auvred Could you double check this?

kirkwaiblinger: @liuxingbaoyu Thanks for calling this out. Good attention to detail!

kirkwaiblinger (dismissed):
Approving contingent on https://github.com/typescript-eslint/typescript-eslint/pull/9209#discussion_r1701610167 being resolved. 

Great work!

cats celebrating

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
auvred: Sorry for such a long reply! I missed the notification :(

---

I think we can't just rely on `services.tsNodeToESTreeNodeMap.get` when dealing with declarations, as these declarations may be contained in other files that `tsNodeToESTreeNodeMap` doesn't know about.

Imagine the following:
```ts
// var-declaration.ts
var varDeclarationFromFixture = 1;

// file.ts
varDeclarationFromFixture!;
```

The rule throws runtime error on this code.


```
FAIL tests/rules/no-unnecessary-type-assertion.test.ts
  ● Test suite failed to run

    TypeError: Converting circular structure to JSON
        --> starting at object with constructor 'Object'
        |     property 'expression' -> object with constructor 'Object'
        --- property 'parent' closes the circle
        at stringify (<anonymous>)

      at messageParent (../../node_modules/jest-worker/build/workers/messageParent.js:29:19)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        5.074 s
Ran all test suites matching /no-unnecessary-type-assertion/i.
```

---

Since it requires more than one file to reproduce, I can't reproduce it in online playground, so here's the patch:

```patch
diff --git a/packages/eslint-plugin/tests/fixtures/tsconfig.json b/packages/eslint-plugin/tests/fixtures/tsconfig.json
index 6ae5e6473..c16815aaf 100644
--- a/packages/eslint-plugin/tests/fixtures/tsconfig.json
+++ b/packages/eslint-plugin/tests/fixtures/tsconfig.json
@@ -12,6 +12,7 @@
     "file.ts",
     "consistent-type-exports.ts",
     "mixed-enums-decl.ts",
-    "react.tsx"
+    "react.tsx",
+    "var-declaration.ts"
   ]
 }
diff --git a/packages/eslint-plugin/tests/fixtures/var-declaration.ts b/packages/eslint-plugin/tests/fixtures/var-declaration.ts
new file mode 100644
index 000000000..5253648a8
--- /dev/null
+++ b/packages/eslint-plugin/tests/fixtures/var-declaration.ts
@@ -0,0 +1 @@
+var varDeclarationFromFixture = 1;
diff --git a/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts b/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
index afc0b3c3b..b636eec84 100644
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
@@ -1088,6 +1088,20 @@ const bar = foo.a;
     },
     {
       code: `
+varDeclarationFromFixture!;
+      `,
+      output: `
+varDeclarationFromFixture;
+      `,
+      errors: [
+        {
+          messageId: 'unnecessaryAssertion',
+          line: 3,
+        },
+      ],
+    },
+    {
+      code: `
 var x = 1;
 x!;
       `,
```

You can run `git apply` to apply it locally.

auvred (approved):
LGTM! :100:

JoshuaKGoldberg (approved):
🙌

Issue #6434, opened 2023-02-07
AST: provide better discriminated unions for ambiguous node types


bradzacher:
Some of the nodes in the AST have certain constraints to which they must adhere which are discriminated by some property on the node.

As an example - if a `MemberExpression` has `.computed = false`, then the `.property` ***must*** only be `Identifier | PrivateIdentifier`. For this specific case we clearly include this discriminated union in the AST types to make it easier for users to consume the AST without performing unnecessary checks to appease the types:
https://github.com/typescript-eslint/typescript-eslint/blob/a9cb860c37590b01222879f07a4c63e6ce95a37d/packages/ast-spec/src/expression/MemberExpression/spec.ts#L7-L28

We have been pretty conservative (read: lazy 😅) with this work though and there are a lot more cases we could discriminate things further.

Specifically we could further discriminate the following cases:

- `Property`
    - Object literal property vs Object pattern property.
        - Pattern
            - An object pattern property `.value` can only ever be `Identifier | ArrayPattern | ObjectPattern | AssignmentPattern`, but an object literal property can be `Expression`.
            - An object pattern property can only ever be `kind = 'init'`
        - Literal
            - An object literal property cannot have `.value = ObjectPattern | ArrayPattern | AssignmentPattern`
    - `.shorthand = true`
        - `.value` must *always* be `Identifier`
        - Cannot have `.computed = true`
        - Cannot have `.kind = 'get' | 'set'`
        - Cannot have `.method = true`
    - `.kind = 'get' | 'set'`
        - Must have `.value = FunctionExpression | TSEmptyBodyFunctionExpression`
        - Cannot have `.shorthand = true`
    - `.kind = 'init' && .method = true`
        - Must have `.value = FunctionExpression | TSEmptyBodyFunctionExpression`
        - Cannot have `.shorthand = true`
- `BinaryExpression`
    - `in` allows private brand checks like `#priv in expr`.
      All other operators do not allow a lone private identifier in the `.left`
- `MethodDefinition`
    - `kind = 'constructor'` cannot have `.computed = true`, `.static = true`
- `ArrowFunctionExpression`
    - `.expression = true`
        - `.value` cannot be `BlockStatement`
- `YieldExpression`
    - `.delegate = true`
        - `.argument` must be defined
- `TSTypePredicate`
    - `.asserts = false`
        - `.typeAnnotation` must be non-null
- `UnaryExpression`
    - `.prefix` must always be `true`

[...]

PR #9211, opened 2024-06-02
feat(ast-spec): tighter types and documentation for declaration/*

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: part of https://github.com/typescript-eslint/typescript-eslint/issues/6434

## Overview

Although the types are tighter, it contains no breaking tree shape changes.

Also added a bunch of documentation.


packages/ast-spec/src/declaration/ExportAllDeclaration/spec.ts
```diff
@@ -12,24 +12,23 @@ export interface ExportAllDeclaration extends BaseNode {
...
   attributes: ImportAttribute[];
   /**
-   * The name for the exported items. `null` if no name is assigned.
+   * The name for the exported items (`as X`). `null` if no name is assigned.
    */
   exported: Identifier | null;
   /**
    * The kind of the export.
    */
-  // TODO(#1852) - breaking change remove this because it is a semantic error to have it
```
Josh-Cena, line 32: TS now allows `export type * from`

packages/ast-spec/src/declaration/ExportDefaultDeclaration/spec.ts
```diff
@@ -10,7 +9,7 @@ export interface ExportDefaultDeclaration extends BaseNode {
    */
   declaration: DefaultExportDeclarations;
   /**
-   * The kind of the export.
+   * The kind of the export. Always `value` for default exports.
    */
-  exportKind: ExportKind;
+  exportKind: 'value';
```
Josh-Cena, line 14: Tightened: `export type default` is not legal

packages/ast-spec/src/declaration/ExportNamedDeclaration/spec.ts
```diff
@@ -64,9 +71,14 @@ export interface ExportNamedDeclarationWithoutSourceWithMultiple
   attributes: ImportAttribute[];
   declaration: null;
   source: null;
-  specifiers: ExportSpecifier[];
```
Josh-Cena, line 67: Removals here are not breaking because these properties are already declared on the base

packages/ast-spec/src/declaration/ImportDeclaration/spec.ts
```diff
@@ -41,5 +41,6 @@ export interface ImportDeclaration extends BaseNode {
    * import 'mod';
    * ```
    */
+  // TODO: we should have a way to differentiate between these two cases
   specifiers: ImportClause[];
```
Josh-Cena, line 45: For the future: `specifiers` should be able to be `null`, like how `typeParameters` can be both empty or `null`

packages/ast-spec/src/declaration/TSDeclareFunction/spec.ts
```diff
@@ -1,11 +1,35 @@
...
 
-// TODO(#1852) - async + declare are semantically invalid together
-export interface TSDeclareFunction extends FunctionBase {
+interface TSDeclareFunctionBase extends FunctionBase {
   type: AST_NODE_TYPES.TSDeclareFunction;
-  body: BlockStatement | undefined;
+  /**
+   * TS1183: An implementation cannot be declared in ambient contexts.
+   */
+  body: undefined;
```
Josh-Cena, line 9: Tightened: I think this is safe?

packages/ast-spec/src/declaration/TSDeclareFunction/spec.ts
```diff
@@ -1,11 +1,35 @@
...
+export interface TSDeclareFunctionWithDeclare extends TSDeclareFunctionBase {
+  /**
+   * TS1040: 'async' modifier cannot be used in an ambient context.
+   */
+  async: false;
+  declare: true;
+  /**
+   * TS1221: Generators are not allowed in an ambient context.
+   */
+  generator: false;
```
Josh-Cena, line 26: Tightened

packages/ast-spec/src/declaration/TSImportEqualsDeclaration/spec.ts
```diff
@@ -19,7 +19,45 @@ export interface TSImportEqualsDeclaration extends BaseNode {
    * import F3 = require('mod');
    * ```
    */
-  moduleReference: EntityName | TSExternalModuleReference;
-  // TODO(#1852) - breaking change remove this as it is invalid
+  moduleReference: Identifier | TSQualifiedName | TSExternalModuleReference;
```
Josh-Cena, line 22: Slight tightening: `EntityName` is a moving target and doesn't really make sense here. I will continue to fix uses of it. It broke when `ThisExpression` became a valid `EntityName`.

packages/ast-spec/src/declaration/TSImportEqualsDeclaration/spec.ts
```diff
@@ -19,7 +19,45 @@ export interface TSImportEqualsDeclaration extends BaseNode {
...
+  moduleReference: TSExternalModuleReference;
+  /**
+   * The kind of the import.
+   */
+  importKind: ImportKind;
+}
+
+export type TSImportEqualsDeclaration =
+  | TSImportEqualsNamespaceDeclaration
+  | TSImportEqualsRequireDeclaration;
```
Josh-Cena, line 63: Changed to union

packages/ast-spec/src/declaration/TSModuleDeclaration/spec.ts
```diff
@@ -50,64 +48,81 @@ interface TSModuleDeclarationBase extends BaseNode {
...
+ * ```
+ */
 export interface TSModuleDeclarationModuleWithIdentifierId
   extends TSModuleDeclarationModuleBase {
   kind: 'module';
   id: Identifier;
-  // modules with an Identifier must always have a body
+  // TODO: we emit the wrong AST for `module A.B {}`
+  // https://github.com/typescript-eslint/typescript-eslint/pull/6272 only fixed namespaces
+  // Maybe not worth fixing since it's legacy
```
Josh-Cena, line 116: Worth pointing out

packages/ast-spec/src/declaration/VariableDeclaration/spec.ts
```diff
@@ -36,18 +40,65 @@ export interface LetOrConstOrVarDeclaration extends BaseNode {
...
+   * In a `const` declaration, the declarators must have initializers.
+   */
+  declarations: ConstDeclaratorWithInit[];
+}
+
+export type LetOrConstOrVarDeclaration =
+  | LetOrVarDeclaredDeclaration
+  | LetOrVarNonDeclaredDeclaration
+  | ConstDeclaredDeclaration
+  | ConstNonDeclaredDeclaration;
```
Josh-Cena, line 98: The biggest change: made `LetOrConstOrVarDeclaration` much stricter.

Josh-Cena (commented):
Some comments

packages/ast-spec/src/declaration/ImportDeclaration/spec.ts
bradzacher: This won't ever happen, sadly. The ESTree spec is pretty well locked in once defined and so we're stuck with the decisions of the past.

If we change something like this then we're liable to break all sorts of rules that assume it's always defined.

In general we can only ever change our AST extension shapes and the "core" AST is fixed.

packages/ast-spec/src/declaration/TSDeclareFunction/spec.ts
bradzacher: This is not correct to do.
The error you cited is a **semantic** error, not a **syntax** error.
This means that TS produces an AST for the error and us we do too.

[playground](https://typescript-eslint.io/play/#ts=5.4.5&showAST=es&fileType=.tsx&code=FAEwpgxgNghgTmABAMwK4DsIBcCWB7dFPPACgEpEBvYRRCAgZy0QA9EBeRARgG5gBfYEA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

packages/ast-spec/src/declaration/TSDeclareFunction/spec.ts
bradzacher: Ditto above - the errors are semantic and TS still produces an AST - AND it includes these flags correctly.
So we can't refine these.

packages/ast-spec/src/declaration/VariableDeclaration/spec.ts
bradzacher: The thing that needs to be checked with these changes is like `const x;` - do we have a syntax error or do we produce an AST.

If it's the latter then we can't refine the types as the types have to reflect the AST we output.

For the specific example of `const x:` - we produce an AST so the types must allow the case of `const` with an `init: null`.

[playground](https://typescript-eslint.io/play/#ts=5.4.5&showAST=es&fileType=.tsx&code=MYewdgzgLgBAHgbgFBA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

packages/ast-spec/src/declaration/ImportDeclaration/spec.ts
Josh-Cena: Hhhh, I forgot this is not TS specific. Fine then.

packages/ast-spec/src/declaration/TSDeclareFunction/spec.ts
Josh-Cena: Aren't we supposed to throw on semantic errors? Wasn't fisker fixing some of these before?

bradzacher (changes requested):
A lot of this is great!
But there's a chunk that I think causes drift between the types and the ASTs we produce.

The important distinction when looking at TS errors is "does TS crash, or does it produce an AST"? In most cases it's the latter.
TS's parser is very forgiving and most errors are semantic errors, not syntactic errors - meaning it produces an AST even if the code has errors.

If TS produces an AST then in most cases we do too. So we need to ensure that the types respect that.

We can always add hard errors to our AST translation step to enforce invariants so that we can change the types! But doing so will be a breaking change for the v8 branch.

Josh-Cena:
> We can always add hard errors to our AST translation step to enforce invariants so that we can change the types! But doing so will be a breaking change for the v8 branch.

Really, even when the AST is nonsense and no program is not going to work? For example https://github.com/typescript-eslint/typescript-eslint/pull/7202 was not breaking

bradzacher:
It generally would require a quick chat with the prettier folk. We want to make sure that we're not breaking their experience by throwing on parsable code.

In the past a lot of the invariants we've encoded were just the invariants that prettier itself encoded (allowing them to delete the checks and make assumptions).

I'm all for a stricter AST transform - just don't want to kill downstream usecases with it.

Josh-Cena:
#9219, #9220, #9221

Josh-Cena:
@bradzacher I am splitting normative changes out into separate PRs

bradzacher (dismissed):

bradzacher:
a few lint errors @Josh-Cena otherwise ready to ship

bradzacher (approved):

PR #9225, opened 2024-06-03
feat(typescript-estree): forbid `.body`, `.async`, `.generator` on `declare function`

### LINKED ISSUE #9219: AST tightening: `TSDeclareFunction` has no body, `declare function` cannot have `async` or `*`
### Suggestion

In https://github.com/typescript-eslint/typescript-eslint/pull/9211: we are planning to enforce the following invariants in our AST for `TSDeclareFunction`:

- `body` must be undefined, so the following is a parsing error:
  ```ts
  declare function foo() {}
  ```
  (I don't think it's possible to have `declare: false` and a body anyway, since that would be a normal function declaration)
- If `declare: true`, then `async` and `generator` are both `false`:
  ```ts
  declare async function foo();
  declare function *foo();
  ```

@sosukesuzuki @fisker Do these look safe to you?

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9219


packages/ast-spec/src/declaration/TSDeclareFunction/spec.ts
```diff
@@ -1,11 +1,35 @@
...
+   */
+  async: false;
+  declare: true;
+  /**
+   * TS1221: Generators are not allowed in an ambient context.
+   */
+  generator: false;
+}
+
+export interface TSDeclareFunctionNoDeclare extends TSDeclareFunctionBase {
```
bradzacher, line 29: A comment to explain where this case is used would be good. Took me a minute to figure out where you could have this case.

```suggestion
/**
 * This case is used for overload declarations:
 * ```
 *    function foo(): void;
 * // ^^^^^^^^^^^^^^^^^^^^ TSDeclareFunction
 *    function foo() { }
 * ```
 */
export interface TSDeclareFunctionNoDeclare extends TSDeclareFunctionBase {
```

packages/ast-spec/src/declaration/TSDeclareFunction/spec.ts
```diff
@@ -1,11 +1,35 @@
...
+  declare: false;
+}
```
bradzacher, line 31:
```ts
function* foo(): Generator<never, void, unknown>;
//      ^ TS1222: An overload signature cannot be declared as a generator.
function* foo(): Generator<never, void, unknown> { }
```
[playground](https://typescript-eslint.io/play/#ts=5.4.5&fileType=.ts&code=GYVwdgxgLglg9mAVAAmHOAKAlALmQcQFMxCAnAQyjlIB4SA3MgGmXrhgBMXwBrMOAO5gAfAG4AUKEiwEKNJlwFiZStTqFGpFm07cwfQSOQBvZAF8gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

Looks like we can exclude and crash on this case as well.

bradzacher (changes requested):

packages/ast-spec/src/declaration/TSDeclareFunction/spec.ts
Josh-Cena: Huh, TIL.

bradzacher (approved):

PR #9186, opened 2024-05-30
feat(eslint-plugin): [no-floating-promises] add option 'allowForKnownSafePromises'

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7008
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

This PR is a continuation of https://github.com/typescript-eslint/typescript-eslint/pull/8502, with a few finishing touches.

A big thank you to @arka1002, who is the primary author of this work! ❤️ 

Co-authored-by: @arka1002

Issue #6861, opened 2023-04-07
Enhancement: Detect declared modules for TypeOrValueSpecifier


marekdedic:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#supported-rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Relevant Package

type-utils

### My proposal is suitable for this project

- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Description

The `PackageSpecifier` check currently checks the declaration file path for the string `node_modules` and the name of the package: 

https://github.com/typescript-eslint/typescript-eslint/blob/v6/packages/type-utils/src/TypeOrValueSpecifier.ts#L173-L179

I propose to use TypeScript's `sourceFileToPackageName` map defined [here](https://github.com/microsoft/TypeScript/blob/85f20246a4a2b424f7d92b463c20886ddf7488f9/src/compiler/types.ts#L4796).

Only problem is that it is currently marked as internal - but maybe we could get TS maintainers to open it up?

### Fail

```typescript
-
```

### Pass

```typescript
-
```

JoshuaKGoldberg:
> maybe we could get TS maintainers to open it up?

That would certainly be nice. Could you file an issue on them? Until they provide guidance, I'm very nervous taking on _even more_ dependencies on internal TypeScript APIs.

marekdedic:
Great, I was waiting for some kind of go-ahead, opened the issue now :)

JoshuaKGoldberg:
This is actually important to us now, as it blocks `node:test` from being matched to `node_modules/@types/node/test.d.ts`.

JoshuaKGoldberg:
As of #6870, we actually do use `sourceFileToPackageName` now. So renaming & retargeting this issue to support separately declared modules like `node:test` properly.

PR #9234, opened 2024-06-03
feat(eslint-plugin): [no-floating-promises] add 'allowForKnownSafeCalls' option

### LINKED ISSUE #8404: Enhancement: [no-floating-promises] add an 'allowForKnownSafeCalls' option

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8404

## Overview

Uses the same format and plumbing as the existing `allowForKnownSafePromises` (#7008 -> #9186). The actual changes are pretty straightforward as a result.

Note that this doesn't yet add `node:test` support, as that's blocked by #6861. I plan on sending a followup PR after this one goes in to fix that - using `Program.sourceFileToPackageName` if we really need.

Sending to the `v8` branch instead of `main` because there are some project service fixes in there that this relies on. Example errors from targeting the `main` branch: https://github.com/typescript-eslint/typescript-eslint/actions/runs/9355798709/job/25751928593?pr=9234#logs. Update: or maybe those are also in `v8`. More investigation required.

💖


packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -176,6 +176,44 @@ returnsSafePromise();
...
+```ts option='{"allowForKnownSafePromiseReturns":[{"from":"file","name":"safe","path":"input.ts"}]}'
+declare function unsafe(...args: unknown[]): Promise<void>;
+
+unsafe('...', () => {});
+```
+
+</TabItem>
+<TabItem value="✅ Correct">
+
+```ts option='{"allowForKnownSafePromiseReturns":[{"from":"file","name":"safe","path":"input.ts"}]}' skipValidation
```
JoshuaKGoldberg, line 208: > `skipValidation`

This was a fun one to discover:

* The playground uses `input.ts`
* `docs.test.ts` uses `file.ts`

I'm stepping out, but separate from this PR we should fix the two to be in sync.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
JoshuaKGoldberg, line 1: Investigation notes for the failing test when the project service is enabled (`TYPESCRIPT_ESLINT_PROJECT_SERVICE=true yarn test no-floating-promises`), in `typeDeclaredInFile`:

| Value                              | `project`                                                          | `projectService`                                    |
| ---------------------------------- | ------------------------------------------------------------------ | --------------------------------------------------- |
| `program` `.getCurrentDirectory()` | `---/packages/eslint-plugin`                                       | `---/packages/eslint-plugin/tests/fixtures`         |
| `relativePath`                     | `tests/fixtures/file.ts`                                           | `tests/fixtures/file.ts`                            |
| `absolutePath`                     | `---/packages/eslint-plugin/tests/fixtures/tests/fixtures/file.ts` | `---/packages/eslint-plugin/tests/fixtures/file.ts` |

Discrepency in the current directory...

JoshuaKGoldberg: Interestingly, looking at `useProgramFromProjectService.ts`, `service.currentDirectory` is `---/packages/eslint-plugin`. So the issue is that the project service is picking up the `tests/fixtures/tsconfig.json` file and using it for the test. Cool!

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
JoshuaKGoldberg: #9269 -> #9271

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -176,6 +176,44 @@ returnsSafePromise();
...
+declare function safe(...args: unknown[]): Promise<void>;
+
+safe('...', () => {});
```
kirkwaiblinger, line 211: i like how this a sneaky `it('is a test', () => {})`

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -176,6 +176,44 @@ returnsSafePromise();
 </TabItem>
 </Tabs>
 
+### `allowForKnownSafePromiseReturns`
```
kirkwaiblinger, line 179: If we are at the stage where naming bikeshedding is still ok, I would say that I think this is a tricky name, and difficult to differentiate visually from `allowForKnownSafePromises`. 

`allowForKnownSafeAsyncFunctions` (technically imprecise) or `allowForKnownSafePromiseFunctions` (more precise, and reuses terminology from https://typescript-eslint.io/rules/promise-function-async/) feel better to me.

If too late to think about minutiae like naming, ignore this comment 🙂

kirkwaiblinger (commented):
couple questions, no significant concerns.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
JoshuaKGoldberg: Not too late at all. I was actually thinking about this too... `allowForKnownSafeCalls` is shorter and more distinct from the others, how about that?

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -118,6 +125,10 @@ export default createRule<Options, MessageId>({
           expression = expression.expression;
         }
 
+        if (isKnownSafePromiseReturn(expression)) {
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Good call. Going with the same attitude we've had before: looking purely at the returned type, not what happened to be called.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
kirkwaiblinger: works for me!

packages/eslint-plugin/src/rules/no-floating-promises.ts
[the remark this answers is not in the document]
kirkwaiblinger: I'm confused because I still don't see a test case with a finally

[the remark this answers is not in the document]
kirkwaiblinger: Oh, I see, you added a `.then(() => {})` which I agree with, and is a good test case to have. The `x.finally(() => {})` case is specifically in question, though, since it triggers the rule to recurse and check `x`.

kirkwaiblinger (changes requested):
just requesting a second look at https://github.com/typescript-eslint/typescript-eslint/pull/9234/files#r1626760319 before it's good to go 🙂

packages/eslint-plugin/src/rules/no-floating-promises.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Right! Added it to ensure the rule still reports, as the `.finally()` gives back a `Promise` rather than the expected known safe type.

PR #9238, opened 2024-06-03
docs: auto-generate support h2 below all blog posts

### LINKED ISSUE #8116: Website: Automatically add "## Supporting typescript-eslint" at the end of blog posts
### Suggestion

I've been copy & pasting this heading + paragraph at the bottom of blog posts:

https://github.com/typescript-eslint/typescript-eslint/blob/04dea84e8e934a415ec1381a90de3cde670d0dc3/packages/website/blog/2023-09-18-parser-options-project-true.md?plain=1#L148-L152

This should be automated, I think.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8116

## Overview

Creates a new internal Docusaurus plugin the same _"Supporting typescript-eslint"_ h2 at the foot of all blog posts that already exists.

~The size of the auto-generated h2 is a bit smaller than the previous manual ones... which I kind of like!~ It was a bug, thanks @Josh-Cena! 

Splits up the existing `generated-rule-docs/utils.ts` into `utils/nodes.ts` and `utils/rules.ts`. The new `blog-footer` plugin only uses `utils/nodes.ts`.

💖


packages/website/docusaurus.config.mts
```diff
@@ -19,6 +20,7 @@ const githubUrl = 'https://github.com/typescript-eslint/typescript-eslint';
 const presetClassicOptions: PresetClassicOptions = {
   blog: {
     blogSidebarCount: 'ALL',
+    beforeDefaultRemarkPlugins: [blogFooter],
```
Josh-Cena, line 23:
```suggestion
    // Allow Docusaurus TOC remark plugin to pick up the injected H2
    beforeDefaultRemarkPlugins: [blogFooter],
```

packages/website/plugins/blog-footer.ts
```diff
@@ -0,0 +1,47 @@
...
+          ],
+          url: 'https://opencollective.com/typescript-eslint',
+          type: 'link',
+        },
+        {
+          type: 'text',
+          value: `. We're a small volunteer team and could use your support to make the ESLint experience on TypeScript great. Thanks! 💖`,
+        },
+      ],
+      type: 'paragraph',
```
Josh-Cena, line 44: FYI: the new title is smaller because you are wrapping it in one extra paragraph here. If you remove this paragraph wrapper the H2 has the same size. Not sure if you prefer that instead.

packages/website/plugins/blog-footer.ts
```diff
@@ -0,0 +1,47 @@
...
+              value: 'Supporting typescript-eslint',
+            },
+          ],
+          depth: 2,
+          type: 'heading',
+        },
+        {
+          type: 'text',
+          value:
+            'If you enjoyed this blog post and/or use typescript-eslint, please consider',
```
Josh-Cena, line 27:
```suggestion
            'If you enjoyed this blog post and/or use typescript-eslint, please consider ',
```

Josh-Cena (changes requested):

packages/website/plugins/blog-footer.ts
JoshuaKGoldberg: Heh, I do prefer the semantics of not having the heading in the paragraph. Thanks!

packages/website/plugins/blog-footer.ts
```diff
@@ -0,0 +1,52 @@
+import type * as mdast from 'mdast';
+import type { Plugin } from 'unified';
+
+import { nodeIsParent } from './utils/nodes';
+
+export const blogFooter: Plugin = () => {
+  return (root, file) => {
+    if (
+      !nodeIsParent(root) ||
+      !(file.value as string).includes('<!--truncate-->')
```
JoshuaKGoldberg, line 10: Agh, noticed this morning - this plugin is also active in the root blog listing (https://deploy-preview-9238--typescript-eslint.netlify.app/blog):

Screenshot of the root blog list, showing the 'Supporting typescript-eslint' h2 at the bottom of the first blog preview

@Josh-Cena is there a better way to stop that? I couldn't find anything searching around.

Josh-Cena: I don't think a better way exists, unfortunately. For context, this is the way it works:

- When Docusaurus renders the overview page, it uses the same syntax to import Markdown, but adds an extra query: `require("some-post.mdx?truncated=true")`
- The blog plugin creates two Webpack loaders, one that reads the resource query, and truncates the source if the `truncated` query exists; one that actually does the MDX transformation, etc.
- Therefore, the MDX loader that runs the plugins doesn't know the protocol of the `truncated` query at all.

First-class truncation is in https://github.com/facebook/docusaurus/issues/6480. We could make plugins know that a file is truncated by giving it the truncation query, for example.

PR #9254, opened 2024-06-04
fix(eslint-plugin): [no-unsafe-return] Differentiate a types-error any from a "true"

developer-bandi:
## PR Checklist

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unsafe-return.ts
+++ b/packages/eslint-plugin/src/rules/no-unsafe-return.ts
@@ -28,7 +28,7 @@ export default createRule({
       requiresTypeChecking: true,
     },
     messages: {
-      unsafeReturn: 'Unsafe return of an `{{type}}` typed value.',
+      unsafeReturn: 'Unsafe return of an {{type}} typed value.',
       unsafeReturnThis: [
         'Unsafe return of an `{{type}}` typed value. `this` is typed as `any`.',
         'You can try to fix this by turning on the `noImplicitThis` compiler option, or adding a `this` parameter to the function.',
@@ -137,6 +137,7 @@ export default createRule({
         }
 
         let messageId: 'unsafeReturn' | 'unsafeReturnThis' = 'unsafeReturn';
+        const isErrorType = tsutils.isIntrinsicErrorType(returnNodeType)
 
         if (!isNoImplicitThis) {
           // `return this`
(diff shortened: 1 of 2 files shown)


developer-bandi:
ci test is fail, but i don't know reason. may be yarn install is fail, but give me a advice to resolve this error?

JoshuaKGoldberg:
@developer-bandi looks it might have been a random transient issue. My advice for that is to first try updating the branch from the latest `main`, or if it's already up-to-date, pushing a small commit. I'm updating from `main` now`.

Just a heads up, it looks like you've got formatting issues in the code that would fail the build. So I'll mark this as `awaiting response` preemptively.

packages/eslint-plugin/tests/rules/no-unsafe-return.test.ts
JoshuaKGoldberg, line 1: [Testing] As with #8304, please include the `data` in existing tests. If there's a bug in the new `isErrorType ? ... : ...` logic, the tests should be able to catch it.

JoshuaKGoldberg (changes requested):
🚀

packages/eslint-plugin/src/rules/no-unsafe-return.ts
```diff
@@ -156,7 +157,11 @@ export default createRule({
           node: reportingNode,
           messageId,
           data: {
-            type: anyType === AnyType.Any ? 'any' : 'any[]',
+            type: isErrorType
+              ? 'error'
```
JoshuaKGoldberg, line 161: #8304 and the other PRs have been going with:

```suggestion
              ? 'error typed'
```

For consistency, let's stick with the same as others.

developer-bandi: @JoshuaKGoldberg It looks good. I just have one question: 
Because the existing error message is `'Unsafe return of an {{type}} typed value.`,if type is `error`, the `error typed` phrase is included. 

Would it be a good idea to edit the error message and change the type to what you suggested?

JoshuaKGoldberg: Oh, I might have misread the code here. Whatever gets it saying _"error typed value"_ in the report.

developer-bandi: thank you Since I think this conversation has been resolved without any modifications, I will process the conversation as resolved.

JoshuaKGoldberg (approved):
The word "Stellar" with a sparkling light traveling through the 'S', over a starfield

PR #9257, opened 2024-06-04
docs: mention non-semver guarantees in Shared Configs docs

### LINKED ISSUE #9073: Docs: Better Callouts for `all` / `strict` / `strict-type-checked` Config Stability

### Suggested Changes

Currently on the Shared Configs page, there is a small blurb near the top of the page that reads:

> With the exception of `all`, `strict`, and `strict-type-checked`, all configurations are considered "stable". Rule additions and removals are treated as breaking changes and will only be done in major version bumps.

However, when a user visits a rule (e.g. [no-dynamic-delete](https://typescript-eslint.io/rules/no-dynamic-delete) ), and clicks on the [link to the config and how to add it to their ESLint config](https://typescript-eslint.io/users/configs/#strict), they are taken past this blurb, and may never see it.

The CAUTIONs under the `strict` and `strict-type-checked` configs, as well as the WARNING under the `all` config, could be updated, or a new admonishment of some kind (ie a CAUTION or WARNING) could be added to each one, to say essentially the inverse of the blurb at the top, with text such as:

> This configuration is not considered "stable" under Semantic Versioning (SemVer); its enabled rules and/or their settings may change even outside of major version updates.

This would better call out their not being considered "stable," especially being under an admonishment instead of a blockquote blurb, and make the wording visible as soon as a user arrives at the config from a rule, instead of relying on their scrolling to the top of the page to get the information. Optionally, the blurb could be removed as well.

### Affected URL(s)

- Shared Configs: https://typescript-eslint.io/users/configs

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9073

## Overview

Adds a `:::note` very similar to @bachmacintosh's phrasing to each of the configs that may change in minor or patch versions.

💖 

Co-authored-by: @bachmacintosh

### DIFF
--- a/docs/users/Shared_Configurations.mdx
+++ b/docs/users/Shared_Configurations.mdx
@@ -184,11 +184,11 @@ module.exports = {
 Some rules also enabled in `recommended` default to more strict settings in this configuration.
 See [`configs/strict.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/strict.ts) for the exact contents of this config.
 
-:::caution
+:::tip
 We recommend a TypeScript project extend from `plugin:@typescript-eslint/strict` only if a nontrivial percentage of its developers are highly proficient in TypeScript.
 :::
 
-:::note
+:::warning
 This configuration is not considered "stable" under Semantic Versioning (semver).
 Its enabled rules and/or their options may change outside of major version updates.
 :::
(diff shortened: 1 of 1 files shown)


docs/users/Shared_Configurations.mdx
```diff
@@ -188,6 +188,11 @@ See [`configs/strict.ts`](https://github.com/typescript-eslint/typescript-eslint
 We recommend a TypeScript project extend from `plugin:@typescript-eslint/strict` only if a nontrivial percentage of its developers are highly proficient in TypeScript.
 :::
 
+:::note
```
Josh-Cena, line 191:
```suggestion
:::warning
```
Same for below

JoshuaKGoldberg: I'd rather not make this stand out quite so much. The warning color is ... a lot. And we already mention this at the top of the file. @Josh-Cena how strongly do you want this? 😄

Josh-Cena: I think this basically means "if you use it it will break", so yes it's quite serious. I think yellow is "just right" for this.

Josh-Cena (approved):

PR #9261, opened 2024-06-04
chore: enable eslint-plugin/no-property-in-node internally

### LINKED ISSUE #8226: Repo: enable eslint-plugin/no-property-in-node
### Suggestion

Sometimes folks will use an `'in'` property check in a rule to narrow the type of a node:

```ts
declare const node: ts.BinaryExpression | ts.ConditionalExpression;

if ('condition' in node) {
  node
  // ^? const node: ts.ConditionalExpression
}
```

I've been advising to prefer using less "dynamic" logical gates: existing type predicate functions, direct `.kind`/`.type` checks, and/or having different functions for different node selectors.

```ts
declare const node: ts.BinaryExpression | ts.ConditionalExpression;

// or:
// if (node.kind === ts.SyntaxKind.ConditionalExpression) {
if (ts.isConditionalExpression(condition)) {
  node
  // ^? const node: ts.ConditionalExpression
}
```

My reasoning has been that "dynamic" logical gates, beyond being a potential deoptimization in JS perf, are susceptible to the types changing unexpectedly. E.g. a new version of TypeScript might add a new node type to a built-in union that adds a case previously meant to be caught by the rule.

My understanding of `in` checks is that they're generally reserved for cases where the more "static" ways of checking aren't available.

Shall we encode this as a lint rule?

Similarly positioned to #4796: if https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/416 is accepted, we can instead enable it in `eslint-plugin-eslint-plugin`.

Edit: https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/326, actually 😄

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8226

## Overview

Adds in a bugfix patch for https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/482. 

💖

### DIFF
--- a/packages/eslint-plugin/src/rules/default-param-last.ts
+++ b/packages/eslint-plugin/src/rules/default-param-last.ts
@@ -24,7 +24,14 @@ export default createRule({
      * @private
      */
     function isOptionalParam(node: TSESTree.Parameter): boolean {
-      return 'optional' in node && node.optional;
+      return (
+        (node.type === AST_NODE_TYPES.ArrayPattern ||
+          node.type === AST_NODE_TYPES.AssignmentPattern ||
+          node.type === AST_NODE_TYPES.Identifier ||
+          node.type === AST_NODE_TYPES.ObjectPattern ||
+          node.type === AST_NODE_TYPES.RestElement) &&
+        node.optional
+      );
     }
 
     /**
--- a/.yarn/patches/eslint-plugin-eslint-plugin-npm-5.5.1-4206c2506d.patch
+++ b/.yarn/patches/eslint-plugin-eslint-plugin-npm-5.5.1-4206c2506d.patch
@@ -1,0 +1,12 @@
+diff --git a/lib/rules/no-property-in-node.js b/lib/rules/no-property-in-node.js
+index f3e2e0c829685213c7807b34158237f59142173c..11e54e8f6d69e0f06b27bb83ed96a96690b5ed9a 100644
+--- a/lib/rules/no-property-in-node.js
++++ b/lib/rules/no-property-in-node.js
+@@ -3,6 +3,7 @@
+ const typedNodeSourceFileTesters = [
+   /@types[/\\]estree[/\\]index\.d\.ts/,
+   /@typescript-eslint[/\\]types[/\\]dist[/\\]generated[/\\]ast-spec\.d\.ts/,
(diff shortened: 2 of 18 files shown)


packages/eslint-plugin/src/rules/default-param-last.ts
```diff
@@ -24,7 +24,14 @@ export default createRule({
      * @private
      */
     function isOptionalParam(node: TSESTree.Parameter): boolean {
-      return 'optional' in node && node.optional;
```
kirkwaiblinger, line 27: this type of code seems fine since it's not being used for narrowing or type operations. It's just asking... "is this parameter optional". it's basically the same as 
```
@ts-expect-error: optional might not be a key, but that's ok, since we're treating it as false.
return Boolean(node.optional);
```

The trouble with enumerating all the subtypes that can possibly be optional is.... what if you forgot one? And they're unnecessary runtime checks.

To mitigate the exhaustiveness concern, though, you can also do 
```ts
switch (node.type) {
    case AST_NODE_TYPES.ArrayPattern:
    case AST_NODE_TYPES.AssignmentPattern:
    // etc
        return node.optional;
    case // explicitly enumerate ones that _should be excluded_:
    default: 
        const __never_case: never = node.type;
        throw new Error('unexpected type');    
}
```
so you get a compiletime and runtime error if you forget to make the switch exhaustive....

kirkwaiblinger (approved):

packages/eslint-plugin/src/rules/default-param-last.ts
kirkwaiblinger: I mean, sure, someone _could_ come along and add a type to the union where `optional` really means _not optional_, or it has shape `{ isOptional: false }`, so it's a truthy object.... but I think it would be reasonable to assume that any `TSESTree.Parameter` would only ever have a sensibly defined `.optional` if it has a `.optional` at all.

kirkwaiblinger: Anyways, the change is fine. I'm just rambling that I don't necessarily think that the change is objectively better _for this specific kind of case_. 🙂

JoshuaKGoldberg: Yeah I think the rule is technically speaking a _little_ overly broad... but the things it saves us from are obnoxious enough that I don't mind a little extra flagging. 🤷

Issue #8517, opened 2024-02-20
Blog: Add post about our suite of Promise rules


JoshuaKGoldberg:
## Overview

One of the most prominently requested, praised, and criticized (sometimes validly) set of lint rules we provide is the Promise/Thenable-related ones:

* [@typescript-eslint/await-thenable](https://typescript-eslint.io/rules/await-thenable)
* [@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises)
* [@typescript-eslint/no-misused-promises](https://typescript-eslint.io/rules/no-misused-promises)
* [@typescript-eslint/prefer-promise-reject-errors](https://typescript-eslint.io/rules/prefer-promise-reject-errors)

Having a nice comprehensive blog post explaining all of them together will be a good way to help folks understand them. Following or alongside #8088, I'd like to write that post explaining:

* What those rules generally do
* Why they're so great & lovely
* Why & how of #7008 & #8404

This'll also be useful once #7008 and #8404 are in, as I expect the new options will be talked about on social media.

JoshuaKGoldberg:
As with #8088, previously blocked on #9816, and still blocked on #8404.

PR #9263, opened 2024-06-04
feat(eslint-plugin): [no-floating-promises] add checkThenables option

### LINKED ISSUE #8433: Enhancement: [no-floating-promises] Don't check coincidentally Promise-like (Thenable) type values by default

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8433

## Overview

Adds a new `checkThenables` `boolean` option to the rule that enables checking all shapes for matching the `.then()` interface. It essentially gates the existing _"does this have a `.then()` with two parameters?"_ logic behind the option.

`checkThenables` is `true` by default in this PR so as to not cause a breaking change. If this PR goes in roughly as-is, I'd like to file a followup issue for v8 to turn it to `false` by default. 

#8433 also mentioned adding in an option for adding additional types to what's flagged by default. I haven't heard anybody mention this as a strong need, so for now I'm holding off on adding that.

💖

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-floating-promises.mdx
+++ b/packages/eslint-plugin/docs/rules/no-floating-promises.mdx
@@ -89,7 +89,7 @@ await Promise.all([1, 2, 3].map(async x => x + 1));
 A ["Thenable"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables) value is an object which has a `then` method, such as a `Promise`.
 Other Thenables include TypeScript's built-in `PromiseLike` interface and any custom object that happens to have a `.then()`.
 
-The `checkAllThenable` option triggers `no-floating-promises` to also consider all values that happen to be Thenable, not just Promises.
+The `checkAllThenable` option triggers `no-floating-promises` to also consider all values that happen to be Thenables with `.then()` two parameters, not just Promises.
 This can be useful if your code works with older `Promise` polyfills instead of the native `Promise` class.
 
 <Tabs>
--- a/packages/eslint-plugin/docs/rules/no-floating-promises.mdx
+++ b/packages/eslint-plugin/docs/rules/no-floating-promises.mdx
@@ -84,6 +84,53 @@ await Promise.all([1, 2, 3].map(async x => x + 1));
 
 ## Options
 
+### `checkThenables`
+
+A ["Thenable"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables) value is an object which has a `then` method, such as a `Promise`.
+Other Thenables include TypeScript's built-in `PromiseLike` interface and any custom object that happens to have a `.then()`.
+
+The `checkAllThenable` option triggers `no-floating-promises` to also consider all values that happen to be Thenable, not just Promises.
+This can be useful if your code works with older `Promise` polyfills instead of the native `Promise` class.
+
+<Tabs>
+<TabItem value="❌ Incorrect">
+
+```ts option='{"checkThenables": true}'
+declare function createPromiseLike(): PromiseLike<string>;
+
+createPromiseLike();
+
+interface MyThenable {
+  then(onFulfilled: () => void, onRejected: () => void): MyThenable;
+}
+
+declare function createMyThenable(): MyThenable;
+
+createMyThenable();
+```
+
+</TabItem>
+<TabItem value="✅ Correct">
+
+```ts option='{"checkThenables": true}'
+declare function createPromiseLike(): PromiseLike<string>;
+
+await createPromiseLike();
+
+interface MyThenable {
+  then(onFulfilled: () => void, onRejected: () => void): MyThenable;
+}
+
+declare function createMyThenable(): MyThenable;
+
+await createMyThenable();
+```
+
+</TabItem>
+</Tabs>
+
+> This option is enabled by default in v7 but will be turned off by default in v8.
+
 ### `ignoreVoid`
 
 This option, which is `true` by default, allows you to stop the rule reporting promises consumed with void operator.
--- a/packages/eslint-plugin/docs/rules/no-floating-promises.mdx
+++ b/packages/eslint-plugin/docs/rules/no-floating-promises.mdx
@@ -89,7 +89,7 @@ await Promise.all([1, 2, 3].map(async x => x + 1));
 A ["Thenable"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables) value is an object which has a `then` method, such as a `Promise`.
 Other Thenables include TypeScript's built-in `PromiseLike` interface and any custom object that happens to have a `.then()`.
 
-The `checkAllThenable` option triggers `no-floating-promises` to also consider all values that happen to be Thenables with `.then()` two parameters, not just Promises.
+The `checkThenables` option triggers `no-floating-promises` to also consider all values that satisfy the Thenable shape (a `.then()` method that takes two callback parameters), not just Promises.
 This can be useful if your code works with older `Promise` polyfills instead of the native `Promise` class.
 
 <Tabs>
@@ -129,7 +129,9 @@ await createMyThenable();
 </TabItem>
 </Tabs>
 
-> This option is enabled by default in v7 but will be turned off by default in v8.
+:::info
+This option is enabled by default in v7 but will be turned off by default in v8.
+:::
 
 ### `ignoreVoid`
 
--- a/packages/eslint-plugin/src/rules/no-floating-promises.ts
+++ b/packages/eslint-plugin/src/rules/no-floating-promises.ts
@@ -8,6 +8,7 @@ import {
   createRule,
   getOperatorPrecedence,
   getParserServices,
+  isSymbolFromDefaultLibrary,
   OperatorPrecedence,
   readonlynessOptionsDefaults,
   readonlynessOptionsSchema,
@@ -16,9 +17,10 @@ import {
 
 type Options = [
   {
-    ignoreVoid?: boolean;
-    ignoreIIFE?: boolean;
     allowForKnownSafePromises?: TypeOrValueSpecifier[];
+    checkThenables?: boolean;
+    ignoreIIFE?: boolean;
+    ignoreVoid?: boolean;
   },
 ];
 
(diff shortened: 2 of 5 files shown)


Josh-Cena:
I don't get the motivation for this, TBH. How can something "happen to be thenable" without noticing? Does `then` have any other meaning other than interoperating with promises? JavaScript itself resolves thenables, so if your thenable is not supposed to quack like a promise you will shoot yourself in the foot everywhere, like returning it from an async function, etc.

packages/eslint-plugin/src/rules/no-floating-promises.ts
```diff
@@ -356,14 +366,10 @@ export default createRule<Options, MessageId>({
...
 
-    // Modified from tsutils.isThenable() to only consider thenables which can be
-    // rejected/caught via a second parameter. Original source (MIT licensed):
-    //
-    //   https://github.com/ajafff/tsutils/blob/49d0d31050b44b81e918eae4fbaf1dfe7b7286af/util/type.ts#L95-L125
     function isPromiseLike(node: ts.Node, type?: ts.Type): boolean {
       type ??= checker.getTypeAtLocation(node);
 
-      // Ignore anything specified by `allowForKnownSafePromises` option.
+      // The highest priority it so allow anything allowlisted
```
kirkwaiblinger, line 372:
```suggestion
      // The highest priority is to allow anything allowlisted
```

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -84,6 +84,53 @@ await Promise.all([1, 2, 3].map(async x => x + 1));
 
 ## Options
 
+### `checkThenables`
+
+A ["Thenable"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables) value is an object which has a `then` method, such as a `Promise`.
+Other Thenables include TypeScript's built-in `PromiseLike` interface and any custom object that happens to have a `.then()`.
+
+The `checkAllThenable` option triggers `no-floating-promises` to also consider all values that happen to be Thenable, not just Promises.
```
kirkwaiblinger, line 92: nit - technically this is very slightly false (since it only considers thenables that can catch a rejected state with a second parameter), but I doubt anyone cares.

kirkwaiblinger (approved):
rocket

Josh-Cena:
Why would one expect a polyfill to _not_ be checked like a real promise?

JoshuaKGoldberg:
One example I've believe I've been told is that if a framework implements its own classes that happen to have a similar API to Promises, it's inconvenient to have the rule fire on them by default.

cc @mcollina as I think this was from you and Fastify, but I might be mistaken / mistinterpreting.

Josh-Cena:
The issue mentions `node:test`, but doesn't `test()` return a real promise?

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
JoshuaKGoldberg: Good point! I'll fix it just to be safe.

JoshuaKGoldberg:
@Josh-Cena the Node issue does, but our #8433 -> this PR isn't around that I don't think.

Josh-Cena:
That issue doesn't mention a valid use case where you want a thenable to not be checked like a promise though.

JoshuaKGoldberg:
Great points. Putting into draft per https://github.com/typescript-eslint/typescript-eslint/issues/8433#issuecomment-2149909831. Thanks!

JoshuaKGoldberg:
#8433 was confirmed as good to go, so un-drafting. 👍

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -733,6 +733,41 @@ promise().then(() => {});
...
+
+declare function createMyThenable(): MyThenable;
+
+createMyThenable();
+      `,
+      options: [{ checkThenables: false }],
+    },
   ],
 
   invalid: [
```
kirkwaiblinger, line 773: (testing) Looks like there aren't currently any invalid test cases with `checkThenables: false` (since for now the default is `checkThenables: true`). I'm thinking we should add:

- Some invalid tests to ensure that ordinary promises still will flag with the `checkThenables: false` option enabled).
- Some invalid tests around whatever the correct behavior with the `allowForKnownSafePromises` also enabled is.... (also - does allowForKnownSafePromises have any point at all with `checkThenables: false`? Won't the implementation just see branded promises as not the global promise type, and therefore not checked anyway?)

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -84,6 +84,53 @@ await Promise.all([1, 2, 3].map(async x => x + 1));
 
 ## Options
 
+### `checkThenables`
```
kirkwaiblinger, line 87: (optional - up to your judgement)
I personally would love to see it spelled out that there are really two cases.
1. Traditional Thenables which are intended to act like promises, and therefore _should_ be subject to the same checks around rejection that ordinary Promises are. (which is probably not terribly important anymore, since people are probably almost all using ordinary Promises, or at least ordinary Promise _types_ with a polyfill provided by a build step, but probably _not_ really using custom Promise polyfill types anymore very frequently) 
1. Custom objects that (ab)use the Thenable/async/await protocol to hijack the `await` syntax for behavior which may bear no resemblance at all to Promises. (of which the only example I've heard come up in discussions is the fastify case)

kirkwaiblinger (changes requested):
Code makes sense to me! only blocker is some test coverage

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
JoshuaKGoldberg: This is good info! Linking to #8517: I think I'll want to put deeper dives / contextual info in the dedicated docs.

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
JoshuaKGoldberg: > ordinary promises still will flag

👍 added, good call

> does allowForKnownSafePromises have any point at all with `checkThenables: false`?

Ooh yes, added a couple `class MyPromise<T> extends Promise<T> { / * ... */ }` tests. They failed until I switched from `isSymbolFromDefaultLibrary` (one shot) to `isBuiltinSymbolLike` (recursive checks of base types) for the built-in `Promise` check. Nice!

kirkwaiblinger (approved):
yay

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -84,6 +84,53 @@ await Promise.all([1, 2, 3].map(async x => x + 1));
...
+
+A ["Thenable"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables) value is an object which has a `then` method, such as a `Promise`.
+Other Thenables include TypeScript's built-in `PromiseLike` interface and any custom object that happens to have a `.then()`.
+
+The `checkAllThenable` option triggers `no-floating-promises` to also consider all values that happen to be Thenables with `.then()` two parameters, not just Promises.
```
Josh-Cena, line 92:
```suggestion
The `checkThenables` option triggers `no-floating-promises` to also consider all values that happen to be Thenables with `.then()` that takes two parameters, not just Promises.
```

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -84,6 +84,53 @@ await Promise.all([1, 2, 3].map(async x => x + 1));
...
+
+declare function createMyThenable(): MyThenable;
+
+await createMyThenable();
+```
+
+</TabItem>
+</Tabs>
+
+> This option is enabled by default in v7 but will be turned off by default in v8.
```
Josh-Cena, line 132:
```suggestion
:::info
This option is enabled by default in v7 but will be turned off by default in v8.
:::
```

JoshuaKGoldberg:
Confirmed with @Josh-Cena that we're good to merge with those docs suggestions. 🚀!

PR #9297, opened 2024-06-06
fix(eslint-plugin): [strict-boolean-expressions] support branded booleans

### LINKED ISSUE #9279: Bug: [strict-boolean-expressions] unsupported branded booleans

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.5&fileType=.tsx&code=FAFwngDgpgBAsmAQge2QGygQwHYwLwwBGqGOMAZDAN4wD6tiASgIIByAIgFwwDkCACgCcoAYwCWyAK4BnFOizYeMAL7A1AM0nYRICbgC2YAGJadegBTF5OAGqY0kqNwRzS2AJTdpIQWOwBzamAYGDF1GEsSBTsHKHcgkJDhEElBXB5kAGseYJUYKDRpWCpcpKgUtN5M5ByQ1VUgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZfKsugIwHs%2BSAIZN6AD2LQUySnyaoMkRNGh9okcNgC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
type MyBoolean = boolean & { __BRAND: 'MyPreciousBoolean' }


function myFunction(booleanValue: MyBoolean): string {
  if (booleanValue) {
    return 'ok'
  } else {
    return 'ko'
  }
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/strict-boolean-expressions": "error"
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expect no error on line `if (booleanValue) {`

### Actual Result

eslint throws an error `Unexpected object value in conditional. The condition is always true` it is seeing the branded boolean as an object.

### Additional Info

_No response_

Vinccool96:
## PR Checklist

- [X] Addresses an existing open issue: fixes #9279 

## Overview

Branded booleans are now supported.

Note: It might be a good idea to verify if branding isn't implemented for the numbers and the strings, so that we don't have an errer if, with the options, we allow numbers and strings. That wasn't done because it wasn't part of the scope of the bug report.


Josh-Cena:
Can we simply check if something is assignable to `boolean`? I thought `isAssignableTo` is a thing now?

JoshuaKGoldberg:
> Can we simply check if something is assignable to `boolean`? I thought `isAssignableTo` is a thing now?

For reference, https://github.com/typescript-eslint/typescript-eslint/discussions/7936:

> Here is our intent in typescript-eslint:

> 1. Do nothing with this information for our current major version, v6 v7
> 2. Once v8 is released, we'll add `isTypeAssignableTo` to the allowlist of private APIs we feel comfortable using, and unblock all issues marked as blocked

Josh-Cena:
Fair then. Also can we make sure that the `allow*` options all work with branded primitives?

Vinccool96:
> > Can we simply check if something is assignable to `boolean`? I thought `isAssignableTo` is a thing now?
> 
> For reference, #7936:
> 
> > Here is our intent in typescript-eslint:
> 
> > 1. Do nothing with this information for our current major version, ~v6~ v7
> > 2. Once v8 is released, we'll add `isTypeAssignableTo` to the allowlist of private APIs we feel comfortable using, and unblock all issues marked as blocked

Then I think I'll do this fix for v8, since it looks like it would be a mess to solve for v7

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -957,3 +962,22 @@ function isArrayLengthExpression(
...
+    return false;
+  }
+
+  const types = (type as ts.IntersectionType).types;
+  return types.some(childType =>
+    tsutils.isTypeFlagSet(
+      childType,
+      ts.TypeFlags.BooleanLiteral | ts.TypeFlags.Boolean,
+    ),
+  );
```
StyleShit, line 982: IIRC, you can do:

```suggestion
  return type.isIntersection() && 
	  type.types.some(childType =>
	    tsutils.isTypeFlagSet(
	      childType,
	      ts.TypeFlags.BooleanLiteral | ts.TypeFlags.Boolean,
	    ),
	  );
```

and then you won't need the type assertion

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -821,7 +821,8 @@ export default createRule<Options, MessageId>({
       | 'string'
       | 'truthy boolean'
       | 'truthy number'
-      | 'truthy string';
+      | 'truthy string'
+      | 'branded boolean';
```
JoshuaKGoldberg, line 825: [Style] Super nitpicky, but - these _were_ in alphabetical order. Up for alphabetizing?

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
JoshuaKGoldberg: 👍 yup, this. Searching for `isIntersection` shows a couple of options.

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
JoshuaKGoldberg, line 1: [Testing] Could you add some tests for how branded booleans interact with existing known edge cases? For example, inside `invalid`, the `if (('' && {}) || (0 && void 0)) { }` is ... scary.

Also: how about `false & { brand }` and `true & { brand }`?

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -909,7 +910,11 @@ export default createRule<Options, MessageId>({
             ),
         )
       ) {
-        variantTypes.add('object');
+        if (types.some(type => isBrandedBoolean(type))) {
+          variantTypes.add('branded boolean');
+        } else {
+          variantTypes.add('object');
+        }
```
JoshuaKGoldberg, line 917: [Style] 🧹 ✨  also nitpicky:

```suggestion
        variantTypes.add(
          types.some(isBrandedBoolean) ? 'branded boolean' : 'object',
        );
```

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -909,7 +910,11 @@ export default createRule<Options, MessageId>({
             ),
         )
       ) {
-        variantTypes.add('object');
+        if (types.some(type => isBrandedBoolean(type))) {
+          variantTypes.add('branded boolean');
```
JoshuaKGoldberg, line 914: [Refactor] Also, I switched this to be `boolean`:

```suggestion
          variantTypes.add('boolean');
```

...and all the current tests pass. Which I think kind of makes sense. I can't think of any reason why the rule would care about the difference between a _branded_ boolean and a non-branded one. Can you?

* If there are reasons to care, they should be in the tests.
* If not, let's stick with just `'boolean'`

JoshuaKGoldberg (changes requested):
✨ A good start! Some simplification and refactors suggested, and I think more comprehensive tests would be good too.

JoshuaKGoldberg:
> verify if branding isn't implemented for the numbers and the strings

I'm ok treating that as a followup - up to you! 🙂

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
Vinccool96: Why is it scary? What would the test case be?

JoshuaKGoldberg: It's scary because it's a complex expression with lots of little interactions between different operators in it. What happens if the `{}` is switched to a branded boolean? How about the `0` with a branded false? Those are the test cases I'm thinking of.

JoshuaKGoldberg (dismissed):
Looks good to me, thanks! 🚀

PR #9306, opened 2024-06-08
fix(typescript-estree): adds support for project services using extended config files

### LINKED ISSUE #9205: Bug(typescript-estree): Project services default project not working with extended tsconfigs

### Issue Description

I have identified two issues with project services.

1. If the default project's `tsconfig.json` has an `extends`, to say a `tsconfig.base.json` with a common set of compiler options, those compiler options are not used.  The example repo is a monorepo with liberal use of project references.  An example published package that uses a similar setup is https://github.com/Effect-TS/effect.
2. Fixing or working around the first issue (both are demonstrated in the example), reveals that `parserOptions.extraFileExtensions` is not working as expected to setup `tsserver` for type checking using project references with non-standard file extensions.  The example demonstrates how `.vue` files are not type checking correctly.

I have created a basic [patch](https://github.com/higherorderfunctor/example-typescript-eslint-bug/blob/main/patches/%40typescript-eslint__typescript-estree%407.11.0.patch) that solves both issues (minus rigorous testing).

The first issue is triggered by `tsserver.readConfigFile` at https://github.com/typescript-eslint/typescript-eslint/blob/5ca7f6e563779157cac1ac1592e2f1b82068715a/packages/typescript-estree/src/create-program/createProjectService.ts#L74

Replacing with `tsserver.getParsedCommandLineOfConfigFile` solves the extends issue.  This was pretty much lifted from https://github.com/typescript-eslint/typescript-eslint/blob/5ca7f6e563779157cac1ac1592e2f1b82068715a/packages/typescript-estree/src/create-program/useProvidedPrograms.ts#L66

Fixing this issue will reveal the second issue with file extensions.  The patch corrects this by forwarding `extraFileExtensions` to `tsserver` using `service.setHostConfiguration`.

https://github.com/higherorderfunctor/example-typescript-eslint-bug

### Repro Steps

Directions are located in the example repo's [README](https://github.com/higherorderfunctor/example-typescript-eslint-bug).

The repo has a patch to fix the bugs (enabled by default) and needs to disabled to demonstrate the bugs.

### Versions

```
node --version
v20.12.2

pnpm --version
8.15.5

pnpm list -r
Legend: production dependency, optional only, dev only

example-typescript-eslint-bug@1.0.0 <REDACTED>/example-typescript-eslint-bug (PRIVATE)

devDependencies:
eslint 8.57.0
eslint-plugin-vue 9.26.0
typescript 5.4.5
typescript-eslint 7.11.0
vue-eslint-parser 9.4.3
vue-tsc 2.0.19

package-a@1.0.0 <REDACTED>/example-typescript-eslint-bug/packages/package-a (PRIVATE)

dependencies:
@types/jsdom 21.1.7
@types/node 20.13.0
vite 5.2.12
vue 3.4.27

cat /etc/os-release
ANSI_COLOR="1;34"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="24.11.20240529.ad57eef"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 24.11 (Vicuña)"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="24.11 (Vicuña)"
VERSION_CODENAME="vicuña"
VERSION_ID="24.11"
```

higherorderfunctor:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9205

## Overview

- Changes `packages/typescript-estree/src/create-program/createProjectService.ts` to use `tsserver.getParsedCommandLineOfConfigFile` instead of `tsserver.readConfigFile` to process extended tsconfigs.
- Moves the `tsserver.getParsedCommandLineOfConfigFile` boilerplate from `packages/typescript-estree/src/create-program/useProvidedPrograms.ts` into a new file `packages/typescript-estree/src/create-program/getParsedConfigFile.ts` to keep the code DRY.
- Adds tests for single and multiple extends to `packages/typescript-estree/tests/lib/createProjectService.test.ts`.  Also adds a couple tests to round out coverage for uncommon code paths to bring `getParsedConfigFile.ts` to 100% (`tsserver.sys === undefined` and triggering `getCanonicalFileName` with `tsserver.formatDiagnostics`).
- Adds a new fixture at `packages/typescript-estree/tests/fixtures/projectServicesComplex/`.

A few implementation notes for the reviewer:

- There is a hard `any` cast in `packages/typescript-estree/src/create-program/createProjectService.ts`.  I left a code comment with reasoning for doing so.  The original `tsserver.readConfigFile` was also `any` with a hard cast.
- `formatDiagnostics` in the new `packages/typescript-estree/src/create-program/getParsedConfigFile.ts` file was placed inside `getParsedConfigFile`.  This was only to share the lazy import of tsserver.  The existing Jest mocks seem to rely on tsserver being a lazy import.  I'm not very familiar with eslint's module import cache-bypassing behavior, so I don't know if there is a performance implication here.  The original lazy import is also still present in `packages/typescript-estree/tests/lib/createProjectService.test.ts`.  We could try adding tsserver as a parameter to `getParsedConfigFile` to remove the extra lazy import if we think it will be a problem.
- Switched from a global `mock` to a `doMock` inside a `beforeAll` for testing `tsserver.env === undefined` case.  I've primarily used mocha/chai, vitest, and bun test; so if it there is a cleaner way to do this in jest I can make that update.
- All previous unit tests in `packages/typescript-estree/tests/lib/createProjectService.test.ts` were truly unit tests mocking all calls to tsserver.  Wanting to test the extends behavior, some of the tests now call through to tsserver.  Since these are really testing integration functionality, please let me know if adjustments are required.  These tests also have slightly looser assertions since tsserver adds some metadata.  A code comment has been added to mark each.
- I did not end up needing to use all of the `projectServicesComplex` fixture, specifically the project references.  It may have future uses, but I can trim the currently unused project references if preferred.

Inline comments work for me on the PR.


packages/typescript-estree/src/create-program/getParsedConfigFile.ts
```diff
@@ -0,0 +1,64 @@
...
+  // parsed is not undefined, since we throw on failure.
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const result = parsed!;
+  if (result.errors.length) {
+    throw new Error(formatDiagnostics(result.errors));
+  }
+
+  return result;
+
+  // scoped to parent function to use lazy typescript import
```
JoshuaKGoldberg, line 54: [Docs] Nit: I think this is implied by the function's placement + contents, and the info isn't generally necessary to understand the program flow. Is there a reason to leave it in? (I won't be offended at all if you just like it and wontfix this nit 😄)

```suggestion
```

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Utility offered by parser to help consumers parse a config file.
+ *
+ * @param configFile the path to the tsconfig.json file, relative to `projectDirectory`
+ * @param projectDirectory the project directory to use as the CWD, defaults to `process.cwd()`
+ */
+function getParsedConfigFile(
+  configFile: string,
+  projectDirectory?: string,
+): ts.ParsedCommandLine {
```
JoshuaKGoldberg, line 16: [Refactor] I'm not a fan of functions throwing errors as a form of control flow. It can't be described in the function's signature and so can easily become hard to keep track of as functions get more complex.

We're seeing an example of here. The thrown error deeper in `getParsedConfigFile` is caught and used to throw a new message in `createProjectService`. I find that hard to keep track of.

What do you think about having this function instead return a `string` message to indicate there was a failure parsing?

I.e.:

```suggestion
): ts.ParsedCommandLine | string {
```

...and then later, instead of `throw new Error(...)`s, it'd return the string message. `onUnRecoverableConfigFileDiagnostic` could set `parsed` to be a `string`:

```ts
let parsingError: string | undefined;

const parsed = tsserver.getParsedCommandLineOfConfigFile(
  // ...
      onUnRecoverableConfigFileDiagnostic: diag => {
        parsingError = formatDiagnostics([diag]);
      },
  // ...
);
```

(or another strategy like that, e.g. having `let parsed: string | ts.ParsedCommandLine`)

packages/typescript-estree/tests/lib/createProjectService.test.ts
```diff
@@ -1,22 +1,42 @@
...
-  server: {
-    ProjectService: class {
-      setCompilerOptionsForInferredProjects =
-        mockSetCompilerOptionsForInferredProjects;
-    },
-  },
-}));
+const origSys = ts.sys;
+const origGetParsedCommandLineOfConfigFile =
+  ts.getParsedCommandLineOfConfigFile;
```
JoshuaKGoldberg, line 13: [Testing] The rewritten test mocking here focuses in part  on mocking out `tsserver.getParsedCommandLineOfConfigFile`. But that function is now no longer being directly called by `createProjectService`. It's inside a `getParsedConfigFile`. I think the tests would be a lot more straightforward by:

* Moving any existing tests for what's now in `getParsedConfigFile` into a new `getParsedConfigFile.test.ts`.
  * This is one of my personal unit testing ideologies: when possible, testing a function's logic by calling the function, not one of the other functions that happens to call that function
* If there are any complex interactions between >2 functions that you'd _really_ like to test at a high level, making a dedicated integration test

It'd be a bit more work now, but IMO would result in better delineated unit tests that don't have to do quite so much deep mocking and un-mocking.

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
```diff
@@ -0,0 +1,64 @@
...
+  // We import this lazily to avoid its cost for users who don't use the service
+  // TODO: Once we drop support for TS<5.3 we can import from "typescript" directly
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const tsserver = require('typescript/lib/tsserverlibrary') as typeof ts;
```
JoshuaKGoldberg, line 20: [Refactor] From the OP:

> We could try adding tsserver as a parameter to `getParsedConfigFile` to remove the extra lazy import if we think it will be a problem.

+1 from me, good idea. I think minimizing the number of ad-hoc `require(...)`s is very good.

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
JoshuaKGoldberg, line 1: [Testing] From the OP:

> I did not end up needing to use all of the projectServicesComplex fixture

I like you added some complex interactions 😄 it's good to test those! But IMO they belong in an integration test. It's not really "unit" level behavior they're testing. +1 to extracting them into an integration test if you have the time - or filing a followup issue if not. No worries either way.

JoshuaKGoldberg (changes requested):
I like the general direction! +1 on extracting out a `getParsedConfigFile`, that's really useful. 🚀 

Requesting changes mostly on test reorganization, and the `require()`ing too. Cheers!

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
higherorderfunctor: I'll get this change in.  I typically use an Either monad to keep errors in the type system, so the objective works for me.  Good call.

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
higherorderfunctor: @JoshuaKGoldberg just to confirm, I should move tests that rely on actually calling into tsserver into `packages/integration-tests/tests`?  These are the only integration tests I am finding after skimming the repo and online documentation?

JoshuaKGoldberg: Sorry, I'm being kind of vague 😅. What I'm suggesting is:

* To test granular logic in `createProjectService`, there should be unit tests in a `createProjectService.test.ts` - which can mock out other dependencies, such as `getParsedConfigFile` and `tsserver`
* To test granular logic in `getParsedConfigFile`, there should be unit tests in a `getParsedConfigFile.test.ts` - which can mock out other dependencies, such as  `tsserver`
* If there's any high-level edge-case-y stuff you want to test, such as making sure the project service logic can handle complex file systems, that would belong in a `packages/integration-tests/*` tests, yes - though I don't plan on requesting that in this PR

Does that seem reasonable to you?

higherorderfunctor: Understood on the refactor request for the unit tests.  I had some tests that call into `tsserver` to better understand its behavior and to test some assumptions having never worked with it before.  I can reevaluate if those are necessary to keep.  Thanks!

JoshuaKGoldberg:
👋 are you waiting on us for anything @higherorderfunctor?

higherorderfunctor:
> 👋 are you waiting on us for anything @higherorderfunctor?

Nope!  I'll be picking this back up this week / weekend.  Had a vacation and some deadlines the last two weeks.

JoshuaKGoldberg:
Great! Let's time box this to a week from now, then - we really want to release v8 as soon as possible.

higherorderfunctor:
Created `getParsedConfigFile.test.ts`.

Will revert most of the changes to `createProjectService.test.ts` except for the use of the new function (mocked).
Decide if dropping or creating the complex tests as an integration (probably drop).
Merge in v8 again and resolve conflicts.

Plan to be done by tonight.

higherorderfunctor:
@JoshuaKGoldberg I will have some time tomorrow to check on the CI for any errors or for any final edits you would like to make.  Thanks!

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (parsed?.errors.length) {
+    parsingError = formatDiagnostics(parsed.errors);
+  }
+
+  if (parsingError !== undefined) {
+    return parsingError;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  return parsed!;
```
JoshuaKGoldberg, line 54: [Style] Nit: a teeny cleanup 😛 

```suggestion
  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  return parsingError ?? parsed!;
```

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
```diff
@@ -0,0 +1,65 @@
...
+): ts.ParsedCommandLine | string {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (tsserver.sys === undefined) {
+    throw new Error(
+      '`getParsedConfigFile` is only supported in a Node-like environment.',
+    );
+  }
+
+  const getCurrentDirectory = (): string =>
+    (projectDirectory && path.resolve(projectDirectory)) || process.cwd();
```
bradzacher, line 26: Nitpick - avoid nested logical short-circuiting as it can be hard to read and reason about.

```suggestion
    projectDirectory ? path.resolve(projectDirectory) : process.cwd();
```

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
```diff
@@ -0,0 +1,65 @@
...
+      fileExists: fs.existsSync,
+      getCurrentDirectory,
+      readDirectory: tsserver.sys.readDirectory,
+      readFile: file => fs.readFileSync(file, 'utf-8'),
+      useCaseSensitiveFileNames: tsserver.sys.useCaseSensitiveFileNames,
+    },
+  );
+
+  if (parsed?.errors.length) {
+    parsingError = formatDiagnostics(parsed.errors);
```
bradzacher, line 46: Question - is it possible for TS to emit an error from the unrecoverable diagnostics hook AND return errors via `parsed.errors`

If yes - then it's possible for us to clobber and lose those unrecoverable diagnostics thanks to this reassignment.

Suggestion: use two separate variables to remove the possibility.

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Utility offered by parser to help consumers parse a config file.
+ *
+ * @param configFile the path to the tsconfig.json file, relative to `projectDirectory`
+ * @param projectDirectory the project directory to use as the CWD, defaults to `process.cwd()`
+ */
+function getParsedConfigFile(
+  tsserver: typeof ts,
+  configFile: string,
+  projectDirectory?: string,
+): ts.ParsedCommandLine | string {
```
bradzacher, line 17: Question: why are we returning a string here instead of just throwing?

It looks like we could simplify code I'm both usages if we just threw directly.

packages/typescript-estree/tests/lib/createProjectService.test.ts
```diff
@@ -1,15 +1,16 @@
 import debug from 'debug';
```
bradzacher, line 1: Side note (not this PR) - we should probably build out a test case using `typescript-estree/tests/fixtures` that allows us to do tests with project service without mocking out the entire world.

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
JoshuaKGoldberg: Haha, I roughly requested the opposite thing previously: https://github.com/typescript-eslint/typescript-eslint/pull/9306#discussion_r1651976054

At this point I would feel bad making higherorderfunctor manually go back the other way.

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
higherorderfunctor: Can do.  I've gotten bitten a few times by empty strings so explicitly checking against undefined is just habit.

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
higherorderfunctor: > Haha, I roughly requested the opposite thing previously: [#9306 (comment)](https://github.com/typescript-eslint/typescript-eslint/pull/9306#discussion_r1651976054)
> 
> At this point I would feel bad making higherorderfunctor manually go back the other way.

@JoshuaKGoldberg I can go back if we think this is the best approach.  Most of this is not my code; I just moved it from [useProvidedPrograms.ts](https://github.com/typescript-eslint/typescript-eslint/pull/9306/files/45551dc2e74379e587396cb1d28748dfca395cac#diff-a4d9d90792db4548a83bc759c16498571e323de4570b37bd6682c2c2dbab730e).

packages/typescript-estree/tests/lib/createProjectService.test.ts
higherorderfunctor: I started to do this but I think the consensus was it fit more into integration: https://github.com/typescript-eslint/typescript-eslint/tree/148a9d0f7c670a74152279974988e9107e4be2f7/packages/typescript-estree/tests/fixtures/projectServicesComplex.

I know there is a desire to release to v8 soon.  I don't have a ton of time this week.  @JoshuaKGoldberg and @bradzacher, should I just open a new issue to follow back up on this?

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
higherorderfunctor: If we throw (below comment) it would just be the first that occurred.  Depending on the response to [my question](https://github.com/typescript-eslint/typescript-eslint/pull/9306#discussion_r1690295031) in that thread, it will dictate the direction here.

packages/typescript-estree/tests/lib/createProjectService.test.ts
JoshuaKGoldberg: A new issue would be very reasonable, yes please!

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
JoshuaKGoldberg: Since Brad is one way and I'm the other, but neither way actually impacts the end-user experience, I'm happy to just step back and say "whatever everyone else prefers".

higherorderfunctor: Reverted to throwing as it simplifies the ask in https://github.com/typescript-eslint/typescript-eslint/pull/9306#discussion_r1689960869

packages/typescript-estree/src/create-program/getParsedConfigFile.ts
higherorderfunctor: Reverted to throwing.

packages/typescript-estree/tests/lib/createProjectService.test.ts
higherorderfunctor: Opened https://github.com/typescript-eslint/typescript-eslint/issues/9641

higherorderfunctor:
All changes made.  Let me know if anything else!

I think the running CI job is stuck, [netlify shows complete](https://app.netlify.com/sites/typescript-eslint/deploys/66a27a1886adfd0008e689a6).

Opened https://github.com/typescript-eslint/typescript-eslint/issues/9641 for the test clean up.

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -123,38 +124,26 @@ export function createProjectService(
...
 
     service.setCompilerOptionsForInferredProjects(
-      (
-        configRead.config as {
-          compilerOptions: ts.server.protocol.InferredProjectCompilerOptions;
-        }
-      ).compilerOptions,
+      // NOTE: The inferred projects API is not intended for source files when a tsconfig
+      // exists.  There is no API that generates an InferredProjectCompilerOptions suggesting
+      // it is meant for hard coded options passed in.  Hard casting as a work around.
```
JoshuaKGoldberg, line 144: Super teeny nitpicky nit: I see "casting" as a runtime thing. "Asserting" is what we do in the type system. 

```suggestion
      // it is meant for hard coded options passed in. Hard asserting as a work around.
```

JoshuaKGoldberg (approved):
🔥 thanks!

PR #9330, opened 2024-06-11
feat(eslint-plugin): [no-unused-vars] report if var used only in typeof

### LINKED ISSUE #3325: [no-unused-vars] report when runtime var is only used within a type (`type MyType = typeof varName`)

**Repro**

```JSON
{
  "rules": {
  }
}
```

```TS
const foo = "bar"

export type Baz = typeof foo
```

Run it locally via https://github.com/happycollision/eslint-no-unused-vars-typescript-repro

**Expected Result**
An eslint error that `foo` is unused. (Just like it would if we removed the type def line.)

**Actual Result**
No eslint error reported.

**Additional Info**
If you agree with my thoughts below, then this is a bug. If you disagree, then I guess it is a feature request that'd need some options to opt into it.

I am of the opinion that if you create a runtime variable, you ought to be expected to use that variable within your runtime code. Using a runtime variable for purely type definition purposes means that when your code is compiled (depending on your compiler of choice and options), you'll have an extra variable handing around that does nothing. In all likelihood, this is an error and you _meant_ to use the variable but forgot to add it to whatever other part of your code needed it. This is exactly why I realized that this bug exists. 😄 

I cannot think of a single case where you'd _want_ such behavior, but I am totally willing to believe they might exist.

**Versions**

| `@typescript-eslint/eslint-plugin` | `4.22.0` |
| `@typescript-eslint/parser`        | `4.22.0` |
| `ESLint`                           | `7.25.0` |
| `node`                             | `v11.12.0` |

StyleShit:
## PR Checklist

- [X] Addresses an existing open issue: fixes #3325

### DIFF
--- a/packages/eslint-plugin/src/rules/no-shadow.ts
+++ b/packages/eslint-plugin/src/rules/no-shadow.ts
@@ -1,12 +1,9 @@
-import type {
-  Definition,
-  ImportBindingDefinition,
-} from '@typescript-eslint/scope-manager';
 import { DefinitionType, ScopeType } from '@typescript-eslint/scope-manager';
 import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
 import { AST_NODE_TYPES, ASTUtils } from '@typescript-eslint/utils';
 
 import { createRule } from '../util';
+import { isTypeImport } from '../util/isTypeImport';
 
 type MessageIds = 'noShadow' | 'noShadowGlobal';
 type Options = [
@@ -99,17 +96,6 @@ export default createRule<Options, MessageIds>({
       return (
         variable.defs[0].type === DefinitionType.Parameter &&
         variable.name === 'this'
-      );
-    }
-
-    function isTypeImport(
-      definition?: Definition,
-    ): definition is ImportBindingDefinition {
-      return (
-        definition?.type === DefinitionType.ImportBinding &&
-        (definition.parent.importKind === 'type' ||
-          (definition.node.type === AST_NODE_TYPES.ImportSpecifier &&
-            definition.node.importKind === 'type'))
       );
     }
 
--- a/packages/eslint-plugin/src/rules/no-use-before-define.ts
+++ b/packages/eslint-plugin/src/rules/no-use-before-define.ts
@@ -3,6 +3,7 @@ import type { TSESTree } from '@typescript-eslint/utils';
 import { AST_NODE_TYPES, TSESLint } from '@typescript-eslint/utils';
 
 import { createRule } from '../util';
+import { referenceContainsTypeQuery } from '../util/referenceContainsTypeQuery';
 
 const SENTINEL_TYPE =
   /^(?:(?:Function|Class)(?:Declaration|Expression)|ArrowFunctionExpression|CatchClause|ImportDeclaration|ExportNamedDeclaration)$/;
@@ -104,24 +105,6 @@ function isNamedExports(reference: TSESLint.Scope.Reference): boolean {
     identifier.parent.type === AST_NODE_TYPES.ExportSpecifier &&
     identifier.parent.local === identifier
   );
-}
-
-/**
- * Recursively checks whether or not a given reference has a type query declaration among it's parents
- */
-function referenceContainsTypeQuery(node: TSESTree.Node): boolean {
-  switch (node.type) {
-    case AST_NODE_TYPES.TSTypeQuery:
-      return true;
-
-    case AST_NODE_TYPES.TSQualifiedName:
-    case AST_NODE_TYPES.Identifier:
-      return referenceContainsTypeQuery(node.parent);
-
-    default:
-      // if we find a different node, there's no chance that we're in a TSTypeQuery
-      return false;
-  }
 }
 
 /**
--- a/packages/eslint-plugin/src/util/collectUnusedVariables.ts
+++ b/packages/eslint-plugin/src/util/collectUnusedVariables.ts
@@ -15,6 +15,9 @@ import {
   TSESLint,
 } from '@typescript-eslint/utils';
 
+import { isTypeImport } from './isTypeImport';
+import { referenceContainsTypeQuery } from './referenceContainsTypeQuery';
+
 interface VariableAnalysis {
   readonly unusedVariables: ReadonlySet<ScopeVariable>;
   readonly usedVariables: ReadonlySet<ScopeVariable>;
@@ -784,6 +787,8 @@ function isUsedVariable(variable: ScopeVariable): boolean {
   const enumDeclNodes = getEnumDeclarations(variable);
   const isEnumDecl = enumDeclNodes.size > 0;
 
+  const isImportedAsType = variable.defs.every(isTypeImport);
+
   let rhsNode: TSESTree.Node | null = null;
 
   return variable.references.some(ref => {
@@ -794,6 +799,7 @@ function isUsedVariable(variable: ScopeVariable): boolean {
     return (
       ref.isRead() &&
       !forItself &&
+      !(!isImportedAsType && referenceContainsTypeQuery(ref.identifier)) &&
       !(isFunctionDefinition && isSelfReference(ref, functionNodes)) &&
       !(isTypeDecl && isInsideOneOf(ref, typeDeclNodes)) &&
       !(isModuleDecl && isSelfReference(ref, moduleDeclNodes)) &&
--- a/packages/eslint-plugin/src/util/isTypeImport.ts
+++ b/packages/eslint-plugin/src/util/isTypeImport.ts
@@ -1,0 +1,27 @@
+import type {
+  Definition,
+  ImportBindingDefinition,
+} from '@typescript-eslint/scope-manager';
+import { DefinitionType } from '@typescript-eslint/scope-manager';
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+
+/**
+ * Determine whether a variable definition is a type import. e.g.:
+ *
+ * ```ts
+ * import type { Foo } from 'foo';
+ * import { type Bar } from 'bar';
+ * ```
+ *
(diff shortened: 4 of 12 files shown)


packages/eslint-plugin/src/rules/no-shadow.ts
```diff
@@ -102,17 +99,6 @@ export default createRule<Options, MessageIds>({
...
-    function isTypeImport(
-      definition?: Definition,
-    ): definition is ImportBindingDefinition {
-      return (
-        definition?.type === DefinitionType.ImportBinding &&
-        (definition.parent.importKind === 'type' ||
-          (definition.node.type === AST_NODE_TYPES.ImportSpecifier &&
-            definition.node.importKind === 'type'))
-      );
-    }
```
StyleShit, line 114: This function has been moved as is to a util so it can be reused

packages/eslint-plugin/src/rules/no-use-before-define.ts
```diff
@@ -106,24 +107,6 @@ function isNamedExports(reference: TSESLint.Scope.Reference): boolean {
...
-    case AST_NODE_TYPES.TSQualifiedName:
-    case AST_NODE_TYPES.Identifier:
-      return referenceContainsTypeQuery(node.parent);
-
-    default:
-      // if we find a different node, there's no chance that we're in a TSTypeQuery
-      return false;
-  }
-}
-
```
StyleShit, line 126: This function has been moved as is to a util so it can be reused

packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -739,6 +744,7 @@ function isUsedVariable(variable: TSESLint.Scope.Variable): boolean {
     return (
       ref.isRead() &&
       !forItself &&
+      !(!isImportedAsType && referenceContainsTypeQuery(ref.identifier)) &&
```
StyleShit, line 747: With the current logic, this is allowed:

```TS
import type { valueVariable } from './a';

export type Foo = typeof valueVariable;
```

Not sure whether it's the right behavior, but there is a test for this already:
https://github.com/typescript-eslint/typescript-eslint/blob/33f723f841badf6c34c5a7092eb2cd2438de9f77/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts#L830-L834

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
```diff
@@ -766,7 +766,7 @@ export function foo() {
     `,
     // https://github.com/typescript-eslint/typescript-eslint/issues/5152
     `
-function foo<T>(value: T): T {
+export function foo<T>(value: T): T {
   return { value };
 }
 export type Foo<T> = typeof foo<T>;
```
StyleShit, line 772: This change means that this PR is actually a breaking change, should we move it to the next major?

bradzacher:
> When I lint it locally, it fails (for valid reasons), but it doesn't fail in the CI. Is it normal? 

It happens, but it's not supposed to.
Usually it's due to the nx cache not being invalidated. (cc @JamesHenry for the resident expert on this)

JamesHenry:
@bradzacher is right, if you look at the logs, the `utils:lint` was a remote cache hit.

The history on this branch looks weird. There is only one run, despite the large number of commits, and the latest commit that actually triggered the CI run is blank:

blank commit screenshot

I have to imagine this funkiness is what has caused the undesired remote cache hit. I am going to resync the branch now via the GitHub UI

JamesHenry:
Oh I guess all merge commits show up like that in the GitHub UI, mine was the same just then...

Let's see what happens with this run and if it's still a cache hit I'll dig further

bradzacher:
https://cloud.nx.app/runs/knS8rHtlsm/task/utils%3Alint

looks like it failed as expected in CI!

JamesHenry:
@bradzacher Cool, I bumped us to latest Nx because I didn't want to start troubleshooting when we were so far behind. I also added some missing build dependencies based on the contents of the `eslint.config.mjs`.

If any of you see the unexpected cache hits again please let me know

packages/eslint-plugin/src/util/collectUnusedVariables.ts
JoshuaKGoldberg: Yeah, this is fine. The variable is imported as a type and used as a type. There's no unused runtime stuff.

packages/eslint-plugin/src/util/isTypeImport.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Determine whether a variable definition is a type import. e.g.:
+ *
+ * ```ts
+ * import type { Foo } from 'foo';
+ * import { type Bar } from 'bar';
+ * ```
+ *
+ * @param definition - The variable definition to check.
+ *
+ * @returns boolean
```
JoshuaKGoldberg, line 18: [Docs] This is an example of when explicit type annotations / comments are not very helpful. This function is a type predicate; explicitly saying `@returns boolean` is _less_ information than that.

```suggestion
```

packages/eslint-plugin/src/util/isTypeImport.ts
JoshuaKGoldberg, line 1: [Praise] I like the reusing of utilities 🙂 nice!

packages/eslint-plugin/src/util/referenceContainsTypeQuery.ts
```diff
@@ -0,0 +1,20 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+
+/**
+ * Recursively checks whether or not a given reference has a type query declaration among it's parents
```
JoshuaKGoldberg, line 5: [Docs] An unnecessary _"or not"_, and an unnecessary apostrophe:

```suggestion
 * Recursively checks whether a given reference has a type query declaration among its parents
```

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
JoshuaKGoldberg: Heh, "breaking" is a loose word. Per https://typescript-eslint.io/users/versioning/#eslint-plugin, _"Fixes incorrect behavior in a rule that may or may not introduce additional reports."_ is not a breaking change.

If we took on a new major for everything that changed rule reports, we'd be on ... well, quite a large number. 😄

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
JoshuaKGoldberg, line 1: [Testing] Can you also add a test for the case of a single variable being used in both a type position and a non-type / runtime?

```ts
const value = 1;

export const alsoValue = value;

export type Type = typeof value;
```

The `.some` in the `usedOnlyAsType` variable value is clever, but makes me worry about how this interacts with multiple uses...

JoshuaKGoldberg (dismissed):
LGTM, nicely done! I like the way you approached this 😄 especially the shared utils and the `.some`. Just requesting changes on some small touchups, so leaving in `1 approval` in case another team member wants to take a look.

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
StyleShit: I think this existing test is doing what you're talking about, no?

https://github.com/StyleShit/typescript-eslint/blob/d99154f3a844dd79e48ad3c935a444369603593a/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts#L1144-L1148

packages/eslint-plugin/src/util/referenceContainsTypeQuery.ts
StyleShit: same as above, copy-pasted
fixed :)

packages/utils/src/ts-eslint/Linter.ts
```diff
@@ -20,6 +20,7 @@ export type MinimalRuleModule<
 > = Partial<Omit<RuleModule<MessageIds, Options>, 'create'>> &
   Pick<RuleModule<MessageIds, Options>, 'create'>;
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
bradzacher, line 23: Question :
Do we want to ignore declared things?
They're not technically values so using them in a typeof only isn't bad.

bradzacher (approved):

packages/utils/src/ts-eslint/Linter.ts
StyleShit: I thought about it as well, but wasn't sure what to do

IMO it's ok not to report them ,WDYT?

Josh-Cena: I think we should keep them reported by default, unless someone complains so we can reconsider. Can't think of a good use case for it at the moment other than global declarations.

StyleShit: I'm ok with that 😄 

Maybe we can add an option in the future if needed

ManorHazaz (approved):

bradzacher:
I'm going to retarget this for v8
We've landed a bunch of changes to the rule in v8 already - so I think it's going to be easier if we rebase this on top of v8 and resolve the conflicts before merging it - instead of landing it in v7 and then trying to resolve conflicts made in 3 or 4 changes.

bradzacher:
conflicts

Yeah, that's what I was afraid of 😅 
@StyleShit please resolve the conflicts and we can get this merged into v8 :)

StyleShit:
@bradzacher I think it should be fine now

PR #9363, opened 2024-06-15
docs: [no-wrapper-object-types] clean up a bit of phrasing

kirkwaiblinger:
## PR Checklist

- [ ] Addresses an existing open issue: ~~fixes #000~~ followup to https://github.com/typescript-eslint/typescript-eslint/pull/9102/files/f25f33396926cf0bb03eac1d02d8636987d3ec28..e078105858728c2cfe6bdf81a145feae291cb43c#r1635229512


packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -10,13 +10,12 @@ import TabItem from '@theme/TabItem';
 > See **https://typescript-eslint.io/rules/no-wrapper-object-types** for documentation.
 
 The JavaScript language has a set of language types, but some of them correspond to two TypeScript types, which look similar: `boolean`/`Boolean`, `number`/`Number`, `string`/`String`, `bigint`/`BigInt`, `symbol`/`Symbol`, `object`/`Object`.
-The difference is that the lowercase variants are compiler intrinsics and specify the actual _runtime types_ (that is, the return value when you use the `typeof` operator), while the uppercase variants are _structural types_ defined in the library that can be satisfied by any user-defined object with the right properties, not just the real primitives.
+The difference is that the lowercase variants are compiler intrinsics and specify the actual _runtime types_ (that is, the type indicated when executing `typeof x` at runtime), while the uppercase variants are _structural types_ defined in the library that can be satisfied by any user-defined object with the right properties, including but not limited to the real primitives.
```
kirkwaiblinger, line 13: made slight edit here to try to clarify that it's about the _runtime_ `typeof` operator in TS vs the type-time `typeof` operator, which also exists.

JoshuaKGoldberg (approved):
🙌 lovely!

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -10,13 +10,12 @@ import TabItem from '@theme/TabItem';
...
 JavaScript also has a "wrapper" class object for each of those primitives, such as [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) and [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number).
 These wrapper objects are assignable to the uppercase types, but not to the lowercase types.
 
-Using the primitives like `0` instead of object wrappers like `new Number(0)` is generally considered a JavaScript best practice.
-JavaScript programs typically work with the real number primitives, rather than objects that "look like" numbers.
-Primitives are simpler to conceptualize and work with `==` and `===` equality checks -- which their object equivalents do notDeepEqual.
+It is widely considered a JavaScript best practice to work directly with the built-in primitives, like `0`, rather than objects that "look like" numbers, like `new Number(0)`.
+Primitives are simpler to conceptualize, work with `==` and `===` equality checks -- which their object equivalents do notDeepEqual -- and have well-known behavior around truthiness/falsiness which is common to rely on.
```
Josh-Cena, line 18:
```suggestion
It is widely considered a JavaScript best practice to work directly with the built-in primitives, like `0`, rather than objects that "look like" numbers, like `new Number(0)`.
Primitives work with `==` and `===` equality checks and have more commonly relied-on behavior around truthiness/falsiness, while two objects that wrap the same primitive numbers do not compare equal, and _all_ wrapper objects are truthy regardless of the wrapped value.
Furthermore TypeScript prevents using `Number` in places that expect `number`—if a function accepts `number`, you cannot pass a value of type `Number` to it.
You also cannot do subtraction on `Number` values.
```

kirkwaiblinger: I'll think about this a bit. Two things I wanna pay attention to 
1. `number` is the specific example in this section, but it's supposed to make a point about the primitives in general, not _just_ numbers
2. If we give _too_ much justification, it can seem like this is something controversial that we're just weirdly strongly opinionated about rather than something that's not even slightly in dispute by anyone. 😆

Josh-Cena: That's true. Well it's up to your decision. I certainly want to keep this phrasing though:

> Primitives work with `==` and `===` equality checks and have more commonly relied-on behavior around truthiness/falsiness, while two objects that wrap the same primitive numbers do not compare equal, and _all_ wrapper objects are truthy regardless of the wrapped value.

kirkwaiblinger: Adapted this a bit to make space for examples, each using a different primitive type. Let me know what you think!

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -10,13 +10,16 @@ import TabItem from '@theme/TabItem';
...
 
-Using the primitives like `0` instead of object wrappers like `new Number(0)` is generally considered a JavaScript best practice.
-JavaScript programs typically work with the real number primitives, rather than objects that "look like" numbers.
-Primitives are simpler to conceptualize and work with `==` and `===` equality checks -- which their object equivalents do notDeepEqual.
+It is widely considered a JavaScript best practice to work directly with the built-in primitives, like `0`, rather than objects that "look like" the corresponding primitive, like `new Number(0)`.
+
+- Primitives have the expected value semantics with `==` and `===` equality checks, whereas their object counterparts are compared by reference. That is to say, `"str" === "str"` but `new String("str") !== new String("str")`.
+- Primitives have well-known behavior around truthiness/falsiness which is common to rely on, whereas all objects are truthy, regardless of the wrapped value (e.g. `new Boolean(false)` is truthy).
+- TS only allows arithmetic operations (e.g. `x - y`) to be performed on numeric primitives, not objects.
```
JoshuaKGoldberg, line 21: Nit: I don't think an abreviation in this user-facing a doc is worth it.

```suggestion
- TypeScript only allows arithmetic operations (e.g. `x - y`) to be performed on numeric primitives, not objects.
```

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
JoshuaKGoldberg, line 1: Sorry to cannonball in here late, but I find the current docs to be a lot of text (tl;dr). I took a stab at cleaning this up locally. WDYT?

```md
The JavaScript language has a set of built-in ("intrinsic") types which look similar: `boolean`/`Boolean`, `number`/`Number`, `string`/`String`, `bigint`/`BigInt`, `symbol`/`Symbol`, and `object`/`Object`.
The lowercase variants are _primitives_ such as `boolean` (e.g. `false`).
The uppercase variants are "wrapper" class objects for each of those primitives, such as [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) (e.g. `new Boolean(false)`).

In TypeScript, the lowercase variants are compiler intrinsics and specify the actual _runtime types_ (that is, the type indicated when executing `typeof x` at runtime).
The uppercase variants are _structural types_ defined in the library that can be satisfied by any user-defined object with the right properties, including but not limited to the real primitives.

It is widely considered a JavaScript best practice to work directly with the built-in primitives, like `0`, rather than objects that "look like" the corresponding primitive, like `new Number(0)`.

- Primitives have the expected value semantics with `==` and `===` equality checks, whereas their object counterparts are compared by reference. That is to say, `"str" === "str"` but `new String("str") !== new String("str")`.
- Primitives have well-known behavior around truthiness/falsiness which is common to rely on, whereas all objects are truthy, regardless of the wrapped value (e.g. `new Boolean(false)` is truthy).
- TS only allows arithmetic operations (e.g. `x - y`) to be performed on numeric primitives, not objects.

As a result, using the lowercase type names like `number` in TypeScript types instead of the uppercase names like `Number` is a better practice that describes code more accurately.
```

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
kirkwaiblinger: yep, good catch

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
kirkwaiblinger: Updated to this 👍

JoshuaKGoldberg (approved):
Yay!

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -9,15 +9,20 @@ import TabItem from '@theme/TabItem';
...
-JavaScript also has a "wrapper" class object for each of those primitives, such as [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) and [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number).
-These wrapper objects are assignable to the uppercase types, but not to the lowercase types.
-
-Using the primitives like `0` instead of object wrappers like `new Number(0)` is generally considered a JavaScript best practice.
-JavaScript programs typically work with the real number primitives, rather than objects that "look like" numbers.
-Primitives are simpler to conceptualize and work with `==` and `===` equality checks -- which their object equivalents do notDeepEqual.
-As a result, using the lowercase type names like `number` instead of the uppercase names like `Number` helps make your code behave more reliably.
+The JavaScript language has a set of built-in ("intrinsic") types which look similar: `boolean`/`Boolean`, `number`/`Number`, `string`/`String`, `bigint`/`BigInt`, `symbol`/`Symbol`, and `object`/`Object`.
```
Josh-Cena, line 12: I think this loses the nuance that we tried hard to keep in the original docs. JavaScript only has eight types: `number`, `object`, `string`, `symbol`, `boolean`, `bigint`, `undefined`, `null`. `Boolean`, `Number`, etc. are subtypes of `object`. It's only in TypeScript-land where they are sort of parallel and can be compared to each other.

Josh-Cena (approved):
I have _some_ reservations but I think this generally looks great

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
kirkwaiblinger: Neither here nor there, but 

> `Boolean`, `Number`, etc. are subtypes of `object`.

TIL that 
```ts
let o: object = false; // <-- TS error
// BUT
let b: Boolean = false;
o = b; // no error lol
``` 

The wrapper types are so bad lol

Josh-Cena: Yep—everything that doesn't claim to be a primitive is implicitly `object`. `object` does not actually mean "non-primitive"; it just means "is not definitely primitive".

packages/eslint-plugin/docs/rules/no-wrapper-object-types.mdx
```diff
@@ -9,15 +9,23 @@ import TabItem from '@theme/TabItem';
...
-JavaScript also has a "wrapper" class object for each of those primitives, such as [`Boolean`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) and [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number).
-These wrapper objects are assignable to the uppercase types, but not to the lowercase types.
-
-Using the primitives like `0` instead of object wrappers like `new Number(0)` is generally considered a JavaScript best practice.
-JavaScript programs typically work with the real number primitives, rather than objects that "look like" numbers.
-Primitives are simpler to conceptualize and work with `==` and `===` equality checks -- which their object equivalents do notDeepEqual.
-As a result, using the lowercase type names like `number` instead of the uppercase names like `Number` helps make your code behave more reliably.
+TypeScript defines several confusing pairs of types that look very similar to each other, but actually mean quite different things: `boolean`/`Boolean`, `number`/`Number`, `string`/`String`, `bigint`/`BigInt`, `symbol`/`Symbol`, `object`/`Object`.
```
JoshuaKGoldberg, line 12: Nit: _"quite"_ is subjective and doesn't add any meaning not described later on:

```suggestion
TypeScript defines several confusing pairs of types that look very similar to each other, but actually mean different things: `boolean`/`Boolean`, `number`/`Number`, `string`/`String`, `bigint`/`BigInt`, `symbol`/`Symbol`, `object`/`Object`.
```

JoshuaKGoldberg (approved):
Looks great to me! Just the one nit, not a blocker if you disagree. Thanks! 🎉

Josh-Cena (approved):
Love this!

Issue #994, opened 2019-09-20
Feature request - rule for await in async try catch return


drabinowitz:
A `Rejected Promise` returned by an `async` function will not be caught by the function's `try-catch` block. This may surprise some users and lead to unexpected behavior:
```typescript
(async () => {
  try {
    // will result in an UnhandledPromiseRejection
    return Promise.reject('Error!');
  } catch (e) {
    // will not run
    handleCleanup();
  }
})();
```

conversely, `await`-ed promises work in a more intuitive fashion:
```typescript
(async () => {
  try {
    // will not result in an UnhandledPromiseRejection
    return await Promise.reject('Error!');
  } catch (e) {
    // will run
    handleCleanup();
  }
})();
```

I understand why this is the case according to the `async-await` function spec, but it feels like confusing behavior and a problem that `typescript-eslint` could help avoid via the following rule:

disallow `return`-ing a `Promise` from inside an `async` function `try-catch` block.

Thus, the first case described above would be invalid according to this rule, but the second case would be valid. We could even, potentially, provide an auto-fix that would add the `await` statement to the promise.

There's also the possibility for the broader rule of disallowing`return`-ing a `Promise` from anywhere inside an `async` function, but I wanted to start with the narrowest case.

Please let me know if there's any other clarifying information I can provide, thanks so much for your time!

drabinowitz:
Yes i think that'd make more sense @ark120202. @bradzacher what are you thoughts on that change should I update the feature request?

bradzacher:
To clarify then,

The idea would be to have a rule (`return-await`) which bans directly returning await statements (`return await foo`) (i.e. requires no type info), except in the case that you're returning a promise from within a try/catch, in which case you want to enforce that you return an await (i.e. requires type info).

Seems like a good candidate for this plugin.

```
declare function returnsPromise(): Promise<string>;
declare function handleException(e: Error): string;

// option: 'in-try-catch'
async function myFunc1(): Promise<string> {
  try {
    return returnsPromise(); // lint error - should await the promise so errors are handled by the surrounding try/catch
  } catch (e) {}
  return Promise.resolve('foo'); // no error
}

// option: 'never'
async function myFunc1(): Promise<string> {
  try {
    return await returnsPromise(); // lint error - should never return await
  } catch (e) {}
  return await Promise.resolve('foo'); // lint error 
}

// option: 'always'
async function myFunc1(): Promise<string> {
  try {
[...]

PR #9364, opened 2024-06-15
enhancement(eslint-plugin): [return-await] add option to report in error-handling scenarios only

### LINKED ISSUE #9030: Enhancement: [return-await] Option for unopinionated "in-try-catch"

https://typescript-eslint.io/rules/return-await

### Description

The current `in-try-catch` option handles two separate concerns 
- _Correctness_ - requiring `await` where it impacts control flow (due to `try-catch-finally`)
- _Stylistic_ - _forbidding_ `await` elsewhere

I would propose for there to be a rule option that handles only the correctness portion. That is to say, it should enforce `await` wherever in-try-catch currently does, but not report where `in-try-catch` forbids `await`.

### Fail

```TypeScript
// same as in-try-catch
async function f() {
  try {
    return returnsPromise();
  } catch { }
}
```

### Pass

```TypeScript
async function f() {
   return returnsPromise();
}

async function f() {
   return await returnsPromise();
}

// same as in-try-catch
async function f() {
  try {
    return await returnsPromise();
  } catch { }
}
```

### Additional Info

I think this would be valuable in order to be able to recommend a less disruptive version of the rule in a preset, see https://github.com/typescript-eslint/typescript-eslint/issues/8667

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9030 
- [x] That issue was marked as ~~[accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)~~ "team assigned"

## Overview

Just adds an option to ignore return-await concerns outside of scenarios relevant for error-handling

### DIFF
--- a/packages/eslint-plugin/docs/rules/return-await.mdx
+++ b/packages/eslint-plugin/docs/rules/return-await.mdx
@@ -203,6 +203,35 @@ async function validAlways3() {
 </TabItem>
 </Tabs>
 
+### `error-handling-correctness-only`
+
+Same as `in-try-catch`, except _only_ enforces Promises be awaited within the
+error handling contexts explained under that option. Outside of error-handling
+contexts, returned Promises are ignored, whether awaited or not.
+
+:::warning
+We recommend you configure either `in-try-catch` or `always` instead of this option.
+While the choice or awaiting or not awaiting promises outside of error-handling contexts is mostly stylistic, it is best to be consistent.
+:::
+
+Examples of additional correct code with `error-handling-correctness-only`:
+
+<Tabs>
+<TabItem value="✅ Correct">
+
+```ts option='"error-handling-correctness-only"'
+async function asyncFunction(): Promise<void> {
+  if (Math.random() < 0.5) {
+    return await Promise.resolve();
+  } else {
+    return Promise.resolve();
+  }
+}
+```
+
+</TabItem>
+</Tabs>
+
 ### `never`
 
 Disallows all `await`ing any returned promises.
--- a/packages/eslint-plugin/src/rules/return-await.ts
+++ b/packages/eslint-plugin/src/rules/return-await.ts
@@ -50,7 +50,12 @@ export default createRule({
     schema: [
       {
         type: 'string',
-        enum: ['in-try-catch', 'always', 'never'],
+        enum: [
+          'in-try-catch',
+          'always',
+          'never',
+          'error-handling-correctness-only',
+        ],
       },
     ],
   },
@@ -283,29 +288,53 @@ export default createRule({
         return;
       }
 
+      if (option === 'error-handling-correctness-only') {
+        if (!isAwait && affectsErrorHandling) {
+          context.report({
+            messageId: 'requiredPromiseAwait',
+            node,
+            // unconditional since always impacts error handling
+            suggest: [
+              {
+                messageId: 'requiredPromiseAwaitSuggestion',
+                fix: fixer =>
+                  insertAwait(
+                    fixer,
+                    node,
+                    isHigherPrecedenceThanAwait(expression),
+                  ),
+              },
+            ],
+          });
+
+          return;
+        }
+      }
+
       if (option === 'in-try-catch') {
         if (isAwait && !affectsErrorHandling) {
           context.report({
             messageId: 'disallowedPromiseAwait',
             node,
-            ...fixOrSuggest(useAutoFix, {
-              messageId: 'disallowedPromiseAwaitSuggestion',
-              fix: fixer => removeAwait(fixer, node),
-            }),
+            // unconditional since never impacts error handling
+            fix: fixer => removeAwait(fixer, node),
           });
         } else if (!isAwait && affectsErrorHandling) {
           context.report({
             messageId: 'requiredPromiseAwait',
             node,
-            ...fixOrSuggest(useAutoFix, {
-              messageId: 'requiredPromiseAwaitSuggestion',
-              fix: fixer =>
-                insertAwait(
-                  fixer,
-                  node,
-                  isHigherPrecedenceThanAwait(expression),
-                ),
-            }),
+            // unconditional since always impacts error handling
+            suggest: [
+              {
+                messageId: 'requiredPromiseAwaitSuggestion',
+                fix: fixer =>
+                  insertAwait(
+                    fixer,
+                    node,
+                    isHigherPrecedenceThanAwait(expression),
+                  ),
+              },
+            ],
           });
         }
 
--- a/packages/eslint-plugin/tests/rules/return-await.test.ts
+++ b/packages/eslint-plugin/tests/rules/return-await.test.ts
@@ -2,6 +2,7 @@ import { noFormat, RuleTester } from '@typescript-eslint/rule-tester';
 
 import rule from '../../src/rules/return-await';
 import { getFixturesRootDir } from '../RuleTester';
+import { InvalidTestCase } from '@typescript-eslint/utils/ts-eslint';
 
 const rootDir = getFixturesRootDir();
 
@@ -56,6 +57,25 @@ ruleTester.run('return-await', rule, {
         }
       }
     `,
+    {
+      code: `
+        async function test() {
+          if (Math.random() < 0.33) {
+            return await Promise.resolve(1);
+          } else if (Math.random () < 0.5) {
+            return Promise.resolve(2);
+          }
+
+          try {
+          } catch (e) {
+            return await Promise.resolve(3);
+          } finally {
+            console.log('cleanup');
+          }
+        }
+      `,
+      options: ['error-handling-correctness-only'],
+    },
     `
       async function test() {
         try {
@@ -435,8 +455,11 @@ async function test() {
         },
       ],
     },
-    {
-      code: `
+
+    ...['error-handling-correctness-only', 'always', 'in-try-catch'].map(
+      option =>
+        ({
+          code: `
         async function test() {
           try {
             return Promise.resolve(1);
@@ -447,15 +470,15 @@ async function test() {
           }
         }
       `,
-      output: null,
-      errors: [
-        {
-          line: 4,
-          messageId: 'requiredPromiseAwait',
-          suggestions: [
+          output: null,
+          errors: [
             {
-              messageId: 'requiredPromiseAwaitSuggestion',
-              output: `
+              line: 4,
+              messageId: 'requiredPromiseAwait',
+              suggestions: [
+                {
+                  messageId: 'requiredPromiseAwaitSuggestion',
+                  output: `
         async function test() {
           try {
             return await Promise.resolve(1);
@@ -466,16 +489,16 @@ async function test() {
           }
         }
       `,
+                },
+              ],
             },
-          ],
-        },
-        {
-          line: 6,
-          messageId: 'requiredPromiseAwait',
-          suggestions: [
             {
-              messageId: 'requiredPromiseAwaitSuggestion',
-              output: `
+              line: 6,
+              messageId: 'requiredPromiseAwait',
+              suggestions: [
+                {
+                  messageId: 'requiredPromiseAwaitSuggestion',
+                  output: `
         async function test() {
           try {
             return Promise.resolve(1);
@@ -486,11 +509,17 @@ async function test() {
           }
         }
       `,
+                },
+              ],
             },
           ],
-        },
-      ],
-    },
+          options: [option],
+        }) satisfies InvalidTestCase<
+          'requiredPromiseAwait' | 'requiredPromiseAwaitSuggestion',
+          [string]
+        >,
+    ),
+
     {
       code: `
         async function test() {
@@ -554,63 +583,6 @@ async function test() {
       options: ['in-try-catch'],
       code: `
         async function test() {
-          try {
-            return Promise.resolve(1);
-          } catch (e) {
-            return Promise.resolve(2);
-          } finally {
-            console.log('cleanup');
-          }
-        }
-      `,
-      output: null,
-      errors: [
-        {
-          line: 4,
-          messageId: 'requiredPromiseAwait',
-          suggestions: [
-            {
-              messageId: 'requiredPromiseAwaitSuggestion',
-              output: `
-        async function test() {
-          try {
-            return await Promise.resolve(1);
-          } catch (e) {
-            return Promise.resolve(2);
-          } finally {
-            console.log('cleanup');
-          }
-        }
-      `,
-            },
-          ],
-        },
-        {
-          line: 6,
-          messageId: 'requiredPromiseAwait',
-          suggestions: [
-            {
-              messageId: 'requiredPromiseAwaitSuggestion',
-              output: `
-        async function test() {
-          try {
-            return Promise.resolve(1);
-          } catch (e) {
-            return await Promise.resolve(2);
-          } finally {
-            console.log('cleanup');
-          }
-        }
-      `,
-            },
-          ],
-        },
-      ],
-    },
-    {
-      options: ['in-try-catch'],
-      code: `
-        async function test() {
           return await Promise.resolve(1);
         }
       `,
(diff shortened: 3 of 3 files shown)


packages/eslint-plugin/docs/rules/return-await.mdx
JoshuaKGoldberg, line 1: > targeted against main for now

+1 from me on landing this in main. We need to frequently merge into v8 anyway. This way, users in both branches get the option sooner.

packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -283,29 +288,53 @@ export default createRule({
         return;
       }
 
+      if (option === 'error-handling-correctness-only') {
+        if (!isAwait && affectsErrorHandling) {
+          context.report({
+            messageId: 'requiredPromiseAwait',
+            node,
+            // unconditional since always impacts error handling
```
JoshuaKGoldberg, line 296: [Docs] Nitty nit: a bit more clear to read?

```suggestion
            // We always report here, as it always impacts error handling
```

JoshuaKGoldberg (approved):
Detective Harris from South Park saying "NICCCE"

packages/eslint-plugin/src/rules/return-await.ts
kirkwaiblinger: Made some further tweaks

packages/eslint-plugin/src/rules/return-await.ts
kirkwaiblinger, line 1: with the addition of yet another option, this code is becoming less and less DRY 🙁 

Basically it just comes down to 

| config | returned promise in error handling context (no autofix) | returned promise in ordinary context (autofix) |
| --- | --- | --- |
| always | `return await promise` | `return await promise` | 
| in-try-catch | `return await promise` | `return promise` | 
| error-handling-correctness-only | `return await promise`  |  don't care 🤷  | 
| never | `return promise` | `return promise` |

Open to suggestions on how to indicate this DRYly in code.

kirkwaiblinger: Also, it occurs to me that having a table like this might be useful in the docs 🤔

JoshuaKGoldberg: 💡 YES!

kirkwaiblinger: Ok, made lots of changes in this area (both code and docs)

packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -271,92 +314,69 @@ export default createRule({
       const type = checker.getTypeAtLocation(child);
       const isThenable = tsutils.isThenableType(checker, expression, type);
```
kirkwaiblinger, line 315: github's diffing gets real confused about what if you don't use the "ignore whitespace" here. Be sure to use that.

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -214,6 +218,35 @@ async function validAlways3() {
...
 
+### `error-handling-correctness-only`
+
+Same as `in-try-catch`, except _only_ enforces Promises be awaited within the
+error handling contexts explained under that option. Outside of error-handling
+contexts, returned Promises are ignored, whether awaited or not.
+
+:::warning
+We recommend you configure either `in-try-catch` or `always` instead of this option.
+While the choice or awaiting or not awaiting promises outside of error-handling contexts is mostly stylistic, it is best to be consistent.
```
JoshuaKGoldberg, line 229: [Docs] Grammar/typo fix, and a bit of clarity:

```suggestion
While the choice of whether to awaiting promises outside of error-handling contexts is mostly stylistic, it's generally best to be consistent.
```

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -260,3 +293,12 @@ async function validNever3() {
...
+| :-----------------------------: | :-------------------------------------------: | :-------------------------------------------------------------: | :--------------------------------------------------------: |
+|             always              |            `return await promise;`            |                     `return await promise;`                     |                          ✅ Yes!                           |
+|          in-try-catch           |               `return promise;`               |                     `return await promise;`                     |                          ✅ Yes!                           |
+| error-handling-correctness-only |                 don't care 🤷                 |                     `return await promise;`                     | 🟡 Okay to use, but the above options would be preferable. |
+|              never              |               `return promise;`               |                        `return promise;`                        |                           ❌ No!                           |
```
JoshuaKGoldberg, line 304: [Docs]

> ❌ No!

Well, if the option should _never_ be used... why would we want to include it at all? It's not clear from this table or the <code>## \`never\`</code> section why we have that. Maybe a solution to this would be to mention that `never` is only included for compatibility with the base rule?

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -260,3 +293,12 @@ async function validNever3() {
 
 </TabItem>
 </Tabs>
+
+### Options Summary
+
+|             Option              | Ordinary Context <br/> (stylistic preference) | Error Handling Context <br/> (catches subtle control flow bugs) |                       Should I use?                        |
```
JoshuaKGoldberg, line 299: [Docs] I think this table is pretty useful. Proposal: how about moving this to within the `## Options` early on, to prime folks with it before they dive into the options?

I also think it would be useful to mention _in each option's section_ the pros & cons of that option. Right now, the `always` section just says _"Requires that all returned promises are `await`ed."_. It doesn't mention the option being preferred, or why it's preferred.

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -214,6 +218,35 @@ async function validAlways3() {
 </TabItem>
 </Tabs>
 
+### `error-handling-correctness-only`
```
JoshuaKGoldberg, line 221: [Refactor] I think the table being useful is a symptom of the options being kind of unclear. What do you think about switching this to be a `boolean` option people have to enable with an options object?

```js
"@typescript-eslint/return-await": [
  "error",
  "always",
  { errorHandlingCorrectnessOnly: true }
]
```

Since we'd rather folks use `'always'` or `'in-try-catch'` anyway, I don't mind the mild user pain of an extra object. And I think it'd make the implementation simpler too!

JoshuaKGoldberg (changes requested):
Thinking on the docs 🤔

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: Documenting a private conversation for visibility:

***

> `never` is only included for compatibility with the base rule

There is no evidence that this is or ever was the case; see https://github.com/typescript-eslint/typescript-eslint/issues/994.

***

Plan to resolve this comment - tentative plan to deprecate 'never' as part of this PR. Getting that process started imminently.

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: > Proposal: how about moving this to within the `## Options` early on, to prime folks with it before they dive into the options?

Yeah, that was actually my first approach, but i was waffling on how to go about defining an error-handling context since that's required before the table... and I went with this approach out of laziness. I agree that putting it first would be better, though, so I'll give it another stab 🙂 

> I also think it would be useful to mention _in each option's section_ the pros & cons of that option. Right now, the `always` section just says _"Requires that all returned promises are `await`ed."_. It doesn't mention the option being preferred, or why it's preferred.

Feels like we're venturing into scope creep territory, but at the same time, we probably don't want this option to go out without good supporting docs 🤔 Will try to incorporate this 👍

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: Documenting private conversation - Leaning against this, on the assumption for now that `never` is on its way out (a la https://github.com/typescript-eslint/typescript-eslint/pull/9364#discussion_r1651963801), in which case the flat options of (always|in-try-catch|error-handling-correctness-only) do make the most sense

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: filed https://github.com/typescript-eslint/typescript-eslint/issues/9433

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -214,6 +218,35 @@ async function validAlways3() {
...
+
+Same as `in-try-catch`, except _only_ enforces Promises be awaited within the
+error handling contexts explained under that option. Outside of error-handling
+contexts, returned Promises are ignored, whether awaited or not.
```
Josh-Cena, line 225: Don't wrap lines by line width

phaux:
Maybe change the option so that instead of a string it's an object: 

```ts
{
  inTryCatch?: "always" | "never" | null, // undefined or null means "don't care"
  other?: "always" | "never" | null,
}
```

And make the string options aliases for backward compatibility:

- "always" - `{ inTryCatch: "always", other: "always" }`
- "in-try-catch" - `{ inTryCatch: "always", other: "never" }`
- "never" - `{ inTryCatch: "never", other: "never" }`

The new option would be `{ inTryCatch: "always", other: null }`

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger, line 1: @phaux Threading https://github.com/typescript-eslint/typescript-eslint/pull/9364#issuecomment-2211750536

> Maybe change the option so that instead of a string it's an object:
> 
> ```ts
> {
>   default: "always" | "never" | "ignore", 
>   inTryCatch?: "always" | "never" | "ignore", // undefined means "same as default" 
> }
> ```
> 
> And make the string options aliases for backward compatibility:
> 
> * "always" - `{ default: "always" }`
> * "in-try-catch" - `{ default: "never", inTryCatch: "always" }`
> * "never" - `{ default: "never" }`
> 
> The new option would be `{ default: "ignore", inTryCatch: "always" }`
> 
> In the future, there could be more conditions added. For example, someone might want to use "always" but make an exception for single-line arrow functions to make them less noisy:
> 
> ```ts
> /* eslint @typescript-eslint/promise-function-async: "error" */
> /* eslint @typescript-eslint/return-await: ["error", { default: "always", inArrowFunction: "ignore" }] */
> 
> promise.then(async (v) => doSomethingAsync(v)) // allowed
> ```

*** 

I think that the options restructure is a potentially a good idea (and +1 to making it non-breaking by providing aliases!) (related - https://github.com/typescript-eslint/typescript-eslint/issues/9030#issuecomment-2083721143). But, it's IMO it's unnecessary, and therefore out of scope for this PR. Would you kindly file a separate issue to propose this? Thanks! 🙂

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: Incorporated #9433 into this PR

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: I made changes in this area. I definitely think the result is much better! Thanks for making the suggestions.

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: Huh. not sure how that happened. In any case this has been fixed/superseded 👍

Josh-Cena:
I thought about the object API (which I mentioned in the linked issue), but I now don't like it. It creates a lot of "dubious combinations", for example you would never want `{ default: "always", inTryCatch: "never" }`.

> In the future, there could be more conditions added. For example, someone might want to use "always" but make an exception for single-line arrow functions to make them less noisy:

You should either not mark it as async (which is even less noisy), or add the await. I don't think we should accommodate that style as that's not really the spirit of this rule.

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: FYI @Josh-Cena, re https://github.com/typescript-eslint/typescript-eslint/pull/9364#issuecomment-2224210061, hoping to thread convo here. But regardless, I think this discussion should be deferred to a separate issue if anyone wants to propose it.

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -21,23 +21,43 @@ The extended rule is named `return-await` instead of `no-return-await` because t
...
+| :-----------------------------: | :----------------------------------------------: | :----------------------------------------------------------: | :--------------------------------------------------------: |
+|             always              |             `return await promise;`              |                   `return await promise;`                    |                          ✅ Yes!                           |
+|          in-try-catch           |                `return promise;`                 |                   `return await promise;`                    |                          ✅ Yes!                           |
+| error-handling-correctness-only |                  don't care 🤷                   |                   `return await promise;`                    | 🟡 Okay to use, but the above options would be preferable. |
+|              never              |                `return promise;`                 | `return promise;` <br/> (⚠️ This behavior may be harmful ⚠️) |             ❌ No. This option is deprecated.              |
+
```
JoshuaKGoldberg, line 52: [Praise] Love this.

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -21,23 +21,43 @@ The extended rule is named `return-await` instead of `no-return-await` because t
...
+| :-----------------------------: | :----------------------------------------------: | :----------------------------------------------------------: | :--------------------------------------------------------: |
+|             always              |             `return await promise;`              |                   `return await promise;`                    |                          ✅ Yes!                           |
+|          in-try-catch           |                `return promise;`                 |                   `return await promise;`                    |                          ✅ Yes!                           |
+| error-handling-correctness-only |                  don't care 🤷                   |                   `return await promise;`                    | 🟡 Okay to use, but the above options would be preferable. |
+|              never              |                `return promise;`                 | `return promise;` <br/> (⚠️ This behavior may be harmful ⚠️) |             ❌ No. This option is deprecated.              |
```
JoshuaKGoldberg, line 51: [Docs] Since this is a code thing, maybe?

```suggestion
|              `never`              |                `return promise;`                 | `return promise;` <br/> (⚠️ This behavior may be harmful ⚠️) |             ❌ No. This option is deprecated.              |
```

(& the others)

JoshuaKGoldberg (dismissed):
Whoohoo! Happy to finally have this thing modernized! 👏 🔥

JoshuaKGoldberg (approved):

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -21,23 +21,43 @@ The extended rule is named `return-await` instead of `no-return-await` because t
...
+The options in this rule distinguish between "ordinary contexts" and "error-handling contexts".
+An error-handling context is anywhere where returning an unawaited promise would cause unexpected control flow regarding exceptions/rejections.
+See detailed examples in the sections for each option.
+
+- If you return a promise within a `try` block, it should be awaited in order to trigger subsequent `catch` or `finally` blocks as expected.
+- If you return a promise within a `catch` block, and there _is_ a `finally` block, it should be awaited in order to trigger the `finally` block as expected.
+- If you return a promise between a `using` or `await using` declaration and the end of its scope, it should be awaited, since it behaves equivalently to code wrapped in a `try` block followed by a `finally`.
+
+Ordinary contexts are anywhere else a promise may be returned.
+The choice of whether to await a returned promise in an ordinary context is mostly stylistic.
```
Josh-Cena, line 42:
```suggestion
The choice of whether to await a returned promise in an ordinary context is mostly stylistic, except for [stack trace benefits](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#improving_stack_trace).
```

packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -62,6 +73,38 @@ export default createRule({
...
+            ordinaryContext: "don't-care",
+            errorHandlingContext: 'await',
+          };
+        case 'in-try-catch':
+          return {
+            ordinaryContext: 'no-await',
+            errorHandlingContext: 'await',
+          };
+      }
+    }
```
Josh-Cena, line 106: Nit: move these outside of the rule creator to top level?

Josh-Cena (approved):
Two optional nits; otherwise 🛳

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: Hmm. The docs page already mentions this and includes the link at the top of the page, so I'm averse to repeating it here. My intention here was to use "mostly" with the context that the reader has already been told this information.

I think that, beyond this specific line, there's still more room for improvement in coherence of the docs anyway, so are you all right with that being addressed in a followup (perhaps as part of https://github.com/typescript-eslint/typescript-eslint/issues/9517?)

Josh-Cena: Ah, I see. Well ignore this then.

Josh-Cena (approved):

PR #9369, opened 2024-06-17
feat(eslint-plugin): [no-misused-object-like] add rule

### LINKED ISSUE #6807: Rule proposal: Prevent `Object.values(...)` usage with `Map`

### Description

Using `Object.values(...)`, `Object.keys(...)`, `Object.entries(...)` with a `Map` object is a foot-gun because it will return empty `[]` regardless of the map contents and is most likely a mistake. These sorts of mistakes are easy to get into during refactors. This issue is spawning from a real-life refactor mistake -> https://github.com/vector-im/element-web/issues/24995#issuecomment-1490776260

Related StackOverflow question: https://stackoverflow.com/questions/72315392/how-to-prevent-the-mistake-of-calling-object-values-on-a-map

As @bergus mentions from the SO question, it's possible to extend `Map` to make a subclass that returns something with the object utilities but this seems a lot more niche compared to the big problem of running them on base `Map` objects. AFAICT, it's not possible to use TypeScript on its own to catch this sort of thing because practically everything inherits from `Object` and function arguments in TypeScript are contravariant (accepts supertypes but doesn't accept subtypes).

### Fail Cases

```typescript
const map = new Map([[ 1, 'one' ],[ 2, 'two' ]]);

// Rule would error for each of these usages, as using these object utilities
// on a Map will just return empty results and is probably a mistake
Object.keys(map); // -> []
Object.values(map); // -> []
Object.entries(map); // -> []
```

### Pass Cases

```typescript
const myObject = { 1: 'one', 2: 'two' };
Object.keys(myObject); // -> ['1', '2']
Object.values(myObject); // -> ['one', 'two']
Object.entries(myObject); // -> [[ 1, 'one' ],[ 2, 'two' ]]

const map = new Map([[ 1, 'one' ],[ 2, 'two' ]]);
map.keys(); // -> ['1', '2']
map.values(); // -> ['one', 'two']
map.entries(); // -> [[ 1, 'one' ],[ 2, 'two' ]]
```

### Additional Info

_No response_

gibson:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6807

## Overview

This MR will close the problem with usage `Object.keys()`, `Object.values()` and `Object.entries()` for object like classes `Map` and `Set`. On this classes that methods always returns empty array instead of any meaningful data and doesn't cause and warnings or errors

### DIFF
--- a/packages/eslint-plugin/src/rules/no-misused-object-likes.ts
+++ b/packages/eslint-plugin/src/rules/no-misused-object-likes.ts
@@ -1,0 +1,149 @@
+import type { Identifier, MemberExpression } from '@typescript-eslint/ast-spec';
+import type { TSESTree } from '@typescript-eslint/utils';
+
+import { createRule, getParserServices } from '../util';
+
+export type Options = [
+  {
+    checkObjectKeysForMap?: boolean;
+    checkObjectValuesForMap?: boolean;
+    checkObjectEntriesForMap?: boolean;
+    checkObjectKeysForSet?: boolean;
+    checkObjectValuesForSet?: boolean;
+    checkObjectEntriesForSet?: boolean;
+  },
+];
+export type MessageIds =
+  | 'objectKeysForMap'
+  | 'objectValuesForMap'
+  | 'objectEntriesForMap'
+  | 'objectKeysForSet'
+  | 'objectValuesForSet'
+  | 'objectEntriesForSet';
+
+export default createRule<Options, MessageIds>({
+  name: 'no-misused-object-likes',
+  defaultOptions: [
+    {
+      checkObjectKeysForMap: true,
+      checkObjectValuesForMap: true,
+      checkObjectEntriesForMap: true,
+      checkObjectKeysForSet: true,
+      checkObjectValuesForSet: true,
+      checkObjectEntriesForSet: true,
+    },
+  ],
+
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Enforce check `Object.values(...)`, `Object.keys(...)`, `Object.entries(...)` usage with Map/Set objects',
+      requiresTypeChecking: false,
+    },
+    messages: {
+      objectKeysForMap: "Don't use `Object.keys()` for Map objects",
+      objectValuesForMap: "Don't use `Object.values()` for Map objects",
+      objectEntriesForMap: "Don't use `Object.entries()` for Map objects",
+      objectKeysForSet: "Don't use `Object.keys()` for Set",
+      objectValuesForSet: "Don't use `Object.values()` for Set",
+      objectEntriesForSet: "Don't use `Object.entries()` for Set",
+    },
+    schema: [
+      {
+        type: 'object',
+        additionalProperties: false,
+        properties: {
+          checkObjectKeysForMap: {
+            description: 'Check usage Object.keys for Map object',
+            type: 'boolean',
+          },
+          checkObjectValuesForMap: {
+            description: 'Check usage Object.values for Map object',
+            type: 'boolean',
+          },
+          checkObjectEntriesForMap: {
+            description: 'Check usage Object.entries for Map object',
+            type: 'boolean',
+          },
+          checkObjectKeysForSet: {
+            description: 'Check usage Object.keys for Set object',
+            type: 'boolean',
+          },
+          checkObjectValuesForSet: {
+            description: 'Check usage Object.values for Set object',
+            type: 'boolean',
+          },
+          checkObjectEntriesForSet: {
+            description: 'Check usage Object.entries for Set object',
+            type: 'boolean',
+          },
+        },
+      },
+    ],
+  },
+
+  create(context, [options]) {
+    const services = getParserServices(context);
+
+    function checkObjectMethodCall(
+      callExpression: TSESTree.CallExpression,
+    ): void {
+      const argument = callExpression.arguments[0];
+      const type = services.getTypeAtLocation(argument);
+      const argumentTypeName = type.getSymbol()?.name;
+      const callee = callExpression.callee as MemberExpression;
+      const objectMethod = (callee.property as Identifier).name;
+
+      if (argumentTypeName === 'Map') {
+        if (objectMethod === 'keys' && options.checkObjectKeysForMap) {
+          context.report({
+            node: callExpression,
+            messageId: 'objectKeysForMap',
+          });
+        }
+        if (objectMethod === 'values' && options.checkObjectValuesForMap) {
+          context.report({
+            node: callExpression,
+            messageId: 'objectValuesForMap',
+          });
+        }
+        if (objectMethod === 'entries' && options.checkObjectEntriesForMap) {
+          context.report({
+            node: callExpression,
+            messageId: 'objectEntriesForMap',
+          });
+        }
+      }
+      if (argumentTypeName === 'Set') {
+        if (objectMethod === 'keys' && options.checkObjectKeysForSet) {
+          context.report({
+            node: callExpression,
+            messageId: 'objectKeysForSet',
+          });
+        }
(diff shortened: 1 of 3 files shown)


SimonSchick:
This is missing support for `ReadonlyMap` `ReadonlySet`, `WeakMap` and `WeakSet` (and possibly others) and will likely not work with inheritance?

It may be worth-while to check if the subject of the check has any `.get(arg)` and `.set(arg1, arg2)` methods to catch issues with `Headers` and friends as well, this may be slow however.

image


Overall this is a good PR, it might also be good to integrate https://github.com/typescript-eslint/typescript-eslint/issues/5677 into this and generalize this rule as `no-misused-map-like-usage`?

packages/eslint-plugin/src/rules/no-misused-object-likes.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Enforce check `Object.values(...)`, `Object.keys(...)`, `Object.entries(...)` usage with Map/Set objects',
+      requiresTypeChecking: false,
+    },
+    messages: {
+      objectKeysForMap: "Don't use `Object.keys()` for Map objects",
```
SimonSchick, line 45: I think this may be overly specific, granularity for `SetLike` and `MapLike` is likely sufficient since it would be weird to allow calling `Object.values(` on a Map but banning `Object.keys`

packages/eslint-plugin/src/rules/no-misused-object-likes.ts
```diff
@@ -0,0 +1,149 @@
...
+      checkObjectEntriesForSet: true,
+    },
+  ],
+
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Enforce check `Object.values(...)`, `Object.keys(...)`, `Object.entries(...)` usage with Map/Set objects',
+      requiresTypeChecking: false,
```
SimonSchick, line 42: Pretty sure this is required?

packages/eslint-plugin/src/rules/no-misused-object-likes.ts
```diff
@@ -0,0 +1,149 @@
...
+          context.report({
+            node: callExpression,
+            messageId: 'objectEntriesForSet',
+          });
+        }
+      }
+    }
+
+    return {
+      'CallExpression[callee.object.name=Object][callee.property.name=keys][arguments.length=1]':
```
SimonSchick, line 141: This is missing various other ones such as `Object.assign/groupBy/hasOwn?`, it might be better to move this check into the `checkObjectMethodCall` function and have a list of banned methods there.

SimonSchick (changes requested):
This would be extremely useful when migrating from object dictionary use to map/set use!

packages/eslint-plugin/src/rules/no-misused-object-likes.ts
```diff
@@ -0,0 +1,149 @@
...
+export type Options = [
+  {
+    checkObjectKeysForMap?: boolean;
+    checkObjectValuesForMap?: boolean;
+    checkObjectEntriesForMap?: boolean;
+    checkObjectKeysForSet?: boolean;
+    checkObjectValuesForSet?: boolean;
+    checkObjectEntriesForSet?: boolean;
+  },
+];
```
JoshuaKGoldberg, line 15: [Question] Why allow such a granular set of options? Is there a known use case where someone would want to skip, say, only `Object.values`?

packages/eslint-plugin/src/rules/no-misused-object-likes.ts
```diff
@@ -0,0 +1,149 @@
...
+    function checkObjectMethodCall(
+      callExpression: TSESTree.CallExpression,
+    ): void {
+      const argument = callExpression.arguments[0];
+      const type = services.getTypeAtLocation(argument);
+      const argumentTypeName = type.getSymbol()?.name;
+      const callee = callExpression.callee as MemberExpression;
+      const objectMethod = (callee.property as Identifier).name;
+
+      if (argumentTypeName === 'Map') {
```
JoshuaKGoldberg, line 98: [Bug] If a user defines their own `Map` class, this will falsely report:

```ts
class Map {
  value = 1;
}

const map = new Map();
Object.values(map);
```

packages/eslint-plugin/src/rules/no-misused-object-likes.ts
```diff
@@ -0,0 +1,149 @@
...
+    const services = getParserServices(context);
+
+    function checkObjectMethodCall(
+      callExpression: TSESTree.CallExpression,
+    ): void {
+      const argument = callExpression.arguments[0];
+      const type = services.getTypeAtLocation(argument);
+      const argumentTypeName = type.getSymbol()?.name;
+      const callee = callExpression.callee as MemberExpression;
+      const objectMethod = (callee.property as Identifier).name;
```
JoshuaKGoldberg, line 96: [Bug] Using an `as` assertion on AST nodes is generally a sign of either:

* 🍎 The node is known to be a more specific type than what you've told the type system
* 🍌 The code isn't properly handling edge cases

The `CallExpression[callee.object.name=Object][callee.property.name=keys][arguments.length=1]` below makes me think it's 🍎. Which means you'd want to use a type for the node.

```ts
    type CallExpressionWithStuff = TSESTree.CallExpression & {
      callee: {
        object: {
          name: 'Object';
      // ...
```

packages/eslint-plugin/src/rules/no-misused-object-likes.ts
JoshuaKGoldberg, line 141: > `object.name=Object`

[Bug] False report case:

```ts
const Object = { keys: (value) => [value] };

export const keys = Object.keys(new Map());
```

packages/eslint-plugin/tests/rules/no-misused-object-likes.test.ts
JoshuaKGoldberg, line 1: [Testing] Some missing test cases:

* Intersections and unions: with `Map`, `Set`, and/or other types
* Readonlies: `ReadonlyMap`, `ReadonlySet`

packages/eslint-plugin/src/rules/no-misused-object-likes.ts
```diff
@@ -0,0 +1,149 @@
...
+        if (objectMethod === 'entries' && options.checkObjectEntriesForSet) {
+          context.report({
+            node: callExpression,
+            messageId: 'objectEntriesForSet',
+          });
+        }
+      }
+    }
+
+    return {
```
JoshuaKGoldberg, line 140: [Bug] Looking at built-in global objects in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects#keyed_collections, I think at least `WeakMap` and `WeakSet` should also be handled by this rule.

JoshuaKGoldberg (changes requested):
A good start! There are some functional gaps that'll need to be filled in around more object support and resiliency to user shenanigans. Excited to see this - thanks for getting it started! 🙌

SimonSchick:
@gibson are you interested in continuing work on this? Otherwise I may be inclined to take a stab at this myself.

gibson:
@SimonSchick will continue in closes time

JoshuaKGoldberg:
👋 @gibson just checking in, do you still plan on working on this?

JoshuaKGoldberg:
Closing this PR as it's been stale for ~6 weeks without activity. Feel free to reopen @gibson if you have time - but no worries if not! If anybody wants to drive it forward, please do post your own PR - and if you use this as a start, consider adding `Co-authored-by: @gibson` at the end of your PR description. Thanks! 😊

SimonSchick:
I might pick this up, I ran into some scenarious/migrations where this rule would've been incredibly valuable.

PR #9380, opened 2024-06-17
fix(eslint-plugin): [prefer-nullish-coalescing] ensure ternary fix does not remove parens

### LINKED ISSUE #9379: Bug: [prefer-nullish-coalescing] quick fix for ternary removes required parens

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.5&fileType=.tsx&code=CYUwxgNghgTiAEEQBd4DcoQK4gFzwGdkYBLAOwHN4AfeLM0AM3JGAG4BYAKG8frGQkA9mXgiwIABRp8RUpQCU8AN7d48OMiwxRGbAgCEAXiN0GIZmVbwA-Okw54%2BaQ4Sm0C7gF8gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6KL2jjokcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
declare let value: string | undefined;

function once(v: string) {
  return value !== undefined ? value : (value = v)
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

Fix is:

```
declare let value: string | undefined;

function once(v: string) {
  return value ?? (value = v)
}
```

Or honestly, 

```
declare let value: string | undefined;

function once(v: string) {
  return value ??= v
}
```

Except that this rule doesn't understand `??=` at all, as far as I can tell (another issue).

### Actual Result

Fix is:

```
declare let value: string | undefined;

function once(v: string) {
  return value ?? value = v
}
```

### Additional Info

_No response_

jakebailey:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9379

## Overview

Parens aren't in the AST, and this fixer's text manipulation did not handle that.

I took the time to move a helper out of another lint rule into a common place. This pattern is seemingly very common, too, so a lot of other lint rules have a bespoke version of this. I can happily go refactor those cases to as they should all become quite a bit simpler.

### DIFF
--- a/packages/eslint-plugin/src/util/getTextWithParentheses.ts
+++ b/packages/eslint-plugin/src/util/getTextWithParentheses.ts
@@ -1,0 +1,34 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+import type { SourceCode } from '@typescript-eslint/utils/ts-eslint';
+import {
+  isClosingParenToken,
+  isOpeningParenToken,
+  isParenthesized,
+  nullThrows,
+  NullThrowsReasons,
+} from '.';
+
+export function getTextWithParentheses(
+  sourceCode: Readonly<SourceCode>,
+  node: TSESTree.Node,
+): string {
+  // Capture parentheses before and after the node
+  let beforeCount = 0;
+  let afterCount = 0;
+
+  if (isParenthesized(node, sourceCode)) {
+    const bodyOpeningParen = nullThrows(
+      sourceCode.getTokenBefore(node, isOpeningParenToken),
+      NullThrowsReasons.MissingToken('(', 'node'),
+    );
+    const bodyClosingParen = nullThrows(
+      sourceCode.getTokenAfter(node, isClosingParenToken),
+      NullThrowsReasons.MissingToken(')', 'node'),
+    );
+
+    beforeCount = node.range[0] - bodyOpeningParen.range[0];
+    afterCount = bodyClosingParen.range[1] - node.range[1];
+  }
+
+  return sourceCode.getText(node, beforeCount, afterCount);
+}
--- a/packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
+++ b/packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
@@ -333,6 +333,26 @@ x ?? 'foo';
           },
         ],
       },
+      {
+        code: code.replace(/y/g, '(z = y)'),
+        output: null,
+        options: [{ ignoreTernaryTests: false }] as const,
+        errors: [
+          {
+            messageId: 'preferNullishOverTernary' as const,
+            line: 1,
+            column: 1,
+            endLine: 1,
+            endColumn: code.replace(/y/g, '(z = y)').length,
+            suggestions: [
+              {
+                messageId: 'suggestNullish' as const,
+                output: 'x ?? (z = y);',
+              },
+            ],
+          },
+        ],
+      },
     ]),
 
     {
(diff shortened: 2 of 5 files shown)


packages/eslint-plugin/src/util/getTextWithParentheses.ts
```diff
@@ -0,0 +1,35 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+import type { SourceCode } from '@typescript-eslint/utils/ts-eslint';
+
+import {
+  isClosingParenToken,
+  isOpeningParenToken,
+  isParenthesized,
+  nullThrows,
+  NullThrowsReasons,
+} from '.';
```
JoshuaKGoldberg, line 10: We generally try not to import `from '.'`, as that generates a cycle:

```suggestion
} from '../util';
```

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
```diff
@@ -333,6 +333,26 @@ x ?? 'foo';
           },
         ],
       },
+      {
+        code: code.replace(/y/g, '(z = y)'),
```
JoshuaKGoldberg, line 337: [Testing] The changed logic includes `beforeCount` and `afterCount`, but there's only one here. Could you add some tests?

* `((z = y))`
* `( /* */ ( z = y ) /* */ )`
* Maybe, a multiline comment or two?
* ...whatever you think is also a good idea to test

JoshuaKGoldberg (changes requested):
The implementation looks great to me! Requesting changes on testing and an import path. Thanks! ✨

packages/eslint-plugin/src/util/getTextWithParentheses.ts
jakebailey: I can do that, though that's still a cycle; those are the same imports, they both point to `index.ts`.

JoshuaKGoldberg: Oh, maybe we need to flatten these out...

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
jakebailey: Yeah, I'm afraid it doesn't handle any of those cases. the `getTextWithParentheses` function I moved isn't fancy enough to handle that.

Is there a helper that already exists to do this that I'm not aware of? Reprinting a node like this sure seems like a common case.

jakebailey: Maybe `getWrappingFixer`? I had looked at that but couldn't reverse engineer how to use it in this application.

JoshuaKGoldberg: That would be the one, yeah. Does its use in `strict-boolean-expressions` help?

jakebailey: Tried it out, and it adds parens in places it shouldn't, and doesn't preserve multiple parens. So, I don't think I have a good idea...

JoshuaKGoldberg: Hmm. Well, I guess you're on your own then. Blurgh.

If there's anything that's an existing bug and would be a lot more work to tackle, I think it's fine as a followup issue. I'll just mark this as ✅ for now.

JoshuaKGoldberg (approved):
🚀

PR #9418, opened 2024-06-24
chore: revert 'chore(deps): update nx to v19.3.0 #9371'

JoshuaKGoldberg:
Reverts #9371.

The weekly release is failing, and I really should have consulted with @JamesHenry before moving us off the canary. Sorry 😄


JamesHenry:
@JoshuaKGoldberg I imagine it's just that the patch is no longer being applied correctly? Sorry that that is still required right now

JoshuaKGoldberg:
Maybe! I'd rather look at it separately, want to get the release out the door heh.

PR #9460, opened 2024-06-30
fix: disable `projectService` in `disabled-type-checked` shared config

### LINKED ISSUE #9316: Bug: `disabled-type-checked` config does not disable project service
The `disable-type-checked` config does not disable the project service - which causes matched files to not be properly excluded from type-aware linting.

The configs are generated and the relevant config is here:
https://github.com/typescript-eslint/typescript-eslint/blob/63e53e28f68f84764ec632d502da3d33095f6a0d/packages/repo-tools/src/generate-configs.mts#L391-L394

Ideally we should fix this separately in both `main` (set `EXPERIMENTAL_useProjectService: false`) and in the `v8` branch (set `projectService: false`).

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9316

### DIFF
--- a/packages/eslint-plugin/src/configs/disable-type-checked.ts
+++ b/packages/eslint-plugin/src/configs/disable-type-checked.ts
@@ -8,7 +8,7 @@
 import type { ClassicConfig } from '@typescript-eslint/utils/ts-eslint';
 
 export = {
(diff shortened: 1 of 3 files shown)


JoshuaKGoldberg (approved):
💯

PR #9462, opened 2024-07-01
Chore(eslint-plugin): re-using logic from builtinSymbolLike in no-implied-eval

YutongZhuu:
## PR Checklist

- [] Addresses an existing open issue: fixes #000

## Overview

As title, I use the function in isBuiltinSymbolLike.ts to replace similar logic. But I am passing a 'FunctionConstructor' to a function instead of Function since there is a test case where the symbol name is FunctionConstructor. I am not sure if this is the right thing to do.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-implied-eval.ts
+++ b/packages/eslint-plugin/src/rules/no-implied-eval.ts
@@ -4,6 +4,7 @@ import * as tsutils from 'ts-api-utils';
 import * as ts from 'typescript';
 
 import { createRule, getParserServices } from '../util';
+import { isBuiltinSymbolLike } from '../../../type-utils/src/builtinSymbolLikes';
 
 const FUNCTION_CONSTRUCTOR = 'Function';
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/src/rules/no-implied-eval.ts
```diff
@@ -76,14 +76,8 @@ export default createRule({
       }
 
       // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
```
auvred, line 78:
```suggestion
```

It looks like this `eslint-disable` directive is no longer required.

auvred (changes requested):
Looks good! Just requesting to remove an unnecessary `eslint-disable` directive.

packages/eslint-plugin/src/rules/no-implied-eval.ts
YutongZhuu: addressed :)

auvred (approved):
💯

JoshuaKGoldberg (approved):
🧹✨

PR #9479, opened 2024-07-02
feat(eslint-plugin): [no-duplicate-type-constituents] prevent unnecessary `| undefined` for optional parameters

### LINKED ISSUE #9203: Enhancement: [no-duplicate-type-constituents] Prevent unnecessary `| undefined` for optional parameters

### Description

When a function parameter (or a class property) is marked as optional with `?`, it implicitly includes `undefined` in the type definition, so there is no need to explicitly add an `| undefined` to the type definition. I'd like to detect these cases and also autofix away the `| undefined`.

I don't think I see an existing rule to do this and it seems too complicated to do with `no-restricted-syntax` (though it wouldn't be able to autofix anyways).

### Fail Cases

```TypeScript
function foo(bar?: number | undefined) {
    return bar
}
```

### Pass Cases

```TypeScript
function foo(bar?: number) {
    return bar
}
```

### Additional Info

_No response_

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9203 


packages/eslint-plugin/src/rules/no-duplicate-type-constituents.ts
```diff
@@ -5,13 +5,10 @@ import type { Type } from 'typescript';
 import { createRule, getParserServices } from '../util';
 
 export type Options = [
-  {
-    ignoreIntersections?: boolean;
-    ignoreUnions?: boolean;
-  },
+  { ignoreIntersections?: boolean; ignoreUnions?: boolean },
```
auvred, line 8: These changes don't seem to be within the scope of this PR. Could you please revert them? Unnecessary changes make the reviewer's job harder :(

packages/eslint-plugin/src/rules/no-duplicate-type-constituents.ts
```diff
@@ -145,63 +139,71 @@ export default createRule<Options, MessageIds>({
...
+                const explicitUndefined = node.types.find(
+                  ({ type }) => type === AST_NODE_TYPES.TSUndefinedKeyword,
+                );
+                if (explicitUndefined) {
+                  report('unnecessary', explicitUndefined);
+                }
+              }
+            }
+          }
+        },
```
auvred, line 206: [Bug] This case isn't reported

```ts
function a(arg?: number | (string | undefined)) {}
```

[playground](https://deploy-preview-9479--typescript-eslint.netlify.app/play/#ts=5.5.2&fileType=.ts&code=GYVwdgxgLglg9mABAQwBTIE4HMD8AuRMEAWwCMBTDRAH0VQGcoMYwsbFwATc4F8zgJQDEAbwC%2BQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNZjGyAQ3yJaRUrTIcmyfJXyxELVBkiJo0DtEjgwAXxB6gA&tsconfig=&tokens=false)

packages/eslint-plugin/src/rules/no-duplicate-type-constituents.ts
```diff
@@ -145,63 +139,71 @@ export default createRule<Options, MessageIds>({
...
-      const beforeUnionOrIntersectionToken =
-        beforeTokens[beforeTokens.length - 1];
-      const bracketBeforeTokens = context.sourceCode.getTokensBetween(
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const beforeUnionOrIntersectionToken = sourceCode
+        .getTokensBefore(node, {
+          filter: token => ['|', '&'].includes(token.value),
+        })
+        .at(-1)!;
+      const bracketBeforeTokens = sourceCode.getTokensBetween(
```
auvred, line 153: [Bug] Rule throws error on this code:

```
getTokensBetween@https://deploy-preview-9479--typescript-eslint.netlify.app/sandbox/index.js:1:451479
o@https://deploy-preview-9479--typescript-eslint.netlify.app/sandbox/index.js:1:1770041
TSUnionType@https://deploy-preview-9479--typescript-eslint.netlify.app/sandbox/index.js:1:1770599
ad</Y/</</<@https://deploy-preview-9479--typescript-eslint.netlify.app/sandbox/index.js:1:1394247
Fp</t.exports/emit/<@https://deploy-preview-9479--typescript-eslint.netlify.app/sandbox/index.js:1:1350213
emit@https://deploy-preview-9479--typescript-eslint.netlify.app/sandbox/index.js:1:1350200...
```

```ts
function a(arg?: undefined | string) {}
```

[playground](https://deploy-preview-9479--typescript-eslint.netlify.app/play/#ts=5.5.2&fileType=.ts&code=GYVwdgxgLglg9mABAQwBTIE4HMD8AuRcAEwFNgYwSjEAfRAZygwqwEpEBvAXyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNZjGyAQ3yJaRUrTIcmyfJXyxELVBkiJo0DtEjgwAXxB6gA&tsconfig=&tokens=false)

auvred (changes requested):

packages/eslint-plugin/src/rules/no-duplicate-type-constituents.ts
abrahamguo: yep! Sorry about that. Wish Prettier was more strict haha

packages/eslint-plugin/src/rules/no-duplicate-type-constituents.ts
abrahamguo: Good catch, it is now possible to report on the first element of a union or intersection, where that wasn't possible before

packages/eslint-plugin/src/rules/no-duplicate-type-constituents.ts
abrahamguo: This is not currently handled even in the existing rule functionality — we'd need to implement some recursion handling for this. I've gone ahead and opened #9496 for further discussion

abrahamguo:
Work done:

- [x] Moved `reportDuplicate` into a more inner scope so that it can read `constituentNode` from scope
- [x] Rename it to `report` because it can now accept a message ID
- [x] Update autofixing logic to now support removing a `&` or `|` _after_ the duplicate node — this is necessary because a duplicate node will never be the first member of an intersection or union, but a useless `undefined` can be the first member of an intersection or union
- [x] Add parameter to `checkDuplicate` to allow additional checks on just unions or just intersections
- [x] Update docs and add new lint message

Ready for review!

packages/eslint-plugin/src/rules/no-duplicate-type-constituents.ts
```diff
@@ -117,94 +131,118 @@ export default createRule<Options, MessageIds>({
...
+                maybeIdentifier.type === AST_NODE_TYPES.Identifier &&
+                maybeIdentifier.optional
+              ) {
+                const maybeFunction = maybeIdentifier.parent;
+                const { type } = maybeFunction;
+                if (
+                  (type === AST_NODE_TYPES.ArrowFunctionExpression ||
+                    type === AST_NODE_TYPES.FunctionDeclaration ||
+                    type === AST_NODE_TYPES.FunctionExpression) &&
+                  maybeFunction.params.includes(maybeIdentifier) &&
```
auvred, line 238: Shouldn't we also report on `TSDeclareFunction`, `TSEmptyBodyFunctionExpression`, `TSMethodSignature`, `TSFunctionType` and `TSAbstractMethodDefinition`?

```ts
declare function f(a?: string | undefined): void

declare class bb {
  f(a?: string | undefined): void
}

interface ee {
  f(a?: string | undefined): void
}

type fn = (a?: string | undefined) => void

abstract class cc {
  abstract f(a?: string | undefined): void
}
```

[playground](https://deploy-preview-9479--typescript-eslint.netlify.app/play/#ts=5.5.2&showAST=es&fileType=.ts&code=CYUwxgNghgTiAEAzArgOzAFwJYHtVIAooB%2BALngGcMYtUBzeAH3jVEVpGAEpyA3HLMABQQ0JFgJxFCvABGs%2BAG8h8QiXJUa9Ji1RsO3PgOEBfEbQwgYiKGAQgEy1YiJlK1Wg2asQ7VJx54fkEhMyEMAE8ABwREfABeeFcND21vPV8DLnh4gD4g4xEoWU1bDHgpGTAwJRV4YtLMNTdNTx0fPwCjEJMgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNZjGyAQ3yJaRUrTIcmyfJXyxELVBkiJo0DtEjgwAXxB6gA&tsconfig=&tokens=false)

packages/eslint-plugin/tests/rules/no-duplicate-type-constituents.test.ts
```diff
@@ -673,5 +674,21 @@ type T = Record<string, A  >;
         },
       ],
     },
+    {
+      code: '(a?: string | undefined) => {};',
```
auvred, line 678: [Testing] Could you add some test cases with `FunctionDeclaration`, `FunctionExpression` and node types mentioned in my comment above?

auvred (changes requested):

packages/eslint-plugin/src/rules/no-duplicate-type-constituents.ts
abrahamguo: Good catch! I added even more node types — I added all node types that had the `params` property I was using.
Also, note that `TSAbstractMethodDefinition` is not needed, because the `value` property of the `TSAbstractMethodDefinition` contains a `TSEmptyBodyFunctionExpression`.

packages/eslint-plugin/src/rules/no-duplicate-type-constituents.ts
```diff
@@ -230,11 +231,9 @@ export default createRule<Options, MessageIds>({
                 maybeIdentifier.optional
               ) {
                 const maybeFunction = maybeIdentifier.parent;
-                const { type } = maybeFunction;
                 if (
-                  (type === AST_NODE_TYPES.ArrowFunctionExpression ||
-                    type === AST_NODE_TYPES.FunctionDeclaration ||
-                    type === AST_NODE_TYPES.FunctionExpression) &&
+                  (isFunctionOrFunctionType(maybeFunction) ||
+                    maybeFunction.type === AST_NODE_TYPES.TSDeclareFunction) &&
```
abrahamguo, line 236: Created #9788 to add `TSDeclareFunction` in `utils`.

auvred:
Let's wait for https://github.com/typescript-eslint/typescript-eslint/pull/9788 to get merged

abrahamguo:
Updated now that #9788 has been merged ✅

auvred (approved):
:100:

jswalden:
Hi, maybe I'm missing something here, but...why does this lint make sense to have for `&`?

For `|`, if you're intending to concatenate type sets rather than combine them, you might well be presuming their disjointedness.

But for `&`, if the type sets don't contain a duplication...aren't you just going to vacuously get the empty type set (which I presume means `never`)?

Maybe I'm just too TypeScript n00b.  But it seems like if you're intersecting type sets not with intention of asserting non-overlap, the only way you get any value from doing so is if there _is_ duplication.  (This was, in fact, exactly what I was intentionally doing when I just ran into this in my own code.)

abrahamguo:
@jswalden `&` can be useful!

One place I like to use it is to combine two object types, like so:
```
type Animal = { name: string };
type Bear = Animal & { honey: boolean };
```

Here, we're using `&` to combine the `Animal` type and add additional properties — this is an example of using `&` without duplication.

jswalden:
@abrahamguo Interesting!  I probably should have realized that possibility.  :slightly_smiling_face: 

I guess I don't have good ideas about `&` where it is not intended to build up an intersection type that way -- other than the explicit disabling comment I eventually resorted to.  I'd suggest at least keep an eye out to see how many other people are in my boat, tho.  If other people run into this, in "sufficient" quantity (you'd be a far better judge of what that is than me!), it might be worth thinking about if there's anything that could be done to treat the two use cases separately "somehow".

abrahamguo:
Can you show the example from your code, where you ended up needing to use a disable comment?

jswalden:
@abrahamguo The lines are here: https://github.com/bitfocus/companion-module-allenheath-sq/blob/50255dd03c4d2c2efc9a1e84e2a97483570c29f2/src/presets.ts#L17

I'll try to briefly summarize the enclosing use, as the line alone is probably obscure.

[Companion](https://github.com/bitfocus/companion) is software you can use to define connections to control devices (or more-abstract stuff like a Slack connection or a Google spreadsheet), by programming buttons on various hardware devices (such as the [Stream Deck](https://www.amazon.com/Elgato-Stream-Deck-XL-customizable/dp/B07RL8H55Z/)).  My code defines a module implementing a connection to control an Allen & Heath sound mixer.

Companion modules in part define a set of "actions" Companion users can invoke to control the underlying device, and a set of "feedbacks" that expose device status changes.  Users can assign series of actions to a hardware button, to be performed when it's pressed/released.  A module _also_ can define "presets": predefined buttons for simple, common use cases.

Actions and feedbacks are identified using separate string ID sets.  In this module, I represent "mute this input or output" action and feedback IDs using the [`MuteActionId`](https://github.com/bitfocus/companion-module-allenheath-sq/blob/50255dd03c4d2c2efc9a1e84e2a97483570c29f2/src/actions/mute.ts#L19) and [`MuteFeedbackId`](https://github.com/bitfocus/companion-module-allenheath-sq/blob/50255dd03c4d2c2efc9a1e84e2a97483570c29f2/src/feedbacks/feedback-ids.ts#L15) enums.

This module has presets that show whether a particular mixer input/output is muted, that will toggle muting when pressed.  When I define these presets, I use the keys of those enums to specify the particular kind of mute button being defined.  I _intersect_ those keys to get a key that identifies both a mute action and a mute feedback.  I can then use `MuteActionId[key]` or `MuteFeedbackId[key]` to get an ID of the desired type, without risking a typo.

I hope that's reasonably clear without being excessively detailed!

abrahamguo:
In this case, shouldn't `MuteFeedbackId` and `MuteActionId` be merged, since they are identical to each other?
That is why `no-duplicate-type-constituents` is reporting — because the two `enum`s are identical to each other, and therefore it's unnecessary to combine them with `&`.

If the two `enum`s were different from each other, with only some properties in common, then the lint rule would not report.

jswalden:
@abrahamguo They're type-identical, but they are not _semantically_ identical.  Feedback IDs aren't the same as action IDs.

PR #9500, opened 2024-07-05
fix(type-utils): also check declared modules for package names in TypeOrValueSpecifier

### LINKED ISSUE #6861: Enhancement: Detect declared modules for TypeOrValueSpecifier

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6861

## Overview

Adds a manual check to `TypeOrValueSpecifier` checks for declaration nodes inside a `declare module '...' {` matching on `packageName`.

The tests in `no-floating-promises` check for this in two ways:
* `'abc'` and `'node:test'`: when declared in the tested file
* `'node:test'`: sourced from `@types/node`

I found the `TypeOrValueSpecifier.ts` to be getting a bit tall, so I split up its main helper functions into their own files.

💖 

Co-authored-by: @RebeccaStevens


packages/type-utils/src/typeOrValueSpecifiers/typeDeclaredInPackage.ts
```diff
@@ -0,0 +1,55 @@
...
+    case ts.SyntaxKind.ModuleDeclaration:
+      return node as ts.ModuleDeclaration;
+    case ts.SyntaxKind.SourceFile:
+      return undefined;
+    default:
+      return findParentModuleDeclaration(node.parent);
+  }
+}
+
+function typeDeclaredInDeclareModule(
```
JoshuaKGoldberg, line 16: This is the main added logic in this PR.

cc @marekdedic @RebeccaStevens as previous authors around the TypeOrValueSpecifier tooling. Sorry it took me so long to follow up on this point!

packages/type-utils/src/typeOrValueSpecifiers/typeDeclaredInPackage.ts
```diff
@@ -0,0 +1,55 @@
...
+  packageName: string,
+  declarations: ts.Node[],
+): boolean {
+  return declarations.some(
+    declaration =>
+      findParentModuleDeclaration(declaration)?.name.text === packageName,
+  );
+}
+
+function typeDeclaredInDeclarationFile(
```
RebeccaStevens, line 26: Might want to make this function's name a bit more clear that it is for package declaration files.

e.g. `typeDeclaredInPackageDeclarationFile`

packages/eslint-plugin/tests/rules/no-floating-promises.test.ts
```diff
@@ -754,6 +754,53 @@ promise().then(() => {});
...
+        it('...', () => {});
+      `,
+      options: [
+        {
+          allowForKnownSafeCalls: [
+            { from: 'package', name: 'it', package: 'node:test' },
+          ],
+        },
+      ],
+    },
```
RebeccaStevens, line 803: Add test for local modules.

e.g. `declare module './foo'`. It should pass with `{ from: 'file', name: '...' }` but fail with `{ from: 'package', package: './foo' }`.

JoshuaKGoldberg: D'oh, I also just realized we already have a dedicated test file for this: `packages/type-utils/tests/TypeOrValueSpecifier.test.ts`. I'll move non-cursory tests there.

RebeccaStevens (approved):
LGTM

I've made this same change to [ts-declaration-location](https://www.npmjs.com/package/ts-declaration-location?activeTab=readme)

abrahamguo:
@JoshuaKGoldberg I was investigating your work in this PR because I'm hoping to use it in #8556.

Here's a question I had:
In the `main` branch, it is impossible to match the built-in type `URL` if `@types/node` is installed ([explanation](https://github.com/typescript-eslint/typescript-eslint/pull/8556#issuecomment-2165608615)).

In this PR, I see that it is now possible to match the built-in type `URL` when `@types/node` is installed.
For example, the specifier `{ from: 'package', name: 'URL', package: 'global' }` matches against `type Test = URL;`.

This is because `findParentModuleDeclaration(declaration)?.name.text` returns `'global'`.

Is this expected? I'm not 100% sure where `global` is coming from — I believe it's a `declare global` somewhere.

packages/type-utils/src/typeOrValueSpecifiers/typeDeclaredInPackageDeclarationFile.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+function typeDeclaredInDeclareModule(
+  packageName: string,
+  declarations: ts.Node[],
+): boolean {
+  return declarations.some(
+    declaration =>
+      findParentModuleDeclaration(declaration)?.name.text === packageName,
```
abrahamguo, line 22: @JoshuaKGoldberg I was investigating your work in this PR because I'm hoping to use it in #8556.

Here's a question I had:
In the `main` branch, it is impossible to match the built-in type `URL` if `@types/node` is installed ([explanation](https://github.com/typescript-eslint/typescript-eslint/pull/8556#issuecomment-2165608615)).

In this PR, I see that it is now possible to match the built-in type `URL` when `@types/node` is installed.
For example, the specifier `{ from: 'package', name: 'URL', package: 'global' }` matches against `type Test = URL;`.

This is because `findParentModuleDeclaration(declaration)?.name.text` returns `'global'`.

Is this expected? I'm not 100% sure where `global` is coming from — I believe it's a `declare global` somewhere.

JoshuaKGoldberg: Huh, that's ... a really interesting bug! Thanks for pointing it out. `findParentModuleDeclaration` should be filtering out `declare global` but isn't. Nice find!

bradzacher (approved):

packages/type-utils/src/typeOrValueSpecifiers/typeDeclaredInPackageDeclarationFile.ts
JoshuaKGoldberg: (I fixed that in this PR)

PR #9518, opened 2024-07-08
feat(typescript-estree): replace `globby` w/ `fast-glob`

### LINKED ISSUE #9453: Enhancement: replace `globby` with `fast-glob`

### Relevant Package

typescript-estree

### Description

Previously, typescript-eslint switched from `glob` to `globby` for `tsconfig.json` file searching: https://github.com/typescript-eslint/typescript-eslint/pull/2418

Under the hood, `globby` is powered by `fast-glob` with extra features, but none of them are utilized by the `typescript-eslint`.

Since the `typescript-eslint` doesn't utilize those extra features, I propose replacing `globby` with `fast-glob` to reduce the installation size.

### Additional Info

**The installation size difference**

[The installation size of `globby` is 605 KiB and 23 transitive dependencies are introduced](https://pkg-size.dev/globby), while [the installation size of `fast-glob` is 503 KiB and 18 transitive dependencies are introduced](https://pkg-size.dev/fast-glob).

**Proof of Concept**

I've tested the proposal in my fork: https://github.com/SukkaW/typescript-eslint/commit/0cec33997c451c8bbbb76ee4a591308a1517cfc0 and all existing test cases are passed on my machine.

**globby's extra features**

Here is globby's feature description:

- Promise API
  - `fast-glob` also has promise-based API, and `typescript-eslint` only uses the sync API
- Multiple patterns
  - `globby`'s multiple patterns support is powered by `fast-glob` directly
- Negated patterns: `['foo*', '!foobar']`
  - `globby`'s negated patterns support is also powered by `fast-glob` directly
- Expands directories: `foo → foo/**/*`
  - `typescript-eslint` doesn't rely on this feature since the exact matches of `tsconfig.json` (-like) files are required
- Supports `.gitignore` and similar ignore config files
  - `typescript-eslint` doesn't rely on this feature since `globby` disabled this feature by default and `typescript-eslint` doesn't enable it.
- Supports URL as `cwd`
  - Currently `typescript-eslint` passed the string path to `cwd`, and we can use Node.js built-in  `url.fileURLToPath` in case we encounter URLs in the future.

SukkaW:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9453
  - Not yet, but likely to be accepted soon though 

## Overview

Addresses #9453. All tests passed on my machine locally.


packages/typescript-estree/tests/lib/parse.test.ts
```diff
@@ -506,8 +506,8 @@ describe('parseAndGenerateServices', () => {
           expect(testParse('ts/notIncluded0j1.ts'))
             .toThrowErrorMatchingInlineSnapshot(`
             "ESLint was configured to run on \`<tsconfigRootDir>/ts/notIncluded0j1.ts\` using \`parserOptions.project\`: 
-            - <tsconfigRootDir>/tsconfig.json
             - <tsconfigRootDir>/tsconfig.extra.json
+            - <tsconfigRootDir>/tsconfig.json
```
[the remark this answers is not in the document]
SukkaW: > Is there a small difference how these libraries find files

Yes, and thus results in an order difference.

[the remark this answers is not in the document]
JoshuaKGoldberg: Hmm, ordering changes would make this either a bug, a bug _fix_, or a breaking change.

It looks to me like this is a bug. The `project: ['./**/tsconfig.json', './**/tsconfig.extra.json'],` indicates that `tsconfig.json` should come first. No?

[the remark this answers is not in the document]
SukkaW: @JoshuaKGoldberg Fixed in the latest commits. Comments are added to explain the change.

packages/typescript-estree/src/parseSettings/resolveProjectList.ts
```diff
@@ -56,15 +56,15 @@ export function resolveProjectList(
...
-    .map(folder => (folder.startsWith('!') ? folder : `!${folder}`));
+  ).reduce<string[]>((acc, folder) => {
+    if (typeof folder === 'string') {
+      acc.push(
+        // prefix with a ! for not match glob
+        folder.startsWith('!') ? folder : `!${folder}`,
+      );
+    }
+    return acc;
+  }, []);
```
SukkaW, line 67: A bonus change, replace the `.reduce.map` chain with a single `.reduce` to avoid unnecessary iteration.

benmccann:
@SukkaW would you consider updating this PR to use `tinyglobby` rather than `fast-glob` as `tinyglobby` has fewer dependencies?

I took a naive stab at it in https://github.com/typescript-eslint/typescript-eslint/pull/9878, but it's failing the CI. I see you have a lot of extra logic here that I don't have, so perhaps that is needed. It seems you're much more familiar with how this project is setup, so if you're able to update this to use `tinyglobby` perhaps I can just close my PR

I was also curious why this is in draft mode as it appears to be working. Perhaps if you're able to update it to `tinyglobby` it could also be marked as ready for review?

SukkaW:
> but it's failing the CI. I see you have a lot of extra logic here that I don't have, so perhaps that is needed.

You can check my comments. `fast-glob` returns files array in arbitrary order. `globby` instead matches patterns one by one (iterating file systems multiple times) and then concat the results to ensure the order is correct. I added extra codes to mimic this behavior. You can try this too.

benmccann:
I've gotten the CI green now on https://github.com/typescript-eslint/typescript-eslint/pull/9878 based on including some of the changes from this PR. I think we could probably just merge that one since `tinyglobby` is both smaller and faster

I didn't include the change from https://github.com/typescript-eslint/typescript-eslint/pull/9518#discussion_r1672292770 since it was unrelated to switching glob libraries. I could update my PR to include it if desired or maybe this PR could be updated to just have that change or it could be submitted as a new PR

bradzacher (approved):

JoshuaKGoldberg (approved):
LGTM, thanks for the cleaup! 🧹

packages/typescript-estree/src/parseSettings/resolveProjectList.ts
```diff
@@ -93,16 +93,23 @@ export function resolveProjectList(
   const nonGlobProjects = sanitizedProjects.filter(project => !isGlob(project));
   const globProjects = sanitizedProjects.filter(project => isGlob(project));
 
+  let globProjectPaths: string[] = [];
+
+  if (globProjects.length > 0) {
+    // Although fast-glob supports multiple patterns, fast-glob returns arbitrary order of results
+    // to improve performance. To ensure the order is correct, we need to call fast-glob for each pattern
```
alfaproject, line 100: OOC, does the order matter?

SukkaW: There is a unit test case against the order, so yes, it does matter.

alfaproject: But why does the order matter? Shouldn't the unit test just check that the needed files are linted?

JoshuaKGoldberg: Folks sometimes do things like `['packages/*/tsconfig.json', 'tsconfig.json']` - and/or entries like `'tsconfig.eslint.jsom'`. The order intent is to first try the lower-level TSConfigs first.

alfaproject: Gotcha, thanks!

PR #9530, opened 2024-07-09
fix(eslint-plugin): [no-unnecessary-type-parameters] descend into all parts of mapped types in no-unnecessary-type-parameters

### LINKED ISSUE #9524: Bug: [no-unnecessary-type-parameters] double false positive

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.2&fileType=.ts&code=KYDwDg9gTgLgBAYwgOwM7wLYEMwHkBGAVnALxwA8A0nKDMMgCapzpQCWyA5gDRwBqvACoA%2BABQQiALjgBvOAG0A1sACecDnEoBdAPzS%2BcAL68AZsmmjlK6ZV4A3LABt9ASlLC4gtyWEAoOHAEhMAIMAB09DDswKjiRC5h2GCikVBqPnBmomE5qWpYzPK2-FouLgDcQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=&tokens=false

### Repro Code

```TypeScript
export const mapObj = <K extends string, V, T>(obj: { [key in K]?: V }, fn: (key: K, val: V) => T) =>
  Object.entries(obj).map(entry => fn(...entry as [K, V]));
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### Expected Result

The rule should not report on `V` — `V` is used 3 times

### Actual Result

The rule claims that `V` is only used once, when it's actually used 3 times.

### Additional Info

_No response_

danvk:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9524 

## Overview

Mapped types have three parts: the type parameter, the template type and the constraint type:

```
> type.checker.typeToString(type)
'{ [key in K]?: V | undefined; }'
> type.checker.typeToString(type.typeParameter)
'key'
> type.checker.typeToString(type.templateType)
'V | undefined'
> type.checker.typeToString(type.constraintType)
'K'
```

The rule previously descended into the typeParameter. It doesn't need to descend into the constraintType because we get that through another code path. It does need to descend into the templateType, though.

One wrinkle is that if the mapped type is something like `{[k in 'a']: T}`, then TypeScript will treat it like `{a: T}` and we'll process it as an object type. In this case, we need to avoid _also_ processing it as a mapped type, even though `isMappedType` returns true.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
@@ -271,12 +271,17 @@ function collectTypeParameterUsageCounts(
     // `isTypeReference` to avoid descending into all the properties of a
     // generic interface/class, e.g. `Map<K, V>`.
     else if (tsutils.isObjectType(type)) {
-      visitSymbolsListOnce(type.getProperties(), false);
+      const properties = type.getProperties();
+      visitSymbolsListOnce(properties, false);
 
       if (isMappedType(type)) {
         visitType(type.typeParameter, false);
-        visitType(type.templateType, false);
-        visitType(type.constraintType, false);
+        if (properties.length === 0) {
+          // TS treats mapped types like `{[k in "a"]: T}` like `{a: T}`.
+          // They have properties, so we need to avoid double-counting.
+          visitType(type.templateType, false);
+          visitType(type.constraintType, false);
+        }
       }
 
       for (const typeArgument of type.aliasTypeArguments ?? []) {
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
@@ -275,6 +275,8 @@ function collectTypeParameterUsageCounts(
 
       if (isMappedType(type)) {
         visitType(type.typeParameter, false);
+        visitType(type.templateType, false);
+        visitType(type.constraintType, false);
       }
 
       for (const typeArgument of type.aliasTypeArguments ?? []) {
@@ -372,6 +374,8 @@ function collectTypeParameterUsageCounts(
 
 interface MappedType extends ts.ObjectType {
   typeParameter?: ts.Type;
+  constraintType?: ts.Type;
+  templateType?: ts.Type;
 }
 
 function isMappedType(type: ts.Type): type is MappedType {
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
@@ -280,7 +280,6 @@ function collectTypeParameterUsageCounts(
           // TS treats mapped types like `{[k in "a"]: T}` like `{a: T}`.
           // They have properties, so we need to avoid double-counting.
           visitType(type.templateType, false);
-          visitType(type.constraintType, false);
         }
       }
 
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
@@ -342,6 +342,12 @@ ruleTester.run('no-unnecessary-type-parameters', rule, {
         (token): token is Exclude<TSESTree.Token, Conditions & ExtractedToken> =>
           tokenType in conditions;
     `,
+    `
+      declare function mapObj<K extends string, V>(
+        obj: { [key in K]?: V },
+        fn: (key: K, val: V) => number,
+      ): number[];
+    `,
   ],
 
   invalid: [
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
@@ -617,15 +617,6 @@ ruleTester.run('no-unnecessary-type-parameters', rule, {
     {
       code: `
         declare function mapObj<K extends string, V>(
-          obj: { [key in K]?: null },
-          fn: (key: K) => number,
-        ): number[];
-      `,
-      errors: [{ messageId: 'sole', data: { name: 'V' } }],
-    },
-    {
-      code: `
-        declare function mapObj<K extends string, V>(
           obj: { [key in K]?: V },
           fn: (key: K) => number,
         ): number[];
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
@@ -605,7 +605,6 @@ ruleTester.run('no-unnecessary-type-parameters', rule, {
     {
       code: "type Fn = <T>() => { [K in 'a']: T };",
       errors: [{ messageId: 'sole', data: { name: 'T' } }],
-      // only: true,
     },
     {
       code: 'type Fn = <T>(value: unknown) => value is T;',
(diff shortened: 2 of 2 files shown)


danvk:
Incidentally @JoshuaKGoldberg I'm not sure why the new test case doesn't trigger the "quick path:"

https://github.com/typescript-eslint/typescript-eslint/blob/bebbd0c274e5009bf6b03b3a382782a3a708248e/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts#L59-L63

`scope.references` is empty for both `K` and `V` in:

```ts
declare function mapObj<
  K extends string,
  V,  // Type parameter V is used only once
>(
  obj: { [key in K]: V },
  fn: (key: K, val: V) => number,
): number[];
```

I would have expected it to contain the references.

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -372,6 +378,8 @@ function collectTypeParameterUsageCounts(
 
 interface MappedType extends ts.ObjectType {
   typeParameter?: ts.Type;
+  constraintType?: ts.Type;
+  templateType?: ts.Type;
```
JoshuaKGoldberg, line 382: Ah, thanks, this reminded me to file an issue on TypeScript: https://github.com/microsoft/TypeScript/issues/59221

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
JoshuaKGoldberg, line 1: [Testing] I think we should add at least one `invalid` case to make sure the added logic plays well with the rule's existing reporting semantics too. The first one that comes to my mind is an unnecessary `V`:

```ts
declare function mapObj<K extends string, V>(
  obj: { [key in K]?: null },
  fn: (key: K) => number,
): number[];
```

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -271,10 +271,16 @@ function collectTypeParameterUsageCounts(
...
-      visitSymbolsListOnce(type.getProperties(), false);
+      const properties = type.getProperties();
+      visitSymbolsListOnce(properties, false);
 
       if (isMappedType(type)) {
         visitType(type.typeParameter, false);
+        if (properties.length === 0) {
+          // TS treats mapped types like `{[k in "a"]: T}` like `{a: T}`.
+          // They have properties, so we need to avoid double-counting.
+          visitType(type.templateType, false);
```
JoshuaKGoldberg, line 282: [Praise] This is very cool, TIL. Nice little collapsing on the TS side.

JoshuaKGoldberg (changes requested):
LGTM, thanks for the fast follow! ⚡ 

Just requesting changes on a bit of testing. This basically an approval otherwise.

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
danvk: Sure, I added that as a test. It's a little odd that this is flagged as "used only once" when it's not used at all. Maybe we should just let TypeScript handle that case? [playground](https://typescript-eslint.io/play/#ts=5.5.2&fileType=.ts&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXwFsoAHAeQCMArAHgGl4QAPDEVYAZ3nYxi1QHMANPABqAPgAUAKHjwcVAFzwA3vADaAaxABPeH3i0AugH4lqZBAjwAvoJlJUSiVu1LaASngBeMfHMFyEBg7dzNkAKC1QwBueCkgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=&tokens=false)

I also added an invalid case where `V` is used exactly once in the "templateType" slot.

JoshuaKGoldberg: > Maybe we should just let TypeScript handle that case? [playground](https://typescript-eslint.io/play/#ts=5.5.2&fileType=.ts&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXwFsoAHAeQCMArAHgGl4QAPDEVYAZ3nYxi1QHMANPABqAPgAUAKHjwcVAFzwA3vADaAaxABPeH3i0AugH4lqZBAjwAvoJlJUSiVu1LaASngBeMfHMFyEBg7dzNkAKC1QwBueCkgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=&tokens=false)

Actually, yes - agreed. You're right 🙂

danvk: 👍 I've removed that test in favor of the single use one. We might want to prevent this rule from erroring on unused type parameters, but that seems out of scope for this bug fix.

danvk:
This should be good to go. I have no idea what the "Semantic Breaking Change PR / Validate Breaking Change PR (pull_request_target)" is about, and the [error logs](https://github.com/typescript-eslint/typescript-eslint/actions/runs/9878160728/job/27281450568?pr=9530) aren't illuminating:

```
Error: Cannot find module '@actions/core'
```

JoshuaKGoldberg:
oop #9533 @auvred - looks like they were necessary after all?

We can ignore them in this PR.

auvred:
> oop #9533 @auvred - looks like they were necessary after all?

My bad, sorry :grimacing: 

https://github.com/typescript-eslint/typescript-eslint/pull/9535

bradzacher (approved):

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
jakebailey: Did you see the comment on https://github.com/microsoft/TypeScript/issues/59221#issuecomment-2223721027? These properties are not safe to use if the checker hasn't actually touched the types to evaluate them for some reason.

danvk: Is there a reliable way for a client to access the parts of a mapped type, then?

jakebailey: I am not aware of any "official" way, but staring at the code, I _think_ one could call some other random checker method to get `resolveStructuredTypeMembers` to run on the mapped type, thus populating those fields (assuming the mapped types have `TypeFlags.Objejct`, anyway). So _maybe_ call `getPropertiesOfType` on it beforehand? We're in the danger zone with this...

danvk: Assuming `getPropertiesOfType()` does the trick, then this code should work as intended.

Here's the relevant code:

https://github.com/typescript-eslint/typescript-eslint/blob/16392ab7a4503843ca3cc59831102971754ab9cc/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts#L385-L387

https://github.com/typescript-eslint/typescript-eslint/blob/16392ab7a4503843ca3cc59831102971754ab9cc/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts#L273-L284

Even our implementation of `isMappedType` relies on `typeParameter` being populated. But a few lines earlier we call `type.getProperties()`, which is a [thin wrapper][1] around `getPropertiesOfType`, so we've done what we can to make that the case.

[1]: https://github.com/microsoft/TypeScript/blob/edd08a570a755e00fd58b971aa4a639545ef3950/src/services/services.ts#L849-L851

PR #9536, opened 2024-07-10
feat(eslint-plugin): [no-unnecessary-type-parameters] special case tuples and parameter location arrays as single-use

### LINKED ISSUE #9529: Enhancement: [no-unnecessary-type-parameters] Special case tuples and parameter arrays as a "single use" generic type

https://typescript-eslint.io/rules/no-unnecessary-type-parameters/

### Description

For purposes of counting whether a generic type parameter is used twice, `T[]` / `Array<T>` should only count as using it once. I'm not aware of any realistic scenario where this would lead to a false positive.

We decided to defer this to a follow-up in the initial implementation (see links below). So here's an official request for the follow-up.

### Fail

```TypeScript
declare function getLength<T>(xs: readonly T[]): number;
```

### Pass

```TypeScript
declare function getLength(xs: readonly unknown[]): number;
```

### Additional Info

See https://twitter.com/jfet97/status/1810703565280759964 and https://github.com/typescript-eslint/typescript-eslint/pull/8173#issuecomment-1879695591 for context on why this doesn't work as expected today.

See https://github.com/typescript-eslint/typescript-eslint/pull/8173#discussion_r1461182890 for previous discussion of special-casing `Array` and friends.

See https://github.com/typescript-eslint/typescript-eslint/pull/8173#pullrequestreview-2006886241 for a [contrived example] of how this could yield a false positive for `Map`.

[contrived example]: https://www.typescriptlang.org/play?target=2#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwygGsQBnAWSgAcAeAaQBp4A1APgAoBbagLnktqMWrAJR8AbjizAA3ACg5hEhWrtUIAO79qNUhhhZUAcyapknAEYgYHESPlywePfE7wAvPHVaBu-YZMvcysbdgBtMIByKFRSDWtIpgAWACYAXTT7RWIyAXYAbzl4YvgjEAx2EgBPPj0DYzEgy2t4AB94NFBEQxBgeEKSwfg4DGQYfGqPd09o2PiYSPgAfnhU+D5OkG71WSKSgF8GPeKAehP4AEEICGGQbkMAgiqqMnhYBENEazg+iyhSXrwPAEAAWCFAkFgUGwwJwiFK5XYIgAdMd4JAQLAkf1DmjQBByiBKiAqiJ+mjBiMxvhEFAIAD5INcYMQf9iaTyUMSlTxkg6Qy0cySgCKtUmOI6cgQGSBlzbqNeZxGQcjoNSFgAF4gPgARlVJUQOBgAFEoGAQewwBYmBgQVhSBcYEYZYKjmiwgBlKqWHAQZFYDDWaFGtJ8bHuViuT3eiy+-2BmDBmBpJH64ogVD+MhhskR1zIjNZ0iptHVUg5jyR-NlkuDCUQKXl+DhyOcZH1xu1krRn1+jA4D3+YwAFSgRlD8EiAgAtBAsCREnJ9lkgA

jfet97:
## PR Checklist

- [X] Addresses an existing open issue: fixes #9529

## Overview

Based on initial work from @danvk, this PR special cases `readonly T[]`, `T[]`, `ReadonlyArray<T>`, `Array<T>` and tuples as single use generic types.


danvk:
Nice! Two quick notes:

> this PR special cases `readonly T[]`, `T[]`, `ReadonlyArray<T>` and `Array<T>` as single use generic types.

It also special-cases tuple types. They don't appear in the `SINGULAR_TYPES` Set, but they're special-cased in code.

I tried running this on a repo of mine and found one false positive:

```ts
declare function fetchJsonArray<T>(url: string): Promise<T[]>;
```

This is considered valid by the current rule but invalid with this PR. The behavior should be the same as for:

```ts
declare function fetchJson<T>(url: string): Promise<T>;
```

which is also considered valid.

The bug (which comes from my version of this PR) is that we can't just set `assumeMultipleUses`. We need to consider the existing value. Here's a commit that fixes the issue and adds a test for it: https://github.com/typescript-eslint/typescript-eslint/commit/f871a35f643bbd0e81f93626170e50fdf7799302

jfet97:
@danvk Thanks for the clarifications! I updated the PR with the contents of that commit and added the test about tuples that you wrote and I completely missed.

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -210,6 +210,7 @@ function collectTypeParameterUsageCounts(
   function visitType(
     type: ts.Type | undefined,
     assumeMultipleUses: boolean,
+    isReturnType = false,
```
danvk, line 213: Does `isTypeParameterRepeatedInAST` need a corresponding update?

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
```diff
@@ -211,6 +211,7 @@ ruleTester.run('no-unnecessary-type-parameters', rule, {
     `,
     'declare function get(): void;',
     'declare function get<T>(param: T[]): T;',
+    'declare function factory<T>(): T[];',
```
danvk, line 214: (If we don't have it already) Add a test like:

```ts
declare function returnsTuple<T>(): [T]
```

This should be invalid, unlike returning `T[]` which we're leaving valid.

jfet97: Added :D

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
jfet97: Don't think so. With the proposed change, single uses of Array and ReadonlyArray always defer the check at a later step, the one with type information. If the single use is as parameter we get an error, whereas the single use is as return type it's fine.

jfet97: I added some tests [here](https://github.com/typescript-eslint/typescript-eslint/pull/9536/commits/fe7efb154c010dbb348fe336ae5b9e0fd8192045#diff-bb675ade954bc4c24d9bf6172222cad53e9d22776514ebf4ea8189015b57b3c3R214-R218) to check everything is fine for `Array` and `ReadonlyArray` too.

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
```diff
@@ -230,6 +211,10 @@ ruleTester.run('no-unnecessary-type-parameters', rule, {
     `,
     'declare function get(): void;',
     'declare function get<T>(param: T[]): T;',
+    'declare function factoryArray<T>(): Array<T>;',
```
JoshuaKGoldberg, line 214: From https://github.com/typescript-eslint/typescript-eslint/issues/9529#issuecomment-2223224801:

> Treating array types in a parameter slot (not return type) as singular.

This is in a return type. Folks could reasonably want this like:

```ts
function factoryArray<T>(): Array<T> {
  console.log("Making an array.");
  return [];
}

const array = factoryArray<number>();
array.push(1);
array.push(2);
```

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
```diff
@@ -16,11 +16,6 @@ const ruleTester = new RuleTester({
 
 ruleTester.run('no-unnecessary-type-parameters', rule, {
   valid: [
-    `
-      class ClassyArray<T> {
-        arr: T[];
-      }
-    `,
```
JoshuaKGoldberg, line 23: Err, shouldn't this still be allowed? `ClassyArray` is fine per this rule.

```ts
const classy = new ClassyArray<number>();
classy.arr = [1, 2, 3];
```

JoshuaKGoldberg (changes requested):
Hey thanks for sending this in & iterating on it! I think the direction here might be a big different from the issue? Did I misread something?

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
jfet97: Yep, that's why we allow it

jfet97:
@JoshuaKGoldberg I was following the discussion on the issue, and what I've implemented are the points which everyone agree on.

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
jfet97: "`ClassyArray` should be invalid, because why not just use an array?" Cit. @danvk.

I agree by the way.

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
JoshuaKGoldberg: ...huh, I don't remember why I wanted to comment here. Ignore me. 😄

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
JoshuaKGoldberg: Gotcha. I don't think we got a consensus agreement on allowing `ClassyArray`. `ClassyArray` is, roughly speaking, code that someone might reasonably write. Expanding on https://github.com/typescript-eslint/typescript-eslint/issues/9529#issuecomment-2221917930:

```ts
class LabeledArray<T> {
  arr: T[] | undefined;
  label: string;

  constructor(label: string) {
    this.arr = [];
    this.label = label;
  }
}

const labeledArray = new LabeledArray<string>("letters");
labeledArray.arr.at(0); // string | undefined
```

I think we have consensus on two things:

* Per @danvk's https://github.com/typescript-eslint/typescript-eslint/issues/9529#issuecomment-2223224801, making _truly safe_ things as safe:
  * Treating tuple types as singular
  * Treating array types in a parameter slot (not return type) as singular
* Adding an _opt-in option_ for things like arrays.

`ClassyArray`/`LabeledArray` wouldn't qualify for _"array types in a parameter slot"_.

jfet97: @JoshuaKGoldberg I added some complexity, but now your example isn't considered an error anymore, while the following are at the moment: 

```ts
class Joiner<T extends string | number> {
  join: (els: T[]) => string;
}

class Joiner<T extends string | number> {
  join(els: T[]) {
    return els.map(el => '' + el).join(',');
  }
}

class Joiner<T extends string | number> {
  join = (els: T[]) => {
    return els.map(el => '' + el).join(',');
  };
}
```

jfet97: Should

```ts
class ClassyArray<T> {
        arr: [T];
        label: string;
}
```

be allowed?

jfet97:
We are ignoring interfaces, but we should error out in cases like the following ones as well:

```ts
// this is equivalent to a function type
interface Joiner<T extends string | number> {
    join(els: T[]): string;
}

// methods and functional property
interface Joiner<T extends string | number> {
   join: (els: T[]) => string;
}

interface Joiner2<T extends string | number> {
  join(els: T[]): string;
}
```

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
danvk: I find @JoshuaKGoldberg's `LabeledArray` example persuasive. Let's allow it.

`Joiner` should be disallowed since the `T[]` only appears in a parameter slot.

`ClassyTuple` should be allowed. I don't think the tuple-ness is really relevant to it, because shouldn't this also be allowed?

```ts
class LabeledBox<T> {
    value: T;
    label: string;
}
```

danvk:
> We are ignoring interfaces, but we should error out in cases like the following ones as well:

Does TS ever infer the type parameter for an interface? I think you need some value-level (not type-level) code for this rule to be relevant.

JoshuaKGoldberg:
> Does TS ever infer the type parameter for an interface

I can't find a way to make it. That would be a nice feature, though...

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
danvk: I see the point about re-usability, but if we accept that as a rationale then I'm not sure we'd have any invalid cases left thanks to [instantiation expressions](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#instantiation-expressions):

```ts
function getLength<T>(array: T[]) {
  return array.length;
}

const getStringArrayLength = getLength<string>;
const getLogLevelArrayLength = getLength<"info" | "warn">;
```

jfet97: Yeah, why should the first be allowed by this rule while the second one is not?

```ts
namespace WithClass {

    declare class LeveledLogger<Level extends string> {
        log(level: Level, ...args: string[]): void;
    }

    const logger = new LeveledLogger<"info" | "warn">();

    logger.log("info", "a"); // Ok
    logger.log("warn", "a"); // Ok

    logger.log("error", "a"); // Not ok

}

namespace WithFunction {
    declare function leveledLogger<Level extends string>
        (level: Level, ...args: string[]): void

    const logger = leveledLogger<"info" | "warn">;

    logger("info", "a"); // Ok
    logger("warn", "a"); // Ok

    logger("error", "a"); // Not ok
}
```

jfet97:
> Does TS ever infer the type parameter for an interface? I think you need some value-level (not type-level) code for this rule to be relevant.

Could you please make an example of it, even if it's not working? I'm not sure I got what you mean. @danvk

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
JoshuaKGoldberg: _Blurgh_ instantiation expressions! I'd forgotten about them. I've never seen someone use them in the wild.

The rule is essentially useless if it considers instantiation expressions a potential multiple use of parameters. That would invalidate roughly all functions! I think there's no reasonable way to support them and this rule. They'd probably be better as something explicitly mentioned in the rule's docs (#9680). cc @kirkwaiblinger 

In my mind, the difference between classes and function instantiation expressions is that classes are _made_ to be reused. I've written code like `LeveledLogger` before and seen others opt into it. But I can see an argument that it's kid of niche. What do you all think? Also I'd want to hear from @typescript-eslint/triage-team

jfet97:
@JoshuaKGoldberg is there something I can do to make this PR land? Are other tests needed? IMO only the documentation was left behind, but I'm not 100% sure on how to properly communicate this change.

JoshuaKGoldberg:
I believe I'm waiting for input from you, @danvk, and/or @typescript-eslint/triage-team on https://github.com/typescript-eslint/typescript-eslint/pull/9536/files#r1715303329?

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
kirkwaiblinger: This is brain-melty stuff. 🫠 I've been thinking about this for far too long and not come up any concrete opinion yet 😆 

One question I have is - is there any difference in behavior between an instantiation expression and providing explicit type arguments? I see that they're represented differently in the AST, but, to my mind, 
```ts
const logger = new LeveledLogger<"info" | "warn">();
```
is the same as
```ts
const logger = new (LeveledLogger<"info" | "warn">)();
```
right?

When we directly instantiate the class with type arguments (or call a function with type arguments, like `f<T>()`) we're doing exactly the same thing as if we aliased the instantiation expression first as in `const ft = f<T>; ft();`, right?

kirkwaiblinger: The reason I ask is that my _very very tentative_ instinct is to say that the current approach in the rule is basically incompatible with classes/functions intended to be used with instantiation expressions or with explicit type arguments.

JoshuaKGoldberg: > to my mind, ... has no observable differences to ...

👍 agreement from me.

> current approach in the rule is basically incompatible with classes/functions intended to be used with instantiation expressions or with explicit type arguments

👍 agreement with me. For those cases, I don't see a way to make the rule actually useful at all.

kirkwaiblinger: Cool.

So what gives me pause is that the logger case is _potentially useful_ and _doesn't introduce type unsafety_ because the type is used in parameter position on a method, rather than return type position. So, yes, the class could be written without, but it doesn't seem to incur any of the negatives that the rule kind of intends to lint against. 🤔 I'm having a hard time thinking about whether this analysis can be generalized into something we can lint for or whether it should be handled by users on a case by case basis. 

Similarly, I don't see how 
```ts
function f<T>(x: T): void {
    // any implementation
}
```
could _ever_ incur type unsafety. But it does in theory have niche use in instantiation expressions/explicit type argument scenarios. So - do we want to allow this (possibly via an option?)? Or just document it as a limitation?

kirkwaiblinger: The trouble with the type unsafety analysis is - I'm not sure that you can incur type unsafety without doing something else that is also problematic. The canonical example, 
```ts
function totallyUnsafe<T>(): T {
    return JSON.parse('"anything"');
}
```
can be defended against with `no-unsafe-return`. And, similarly, anything that uses `as T` is overtly unsafe regardless of the single-use generic. 

So, the question in my mind is - does this rule _really_, _really_, matter for type safety reasons? Or is it a purely stylistic rule? 

I think deciding on an answer to that question may inform the judgement calls we'll make here and elsewhere.

JoshuaKGoldberg: I see it as for both stylistic and type safety. The removal of unnecessary constructs (`function log<T>(value: T): void`) is useful, as is hiding assertions (`function parse<T>(): T`).

> whether this analysis can be generalized into something we can lint for or whether it should be handled by users on a case by case basis.

So it sounds like, for now, the conclusion is _no_: that's just not a pattern compatible with this rule. So that'd be a good FAQ. Am I misreading the conclusion?

JoshuaKGoldberg: 👋 following up @jfet97, @kirkwaiblinger and I chatted and we're still on this opinion.  That in _this_ PR the only change to be made is special casing tuples and parameter arrays as single-use.

JoshuaKGoldberg:
👋 @jfet97: I _think_ all the open questions are answered. But this PR has a lot of threads and I'm not confident I didn't miss one 😅. Is there anything else on your mind?

jfet97:
Nothing else! Could you please go through the tests I added and change/modify what's not aligned with the "new" goal of this PR? So that I'll change the implementation accordingly.

JoshuaKGoldberg:
👍 good idea. Pushed, how does it look to you now?

jfet97:
@JoshuaKGoldberg there are some problems with the tests that still fail:

1) We said that we wanna allow declarations like `declare function makeReadonlyTupleMulti<T>(): readonly [T, T]`, that's failing, because they are kind of builders. In fact, tests contradict themselves because `declare function makeArray<T>(): T[]` and some friends are under the `valid` cases instead. I'm confused 😅

2) There are missing suggestions, but I'd say you were expecting that

3) Declarations like `function takeTupleMulti<T>(input: [T, T]): void` should be considered valid, I think we talk about this in one of the threads. You are, in fact, using the type parameter to express a constraint: the two elements of the tuple must have the same type. You can't just replace it with `function takeTupleMulti(input: [unknown, unknown]): void`, this won't give you the same guarantees in the implementation because it won't force you to call the function with a tuple containing two elements of the same type.

JoshuaKGoldberg:
> 1. We said that we wanna allow declarations like `declare function makeReadonlyTupleMulti<T>(): readonly [T]`, that's currently failing though, because they are kind of builders. In fact, tests contradict themselves because `declare function makeArray<T>(): T[]` and some friends are under the `valid` cases instead. I'm confused 😅

You and me both 😜 this whole situation is hard to keep track of...

The difference between `(): readonly [T]` and `(): T[]` is:
* `(): readonly [T]`: consumers can't add their own values of type `T`. There's no `.push(...)` or `[0] = ...` allowed. It's purely a container for whatever the function gives it internally.
* `(): T[]` _does_ allow consumers to modify it with `.push()`, etc. So the type parameter `T` _does_ impact what users can make safely.

Example of a `(): T[]` that is reasonable to write:

```ts
// (hypothetically in e.g. some project that's tracking object instantiations)
function makeArray<T>(): T[] {
    const values: T[] = []
    console.log("New array created");
    return values;
}
```

> There are missing suggestions, but I'd say you were expecting that

Ha, yeah, I didn't want to be prescriptive in them. In case you ended up changing the details. Also I was being lazy. 😄 

3) Declarations like `function takeTupleMulti<T>(input: [T, T]): void` should be considered valid, I think we talked about this in one of the threads. You are, in fact, using the type parameter to express a constraint: the two elements of the tuple must have the same type. You can't just replace it with `function takeTupleMulti(input: [unknown, unknown]): void`, this won't give you the same guarantees in the implementation because it won't force you to call the function with a tuple containing two elements of the same type.

You're right! I forgot about that one. Mistake on my end. I'll fix up the tests.

jfet97:
@JoshuaKGoldberg Other than missing suggestions there are just a couple of failing tests and I'll try to argue against them 😝 

They are:

```ts
declare function makeReadonlyTupleMulti<T>(): readonly [T, T];

declare function makeReadonlyTupleMultiNullish<T>(): readonly [
  T | null,
  T | null,
];      
```

I get why we can't consider them as builders, even if the latter one could return `[null, null]` and call it a day. OTOH the purpose of this rule is to count the number of references to a type parameter, and that's what the overall code does. Even if writing a declaration like one of those is non-sensical, there are in fact at least two usages of the type parameter `T` in each of those.

The current version of this PR doesn't consider any type reference as multiple usage of a type variable anymore, like eslint is doing now instead. When a tuple or an array type comes into play, we do not assume multiple usages by default anymore for them unless the array/tuple is in a return position and it's mutable.

But this does not interfere with the overall usages counting. A tuple isn't considered as multiple usage anymore, good, so the code actually counts how many time a type parameter is used inside it. 

I fear special casing this would go outside of the scope of this rule; the type parameter is used multiple times, in a stupid manner I'd say, but it is.

JoshuaKGoldberg:
> OTOH the purpose of this rule is to count the number of references to a type parameter, and that's what the overall code does. 

I interpreted the counting as just an implementation detail. The rule's general purpose is to flag unnecessary type parameters. And in unless there's a use case I haven't thought of, those code snippets have unnecessary `T`s.

The way I understand the algorithm is: it crawls through the available info until it finds the second justification for making the type parameter necessary. Those `readonly` `T`s in the tuples don't do anything to justify the type parameter. So to me this is a direct conceptual bug in how the algorithm is implemented.

jfet97:
What about `declare function makeFoo<T>(): Readonly<{ 0: T, 1: T }>` then? In principle it should be treated the same as the tuple case, because even if there are two occurrences of the type parameter they are readonly and the overall thing becomes useless, or unsafe.

And you can complicate the return type how much you'd like, by nesting mutable and non mutable object types where `T` may or may not appear. Counting occurrences can't be treated as just an implementation detail. It's the only reasonable general way to check if the type parameter is not used just once.

We can have exceptions of course like arrays of tuples, but there is a limit to that I'd say.

JoshuaKGoldberg:
Ha, you're right, I would say that that is an unnecessary type parameter. And you're right that handling all sorts of `readonly ...` / `Readonly<...>` cases is a whole extra ordeal. I think you've swayed me into thinking that that should be a separate followup issue.

cc @typescript-eslint/triage-team

jfet97:
Or maybe a different rule? I mean that article by @danvk was really focused on counting references if I'm not mistaken. I have not written the majority of the code of this rule, I suppose @danvk started it? That's why everything seems to be based on counting. I'd say that's more than reasonable as a general strategy.

Now, let's spend three more months to define the precise meaning of `unnecessary` /s

JoshuaKGoldberg:
Btw @jfet97 since it's been a couple of weeks without other team members weighing in, I think you're pretty solidly unblocked from going forward with your proposal. I.e. considering handling all sorts of `readonly ...` / `Readonly<...>` cases as separate.

jfet97:
@JoshuaKGoldberg I'd say only suggestions are missing at this point, if you like my changes to the documentation.

JoshuaKGoldberg (commented):
Yup, LGTM - thanks for the iteration!

packages/eslint-plugin/docs/rules/no-unnecessary-type-parameters.mdx
```diff
@@ -76,16 +76,22 @@ function getProperty<T, K extends keyof T>(obj: T, key: K) {
 ## Limitations
 
 Note that this rule allows any type parameter that is used multiple times, even if those uses are via a type argument.
-For example, the following `T` is used multiple times by virtue of being in an `Array`, even though its name only appears once after declaration:
+For example, the following `T` is used multiple times by virtue of being in an `Set`, even though its name only appears once after declaration:
```
JoshuaKGoldberg, line 79: These no longer need to be changed now that this PR doesn't touch _return_ location arrays. I'll touch these docs up..

```suggestion
For example, the following `T` is used multiple times by virtue of being in an `Array`, even though its name only appears once after declaration:
```

JoshuaKGoldberg, line 79: Now that this PR doesn't touch _return_ location arrays, only _parameter_ location ones, these docs don't need to change anymore. I'll go ahead and revert the now-unnecessary changes.

```suggestion
For example, the following `T` is used multiple times by virtue of being in an `Array`, even though its name only appears once after declaration:
```

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -380,9 +395,31 @@ function collectTypeParameterUsageCounts(
...
     // Tuple types like `[K, V]`
     // Generic type references like `Map<K, V>`
-    else if (tsutils.isTupleType(type) || tsutils.isTypeReference(type)) {
+    else if (tsutils.isTypeReference(type)) {
       for (const typeArgument of type.typeArguments ?? []) {
-        visitType(typeArgument, true);
+        // currently, if we are in a "class context", everything is accepted
+        let thisAssumeMultipleUses = fromClass || assumeMultipleUses;
+
+        if (!thisAssumeMultipleUses) {
```
kirkwaiblinger, line 403: I had a really hard time parsing this section on first read.... What do you think of something along these lines? Helps keep the logic simpler by not intermixing the tuple and array logic:

```ts
        thisAssumeMultipleUses ||= ((): boolean => {
          // special cases - readonly arrays/tuples are considered only to use the
          // type parameter once. Mutable arrays/tuples are considered to use the
          // type parameter multiple times if and only if they are returned.
          if (checker.isTupleType(type.target)) {
            const isReadonlyTupleType =
              // @ts-expect-error -- why is this a type error?
              (type.target as ts.TupleType).readonly;

            return isReturnType && !isReadonlyTupleType;
          }

          if (checker.isArrayType(type.target)) {
            const isReadonlyArray =
              (type.symbol as ts.Symbol | undefined)?.getName() ===
              'ReadonlyArray';

            return isReturnType && !isReadonlyArray;
          }

          // other kind of type references always count as multiple uses
          return true;
        })();
```

kirkwaiblinger (changes requested):

kirkwaiblinger:
There's a parenthesization bug around the suggestion fixer - it fixes this:
```ts
class Joiner {
  join<T extends string | number>(els: T[]) {
    return els.map(el => '' + el).join(',');
  }
}
```
to this:
```ts
class Joiner {
  join(els: string | number[]) {
    return els.map(el => '' + el).join(',');
  }
}
```

That might be worsened by this PR due to it now reporting on `T[]` syntax, but I think it's an existing bug (you could probably come up with a test case with some wonky intersections/unions precedence), so we can handle it in a followup.

kirkwaiblinger (dismissed):
Exciting!

kirkwaiblinger:
Filed https://github.com/typescript-eslint/typescript-eslint/issues/10894 to follow up on https://github.com/typescript-eslint/typescript-eslint/pull/9536#issuecomment-2685621735

JoshuaKGoldberg (approved):
So relieved to have this through 🫠 thanks so much everyone for working on this! 🚀

JoshuaKGoldberg:
CI failures are unrelated on `main` too. 🚢

PR #9540, opened 2024-07-11
chore: enable no-unreachable-loop

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes part of #9523 

## Overview

Enable `no-unreachable-loop` ([docs](https://eslint.org/docs/latest/rules/no-unreachable-loop))


packages/eslint-plugin/tests/rules/no-unsafe-return.test.ts
```diff
@@ -127,6 +127,7 @@ function foo(): Set<number> {
         return [] as any[];
       }
     `,
+    'const foo: (() => void) | undefined = () => 1;',
```
abrahamguo, line 130: `functionType.getCallSignatures()` can return an empty array, but that code path was not being tested. This ensures that the code path is now covered.

packages/eslint-plugin/src/rules/no-mixed-enums.ts
```diff
@@ -170,18 +170,15 @@ export default createRule({
...
-              ? tsutils.isTypeFlagSet(
-                  typeChecker.getTypeAtLocation(member.initializer),
-                  ts.TypeFlags.StringLike,
-                )
-                ? AllowedType.String
-                : AllowedType.Number
-              : AllowedType.Number;
-          }
-        }
+        const [{ initializer }] = (declarations[0] as ts.EnumDeclaration)
```
Josh-Cena, line 173: Love this because this exposes a bug (not sure if it's intentional)! Checking `declarations[0]` is almost always a bug.

packages/eslint-plugin/tests/rules/no-unsafe-return.test.ts
abrahamguo: This was brought to light because when I initially refactored the rule to assume that `getCallSignatures()[0]` was always defined, the tests passed, but the rule crashed when linting this repo, therefore showing that a test was missing.

packages/eslint-plugin/src/rules/no-mixed-enums.ts
abrahamguo: Is this truly a bug? Is it possible for a symbol to have multiple declarations, that would be different from each other?

Josh-Cena: No; if a symbol returns multiple declarations, that means these sites are merged together and should be treated as one. That's my understanding; I haven't verified it.

kirkwaiblinger (dismissed):
Appreciate you being proactive about adding that test 🙂

Josh-Cena (approved):

kirkwaiblinger (approved):

JoshuaKGoldberg (dismissed):
A `chore` so good, it got promoted to a `fix(eslint-plugin)`. Nice! 💪

Josh-Cena:
@JoshuaKGoldberg I don't think there are behavior differences with this PR? The logic of no-unsafe-return wasn't buggy, I think (well it could be, since only caring about the first signature still seems fishy).

JoshuaKGoldberg:
Oh, gotcha - I misinterpreted. Never mind then. 🙂

Josh-Cena (approved):

PR #9563, opened 2024-07-14
chore: enable radix

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes part of #9523 

## Overview

Enable `radix` ([docs](https://eslint.org/docs/latest/rules/radix))


eslint.config.mjs
```diff
@@ -211,6 +211,7 @@ export default tseslint.config(
       'one-var': ['error', 'never'],
       'prefer-arrow-callback': 'error',
       'prefer-rest-params': 'error',
+      radix: 'error',
```
JoshuaKGoldberg, line 214: Huh, from reading https://eslint.org/docs/latest/rules/radix:

> On the other hand, if the code is targeting only ES5-compliant environments passing the radix `10` may be redundant. In such a case you might want to disallow using such a radix.

Maybe we don't need this?

Josh-Cena: `parseInt` still auto-detects `0x`, so I'd say this is still nice to have, especially when this PR is moving away from `parseInt` to `Number`, which I vastly prefer.

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -160,7 +160,7 @@ type RequireKeys<
 > = ExcludeKeys<Obj, Keys> & { [k in Keys]-?: Exclude<Obj[k], undefined> };
 
 function getEnumNames<T extends string>(myEnum: Record<T, unknown>): T[] {
-  return Object.keys(myEnum).filter(x => isNaN(parseInt(x))) as T[];
+  return Object.keys(myEnum).filter(x => isNaN(+x)) as T[];
```
kirkwaiblinger, line 163: Why wouldn't we use this?
```suggestion
  return Object.keys(myEnum).filter(x => isNaN(parseInt(x, 10))) as T[];
```

abrahamguo: I personally prefer using an operator instead of a function when available, especially since the `+` operator works more safely than `parseInt` does by default. However, I'm happy to change it if you'd like

kirkwaiblinger: Gotcha, I'm cool with it, just wanted to check to make sure there was intentionality 👍 . I can never quite keep straight the differences between `parseInt()` and `Number()`/`+` but I trust and defer to your judgement here 🙂.

I would personally say I do much prefer the explicit style....
```suggestion
  return Object.keys(myEnum).filter(x => isNaN(Number(x))) as T[];
```
Up to you though!

kirkwaiblinger (dismissed):

packages/eslint-plugin/src/util/misc.ts
Josh-Cena: I also prefer `Number(x)` :)

Issue #9571, opened 2024-07-17
Performance: parserOptions.projectService no longer outperforms parserOptions.project


JoshuaKGoldberg:
## Overview

When we first started working with the new [`parserOptions.projectService` (formerly `parserOptions.EXPERIMENTAL_useProjectService`)](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8-beta/#project-service), it outperformed equivalent `parserOptions.project` setups by ~10-15%. We still sometimes have users report to us that it improved their linting speed

However, at some point since then, we've started to notice _worse_ performance of `parserOptions.projectService` in our own measurements. https://github.com/typescript-eslint/performance contains tooling to test this. From https://github.com/typescript-eslint/performance/blob/45f77c00c6d4384d466c39e8eb633a76f33dd019/README.md#results:

```plaintext
┌───────┬──────────────────────┬──────────────────────┬──────────────────────┬──────────────────────┐
│ files │ project (even)       │ project (references) │ service (even)       │ service (references) │
┼───────┼──────────────────────┼──────────────────────┼──────────────────────┼──────────────────────┤
│ 128   │ '1.149 s ±  0.030 s' │ '1.135 s ±  0.008 s' │ '1.178 s ±  0.010 s' │ '1.736 s ±  0.012 s' │
│ 512   │ '1.636 s ±  0.009 s' │ '1.656 s ±  0.004 s' │ '1.895 s ±  0.007 s' │ '2.613 s ±  0.020 s' │
│ 1024  │ '2.353 s ±  0.013 s' │ '2.399 s ±  0.016 s' │ '3.130 s ±  0.017 s' │ '4.034 s ±  0.061 s' │
┴───────┴──────────────────────┴──────────────────────┴──────────────────────┴──────────────────────┘
```

Summarizing those results, there seem to be two areas of slowness. Comparing `project` to `projectService` at ~1024 files:

* Overhead per-file (`(even)`), jumping from ~2.35s to ~3.13s 
* Overhead switching to project references (`(references)`): jumping from ~2.4s to ~4.0s`

⚠️ **These measurements might be flawed.** We are not sure there's not some external data polluting the results. This issue might just be tracking that we're measuring them incorrectly.

**Also note** that `parserOptions.project` has been in production for years, with many performance optimizations around it. `parserOptions.projectService` is very new and not yet stable. There is likely much more "low-hanging fruit" to optimize around `parserOptions.projectService`.

Relevant past issues:

[...]

PR #9586, opened 2024-07-18
fix(typescript-estree): disable includePackageJsonAutoImports in ProjectService

jakebailey:
## PR Checklist

- [x] Addresses an existing open issue: for #9571

## Overview

Per https://github.com/microsoft/TypeScript/issues/59338#issuecomment-2236857239, disable this so that the auto import projects are not used.

### DIFF
--- a/packages/typescript-estree/src/create-program/createProjectService.ts
+++ b/packages/typescript-estree/src/create-program/createProjectService.ts
@@ -67,6 +67,12 @@ export function createProjectService(
     jsDocParsingMode,
   });
 
+  service.setHostConfiguration({
+    preferences: {
+      includePackageJsonAutoImports: 'off',
+    },
+  });
+
   if (options.defaultProject) {
     let configRead;
 
@@ -93,12 +99,6 @@ export function createProjectService(
         )}`,
       );
     }
-
-    service.setHostConfiguration({
-      preferences: {
-        includePackageJsonAutoImports: 'off',
-      },
-    });
 
     service.setCompilerOptionsForInferredProjects(
       (
--- a/packages/typescript-estree/src/create-program/createProjectService.ts
+++ b/packages/typescript-estree/src/create-program/createProjectService.ts
@@ -67,12 +67,6 @@ export function createProjectService(
     jsDocParsingMode,
   });
 
-  service.setHostConfiguration({
-    preferences: {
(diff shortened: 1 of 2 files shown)


jakebailey:
The heck

packages/typescript-estree/src/create-program/createProjectService.ts
```diff
@@ -94,6 +94,12 @@ export function createProjectService(
       );
     }
 
+    service.setHostConfiguration({
+      preferences: {
+        includePackageJsonAutoImports: 'off',
+      },
+    });
```
JoshuaKGoldberg, line 101: @jakebailey Is this intentionally inside the `if (options.defaultProject) {`?

jakebailey: Nope...

jakebailey: 2 space indent is way too small for me to notice :(((

bradzacher:
> auto import projects

What are those..? Will that cause us to lose any type coverage?

JoshuaKGoldberg:
@bradzacher this is our side of https://github.com/microsoft/TypeScript/issues/59338#issuecomment-2237357522. There's more work to be done, but this should be a good improvement to start.

JoshuaKGoldberg (approved):
Thanks (as usual) for hopping on this @jakebailey + @sheetalkamat! ❤️

jakebailey:
> What are those..? Will that cause us to lose any type coverage?

TS needs a way to analyze code for auto imports, but it can't do that in the main Program for the current code because that would end up loading stuff like potentially unreferenced ambients. So, it instead loads those other libraries in a AutoImportProviderProject separately so they can be analyzed and provide completions / fixes. There's no need for that in your case, so it's just wasted work.

bradzacher:
> this is our side of microsoft/TypeScript#59338 (comment).

The issue doesn't say what they are!
What are auto import projects..?
Will turning them off cause us to lose any type coverage?

jakebailey:
Did you see my explanation above? This will have no effect on your usage at all.

bradzacher:
I didn't see it cos the GH mobile app is so bad at doing anything useful 😅

Makes sense! Kill it!

bradzacher (approved):

PR #9595, opened 2024-07-19
feat(eslint-plugin): [return-await] promote to recommended: strict

### LINKED ISSUE #8667: Configs: include `return-await` in `strict-type-checked` preset

### Description

I recently debugged a difficult bug in my code where I was try-catching an async call that I was returning right away.

```ts
async function asyncFoo() {
  try {
    return asyncBar() // no await
  } catch (err) {
    // so my error handling code here wasn't getting hit
  }
}
```

I was surprised when I finally figured this out, because I love and rely on e.g. [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises/) to catch missing `await`s, but it hadn't caught this one.

I found that [`return-await`](https://typescript-eslint.io/rules/return-await/) exists to help here, but was then also surprised that it wasn't included in the [`recommended-type-checked`](https://typescript-eslint.io/users/configs#recommended-type-checked) preset that we use.

I thought the whole purpose of that preset was to at least help ensure _correctness_ (separate from style) — and this was definitely a correctness issue.

I've now enabled `return-await` in my codebase, and am feeling great. But can I suggest adding this to `recommended-type-checked` to prevent other developers from hitting easily preventable bugs like this?

Thank you for your consideration and for a great tool!

### Impacted Configurations

[`recommended-type-checked`](https://typescript-eslint.io/users/configs#recommended-type-checked)

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8667

## Overview

Adds `return-await` to the `strict-type-checked` preset shared config.

This is not a breaking change, as `strict` configs are explicitly allowed to be modified in minor versions. But targeting to `v8` to be friendly.

💖

### DIFF
--- a/packages/eslint-plugin/src/rules/return-await.ts
+++ b/packages/eslint-plugin/src/rules/return-await.ts
@@ -29,6 +29,7 @@ export default createRule({
   meta: {
     docs: {
       description: 'Enforce consistent awaiting of returned promises',
+      recommended: 'strict',
       requiresTypeChecking: true,
       extendsBaseRule: 'no-return-await',
     },
--- a/packages/eslint-plugin/src/configs/strict-type-checked-only.ts
+++ b/packages/eslint-plugin/src/configs/strict-type-checked-only.ts
@@ -64,6 +64,8 @@ export = {
         allowNever: false,
       },
     ],
+    'no-return-await': 'off',
+    '@typescript-eslint/return-await': 'error',
     '@typescript-eslint/unbound-method': 'error',
     '@typescript-eslint/use-unknown-in-catch-callback-variable': 'error',
   },
--- a/packages/eslint-plugin/src/configs/strict-type-checked.ts
+++ b/packages/eslint-plugin/src/configs/strict-type-checked.ts
@@ -97,6 +97,8 @@ export = {
         allowNever: false,
       },
     ],
+    'no-return-await': 'off',
+    '@typescript-eslint/return-await': 'error',
     '@typescript-eslint/triple-slash-reference': 'error',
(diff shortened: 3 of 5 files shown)


packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -29,6 +29,7 @@ export default createRule({
   meta: {
     docs: {
       description: 'Enforce consistent awaiting of returned promises',
+      recommended: 'strict',
```
JoshuaKGoldberg, line 32: @kirkwaiblinger surfacing what you mentioned a bit back in a DM: that we might want to customize settings for this rule in the strict configs. I'm interpreting https://github.com/typescript-eslint/typescript-eslint/issues/8667#issuecomment-2148569319 to mean that:

* For `strict`: we want `in-type-checked`, as it avoids stylistic things
* For `stylistic`: that'd be a separate issue

...and since this already has `defaultOptions: ['in-try-catch']`, we're cleared to just say `recommended: 'strict'`. Is that right?

kirkwaiblinger: negative - my understanding was that the only thing that we are currently proceeding with is putting `return-await` in the `strict` _with option `error-handling-correctness-only`_.

JoshuaKGoldberg: Right! Makes sense, thanks.

Since that's a little bit more plumbing, as you suggested it would be, I'm going to close this one out & let you take lead. You've already taken the lead on so much around it & other async things, I'd rather not butt in.

PR #9603, opened 2024-07-22
feat(rule-tester): switched to flat config

### LINKED ISSUE #9017: Enhancement(rule-tester): FlatRuleTester -> RuleTester

### Relevant Package

rule-tester

### Description

Splitting out of #8211 -> #8969: this issue tracks any updates we'll need corresponding to https://github.com/eslint/eslint/pull/17922.

### Additional Info

_No response_

JoshuaKGoldberg:
BREAKING CHANGE:
Switches `RuleTester` from legacy configs to flat configs.

## PR Checklist

- [x] Addresses an existing open issue: fixes #9017

## Overview

Switches the internals of our forked `RuleTester` to work with `new Linter({ configType: 'flat' });` instead of `configType: 'eslintrc'`. This is a pretty substantial breaking change internally:

* `this.#linter` is no longer able to `defineRule()` ad hoc; instead, the `config: TesterConfigWithDefaults` has to have `plugins` including the rule.
* Because the `@/` plugin can't be redefined, a `@rule-tester/` prefix is added in front of rule names.
* Per flat config, any `parser`, `parserOptions`, and other language-specific settings are moved to `languageOptions`. 
  * This caused a lot of churn in our test files
* Per flat config, `parser` has to be the actual parser, not a string descriptor.
  * As a convenience, I formally defaulted it to our parser
  * I went ahead and removed all the redundant `parser: '@typescript-eslint/parser'`, `parserOptions: { sourceType: 'module' }`, and other seemingly unnecessary options from our test files
* Marks the `@typescript-eslint/utils` `RuleTester` re-export as deprecated

I think we'll want to follow this up with added info in the v8 beta blog post. See our `packages/eslint-plugin(-internal)/**/*.test.ts` file changes. There's a fair amount users will have to do:

* Nest `lib`, `parser`, `parserOptions`, `sourceType`, and any other parser options inside `languageOptions`
* Switch `parser` from a `string` to the actual parser
* The default `ecmaVersion` and `sourceType` are now `'latest'` and `'module'`, respectively

💖


bradzacher:
> Per flat config, any parser and parserOptions, and other language-specific settings are moved to languageOptions.

Thought: I wonder if we should have hoisted up our stuff / the common stuff to avoid churn. Cos this will mess with all the ecosystem as well. I know that was their intent but it is going to block plugins hard.

JoshuaKGoldberg:
> > Per flat config, any parser and parserOptions, and other language-specific settings are moved to languageOptions.
> 
> Thought: I wonder if we should have hoisted up our stuff / the common stuff to avoid churn. Cos this will mess with all the ecosystem as well. I know that was their intent but it is going to block plugins hard.

Yeah I don't think we're in a good place to do anything here. +1 that this will be work for plugins, but:

* Switching our `RuleTester` from a rough drop-in to something noticeably different from the core config system would be a lot of work
* Most plugins will just need to wrap some stuff in `languageOptions`

packages/rule-tester/src/RuleTester.ts
```diff
@@ -66,15 +66,61 @@ const DUPLICATE_PARSER_ERROR_MESSAGE = `Do not set the parser at the test level
  * the initial default configuration
  */
 const testerDefaultConfig: Readonly<TesterConfigWithDefaults> = {
-  parser: TYPESCRIPT_ESLINT_PARSER,
-  rules: {},
   defaultFilenames: { ts: 'file.ts', tsx: 'react.tsx' },
+  languageOptions: {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-require-imports
+    parser: require(TYPESCRIPT_ESLINT_PARSER),
```
bradzacher, line 72:
```suggestion
    // eslint-disable-next-line @typescript-eslint/no-require-imports
    parser: require(TYPESCRIPT_ESLINT_PARSER) as typeof parser,
```

packages/rule-tester/src/RuleTester.ts
```diff
@@ -115,11 +161,16 @@ function getUnsubstitutedMessagePlaceholders(
 }
 
 export class RuleTester extends TestFramework {
+  // instead of creating a hard dependency, just use a soft require
+  // a bit weird, but if they're using this tooling, it'll be installed
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  readonly #defaultParser = require(
+    TYPESCRIPT_ESLINT_PARSER,
+  ) as typeof ParserType;
```
bradzacher, line 169: we've got this `require` plus the `require` up on L72.
we should extract it to a global and reuse the `require`.

bradzacher (approved):

PR #9629, opened 2024-07-24
chore: enable unicorn/no-array-for-each

abrahamguo:
## PR Checklist

## Overview

Enable `unicorn/no-array-for-each` ([docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md)).

### DIFF
--- a/packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
+++ b/packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
@@ -118,9 +118,11 @@ export default createRule<Options, MessageIds>({
           const extraComments = new Set(
             context.sourceCode.getCommentsInside(lhs.parent),
           );
-          context.sourceCode
-            .getCommentsInside(lhs.typeArguments)
-            .forEach(c => extraComments.delete(c));
+          for (const c of context.sourceCode.getCommentsInside(
+            lhs.typeArguments,
+          )) {
+            extraComments.delete(c);
+          }
           context.report({
             node,
             messageId: 'preferConstructor',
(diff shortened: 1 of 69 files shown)


packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
```diff
@@ -118,9 +118,11 @@ export default createRule<Options, MessageIds>({
...
             context.sourceCode.getCommentsInside(lhs.parent),
           );
-          context.sourceCode
-            .getCommentsInside(lhs.typeArguments)
-            .forEach(c => extraComments.delete(c));
+          for (const c of context.sourceCode.getCommentsInside(
+            lhs.typeArguments,
+          )) {
+            extraComments.delete(c);
+          }
```
Josh-Cena, line 125: This is one of the cases where I strongly believe the original case is more readable and thus the rule is not useful to be enforced.

abrahamguo: I think I agree. And enabling this rule did not surface any type-related issues

JoshuaKGoldberg: Yeah it's a pity, a lot of the other cases I do like more, but this one not so much... it would be nice if there'd be a way to encodify this. Maybe something around avoiding changing places with this kind of chaining?

abrahamguo: Yeah, I agree. Unfortunately `no-array-for-each` has no configuration options.

At work I do use `no-array-for-each`, because I feel like the upsides of better consistency and better readability _most of the time_ is better than the downside of worse readability occasionally.

abrahamguo:
Me too. At work I do use no-array-for-each, because I feel like the upsides of better consistency and better readability most of the time is better than the downside of worse readability occasionally.

PR #9640, opened 2024-07-25
chore: enable unicorn/no-array-reduce

abrahamguo:
## PR Checklist

## Overview

Enable `unicorn/no-array-reduce` ([docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-reduce.md)).
Note that `no-array-reduce` does allow ["simple operations"](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-reduce.md#allowsimpleoperations), which it defines as [a body that is only a `BinaryExpression`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/no-array-reduce.js#L29), so those were left unchanged.
I strongly feel that every place that was refactored has better readability than it did before.
In _most_ places, there was a more appropriate method to use, such as `Array.flatMap` or `Object.fromEntries`.
In other places, it was unnecessarily confusing to return the accumulator every iteration when the accumulator never changes, and so it was better to use a `const` variable in conjunction with a `for..of` loop.

I recommend selecting `Hide whitespace` when looking at this diff 😄


abrahamguo:
I recommend selecting `Hide whitespace` when looking at this diff 😄

packages/repo-tools/src/generate-sponsors.mts
```diff
@@ -118,25 +118,19 @@ async function main(): Promise<void> {
...
+
+  const totalDonationsById: Record<string, number> = {};
+  for (const member of collective.members.nodes) {
     const name = member.account.name || member.account.id;
-    accumulator[name] ||= 0;
-    accumulator[name] += member.totalDonations.valueInCents;
-    return accumulator;
-  }, {});
+    totalDonationsById[name] =
+      (totalDonationsById[name] ?? 0) + member.totalDonations.valueInCents;
```
kirkwaiblinger, line 132: Not sure that changing the `+=` logic is relevant 🤔 

Revert?
```suggestion
    totalDonationsById[name] ||= 0;
    totalDonationsById[name] += member.totalDonations.valueInCents;
```

packages/repo-tools/src/generate-sponsors.mts
```diff
@@ -118,25 +118,19 @@ async function main(): Promise<void> {
...
-      ...accumulator[name],
-      ...account.fromAccount,
-    };
-    return accumulator;
-  }, {});
-
-  const totalDonationsById = collective.members.nodes.reduce<
-    Record<string, number>
-  >((accumulator, member) => {
+  const accountsById = Object.fromEntries(
```
kirkwaiblinger, line 121: I'm not sure this has the same behavior.... The previous implementation copies and modifies values corresponding to existing keys in the `accumulator`, rather than overwriting them completely, which is what `Object.entries` does. Maybe that matters, maybe it doesn't, but I figure we should probably preserve the behavior as faithfully as possible with this PR...
 Maybe this one is better as a straightforward `for` loop?
 
 ```ts
 ```

kirkwaiblinger: (note that converting to a for loop also forces you to clarify some confusing variable name shadowing)

packages/utils/src/eslint-utils/deepMerge.ts
```diff
@@ -22,28 +22,27 @@ export function deepMerge(
...
+            ? // object type
+              deepMerge(firstValue, secondValue)
+            : // value type
+              secondValue
+          : firstHasKey
+            ? firstValue
+            : secondValue,
+      ];
+    }),
+  );
```
kirkwaiblinger, line 45: Consider keeping the diff simpler like so:
```suggestion
      let value: unknown;

      if (firstHasKey && secondHasKey) {
        if (isObjectNotArray(firstValue) && isObjectNotArray(secondValue)) {
          // object type
          value = deepMerge(firstValue, secondValue);
        } else {
          // value type
          value = secondValue;
        }
      } else if (firstHasKey) {
        value = firstValue;
      } else {
        value = secondValue;
      }

      return [key, value];
```

packages/website/src/components/lib/jsonSchema.ts
```diff
@@ -176,30 +176,33 @@ export function getTypescriptOptions(): DescribedOptionDeclaration[] {
...
-        enum: Array.from(item.type.keys()),
-      };
-    }
-    return options;
-  }, {});
+          description: item.description.message,
+          enum: Array.from(item.type.keys()),
+        };
+      }
+      return value ? [[item.name, value]] : [];
```
kirkwaiblinger, line 203: (optional) Similar to a previous comment, I'd much prefer a conditional push for these rather than flatmap with a 1- or 0-length array.

kirkwaiblinger:
I like all of these refactors! I'm still not really sold on a total ban on `.reduce()` in principle 🤷, but maybe it's worthwhile as a practical matter. So, +1 to the code changes, 🤷 on the config change

Josh-Cena:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#examples summarizes all valid use cases of `reduce` I could conceive of at the time of writing. I think a blanket-ban is fine.

fisker:
Maybe too late. [prefer-object-from-entries](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-object-from-entries.md) should enabled first, it's a special version of the `no-array-reduce` fixes cases creating an object with `array.reduce`.

packages/utils/src/eslint-utils/deepMerge.ts
```diff
@@ -22,28 +22,27 @@ export function deepMerge(
   // get the unique set of keys across both objects
   const keys = new Set(Object.keys(first).concat(Object.keys(second)));
 
-  return Array.from(keys).reduce<ObjectLike>((acc, key) => {
-    const firstHasKey = key in first;
-    const secondHasKey = key in second;
-    const firstValue = first[key];
-    const secondValue = second[key];
+  return Object.fromEntries(
+    Array.from(keys).map(key => {
```
fisker, line 26:
```suggestion
    Array.from(keys, key => {
```

packages/repo-tools/src/generate-configs.mts
```diff
@@ -72,9 +72,8 @@ async function main(): Promise<void> {
   }
 
   const RULE_NAME_PREFIX = '@typescript-eslint/';
-  const MAX_RULE_NAME_LENGTH = Object.keys(eslintPlugin.rules).reduce(
-    (acc, name) => Math.max(acc, name.length),
-    0,
+  const MAX_RULE_NAME_LENGTH = Math.max(
+    ...Object.keys(eslintPlugin.rules).map(name => name.length),
```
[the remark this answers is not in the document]
abrahamguo: Won't affect anything since this constant is only given to `String.padEnd`, and, from MDN,
> If the value is less than or equal to str.length, the current string will be returned as-is.

I wouldn't want to add `0,`, because it could seem like the `0,` is required, to a future reader, when it actually won't affect anything.

packages/website/src/components/lib/jsonSchema.ts
abrahamguo: Updated to use `.map.filter` because I like functional methods 😁

packages/parser/src/parser.ts
```diff
@@ -42,13 +42,9 @@ function validateBoolean(
...
   if (compilerOptions.lib) {
-    return compilerOptions.lib.reduce<Lib[]>((acc, lib) => {
+    return compilerOptions.lib.flatMap(lib => {
       const match = LIB_FILENAME_REGEX.exec(lib.toLowerCase());
-      if (match) {
-        acc.push(match[1] as Lib);
-      }
-
-      return acc;
+      return match ? (match[1] as Lib) : [];
```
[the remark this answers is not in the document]
abrahamguo: Updated to use `.map.filter` because I like functional methods 😁 but agree that this use of `flatMap` was weird

packages/repo-tools/src/generate-sponsors.mts
abrahamguo: Good catch on the variable name that was just plain wrong — should be `order`, not `member`, `account` or `accountNode`. (The TypeScript type is `OrderNode`)

There is guaranteed to be no functionality change because:
=> no properties of the `OrderNode` type are optional (i.e. all properties are always there)
=> (I confirmed that was correct, on the actual response data)
=> so `accountNode.fromAccount` always overwrites `accountsById[name]`
=> `Object.fromEntries` picks the last value if two entries have the same key

kirkwaiblinger:
Closing this based on https://github.com/typescript-eslint/typescript-eslint/issues/9623#issuecomment-2254267973.

I personally would still welcome most of the changes here as miscellaneous refactors, though, _especially_ fixing the variable name shadowing oddity!

abrahamguo:
@kirkwaiblinger  would you be open to using this PR if I removed the lint rule but kept all the refactors?

Josh-Cena:
+1 to do the necessary refactors.

kirkwaiblinger:
@abrahamguo hehe.

> would you be open to using this PR...

yes 🚀 

> ...but kept all the refactors?

In good faith, I think we should take out any refactors that are _purely_ of the variety that @bradzacher is -1e9999 on (i.e. the case mentioned in https://github.com/typescript-eslint/typescript-eslint/issues/9623#issuecomment-2254267973) 🙂

kirkwaiblinger (changes requested):
Changes discussed in top-level comments

abrahamguo:
Any refactors that look like
```
const dataStructure;
for (const ... of ...) {
  ...append to dataStructure...
}
```
have been removed 👍

JoshuaKGoldberg (approved):
This form I'm happy with 🙂 thanks!

packages/eslint-plugin/src/rules/member-ordering.ts
```diff
@@ -299,75 +299,73 @@ export const defaultOrder: MemberType[] = [
...
-      'set',
-      'static-initialization',
-    ] as const
-  ).reduce<Set<MemberType>>((all, type) => {
-    all.add(type);
-
-    (['public', 'protected', 'private', '#private'] as const).forEach(
-      accessibility => {
-        if (
+const allMemberTypes = [
```
Josh-Cena, line 302: Wow this is a giant one. Definitely an improvement.

tools/scripts/generate-configs.mts
```diff
@@ -72,9 +72,8 @@ async function main(): Promise<void> {
   }
 
   const RULE_NAME_PREFIX = '@typescript-eslint/';
-  const MAX_RULE_NAME_LENGTH = Object.keys(eslintPlugin.rules).reduce(
-    (acc, name) => Math.max(acc, name.length),
-    0,
+  const MAX_RULE_NAME_LENGTH = Math.max(
+    ...Object.keys(eslintPlugin.rules).map(name => name.length),
   );
```
Josh-Cena, line 77: I believe the original intention is that argument count must be <65536. Not sure it can be hit here 😅

Josh-Cena (approved):
Looks great to me :)

packages/eslint-plugin/tests/rules/no-unsafe-assignment.test.ts
```diff
@@ -12,55 +12,51 @@ type Options = InferOptionsTypeFromRule<typeof rule>;
...
+        {
+          messageId: 'unsafeArrayPatternFromTuple',
+          line: 1,
+          column: column + 13,
+          endColumn: endColumn + 13,
+        },
+      ],
+    },
+    // AssignmentExpression
+    ...(skipAssignmentExpression
```
kirkwaiblinger, line 44: still think these types of cases should be explicit like below rather than using lots of `...` operator to conditionally push within an expression.
```ts
const mappedTests = [ 
  // ... lots of cases
]
if (skipAssignmentExpression !== true) {
    mappedTests.push(/* etc */);
}
```
(also note that `skipAssignmentExpression !== true` isn't equivalent to `if (!skipAssignmentExpression)` here - let's preserve that).

kirkwaiblinger (approved):
A few things that I'd personally change, but nothing that's a blocker if we want to merge. 

Thanks for trudging through all these super confusing `reduce` constructions!

kirkwaiblinger (approved):

PR #9656, opened 2024-07-29
fix(eslint-plugin): [no-misused-promises] perf: avoid getting types of variables/functions if the annotated type is obviously not a function

bradzacher:
## Overview

I sat down for a few minutes to try perf profiling our codebase (writeup coming slowly in #6366) to look for some obvious improvements.

One thing that immediately stood out was that almost 10% of our lint run was spent in `no-misused-promises`' `checkVariableDeclaration` function. We can fix this easily by doing a quick syntactic check to filter out cases of (a) not-annotated variables and (b) annotated types that can never be function types.

This completely removes the function from the left-heavy cpu profile flame graph.

It ultimately doesn't change the lint run performance, however, because types are lazily calculated - so it has only shifted the workload to another rule -- but if we can do this enough we can shift that needle.


eslint.config.mjs
JoshuaKGoldberg, line 1: (merge shenanigan)

JoshuaKGoldberg (approved):
Awesome! ⚡

eslint.config.mjs
bradzacher: nah i added this cos I was running from the root with `node --cpu-prof --max-old-space-size=10000 ./node_modules/.bin/eslint .` and our lint run was matching reporting on files in these two folders.

the `/` should make eslint ignore the entire folder without globbing etc

PR #9658, opened 2024-07-29
chore: enable prefer-regexp-test

abrahamguo:
## PR Checklist

## Overview

Enable `unicorn/prefer-regexp-test` ([docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-regexp-test.md)).


Josh-Cena (approved):

JoshuaKGoldberg (approved):
We already have https://typescript-eslint.io/rules/prefer-regexp-exec. I wonder if we should also have `prefer-regexp-test`? 🤷 

...or maybe move both to `eslint-plugin-regexp`?

abrahamguo:
@JoshuaKGoldberg Not saying they wouldn't be open to it, but it looks like all of `eslint-plugin-regexp`'s lint rules relate to the contents of regular expressions themselves, whereas rules like `prefer-regexp-test` and `prefer-regexp-exec` relate to the JS methods surrounding the regular expressions.

Josh-Cena:
That's two +1s—let's do this!

It's indeed weird that we have `prefer-regexp-exec`. Looks like the Unicorn rule doesn't use type information though, so it may false positive though chances are sparse.

Josh-Cena (approved):

PR #9684, opened 2024-07-31
docs: add guide for building ESLint plugins

### LINKED ISSUE #5444: Docs: Add guidance for third party ESLint plugins and type checking

### Suggested Changes

Forking from https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/812#issuecomment-1207694721: ESLint plugins want to be able to use type checking in their rules. For example, `eslint-plugin-jsx-a11y`'s `jsx-a11y/alt-text` wants to be able to determine whether a spread provides a prop such as `alt` to an image that would make it obviously inaccessible.

```tsx
// Is this statically known to be inaccessible?
// We need type information to know!
<img src="something.jpg" {...myProps} />
```

Copying https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/812#issuecomment-1207694217:

> A couple of implementation detail points that make being type-aware tricky:
>
> * If the rule uses the typescript-eslint tooling to be type-aware, it'll be locked into TypeScript (i.e. Flow and any future competitors won't work for its type checking)
> * In my (and other typescript-eslint maintainers') experience, it's generally confusing for users when rules provide different -even better!- results when the config changes to allow type checking
>
> Those concerns might be small and not more important than getting better info. Either way, this scenario is something we probably need to make explicit docs for on typescript-eslint.io.
>
> Out of curiosity @ljharb, have you seen other plugins add type-aware linting the way this issue is suggesting before?

Copying https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/812#issuecomment-1207694721:

> eslint-plugin-react and eslint-plugin-import have lots of examples to look at.
>
> The one trick tho will be that we'll probably need to add the infrastructure for testing with the TS parser.

---

We'll want to document, at the very least:

* Guidance for how to turn on typed linting for rules (explicit opt-in? implicit-if-parser-services-are-available?)
* How to set up tests, if not already covered in https://typescript-eslint.io/docs/development/custom-rules
* ...anything else useful?

I would mention other maintainers on those repos to ask for input too, but... @ljharb are you the only one? Are there other people we should loop in? 😂 

### Affected URL(s)

https://typescript-eslint.io/docs/development/custom-rules should probably link to this new page

Maybe https://typescript-eslint.io/docs/development/custom-plugins or some similar URL?

### LINKED ISSUE #7066: Docs: instructions on how to integrate custom rules into your project

### Suggested Changes

typescript-eslint has some [great documentation](https://typescript-eslint.io/custom-rules) and [utils](https://typescript-eslint.io/architecture/utils/) to write a custom linting rule 👍 The first line reads "_custom ESLint rules generally work the same way for JavaScript and TypeScript code_", so the [plugin docs from eslint](https://eslint.org/docs/latest/extend/plugins) are a good starting point. However, it still left me with some questions around how to utilize such a rule in a project. 

Writing custom rules on top of typescript-eslint might not be incredibly common, so I don't expect there to be a extremely detailed step-by-step guide. But I think some pointers around the questions listed below would be a great addition to the documentation:

1. Can the `rule` from the [code snippet in the documentation](https://typescript-eslint.io/custom-rules/#rulecreator) be directly consumed in your eslint config, or does it need to be wrapped in a plugin?
   -  my guess is this is generally the same as eslint, but is anything like "[Runtime Rules](https://eslint.org/docs/latest/extend/custom-rules#runtime-rules)" also possible?
2. Do rules need to be compiled to JS before being usable? 
   - my assumption is yes, but this was optional in `tslint` if you utilized `ts-node`. So it would be nice to mention this explicitly.
3. What tsconfig is recommended when compiling a rule?
   - I tried copying configuration from the `@typescript-eslint/eslint-plugin` package (given that one contains the rules). Not all of it seems to be relevant for creating rules though
4. If you want to publish a package with a rule: 
   - are there package naming recommendations on top of the [naming instructions](https://eslint.org/docs/latest/extend/plugins#name-a-plugin) from eslint itself?
   - should `typescript` and `@typescript-eslint/***` packages be part of your (peer/dev-)dependencies?

This list is not intended as a literal proposal of questions that should all be addressed in the docs. Rather they're meant as examples of a topic that I think could use some clarification. Even referencing an example repo with a good setup would already help. Alternatively, there's interest in having a generator similar to [eslint/generator-eslint](https://github.com/eslint/generator-eslint). There's [an open issue](https://github.com/eslint/generator-eslint/issues/120) to create a TypeScript template, but at this moment it seems unclear if they want to move forward with that.

### Affected URL(s)

https://typescript-eslint.io/custom-rules

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5444; fixes #7066

## Overview

Also updates `RuleTester` docs to use `parserOptions.projectServer` instead of `parserOptions.project`. Both still work, but the new form doesn't require a `fixtures/` directory on disk (🥳!). 

Backing PR to add the example project: https://github.com/typescript-eslint/examples/pull/13

💖


docs/developers/Custom_Rules.mdx
```diff
@@ -305,7 +305,6 @@ This rule bans for-of looping over an enum by using the TypeScript type checker
 
 ```ts
 import { ESLintUtils } from '@typescript-eslint/utils';
-import * as tsutils from 'ts-api-utils';
```
JoshuaKGoldberg, line 308: There's really no need to ask users to load in a full third-party-package for a single enum helper function. Especially not for a package outside of the org's ownership.

docs/developers/ESLint_Plugins.mdx
```diff
@@ -0,0 +1,91 @@
...
+This page describes how to write your own custom ESLint plugins using typescript-eslint.
+You should be familiar with [ESLint's plugins guide](https://eslint.org/docs/latest/extend/plugins) and [typescript-eslint Custom Rules](./Custom_Rules.mdx) before writing custom plugins.
+:::
+
+Custom plugins that support TypeScript code and typed linting look very similar to any other ESLint plugin.
+Follow the same general steps as [ESLint's plugins guide > _Creating a plugin_](https://eslint.org/docs/latest/extend/plugins#creating-a-plugin) to set up your plugin.
+The required differences are noted on this page.
+
+:::tip
+See [**`eslint-plugin-example-typed-linting`**](https://github.com/typescript-eslint/examples/tree/39a6f7bb310645c88382a25aa8bed719ccd61a0c/packages/eslint-plugin-example-typed-linting) for an example plugin that supports typed linting.
```
Josh-Cena, line 17: Use `main` instead of a hash; it's better for the link to be broken (and fixed) rather than outdated by default.

docs/developers/Custom_Rules.mdx
```diff
@@ -331,6 +330,7 @@ export const rule = createRule({
   meta: {
     docs: {
       description: 'Avoid looping over enums.',
+      requiresTypeChecking: true,
```
Josh-Cena, line 333: Maybe add a comment saying what it's useful for

docs/developers/ESLint_Plugins.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+Most ESLint plugins export a _"`recommended`"_ [ESLint shared config](https://eslint.org/docs/latest/extend/shareable-configs).
+Many ESLint users assume enabling a plugin's `recommended` config is enough to enable all its relevant rules.
+
+However, at the same time, not all users want to or are able to enabled typed linting.
+If your plugin's rules heavily use type information, it might be difficult to enable those in a `recommended` config.
+
+You have roughly two options:
+
+- Have your plugin's `recommended` config require enabling type information _(recommended for new plugins)_
```
Josh-Cena, line 80: I don't think this has anything to do with age. It's up to the plugin to decide whether the plugin needs type information to be useful at all. For example if I make a Unicorn-style plugin today I would certainly still have a separate `recommendedTypeChecked` config.

Josh-Cena (approved):

docs/developers/ESLint_Plugins.mdx
JoshuaKGoldberg: I'll go ahead and merge https://github.com/typescript-eslint/examples/pull/13 to get it on that repo's `main`. 👍

docs/developers/Custom_Rules.mdx
JoshuaKGoldberg: I can't think of anything to say that isn't implied by the name. 😕

Josh-Cena: I meant for example: the rule tester and the doc generator use this info to do xyz, etc.

JoshuaKGoldberg: Oh right now that we have the custom `meta.docs` types, this isn't even needed anymore! I'll just remove it.

PR #9698, opened 2024-08-01
chore: enabled eslint-plugin-perfectionist on utils

### LINKED ISSUE #8479: Repo: Enable alphabetical (natural) sorting of fields
### Suggestion

Spinning out of https://github.com/typescript-eslint/typescript-eslint/issues/8477#issuecomment-1946131383: we're interested in maybe enabling https://eslint-plugin-perfectionist.azat.io/configs/recommended-natural or similar to enforce alphabetical sorting of fields. That likely includes methods and properties on classes, interfaces, objects, type aliases, and so on.

For reference I've used that rule in my http://github.com/JoshuaKGoldberg/create-typescript-app for quite a while now and have been very happy with it.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: starts on #8479 ... in that it _starts_ on it; I figure we'd want to split out followup issues for other packages

## Overview

Enables [`eslint-plugin-perfectionist`](https://perfectionist.dev) on just one package to start: `packages/utils`.


packages/utils/src/ast-utils/helpers.ts
```diff
@@ -8,14 +8,14 @@ type ObjectEntries<BaseType> = ObjectEntry<BaseType>[];
...
-    node: TSESTree.Node | null | undefined,
+    node: null | TSESTree.Node | undefined,
   ): node is Extract<TSESTree.Node, { type: NodeType }> =>
     node?.type === nodeType;
 
 export const isNodeOfTypes =
   <NodeTypes extends readonly AST_NODE_TYPES[]>(nodeTypes: NodeTypes) =>
   (
-    node: TSESTree.Node | null | undefined,
+    node: null | TSESTree.Node | undefined,
```
Josh-Cena, line 18: This looks worse to me, FWIW.

abrahamguo: I agree, which is why I still prefer (and use) `@typescript-eslint/sort-type-constituents` in my own projects even though it's deprecated. I'm hoping there's some way to achieve the same sorting with `eslint-plugin-perfectionist`, or that `@typescript-eslint/sort-type-constituents` gets brought back somehow... 😬

JoshuaKGoldberg: Heh yeah. I also like having `null` and `undefined` at the end. IMO they're separate primitives that should be after the normal flow of defined types.

This is somewhat-coincidentally controllable by the [`ignoreCase` rule option](https://perfectionist.dev/rules/sort-union-types#ignorecase)... I'll apply that now.

It'd be nice if we could set the general settings for the plugin once - rather than having a setting for each rule... filed: https://github.com/azat-io/eslint-plugin-perfectionist/issues/202

eslint.config.mjs
```diff
@@ -568,4 +569,36 @@ export default tseslint.config(
       'import/no-default-export': 'off',
     },
   },
+  {
+    ...perfectionistPlugin.configs['recommended-alphabetical'],
```
bradzacher, line 573:
```suggestion
    extends: [perfectionistPlugin.configs['recommended-alphabetical']],
```

eslint.config.mjs
```diff
@@ -568,4 +569,36 @@ export default tseslint.config(
...
+    files: ['packages/utils/src/**/*.ts'],
+    rules: {
+      ...perfectionistPlugin.configs['recommended-alphabetical'].rules,
```
bradzacher, line 576:
```suggestion
```

bradzacher (approved):

eslint.config.mjs
JamesHenry: I didn't know this was even supported any more, it is not listed in the configuration docs on the eslint site

bradzacher: @JamesHenry It's not part of ESLint - it's part of our config util 😄 
https://typescript-eslint.io/packages/typescript-eslint#flat-config-extends

I added it to our util because doing this sort of thing is so common that manually doing what Josh had to do (manually pull apart an object... or is it an array... who knows!)

packages/utils/src/ast-utils/helpers.ts
abrahamguo: so `ignoreCase` fixes this particular case just by coincidence.

But it doesn't fix the issue generally, i.e.:
image

do you think that adding support for `null` and `undefined` at the end is something that `eslint-plugin-perfectionist` would add?

JoshuaKGoldberg: Maybe! It'd be a good issue to file on them for sure.

bradzacher (dismissed):

kirkwaiblinger:
I'm -0.5 on proceeding with this due to the `null | string` behavior that exists for now (https://github.com/typescript-eslint/typescript-eslint/pull/9698#discussion_r1700883926). But if that's deemed an accepted drawback, then I approve this PR.

bradzacher:
Have we got an issue filed upstream?
I feel like a "put nullish types last" option probably would be a good default.

JoshuaKGoldberg:
Filed https://github.com/azat-io/eslint-plugin-perfectionist/issues/223. There's a `groups` option but I couldn't tell from the docs what the sanctioned way to do this was.

eslint.config.mjs
```diff
@@ -575,4 +576,36 @@ export default tseslint.config(
...
+          partitionByComment: true,
+          type: 'natural',
+        },
+      ],
+      'perfectionist/sort-union-types': [
+        'error',
+        {
+          order: 'asc',
+          groups: ['unknown', 'nullish'],
+          ignoreCase: false,
```
kirkwaiblinger, line 604: Shall we revert this now that the `| null | undefined` issue is resolved?

kirkwaiblinger (approved):
🚀😁

packages/utils/src/ast-utils/helpers.ts
JoshuaKGoldberg: https://github.com/typescript-eslint/typescript-eslint/pull/9698#issuecomment-2296427550 🚀

eslint.config.mjs
JoshuaKGoldberg: Yeah, I'd been using it to put lowercase keywords like `string` after capitalized things... `groups: ['unknown', 'keyword', 'nullish']` does that too and a little better.

kirkwaiblinger (approved):

packages/utils/src/ts-eslint/Config.ts
```diff
@@ -18,13 +18,13 @@ export namespace SharedConfig {
...
-    | 'writable'
-    | /** @deprecated use `'writable'` */ 'writeable'
-    | /** @deprecated use `'readonly'` */ 'readable';
+    | /** @deprecated use `'writable'` */ 'writable'
+    | /** @deprecated use `'readonly'` */ 'writeable';
   export type GlobalVariableOptionBoolean =
-    | /** @deprecated use `'writable'` */ true
-    | /** @deprecated use `'readonly'` */ false;
+    | /** @deprecated use `'writable'` */ false
+    | /** @deprecated use `'readonly'` */ true;
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Oh, nice! Thanks for the catch :)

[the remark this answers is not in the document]
JoshuaKGoldberg: #9856

Issue #8054, opened 2023-12-12
Bug: no-misused-promises ignores conversion of Promise to void for generic higher-order-functions with fixed type parameter


mx-bernhard:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.3.2&fileType=.ts&code=C4TwDgpgBAqgdgazgewO5wM4DECucDGUAvFABQB0lAhgE4DmGAXFHkmnANoC6AlMQHwtEKdAG4AUONCQoAJQjAcNTADVkASwAmxMnyKCAbhs0TxmiPgA2taPmSZgLDBADCVS5YBGVfAkbioKAAeABVcAmB1eygIAA9gCDhNDFhhdmw8fEFSADM4ZjDMyPs9QUKIqLhTcysbKDsHJ1d3Lx8EeUVldTg6NS1maQhkHKa3D29fII6lVWN%2BUxxnMdbfaa6evs1SUioMEAJdASgAbygAXx4eUSgAehv65BoaC2BJRebxtqmFGYxN-h2ewOpFKJ3OV1u9wwAAtkDhLNpPNBnmBHgltLsoKgIB5xEA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oFtLlZkiACa1i0Dr0GoMAbXDYciaOOiQANPIVYF2SGQAWiMgGtkANQ6UhAJUT5Y0JujD44iTdgC%2BHgLrzvnkA&tsconfig=&tokens=false

### Repro Code

```TypeScript
type UnknownsFunc = (...args: unknown[]) => unknown;

type ReturnsVoid = () => void;

declare const useCallback:
  <TFunction extends UnknownsFunc> (fn: TFunction) => TFunction;

declare const useCallbackReturningVoid: typeof useCallback<ReturnsVoid>;

useCallbackReturningVoid((async () => { })); // correct

[...]

Issue #8676, opened 2024-03-14
Add a “recommended-next” preset for rules that will be recommended in next major


voxpelli:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22preset+config+change%22+) and found none that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Description

Maybe add an “recommended-next” / “recommended-unstable” / “recommended-beta” where new rules like https://github.com/typescript-eslint/typescript-eslint/pull/8067 can be added without having to wait for the next major – then when the new major hits those are simply moved from there to the stable recommendation

That way the future of new rules becomes more documented.

A staging area for preset changes pretty much.

### Additional Info

I originally posted in the Discord and @JoshuaKGoldberg asked me to post it here

bradzacher:
The big issue I see here is that we don't usually decide upon what the recommended set should be in the next major until we're pepping for the next major.

We might have some ideas about new rules and whether or not they make a good recommended candidate - but we don't usually provide much more thought than that.

voxpelli:
Here's the Discord thread: https://discord.com/channels/1026804805894672454/1217588154710622271/1217896050745147423

There was some confusion as to why `no-array-delete` wasn't recommended and it wasn't easy to see that it was a new rule that likely would be considered for inclusion in the next major, leading to questions being asked, so that's the problem I suggested this to solve 😌

bradzacher:
Yeah for sure - I get the idea!
I'm just explaining that it would be a big change in our workflow for us to build this.

Right now our workflow is that we don't fully consider changes to the recommended set until we are preparing for the next major - at which point we generate an issue with the list of rules and we go through each rule to decide (a) if it's in the list - should it stay in the list and (b) if it's not in the list - should it be added.

Essentially we write everything out and we open it for community discussion and then we make some decisions and commit them for the major.

So you can see how trying to maintain a list before the major is a shift in things - we need to build out this issue and remember to keep it updated as we merge new rules in so that we don't lose track of anything.

JoshuaKGoldberg:
There's also the issue of #8695, that we leak internal `meta.docs` types. I'd think we'd want to wait until that's resolved before adding _yet another_ option to these types (previous addition: #8364).

[...]

PR #9707, opened 2024-08-02
feat(eslint-plugin): add rule [strict-void-return]

### LINKED ISSUE #1744: Rule proposal check void-covariant return types of implemented interfaces and extended classes
**What code were you trying to parse?**

```ts
interface X {
  a(): void;
}

class Y implements X {
  async a(): Promise<void> {
  }
}
```

**What did you expect to happen?**

This should (configurably) be an error, especially concerning floating promises.

**What actually happened?**

No error.

**Versions**

```
    "@typescript-eslint/eslint-plugin": "^2.24.0",
    "@typescript-eslint/eslint-plugin-tslint": "^2.24.0",
    "@typescript-eslint/parser": "^2.24.0",
    "eslint": "^6.8.0",
    "eslint-config-prettier": "^6.10.0",
    "typescript": "^3.8.3"
```

**Other**

This might be related: https://github.com/microsoft/TypeScript/issues/4544

### LINKED ISSUE #2988: [no-confusing-void-expression] Add an option to warn on non void arrow callbacks
**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/no-confusing-void-expression": [
      "warn",
      { ignoreArrowShorthand: true },
    ],
  }
}
```

```TS
// setState return void, this is ok and shorter
<Modal onClose={() => setState(undefined)} /> 
// We are returning a number to a void callback. This is probably a mistake
<Counter onChange={(value) => Math.abs(value)}   /> 
```

**Expected Result**

Warn in the second case. 

I would like the rule to report on return type mismatch like `no-misused-promises`, while keeping the the possibility to use arrow function when callback returns nothing. Would an option for this be accepted? I can work on it next week if I have a green light

We could also change the default behaviour of ignoreArrowShorthand, but it would be a breaking change

**Actual Result**

Warn in both cases by default, non with `{ ignoreArrowShorthand: true }`

**Versions**

| `@typescript-eslint/eslint-plugin` | `4.14.1` |
| `@typescript-eslint/parser`        | `4.14.1` |
| `TypeScript`                       | `4.0.5` |
| `ESLint`                           | `7.19.0` |
| `node`                             | `4.13.1` |

phaux:
## PR Checklist

  - fixes #2988 
  - fixes #1744

## Overview

So basically I implemented the void checking as requested by #2988. For every checked node (arguments, assignments, returns, etc) I take the actual function type and the contextual function type and compare return types. Only object method shorthand required slightly different logic. 

That already worked pretty well, but I also found #1744 and decided to include it in this rule as well, since I already had a similar thing implemented for object shorthand methods. I added this as an option `considerBaseClass` and `considerImplementedInterfaces`, enabled by default.

Then I noticed that callback for `addEventListener` is not detected as void context. That's because it has another signature where the callback can return `any`. I was stuck on this for a long time. Ultimately I looked at how `no-misused-promises` does this and implemented something similar as `considerOtherSignatures` option, enabled by default.

At this point this rule already did everything `no-misused-promises`'s `checkVoidReturn` did, but better. It doesn't have problems like #8054 or #8739. Maybe it's worth splitting `no-misused-promises` into 3 separate rules in the future?

I also added many autofixes and suggestions. They are possible when the provided function is a function literal and we can inspect its body. Some of them are the same as in `no-confusing-void-expression` so I moved them into utils. It might make sense to change some autofixes into suggestions instead so they don't accidentally remove a big chunk of code. Let me know if that's a good idea.

The biggest feature is automatic suggestions which I and probably others had to type manually a thousand of times:

```js
takesCallback(async () => {
  block;
});
``` 

into 

```js
takesCallback(() => {
  (async () => {
    block;
  })().catch(err => {});
});
```

or

```js
// eslint-disable-next-line @typescript-eslint/no-misused-promises
takesCallback(async () => {
  try {
    block;
  } catch {}
});
```

To allow the second suggestion without the need of ignoring the line I added the option `allowReturnPromiseIfTryCatch`. It's just a simple extra `if` near the end of the long routine that checks everything that could be wrong in the function body. I hope it can stay.


JoshuaKGoldberg:
👋 Exciting PR, really looking forward to the rule! Just marking as a draft because there are unit test failures. This keeps getting me excited when it pops up in my notifications 😄. Let us know if you want to talk or ask questions about any of them.

phaux:
Fixed and I'm not planning any more changes so I'm undrafting it I guess.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+            .unionTypeParts(funcType)
+            .flatMap(type =>
+              ts.isCallExpression(callTsNode)
+                ? type.getCallSignatures()
+                : type.getConstructSignatures(),
+            );
+          const argExpectedReturnTypes = funcSignatures
+            .map(s => s.parameters[argIdx])
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Indexing can return undefined
+            .filter(param => param != null)
```
JoshuaKGoldberg, line 363: [Style] Minor nitpick, but helps get rid of the comment 😄 

```suggestion
            .filter(Boolean)
```

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
+import assert from 'node:assert';
```
JoshuaKGoldberg, line 1: [Refactor] Rules can run in environments other than Node. But, even if we did want to take a dependency on `node:assert`, the kinds of assertions in this rule aren't what we'd do. Lint rules run in a controlled enough environment (known AST inputs) that we should be able to handle all cases without crashing. 

If something is known to be a specific type and there's no good way to tell the type system that, use an `as`.

Aside: it's common that the type system is actually correct when it's yelling about a node being the wrong type. Or at least - it's correct per the AST's types. If it's yelling and you think it shouldn't then that might be a bug in the AST types? For example, in the `if (funcNode.generator)` check, `funcNode.body` is typed as `TSESTree.BlockStatement | TSESTree.Expression`... even though if `funcNode.generator` is true, we know it's not an arrow expression. A bug in the AST types maybe?

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+          if (varName != null) {
+            checkExpressionNode(node.right, 'Var', { varName });
+          } else {
+            checkExpressionNode(node.right, 'Other');
+          }
+        }
+      },
+      ObjectExpression: (node): void => {
+        for (const propNode of node.properties) {
+          if (propNode.type !== AST_NODE_TYPES.SpreadElement) {
```
JoshuaKGoldberg, line 236: [Testing] If I remove this `if` and just leave its body, all unit tests still pass. That means there's at least a missing test case.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+          if (
+            elemNode != null &&
+            elemNode.type !== AST_NODE_TYPES.SpreadElement
+          ) {
+            checkExpressionNode(elemNode, 'Other');
+          }
+        }
+      },
+      ArrowFunctionExpression: (node): void => {
+        if (node.body.type !== AST_NODE_TYPES.BlockStatement) {
```
JoshuaKGoldberg, line 252: [Testing] If I remove this `if` and just leave its body, all unit tests still pass. That means there's at least a missing test case.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+    const sourceCode = context.sourceCode;
+    const parserServices = util.getParserServices(context);
+    const checker = parserServices.program.getTypeChecker();
+
+    return {
+      'CallExpression, NewExpression': (
+        node: TSESTree.CallExpression | TSESTree.NewExpression,
+      ): void => {
+        checkFunctionCallNode(node);
+      },
```
JoshuaKGoldberg, line 200: [Style] Nit: these can be inlined. Anything that just curries a single param ✨ 

```suggestion
      'CallExpression, NewExpression': checkFunctionCallNode,
```

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+          if (propNode.type !== AST_NODE_TYPES.SpreadElement) {
+            checkObjectPropertyNode(propNode);
+          }
+        }
+      },
+      ArrayExpression: (node): void => {
+        for (const elemNode of node.elements) {
+          if (
+            elemNode != null &&
+            elemNode.type !== AST_NODE_TYPES.SpreadElement
```
JoshuaKGoldberg, line 245: [Testing] If I remove this `&& elemNode.type !== AST_NODE_TYPES.SpreadElement`, all unit tests still pass. That means there's at least a missing test case.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+          const varName = util.getNameFromExpression(sourceCode, node.id);
+          if (varName != null) {
+            checkExpressionNode(node.init, 'Var', { varName });
+          } else {
+            checkExpressionNode(node.init, 'Other');
+          }
+        }
+      },
+      AssignmentExpression: (node): void => {
+        if (['=', '||=', '&&=', '??='].includes(node.operator)) {
```
JoshuaKGoldberg, line 225: [Testing] If I remove this `if` and just leave its body, all unit tests still pass. That means there's at least a missing test case.

packages/eslint-plugin/src/util/getBaseTypesOfClassMember.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as ts from 'typescript';
+
+/**
+ * Given a member of a class which extends another class or implements an interface,
+ * yields the corresponding member type for each of the base class/interfaces.
+ */
+export function* getBaseTypesOfClassMember(
+  checker: ts.TypeChecker,
+  memberTsNode: ts.PropertyDeclaration | ts.MethodDeclaration,
+  heritageToken: ts.SyntaxKind.ExtendsKeyword | ts.SyntaxKind.ImplementsKeyword,
```
JoshuaKGoldberg, line 12: [Refactor] Both places that call `getBaseTypesOfClassMember` look roughly like:

```ts
for (const { baseMemberType } of util.getBaseTypesOfClassMember(checker, propTsNode, ts.SyntaxKind.ExtendsKeyword) {
    if (isVoidReturningFunctionType(baseMemberType)) {
      reportIfNonVoidFunction(propNode.value);
  }
}

for (const { baseMemberType } of util.getBaseTypesOfClassMember(checker, propTsNode, ts.SyntaxKind.ExtendsKeyword) {
    if (isVoidReturningFunctionType(baseMemberType)) {
      reportIfNonVoidFunction(propNode.value);
  }
}
```

There's some duplication there. I think the `heritageToken` parameter can be removed, to have the function always check both in one call, rather than one each in two calls.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+          ts.SyntaxKind.ExtendsKeyword,
+        )) {
+          const baseName = baseType.getSymbol()?.name ?? 'class';
+          if (isVoidReturningFunctionType(baseMemberType)) {
+            reportIfNonVoidFunction(propNode.value, 'ExtMember', {
+              memberName,
+              className,
+              baseName,
+            });
+          }
```
JoshuaKGoldberg, line 485: [Bug] It's kind of unusual to report twice in one place. Plus, for better performance, it's nice to not have to do checks later on once there's already a known report. For here and elsewhere, let's skip later reports as soon as one is made:

```suggestion
            });
            return;
          }
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
JoshuaKGoldberg, line 1: > At this point this rule already did everything `no-misused-promises`'s `checkVoidReturn` did, but better. It doesn't have problems #8054 or #8739. Maybe it's worth splitting `no-misused-promises` into 3 separate rules in the future? (this one being one of them)

This is a great question. `no-misused-promises` was already largely overlapped by `no-unnecessary-condition`. This new `strict-void-return` pretty much takes on the rest of `no-misused-promises`, making `no-misused-promises` redundant if you have both `no-unnecessary-condition` and `strict-void-return`... 

I'd be in favor of deprecated `no-misused-promises` in favor of using `no-unnecessary-condition` + `strict-void-return`. The only benefit I can think of for `no-misused-promises` would be projects that want to _only_ apply the checks for Promises... Maybe these two rules could each be given some kind of _"only check Promises"_ option?

Also of note is that `no-misused-promises`'s `checkVoidReturns` is pretty configurable. Maybe, if this rule is to replace `no-misused-promises`, it'd be useful to have each of those configurable options? Or, on the other hand, maybe those options are holdovers that real-world don't generally use? Investigation needed. I think those options can be a followup & shouldn't block this PR.

What do you think?

Also cc: @typescript-eslint/triage-team in general, and @kirkwaiblinger + @alythobani from #8765.

JoshuaKGoldberg (changes requested):
OK! Very happy to have finally made it to this PR: it's a great piece of work. As you said, it covers a heck of a lot of functionality, and does so in ways that are really solid compared to previous approaches. Fantastic! 👏 

It's also a _lot_ of code that was hard to read through. For an initial version of the rule, I think nuance around suggestions aren't necessary. And the fixers would need to be suggestions given they change code behavior.

I left requests for simplification through the code: for messages, options, and the suggestions.

But, my advice would be to hold off applying that large set of removals until the conversation in _(will switch to link)_ is resolved. The consensus might end up being that the options are good and useful after all.

Keanu Reaves as Neo in The Matrix saying 'whoa'

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: Why not add `assert` to utils like `nullThrows`?

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+      if (
+        funcNode.type !== AST_NODE_TYPES.ArrowFunctionExpression &&
+        funcNode.type !== AST_NODE_TYPES.FunctionExpression
+      ) {
+        // The provided function is not a function literal.
+        // Report a generic error.
+        return context.report({
+          node: funcNode,
+          messageId: `nonVoidFuncIn${msgId}`,
+          data,
```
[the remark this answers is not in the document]
phaux: I considered having context as an argument to messages like `"Unexpected value returned in {context}..."`, but then the context would be untranslatable, which I think is the whole point of using message IDs, no? 

Should I change it to taking context like `` `argument for ${funcName}` `` ?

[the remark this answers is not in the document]
JoshuaKGoldberg: That seems reasonable to me!

alythobani (commented):
Very cool PR, great job sticking it out! Left some thoughts/questions, hope they're helpful

packages/eslint-plugin/docs/rules/strict-void-return.mdx
kirkwaiblinger: Maybe a lame response, but is there a compelling reason not to land this first, then consider the no-misused-promises deprecation, and which options we might need to port or create in order to do so, afterwards?

Just thinking, deprecating no-misused-promises might have some strings attached, such as some nontrivial updating of the docs in no-floating-promises that explain how to lint against promise antipatterns outside of ExpressionStatements.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+        if (node.argument != null) {
+          checkExpressionNode(node.argument, 'Return');
+        }
+      },
+      PropertyDefinition: (node): void => {
+        checkClassPropertyNode(node);
+      },
+      MethodDefinition: (node): void => {
+        checkClassMethodNode(node);
+      },
```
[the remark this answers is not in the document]
JoshuaKGoldberg: I'd say let's wait on that, yeah. Maybe in a future PR.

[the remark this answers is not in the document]
alythobani: Including for `no-misused-promises`?

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,436 @@
...
+### Unsafety
+
+Passing a value-returning function in a place expecting a void function can be unsound.
+
+<Tabs>
+<TabItem value="❌ Incorrect">
+
+```ts
+const bad: () => void = () => 2137;
+const func = Math.random() > 0.1 ? bad : prompt;
```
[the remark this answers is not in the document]
phaux: I tried

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,991 @@
...
+    ],
+  },
+  defaultOptions: [
+    {
+      considerOtherOverloads: true,
+      considerBaseClass: true,
+      considerImplementedInterfaces: true,
+      allowReturnPromiseIfTryCatch: true,
+      allowReturnUndefined: true,
+      allowReturnNull: true,
```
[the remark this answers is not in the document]
phaux: Makes sense

packages/eslint-plugin/src/util/getBaseTypesOfClassMember.ts
phaux: Done

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: That's because technically this condition is not necessary, because if it's a spread argument then the contextual type wont be a void function. I don't know how to write a test for this. Is it even possible?

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,496 @@
...
+Passing a value-returning function in a place expecting a void function can be unsound.
+
+<Tabs>
+<TabItem value="❌ Incorrect">
+
+```ts
+const getNothing: () => void = () => 2137;
+const getString: () => string = () => 'Hello';
+const getStringMaybe = Math.random() > 0.1 ? getNothing : getString;
+const maybeString = getStringMaybe();
```
kirkwaiblinger, line 30: (praise) This version of the example is much easier to follow. Thanks! 🙂 

(nit) why not inline the `getStringMaybe` variable still?
```suggestion
const maybeString = Math.random() > 0.1 ? getNothing() : getString();
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,496 @@
...
+## Rule Details
+
+TypeScript considers functions returning a value to be assignable to a function returning void.
+Using this feature of TypeScript can lead to bugs or confusing code.
+
+## Examples
+
+### Return type unsafety
+
+Passing a value-returning function in a place expecting a void function can be unsound.
```
kirkwaiblinger, line 21: What do you think of a little more verbosity here? 

If I were writing it, I might do something like this...

````suggestion
Passing a value-returning function in a place expecting a void function can be unsound. TypeScript generally treats the `void` type as though it has the same runtime behavior as `undefined`, so this pattern will cause a mismatch between the runtime behavior and the types.

```ts
// TypeScript errors on overtly wrong ways of populating the `void` type...
function returnsVoid(): void {
    return 'not void!'; // TS Error:  Type 'string' is not assignable to type 'void'.
}

// ...but allows more subtle ones
const returnsVoid: () => void = () => 1234;

// Likewise, TypeScript errors on overtly wrong usages of `void` as a runtime value...
declare const v: void;
if (v) { // TS Error: An expression of type 'void' cannot be tested for truthiness.
    // ... do something
}

// ...but silently treats `void` as `undefined` in more subtle scenarios
declare const voidOrString: void | string;
if (voidOrString) {
     // voidOrString is narrowed to string in this branch, so this is allowed.
     console.log(voidOrString.toUpperCase());
}
```

Putting these behaviors together, we can construct a scenario where TS will throw at runtime, despite not reporting a type error:

````

File: packages/eslint-plugin/src/rules/strict-void-return.ts (hunk 11 above)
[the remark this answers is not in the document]
Comment (alythobani) [reply packages/eslint-plugin/src/rules/strict-void-return.ts]: Separate from this PR for sure, but I meant do we want to postpone that indefinitely until we have a better idea of what we're going to do with `no-misused-promises`?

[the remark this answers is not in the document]
Comment (JoshuaKGoldberg) [reply packages/eslint-plugin/src/rules/strict-void-return.ts]: Oh, gotcha - no I think that's a different thing. Whether a rule properly checks index signatures is a _bug_ scoped _with the rule_. High-level direction discussions like this might change what rule(s) exist or where they look, but don't really impact small bugs like that.

**Comment by JoshuaKGoldberg** [comment]
👋 @phaux is this ready for re-review?

**Comment by phaux** [comment]
Not yet, but I am thinking every day about how to simplify messages in this PR

**Comment by JoshuaKGoldberg** [comment]
I'll switch this to draft then 👍.

**Comment by JoshuaKGoldberg** [comment]
Checking in again @phaux - is this ready for review? Anything we can help with?

**Comment by phaux** [comment]
I addressed most of the issues. 

I left some unrelated refactors in another rule because they were needed for this rule's auto fixes, which are now removed and will be sent as separate PR. If that's important I can try to remove the unrelated changes too for now. I have to do it manually because they were part of the big initial commit and I was lazy.

The messages still use dynamic IDs, but I reduced their number by a lot. If that's still too many messages, let me know and I will just replace all of them with a single generic message like "value returned in void context" and move them to a separate PR so it can be discussed later too.

File: packages/eslint-plugin/src/util/discardReturnValueFix.ts [hunk 16]
```diff
@@ -0,0 +1,91 @@
...
+): TSESLint.RuleFix {
+  const argumentNode = nullThrows(
+    returnNode.argument,
+    'missing return argument',
+  );
+  let newReturnText = sourceCode.getText(argumentNode);
+  if (newReturnText[0] === '{') {
+    // The value would be interpreted as a block statement,
+    // so we need to wrap it in parentheses.
+    newReturnText = `(${newReturnText})`;
```
JoshuaKGoldberg, line 77: [Testing] This isn't covered by any tests - either it's a valid case that should be tested, or it's not and we can remove it.

packages/utils/src/eslint-utils/assert.ts
JoshuaKGoldberg, line 1: [Refactor] -1 from me in general on this extra set of assertions. There's a reason we haven't leaned into this pattern more: values should be their expected type before being used. 

E.g. instead of:

```ts
memberTsNode: ts.MethodDeclaration | ts.PropertyDeclaration,
// ...
ESLintUtils.assert(ts.isClassLike(memberTsNode.parent), '...');
```

...we'd want either:

* If `ts.MethodDeclaration | ts.PropertyDeclaration`'s `.parent` _is_ always a `ts.ClassLikeDeclaration`, the AST's types should reflect that
* If `ts.MethodDeclaration | ts.PropertyDeclaration`'s `.parent` _isn't_ always a `ts.ClassLikeDeclaration`, then `memberTsNode`'s type should have something like `{ parent: ts.ClassLikeDeclaration }` added

In other words: I think these assertions are surface-level fixes for type issues that should be fixed at a deeper level.

---

I'll also note that we generally try to work with the TSESTree AST rather than TypeScript's, when possible. Will mention this on `getBaseTypesOfClassMember` itself.

packages/eslint-plugin/src/util/getBaseTypesOfClassMember.ts
```diff
@@ -0,0 +1,42 @@
+import { ESLintUtils } from '@typescript-eslint/utils';
+import * as ts from 'typescript';
+
+/**
+ * Given a member of a class which extends another class or implements an interface,
+ * yields the corresponding member type for each of the base class/interfaces.
+ */
+export function* getBaseTypesOfClassMember(
+  checker: ts.TypeChecker,
+  memberTsNode: ts.MethodDeclaration | ts.PropertyDeclaration,
```
JoshuaKGoldberg, line 10: Continuing from the comment on assertions: I'll also note that we generally try to work with the TSESTree AST rather than TypeScript's, when possible. We have more flexibility and power to make the AST more specific because we're the ones who define it (and we define it as a discriminated union).

In this case it'd be more typical for a typed lint rule to take in the TSESTree node. You can also pass in the `services: ParserServicesWithTypeInformation` to use APIs like `services.getSymbolAtLocation`.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,636 @@
...
+    },
+    fixable: 'code',
+    hasSuggestions: false,
+    messages: {
+      asyncFuncInArg:
+        'Async callback passed as an argument to `{{funcName}}`, which expects a void callback.',
+      asyncFuncInAttr:
+        'Async event handler `{{attrName}}` passed as a prop to `{{elemName}}`, which expects a void event handler.',
+      asyncFuncInExtMember:
+        'Async function provided as `{{memberName}}` method of `{{className}}`, whose base class `{{baseName}}` declares it as a void method.',
```
JoshuaKGoldberg, line 49: [Typo] Grammar fix:

```suggestion
        'Async function provided as `{{memberName}}` method of `{{className}}`, whose base class `{{baseName}}` declares as a void method.',
```

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
JoshuaKGoldberg, line 1: [Testing] A few more edge cases to test

* `extends Bar implements Foo, Baz`
* `extends class { ... }` and  `extends class WithName { ... }`
* `implements class { ... }` and `implements class WithName { ... }`

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,636 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Disallow passing a value-returning function in a position accepting a void function',
+      requiresTypeChecking: true,
+    },
+    fixable: 'code',
+    hasSuggestions: false,
+    messages: {
```
JoshuaKGoldberg, line 43: From https://github.com/typescript-eslint/typescript-eslint/pull/9707#issuecomment-2575990827:

> The messages still use dynamic IDs, but I reduced their number by a lot. If that's still too many messages, let me know and I will just replace all of them with a single generic message like "value returned in void context" and move them to a separate PR so it can be discussed later too.

For me, dynamic IDs are a hard blocker. All IDs should be statically constructed for readability - both for us humans and the tools around them.

I'm not opposed to this PR having more specific reports for different cases. I think a few of the delineations are helpful:

* Declaring a class/object method/property _vs._ providing an argument to a function
* Explicitly stating which base class or interface made a method/property report

Looking at the code, it's ~7 archetypes of messages repeated across each of 3 groups. How about reducing to just 7 messages overall?

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
JoshuaKGoldberg, line 1: From https://github.com/typescript-eslint/typescript-eslint/pull/9707#issuecomment-2575990827:

> I left some unrelated refactors in another rule because they were needed for this rule's auto fixes, which are now removed and will be sent as separate PR. If that's important I can try to remove the unrelated changes too for now. I have to do it manually because they were part of the big initial commit and I was lazy.

Yes, please revert any changes that aren't necessary for the specific added rule in this PR.

FWIW this is part of why we don't require rebasing and actually request not to force-push in https://typescript-eslint.io/contributing/pull-requests 😄. We don't want to add "keeping the Git history pristine" to the tasks for you the PR author or us as maintainers.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,636 @@
...
+          data,
+        });
+      }
+
+      if (funcNode.async) {
+        // The provided function is an async function.
+
+        if (!options.allowReturnPromiseIfTryCatch) {
+          // Async functions aren't allowed.
+          // TODO: Suggest wrapping its body in an async IIFE.
```
JoshuaKGoldberg, line 529: [Docs] Could you remove these TODOs actually? We haven't decided what the right next step would be. I'm personally loathe to suggest async IIFEs; IMO they're 99% of the time an anti-pattern that results in dangerous async patterns.

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
```diff
@@ -0,0 +1,2767 @@
...
+        {
+          column: 42,
+          data: { varName: 'foo' },
+          line: 2,
+          messageId: 'asyncFuncInVar',
+        },
+      ],
+      options: [{ allowReturnUndefined: false }],
+    },
+    // TODO: Check every union type separately
```
JoshuaKGoldberg, line 1678: [Question] Is this still pending? What's up?

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,636 @@
...
+              "Whether to allow functions returning `undefined` to be used in place expecting a `void` function. When disabled, `void` operator can't be used to discard the return value, because it evaluates to undefined. Disable this to enforce a consistent style across the codebase.",
+          },
+        },
+      },
+    ],
+  },
+  defaultOptions: [
+    {
+      allowReturnNull: false,
+      allowReturnPromiseIfTryCatch: true,
```
JoshuaKGoldberg, line 121: [Feature] In this PR's OP, `allowReturnPromiseIfTryCatch` is only mentioned in the collapsed note around fixers & suggestions. Seems to me it's no longer necessary, right? 

ACK that it's not much added code but I don't think we have consensus that it's a strategy we'd want to add to the rule. Could you please split it out? We can always talk about it more in a followup issue - either its own and/or the one around fixers & suggestions.

packages/eslint-plugin/src/rules/strict-void-return.ts
JoshuaKGoldberg: This is still the case.

packages/eslint-plugin/src/rules/strict-void-return.ts
JoshuaKGoldberg: Oop, sorry I missed this question. Looking at the non-spread test case that hits this line:

```ts
declare function cb(): void;
const foo: Record<string, () => void> = {
  cb1: cb,
  cb2: cb,
};
```

We'd be able to trigger it by switching one of those properties to a spread:

```ts
declare function cb(): void;
const foo: Record<string, () => void> = {
  cb1: cb,
  ...{},
};
```

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,636 @@
...
+  name: 'strict-void-return',
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Disallow passing a value-returning function in a position accepting a void function',
+      requiresTypeChecking: true,
+    },
+    fixable: 'code',
+    hasSuggestions: false,
```
JoshuaKGoldberg, line 42:
```suggestion
```

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,636 @@
...
+          },
+        },
+      },
+    ],
+  },
+  defaultOptions: [
+    {
+      allowReturnNull: false,
+      allowReturnPromiseIfTryCatch: true,
+      allowReturnUndefined: true,
```
JoshuaKGoldberg, line 122: [Question] Separate from the promise/try-catch options, I don't know that we really need the remaining `allowReturn*` options. They're added complexity for the first version of a rule we haven't tried out in the real world yet. Do you see a strong need for them? (if not, I'm -1 on including them)

JoshuaKGoldberg (changes requested):
Progress! 🚀 

Requesting some more simplification changes that'll help us review please.

JoshuaKGoldberg:
👋 Checking in @phaux, do you think you'll have the time to drive this to completion over the next month or two? I really like the idea of the rule but it's been sitting for a while - it'd be nice to either really push it forward or open it up to get worked on.

phaux:
@JoshuaKGoldberg Sorry for the wait. I recreated this branch from scratch and the next version will be as minimal as possible so hopefully it finally passes review :) I will send it in few days probably.

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
phaux: This one was non-trivial to implement and I think it's not worth it.

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
phaux: `class implements class {} {}` is already invalid TS

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: removed

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: Added a test but it doesn't matter:

```ts
declare let foo: () => void;
foo += () => 1;
```

still doesn't fail with condition removed.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: This wont work because the message id and parts are decided at the beginning and they would still need to be passed to `reportIfNonVoidFunction` via argument and would be dynamic anyways.

So there can't be "where" distinction (variable / argument / class member), only the "what" distinction (return / async func / non-void func), because it's computed at the end. I agree the first is the more useful one.

I will remove all messages and leave a single generic one.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: I agree that AIIFEs suck, but that's the only way (that I know) to make this rule (or no-misused-promises) happy. That's why I added that suggestion and ignoreAsyncInTryCatch.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: Done

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: Removed

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: **Unless** we remove the "what" distinction and the whole `reportIfNonVoidFunction` so that every node check calls `context.report` directly with a static messageId.

But that distinction will be necessary for the autofixes anyways so it's not possible neither.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: fixed

packages/eslint-plugin/src/util/getBaseTypesOfClassMember.ts
phaux: Done

phaux:
Ready

packages/eslint-plugin/src/rules/strict-void-return.ts
JoshuaKGoldberg: Sounds to me like the `if` can be removed, then! 🔪 

Change request: remove the `if`?

Just noting for posterity, `node.operator` in an `AssignmentExpression` can be any of:

```ts
"=" | "+=" | "-=" | "*=" | "**=" | "/=" | "%=" | "<<=" | ">>=" | ">>>=" | "&=" | "|=" | "||=" | "&&=" | "??=" | "^="
```

The operators not mentioned in the `if` all produce TypeScript errors when given functions. So I agree with adding the `valid` test case containing `foo += () => 1`. That makes sure the removal of the `if` doesn't break those cases.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,452 @@
...
+        }
+      },
+      VariableDeclarator: (node): void => {
+        if (node.init != null) {
+          checkExpressionNode(node.init);
+        }
+      },
+    };
+
+    /** Checks whether the type is a void-returning function type. */
```
JoshuaKGoldberg: [Docs] Nit: I think the _"is void returning function type"_ title says the same thing as this comment.

```suggestion
```

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,452 @@
...
+    function isVoidReturningFunctionType(type: ts.Type): boolean {
+      const returnTypes = tsutils
+        .getCallSignaturesOfType(type)
+        .flatMap(signature =>
+          tsutils.unionTypeParts(signature.getReturnType()),
+        );
+      return (
+        // At least one return type is void
+        returnTypes.some(type =>
+          tsutils.isTypeFlagSet(type, ts.TypeFlags.Void),
```
JoshuaKGoldberg: [Question] This uses `Void` but the next check uses `VoidLike`. Swapping to any other combination of the type flags doesn't produce any test failures. Was there a reason to choose one vs. the other?

I'm genuinely asking, I don't know 😄

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,416 @@
...
+```
+
+</TabItem>
+</Tabs>
+
+### Unhandled returned promises
+
+If a promise is returned from a callback that should return void,
+it probably won't be awaited and its rejection will be silently ignored
+or crash the process depending on runtime.
```
JoshuaKGoldberg: [Docs] Some users bristle at definitive, declarative statements like _"if you do X, then Y will happen"_ when their design allows X without Y. Even if X and Y are not good.

```suggestion
If a callback is meant to return void, values returned from functions are likely ignored.
Ignoring a returned Promise means any Promise rejection will be silently ignored
or crash the process depending on runtime.
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,416 @@
...
+});
+```
+
+</TabItem>
+</Tabs>
+
+### Probable mistakes
+
+Returning a value from a void function is likely a mistake on part of the programmer.
+This rule will often warn you early when using a function in a wrong way.
```
JoshuaKGoldberg:
```suggestion
Returning a value from a void function often is an indication of incorrect assumptions about APIs.
Those incorrect assumptions can often lead to bugs.

The following `forEach` loop is a common mistake: its author likely either meant to add `console.log` or meant to use `.map` instead.
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,416 @@
...
+<TabItem value="✅ Correct">
+
+```ts
+['Kazik', 'Zenek'].forEach(name => console.log(`Hello, ${name}!`));
+```
+
+</TabItem>
+</Tabs>
+
+### Void context examples
```
JoshuaKGoldberg: [Docs] This is a solid deep dive, but I think it's a little too deep for the docs. Maybe it'd be good content for a blog post? We've never had a guest blog post before but this would be a lovely area to start with. 😄  

Suggestion: let's remove the _Void context examples_ and _Void context from overloads_, with a followup issue about a longer form blog post?

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,416 @@
...
+});
+```
+
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+Primitive values returned from void functions are usually safe.
+If you don't care about returning them you can use [`no-misused-promises`](./no-misused-promises.mdx) instead.
```
JoshuaKGoldberg:
```suggestion
Some projects are architected so that values returned from synchronous void functions are generally safe.
If you only want to check for misused voids with asynchronous functions then you can use [`no-misused-promises`](./no-misused-promises.mdx) instead.
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+In browser context, an unhandled promise will be reported as an error in the console.
+It's a always a good idea to also show some kind of indicator on the page that something went wrong,
```
JoshuaKGoldberg:
```suggestion
It's generally a good idea to also show some kind of indicator on the page that something went wrong,
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,416 @@
...
+but if you are just prototyping or don't care about that, the default behavior might be acceptable.
+In such case, instead of handling the promises and `console.error`ing them anyways, you can just disable this rule.
+
+Similarly, the default behavior of crashing the process on unhandled promise rejection
+might be acceptable when developing, for example, a CLI tool.
+If your promise handlers simply call `process.exit(1)` on rejection,
+you might as well not use this rule and rely on the default behavior.
```
JoshuaKGoldberg:
```suggestion
you may prefer to avoid this rule and rely on the default behavior.
```

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,452 @@
...
+    };
+
+    /** Checks whether the type is a void-returning function type. */
+    function isVoidReturningFunctionType(type: ts.Type): boolean {
+      const returnTypes = tsutils
+        .getCallSignaturesOfType(type)
+        .flatMap(signature =>
+          tsutils.unionTypeParts(signature.getReturnType()),
+        );
+      return (
```
JoshuaKGoldberg, line 114: [Refactor] This function includes the same _"is this set of return types a void return?"_ logic as `checkFunctionCallNode`. They even have the same two comments!

Since the logic is really the same, how about sharing a function?

```ts
function isVoidReturningReturnType(returnTypes: ts.Type[]) {
  return (
    // At least one return type is void
    returnTypes.some(type =>
      tsutils.isTypeFlagSet(type, ts.TypeFlags.Void),
    ) &&
    // The rest are nullish or any
    returnTypes.every(type =>
      tsutils.isTypeFlagSet(
        type,
        ts.TypeFlags.VoidLike |
          ts.TypeFlags.Undefined |
          ts.TypeFlags.Null |
          ts.TypeFlags.Any |
          ts.TypeFlags.Never,
      ),
    )
  );
}
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,416 @@
...
+
+document.addEventListener('click', () => {
+  console.log('Clicked');
+});
+```
+
+</TabItem>
+</Tabs>
+
+### Void context from base class
```
JoshuaKGoldberg: [Docs] The other headings are plural, so for consistency:

```suggestion
### Void context from base classes
```

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,452 @@
...
+        .flatMap(type =>
+          ts.isCallExpression(callTsNode)
+            ? type.getCallSignatures()
+            : type.getConstructSignatures(),
+        );
+
+      for (const [argIdx, argNode] of callNode.arguments.entries()) {
+        if (argNode.type === AST_NODE_TYPES.SpreadElement) {
+          continue;
+        }
```
JoshuaKGoldberg, line 174: [Testing] Removing this doesn't fail any unit tests. Either this check should exist and tests are missing, or this is dead code that can be removed.

If the reason it's here is to narrow `argNode`'s type down to `TSESTree.Expression` then maybe either the types are suggesting a legit case to handle, or are wrong?

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,452 @@
...
+     */
+    function checkObjectPropertyNode(
+      propNode: TSESTree.MethodDefinition | TSESTree.Property,
+    ): void {
+      if (
+        propNode.value.type === AST_NODE_TYPES.AssignmentPattern ||
+        propNode.value.type === AST_NODE_TYPES.TSEmptyBodyFunctionExpression
+      ) {
+        return;
+      }
```
JoshuaKGoldberg: [Testing] Removing this `if` check doesn't fail any unit tests. Either this check should exist and tests are missing, or this is dead code that can be removed.

Similar note to earlier on the type narrowing.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,452 @@
...
+     * In addition to the regular check against the contextual type,
+     * we also check against the base class property (when the class extends another class)
+     * and the implemented interfaces (when the class implements an interface).
+     */
+    function checkClassPropertyNode(
+      propNode: TSESTree.PropertyDefinition,
+    ): void {
+      if (propNode.value == null) {
+        return;
+      }
```
JoshuaKGoldberg, line 271: [Testing] Removing this `if` check doesn't fail any unit tests. Either this check should exist and tests are missing, or this is dead code that can be removed.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,452 @@
...
+     */
+    function checkClassMethodNode(methodNode: TSESTree.MethodDefinition): void {
+      if (
+        methodNode.value.type === AST_NODE_TYPES.TSEmptyBodyFunctionExpression
+      ) {
+        return;
+      }
+      if (methodNode.kind !== 'method') {
+        return;
+      }
```
JoshuaKGoldberg: [Testing] Removing this `if` checks doesn't fail any unit tests. Either they should exist and tests are missing, or this is dead code that can be removed.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,452 @@
...
+      AssignmentExpression: (node): void => {
+        if (['=', '||=', '&&=', '??='].includes(node.operator)) {
+          checkExpressionNode(node.right);
+        }
+      },
+      'CallExpression, NewExpression': checkFunctionCallNode,
+      JSXAttribute: (node): void => {
+        if (
+          node.value?.type === AST_NODE_TYPES.JSXExpressionContainer &&
+          node.value.expression.type !== AST_NODE_TYPES.JSXEmptyExpression
```
JoshuaKGoldberg, line 82: [Testing] Removing the `&& node.value... !== JSXEmptyExpression` doesn't fail any unit tests. Either this check should exist and tests are missing, or this is dead code that can be removed.

If the reason it's here is to narrow `node.value.expression`'s type down to `TSESTree.Expression` then maybe either the types are suggesting a legit case to handle, or are wrong?

packages/eslint-plugin/docs/rules/strict-void-return.mdx
JoshuaKGoldberg: Agreed with landing this first, then considering a deprecation as a followup. 

In fact, this rule is pretty big and scary. We don't really have a process for declaring rules as "canary" or "experimental". #8676 is the closest we have to a feature request. Maybe we should set a precedent?

(I don't think this PR should be blocked on that)

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: Added valid tests for JSXEmptyExpression (`prop={}`)  and JSXElement (`prop=<elem/>`) but I dunno how to get JSXSpreadChild. 
I tried `prop={...value}` but it doesn't work.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: This is just a coincidence. I simplified this condition so that it doesn't work for the other function anymore and now they are different. I think it makes sense to keep them separate.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: dunno

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: Spread arguments aren't checked because that would be non-trivial. So I don't know what to do here.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: I think in this case the types are wrong. 

AFAIK it's not possible to have TSEmptyBodyFunctionExpression as an object's property value. `{ f() }` is an syntax error. This is only valid in an abstract class body.

AssignmentPattern is also not valid as value of Property of ObjectLiteral, but it is valid for Property of ObjectPattern. Types could be improved by splitting Property into 2 types.

I will add type assertions for now.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: There is already a test for this and if I remove this check then the types are wrong. TS functions say they don't accept null but if you pass null they just do nothing and return more null so nothing crashes.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: Added test for empty body func.

Removing `kind == 'method'` actually makes the rule catch more cases.

packages/eslint-plugin/src/rules/strict-void-return.ts
JoshuaKGoldberg: Oh if that's just not doable in the AST then we can just ignore it. A good followup issue would be to tighten up the AST types to not include it.

packages/eslint-plugin/docs/rules/strict-void-return.mdx
phaux: Removed and changed other headings slightly

packages/eslint-plugin/docs/rules/strict-void-return.mdx
phaux: I already changed it but I just remembered my reasoning: You can only extend one class in a class declaration (unlike many interfaces) that's why this one was singular :laughing:

phaux:
I need help with some of the review comments

packages/eslint-plugin/docs/rules/strict-void-return.mdx
phaux: I can make a blog post WIP PR directly later if you don't mind. It would contain mostly the part I removed for now.

phaux:
@JoshuaKGoldberg ping

JoshuaKGoldberg:
Hi, sorry for the delay! I'll try to get to this Soon ™️

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,428 @@
...
+        type: 'object',
+        additionalProperties: false,
+        properties: {
+          allowReturnAny: {
+            type: 'boolean',
+            description:
+              'Whether to allow functions returning `any` to be used in place expecting a `void` function.',
+          },
+        },
+      },
```
JoshuaKGoldberg, line 45: [Feature] This is what [`@typescript-eslint/no-unsafe-return`](https://typescript-eslint.io/rules/no-unsafe-return) already does, right? I don't think we need this rule to also check for `any` safety - it's not the same area as strict `void` semantics. I think this rule should effectively hardcode `allowReturnAny` to enabled and not let users change that.

```suggestion
```

packages/eslint-plugin/src/rules/strict-void-return.ts
JoshuaKGoldberg: @phaux this is still the case - removing the `if` and leaving the `checkObjectPropertyNode(propNode)` still doesn't fail unit tests.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,428 @@
...
+      ObjectExpression: (node): void => {
+        for (const propNode of node.properties) {
+          if (propNode.type !== AST_NODE_TYPES.SpreadElement) {
+            checkObjectPropertyNode(propNode);
+          }
+        }
+      },
+      PropertyDefinition: checkClassPropertyNode,
+      ReturnStatement: (node): void => {
+        if (node.argument != null) {
```
JoshuaKGoldberg, line 97: [Testing] Removing this `if` and just calling `checkExpressionNode(node.argument);` doesn't fail any unit tests. Unnecessary code, or missing tests?

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,428 @@
...
+          checkExpressionNode(node.argument);
+        }
+      },
+      VariableDeclarator: (node): void => {
+        if (node.init != null) {
```
JoshuaKGoldberg, line 102: [Testing] Removing this `if` and just calling `checkExpressionNode(node.init)` doesn't fail any unit tests. Unnecessary code, or missing tests?

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -0,0 +1,428 @@
...
+        .flatMap(type =>
+          ts.isCallExpression(callTsNode)
+            ? type.getCallSignatures()
+            : type.getConstructSignatures(),
+        );
+
+      for (const [argIdx, argNode] of callNode.arguments.entries()) {
+        if (argNode.type === AST_NODE_TYPES.SpreadElement) {
+          continue;
+        }
```
JoshuaKGoldberg, line 174: [Testing] Removing this `if` altogether doesn't fail any unit tests. Unnecessary code, or missing tests?

packages/eslint-plugin/src/rules/strict-void-return.ts
JoshuaKGoldberg, line 1: [Praise] I normally requests folks use fewer comments, as they often just restate what's directly inferable from code. But this rule by nature deals with a lot of tricky points and I think the comments are very helpful here. Thank you for making it easier to read + review! 🙌

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,321 @@
...
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/strict-void-return** for documentation.
+
+## Rule Details
+
+TypeScript considers functions returning a value to be assignable to a function returning void.
```
JoshuaKGoldberg:
```suggestion
TypeScript considers functions returning a value to be assignable to a function returning `void`.
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,321 @@
...
+Using this feature of TypeScript can lead to bugs or confusing code.
+
+## Examples
+
+### Return type unsafety
+
+Passing a value-returning function in a place expecting a void function can be unsound.
```
JoshuaKGoldberg:
```suggestion
Passing a value-returning function in a place expecting a `void`-returning function can be unsound.
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,321 @@
...
+});
+```
+
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+Some projects are architected so that values returned from synchronous void functions are generally safe.
+If you only want to check for misused voids with asynchronous functions then you can use [`no-misused-promises`](./no-misused-promises.mdx) instead.
```
JoshuaKGoldberg:
```suggestion
If you only want to check for misused `void`s with asynchronous functions then you can use [`no-misused-promises`](./no-misused-promises.mdx) instead.
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,321 @@
...
+  someUntypedApi();
+});
+```
+
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+Some projects are architected so that values returned from synchronous void functions are generally safe.
```
JoshuaKGoldberg:
```suggestion
Some projects are architected so that values returned from synchronous `void` functions are generally safe.
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,321 @@
...
+If you only want to check for misused voids with asynchronous functions then you can use [`no-misused-promises`](./no-misused-promises.mdx) instead.
+
+In browser context, an unhandled promise will be reported as an error in the console.
```
JoshuaKGoldberg:
```suggestion
In browser contexts, an unhandled Promise will be reported as an error in the console.
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,321 @@
...
+It's generally a good idea to also show some kind of indicator on the page that something went wrong,
+but if you are just prototyping or don't care about that, the default behavior might be acceptable.
+In such case, instead of handling the promises and `console.error`ing them anyways, you can just disable this rule.
```
JoshuaKGoldberg:
```suggestion
In such cases you can disable this rule.
```

packages/eslint-plugin/docs/rules/strict-void-return.mdx
```diff
@@ -0,0 +1,321 @@
...
+
+Similarly, the default behavior of crashing the process on unhandled promise rejection
```
JoshuaKGoldberg:
```suggestion
Similarly, the default server runtime behaviors of crashing the process on unhandled Promise rejections
```

JoshuaKGoldberg (dismissed):
Ok! I must apologize for another long gap in reviewing. This rule intimidated me and I procrastinated on it far too long. Again, sorry.

I think it's roughly ready for merge. It's big and intricate enough that I'm positive we'll get a bunch of issues filed once people start using it. But I think a gradual rollout to let folks try it will make sense:

1. Merge roughly as-is, not in any preset configs
2. We can put a call out on social media for folks to try the rule
3. Eventually we can add it to `strictTypeChecked` - which'll get a lot of additional testing (recent prior art: `no-useless-default-assignment`)

I left some comments on things I think should block merge, but I feel comfortable applying them myself to save you the time (especially the unit tests splitting-up request). I'll just wait for another review from @typescript-eslint/triage-team.

Thanks again! ❤️

packages/eslint-plugin/src/rules/strict-void-return.ts
bradzacher: I think what you'll find is that without this guard everything still works cos it's not valid to have `...foo` where `foo: void`. That state is a typecheck error.

So for all intents and purposes you _could_ remove this condition and the rule would work the same. But having this here is mainly to keep the types sane.

bradzacher (dismissed):

bradzacher:
@JoshuaKGoldberg / @phaux whoever wants to get to it first -- once the comment are resolved we can get this landed.

JoshuaKGoldberg:
I'll start work on them now, to get them ready in time for today's release. 🚀

packages/eslint-plugin/src/rules/strict-void-return.ts
JoshuaKGoldberg: It was harder than I thought to get a failing test case around null variable inits. These are directly passed to `checker.getContextualType()`, which returns undefined in the case of an unknown/missing node. Ah well.

packages/eslint-plugin/src/rules/strict-void-return.ts
JoshuaKGoldberg: Heh, I tried for 10 minutes and couldn't get this to work either. Tricky!

PR #9714, opened 2024-08-03
fix(eslint-plugin): [no-unnecessary-type-parameters] ignore type arguments for infinite recursion protection

### LINKED ISSUE #9705: Bug: [no-unnecessary-type-parameters] Maximum call stack size exceeded

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.2&fileType=.ts&code=KYDwDg9gTgLgBDAnmYcCCA7AhjCBbRASQwGcYsMBjYAHgBU5QZgMATEuMqASwwHMAfHAC8cAPJ5uMGphz5E9AQBo4AcjBZYJVQIBQu0JFhxezKADMs1dNlwF6jEMzYcuvQXADeuuI%2BesAdSkACwAuOBoANT8Wdk4YHn4BAAoAOnSNLXDIgG0AXQBKESFZOyJScipaBgAfOEi9AF9dIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=&tokens=false

### Repro Code

```TypeScript
export type AnatomyInstance<T extends string> = Omit<Anatomy<T>, 'parts'>

export interface Anatomy<T extends string> {
  extendWith: <V extends string>(...parts: V[]) => AnatomyInstance<T | V>
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-unnecessary-type-parameters": "error"
  }
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

No error

### Actual Result

Maximum call stack size exceeded

### Additional Info

_No response_

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9705 

## Overview

As mentioned in https://github.com/typescript-eslint/typescript-eslint/issues/9705#issuecomment-2266685430, this rule has protection against infinite recursion, but it does not handle the case of infinite recursion with infinitely-expanding type arguments. Thus, I remove the type argument values from the recursion check.


abrahamguo:
Still in progress; need to investigate lint failures and make further updates

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -323,8 +323,11 @@ function collectTypeParameterUsageCounts(
   }
 
   function incrementTypeUsages(type: ts.Type): number {
-    const count = (typeUsages.get(type) ?? 0) + 1;
-    typeUsages.set(type, count);
+    const typeWithoutArguments = checker
+      .typeToString(type)
+      .replaceAll(/<.+?>/g, '');
```
JoshuaKGoldberg, line 328: Just commenting while it's in draft: we really, really try to avoid `typeToString`. Types can look exactly the same but be in two different places, and/or might only differ by whitespace.

```ts
export function one() {
  interface Box {
    value: ` ${" " | ""}`;
  }

  const box: Box = { value: " " };

  return box.value;
}

export function two() {
  interface Box {
    value: ` ${" " | ""}`;
  }

  const box: Box = { value: " " };

  return box.value;
}
```

https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVx8ACgEo4BvAKDjkxRmCgQEM1g4AhCEMyquAN0YAbJMABccAAZwAJKQBEcBQB9FcgL4SA3DzXkeaHAGd4AI07iOXALxl+QkeIUK12nlGAwkUXGZAA6AWFgbV1yUEhYRFQMbFwYAHcIYm4qGjoGZlZLFN5Ah0kZeUU4FTl1LR09KgMUYzhfC044G1I7IMdilyq4d09ves4A+2DyNSA

JoshuaKGoldberg:
Hmm, could you skip union types and just look at their constituents directly?

abrahamguo:
We could, but that would only handle the union case, and would not be a general solution. Like I mentioned above, it's possible to get into other recursive loops like `A<B[]>` => `A<B[][]>` => ...

and I haven't tested, but I assume that it's possible to create other loops, like `A<B & B & C>`, `A<CustomType<CustomType<T>>`, or `A<{ [K in key]: { [K in key]: T } }>`.

JoshuaKGoldberg:
Do you have repros for those? I'd think there's a very small list of cases we'd have to explicitly handle, and each should have a pretty clear way to avoid infinite recursion:

* Arrays (`T[]`) and tuples (`[T]`): skip into the type argument
* Intersections (`T & T`) and unions (`T | T`): skip into constituent types
* ...I don't follow how the other ones could happen, but if there's a repro I can look more closely

PR #9718, opened 2024-08-04
feat(eslint-plugin): add suggestion to `require-await` to remove `async` keyword

### LINKED ISSUE #9621: Enhancement: [require-await] Add suggestion to remove async keyword

https://typescript-eslint.io/rules/require-await

### Description

#6509 proposed adding an automatic fixer for the `require-await` rule, however that was rightly rejected because automatically applying a fix could change the type of the function and therefore break code.

A better option would be to add a [_suggestion_](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions) so that the problem can be semi-automatically fixed, _but only when you ask the editor_ to apply the suggestion.

### Fail

```TypeScript
const arrow = async () => {
    return;
};
```

### Pass

```TypeScript
const arrow = () => {
    return;
};
```

### Additional Info

_No response_

reduckted:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9621

## Overview

Implements the suggestion that was added in https://github.com/eslint/eslint/issues/18713 (pull request: https://github.com/eslint/eslint/pull/18716)

The `require-await` rule now includes a suggestion to remove the `async` keyword.

This mirrors the behavior of the underlying ESLint rule's suggestion. If the function has a return type annotation of `Promise<T>`, the suggestion will also change the return type annotation to `T`, or if it has a return type annotation of `AsyncGenerator<T>`, the suggestion will also change the return type annotation to `Generator<T>`.


reduckted:
@kirkwaiblinger Could you please reopen #9621, as the feature has now been merged into ESLint.

packages/eslint-plugin/src/rules/require-await.ts
kirkwaiblinger, line 1: Question regarding the description: 

> This mirrors the behavior of the underlying ESLint rule's suggestion, with one difference. When the function has a type annotation for the return type, the suggestion will not be provided because removing the `async` keyword in those scenarios would result in broken code.
> 
> For example, suggesting to remove the `async` keyword in this case is fine. The inferred return type will change, but the function itself is still valid:
> 
> ```ts
> async function foo() {
>   return 0;
> }
> 
> // Result of applying the suggestion:
> function foo() {
>   return 0;
> }
> ```
> 
> But removing the return type in this case would cause a compilation error:
> 
> ```ts
> async function foo(): Promise<number> {
>   return 0;
> }
> 
> // Result of applying the suggestion:
> function foo(): Promise<number> {
>   return 0;
> //~~~~~~ Type 'number' is not assignable to type 'Promise<number>'. ts(2322)
> }
> ```

I'm kind of thinking that it's still ok to put the suggestion in there. After all, you can easily break the build in the other case too....

```ts
const inferredReturnType = async () => 42;
inferredReturnType().then(console.log);
```
suggests to
```ts
const inferredReturnType = () => 42;
inferredReturnType().then(console.log);
```

In a sense, you could argue that the inferred return version is more disruptive since it results in broken code at a distance rather than in the function signature that has just been changed. 

But my thought is both cases are fair game for suggestions. It's just the user will have to do some cleanup afterwards, which the compiler will help them with.

What do you think?

kirkwaiblinger (commented):
Thanks for sending this in! I don't have time for a full review right now, but I wanted to get the question about return types in front of you to contemplate early, in case it affects the PR substantially.

packages/eslint-plugin/src/rules/require-await.ts
reduckted: Yeah, that's a good point. I'm happy to make that change.

🤔 I wonder if it should also change the return type annotation from `Promise<T>` to `T`.

kirkwaiblinger: Good question! I tentatively think it would be the right move to turn `Promise<T>` into `T` since our extension specifically checks that you're not returning a promise before reporting. 👍

But if some counterargument or tricky example comes to mind we can rethink

packages/eslint-plugin/tests/rules/require-await.test.ts
```diff
@@ -305,6 +325,38 @@ const numberOne = async function (): Promise<number> {
...
+            {
+              output: 'const numberOne = (): number => 1;',
+              messageId: 'removeAsync',
+            },
+          ],
+        },
+      ],
+    },
+    {
+      // Return type with nested generic type
```
kirkwaiblinger, line 338: nit 
```suggestion
      // Return type with nested type argument
```

packages/eslint-plugin/tests/rules/require-await.test.ts
```diff
@@ -338,6 +402,16 @@ async function* foo(): void {
           data: {
             name: "Async generator function 'foo'",
           },
+          suggestions: [
+            {
+              output: `
+function* foo(): void {
```
kirkwaiblinger, line 408: (no action requested) 

Oh... This is a silly test case, since generators _cannot_ have `void` return.....

But, granted the rule flags, I think this is the correct suggestion; just remove the `async` and don't do anything else.

packages/eslint-plugin/tests/rules/require-await.test.ts
kirkwaiblinger, line 1: Please add test case for 
-  ```ts
    // intentional TS error
    async function* foo(): Promise<number> {
       yield 1;
    }
    
    // suggest to the following...
    // in this case, do not mess with the return type, since `Promise<T>` is not valid for a generator. 
    function* foo(): Promise<number> {
        yield 1;
    }
     ```

packages/eslint-plugin/tests/rules/require-await.test.ts
```diff
@@ -667,6 +939,169 @@ for await (let num of asyncIterable) {
...
+              messageId: 'removeAsync',
+            },
+          ],
+        },
+      ],
+    },
+    {
+      // This test must not have a semicolon after "foo".
+      // eslint-disable-next-line @typescript-eslint/internal/plugin-test-formatting
+      code: `
```
kirkwaiblinger, line 1045: nit - we have a `noFormat` tag we like to use for intentionally unformatted test cases 
```suggestion
      code: noFormat`
```

packages/eslint-plugin/tests/rules/require-await.test.ts
```diff
@@ -667,6 +939,169 @@ for await (let num of asyncIterable) {
...
+        foo
+        async () => {
+          return 0;
+        }
+      `,
+      languageOptions: {
+        parserOptions: {
+          ecmaVersion: 2022,
```
kirkwaiblinger, line 1053: I'm not sure this is necessary? Or if so, leave a comment?

packages/eslint-plugin/src/rules/require-await.ts
```diff
@@ -75,13 +84,108 @@ export default createRule({
...
+              (nextToken.value === '[' || nextToken.value === '(') &&
+              (nodeWithAsyncKeyword.type === AST_NODE_TYPES.MethodDefinition ||
+                isStartOfExpressionStatement(nodeWithAsyncKeyword)) &&
+              needsPrecedingSemicolon(context.sourceCode, nodeWithAsyncKeyword);
+
+            const changes = [
+              { range: asyncRange, replacement: addSemiColon ? ';' : '' },
+            ];
+
+            // If there's a return type annotation and it's a
```
kirkwaiblinger, line 134: Let's _not_ do this if the function is a generator function (see also comment on test file), since it doesn't make semantic sense. 

We _could_ consider doing `AsyncGenerator` -> `Generator` if provided... I leave it to you as an option whether to a) do nothing with the return type if specified on a generator, or b) change the type if and only if it's an `AsyncGenerator`.  Please add a test case for whichever route you go 🙂. 

PS - today I learned that this rule inspects generator functions 🙃.

kirkwaiblinger (changes requested):
Another high-level pass, a few minor changes requested. Main blocker is the question about async generators. 

Good stuff!

packages/eslint-plugin/tests/rules/require-await.test.ts
reduckted: Interesting. It was needed in the ESLint tests because the parser didn't support the syntax for some of these tests (the test above needed it in the `eslint` repo, but I may have just copied it for the test corresponding to this one). You're correct that it's not necessary here (for any of the tests), so I'll remove it. Must be a difference from the espree parser.

packages/eslint-plugin/src/rules/require-await.ts
reduckted: I choose to go with replacing `AsyncGenerator` with `Generator` since it was easy enough to do.

packages/eslint-plugin/tests/rules/require-await.test.ts
```diff
@@ -134,14 +134,6 @@ async function testFunction(): Promise<void> {
       async value => Promise.resolve(value),
     ),
   );
-}
-    `,
-    'async function* run() {}',
```
reduckted, line 139: I reviewed the valid test cases from ESLint and found that some of them were missing. This was one of them:
https://github.com/eslint/eslint/blob/0dd38631b8dd60f93807e9cee1df3e752b86ab51/tests/lib/rules/require-await.js#L77

...as was the test below:
https://github.com/eslint/eslint/blob/0dd38631b8dd60f93807e9cee1df3e752b86ab51/tests/lib/rules/require-await.js#L68-L72

...and the one a bit further down on line 217 (except that one had a return type annotation:
https://github.com/eslint/eslint/blob/0dd38631b8dd60f93807e9cee1df3e752b86ab51/tests/lib/rules/require-await.js#L81

I've moved them down into the "base eslint tests" section.

packages/eslint-plugin/src/rules/require-await.ts
```diff
@@ -75,13 +84,118 @@ export default createRule({
...
+                    token.value === '<',
+                );
+                const closeAngle = context.sourceCode.getLastToken(
+                  node.returnType.typeAnnotation,
+                  token =>
+                    token.type === AST_TOKEN_TYPES.Punctuator &&
+                    token.value === '>',
+                );
+                if (openAngle && closeAngle) {
+                  changes.unshift(
```
kirkwaiblinger, line 167: If I change this to use `.push()`, all the test cases pass, and it's easier to read. Any reason to keep it as `unshift()`? (here and above)
```suggestion
                  changes.push(
```

packages/eslint-plugin/src/rules/require-await.ts
```diff
@@ -75,13 +84,118 @@ export default createRule({
...
+                    // Remove the closing angled bracket.
+                    { range: closeAngle.range, replacement: '' },
+                    // Remove the "Promise" identifier
+                    // and the opening angled bracket.
+                    {
+                      range: [
+                        node.returnType.typeAnnotation.typeName.range[0],
+                        openAngle.range[1],
+                      ],
+                      replacement: '',
```
kirkwaiblinger, line 177: (nit) Rather than using `replaceTextRange()` with `replacement: ''`, would you mind refactoring this to use `fixer.removeTextRange()`? Just for simplicity. I realize this will require you to move a few things around.

packages/eslint-plugin/src/rules/require-await.ts
```diff
@@ -75,13 +84,118 @@ export default createRule({
...
+                  hasTypeName(node.returnType.typeAnnotation, 'AsyncGenerator')
+                ) {
+                  changes.unshift({
+                    range: node.returnType.typeAnnotation.typeName.range,
+                    replacement: 'Generator',
+                  });
+                }
+              } else if (
+                hasTypeName(node.returnType.typeAnnotation, 'Promise') &&
+                !!node.returnType.typeAnnotation.typeArguments
```
kirkwaiblinger, line 152: (nit, optional) my personal preference, feel free to ignore
```suggestion
                node.returnType.typeAnnotation.typeArguments != null
```

packages/eslint-plugin/src/rules/require-await.ts
```diff
@@ -75,13 +84,118 @@ export default createRule({
...
+              if (scopeInfo.isGen) {
+                if (
+                  hasTypeName(node.returnType.typeAnnotation, 'AsyncGenerator')
+                ) {
+                  changes.unshift({
+                    range: node.returnType.typeAnnotation.typeName.range,
+                    replacement: 'Generator',
+                  });
+                }
+              } else if (
```
kirkwaiblinger, line 150: (no action requested) grrrr this is why i hate `no-lonely-if` 😆 https://github.com/typescript-eslint/typescript-eslint/pull/9547

kirkwaiblinger (changes requested):
This is great stuff! 

I haven't seen any issues with the actual rule logic, just requesting a some changes around our repo's code style and some unnecessarily defensive conditions. 

Awesome work! Getting close!

packages/eslint-plugin/src/rules/require-await.ts
```diff
@@ -75,13 +84,118 @@ export default createRule({
...
+            node.parent.value === node)
+            ? node.parent
+            : node;
+
+        const asyncToken = context.sourceCode.getFirstToken(
+          nodeWithAsyncKeyword,
+          token => token.value === 'async',
+        );
+
+        if (asyncToken) {
```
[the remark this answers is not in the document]
reduckted: Oh, sweet. I didn't know about `nullThrows`. I initially added a non-null assertion, but of course the the lint rule prohibits that, hence why it ended up with the conditions. I'll remove all the unnecessary conditions then. 👍

packages/eslint-plugin/src/rules/require-await.ts
reduckted: Ah, I think my assumption about how multiple edits are handled was incorrect. My thinking was that because removing the `async` keyword is always the first change, we'd want to apply this change first so that the range we specify here doesn't end up referring to a different piece of text once the `async` keyword is removed. 

Given how that doesn't seem to be the case, I'm now assuming that ESLint must adjust subsequent ranges when it applies an edit. I took a look at what ESLint does, but couldn't work out where it actually applies the changes. 😕

packages/eslint-plugin/src/rules/require-await.ts
kirkwaiblinger: Oh, for this I was thinking more like 
```ts
const fix: ReportFixFunctino = fixer => {
  const fixes: RuleFix[] = [];
    
  // ...
    
  fixes.push(
    fixer.replaceTextRange(asyncRange, addSemiColon ? ';' : ''),
  );
    
  // ...
    
  fixes.push(fixer.removeRange(closeAngle.range));
  fixes.push(
    fixer.removeRange([
      node.returnType.typeAnnotation.typeName.range[0],
      openAngle.range[1],
    ]),
  );
  
  // ....
  
  return fixes;
}
```

if that makes sense. But the way you have it is fine to keep too! 🙂

kirkwaiblinger (approved):
This is great stuff! Up to you if you want to make any changes with respect to https://github.com/typescript-eslint/typescript-eslint/pull/9718#discussion_r1717701460 or keep as is 🙂 

Thanks for handling this!

JoshuaKGoldberg (approved):
The comments in there were really helpful for review, thanks 😄

PR #9737, opened 2024-08-06
fix(eslint-plugin): [no-unnecessary-type-parameters] clarify message

### LINKED ISSUE #9664: Bug: [no-unnecessary-type-parameters] misleading error messages

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.2&fileType=.ts&code=GYVwdgxgLglg9mABABwIYCcDOBTAUphAHgBVFsAPKbMAE00UynRjAHMA%2BACheRCgC4GTFqwCUg0gG8AUIkTpsUEOiS4AygHkAcgDo0WbNzC8ooxKnrEA3NIC%2BQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false

### Repro Code

```TypeScript
// T is used twice - maybe it should say "once in the signature"?
function parseJson<T extends string>(input: string): T {
  return JSON.parse(input) as T;
}

// T is used zero times
function parseJson<T extends string>(input: string): any {
  return JSON.parse(input) as any;
}
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/no-unnecessary-type-parameters": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true
  }
}
```

### Expected Result

I expect the error message should reflect correctly the number of times the type parameter is used.

### Actual Result

"Type parameter T is used only once." in both cases

### Additional Info

Let's also include a doc change to the docs page, if counting parameters is only supposed to be counting the usages in the signature?

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9664

## Overview

Adds _"in the (class|function) signature"_ to clarify the message. Also, shoutout @kirkwaiblinger, updated the rule's description and message to be clear that it reports when a type parameter _isn't used multiple times_. Using a type parameter 0 times or 1 time both cause a report.

💖


packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -21,7 +21,7 @@ export default createRule({
       recommended: 'strict',
     },
     messages: {
-      sole: 'Type parameter {{name}} is used only once.',
+      sole: 'Type parameter {{name}} is used only once in the function signature.',
```
JoshuaKGoldberg, line 24: @kirkwaiblinger I added _"function"_ because I think some users won't know what _"signature"_ means.

kirkwaiblinger: > @kirkwaiblinger I added _"function"_ because I think some users won't know what _"signature"_ means.

Yep, no opinion on this particular minutia 👍 

***

I wonder, though, don't we need to be a little careful to get the node type correct?

The rule reports on types that aren't necessarily in function/method signatures, for example: 
```ts
        class Joiner<T extends string | number> {
          join(el: T, other: string) {
            return [el, other].join(',');
          }
        }
```

JoshuaKGoldberg: Ah true... hmm........

JoshuaKGoldberg: Switched the descriptor to _"class signature"_ for those.

kirkwaiblinger (approved):
Looks good to me on the "in the signature" changes, with appropriate node type 👍 

***

Do we want to consider the part of the issue too that complains about it saying "used once" when it's used zero times? I think that change _should_ be pretty simple since I think we can just use the `identifierCounts` right above the `context.report()` call to differentiate between 0 and 1 usages.  But no need to include this if you think it's inadvisable

kirkwaiblinger (approved):
Yay!
happy Monday coffee

JoshuaKGoldberg:
I'll fix up the merge conflict & formatting issues locally. 🚀

JoshuaKGoldberg:
Oh and fyi @danvk - we're hoping to get this in today, but if you have more changes you'd want, let me know! I'm on task to apply them.

danvk:
LGTM on the wording.

I'm confused why we care about zero times in the context of this rule, though. That would be an unused type parameter, and TypeScript already [complains about those][1].

image

[1]: https://typescript-eslint.io/play/#ts=5.5.2&fileType=.tsx&code=GYVwdgxgLglg9mABMOcA8AVAfACgJSIDeAUIogE4CmUI5SAjAEwDcxAvsUA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

JoshuaKGoldberg:
Ah, that's a very good point. Agreed - the rule shouldn't handle that in the first place! I'm stepping away and can't file an issue at the moment, but should have time later this week if you haven't by then.

Between this, #9740, and #9741, I'm feeling good about solidifying this rule. 🙂

PR #9740, opened 2024-08-06
fix(eslint-plugin): [no-unnecessary-type-parameters] check mapped constraint types if necessary

### LINKED ISSUE #9667: Bug: [no-unnecessary-type-parameters] false positive when parameter is used in inferred mapped return type

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.2&showAST=es&fileType=.tsx&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXwFsoAHYkYAJRA2RlQBUBPMgHnvhAA8MRVgBnePwwwsqAOYA%2BABScAXPHoBKBQG94AbQDS8MYoC6CgMIksGKBCwAvECy2T4AXwBQzlOmx5dqRCBhxgAFkSMkpqWgZmW3YuHj5BYVEJGXlFJXhVZ3h4OBo6QhDyKjzIslklAG5nF2cAN1hvX39yeABeRr8A4NIi8LomMoAicV4-LDBdQWR%2BckGlZyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJADuIppHBgAviF1A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false

### Repro Code

```TypeScript
declare function mappedReturnType<T extends string>(x: T): { [K in T]: Capitalize<K> }

function inferredMappedReturnType<T extends string>(x: T) {
  return mappedReturnType(x);
}

var inferred = inferredMappedReturnType("generic is used")
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-unnecessary-type-parameters": "warn"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true
  }
}
```

### Expected Result

I expected no error

### Actual Result

Error on the type parameter in `inferredMappedReturnType` - "Type parameter T is used only once".

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9667

## Overview

When a mapped type is implicitly the return type of a function, and its reference to the type parameter is a _constraint_, we need to check that too. That'll happen in the case of `type.templateType` not being filled out I think.

💖


packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -303,7 +303,7 @@ function collectTypeParameterUsageCounts(
         if (properties.length === 0) {
           // TS treats mapped types like `{[k in "a"]: T}` like `{a: T}`.
           // They have properties, so we need to avoid double-counting.
-          visitType(type.templateType, false);
+          visitType(type.templateType ?? type.constraintType, false);
```
danvk, line 306: From https://github.com/typescript-eslint/typescript-eslint/pull/9530:

> The rule previously descended into the typeParameter. It doesn't need to descend into the constraintType because we get that through another code path. It does need to descend into the templateType, though.

I wish I'd included more details about what "another code path" was and what my example was. Since you touched this more recently, do you have an idea? Double-counting would be problematic for this rule since it would lead to false negatives. On the other hand, if we only require one appearance of `T` in the return slot, then we can double-count all we want.

I'd also like to double-check whether this comment has any implications for this change:

> One wrinkle is that if the mapped type is something like {[k in 'a']: T}, then TypeScript will treat it like {a: T} and we'll process it as an object type. In this case, we need to avoid also processing it as a mapped type, even though isMappedType returns true.

danvk (changes requested):

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
JoshuaKGoldberg: I ... don't 😅. I couldn't figure out how to construct a test case that shows a flaw in this changed logic.

Same with the mapped types like `{ [k in 'a'] ... }`.

kirkwaiblinger (approved):
I don't have a deep understanding of this, but, approving for the fact that it seems plausible, and the tests pass.

PR #9741, opened 2024-08-06
fix(eslint-plugin): [no-unnecessary-type-parameters] check mapped alias type arguments

### LINKED ISSUE #9721: Bug: [no-unnecessary-type-parameters] False positive on curried functions

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.4&showAST=types&fileType=.ts&code=C4TwDgpgBAYg9nAPAFQDRQMoD4oF5NQQAewEAdgCYDOUARFXALYQBGcFItAsAFBRQB%2BKMkIly1OnDIBjCNz6C6wOABsK8-gC46zDVG1kIANwgAnANy9eFCNJUBDU9ABmAVxnAAllKjOEKLAAKCntge21kAEptRGxAuGAACzNtDEi8HHgkNEwsSx4gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
type Foo<T, S> = S extends "somebody"
  ? T extends "once"
  ? "told"
  : "me"
  : never;

declare function foo<T>(data: T): <S>(other: S) => Foo<T, S>;
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### tsconfig

_No response_

### Expected Result

T shouldn't be surfaced by this rule; it has a meaningful usage in the type definitions.

### Actual Result

T is surfaced as only being used once, incorrectly.

### Additional Info

I maintain the [Remeda typescript utility library](https://github.com/remeda/remeda). We rely heavily on curried functions, which are functions that return functions. The rule is firing for almost all our complex types where the type is passed on to the return type of the curried function. The non-curried functions, which are identical in terms of how the type param is used, are not falsely detected, so this has something to do with the param being passed on "through" to the return type of its returned function type.

When trying to reproduce this, I noticed that simpler cases where the T is used directly in the type aren't falsely detected; the issue can only be reproduced once we start introducing ternary conditions on other type params.

You can see a fuller example via the PR, which tries to introduce this rule into our codebase: https://github.com/remeda/remeda/pull/801

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9721

## Overview

I'm not 100% on my understanding here. But what I think is happening is that TypeScript is inlining the resolved parts of the conditional type when possible. `type.aliasTypeArguments` is the piece that still has a reference to the type arguments `<T, S>`.

💖


kirkwaiblinger (dismissed):
I _think_ this makes sense. Yay for test cases 😅

danvk (changes requested):
I don't think this is the right way to handle this case.

I pulled this change down locally and set a breakpoint on line 292 to see which tests hit the new code path. This was the first one:

```
> type
"Extract<AccessorPropertyComputedName, { type: NodeType; }>"
> type.aliasSymbol.declarations[0].getText()
'type Extract<T, U> = T extends U ? T : never;'
> > type.aliasTypeArguments.map(t => type.checker.typeToString(t))
[ 'AccessorPropertyComputedName', '{ type: NodeType; }' ]
> checkType
AccessorPropertyComputedName
> extendsType
{ type: NodeType; }
```

It's pretty weird that TS "inlines" type aliases that are conditional types like this. But since the same types appear in both `type.aliasTypeArguments` as `type.checkType` and `type.extendsType`, this change means we'll start double-counting them.

#9721 shows that we need the recursive call via `aliasTypeArguments`. So in the case that a type is both a type alias instantiation _and_ a conditional type, I think it's safer to only treat it as a type alias instantiation and ignore the extra information.

In practice, that just means moving the last "catch-all" higher up. This lets us remove some duplicated logic. Here's a commit that does that, all tests still pass:
https://github.com/danvk/typescript-eslint/commit/bacd28d0695202c6816e69e57e3f9b7ca36585f8

JoshuaKGoldberg:
Oh awesome! @danvk do you have bandwidth to send a PR? I can add that better change to this PR if not.

danvk:
> Oh awesome! @danvk do you have bandwidth to send a PR? I can add that better change to this PR if not.

Sure, here's a PR to merge that commit into your PR branch https://github.com/JoshuaKGoldberg/typescript-eslint/pull/323

kirkwaiblinger (approved):

danvk (approved):

PR #9079, opened 2024-05-12
feat(eslint-plugin): apply initial config changes for v8

JoshuaKGoldberg:
BREAKING CHANGE:
Adjusts the shared configs.

## PR Checklist

- [x] Addresses an existing open issue: fixes #9078
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Applies the changes that were suggested in #8914.

Note that this is into the `v8` branch, not `main`. It'll only impact the `v8` line of versions. We'll try this out on a set of community repositories to see how it feels.

💖

PR #9743, opened 2024-08-07
chore: set `reportUnusedDisableDirectives` level to `error`

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/pull/9119#discussion_r1661294146

## Overview

Now, unused disable directives will block the pipeline. I cannot think of any reason not to have this, and per the linked comment, @auvred and @JoshuaKGoldberg support it 👍

### DIFF
--- a/packages/typescript-eslint/src/index.ts
+++ b/packages/typescript-eslint/src/index.ts
@@ -71,7 +71,6 @@ const configs = {
 export type Config = TSESLint.FlatConfig.ConfigFile;
 export type { ConfigWithExtends };
 /*
-eslint-disable-next-line import/no-default-export --
 we do both a default and named exports to allow people to use this package from
 both CJS and ESM in very natural ways.
 
--- a/packages/typescript-estree/tests/lib/convert.test.ts
+++ b/packages/typescript-estree/tests/lib/convert.test.ts
@@ -29,7 +29,6 @@ describe('convert', () => {
       function fakeUnknownKind(node: ts.Node): void {
         ts.forEachChild(node, fakeUnknownKind);
         // @ts-expect-error -- intentionally writing to a readonly field
-        // eslint-disable-next-line deprecation/deprecation, @typescript-eslint/no-unused-expressions
         node.kind = ts.SyntaxKind.UnparsedPrologue;
       }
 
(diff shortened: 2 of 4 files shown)


packages/typescript-estree/tests/lib/convert.test.ts
```diff
@@ -437,10 +436,10 @@ describe('convert', () => {
     it('allows writing to the deprecated getter property as a new enumerable value', () => {
       const tsMappedType = getEsTsMappedType();
 
-      // eslint-disable-next-line deprecation/deprecation, @typescript-eslint/no-unused-expressions
+      // eslint-disable-next-line deprecation/deprecation
       tsMappedType.typeParameter = undefined!;
 
-      // eslint-disable-next-line deprecation/deprecation, @typescript-eslint/no-unused-expressions
+      // eslint-disable-next-line deprecation/deprecation
```
abrahamguo, line 442: too many `@typescript-eslint/no-unused-expressions` were accidentally added in https://github.com/typescript-eslint/typescript-eslint/pull/9079/commits/aa5525f0dfd15265ebd4a364dcc5aca12bfc676d

packages/typescript-estree/tests/lib/convert.test.ts
```diff
@@ -29,7 +29,6 @@ describe('convert', () => {
       function fakeUnknownKind(node: ts.Node): void {
         ts.forEachChild(node, fakeUnknownKind);
         // @ts-expect-error -- intentionally writing to a readonly field
-        // eslint-disable-next-line deprecation/deprecation, @typescript-eslint/no-unused-expressions
```
abrahamguo, line 32: re: `deprecation/deprecation`, `ts.SyntaxKind` has actually been removed from TypeScript altogether (!) — this line is not throwing any errors, though, because of the unrelated `ts-expect-error`

- `fakeUnknownKind` was originally added in #1525 
- `deprecation/deprecation` suppression added in #6570 
- `TSUnparsedPrologue` removed from snapshots (and TypeScript completely) in #9397

packages/typescript-eslint/src/index.ts
```diff
@@ -71,7 +71,6 @@ const configs = {
 export type Config = TSESLint.FlatConfig.ConfigFile;
 export type { ConfigWithExtends };
 /*
-eslint-disable-next-line import/no-default-export --
```
abrahamguo, line 74: Looks like this rule is actually not working correctly in any file — `context.parserOptions.sourceType` always comes back as `undefined`, whereas this rule needs it to be `module`

packages/utils/src/ts-eslint/ESLint.ts
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable @typescript-eslint/no-namespace */
```
abrahamguo, line 1: Namespaces were moved to other files in #8861

auvred (approved):
Nice!

packages/typescript-eslint/src/index.ts
bradzacher: That's because we don't set it via parser options.
https://github.com/typescript-eslint/typescript-eslint/blob/main/eslint.config.mjs#L90-L94

It shouldn't be reading the parser options and should instead be reading it from the parsed result, really.

Probs worth reporting upstream.

JoshuaKGoldberg (approved):
Love it!

Great descriptions inline too 😍

PR #9777, opened 2024-08-11
fix(eslint-plugin): [no-unsafe-enum-comparison] add logic to see through intersections

### LINKED ISSUE #7114: Bug: [no-unsafe-enum-comparison] false positive with enums in a union with branded types

### Playground Link

https://typescript-eslint.io/play/#ts=5.0.3&sourceType=module&showAST=es&code=KYOwrgtgBAYg9nKBvAsAKClAglAvFAcgEMCAadAX3XQBcBPAB2CgCEiAnPKACgGcb2ASxABzKADJkUAPrSARuyIgAJgC4oYEAGsQcAO4goFAJRQAPrAQBuammXAAxgBsOzB3BD8ocjurbsbNHRBADMeH05cKMI5AlNUDCgAeiTKdCA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tieQEMAZonpNYAW1pkOY4nyrIOTdFETRoHaJHBgAviB1A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
enum Foo {
  A = 'a',
}

type Bar = (string & { __brand: unknown }) | Foo;

declare const bar: Bar;

if (bar === 'b') {
  //
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-unsafe-enum-comparison": ["error"],
  },
};
```

### tsconfig

_No response_

### Expected Result

No error

### Actual Result

The rule errors on the comparison because it doesn't recognise the branded type as the base type

### Additional Info

_No response_

Zamiell:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7114

## Overview

Added additional logic for the rule to "see through" intersections, allowing the rule to work properly in the case outlined in the linked issue.

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
@@ -227,16 +227,6 @@ ruleTester.run('strict-enums-comparison', rule, {
         Beet = 'beet',
         Celery = 'celery',
       }
-      type WeirdString = string & { __someBrand: void };
-      declare const weirdString: WeirdString;
-      Vegetable.Asparagus === weirdString;
-    `,
-    `
-      enum Vegetable {
-        Asparagus = 'asparagus',
-        Beet = 'beet',
-        Celery = 'celery',
-      }
       const foo = {};
       const vegetable = Vegetable.Asparagus;
       vegetable in foo;
--- a/packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
@@ -1213,5 +1213,38 @@ ruleTester.run('no-unsafe-enum-comparison', rule, {
       `,
       errors: [{ messageId: 'mismatchedCondition' }],
     },
+    {
+      code: `
+        enum Fruit {
+          Apple,
+        }
+        declare const foo: number & { __someBrand: void };
+        if (foo === Fruit.Apple) {
+        }
+      `,
+      errors: [{ messageId: 'mismatchedCondition' }],
+    },
+    {
+      code: `
+        enum Vegetable {
+          Asparagus = 'asparagus',
+        }
+        declare const foo: string & {};
+        if (foo === Vegetable.Asparagus) {
+        }
+      `,
+      errors: [{ messageId: 'mismatchedCondition' }],
+    },
+    {
+      code: `
+        enum Vegetable {
+          Asparagus = 'asparagus',
+        }
+        declare const foo: string & { __someBrand: void };
+        if (foo === Vegetable.Asparagus) {
+        }
+      `,
+      errors: [{ messageId: 'mismatchedCondition' }],
+    },
   ],
 });
--- a/packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
@@ -1212,5 +1212,16 @@ ruleTester.run('strict-enums-comparison', rule, {
         },
       ],
     },
+    {
+      code: `
+        enum Fruit {
+          Apple,
+        }
+        declare const foo: number & {};
+        if (foo === Fruit.Apple) {
+        }
+      `,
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
Yeah, `AtomicsWaitError` has been happening once in a while... cc @JamesHenry, I see it a few times a month 😞

packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
```diff
@@ -221,16 +221,6 @@ ruleTester.run('no-unsafe-enum-comparison', rule, {
...
-    `
-      enum Vegetable {
-        Asparagus = 'asparagus',
-        Beet = 'beet',
-        Celery = 'celery',
-      }
-      type WeirdString = string & { __someBrand: void };
-      declare const weirdString: WeirdString;
-      Vegetable.Asparagus === weirdString;
-    `,
```
JoshuaKGoldberg, line 233: [Testing] Could you keep this case in the file, please? IMO it's good to also check the case of a property like `__someBrand`. Sometimes empty object types in an intersection are handled differently on the TypeScript side.

packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
JoshuaKGoldberg, line 1: [Testing] This PR is touching both the _number_ and _string_ areas of logic. But its test changes are only impacting a _string_ typed enum. Could you add some testing for the _number_ case too please?

JoshuaKGoldberg (dismissed):
This is looking nice and clean! Just requesting changes on tests 🚀

packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
Zamiell: I don't understand the test. Specifically, consider the following code:

```
type String = string;
declare const string: String;
Vegetable.Asparagus === string;
```

Currently, this code flags the "no-unsafe-enum-comparison" rule, which seems correct, because it is not possible for the `String` type to contain an enum. So far so good.

So why would this logic change we expand the scope to intersection? Even though `WeirdString` is an intersection, the same logic should apply. In other words, it is not possible for the intersection of `WeirdString` to *ever* contain a vegetable, so therefore we should want the rule to flag this.

packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
Zamiell: I added some more test cases, let me know if this is sufficient

packages/eslint-plugin/tests/rules/no-unsafe-enum-comparison.test.ts
JoshuaKGoldberg: It's a [Branded Type](https://github.com/LearningTypeScript/site/pull/135). Folks do this sometimes and use strings as the actual runtime value.

Zamiell: Right, I used branded types all the time myself. But how does testing for a branded type make sense with respect to what I wrote in my previous post?

JoshuaKGoldberg: I just want to make sure that the logic _doesn't_ change.

Zamiell: I don't know what you mean by "logic". After my PR, the rule will trigger in more situations than it did before, which causes the test case failure.

JoshuaKGoldberg: By "logic" I mean the rule's handling of this case. As in, I want to make sure that branded types are still something in the tests. 

👍 you do have them separately though so I think this thread is resolved. Thanks 😄

Issue #8988, opened 2024-04-24
Rule proposal: Warn when deprecated APIs are used (eslint-plugin-deprecation)


JoshuaKGoldberg:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+new+plugin+rule%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### My proposal is suitable for this project

- [X] My proposal specifically checks TypeScript syntax, or it proposes a check that requires type information to be accurate.
- [X] My proposal is not a "formatting rule"; meaning it does not just enforce how code is formatted (whitespace, brace placement, etc).
- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Description

Re-opening #1223: being able to lint against usage of an API marked as `@deprecated` is a common request for TypeScript developers. It was a [TSLint rule](https://palantir.github.io/tslint/rules/deprecation) back in those days, and now is provided primarily by https://github.com/gund/eslint-plugin-deprecation. `eslint-plugin-deprecation` existing is why #1223 was closed by a docs change pointing folks externally.

But! Since then:

* We've stabilized as a repository and IMO this is an important + often-useful enough rule to justify being included (so folks don't have to use a third-party plugin)
* `eslint-plugin-deprecation` isn't 100% actively maintained, and has an [issue template notice indicating its maintainer doesn't have the bandwidth to ramp up on it & actively fix issues](https://github.com/gund/eslint-plugin-deprecation/blob/ae8ca953b1b4cc2021525089f56ede97613e9311/.github/ISSUE_TEMPLATE/issue-template.md#L16-L17)

Proposal: let's write a new implementation of a `deprecation` rule in this repo?

### Fail Cases

```TypeScript
/** @deprecated */
declare function v1(): unknown;

v1();
```

### Pass Cases

```TypeScript
declare function v2(): unknown;

v2();
```

### Additional Info

Note that the licensing here is tricky. https://github.com/gund/eslint-plugin-deprecation includes the following README.md note:

> This rule was originally ported from the [SonarJS repository](https://github.com/SonarSource/SonarJS).

[...]

PR #9783, opened 2024-08-12
WIP: no-deprecation

JoshuaKGoldberg:
## PR Checklist

## Overview

Sending as draft for reference.

💖

### DIFF
--- a/packages/eslint-plugin/src/rules/no-deprecated.ts
+++ b/packages/eslint-plugin/src/rules/no-deprecated.ts
@@ -1,0 +1,214 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+import * as ts from 'typescript';
+
+import { createRule, getParserServices } from '../util';
+
+type IdentifierLike = TSESTree.Identifier | TSESTree.JSXIdentifier;
+
+const functionLikeSymbolKinds = new Set<ts.SyntaxKind | undefined>([
+  ts.SyntaxKind.FunctionDeclaration,
+  ts.SyntaxKind.FunctionExpression,
+  ts.SyntaxKind.MethodDeclaration,
+  ts.SyntaxKind.MethodSignature,
+]);
+
+const importTypes = new Set([
+  AST_NODE_TYPES.ImportDeclaration,
+  AST_NODE_TYPES.ImportExpression,
+]);
+
+const parentDeclarationTypes = new Set([
+  AST_NODE_TYPES.ArrowFunctionExpression,
+  AST_NODE_TYPES.FunctionDeclaration,
+  AST_NODE_TYPES.FunctionExpression,
+  AST_NODE_TYPES.MethodDefinition,
+  AST_NODE_TYPES.TSDeclareFunction,
+  AST_NODE_TYPES.TSEmptyBodyFunctionExpression,
+  AST_NODE_TYPES.TSEnumDeclaration,
+  AST_NODE_TYPES.TSInterfaceDeclaration,
+  AST_NODE_TYPES.TSMethodSignature,
+  AST_NODE_TYPES.TSModuleDeclaration,
+  AST_NODE_TYPES.TSParameterProperty,
+  AST_NODE_TYPES.TSPropertySignature,
+  AST_NODE_TYPES.TSTypeAliasDeclaration,
+  AST_NODE_TYPES.TSTypeParameter,
+]);
+
+export default createRule({
+  name: 'no-deprecated',
+  meta: {
+    docs: {
+      description: 'Disallow using code marked as `@deprecated`',
+      recommended: 'recommended',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      deprecated: `\`{{name}}\` is deprecated.{{reason}}`,
+    },
+    schema: [],
+    type: 'problem',
+  },
+  defaultOptions: [],
+  create(context) {
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    function isDeclaration(node: IdentifierLike): boolean {
+      const { parent } = node;
+
+      switch (parent.type) {
+        case AST_NODE_TYPES.ArrayPattern:
+          return parent.elements.includes(node as TSESTree.Identifier);
+
+        case AST_NODE_TYPES.ClassExpression:
+        case AST_NODE_TYPES.ClassDeclaration:
+        case AST_NODE_TYPES.VariableDeclarator:
+        case AST_NODE_TYPES.TSEnumMember:
+          return parent.id === node;
+
+        case AST_NODE_TYPES.MethodDefinition:
+        case AST_NODE_TYPES.PropertyDefinition:
+          return parent.key === node;
+
+        case AST_NODE_TYPES.Property:
+          return (
+            (parent.shorthand && parent.value === node) ||
+            parent.parent.type === AST_NODE_TYPES.ObjectPattern
+          );
+
+        case AST_NODE_TYPES.AssignmentPattern:
+          return (
+            parent.left === node &&
+            !(
+              parent.parent.type === AST_NODE_TYPES.Property &&
+              parent.parent.shorthand
+            )
+          );
+
+        default:
+          return parentDeclarationTypes.has(parent.type);
+      }
+    }
+
+    function isInsideImport(node: TSESTree.Node): boolean {
+      return context.sourceCode
+        .getAncestors(node)
+        .some(ancestor => importTypes.has(ancestor.type));
+    }
+
+    function isFunctionLikeSymbol(symbol: ts.Symbol): boolean {
+      return functionLikeSymbolKinds.has(symbol.getDeclarations()?.at(0)?.kind);
+    }
+
+    function getJsDocDeprecation(
+      symbol: ts.Symbol | undefined,
+    ): string | undefined {
+      const tag = symbol
+        ?.getJsDocTags(checker)
+        .find(tag => tag.name === 'deprecated');
+
+      if (!tag) {
+        return undefined;
+      }
+
+      const displayParts = tag.text;
+
+      return displayParts ? ts.displayPartsToString(displayParts) : '';
+    }
+
+    function isCallLike(node: TSESTree.Node): boolean {
+      const [callee, parent] =
+        node.parent?.type === AST_NODE_TYPES.MemberExpression &&
+        node.parent.property === node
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/no-deprecated.ts
JoshuaKGoldberg, line 1: Note: @Zamiell also did some work on a rule (https://github.com/typescript-eslint/typescript-eslint/issues/8988#issuecomment-2284452160).

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -0,0 +1,221 @@
...
+      if (
+        node.parent.type === AST_NODE_TYPES.AssignmentPattern ||
+        node.parent.type === AST_NODE_TYPES.Property
+      ) {
+        return services
+          .getTypeAtLocation(node.parent.parent)
+          .getProperty(node.name);
+      }
+
+      // Identifier CallExpression
```
JoshuaKGoldberg, line 167: (little notes from debugging tests, heh)

```suggestion
```

yarn.lock
```diff
@@ -6056,16 +6002,6 @@ __metadata:
...
 
-"@typescript-eslint/visitor-keys@npm:6.21.0":
-  version: 6.21.0
-  resolution: "@typescript-eslint/visitor-keys@npm:6.21.0"
-  dependencies:
-    "@typescript-eslint/types": 6.21.0
-    eslint-visitor-keys: ^3.4.1
-  checksum: 67c7e6003d5af042d8703d11538fca9d76899f0119130b373402819ae43f0bc90d18656aa7add25a24427ccf1a0efd0804157ba83b0d4e145f06107d7d1b7433
-  languageName: node
-  linkType: hard
```
JoshuaKGoldberg, line 6067: Aside: it is very pleasing to no longer have a dependency on v6 versions of our packages. 🙃

eslint.config.mjs
```diff
@@ -31,9 +30,7 @@ export default tseslint.config(
     plugins: {
       ['@typescript-eslint']: tseslint.plugin,
       ['@typescript-eslint/internal']: tseslintInternalPlugin,
-      // https://github.com/gund/eslint-plugin-deprecation/issues/78
       // https://github.com/typescript-eslint/typescript-eslint/issues/8988
```
kirkwaiblinger, line 33: guess this comment can be removed too?

packages/eslint-plugin/docs/rules/no-deprecated.mdx
```diff
@@ -0,0 +1,68 @@
...
+## Examples
+
+<Tabs>
+<TabItem value="❌ Incorrect">
+
+```ts
+/** @deprecated Use @see {@link apiV2} instead. */
+declare function apiV1(): Promise<string>;
+
+declare function apiV1(): Promise<string>;
```
kirkwaiblinger, line 30:
```suggestion
declare function apiV2(): Promise<string>;
```

packages/eslint-plugin/docs/rules/no-deprecated.mdx
```diff
@@ -0,0 +1,68 @@
...
+```
+
+</TabItem>
+<TabItem value="✅ Correct">
+
+```ts
+/** @deprecated Use @see {@link apiV2} instead. */
+declare function apiV1(): Promise<string>;
+
+declare function apiV1(): Promise<string>;
```
kirkwaiblinger, line 49:
```suggestion
declare function apiV2(): Promise<string>;
```

packages/eslint-plugin/docs/rules/no-deprecated.mdx
```diff
@@ -0,0 +1,68 @@
...
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+If portions of your project heavily use deprecated APIs and have no plan for moving to non-deprecated ones, you might want to disable this rule in those portions.
+
+## Related To
+
+- [`eslint-plugin-deprecation`](https://github.com/gund/eslint-plugin-deprecation)
```
kirkwaiblinger, line 68: IMO this is a little confusing without an explanation of why we've made our own rule. As a user, I'd wonder - well, which one should I use?

kirkwaiblinger (changes requested):
Looks great to me, just a few questions!

packages/eslint-plugin/docs/rules/no-deprecated.mdx
```diff
@@ -0,0 +1,69 @@
...
+</Tabs>
+
+## When Not To Use It
+
+If portions of your project heavily use deprecated APIs and have no plan for moving to non-deprecated ones, you might want to disable this rule in those portions.
+
+## Related To
+
+- [`import/no-deprecated`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-deprecated.md) and [`import-x/no-deprecated`](https://github.com/un-ts/eslint-plugin-import-x/blob/master/docs/rules/no-deprecated.md): Does not use type information, but does also support [TomDoc](http://tomdoc.org)
+- [`eslint-plugin-deprecation`](https://github.com/gund/eslint-plugin-deprecation): Predecessor to this rule in a separate plugin
```
JoshuaKGoldberg, line 69: @gund @kirkwaiblinger I'm unsure what phrasing to use here. I don't want to phrase anything negative about eslint-plugin-deprecation, but do want to position it as something that's come earlier & we think we're a next-step from... 🤔

gund: To me the wording sounds fine - it's a predecessor pretty much nothing negative about it imo.
Great to see this rule becoming a part of this repo, thanks for your work on it!
Once this is released I would update my repo to point to this rule here and archive it.

kirkwaiblinger (dismissed):
Awesome! 

Looks like the snapshots need re-updating. Feel free to merge without my re-approval when you do that.

kirkwaiblinger (approved):

Josh-Cena (dismissed):
Maybe mention `deprecation/deprecation` and why we have this rule?

kirkwaiblinger (approved):

packages/eslint-plugin/tests/docs-eslint-output-snapshots/no-deprecated.shot
```diff
@@ -0,0 +1,45 @@
...
+exports[`Validating rule docs no-deprecated.mdx code examples ESLint output 1`] = `
+"Incorrect
+
+/** @deprecated Use @see {@link apiV2} instead. */
+declare function apiV1(): Promise<string>;
+
+declare function apiV2(): Promise<string>;
+
+await apiV1();
+      ~~~~~ \`apiV1\` is deprecated. Use
```
bradzacher, line 12: this doesn't seem right?

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -0,0 +1,243 @@
...
+            current = current.parent;
+        }
+      }
+    }
+
+    function getJsDocDeprecation(
+      symbol: ts.Signature | ts.Symbol | undefined,
+    ): string | undefined {
+      const tag = symbol
+        ?.getJsDocTags(checker)
```
bradzacher, line 115: this forces us to do "jsdoc parsing" by default :(
I kind of wish we could do it without using this -- eg with a regex so we can speed up parsing.

bradzacher (dismissed):

packages/eslint-plugin/tests/docs-eslint-output-snapshots/no-deprecated.shot
JoshuaKGoldberg: Ah, I'm using the tags wrong apparently. `@see` can't be nested inside `@deprecated`. They're two different tags from `symbol.getJsdocTags()`.

Screenshot of hovering over apiV1 from the linked playground, showing @deprecated and @see on a new line each

I'll just remove the extra tags.

packages/eslint-plugin/src/rules/no-deprecated.ts
JoshuaKGoldberg: _OH_, this just reminded me we have `JSDocParsingMode`. Adding `jsDocParsingMode: 'none'` to the test file's `languageOptions.parserOptions` causes all the `invalid` tests to fail for not having any parsed JSDocs. Great.

I think we're going to have to do our own regex-based parsing. I'll put the PR back in draft in the meantime.

bradzacher: I think this is one of the few usecases of jsdoc via TS AST in the ecosystem. So yeah if we can remove this then we can consider flipping the default in the next version to speed up parsing in the general case!

JoshuaKGoldberg: Well, I tried using on `symbol.valueDeclaration` and `symbol.valueDeclaration.parent` in the first `invalid` test case:
* `ts.getJSDocCommentsAndTags`: Nothing retrieved
* `tsutils.forEachComment`: Retrieved the comment, but took ~750ms 

`deprecation/deprecation` also uses `symbol.getJsDocTags` and suffers from the same issue. We don't (yet?) default `jsdocParsingMode` to `'none'` or `'type-info'` for lint users. I added a thrown error for those modes in 5cd8b823d.

Back-linking to past conversation (cc @jakebailey as FYI): https://github.com/typescript-eslint/typescript-eslint/pull/7997#issuecomment-1828891453

bradzacher: Yeah for sure - file an issue.
Wasnt sposed to be a blocking comment. More of a "hey if we can that'd be good" thing

bradzacher (approved):

packages/eslint-plugin/src/rules/no-deprecated.ts
jakebailey: Perhaps it's worth adding `@deprecated` to the list of tags that we parse out unconditionally in the API? Would that be simpler?

jakebailey: I'd have to test and see. Unfortunate though as my enum names were too clever, `@deprecated` is not used for checking :(

FloEdelmann:
The rule was renamed to `no-deprecated`, could you please adjust the PR title and the changelog, to reduce confusion?

JoshuaKGoldberg:
Alas, changing the commit history & linked changelogs would be hard 😞 sorry about that. I edited the PR title just to help folks find it.

PR #9795, opened 2024-08-14
docs: added note on async generator support in require-await

### LINKED ISSUE #9769: Docs: [require-await] document behavior around generator functions

### Suggested Changes

The [base rule](https://eslint.org/docs/latest/rules/require-await) explicitly mentions that it does _not_ flag on generators. 

Our rule intentionally departs from that behavior. However, we don't say so in the docs, so, it's surprising when you encounter a require-await error in a generator function. Let's document this!

See, e.g. https://github.com/typescript-eslint/typescript-eslint/pull/1782 and https://github.com/typescript-eslint/typescript-eslint/pull/8003

### Affected URL(s)

https://typescript-eslint.io/rules/require-await/

xaos7991:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9769

## Overview

Added a note to the [@typescript-eslint/require-await](https://typescript-eslint.io/rules/require-await/) documentation indicating that the rule supports async generator functions. This clarification ensures users are aware that async generators are subject to the same enforcement as other async functions.


packages/eslint-plugin/docs/rules/require-await.mdx
```diff
@@ -12,6 +12,10 @@ import TabItem from '@theme/TabItem';
 This rule extends the base [`eslint/require-await`](https://eslint.org/docs/rules/require-await) rule.
 It uses type information to allow promise-returning functions to be marked as `async` without containing an `await` expression.
 
+:::note
+Unlike the base rule, our implementation also supports [async generator functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*).
```
kirkwaiblinger, line 16: It's not 100% unambiguous what it means to "support" generator functions (and it's tricky to describe in words). Would you mind adding one example to the docs for the generator function case? 🙂

kirkwaiblinger (changes requested):
Hi @xaos7991!

Thanks for sending in your first contribution to typescript-eslint!! Just requesting an example be added for clarity. But this is great! 🙂

packages/eslint-plugin/docs/rules/require-await.mdx
xaos7991: You want to delete the note and add an example instead?

kirkwaiblinger: Nah, definitely need the words too 👍

xaos7991: I added a new example.

packages/eslint-plugin/docs/rules/require-await.mdx
```diff
@@ -21,5 +25,11 @@ async function returnsPromise1() {
   return Promise.resolve(1);
 }
 
+async function* asyncGenerator() {
```
kirkwaiblinger, line 28: Ah - so, the trouble now, is, this added example is correct in the base rule, too. So, we'd want to add an additional _incorrect_ example (in a separate tab/section - see other rules for examples) that _will_ flag.

kirkwaiblinger (changes requested):

packages/eslint-plugin/docs/rules/require-await.mdx
xaos7991: When I added the example I checked that the base rule doesn't support that.
[ESLint playground(https://eslint.org/play/#eyJ0ZXh0IjoiYXN5bmMgZnVuY3Rpb24gc29tZXRoaW5nKCkge1xuICBjb25zdCBudW0gPSBQcm9taXNlLnJlc29sdmUoMSk7XG4gIHJldHVybiAxICsgbnVtO1xufVxuXG5hc3luYyBmdW5jdGlvbiogYXN5bmNHZW5lcmF0b3IoKSB7XG4gIGNvbnN0IG51bSA9IFByb21pc2UucmVzb2x2ZSgxKTtcbiAgeWllbGQgMSArIG51bTtcbn1cbiIsIm9wdGlvbnMiOnsicnVsZXMiOnsicmVxdWlyZS1hd2FpdCI6WyJlcnJvciJdfSwibGFuZ3VhZ2VPcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYUZlYXR1cmVzIjp7fX19fX0=)
[Typescript-eslint playground](https://typescript-eslint.io/play/#ts=5.5.2&fileType=.ts&code=IYZwngdgxgBAZgV2gFwJYHsIxOgtgU2QAtUIBzACgEoYBvAWACgYYpMRkYIFcYBeGAAUATnlQh8AOmH4cAGwBu%2BCgEYqAbiYsZyBMKwqYAai49NjAL5MmoSLEQoMEAFQxb0AOL4I%2BYcGTowtR0WqzsnNy8AiJiEtKy6IrKauYsYKj4cgAmMIYmkeZWjExAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEI0QR1gEtpEBaAQwHdzCAXdKAewDNnIAacbSAAVoE8ADigDG0QoNqkU8QgDtaAehIFiZKjXoZIiaNEbRIXAL4hjQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

kirkwaiblinger: I'm not following, sorry. Could you clarify?

xaos7991: I clarified this in the comment above. Could you check the links and see the difference?

kirkwaiblinger: Oh, looking at this now, I think I looked only in the "correct" tab, here, and failed to see the corresponding code in the "incorrect" tab. In any case this looks good. Sorry for the confusion!

packages/eslint-plugin/docs/rules/require-await.mdx
```diff
@@ -12,14 +12,45 @@ import TabItem from '@theme/TabItem';
...
+<TabItem value="❌ Incorrect">
+
+```ts
+async function returnNumber() {
+  return 1;
+}
+
+async function* asyncGenerator() {
+  const num = Promise.resolve(1);
+  yield 1 + num;
```
kirkwaiblinger, line 31: One more complaint (sorry!) - we try to avoid putting type errors in our examples that might be seen as red herrings. Here, TS will complain that `number` and `Promise<number>` can't be added to each other, which isn't relevant to what this rule is about. Let's fix that up so we don't confuse readers 🙂 

i.e. something like this?
```ts
// incorrect
async function* asyncGenerator() {
   yield 1;
}

// correct
function* asyncGenerator() {
   yield 1;
}
```

kirkwaiblinger (changes requested):

packages/eslint-plugin/docs/rules/require-await.mdx
```diff
@@ -12,14 +12,45 @@ import TabItem from '@theme/TabItem';
 This rule extends the base [`eslint/require-await`](https://eslint.org/docs/rules/require-await) rule.
 It uses type information to allow promise-returning functions to be marked as `async` without containing an `await` expression.
 
+:::note
+Unlike the base rule, our implementation also supports [async generator functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*).
```
Josh-Cena, line 16:
```suggestion
`yield` expressions in [async generator functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*) behave differently from sync generator functions, so the base rule never checks async generator functions. On the other hand, our rule uses type information and can detect async generator functions that also always yield non-promise values.
```

kirkwaiblinger (approved):
This looks good to me. Thanks for taking care of this! 
congratulatory gif

Josh-Cena (approved):

PR #9798, opened 2024-08-15
chore: enable unicorn/prefer-string-replace-all

abrahamguo:
## PR Checklist

## Overview

Enable `unicorn/prefer-string-replace-all` ([docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-replace-all.md)). 100% autofixable.

### DIFF
--- a/packages/ast-spec/tests/fixtures.test.ts
+++ b/packages/ast-spec/tests/fixtures.test.ts
@@ -77,7 +77,7 @@ const FIXTURES: readonly Fixture[] = [...VALID_FIXTURES, ...ERROR_FIXTURES].map(
       isError: /[\\/]_error_[\\/]/.test(absolute),
       isJSX: ext.endsWith('x'),
       name,
-      relative: path.relative(SRC_DIR, absolute).replace(/\\/g, '/'),
+      relative: path.relative(SRC_DIR, absolute).replaceAll('\\', '/'),
       segments,
       snapshotFiles: {
         success: {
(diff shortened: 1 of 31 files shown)


packages/ast-spec/tests/fixtures.test.ts
```diff
@@ -77,7 +77,7 @@ const FIXTURES: readonly Fixture[] = [...VALID_FIXTURES, ...ERROR_FIXTURES].map(
       isError: /[\\/]_error_[\\/]/.test(absolute),
       isJSX: ext.endsWith('x'),
       name,
-      relative: path.relative(SRC_DIR, absolute).replace(/\\/g, '/'),
+      relative: path.relative(SRC_DIR, absolute).replaceAll('\\', '/'),
```
JoshuaKGoldberg, line 80: 🤔 this new version is more characters and IMO equally clear. I'm -0.1.

abrahamguo: In this specific instance, or in the whole PR?

Josh-Cena: I'm +1. Regex escaping is too footguny. We should prioritize string replacement whenever possible.

JoshuaKGoldberg: I meant kind of halfway between this instance & the whole PR. This instance and the ones like it...

...but yeah, Josh brings up a good point on the footguns. I'm ok with the average being +0.9.

kirkwaiblinger (approved):

PR #9820, opened 2024-08-19
docs: add note about typescript-eslint as peer dependency

### LINKED ISSUE #9814: Docs: Issue with Flat Config and typescript-eslint Parser in Custom ESLint Plugin

### Suggested Changes

I am currently working on adding flat config support to my custom ESLint plugin, which is designed to work with TypeScript. 

**Problem Description**

I've followed the docs and added @typescript-eslint/parser as a peer dependency. However, with the flat-config, I need to provide the parser from typescript-eslint, which is not a peer dependency. This leads to an issue where users are reporting compilation errors due to the missing typescript-eslint dependency.

**Reference Example**

I referenced the eslint-plugin-example-typed-linting repository for guidance and noticed that it does not define a parser explicitly in its configuration. This allows it to avoid the issue altogether, likely because the ESLint config pulls the parser from another plugin that has it as a dependency.
What to do?

1. Add typescript-eslint as a direct dependency in my plugin to ensure its availability and prevent compilation errors?
2. Skip defining the parser in my configuration, assuming the user’s ESLint setup will provide the necessary parser?

---

I think that should be discussed in the docs.

### Affected URL(s)

https://typescript-eslint.io/developers/eslint-plugins

rainerhahnekamp:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9814

## Overview

Explain, that there is not need to define the parser and therefore include `typescript-eslint` as peer dependency.

### DIFF
--- a/docs/developers/ESLint_Plugins.mdx
+++ b/docs/developers/ESLint_Plugins.mdx
@@ -42,6 +42,10 @@ Include the following to enforce the version range allowed without making users'
 
 Those are all packages consumers are expected to be using already.
 
+The parser for TypeScript files generally requires `typescript-eslint`.
+However, you do not need to add the parser, and therefore `typescript-eslint`, as a peer dependency.
+Our setup guide ensures that the parser is automatically registered when configuring ESLint.
+
 ## `RuleCreator` Usage
 
 We recommend including at least the following three properties in your plugin's [`RuleCreator` extra rule docs types](./Custom_Rules.mdx#extra-rule-docs-types):
--- a/docs/developers/ESLint_Plugins.mdx
+++ b/docs/developers/ESLint_Plugins.mdx
@@ -42,8 +42,7 @@ Include the following to enforce the version range allowed without making users'
 
 Those are all packages consumers are expected to be using already.
 
-The parser for TypeScript files generally requires `typescript-eslint`.
-However, you do not need to add the parser, and therefore `typescript-eslint`, as a peer dependency.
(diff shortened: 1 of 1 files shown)


docs/developers/ESLint_Plugins.mdx
```diff
@@ -42,6 +42,10 @@ Include the following to enforce the version range allowed without making users'
 
 Those are all packages consumers are expected to be using already.
 
+The parser for TypeScript files generally requires `typescript-eslint`.
+However, you do not need to add the parser, and therefore `typescript-eslint`, as a peer dependency.
+Our setup guide ensures that the parser is automatically registered when configuring ESLint.
```
JoshuaKGoldberg, line 47: > The parser for TypeScript files generally requires `typescript-eslint`.

No, it's the other way around. There's no dependency on the `typescript-eslint` package in the parser.

> you do not need to add the parser ... as a peer dependency

This goes against what's stated earlier in this section - that `@typescript-eslint/parser` should be a peer dependency.

---

How about...

```suggestion
You don't need to declare any dependencies on `typescript-eslint` or `@typescript-eslint/eslint-plugin`.
Our setup guide ensures that the parser is automatically registered when configuring ESLint.
```

JoshuaKGoldberg (changes requested):
Now I'm not so sure what would actually be useful in the docs 🤔

JoshuaKGoldberg (approved):
I'm going to go ahead and make an executive decision saying this is sufficient. We really only need the `@typescript-eslint/parser` as a `peerDependency`; everything else not mentioned in the doc is userland config.

Thanks for sending this in @rainerhahnekamp!

PR #9834, opened 2024-08-20
chore: enable unicorn/prefer-spread

abrahamguo:
## PR Checklist

## Overview

Enable `unicorn/prefer-spread` ([docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-spread.md)).

I particularly like the replacements of the style `a.concat(b)` => `[...a, ...b]` because I like that `a` and `b` are given "equal weight" visually when they are both spread into a new array, whereas with `concat` one is before the `concat` and one is inside the `concat` and so visually they don't have "equal weight".


kirkwaiblinger:
I agree with all these changes except the ones that use `[...expr]` instead of `Array.from(expr)`

abrahamguo:
My thoughts are — it's better to use a language construct (`...`) rather than a function (`Array.from`) when available. It's shorter as well 🤷‍♂️

kirkwaiblinger (dismissed):
Ok, works for me 🙂

JoshuaKGoldberg (approved):
I also like this :)

PR #9843, opened 2024-08-20
chore: enable eslint-plugin-perfectionist on eslint-plugin package, with ordering tweaks

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: continues #8479

## Overview

Enables sorting on `packages/eslint-plugin`.

This by nature ~includes a couple changes to rule ordering preferences~ edit: after review feedback, this now sets the config to explicitly override alphabetrization:

* Putting `create` after metadata such as `meta` in rule definitions
* Putting `start` before `end` in rule reports and test cases
* Putting `type` before other properties in rule `meta` and schema objects
* Putting `valid` entries before `invalid` ones in rule tests

💖


packages/eslint-plugin/src/rules/adjacent-overload-signatures.ts
```diff
@@ -21,24 +22,11 @@ type MemberDeclaration =
...
-      description: 'Require that function overload signatures be consecutive',
-      recommended: 'stylistic',
-    },
-    schema: [],
-    messages: {
-      adjacentSignature: 'All {{name}} signatures should be adjacent.',
-    },
-  },
-  defaultOptions: [],
   create(context) {
```
bradzacher, line 25: This might be a case where we want to ditch the alpha sorting.
I think that plugin defs are better sorted with the meta-properties being first and `create` being last.

packages/eslint-plugin/tests/rules/adjacent-overload-signatures.test.ts
```diff
@@ -5,924 +5,924 @@ import rule from '../../src/rules/adjacent-overload-signatures';
 const ruleTester = new RuleTester();
 
 ruleTester.run('adjacent-overload-signatures', rule, {
-  valid: [
+  invalid: [
```
bradzacher, line 8: yeah I definitely don't like this.
I think that the standard of valid, invalid is better.
I'm fine with everything else being sorted -- but I think this top-level would be best left as is.

bradzacher (changes requested):
my 2c: big no no here for rule definitions and test definitions.

kirkwaiblinger (changes requested):
Fully agree with @bradzacher here.

Also, even if I agreed the sorted version were preferable, I would say this is a _huge_ disruption to the git history for marginal gain at best, and therefore to pass on, granted the current the existing codebae state.

bradzacher:
There's a sorting rule in `eslint-plugin-eslint-plugin` right? That one maintains a sort order closer to what we do IIRC

packages/eslint-plugin/src/rules/ban-ts-comment.ts
```diff
@@ -197,11 +198,10 @@ export default createRule<[Options], MessageIds>({
             if (directive === 'ignore') {
               // Special case to suggest @ts-expect-error instead of @ts-ignore
               context.report({
-                node: comment,
                 messageId: 'tsIgnoreInsteadOfExpectError',
+                node: comment,
                 suggest: [
                   {
-                    messageId: 'replaceTsIgnoreWithTsExpectError',
```
kirkwaiblinger, line 204: i also think putting the messageId first and the fix afterwards is beneficial - can that be added to the sorting rule?

packages/eslint-plugin/src/rules/ban-ts-comment.ts
```diff
@@ -214,14 +214,15 @@ export default createRule<[Options], MessageIds>({
...
+                    messageId: 'replaceTsIgnoreWithTsExpectError',
                   },
                 ],
               });
             } else {
               context.report({
                 data: { directive },
-                node: comment,
                 messageId: 'tsDirectiveComment',
+                node: comment,
```
kirkwaiblinger, line 225: ideally `node`/`loc` would be first in the `context.report` as it's the most important piece of it...
IMO my ordering would be
1. `loc`
2. `node`
3. `messageId`
4. `data`
5. `fix`

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -139,8 +140,8 @@ export default createRule<Options, MessageIds>({
     ) {
       context.report({
         loc: {
-          start: { line: 0, column: 0 },
-          end: { line: 0, column: 0 },
+          end: { column: 0, line: 0 },
```
kirkwaiblinger, line 143: oof. 
you know how i'll feel about this 😆

packages/eslint-plugin/src/rules/naming-convention-utils/enums.ts
```diff
@@ -125,11 +125,11 @@ enum Modifiers {
...
 enum TypeModifiers {
+  array = 1 << 21,
   boolean = 1 << 17,
-  string = 1 << 18,
-  number = 1 << 19,
   function = 1 << 20,
-  array = 1 << 21,
+  number = 1 << 19,
+  string = 1 << 18,
 }
```
JoshuaKGoldberg, line 133: Ah, enums...

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
kirkwaiblinger: Oh, actually - I don't really care about the line vs column thing, but it's the `{ end, start }` that grinds my gears

JoshuaKGoldberg: what is time...

eslint.config.mjs
```diff
@@ -578,33 +578,58 @@ export default tseslint.config(
...
     extends: [perfectionistPlugin.configs['recommended-alphabetical']],
-    files: ['packages/utils/src/**/*.ts'],
+    ignores: ['packages/eslint-plugin/src/configs/*'],
+    files: [
+      'packages/eslint-plugin/{src,tests,tools,typings}/**/*.ts',
+      'packages/utils/src/**/*.ts',
+    ],
     rules: {
-      'perfectionist/sort-classes': [
+      '@typescript-eslint/sort-type-constituents': 'off',
```
OlivierZal, line 587: Perfectionist's documentation says it can conflict with `sort-intersection-types`, but I never encountered any conflict issue myself.

JoshuaKGoldberg: Yeah it's mostly around small nuances in sorting, or when you have custom sorts.

kirkwaiblinger (dismissed):
ok, i guess i have been picky enough 😆 Thanks for indulging me

eslint.config.mjs
```diff
@@ -611,32 +612,67 @@ export default tseslint.config(
...
         'error',
         {
-          order: 'asc',
-          partitionByComment: true,
-          type: 'natural',
+          customGroups: {
+            primary: ['loc', 'name', 'node', 'type'],
+            secondary: ['meta', 'messageId', 'start'],
+            tertiary: ['defaultOptions', 'data', 'end'],
+            quaternary: ['fix'],
```
bradzacher, line 659: You know.... You could have just called them "first" "second" "third" and "fourth" haha

packages/ast-spec/src/declaration/TSModuleDeclaration/spec.ts
```diff
@@ -9,11 +9,13 @@ import type { Literal } from '../../unions/Literal';
 export type TSModuleDeclarationKind = 'global' | 'module' | 'namespace';
 
 interface TSModuleDeclarationBase extends BaseNode {
+  type: AST_NODE_TYPES.TSModuleDeclaration;
   /**
    * The body of the module.
    * This can only be `undefined` for the code `declare module 'mod';`
    */
   body?: TSModuleBlock;
+  // TODO - remove this in the next major (we have `.kind` now)
```
bradzacher, line 18: This shouldn't have moved - it applies to the `global` property.

packages/eslint-plugin/src/configs/all.ts
bradzacher, line 1: Probs undo this? It's nice to have them next to the extension rules that are replacing them.

packages/eslint-plugin/src/configs/recommended-type-checked-only.ts
bradzacher, line 1: Ditto - probs undo this

packages/eslint-plugin/src/configs/recommended-type-checked.ts
bradzacher, line 1: Ditto probs undo this

packages/eslint-plugin/src/configs/recommended.ts
bradzacher, line 1: Ditto probs undo this

packages/eslint-plugin/src/configs/strict-type-checked-only.ts
bradzacher, line 1: Ditto probs undo this

packages/eslint-plugin/src/configs/strict-type-checked.ts
bradzacher, line 1: Ditto probs undo this

packages/eslint-plugin/src/configs/strict.ts
bradzacher, line 1: Ditto probs undo this

packages/eslint-plugin/src/configs/stylistic-type-checked-only.ts
bradzacher, line 1: Ditto probs undo this

packages/eslint-plugin/src/configs/stylistic-type-checked.ts
bradzacher, line 1: Ditto probs undo this

packages/eslint-plugin/src/configs/stylistic.ts
bradzacher, line 1: Ditto probs undo this

bradzacher (approved):

eslint.config.mjs
JoshuaKGoldberg: lol I liked these

PR #9846, opened 2024-08-20
chore: enable eslint-plugin-perfectionist on rule-schema-to-typescript-types package

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: continues #8479.

## Overview

Enables sorting on `packages/rule-schema-to-typescript-types`.

💖


packages/rule-schema-to-typescript-types/src/generateArrayType.ts
```diff
@@ -1,13 +1,15 @@
-import { TSUtils } from '@typescript-eslint/utils';
 import type {
   JSONSchema4,
   JSONSchema4ArraySchema,
 } from '@typescript-eslint/utils/json-schema';
 
+import { TSUtils } from '@typescript-eslint/utils';
+
+import type { ArrayAST, AST, RefMap, TupleAST, UnionAST } from './types';
+
```
bradzacher, line 9: 2c: I dislike this style of splitting `type`-only imports from value imports.
I personally didn't really like the grouping at all, but `simple-import-sort` used to be the "best autofixable" rule.

Question: Can we turn import grouping off entirely?
Given all relative imports are prefixed with `./` or `../` -- the relative imports are already "grouped".

If not, w/e -- it's autofixed.

bradzacher (approved):

packages/rule-schema-to-typescript-types/src/generateArrayType.ts
JoshuaKGoldberg: Yeah, we should be able to. I commented https://github.com/typescript-eslint/typescript-eslint/issues/8479#issuecomment-2307367562 to make sure it's done as part of #8479.

PR #9847, opened 2024-08-20
chore: enable eslint-plugin-perfectionist on rule-tester package

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: continues #8479

## Overview

Enables sorting on `packages/rule-tester`.

💖


eslint.config.mjs
```diff
@@ -578,7 +578,10 @@ export default tseslint.config(
   },
   {
     extends: [perfectionistPlugin.configs['recommended-alphabetical']],
-    files: ['packages/utils/src/**/*.ts'],
+    files: [
+      'packages/rule-tester/{src,tests,typing}/**/*.ts',
```
JoshuaKGoldberg, line 582:
```suggestion
      'packages/rule-tester/{src,tests,typings}/**/*.ts',
```

packages/rule-tester/src/types/DependencyConstraint.ts
```diff
@@ -6,12 +6,12 @@ export interface RangeOptions {
...
 
 export interface SemverVersionConstraint {
-  readonly range: string;
   readonly options?: RangeOptions | boolean;
+  readonly range: string;
 }
 export type AtLeastVersionConstraint =
-  | `${number}.${number}.${number}-${string}`
   | `${number}.${number}.${number}`
+  | `${number}.${number}.${number}-${string}`
```
bradzacher, line 14: weird that this sorted this way.

Why does `-` come AFTER "nothing", but `.` comes BEFORE nothing.
Is this a bug? What the what?

packages/rule-tester/src/utils/config-validator.ts
```diff
@@ -25,8 +26,8 @@ const ruleValidators = new WeakMap<AnyRuleModule, ValidateFunction>();
 let validateSchema: ValidateFunction | undefined;
 const severityMap = {
   error: 2,
-  warn: 1,
   off: 0,
+  warn: 1,
```
bradzacher, line 30: this is another case where the value ordering is better than the prop name ordering

packages/rule-tester/src/utils/flat-config-schema.ts
```diff
@@ -10,17 +10,17 @@ import { normalizeSeverityToNumber } from './severity';
...
 
 const ruleSeverities = new Map<SharedConfig.RuleLevel, SharedConfig.Severity>([
-  [0, 0],
+  ['error', 2],
   ['off', 0],
-  [1, 1],
   ['warn', 1],
+  [0, 0],
+  [1, 1],
   [2, 2],
```
bradzacher, line 23: this is another case where the value ordering is better than the prop name ordering

bradzacher (dismissed):

packages/rule-tester/src/types/DependencyConstraint.ts
JoshuaKGoldberg: I think it's that <code>`</code> is being treated as before `-` alphabetically?

bradzacher: That must be it - it must be considering the quote character as part of the thing.

Would be worth filing a bug for that.
"if the two types are the same literal type (eg both template literal types) then the quote character should be excluded from the sort comparison.

JoshuaKGoldberg: 👍 noted in https://github.com/typescript-eslint/typescript-eslint/issues/8479#issuecomment-2307391859

Issue #9731, opened 2024-08-06
Bug: `projectService` causes issues with `vscode-eslint` extension.


tetarchus:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Abug) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Issue Description

We've been using the v8 alpha-30 release to make use of ESLint v9 and have now switched to using the full release.
When using the `projectService` option for `typescript-eslint`'s parser in v8.0.0, creating a new file causes the extension to produce the error:

```
Parsing error: [path-to-newly-created-file] was not found by the project service. Consider either including it in the tsconfig.json or including it in allowDefaultProject.
```

Requiring that the extension be restarted to force the project service to pick up the new file. 
The alpha version (30) of v8 previously raised `This rule requires the 'strictNullChecks' compiler option to be turned on to function correctly.` on the newly created file, likely caused by the same issue (the project service was not aware of the new file as from the command line it will be started _after_ the file is created).

## Background
- The repo is an NX monorepo using a custom flat config.
- Each project contains a `tsconfig.json` that extends a base config, and `references` several additional tsconfig files (e.g `tsconfig.lib.json`/`tsconfig.spec.json`).
- Linting is run from the root of the workspace on all projects
- There is also a root level `tsconfig.json` (as well as the base config) that includes all files outside of nested projects (which theoretically negates the need for a `defaultProject` if I understand correctly).
- The ESLint configuration uses a function to return the config at runtime based on environment variables and dynamic values - this all works as expected when editing existing files.
- Running ESLint from the command line works as expected.

This [issue](https://github.com/microsoft/vscode-eslint/issues/1911) was also opened on the vscode-eslint repo.

CC @dbaeumer

### Reproduction Repository Link

https://github.com/tetarchus/project-service

### Repro Steps

- Clone the repo
- Open the repo in VSCode
- Install dependencies
- Run `npm run lint` in the terminal and observe no errors
- Ensure that the [vscode extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) is installed and active
[...]

PR #9853, opened 2024-08-20
feat(typescript-estree): reload project service once when file config isn't found

### LINKED ISSUE #9772: Enhancement: Trigger a project service reload on first out-of-project client file

### Relevant Package

typescript-estree

### Description

Coming over from #9731 -> https://github.com/microsoft/vscode-eslint/issues/1911: when `parserOptions.projectService` is enabled and a file isn't found in the project service, we throw an error like:

> ```plaintext
> ${parseSettings.filePath} was not found by the project service. Consider either including it in the tsconfig.json or including it in allowDefaultProject.
> ```

That error triggers in the case of a user adding a file in an editor and not reloading the editor's ESLint extension. The extension has no way of knowing that its project service needs to pick up the new file. https://github.com/microsoft/vscode-eslint/issues/1911 tracks discussing adding in an API to notify and/or be notified by the extension. However:
* Any API provided by an editor extension would only be useful for that specific editor.
* Any API we provide incurs extra work for editor extensions

I think it'd still be useful to solve the _"not found by the project service"_ issue for the general case. Proposal: if the file would throw the error, we should try reloading project service projects with a throttle.

I.e. the first suggestion from https://github.com/microsoft/vscode-eslint/issues/1911#issuecomment-2273674560:

```ts
// openClientFileFromProjectService
let opened = openClientFile();

if (
  !opened.configFileName &&
  !isDefaultProjectAllowed &&
  performance.now() - serviceSettings.lastRefoundFileTimestamp > 250
) {
  serviceSettings.service.reloadProjects();
  opened = openClientFile();
  if (opened.configFileName) {
    serviceSettings.lastRefoundFileTimestamp = performance.now();
  }
}
```

This wouldn't fix for all cases of out-of-date file information (https://github.com/microsoft/vscode-eslint/issues/1774). We'd still want to work on receiving file watchers from editors. But this would help in the general case for `parserOptions.projectService` and file modifications on disk, I think.

### Additional Info

💖

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9772

## Overview

If a file isn't found by `parserOptions.projectService`, then this will attempt to reload the project service. Doing so refreshes how it sees TSConfigs on disk. This should fix most -but not all- cases mentioned in #9731 -> https://github.com/microsoft/vscode-eslint/issues/1911 of a new file being created & ESLint not knowing to update parser information.

This includes several limits on it to avoid slowing down unrelated situations:

* It won't happen in single-run mode at all (i.e. CLI calls without `--fix`)
* It's throttled to 250ms so it shouldn't slow down editors if multiple files have this issue

💖

### DIFF
--- a/packages/typescript-estree/src/create-program/createProjectService.ts
+++ b/packages/typescript-estree/src/create-program/createProjectService.ts
@@ -32,6 +32,7 @@ export type TypeScriptProjectService = ts.server.ProjectService;
 
 export interface ProjectServiceSettings {
   allowDefaultProject: string[] | undefined;
+  lastReloadTimestamp: number;
   maximumDefaultProjectFileMatchCount: number;
   service: TypeScriptProjectService;
 }
@@ -148,6 +149,7 @@ export function createProjectService(
 
   return {
     allowDefaultProject: options.allowDefaultProject,
+    lastReloadTimestamp: performance.now(),
     maximumDefaultProjectFileMatchCount:
       options.maximumDefaultProjectFileMatchCount_THIS_WILL_SLOW_DOWN_LINTING ??
       DEFAULT_PROJECT_MATCHED_FILES_THRESHOLD,
--- a/packages/typescript-estree/src/useProgramFromProjectService.ts
+++ b/packages/typescript-estree/src/useProgramFromProjectService.ts
@@ -16,6 +16,8 @@ import type {
 import { DEFAULT_PROJECT_FILES_ERROR_EXPLANATION } from './create-program/validateDefaultProjectForFilesGlob';
 import type { MutableParseSettings } from './parseSettings';
 
+const RELOAD_THROTTLE_MS = 250;
+
 const log = debug(
   'typescript-eslint:typescript-estree:useProgramFromProjectService',
 );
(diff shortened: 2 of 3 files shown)


jakebailey:
How is tsserver able to handle this gracefully? All the editor is doing is opening files, same as ts-eslint would...

bradzacher (approved):
is there a way to fetch the globs for the project service that it "should" be matching?

it would be much cheaper for us to go like
"should this file be in the project? if yes -> reload, if no -> error"

If not then 👍

bradzacher:
> How is tsserver able to handle this gracefully?

@jakebailey Probably because it has disk watchers too so it can be updated based on disk changes. We don't have any disk watchers :(

JoshuaKGoldberg:
> is there a way to fetch the globs for the project service that it "should" be matching?

Without reloading the projects? Not that I know of. @jakebailey might?

jakebailey:
I'm not 100% certain what you mean; either you open a file and the ProjectService does a bunch of magic to find which project to use, or it fails and it goes into the inferred bucket. Either way, the file is "valid" to use from its eye.

kirkwaiblinger (approved):

PR #9856, opened 2024-08-21
chore: correct a few comment placements in utils

JoshuaKGoldberg:
## Overview

Fixes up a few comments that were put in an incorrect new location in #9698. Thanks @FloEdelmann for catching them!

💖 

Co-authored-by: @FloEdelmann

### DIFF
--- a/packages/utils/src/ts-eslint/Config.ts
+++ b/packages/utils/src/ts-eslint/Config.ts
@@ -18,13 +18,13 @@ export namespace SharedConfig {
 
   export type GlobalVariableOptionBase =
     | 'off'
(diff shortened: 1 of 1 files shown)


kirkwaiblinger (approved):

bradzacher (approved):

bradzacher:
We should report these bugs upstream.

The JSDoc comments need to be attached properly during fixing so that they don't introduce breakages!

JoshuaKGoldberg:
cc @azat-io - I added this to my personal todo to make isolated repros for, not sure when I'll be able to.

PR #9878, opened 2024-08-23
chore: replace globby with tinyglobby

benmccann:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/9453

## Overview

Replaces `globby` with `tinyglobby` for a smaller more efficient implementation with fewer dependencies

### DIFF
--- a/packages/typescript-estree/tests/lib/parse.test.ts
+++ b/packages/typescript-estree/tests/lib/parse.test.ts
@@ -2,7 +2,7 @@ import { join, resolve } from 'node:path';
 
 import type { CacheDurationSeconds } from '@typescript-eslint/types';
 import debug from 'debug';
-import * as globbyModule from 'globby';
+import * as globbyModule from 'tinyglobby';
 import type * as typescriptModule from 'typescript';
 
 import * as parser from '../../src';
@@ -39,11 +39,11 @@ jest.mock('typescript', () => {
   };
 });
 
-jest.mock('globby', () => {
-  const globby = jest.requireActual<typeof globbyModule>('globby');
+jest.mock('tinyglobby', () => {
+  const globby = jest.requireActual<typeof globbyModule>('tinyglobby');
   return {
     ...globby,
-    sync: jest.fn(globby.sync),
+    globSync: jest.fn(globby.globSync),
   };
 });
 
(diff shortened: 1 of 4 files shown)


packages/typescript-estree/tests/lib/parse.test.ts
```diff
@@ -43,7 +43,7 @@ jest.mock('globby', () => {
   const globby = jest.requireActual<typeof globbyModule>('globby');
```
bradzacher, line 43: This would also need to change

bradzacher (changes requested):

packages/typescript-estree/tests/lib/parse.test.ts
benmccann: done. thanks!

Josh-Cena:
Why do we have both this and https://github.com/typescript-eslint/typescript-eslint/pull/9518?

benmccann:
Oh, thanks for pointing that out. I hadn't seen it. I think `tinyglobby` is better as it has many fewer dependencies, but that PR is passing the CI whereas this one isn't. I'll check if the author is able to switch to `tinyglobby` in the other PR and if so perhaps we can close this one

SukkaW:
It seems that your CI failed due to incorrect spying of glob functions. Might wanna check that out.

benmccann:
> I just want to point out that this change added 2 new dependencies to the project and did not actually remove any included dep, if you check yarn.lock. So this feels kinda pointless change this time.
(note, I’m just a user of this plugin, so my opinions do not reflect what the maintainers may think)

The `yarn.lock` isn't what's important here because it includes both dependencies and `devDependencies`. `globby` remains only for developers of this project. If you look at where it's used now it's in things like docusaurus. This PR would remove about 40% of dependencies for all consumers of `typescript-eslint` going from 44 down to 26

benmccann:
> It seems that your CI failed due to incorrect spying of glob functions. Might wanna check that out.

I found the line I missed updating before. It's now correctly mocking. However, it's returning things in a different order. I think we basically have to combine our two PRs to get it working

SukkaW:
> However, it's returning things in a different order.

See https://github.com/typescript-eslint/typescript-eslint/pull/9518#issuecomment-2308461397

bradzacher:
As per https://github.com/typescript-eslint/typescript-eslint/issues/9453#issuecomment-2309080805 we're going to go with #9518 for now.
We may revisit this in future and switch from `fast-glob` to `tinyglobby`.

PR #9886, opened 2024-08-27
chore: enable `unicorn/prefer-ternary`

abrahamguo:
## PR Checklist

## Overview

Enable `unicorn/prefer-ternary` ([docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-ternary.md)).
I believe many of these cases are much simpler than they were previously. Several `let`s could be changed to `const`s — always a good thing.
I recommend viewing the diff with whitespace hidden 😉


Josh-Cena:
I thought from the issue that pretty much everyone were negative/ambivalent on this rule? I personally wouldn't want it enforced though I may enjoy its style.

abrahamguo:
Sure thing! I wanted to open a PR since I felt it might help to see the actual changes recommended and gather a bit more feedback as to whether it would be beneficial to enforce. I'm also open to removing the lint rule and keeping the changes, a la #9640

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
kirkwaiblinger, line 1: I'm kind of surprised this doesn't check this code:
https://github.com/typescript-eslint/typescript-eslint/blob/243fb5f5be07f2124536aeed8b1ffc1e47f895c1/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts#L457-L463
which could just be converted to 
```ts
        const typeArgumentsText =
          node.typeArguments == null
            ? ''
            : sourceCode.getText(node.typeArguments);
```

fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1221

packages/eslint-plugin/src/rules/consistent-type-exports.ts
```diff
@@ -193,46 +193,20 @@ export default createRule<Options, MessageIds>({
             const allExportNames = report.typeBasedSpecifiers.map(
               specifier => specifier.local.name,
             );
-
-            if (allExportNames.length === 1) {
-              const exportNames = allExportNames[0];
```
bradzacher, line 198: This line was lost in the transition.

bradzacher (commented):
My 2c - all these changes are less clear code. I don't think compacting the code down improves anything and I think instead you lose out. So I'm a -1 for this rule.

packages/eslint-plugin/src/rules/consistent-type-exports.ts
abrahamguo: this was actually not needed, because `formatWordList` is able to handle a `1`-length array

bradzacher:
@typescript-eslint/triage-team - do we have the consensus to reject this rule? Based on thumbs up I think we do but just confirming before I close.

Josh-Cena:
+1 to reject

PR #9896, opened 2024-08-28
chore: use a unit test to require rule options schema properties include a description

### LINKED ISSUE #9866: Repo: Use a lint rule to require rule options schema properties include a description
### Suggestion

Following #5392: it'd be good to require that rule options include a `description` property. #5392 and https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/334 have good explanations of why.

We could enforce this with the lint rule described in https://github.com/eslint-community/eslint-plugin-eslint-plugin/issues/334... except, that doesn't exist yet. Which makes this issue blocked on it.

:sparkling

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9866 

## Overview

I did this via a unit test rather than a lint rule.

### DIFF
--- a/packages/eslint-plugin/tests/docs.test.ts
+++ b/packages/eslint-plugin/tests/docs.test.ts
@@ -18,6 +18,7 @@ import type * as UnistUtilVisit from 'unist-util-visit' with { 'resolution-mode'
 import rules from '../src/rules';
 import { areOptionsValid } from './areOptionsValid';
 import { getFixturesRootDir } from './RuleTester';
+import type { JSONSchema4 } from '@typescript-eslint/utils/json-schema';
 
 const docsRoot = path.resolve(__dirname, '../docs/rules');
 const rulesData = Object.entries(rules);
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/tests/docs.test.ts
```diff
@@ -575,6 +576,29 @@ describe('Validating rule metadata', () => {
           rule.meta.docs?.requiresTypeChecking ?? false,
         );
       });
+
+      it('all properties of the rule options schema must include a description', () => {
```
JoshuaKGoldberg, line 580: 😬 I have to be honest, I really don't think a unit test is the right approach here. The issue asks for a lint rule because that's something every ESLint plugin project can use. An internal Jest unit test isn't shareable in that way.

JoshuaKGoldberg (changes requested):
Thanks, but this isn't a direction I think we should take here.

packages/eslint-plugin/tests/docs.test.ts
abrahamguo: Sounds good! I figured

- this could be something to help us move forward without waiting for an upstream update
- as well as potentially better accommodating any potential cases where the options schema is somehow dynamically generated and cannot be statically analyzed by a lint rule

but no worries if you'd prefer not to go this way!

JoshuaKGoldberg:
Now that https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/490 is merged, we should be able to get away with enabling it in our linter. I'll close this out now. Cheers!

PR #9912, opened 2024-09-03
feat(eslint-plugin): [no-unnecessary-condition] check switch cases

### LINKED ISSUE #9802: Enhancement: [no-unnecessary-condition] check nullish switch cases

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.2&fileType=.tsx&code=MYewdgzgLgBAhgLhgcgILJgHxQIQwXhgCkBlAeQDkA6ABzgCcIBTACmSiemQEoAoXiAHcAllGAALGCzjcYAb14wYwOMxToEipcvAQQAGyZV9IAOZtxwnlqUAjekzgBrANxaVa5Hk3adkA0Ym5si2AJ5M1r72jq7uqkwwYACu%2BvpIAPTpMBDiICkAJjBM9PQg9DZ%2BeobGZmyC4nBQkdrRzrwAvvyZMMIAZlJwMPjD6jzyvN1KoP7VQRZWfN3tRfpqfQNDI8mpsnIw3cWljDA0pTTF%2BqETWVO6ATXBYRGLWe1AA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK%2BSpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
const a: 'A' | 'B' = JSON.parse('test')

switch (a) {
  case 'A':
    console.log('hi')
    break;
  case 'B':
    console.log('bye')
    break;
  case null: // should error
    console.log('what')
    break
}

// if (a === 'A') {
//   console.log('hi')
// } else if (a === null) { // errors properly
//   console.log('bye')
// }
```

### ESLint Config

```javascript
{
  "rules": {
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

The null switch case should have errored

### Actual Result

No error

### Additional Info

Works fine (real error) with an if-else

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9802 

## Overview

This PR is the satisfying pinnacle of "big feature, tiny code change" 🌈


abrahamguo:
Ready for review!

abrahamguo:
Ready for review! CI failed due to lint flake.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -364,13 +364,15 @@ export default createRule<Options, MessageId>({
...
     function checkIfBinaryExpressionIsNecessaryConditional(
-      node: TSESTree.BinaryExpression,
+      node: TSESTree.Node,
+      {
+        left,
+        right,
+        operator,
+      }: Pick<TSESTree.BinaryExpression, 'left' | 'right' | 'operator'>,
     ): void {
-      if (!BOOL_OPERATORS.has(node.operator)) {
```
kirkwaiblinger, line 369: (nit) - i'm not sure i agree with moving this check outside the function. The function's title implies that it can accept any binary expression, so it would seem correct for it to validate first that the expression is a conditional, unless the `TSESTree.BinaryExpression` type itself could guarantee that the binary expression is in `BOOL_OPERATORS`.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -719,13 +721,27 @@ export default createRule<Options, MessageId>({
...
       CallExpression: checkCallExpression,
       ConditionalExpression: (node): void => checkNode(node.test),
       DoWhileStatement: checkIfLoopIsNecessaryConditional,
       ForStatement: checkIfLoopIsNecessaryConditional,
       IfStatement: (node): void => checkNode(node.test),
       LogicalExpression: checkLogicalExpressionForUnnecessaryConditionals,
+      SwitchCase({ test, parent }): void {
+        // only check `case ...:`, not `default:`
+        if (test) {
+          checkIfBinaryExpressionIsNecessaryConditional(test, {
```
kirkwaiblinger, line 738: clever way of reusing this function 👍

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
kirkwaiblinger, line 1: (testing) Please add test coverage... The two test cases added here already pass on main and do not test anything related to the issue.

kirkwaiblinger (changes requested):
The code change looks right to me, but requesting changes since there aren't currently any relevant tests.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
abrahamguo: Rather than adding a runtime check that would run unnecessarily for `SwitchCase`s, I instead
- kept the runtime check to continue only checking on `BinaryExpression`s
- used TypeScript to ensure only valid operators are passed into the function (since why use runtime checks when we can use type checks? 🤷‍♂️ )
- renamed the function to be consistent with the existing `BOOL_OPERATORS` constant and make clear that it is only for a subset of `BinaryExpression`s

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
abrahamguo: Thanks!

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
abrahamguo: Good catch! I updated the two tests so that the `invalid` test for `switch` now fails on `main`.

Note that the overall workflow of the existing tests is that they have these two tests, commented as `// Ensure that it's checking in all the right places`. These two tests are the only places where some constructs, such as `for`/`while`/`do`-`while` appear. Then the rest of the tests test the common boolean logic shared between various constructs.

I followed this pattern and only added two valid and two invalid tests for `switch`es (one each for `case` and `default`) but if you'd like me to add any other tests I can! 👍

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
```diff
@@ -15,7 +15,7 @@ Conversely, any expression that always evaluates to truthy or always evaluates t
 The following expressions are checked:
 
 - Arguments to the `&&`, `||` and `?:` (ternary) operators
-- Conditions for `if`, `for`, `while`, and `do-while` statements
+- Conditions for `if`, `switch`, `for`, `while`, and `do-while` statements
```
kirkwaiblinger, line 18: (nit) I think "conditions for `switch` statements" isn't _quite_ as self-explanatory as the rest of these. What about the following? Or some other way of being explicit that it's the `case`s that are checked?

```suggestion
- Conditions for `if`, `for`, `while`, and `do-while` statements
- `case`s in `switch` statements
```

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
kirkwaiblinger: Yep, using the type system to enforce it is good, too. Nice 👍

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -1,3 +1,5 @@
+import console from 'node:console';
+
```
kirkwaiblinger, line 2:
```suggestion
```

abrahamguo:
Failed due to unrelated CI failure on `main`.

JoshuaKGoldberg (approved):
LGTM!

kirkwaiblinger:
I'm happy! 

Thanks! 🙂

kirkwaiblinger (approved):

kirkwaiblinger:
afterthought - the message "Unnecessary conditional, both sides of the expression are literal values." is a little awkward for the `case` values. But I think there's no need to worry about special-casing the wording unless someone complains. I think it's pretty clear in context what's meant by the report

PR #9921, opened 2024-09-03
fix(eslint-plugin): [consistent-type-assertions] access parser services lazily

### LINKED ISSUE #9860: Bug: [consistent-type-assertions] should not error when parser services are not available

### Issue Description

See https://github.com/typescript-eslint/typescript-eslint/issues/9855#issuecomment-2302246491:

> The issue is that the rule is using the TS AST which comes from parser services.
> 
> The util shouldn't eagerly error in this case though, I don't think. It should lazily error when the rule accesses the services instead - I.e. Make the properties getters that throw.
> 
> Also we could probably fix the rule to not use the TS AST. I think we updated the utils to allow us to leverage the estree ast for operator precedence now?
> 
> ---
>
> My opinion: non-type-aware rules should never crash on a file for which they match no selectors.
> 
> I.e. Json or angular template files that don't use the same ast structure shouldn't crash in our rules like this.

In the reproduction for this issue (see Stackblitz link below), the rule is enabled on a JavaScript file parsed with Espree, not typescript-eslint's parser. Instead of doing nothing, the rule errors:

```
Error while loading rule '@typescript-eslint/consistent-type-assertions': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
Parser: (unknown)
Note: detected a parser other than @typescript-eslint/parser. Make sure the parser is configured to forward "parserOptions.project" to @typescript-eslint/parser.
```

https://stackblitz.com/edit/vitejs-vite-fgbcrl?file=eslint.config.js

### Repro Steps

1. `npm install`
2. `npm run lint`

### Versions

```json
{
    "eslint": "9.9.0",
    "typescript": "5.3.3",
    "typescript-eslint": "8.2.0"
}
```

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9860

## Overview

Adds `espree` as a `devDependency` to be able to use it as a parser in the unit test.

💖

### DIFF
--- a/packages/eslint-plugin/src/rules/consistent-type-assertions.ts
+++ b/packages/eslint-plugin/src/rules/consistent-type-assertions.ts
@@ -124,6 +124,7 @@ export default createRule<Options, MessageIds>({
         fix:
           messageId === 'as'
             ? (fixer): TSESLint.RuleFix => {
+                // lazily access parserServices to avoid crashing on non TS files (#9860)
                 const tsNode = getParserServices(
                   context,
                   true,
--- a/packages/eslint-plugin/src/rules/consistent-type-assertions.ts
+++ b/packages/eslint-plugin/src/rules/consistent-type-assertions.ts
@@ -94,8 +94,6 @@ export default createRule<Options, MessageIds>({
     },
   ],
   create(context, [options]) {
-    const parserServices = getParserServices(context, true);
-
     function isConst(node: TSESTree.TypeNode): boolean {
       if (node.type !== AST_NODE_TYPES.TSTypeReference) {
         return false;
@@ -126,9 +124,10 @@ export default createRule<Options, MessageIds>({
         fix:
           messageId === 'as'
             ? (fixer): TSESLint.RuleFix => {
-                const tsNode = parserServices.esTreeNodeToTSNodeMap.get(
-                  node as TSESTree.TSTypeAssertion,
-                );
+                const tsNode = getParserServices(
+                  context,
+                  true,
+                ).esTreeNodeToTSNodeMap.get(node as TSESTree.TSTypeAssertion);
 
                 const expressionCode = context.sourceCode.getText(
                   node.expression,
--- a/packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
+++ b/packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
@@ -149,6 +149,7 @@ ruleTester.run('consistent-type-assertions', rule, {
     {
       code: '123;',
       languageOptions: {
+        // simulate a 3rd party parser that doesn't provide parser services
         parser: {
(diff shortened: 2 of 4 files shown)


packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
```diff
@@ -144,6 +145,12 @@ ruleTester.run('consistent-type-assertions', rule, {
         },
       ],
     },
+    {
+      code: "console.log('Hello, world!');",
+      languageOptions: {
+        parser: espree,
```
bradzacher, line 151: Instead of bringing in espree and updating it manually etc.
Could we just do a similar thing to what ESLint does for TS parsing -- parse it with the parser, dump the json blob out and then define a fake parser module eg
```
languageOptions: {
  parser: {
    parse: () => { type: 'Program', ... },
  },
},
```

WDYT?

packages/eslint-plugin/src/rules/consistent-type-assertions.ts
```diff
@@ -126,9 +124,10 @@ export default createRule<Options, MessageIds>({
         fix:
           messageId === 'as'
             ? (fixer): TSESLint.RuleFix => {
-                const tsNode = parserServices.esTreeNodeToTSNodeMap.get(
-                  node as TSESTree.TSTypeAssertion,
-                );
+                const tsNode = getParserServices(
```
bradzacher, line 127:
```suggestion
                // lazily access parserServices to avoid crashing on non TS files (#9860)
                const tsNode = getParserServices(
```

bradzacher (dismissed):
actual change LGTM -- just the one question re espree dep

packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
kirkwaiblinger: +1 to this suggestion

JoshuaKGoldberg: I did this, then realized we call `parser.parse` manually in another file. I did that with a separate string literal.

packages/eslint-plugin/tests/rules/consistent-type-assertions.test.ts
```diff
@@ -144,6 +146,14 @@ ruleTester.run('consistent-type-assertions', rule, {
         },
       ],
     },
+    {
+      code: '123;',
+      languageOptions: {
+        parser: {
+          parse: (): TSESTree.Program => parser.parse('123;'),
+        },
```
bradzacher, line 154:
```suggestion
        // simulate a 3rd party parser that doesn't provide parser services
        parser: {
          parse: (): TSESTree.Program => parser.parse('123;'),
        },
```

bradzacher (approved):

PR #9924, opened 2024-09-03
feat(eslint-plugin): [prefer-nullish-coalescing] add option `ignoreBooleanCoercion`

### LINKED ISSUE #9080: Enhancement: [prefer-nullish-coalescing] should ignore Boolean constructor

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.3&fileType=.ts&code=DYUwLgBAhgXBDOYBOBLAdgcwgHwsgriDhPmgCYgBm6IZA3ALABQokARnIqpsWwPZ9QUNMVIVqaWoxbgIAYzj9BIYaPJUa9ZswD0OvCEQQAjBAAWIJEQC2KtPAR9bEPpTwXobORBQOCYMwBPZjk%2Be0gwQzBTAF4IACEBITQACigcXDYM%2BQBKaVDwg0QAJgg4xOVhNJzsiuSUthrsXDq7FLkc7SYC%2BGUAOmA%2BDBTIxGM85iA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6MAG1w2HAHdJ0JpAA0a9VnXZIleU3GcAwuKYATSvkp3pAFRT5UGfHESHsAF8AgF01YMCgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
let a: string | true | undefined;
let b: string | boolean | undefined;
let c: boolean | undefined;

// test 1 here means some of the abc is truthy
const test1 = Boolean(a || b || c);
const test2 = Boolean(a) || Boolean(b) || Boolean(c)

console.log(test1);
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/prefer-nullish-coalescing": [
      "warn",
      {
        "ignoreConditionalTests": true
      }
    ]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

Eslint should not try to fix the code, as this is a correct usage.

### Actual Result

The fix is invalid, as when:

```
a = ''
b = 'something'
```

The fix result `Boolean((a ?? b) ?? c)` is false, while the snippet is meant to be true.

I understand that I can write the way in `test2`, but it will have performance issues that incease code exec steps, and making the output bigger. The using of `Boolean` is that we usually need a clean state type with `boolean` to use it in other places (e.g.: Array.filter)

### Additional Info

IMO, when setting `ignoreConditionalTests: true`, the `Boolean` constructor and `!(/* codes */)` should also be checked. This means users are wanting to have a full truthy check on all of them.

I have encounted 20+ times with similar issues in my code repo, as it's really common to get a state which is "some of the conditions are truthy"

### LINKED ISSUE #10153: Bug: [prefer-nullish-coalescing] faulty detection of conditional tests with `ignoreConditionalTests: true`

### Playground Link

[link](https://typescript-eslint.io/play/#ts=5.5.2&showAST=es&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3lAXPTMAligObwA%2B8KArhBANwBQoksCy6WARngcWZRp0mzcNDjwAZtRRgMhVFIAUADzwyA1iiQB3FAEp1KLbpQiA9OcRIYcORACe8OAAcbGNI0lKoFSl30LKzQACyRaYHguaixgJBA0FAByLFd3RkJJeCVvX3J-fX14AG8AXyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6MAG1w2HAHdJ0JpAA0a9VnXZIleU3GcAwuKYATSvkp3pAFRT5UGfHESHsAF8AgF01YMCgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

### Repro Code

```TypeScript
declare const a: string | null;
declare const b: string | null;

declare function f(x: unknown): unknown;

// correctly reports
f(a || b);

// should but doesn't report
if (f(a || b)) {}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/prefer-nullish-coalescing": [
      "warn",
      {
        "ignoreConditionalTests": true
      }
    ]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expect that `if (f(a || b)) {}` should report, since `a || b` is being used as a function argument, not the test of an `if` statement.

### Actual Result

it doesn't report

### Additional Info

Just popping this up as a standalone issue for visibility/tracking. It will likely be addressed in https://github.com/typescript-eslint/typescript-eslint/pull/9924#discussion_r1795649396

cc @developer-bandi

developer-bandi:
## PR Checklist

   - fixes #9080
   - fixes #10153

## Overview
I read the comments on the issue, and it seemed better to create a new option rather than adding a function to ignoreConditionalTests, so I created a new option.

The reason is that I thought it was reasonable to say that there is no case in typescript-eslint where the content raised in the issue is judged as a condition/boolean.


packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -436,6 +448,43 @@ function isConditionalTest(node: TSESTree.Node): boolean {
...
+  while (current) {
+    parents.add(current);
+
+    if (
+      (current.type === AST_NODE_TYPES.UnaryExpression &&
+        current.operator === '!') ||
+      (current.type === AST_NODE_TYPES.CallExpression &&
+        current.callee.type === AST_NODE_TYPES.Identifier &&
+        // eslint-disable-next-line @typescript-eslint/internal/prefer-ast-types-enum
+        current.callee.name === 'Boolean')
```
JoshuaKGoldberg, line 463: [Bug] Folks sometimes write their own `function Boolean( ... ) { ... }`. This rule will need to handle that well.

Previous example: https://github.com/typescript-eslint/typescript-eslint/pull/10005#discussion_r1771754649

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -436,6 +448,43 @@ function isConditionalTest(node: TSESTree.Node): boolean {
...
+        AST_NODE_TYPES.ArrowFunctionExpression,
+        AST_NODE_TYPES.FunctionExpression,
+      ].includes(current.type)
+    ) {
+      /**
+       * This is a weird situation like:
+       * `if (() => a || b) {}`
+       * `if (function () { return a || b }) {}`
+       */
+      return false;
```
JoshuaKGoldberg, line 479: [Testing] Good to see that the cases tripping this up are known 😄.  

Also to test are other unusual things in `if`s:

* `class` expressions
* `new` expressions
* Function decorators
* ...I'm drawing a blank on more spooky things at the moment, but there's a good chance more exist.

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.mdx
```diff
@@ -172,6 +172,34 @@ foo ?? 'a string';
 
 Also, if you would like to ignore all primitives types, you can set `ignorePrimitives: true`. It is equivalent to `ignorePrimitives: { string: true, number: true, bigint: true, boolean: true }`.
 
+### `ignoreMakeBoolean`
```
JoshuaKGoldberg, line 175: [Docs] _"make"_ is a general word. Let's go with `ignoreBooleanCall`:

```suggestion
### `ignoreBooleanCall`
```

JoshuaKGoldberg (changes requested):
Hey sorry for the long wait! This is a good start, but I think has some edge case handling and testing to go. 🚀

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
developer-bandi: 1. Because I was copying the isConditionalTest function, I forgot to change the example case. In practice, an example would be `Boolean(()=> a || b)`. 
Therefore, it seems that the situations you mentioned should actually apply to the isConditionalTest function as well. Should the work be done in this pr?

2. Actually, I did not exactly understand the three cases you mentioned. Could you please give an example if possible?

3. The additional case I was considering is an object. Should cases like `Boolean({test:a||b})` also be included?

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -103,6 +104,11 @@ export default createRule<Options, MessageIds>({
               'Whether to ignore any ternary expressions that could be simplified by using the nullish coalescing operator.',
             type: 'boolean',
           },
+          ignoreBooleanCoercion: {
+            description:
+              'Whether to ignore any make boolean type value like `Boolan`, `!`',
```
kirkwaiblinger, line 109:
```suggestion
              'Whether to ignore any make boolean type value like `Boolean`, `!`',
```

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -436,6 +448,60 @@ function isConditionalTest(node: TSESTree.Node): boolean {
...
+      return false;
+    }
+
+    current = current.parent;
+  }
+
+  return false;
+}
+
+function isBuiltInBoolanCall(
```
kirkwaiblinger, line 489:
```suggestion
function isBuiltInBooleanCall(
```

kirkwaiblinger (changes requested):
Why does this option include allowing arguments to the `!` operator? The discussion in the issue explicitly contradicts this https://github.com/typescript-eslint/typescript-eslint/issues/9080#issuecomment-2106370534

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -436,6 +448,58 @@ function isConditionalTest(node: TSESTree.Node): boolean {
...
+  while (current) {
+    parents.add(current);
+
+    if (
+      current.type === AST_NODE_TYPES.CallExpression &&
+      isBuiltInBooleanCall(current, context)
+    ) {
+      return true;
+    }
+
```
kirkwaiblinger, line 466: So, the algorithm in this function definitely has some problems.... 

For example, the following code will be ignored by the new option
```ts
let a: string | true | undefined;
let b: string | boolean | undefined;

declare function f(x: unknown): unknown;

const x = Boolean(f(a || b));
```

However, I'm also noticing that the existing code works quite similarly.... 

https://github.com/typescript-eslint/typescript-eslint/blob/723d0a8ec655f89f0383ca8ed76158368b6d699b/packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts#L414-L449

So I'm wondering whether that has bugs too....

Without that context, I'd think we'd want to model this algorithm closer to what's in other rules, for example, no-floating-promises or no-extra-boolean-cast. Those rules recurse specifically on logical expressions, `??` expressions, ternaries, chain expressions, sequence expressions.

kirkwaiblinger (changes requested):

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -436,6 +448,58 @@ function isConditionalTest(node: TSESTree.Node): boolean {
   return false;
 }
 
+function isMakeBoolean(
```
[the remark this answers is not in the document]
developer-bandi: changed to isBooleanConstructorContext.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -436,6 +448,58 @@ function isConditionalTest(node: TSESTree.Node): boolean {
...
+  return false;
+}
+
+function isBuiltInBooleanCall(
+  node: TSESTree.CallExpression,
+  context: Readonly<TSESLint.RuleContext<MessageIds, Options>>,
+): boolean {
+  if (
+    node.callee.type === AST_NODE_TYPES.Identifier &&
+    node.callee.name === AST_TOKEN_TYPES.Boolean &&
```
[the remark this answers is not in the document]
developer-bandi: A thought occurred to me while looking at the code below. Wouldn't it be necessary to include the case of calling Boolean with the new operator?

https://github.com/eslint/eslint/blob/main/lib/rules/no-extra-boolean-cast.js#L84-L90

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
developer-bandi: upon checking, it appears that similar code is ignored when the ignoreConditionalTests option is true [playground](https://typescript-eslint.io/play/#ts=5.5.2&showAST=es&fileType=.ts&code=CYUwxgNghgTiAEEQBd5QFzwM7JgSwDsBzeAH3gIFcIIBuAKFElgTAHsCd4AjTHfYmQrU69RuGhx4AM0oEwyPBxkAKAB6Y5AawJsA7gQCUmgjv0EGeafBXSVUMuW6HD8AN4BfenoAWeJDZ2DqROLu5eQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6MAG1w2HL2jjokADRr1WddkiV5TbYgDC4pgBNK%2BSrekAVFPlQZ8cRAewAvv4AumpBAUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

```
declare let a: string | null;
declare const b: string | null;

declare function f(x: unknown): unknown;
if (f(a || b)) {}
while (f(a || b)) {}
```
so i think isConditionalTest function need to modified. Would it be a good idea to open a new issue for this problem?
_________________________________________________________________to modify algorithm, I looked at the `no-extra-boolean-cast` code you mentioned.

considering what you mentioned, it seems like you can improve it by referring to the algorithm of the function below, but I am wondering if my understanding is correct.

https://github.com/eslint/eslint/blob/main/lib/rules/no-extra-boolean-cast.js#L117-L167

kirkwaiblinger: What fun, haha. Thanks for checking on this!

I think file no matter what, and it's up to you if you feel motivated and want to fix it in this PR or prefer to not worry about it in this work. 👍

kirkwaiblinger: > considering what you mentioned, it seems like you can improve it by referring to the algorithm of the function below, but I am wondering if my understanding is correct.

Yeah, exactly, that's the pattern I'm suggesting to study 👍 (just the first thing that came to my mind, since [I have previously touched that code](https://github.com/eslint/eslint/pull/18222/files))

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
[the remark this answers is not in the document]
kirkwaiblinger: good question. IMO no; `new Boolean(expr)` is a) rarely used, b) usually a mistake if it is used, and c) a different thing entirely from `Boolean(expr)`, considering the result.... isn't a boolean. But, I would say maybe propose it on the issue if you want input from others? In my opinion, though, I think you're 100% fine to omit `new Boolean(expr)` and if someone wants that as a feature one day they can request it specifically.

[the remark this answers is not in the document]
developer-bandi: thank you I think it would be better to just skip it!

developer-bandi:
i edit the `isConditionalTest` and `isBooleanConstructorContext` with no-extra-boolean-cast rule's code. 

These functions have been modified in almost the same way, changing from recursively searching if a parent exists to recursively searching only if the logical operation conditions are satisfied.

The logical operation conditions are the same for both functions, but recursive search is allowed only when the result of the logical operation is used as the value. 

**recursive is not allowed**
- `if(f(a||b))`
- `if(()=>a||b)`
- `if((a||b)?.[c])`
....

**recursive is allowed **
- `if (a || (b && c)) {}`
- `if (a ?? (b || c)) {}`
...

For more examples, please refer to the test code.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -399,40 +411,83 @@ export default createRule<Options, MessageIds>({
   },
 });
 
-function isConditionalTest(node: TSESTree.Node): boolean {
-  const parents = new Set<TSESTree.Node | null>([node]);
-  let current = node.parent;
-  while (current) {
-    parents.add(current);
+function isConditionalTest(node: TSESTree.Node,parents=new Set<TSESTree.Node | null>([node])): boolean {
```
kirkwaiblinger, line 414: I was playing around with this locally and was able to simplify the implementations of these functions and maintain all the tests passing.... WDYT?

The main changes are: 
- The `parents` set does not appear to be needed at all and can be removed.
- I've checked for nullish parents at the beginning of the functions to be explicit and also cut down on `?.` usages.
- `isBuiltInBooleanCall(node)` shouldn't even get called if `node` is null, so I've changed the signature and moved some of the null checking.
- (optional) I've aliased `node.parent` for simplicity but that's just my personal preference.

```ts
function isConditionalTest(node: TSESTree.Node): boolean {
  const parent = node.parent;
  if (!parent) {
    return false;
  }

  if (parent.type === AST_NODE_TYPES.LogicalExpression) {
    return isConditionalTest(parent);
  }

  if (
    parent.type === AST_NODE_TYPES.ConditionalExpression &&
    (parent.consequent === node || parent.alternate === node)
  ) {
    return isConditionalTest(parent);
  }

  if (
    parent.type === AST_NODE_TYPES.SequenceExpression &&
    parent.expressions.at(-1) === node
  ) {
    return isConditionalTest(parent);
  }

  if (
    (parent.type === AST_NODE_TYPES.ConditionalExpression ||
      parent.type === AST_NODE_TYPES.DoWhileStatement ||
      parent.type === AST_NODE_TYPES.IfStatement ||
      parent.type === AST_NODE_TYPES.ForStatement ||
      parent.type === AST_NODE_TYPES.WhileStatement) &&
    parent.test === node
  ) {
    return true;
  }

  return false;
}

function isBooleanConstructorContext(
  node: TSESTree.Node,
  context: Readonly<TSESLint.RuleContext<MessageIds, Options>>,
): boolean {
  const parent = node.parent;
  if (parent == null) {
    return false;
  }

  if (parent.type === AST_NODE_TYPES.LogicalExpression) {
    return isBooleanConstructorContext(parent, context);
  }

  if (
    parent.type === AST_NODE_TYPES.ConditionalExpression &&
    (parent.consequent === node || parent.alternate === node)
  ) {
    return isBooleanConstructorContext(parent, context);
  }

  if (
    parent.type === AST_NODE_TYPES.SequenceExpression &&
    parent.expressions.at(-1) === node
  ) {
    return isBooleanConstructorContext(parent, context);
  }

  return isBuiltInBooleanCall(parent, context);
}

function isBuiltInBooleanCall(
  node: TSESTree.Node,
  context: Readonly<TSESLint.RuleContext<MessageIds, Options>>,
): boolean {
  if (
    node.type === AST_NODE_TYPES.CallExpression &&
    node.callee.type === AST_NODE_TYPES.Identifier &&
    // eslint-disable-next-line @typescript-eslint/internal/prefer-ast-types-enum
    node.callee.name === 'Boolean' &&
    node.arguments[0]
  ) {
    const variable = scope.set.get(AST_TOKEN_TYPES.Boolean);
    return variable == null || variable.defs.length === 0;
  }
  return false;
}
```

developer-bandi: It looks good. We will reflect all the changes you mentioned. It's not important, but I have a question. 

The isConditionalTest function returns a false value, 
isBooleanConstructorContext compares null for equality. Is there a reason there is a difference between the two functions?

```
function isConditionalTest(node: TSESTree.Node): boolean {
  const parent = node.parent;
  if (!parent) {
    return false;
  }

function isBooleanConstructorContext(
  node: TSESTree.Node,
  context: Readonly<TSESLint.RuleContext<MessageIds, Options>>,
): boolean {
  const parent = node.parent;
  if (parent == null) {
    return false;
  }
```

kirkwaiblinger: oh - no reason! I have a personal thing where I prefer to avoid truthy/falsey checks and use explicit `== null` conditions instead, so I probably just applied that habit to one case and not the other. Feel free to make those consistent whichever way you prefer 👍

kirkwaiblinger (approved):
Thanks for going above and beyond by fixing issues in the existing code in addition to the added functionality!

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -103,6 +104,11 @@ export default createRule<Options, MessageIds>({
               'Whether to ignore any ternary expressions that could be simplified by using the nullish coalescing operator.',
             type: 'boolean',
           },
+          ignoreBooleanCoercion: {
+            description:
+              'Whether to ignore any make boolean type value like `Boolean`',
```
kirkwaiblinger, line 109:
```suggestion
              'Whether to ignore arguments to the `Boolean` constructor',
```

developer-bandi:
added commit to resolve conflicts.

kirkwaiblinger (approved):

bradzacher (approved):
thanks!

Issue #5393, opened 2022-07-28
Repo: Use mdast-util-from-markdown in Docusaurus plugins


JoshuaKGoldberg:
### Suggestion

Right now our [`generated-rule-docs.ts` file](https://github.com/typescript-eslint/typescript-eslint/blob/5bc4d43bed1ccb0b5c6bebe049409e5fc6d937eb/packages/website/plugins/generated-rule-docs.ts#L27) manually inserts POJO objects for the MDAST. It'd be great if we could use a package like https://github.com/syntax-tree/mdast-util-from-markdown to clean up that code.

However, that package is written in ESM, which Docusaurus doesn't support yet. Adding this inside a plugin:

```js
import { fromMarkdown } from 'mdast-util-from-markdown';
```

...gives a loading failure:

```plaintext
[INFO] Starting the development server...
[ERROR] Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/josh/repos/typescript-eslint/node_modules/mdast-util-from-markdown/index.js from /Users/josh/repos/typescript-eslint/packages/website/plugins/generated-rule-docs.ts not supported.
Instead change the require of index.js in /Users/josh/repos/typescript-eslint/packages/website/plugins/generated-rule-docs.ts to a dynamic import() which is available in all CommonJS modules.
    at Object.require.extensions.<computed> [as .js] (/Users/josh/repos/typescript-eslint/node_modules/ts-node/dist/index.js:785:20)
    at Object.<anonymous> (/Users/josh/repos/typescript-eslint/packages/website/plugins/generated-rule-docs.ts:8:30)
    at Module.m._compile (/Users/josh/repos/typescript-eslint/node_modules/ts-node/dist/index.js:791:29)
    at Object.require.extensions.<computed> [as .ts] (/Users/josh/repos/typescript-eslint/node_modules/ts-node/dist/index.js:793:16)
    at Object.<anonymous> (/Users/josh/repos/typescript-eslint/packages/website/docusaurusConfig.ts:5:26)
    at Module.m._compile (/Users/josh/repos/typescript-eslint/node_modules/ts-node/dist/index.js:791:29)
    at Object.require.extensions.<computed> [as .ts] (/Users/josh/repos/typescript-eslint/node_modules/ts-node/dist/index.js:793:16)
    at Object.<anonymous> (/Users/josh/repos/typescript-eslint/packages/website/docusaurus.config.js:10:18)
[...]

PR #9925, opened 2024-09-03
docs: inject option descriptions into rule docs when possible

### LINKED ISSUE #9867: Docs: Automatically inject rule option descriptions into their docs

### Suggested Changes

Following #5392, most rule options will have `description`s in their schemas. For many of those options, their docs either:

* Don't have an actual description of the option (see the `allowAny` example later)
* As of #9868, roughly -or exactly- match what's on their docs pages

Proposal: let's have the rule docs generation also inject the `description` and default value for each rule option into its `###` heading?

For example, the `allowAny` `description` in `restrict-plus-operands`:

https://github.com/typescript-eslint/typescript-eslint/blob/ce79c552da146eeb565a1be37f2568ba4f281881/packages/eslint-plugin/src/rules/restrict-plus-operands.ts#L61-L62

...could be used as a first line in the docs page under its h3:

https://github.com/typescript-eslint/typescript-eslint/blob/a7bdd1c1eb3a403c919e516d58b1d667ea23219a/packages/eslint-plugin/docs/rules/restrict-plus-operands.mdx#L75-L77

We could also mention the default value.

In other words, the end result I'm proposing is:

```diff
### `allowAny`

+ Whether to allow `any` typed values.
+ Default: `false`.

Examples of code for this rule with `{ allowAny: true }`:
```

We'd have to do a pass to remove redundant lines in the `.mdx` files.

### Affected URL(s)

💖

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9867

## Overview

For each rule whose `meta.schema[0]` is an object with `type: "properties"` or `type: "oneOf"`, this adds the `description` for each of those option properties to the rule's docs page.

💖


packages/website/plugins/generated-rule-docs/RuleDocsPage.ts
```diff
@@ -1,8 +1,9 @@
 import type { ESLintPluginRuleModule } from '@typescript-eslint/eslint-plugin/use-at-your-own-risk/rules';
+import { fromMarkdown } from 'mdast-util-from-markdown';
```
JoshuaKGoldberg, line 2: I accidentally coincidentally found in this PR that #5393 is no longer blocked! Hooray! Will split that out into a separate PR when I get a chance.

JoshuaKGoldberg: Merged.

packages/eslint-plugin/docs/rules/consistent-type-assertions.mdx
```diff
@@ -42,8 +42,6 @@ Some codebases like to go for an extra level of type safety, and ban assertions
 
 ### `objectLiteralTypeAssertions`
 
-Whether to always prefer type declarations for object literals used as variable initializers, rather than type assertions.
```
kirkwaiblinger, line 45: Hmm. I think this implicit injection makes the rule docs source hard to understand. If we do inject like this, can it be something explicitly reflected in the source, so that the source reads more like the following? 

```mdx
### `objectLiteralTypeAssertions`

<RuleOptionDescription for='consistent-type-assertions' option='objectLiteralTypeAssertions'/>

For example, this would prefer `const x: T = { ... };` to `const x = { ... } as T;` (or similar with angle brackets).

(etc)
```

I don't have a lot of familiarity with best practices here, but I'm worried that the change as-is really hampers understandability of the doc page source especially for community contributors.

JoshuaKGoldberg: Hmm, I can see what you're saying... but that's so much manual work for something we _know_ for certain on each rule's option. 😬

Even if we pare it down to `<RuleOptionDescription option="objectLiteralTypeAssertions" />` using current file info (which we can), we'd then also have to write a unit test or lint rule to make sure they don't fall out of sync.

The rule docs are already pretty automated. My leaning is towards continuing that.

Arbiter and Master Chief back to back in Halo 3, with Gravemind tentacles behind them. Arbiter is speaking: "we trade one villain for another"

JoshuaKGoldberg: Chatted a bit with @Josh-Cena and @kirkwaiblinger - I'll switch to something more explicit. Both are in favor of making it more clear where stuff will be injected.

kirkwaiblinger (approved):
I'm happy with this approach!

Issue #9934, opened 2024-09-04
Bug: Vue SFC with lang `tsx` does not parse as tsx with projectService 


blake-newman:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Abug) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Relevant Package

typescript-estree

### Repro Code

```TypeScript
<script lang="tsx">
defineComponent({
  setup() {
    return () => <div />
  }
})
</script>
```

### ESLint Config

```javascript
{
    files: ['**/*.vue'],
    languageOptions: {
      parser: vueEslintParser,
      parserOptions: {
        projectService: true,
        parser: tseslint.parser,
        extraFileExtensions,
        tsconfigRootDir: import.meta.dirname,
        ecmaFeatures: {
          jsx: true,
        },
      },
    },
  },

  // .ts/tsx files
  {
    files: ['**/*.{ts,tsx}'],
    languageOptions: {
      parser: tseslint.parser,
      parserOptions: {
        projectService: true,
        extraFileExtensions,
        tsconfigRootDir: import.meta.dirname,
        ecmaFeatures: {
          jsx: true,
        },
      },
    },
  },
```

### tsconfig

```jsonc
Not applicable
```

### Expected Result

No issues parsing `.vue` files that have `lang="tsx"`

### Actual Result

 > 49:17  error  Parsing error: '>' expected

### Additional Info

This is caused as the project service is allowed to infer the script kind by default:

```
serviceSettings.service.openClientFile(filePathAbsolute, parseSettings.codeFullText,  /* scriptKind */ undefined, /*...*/)
```

In order to resolve this `jsx` is defined on parser options then we can derive the script kind for the file if the extension matches the `extraFileExtensions`. 

Alternatively we could allow `extraFileExtensions` to be an object, which can configure the extension types and there script type. 

As `vue-eslint-parser` only parses the extracted script code there is no way for that parser to feed the correct type. However; i think it's save to assume that if you specify `ecmaFeatures.jsx` in eslint configurations that you are explicitly using TSX and want to parse vue files this way.

```
extensions: {

}
```

### Versions

[...]

PR #9935, opened 2024-09-04
feat(typescript-estree): support TSX with extra file extensions

blake-newman:
If opening a file is part of extra file extensions and JSX is set, then set the ScriptKind to TSX so that the project service can correctly parse the code as TSX, which by default it is infered as `ScriptKind.TS`.

resolves: #9934


bradzacher:
Thanks for the proactive PR.
As per our contributing guide we generally prefer waiting to implement until an issue is marked as accepting PRs so that we can discuss and agree upon a course of action and avoid anyone working in the wrong direction.

Closing for the moment.
If we decide in the issue this is the right direction we can reopen and review!

PR #9937, opened 2024-09-04
feat(type-utils): isNullableType add Void logic

### LINKED ISSUE #8974: Repo: Incorporate void check into isNullableType?
### Suggestion

In https://github.com/typescript-eslint/typescript-eslint/pull/8912/files#r1572612158, the `void` type is treated as a nullable type for the purposes of no-unnecessary-type-assertion, the same as `null` and `undefined` would be. Consider whether checking for the `void` type should be incorporated into the helper method `isNullableType` already used for checking for the `null` and `undefined` type, and assess whether its impact on other rules that use that helper is a correct behavior change or not.

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8974

## Overview

Since there are two rules using this utility function, I removed the same logic from the rule and added test code.

And I tried to add a unit test for isNullableType, but I couldn't find the original unit test file. Is it true that unit tests are not originally written for this utility file?

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -644,10 +644,9 @@ export default createRule<Options, MessageId>({
             ? !isCallExpressionNullableOriginFromCallee(node)
             : true;
 
-      const possiblyVoid = isTypeFlagSet(type, ts.TypeFlags.Void);
       return (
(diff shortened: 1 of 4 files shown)


kirkwaiblinger (approved):
void gif

JoshuaKGoldberg (approved):
🆒 

Yeah, tests are a bit lacking in some of the ancillary packages.

Issue #8088, opened 2023-12-18
Docs: Overhaul no-floating-promise messages and docs


JoshuaKGoldberg:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

Right now, [`@typescript-eslint/no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises) logs quite a mouthful for its messages:

```plaintext
Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.
```

So much text! Longer messages are less likely to be read by users - especially "error" (or "warning") messages like this one.

In addition, it's also missing the point that storing the Promise in a variable is considered an acceptable workaround by the rule (assuming you have unused variable detection enabled). https://github.com/typescript-eslint/typescript-eslint/discussions/8046 is an example of a discussion asking about that.

This is a nuanced area. See replies to: https://twitter.com/flybayer/status/1469098848958222337, https://twitter.com/threepointone/status/1471458830626299911, https://twitter.com/nandafyi/status/1736164705448956388, and other threads from searching on `no-floating-promises`). I think we should rework the rule a bit to coach/encourage developers into treating the nuance with care:

* Make the error message much shorter (so folks will actually read the whole thing), and perhaps encourage them to read the docs for more info
* Mention the unused variable allowance in the docs
* Link to a blog post explaining how to enable our Promise/Thenable-area rules (perhaps a separate issue?)

### Affected URL(s)

https://typescript-eslint.io/rules/no-floating-promises

JoshuaKGoldberg:
Since #7008 was added to the 8 milestone for tracking, adding this as well. I think we'll really need to have good docs to make sure folks don't misuse this.

JoshuaKGoldberg:
Btw, referencing https://github.com/typescript-eslint/typescript-eslint/issues/7008 & https://github.com/nodejs/node/issues/51292 & https://github.com/reduxjs/redux-toolkit/pull/4102: we should definitely mention strategies for frameworks/libraries to mark Promises as safe (or not expose them).

JoshuaKGoldberg:
[...]

PR #9949, opened 2024-09-06
docs: [no-floating-promises] clarify that void does not resolve promises

### LINKED ISSUE #9947: Docs: [no-floating-promises] voiding Promise doesn't make it handled

### Suggested Changes

Hi!

[[no-floating-promises]](https://typescript-eslint.io/rules/no-floating-promises/) cite what follows:

```
[...]

This rule reports when a Promise is created and not properly handled. Valid ways of handling a Promise-valued statement include:

[...]
- voiding it
[...]
```

However, **voiding a Promise doesn't actually make it handled**. The result is just ignored.

If a voided Promise is rejected, an error is still thrown and thus the failure is **just ignored**. However, the failure is still there. In fact:
 - a listener for `unhandledrejection` still gets invoked;
 - an error is still logged in the console and eventually in enterprise logging systems (e.g. Sentry);

Therefore, it is not exactly correct to say that `void` operator is a valid way of handling a Promise. **A promise should still be followed by a .catch() (even with a noop function) right after the invokation (not after or in an if)**.

This code proves what happens. I trusted this package, I voided some Promises I didn't expect could reject and, after enabling Sentry, I ended up with a ton of reports about unhandled rejections.

```javascript
	window.addEventListener("unhandledrejection", (err) => {
		console.log("Unhandled Rejection", err);
	});

	async function f1() {
		return Promise.reject("Timeout expired");
	}

	void f1();
	console.log("test");
```

Logs orders:
 - (log) `"test"`
 - (log) `"Unhandled Rejection"`
 - (error) `Uncaught (in promise) Timeout expired`

immagine

I'd improve the documentation by putting a warning on void about this behavior, both above in the list (add a link to `ignoreVoid`) and under the `ignoreVoid`.

Runtime error messages might get improved as well, but as I'm on an older version of the extension (v6 if I see correctly), I still went to see the documentation to check if the usage is correct.

Code could be improved by requiring the `.catch(() => { ... })` even for voided Promises or by setting `{ ignoreVoid: false }` by default, as the behavior is potentially annoying.

Thank you

### Affected URL(s)

https://typescript-eslint.io/rules/no-floating-promises/

alexandercerutti:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9947

## Overview

Reworded some points of no-floating-promises and added a warn under `ignoreVoid` option.


packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -13,13 +13,13 @@ A "floating" Promise is one that is created without any code set up to handle an
...
-- `await`ing it
-- `return`ing it
-- `void`ing it
-- Calling its `.then()` with two arguments
-- Calling its `.catch()` with one argument
+- `await`ing it (❌ unhandled)
+- `return`ing it (❌ unhandled)
+- [`void`ing it](#ignorevoid) (❌ unhandled)
+- Calling its `.then()` with two arguments (✅ handled)
+- Calling its `.catch()` with one argument (✅ handled)
```
kirkwaiblinger, line 22: I understand why you'd want to have these `(❌ unhandled)`/`(✅ handled)` indicators, but I'm not sure they're an improvement here, without a deeper dive into unhandled rejections. 

The reason I say this is that code like `promise.catch(() => { throw new Error('unhandled!') });` _is_ still unhandled, but (intentionally) not flagged by the rule.

Furthermore, `await`ing or `return`ing promises _do_ prevent a unhandled rejections (assuming the caller is handled appropriately).

For now, my thought would be to leave these markers off (but maybe this can be revisited when we do futher docs improvements for the promise rules, which is already on our redar).

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -13,13 +13,13 @@ A "floating" Promise is one that is created without any code set up to handle an
 Floating Promises can cause several issues, such as improperly sequenced operations, ignored Promise rejections, and more.
 
 This rule reports when a Promise is created and not properly handled.
-Valid ways of handling a Promise-valued statement include:
+Valid ways of silencing a Promise-valued statement include:
```
kirkwaiblinger, line 16: What do you think of something more along these lines?

```suggestion
This rule will report Promise-valued statements that are not treated in one of the following ways:
```

The idea I'm getting at is to just say what's reported or not reported, and dodge unpacking every scenario that does or doesn't lead to unhandled rejection, and why the rule cannot report in 1:1 correspondence with those scenarios.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -13,13 +13,13 @@ A "floating" Promise is one that is created without any code set up to handle an
...
 
-- `await`ing it
-- `return`ing it
-- `void`ing it
-- Calling its `.then()` with two arguments
-- Calling its `.catch()` with one argument
+- `await`ing it (❌ unhandled)
+- `return`ing it (❌ unhandled)
+- [`void`ing it](#ignorevoid) (❌ unhandled)
```
kirkwaiblinger, line 20: (praise) Good call on adding the `#ignorevoid` link!

kirkwaiblinger, line 20: Maybe it would be good to call out here 
```suggestion
- [explicitly marking the promise as ignored by `void`ing it](#ignorevoid)
```

What do you think?

(and i personally would move it to the end of the list since it's an outlier in that it is just a way to disable the rule, not actually handle the promise, but up to you)

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -134,6 +134,12 @@ await createMyThenable();
 This option, which is `true` by default, allows you to stop the rule reporting promises consumed with void operator.
 This can be a good way to explicitly mark a promise as intentionally not awaited.
 
+:::warning
+Voiding a Promise doesn't handle it, but just ignores the outcome. Such Promise rejection will still lead to an
+[`unhandledrejection` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event)
+being fired and error logs being printed. Refer to the list at the top of the page for the possible ways to handle a promise.
+:::
```
kirkwaiblinger, line 141: (praise) I quite like that you've used the `:::warning` block here 🙂 

... but I feel like the wording wanders a bit off topic. What do you think of something like this?

```suggestion
:::warning
Voiding a promise doesn't handle it or change its behavior in any way. It is equivalent to an `// eslint-disable-next-line @typescript-eslint/no-floating-promises` comment.
:::
```

kirkwaiblinger (changes requested):
Thanks for sending this in! Sorry for leaving so many comments - the promise docs in general, and especially this rule, deal with quite a bit of subtle nuance.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: I see your point. I'd ask you why this is intentional, but I guess that would be pointless.

`await`ing and `return`ing do not prevent unhandled rejections but delegate the handling to, respectively a `try-catch` and the parent function.

As you surely have seen, these markers are meant also for the `ignoreVoid` section warning. So I wonder if there's another way to include such markers in a softer way, maybe by preparing it for the change you mentioned...

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: Yeah, I chose to just use "silencing" because of the issue I've opened before this PR.
But yeah, I like this.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: I'm neutral to this suggestion. I mean, using "`void`ing" is coherent with "`await`ing" and "`return`ing" wording, so I'm not actually sure.

About moving to the bottom, this is perfectly fine. I like it.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: Quoting the `disable-next-line` fact is great and I can add it. However, my idea was to exactly represent what happend to me and what I described in the issue (the code example).

So, I think it is important to include a link to mdn page about `unhandledrejection`, which can explain surely better what is it.

About the comment above, I think could be interesting to give a hint about how can a person handle a promise.

Josh-Cena:
I'll let Kirk take on this since apparently he has a lot of opinions while I have much fewer, but if either of you want to add anything to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#error_handling, and maybe eventually link to it on this doc page, I'd be glad to review a PR over there on MDN.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: I've removed the markers as I've improved the message in `ignoreVoid`.
Let me know if it is okay to you

alexandercerutti:
@Josh-Cena I've added a link to MDN page you linked in order to improve the warning explanation.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
kirkwaiblinger: > I'd ask you why this is intentional, but I guess that would be pointless

Hardly! 🙂 The short version is just that it's impossible to know that a function won't throw in JS, so, in reality, `.then()` and `.catch()` can never guaranteeably prevent a floating promise. 

For example, `promise.catch(() => maybeThrows())` can never be known not to throw, so _technically_ we could mark it as unhandled. But then, _everything_ would be unhandled! So instead, we say, `promise.catch(() => maybeThrows())` _looks_ like it's handled, so the rule should not report.

Similarly, for `promise.then(() => onFulfilled(), () => onRejected())`, we know that `onFulfilled` _could_ throw, and isn't any different than `promise.then(() => onFulFilled())`, which we _would_ mark as unhandled. However, because the `onRejected` handler is provided, we say that the usage looks like the developer was aware of rejections when they wrote the code, and, therefore, it it _looks_ handled. 

So, no-floating-promises really becomes about putting up reasonable guardrails, rather than attempting to make guarantees.

There are quite a few issues that discuss these tradeoffs.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -134,6 +133,18 @@ await createMyThenable();
...
 This can be a good way to explicitly mark a promise as intentionally not awaited.
 
+:::warning
+Voiding a Promise doesn't handle it or change the runtime behavior. The outcome is just ignored, like disabling the rule with a
+
+```
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+```
+
+Such Promise rejection will still lead to an [`unhandledrejection` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event)
```
kirkwaiblinger, line 143: Continuing from https://github.com/typescript-eslint/typescript-eslint/pull/9949#discussion_r1748907511,

> ... my idea was to exactly represent what happend to me and what I described in the issue (the code example).

Right, exactly. But the details of exactly what happened to you are a little specific imo since it's dependent on node vs web, and on whether a global promise rejection handler was or was not set up.

> So, I think it is important to include a link to mdn page about `unhandledrejection`, which can explain surely better what is it.
> 
> About the comment above, I think could be interesting to give a hint about how can a person handle a promise.

To be clear, I definitely agree with putting this info on this page! 🙂  I'm just thinking it isn't _specific_ to the `void` operator - it's essential knowledge whether or not you're using the `ignoreVoid` option - so I'd prefer it to be in the general info on handling promises, rather than in this particular warning block (which I wouldn't bother reading if i had the `ignoreVoid` option disabled).

Note that the MDN page link (albeit different section) is included already, down in the "Further Reading" section. Maybe we can find a better place to link the "error handling" section in the intro?

What do you think?

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -29,7 +28,7 @@ This rule also reports when an Array containing Promises is created and not prop
 - `Promise.race()`
 
 :::tip
-`no-floating-promises` only detects unhandled Promise _statements_.
+`no-floating-promises` only detects unhandled Promise _statements_ and voided ones.
```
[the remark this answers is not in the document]
alexandercerutti: Thanks for the explanation. I'm going to revert it

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: > Right, exactly. But the details of exactly what happened to you are a little specific imo since it's dependent on node vs web, and on whether a global promise rejection handler was or was not set up.

I just tested the difference between node and web. I see some differences I'm going to document here.

 - `unhandledrejection` (web) vs `unhandledRejection` (node)
 - `window` vs `process`
 - if the listener is set up, error log still prints on web, but doesn't on node

So, I see your point.

> so I'd prefer it to be in the general info on handling promises, rather than in this particular warning block (which I wouldn't bother reading if i had the ignoreVoid option disabled).

I agree. The link to `ignoreVoid` was serving that exact purpose, but it might not be enough. There is no section about handling promises, right?

The fact the MDN link is on the end section might not help making it being read. Personally I don't ever read to the end of a document. This behavior might be wrong, but I organize my document sections to give the important resources first.

I'm going to try to put down a few lines

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: The fact it "looks like it handled" could be a thing worth to be specified in a new paragraph about handling promises.

alexandercerutti:
@kirkwaiblinger I've written a paragraph about Promises handling and linked it to the `ignoreVoid` warning. Let me know what do you thing about it.

Once the changes are ready, I'll merge the base branch into this (again).

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -82,6 +83,24 @@ await Promise.all([1, 2, 3].map(async x => x + 1));
 </TabItem>
 </Tabs>
 
+## Handling Promises rejections
```
kirkwaiblinger, line 86: Oh, sorry. I feel I may have led you in the wrong direction here. I'm thinking, for now, for this PR, let's not add this section (sorry for the time you spent on this!). 

But it _would_ be in scope to prominently link to the MDN error-handling writeup, maybe by a simple `See [the MDN docs](link) for a detailed writeup on promise error-handling.` right below the list of ways to satisfy the rule?

kirkwaiblinger (changes requested):

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: I love writing things and explanations for people to read.

I studied this paragraph with the idea of it linking to MDN, not replacing it, but also to share the knowledge we discussed in here and adding some values related to this rule.

I think that removing such paragraph and reducing it to be just a link, would almost completely void (_pun intended_) the purpose of this PR, which I guess could be closed then.

We would be going to remove the whole concept of "looks like to be handled" and "guardrails" you highlighted before in this discussion, which are two concepts that I think are worth to me mentioned in terms of mental model.

I can move it below the list, for sure, and I can tune the paragraph a bit.

However, I'm of the idea I already spent too much energy on this hoping it would be okay and reducing it to a link makes me feel I wasted my time and energy. I hope you understand how I'm feeling after this comment.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -28,6 +27,8 @@ This rule also reports when an Array containing Promises is created and not prop
 - `Promise.any()`
 - `Promise.race()`
 
+Refer to [Handling Promises Rejections](#handling-promises-rejections) below for more information about this rule.
+
```
JoshuaKGoldberg, line 31: I don't see any real value to including this. The page is tall; people will know that scrolling down will give them more information.

alexandercerutti: I wouldn't make this assumption. I'm speaking for myself, but I often try to scan the page fastly and perform a search to get what I want. I don't read everything.

However, this is fine if I move the paragraph below the list, as suggested in the above.

alexandercerutti: I moved the paragraph up and removed the phrase. Let me know.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: I tried to rework the paragraph a little bit and moved it up in the discussion. Let me know how it feels now.

kirkwaiblinger: @alexandercerutti, I hear you. But, it's deliberate that the issue was scoped narrowly to be essentially about the `void` wording only, see https://github.com/typescript-eslint/typescript-eslint/issues/9947#issuecomment-2334294112. We have #8088 and #8517 already to address the larger scale efforts of improving the promise docs, deep-diving on the technical details, and ensuring that the various promise rules' docs are coherent with each other, rather than just this page alone.

So, we all agree that the deep-dive info on promise rejections needs to exist, but, unfortunately, I need to insist that this particular PR is not the time and place to make it happen, sorry.

We'll certainly keep this discussion in mind, though for when it does happen in the linked issues!

Best, Kirk

kirkwaiblinger (changes requested):
just need resolution on https://github.com/typescript-eslint/typescript-eslint/pull/9949#discussion_r1760523306

alexandercerutti:
@kirkwaiblinger done, removed. Should I remove even the warn?

kirkwaiblinger:
> @kirkwaiblinger done, removed. Should I remove even the warn?

No, the warn admonition is great! 👌

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -134,6 +133,16 @@ await createMyThenable();
...
 This can be a good way to explicitly mark a promise as intentionally not awaited.
 
+:::warning
+Voiding a Promise doesn't handle it or change the runtime behavior. The outcome is just ignored, like disabling the rule with a
+
+```
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+```
+
+Such Promise rejection will still be unhandled, as explained in the above section [_Handling Promises Rejections_](#handling-promises-rejections).
```
kirkwaiblinger, line 143:
```suggestion
Such Promise rejections will still be unhandled.
```

kirkwaiblinger (approved):
Great! Thanks for bearing with us on this.
celebratory gif

kirkwaiblinger:
(once another member gets a chance to review, they'll merge it in, and if all goes well this will go out with our next Monday release 🙂)

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -134,6 +135,16 @@ await createMyThenable();
...
 This can be a good way to explicitly mark a promise as intentionally not awaited.
 
+:::warning
+Voiding a Promise doesn't handle it or change the runtime behavior. The outcome is just ignored, like disabling the rule with a
+
+```
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+```
+
+Such Promise rejections will still be unhandled.
```
JoshuaKGoldberg, line 145: Lots of vertical space for a point that could be made with a link, no?

```suggestion
Voiding a Promise doesn't handle it or change the runtime behavior.
The outcome is just ignored, like disabling the rule with an [ESLint disable comment](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1).
Such Promise rejections will still be unhandled.
```

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
```diff
@@ -29,8 +28,10 @@ This rule also reports when an Array containing Promises is created and not prop
 - `Promise.race()`
 
 :::tip
-`no-floating-promises` only detects unhandled Promise _statements_.
+`no-floating-promises` only detects apparently unhandled Promise _statements_.
 See [`no-misused-promises`](./no-misused-promises.mdx) for detecting code that provides Promises to _logical_ locations such as if statements.
+
+See [_Using promises (error handling) on MDN_](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#error_handling) for a detailed writeup on promise error-handling.
```
JoshuaKGoldberg, line 34:
```suggestion
See [_Using promises (error handling) on MDN_](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#error_handling) for a detailed writeup on Promise error-handling.
```

JoshuaKGoldberg (approved):
LGTM! Just some small suggestions.

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: I think this is not mine, but I still accepted it

packages/eslint-plugin/docs/rules/no-floating-promises.mdx
alexandercerutti: Yes, why not

kirkwaiblinger (approved):

Josh-Cena (approved):

JoshuaKGoldberg (approved):
yay!

PR #9954, opened 2024-09-08
feat(eslint-plugin): [switch-exhaustiveness-check] add allowDefaultCaseMatchUnionMember option

### LINKED ISSUE #8247: Enhancement: [switch-exhaustiveness-check] Add ability to ignore default case

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.md

### Description

https://github.com/typescript-eslint/typescript-eslint/issues/3616#issuecomment-1673619717

This comment was never implemented. The case that is not possible to detect is case 1 from the original thread

This should throw an error because all cases are not enumerated (despite there being a fallthrough).

### Fail

```TypeScript
declare const literal: "a" | "b";

switch (literal) {
  case "a": break;
  default: break;
}
```

### Pass

```TypeScript
declare const literal: "a" | "b";

// Valid
switch (literal) {
  case "a": break;
  case "b": break;
}

// Valid
switch (literal) {
  case "a": break;
  case "b": break;
  default: break;
}
```

### Additional Info

Why is this useful? Say you want to define explicit behavior for each enum value. I then add an enum value to my type. Having an error in this case will be an extra catch that will allow me to ensure that I remember to update the switch statements. Essentially I want to be able to always define behavior for known types.

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8247

## Overview
- add allowDefaultCaseMatchUnionMember option to dismiss default case when decide exhaustiveness
- In this situation, even if there is a default case, an error must be added to the key that does not exist, so we changed some of the code modification logic.


packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -53,6 +53,27 @@ switch (value) {
 
 Since `value` is a non-union type it requires the switch case to have a default clause only with `requireDefaultForNonUnion` enabled.
 
+### `allowDefaultCaseMatchUnionMember`
+
+Defaults to true. If set to `true`, the `default` clause is used to determine whether the switch statement is Exhaustive for union type.
+
+In a switch case, if you enter the default statement, it can be considered that Exhaustive is satisfied, but if you think that entering the default statement did not solve the exhaustive problem of the union type, please set to option `true`.
```
JoshuaKGoldberg, line 60:
```suggestion
By default, if a `switch` statement over a union type includes a `default` case, then it's considered exhaustive.
`allowDefaultCaseMatchUnionMember` enforces explicitly handling every constituent of the union type with their own explicit `case`.
This can be useful if you want to make sure every value added to the union receives explicit handling, with the `default` case reserved for reporting an error.
```

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -36,6 +36,13 @@ type Options = [
      * @default false
      */
     requireDefaultForNonUnion?: boolean;
+
+    /**
+     * If `true`, the `default` clause is used to determine whether the switch statement is Exhaustive for union type.
+     *
+     * @default true
+     */
+    allowDefaultCaseMatchUnionMember?: boolean;
```
JoshuaKGoldberg, line 45: [Refactor] I'm not a fan of truthy defaults for unspecified values. The truthiness/not-defined-ness gets confusing. Proposal: let's have this be falsy by default?

```suggestion
    requireDefaultCaseForUnions?: boolean;
```

Also, tip: _"member"_ refers to methods/properties on a class or object. _"constituent"_ refers to the types in an intersection or union type.

packages/eslint-plugin/tests/rules/switch-exhaustiveness-check.test.ts
JoshuaKGoldberg, line 1: [Testing] Missing tests fo the new option, I think:

* Usage with enums
* Usage with a primitive, to make sure that works
* Usage with an object shape, to make sure that works

JoshuaKGoldberg (changes requested):
🆒

packages/eslint-plugin/tests/rules/switch-exhaustiveness-check.test.ts
developer-bandi: I wrote some test cases, but I'm not sure if they match what you said.

Especially in the case of object shape, I could not write a test case because I did not understand the case. 

Please check if the above test case is correct, and could you provide additional explanation or example code about the object shape?

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -155,10 +173,13 @@ export default createRule<Options, MessageIds>({
       const { missingLiteralBranchTypes, symbolName, defaultCase } =
         switchMetadata;
 
-      // We only trigger the rule if a `default` case does not exist, since that
-      // would disqualify the switch statement from having cases that exactly
+      // We only trigger the rule if a `default` case does not exist when requireDefaultCaseForUnions option
```
kirkwaiblinger, line 176: (clarity nit) 

Maybe just have a standalone condition like

```ts
// Unless requireDefaultCaseForUnions is enabled, the presence of a default case 
// always makes the switch exhaustive.
if (!requireDefaultCaseForUnions && defaultCase != null) {
    return;
}
```

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -243,6 +264,15 @@ export default createRule<Options, MessageIds>({
         .join('\n');
 
       if (lastCase) {
+        const isLastCaseDefaultCase = lastCase.test == null;
+        if (isLastCaseDefaultCase) {
```
kirkwaiblinger, line 268: (optional)

food for thought - what if the default case isn't the last case?

This change makes it so that we avoid suggesting to fix
```ts
declare const literal: 'a' | 'b';

switch (literal) {
  case 'b':
  default:
}
```
to 
```ts
declare const literal: 'a' | 'b';

switch (literal) {
  case 'b':
  default:
  case "a": { throw new Error('Not implemented yet: "a" case') }
}
```

and instead fix it to 
```ts
declare const literal: 'a' | 'b';

switch (literal) {
  case 'b':
  case "a": { throw new Error('Not implemented yet: "a" case') }
  default:
}
```
. (also note that this has its own issue of no longer having `case 'b'` be handled the same as `default`, instead having it throw as not implemented yet).

But it will happily fix 
```ts
declare const literal: 'a' | 'b';

switch (literal) {
  default:
  case 'b':
}
```
to
```ts
declare const literal: 'a' | 'b';

switch (literal) {
  default:
  case 'b':
  case "a": { throw new Error('Not implemented yet: "a" case') }
}
```
which still has the added case as part of the default 🤔, even though the first examples implies this code is trying to avoid it. Maybe to be more consistent we could say "if default case is present, add it above it, else add it at the end"?

***

I don't think that this is super important since the user will have to do manual cleanup regardless, but just throwing this out there in case you want to tweak some edge cases. I'm ok with either way though.

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -65,6 +72,10 @@ export default createRule<Options, MessageIds>({
             description: `If 'true', require a 'default' clause for switches on non-union types.`,
             type: 'boolean',
           },
+          requireDefaultCaseForUnions: {
+            description: `If 'true', the 'default' clause is used to determine whether the switch statement is Exhaustive for union type`,
```
kirkwaiblinger, line 76:
```suggestion
            description: `If 'true', the 'default' clause is used to determine whether the switch statement is exhaustive for union type`,
```

kirkwaiblinger (commented):
The code logic looks great to me, but some cleanup is needed for the option naming and docs.

`allowDefaultCaseMatchUnionMember` is ungrammatical, and `requireDefaultCaseForUnions` is not accurate, to my understanding.

Maybe `requireExplicitCasesForUnion` or `considerDefaultExhaustiveForUnion` depending on which direction the true/false goes?

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
kirkwaiblinger, line 1: The line 
```md
Additionally, if a new value is added to the union type, a `default` would prevent the `switch-exhaustiveness-check` rule from reporting on the new case not being handled in the `switch` statement.
```
in the `allowDefaultCaseForExhaustiveSwitch` section should be updated, since this option will address that issue

kirkwaiblinger:
(Further thoughts on naming and options) - 

My understanding is that there are two independent things going on with union types and defaults.

1 - When _all_ union constituents _do_ have their own explicit case, whether a `default` case is prohibited|ignored|required. This is currently controlled by the existing `allowDefaultCaseForExhaustiveSwitch` option, and the only possibilities currently supported are prohibited|ignored (though I think you can use other lint rules to enforce the default case _must_ be present).
2 - When some union constituents _do not_ have their own explicit case, whether a `default` case will silence the rule. This is the new option being created.

Perhaps thinking about it in this framework can help clarify the docs and the naming.

developer-bandi:
As for the option name, I would suggest `considerDefaultExhaustiveForUnion` as suggested.

It seems that including both `Default` and `Exhaustive` keywords is more appropriate in the context you explained.

+ Once the option name is decided, related modifications will be reflected.

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -22,6 +22,7 @@ Defaults to true. If set to false, this rule will also report when a `switch` st
 
 When a `switch` statement over a union type is exhaustive, a final `default` case would be a form of dead code.
 Additionally, if a new value is added to the union type, a `default` would prevent the `switch-exhaustiveness-check` rule from reporting on the new case not being handled in the `switch` statement.
+If you want `default` to prevent `switch-exhaustiveness-check` when new values ​​are added to the union type, try using the `requireDefaultCaseForUnions` option
```
kirkwaiblinger, line 25: how about
```suggestion
Additionally, if a new value is added to the union type and you're using `[considerDefaultExhaustiveForUnions](#considerDefaultExhaustiveForUnions)`, a `default` would prevent the `switch-exhaustiveness-check` rule from reporting on the new case not being handled in the `switch` statement.
```

or just omitting this check now that there is a recourse 🤷

kirkwaiblinger:
looks pretty close. Let's just change the option name to `considerDefaultExhaustiveForUnions` and if anyone else has a conflicting opinion they can object before it gets merged 👍

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
developer-bandi: It seemed better to change it like the first suggestion, so we reflected it.

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -53,6 +53,27 @@ switch (value) {
 
 Since `value` is a non-union type it requires the switch case to have a default clause only with `requireDefaultForNonUnion` enabled.
 
+### `considerDefaultExhaustiveForUnions`
+
+Defaults to false. Whether a `default` clause will be considered as making a switch statement over a union type exhaustive.
+
+By default, if a `switch` statement over a union type includes a `default` case, then it's considered exhaustive.
```
kirkwaiblinger, line 60: this is the opposite way (since the option defaults to false). By default, every branch will be required to be explicitly handled, and one may opt out by enabling `considerDefaultExhaustiveForUnions`

kirkwaiblinger (changes requested):
Just the nit in https://github.com/typescript-eslint/typescript-eslint/pull/9954/files#r1801500418

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
developer-bandi: thank you I think I missed changing it at the same time when changing the default value.

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -53,6 +53,27 @@ switch (value) {
 
 Since `value` is a non-union type it requires the switch case to have a default clause only with `requireDefaultForNonUnion` enabled.
 
+### `considerDefaultExhaustiveForUnions`
+
+Defaults to false. Whether a `default` clause will be considered as making a switch statement over a union type exhaustive.
+
+If set to true, `switch` statement over a union type includes a `default` case is considered exhaustive.
+`considerDefaultExhaustiveForUnions` enforces explicitly handling every constituent of the union type with their own explicit `case`.
+This can be useful if you want to make sure every value added to the union receives explicit handling, with the `default` case reserved for reporting an error.
```
kirkwaiblinger, line 62: https://github.com/typescript-eslint/typescript-eslint/pull/9954#discussion_r1801500418 applies to the other lines here too...

```suggestion
If set to true, a `switch` statement over a union type that includes a `default` case is considered exhaustive.
Otherwise, the rule enforces explicitly handling every constituent of the union type with their own explicit `case`.
Keeping this option disabled can be useful if you want to make sure every value added to the union receives explicit handling, with the `default` case reserved for reporting an error.
```

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -53,6 +53,27 @@ switch (value) {
...
+
+Examples of additional **incorrect** code for this rule with `{ considerDefaultExhaustiveForUnions: true }`:
```
kirkwaiblinger, line 64:
```suggestion
Examples of additional **correct** code for this rule with `{ considerDefaultExhaustiveForUnions: true }`:
```

kirkwaiblinger (changes requested):

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
developer-bandi: I thought it would be good to provide detailed example cases, so I added a few cases each correct and incorrect.

kirkwaiblinger (dismissed):
Good stuff! I'm super excited to get this change released! This has been a pain point for me as a user of the rule, and I'm glad to see it addressed 🙂 

FYI - I've made a few edits to your branch primarily in order to get the website to build (we added in the `{/* insert option description */}` thing recently, which prevented the CI from building).

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
kirkwaiblinger: Makes sense 👍  I just went ahead and refactored this to our typical tabs setup

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -23,7 +23,7 @@ This rule reports when a `switch` statement over a value typed as a union of lit
 If set to false, this rule will also report when a `switch` statement has a case for everything in a union and _also_ contains a `default` case. Thus, by setting this option to false, the rule becomes stricter.
 
 When a `switch` statement over a union type is exhaustive, a final `default` case would be a form of dead code.
-Additionally, if a new value is added to the union type, a `default` would prevent the `switch-exhaustiveness-check` rule from reporting on the new case not being handled in the `switch` statement.
+Additionally, if a new value is added to the union type and you're using `[considerDefaultExhaustiveForUnions](#considerDefaultExhaustiveForUnions)`, a `default` would prevent the `switch-exhaustiveness-check` rule from reporting on the new case not being handled in the `switch` statement.
```
JoshuaKGoldberg, line 26:
```suggestion
Additionally, if a new value is added to the union type and you're using [`considerDefaultExhaustiveForUnions`](#considerDefaultExhaustiveForUnions), a `default` would prevent the `switch-exhaustiveness-check` rule from reporting on the new case not being handled in the `switch` statement.
```

JoshuaKGoldberg (dismissed):
LGTM :)

PR #9972, opened 2024-09-12
feat: support TypeScript 5.6

### LINKED ISSUE #9653: TypeScript 5.6 Support
Edit(JoshuaKGoldberg): We're currently blocked on https://github.com/microsoft/rushstack/issues/4919

We haven't had the bandwidth to investigate and try helping them fix it. If anyone is so motivated - fixing that would unblock us and allow us to support 5.6.

---

https://devblogs.microsoft.com/typescript/announcing-typescript-5-6-beta/

Please read through our TypeScript version policy: https://typescript-eslint.io/users/dependency-versions#typescript

_Please be patient._

---

# 🔗 Disallowed Nullish and Truthy Checks: #9973

> Moved to issue #9973. This does not block supporting TypeScript 5.6.

As they mention in the post - this is essentially a weaker version of [`no-constant-binary-expression`](https://eslint.org/docs/latest/rules/no-constant-binary-expression). There will be a lot of overlap between the rules.

Perhaps we should create an extension rule to remove the duplicates so users can have the rest of the rule without the duplicate reports?

# 🏗 Support for Arbitrary Module Identifiers: #9972

We will need to update our AST types as per [the estree spec](https://github.com/estree/estree/blob/3851d4a6eae5e5473371893959b88b62007469e8/es2022.md#exportspecifier).

# 🏗 Search Ancestor Configuration Files for Project Ownership: #9972

# 🏗 lib.d.ts Updates: #9972

- Iterator Helper Methods
- Strict Builtin Iterator Checks (and --strictBuiltinIteratorReturn)
- The --noUncheckedSideEffectImports Option
- The --noCheck Option
- Allow --build with Intermediate Errors
- Region-Prioritized Diagnostics in Editors
- Notable Behavioral Changes
    - .tsbuildinfo is Always Written
    - Respecting File Extensions and package.json from within node_modules
    - Correct override Checks on Computed Properties

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9653

## Overview

Co-authored-by: @jakebailey


packages/ast-spec/src/special/ExportSpecifier/spec.ts
```diff
@@ -2,9 +2,10 @@ import type { AST_NODE_TYPES } from '../../ast-node-types';
 import type { BaseNode } from '../../base/BaseNode';
 import type { ExportKind } from '../../declaration/ExportAndImportKind';
 import type { Identifier } from '../../expression/Identifier/spec';
+import type { StringLiteral } from '../../expression/literal/StringLiteral/spec';
 
 export interface ExportSpecifier extends BaseNode {
-  exported: Identifier;
+  exported: Identifier | StringLiteral;
   exportKind: ExportKind;
   local: Identifier;
```
bradzacher, line 10: You have missed one case here

```suggestion
  local: Identifier | StringLiteral;
```

However note from [the estree spec](https://github.com/estree/estree/blob/3851d4a6eae5e5473371893959b88b62007469e8/es2022.md#exportspecifier)

> local can be Literal only if the source of the ExportNamedDeclaration of the parent of this node is not null. e.g. export { "foo" as "foo" } from "mod" is valid, export { "foo" as "foo" } is invalid.

So we should create two separate `ExportSpecifier`'s:
- One for the non-`source` export that only allows `local: Identifier`
- One for the `source` export that allows `local: Identifier | StringLiteral`

packages/ast-spec/src/special/ImportSpecifier/spec.ts
```diff
@@ -2,9 +2,10 @@ import type { AST_NODE_TYPES } from '../../ast-node-types';
 import type { BaseNode } from '../../base/BaseNode';
 import type { ImportKind } from '../../declaration/ExportAndImportKind';
 import type { Identifier } from '../../expression/Identifier/spec';
+import type { StringLiteral } from '../../expression/literal/StringLiteral/spec';
 
 export interface ImportSpecifier extends BaseNode {
-  imported: Identifier;
+  imported: Identifier | StringLiteral;
```
bradzacher, line 8: So there is one problem with this definition:
```ts
import { "😭" } from 'module';
```

This is parser valid TS code, but is invalid ESTree AST code.

So we need to ban this in the converter.

bradzacher (changes requested):
could you also add some tests for the import/export changes please?

packages/ast-spec/package.json
```diff
@@ -48,7 +48,7 @@
     "@babel/core": "*",
     "@babel/eslint-parser": "*",
     "@jest/types": "29.6.3",
-    "@microsoft/api-extractor": "^7.43.0",
+    "@microsoft/api-extractor": "^7.47.8",
```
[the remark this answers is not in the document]
bradzacher: angry_bear

packages/ast-spec/src/special/ImportSpecifier/spec.ts
JoshuaKGoldberg: Is it valid in the TS parser? I'm getting errors. https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgIkLwbhaPeXAvnAZlBCHAOQgQAmArgDYCmJA3EA

packages/ast-spec/src/special/ExportSpecifier/spec.ts
JoshuaKGoldberg: I started on this in ced3289ca, but paused. These nodes don't have a `parent` in them. So is there really any difference on the `ast-spec` side?

JoshuaKGoldberg:
I started on tests in ced3289ca, but am roughly waiting on api-extractor to be patched.

packages/ast-spec/src/special/ExportSpecifier/spec.ts
bradzacher: Yeah the difference is on the `ExportDeclaration` side of things where we can specify via the types and the conversion logic that an export without a source cannot have a specifier with a literal local.

packages/ast-spec/package.json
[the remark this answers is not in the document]
jakebailey: https://github.com/microsoft/rushstack/issues/4919#issuecomment-2412077019 contains a patch which I believe would unblock you.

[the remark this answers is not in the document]
jakebailey: https://github.com/microsoft/rushstack/pull/4973 is the actual code change.

JoshuaKGoldberg:
Just https://github.com/typescript-eslint/typescript-eslint/pull/9972#discussion_r1758529218 is left here, hoping to tackle tomorrow

packages/ast-spec/src/special/ExportSpecifier/spec.ts
JoshuaKGoldberg: Hmm, I can't find any cases that TS doesn't directly error on. Am I missing something?

packages/ast-spec/src/special/ImportSpecifier/spec.ts
bradzacher: You're getting a *_SEMANTIC_* error, not a *_SYNTACTIC_* error.
The code parses fine and then TS emits a semantic error.

So we need to check for this case and throw our own error.

packages/ast-spec/src/special/ExportSpecifier/spec.ts
bradzacher: You're likely getting **_SEMANTIC_** errors, not **_SYNTACTIC_** errors.

Syntactic errors are invariants and will crash out the TS parser -- i.e. we can just assume they'll crash and be happy.
Semantic errors will parse just fine and will not crash out the TS parser -- which means we have to throw our own errors.

bradzacher (approved):

bradzacher:
@JoshuaKGoldberg if you're happy with the changes in [`c49e85f` (#9972)](https://github.com/typescript-eslint/typescript-eslint/pull/9972/commits/c49e85fddae12b4351926f2aa77614cdf932e092) then feel free to land this!

.yarn/patches/@microsoft-api-extractor-npm-7.47.9-d69a694fc6.patch
```diff
@@ -0,0 +1,44 @@
+diff --git a/lib/analyzer/ExportAnalyzer.js b/lib/analyzer/ExportAnalyzer.js
```
jakebailey, line 1: I believe this change was released, so the patch shouldn't be required anymore.

.prettierignore
JoshuaKGoldberg, line 1: Yeah, I as thinking of cleaning up the old globs here. Nice.

JoshuaKGoldberg (commented):
Great! (approving)

Issue #9828, opened 2024-08-19
Bug: [no-unnecessary-type-parameters] false positive when used in the implementation of the method


ST-DDT:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.4&fileType=.ts&code=KYDwDg9gTgLgBAMwK4DsDGMCWEVxsAZxgB4AVOUfFAEwLiKkxQHMA%2BACgFtgYALCagC44pAJTCGTZnADeAWABQcOGhxE4AQyjM6AXjgByA5rqkA3IuVQeSKLm58BcANSbtBCwoC%2BQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=N4XyA&tokens=false

### Repro Code

```TypeScript
export function test<T extends string>(method: T): string {
  const args = '' as T;
  return method + args;
}
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/no-unnecessary-type-parameters": "error"
  }
}
```

### tsconfig

```jsonc
{}
```

### Expected Result

No Error

### Actual Result

Type parameter T is used only once in the function signature.eslint[@typescript-eslint/no-unnecessary-type-parameters](https://typescript-eslint.io/rules/no-unnecessary-type-parameters)

### Additional Info

Original Code: https://github.com/faker-js/faker/blob/cc7715bebded681655b95f5afc368cb250a2eadf/test/support/seeded-runs.ts#L172-L192

The error was first observed in v8.1.0 it does not appear in 7.18.0

The rule seems to ignore any usage of the type inside the actual implementation.

JoshuaKGoldberg:
> The rule seems to ignore any usage of the type inside the actual implementation

[...]

PR #9975, opened 2024-09-12
docs: [no-unnecessary-type-parameters] add FAQ section

### LINKED ISSUE #9680: Docs: [no-unnecessary-type-parameters] add FAQ for some tricky cases

### Suggested Changes

NUTP is an incredibly subtle rule - huge shoutout to @danvk and @JoshuaKGoldberg for their work to implement it!

From testing out this rule, some of the cases that it reports are unusually tricky to fix, compared to most lint rules. In particular, while fixing the reported function signature is generally straightforward, it can have unexpected impacts at the usage sites.

Let's document some of these subtleties on the rule page, to help users out. 

Below is one case that I've had trouble with... I intend to update this post with a few more when I get around to it. I would also welcome anyone else to document tricky cases here!

<details>
<summary>Excess property errors at call sites</summary>
In the following example, the generic property appears unnecessary, and is flagged by the rule.

```ts
interface SomeProperties {
   requiredProperty1: string;
   requiredProperty2: number;
}

function serialize<T extends SomeProperties>(x: T): string {
//                 ~~~~~~~~~~~~~~~~~~~~~~~~
    return JSON.stringify(x);
}
```

The obvious fix would be to replace the generic with its constraint, like so 
```ts
function serialize(x: SomeProperties): string {
    return JSON.stringify(x);
}
```

However, this is a (slightly) breaking change for consumers! (affecting the type system only, not runtime behavior)

Before the change, this usage is just fine:
```ts
serialize({ 
   requiredProperty1: 'foo',
   requiredProperty2: 123,
   excessProperty: 'no problem'
});
```

But afterwards, we get a TS error that `excessProperty` does not exist on type `SomeProperties` (this happens specifically because we're supplying an object literal as the argument; a variable won't trigger this check). And, often, that's the right behavior! There is usually no need for excess properties.

But, sometimes (as, possibly, in the example given here), we actually did want to allow excess properties, and perhaps that's even why the original author wrote the function that way. 

In order to preserve that behavior, the fix is to reflect reasonable excess properties in the index signature of `SomeProperties`. Changing it to 

```ts
interface SomeProperties {
   requiredProperty1: string;
   requiredProperty2: number;
   [anyOtherProperties: string | number | symbol]: unknown;
}
```

we get our original, permissive behavior back. 

Deciding whether to fix the function signature type or the call sites is a case-by-case matter.
</details>

### Affected URL(s)

https://typescript-eslint.io/rules/no-unnecessary-type-parameters/#FAQ

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9680 

## Overview

Add FAQ section to rule's doc page.

Tried to mention all/most of the issues linked to #9680 in some capacity.

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-type-parameters.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-type-parameters.mdx
@@ -19,7 +19,7 @@ At best unnecessary type parameters make code harder to read.
 At worst they can be used to disguise unsafe type assertions.
 
 :::warning
-This rule was recently added, and has a surprising amount of hidden complexity compared to most of our rules. If you encounter unexpected behavior with it, please check closely the [Limitations](#limitations) section below and our [issue tracker](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+no-unnecessary-type-parameters).
+This rule was recently added, and has a surprising amount of hidden complexity compared to most of our rules. If you encounter unexpected behavior with it, please check closely the [Limitations](#limitations) and [FAQ](#faq) sections below and our [issue tracker](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+no-unnecessary-type-parameters).
 If you don't see your case covered, please [reach out to us](https://typescript-eslint.io/contributing/issues)!
 :::
 
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-type-parameters.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-type-parameters.mdx
@@ -181,8 +181,8 @@ type Compute<A> = A extends Function ? A : { [K in keyof A]: Compute<A[K]> };
 type Equal<X, Y> =
   (<T1>() => T1 extends Compute<X> ? 1 : 2) extends
     (<T2>() => T2 extends Compute<Y> ? 1 : 2)
-      ? true
-      : false;
+  ? true
+  : false;
 ```
 
 Use eslint-disable comments as appropriate to suppress the rule in these cases.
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/docs/rules/no-unnecessary-type-parameters.mdx
```diff
@@ -87,6 +87,108 @@ This is because the type parameter `T` relates multiple methods in the `T[]` tog
...
 That includes type arguments given to global types such as `Array` (including the `T[]` shorthand and in tuples), `Map`, and `Set`.
 
+## FAQ
+
+### I'm using the type parameter inside the function, so why is the rule reporting?
+
+You might be surprised to that the rule reports on a function like this:
+
+```ts
+function f<T extends string>(string1: T): void {
```
kirkwaiblinger, line 97: I would love if someone has a better example for this. I didn't just copy the example from https://github.com/typescript-eslint/typescript-eslint/issues/9828 since it didn't seem very organic, but I'm not sure this is any better 🤔

packages/eslint-plugin/docs/rules/no-unnecessary-type-parameters.mdx
```diff
@@ -87,6 +87,111 @@ This is because the type parameter `T` relates multiple methods in the `T[]` tog
...
+That is to say, the function can always be rewritten as:
+
+```ts
+function log(string1: string): void {
+  const string2: string = string1;
+  console.log(string2);
+}
+```
+
+Therefore, this rule only counts usages of a type parameter in the _signature_ of a function, method, or class, but not in the implementation. See also [#9735](https://github.com/typescript-eslint/typescript-eslint/issues/9735)
```
JoshuaKGoldberg, line 114: I'm guessing users will be confused by implicit return types:

```ts
// "wait, this is only used in the body, not in the signature?"
function identity<T>(value: T) {
  return value;
}
```

Do you think it would be useful to mention those?

JoshuaKGoldberg (approved):
LGTM so far! Just waiting on the todo note.

packages/eslint-plugin/docs/rules/no-unnecessary-type-parameters.mdx
kirkwaiblinger: good idea. added an entry

kirkwaiblinger:
> Just waiting on the todo note.

I was thinking to defer that to a followup once the linked conversation is fully resolved. Thoughts?

JoshuaKGoldberg (approved):
Makes sense to me. Let's do it!

PR #9989, opened 2024-09-14
fix(typescript-estree): don't throw on missing tsconfig.json by default in project service

### LINKED ISSUE #9985: Bug: Upgrade to 8.5.0 breaks linting setup with NX and `projectService: true`

### Issue Description

After upgrading `typescript-eslint` to the latest version `8.5.0` my linting setup with NX broke. 

The respective error message is `Parsing error: Could not read project service default project 'tsconfig.json': error TS5012: Cannot read file 'tsconfig.json': ENOENT: no such file or directory`. The respective files are included in a `tsconfig.lib.json` file which is being referenced from `tsconfig.json`. The same applies to files included by `tsconfig.spec.ts` (also referenced by `tsconfig.json`).

I am using `projectService: true` with a NX monorepo (see reproduction repo). The same setup worked fine with version `8.4.0`. The breaking change seems to be #9893.

https://github.com/hoi4/typescript-eslint-issue

### Repro Steps

1) clone the repo
2) `npm i`
3) `npx nx lint nx19 --skip-nx-cache`

Alternatively, when using `eslint` directly with an example file: `npx eslint apps/nx19/src/app/app.component.ts`

The latest commit uses version `8.5.0` which has the described issue. The previous commit holds a working state with version `8.4.0`. 

### Versions

| `@typescript-eslint/eslint-plugin`     | `8.5.0` |
| `@typescript-eslint/parser`            | `8.5.0` |
| `@typescript-eslint/utils`             | `8.5.0` |
| `TypeScript`                           | `~5.5.2` |
| `ESLint`                               | `9.10.0` |
| `node`                                 | `20.15.0` |

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9985

## Overview

Changes error throwing to only occur if the user explicitly provides a `parserOptions.defaultProject`.

💖

### DIFF
--- a/packages/typescript-estree/src/create-program/createProjectService.ts
+++ b/packages/typescript-estree/src/create-program/createProjectService.ts
@@ -42,9 +42,10 @@ export function createProjectService(
   jsDocParsingMode: ts.JSDocParsingMode | undefined,
   tsconfigRootDir: string | undefined,
 ): ProjectServiceSettings {
+  const optionsRawObject = typeof optionsRaw === 'object' ? optionsRaw : {};
   const options = {
     defaultProject: 'tsconfig.json',
-    ...(typeof optionsRaw === 'object' && optionsRaw),
+    ...optionsRawObject,
   };
   validateDefaultProjectForFilesGlob(options.allowDefaultProject);
 
@@ -126,7 +127,7 @@ export function createProjectService(
   });
 
   log('Enabling default project: %s', options.defaultProject);
-  let configFile: ts.ParsedCommandLine;
+  let configFile: ts.ParsedCommandLine | undefined;
 
   try {
     configFile = getParsedConfigFile(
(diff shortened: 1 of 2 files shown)


auvred (approved):
:+1:

ex0ns:
I have been troubleshooting this issue on my side without realizing there was an open issue/pr to already fix that. I was checking about the fix and wondering about the `if` inside the try, that involve knowing the subtle difference between the `config` and `optionRawObject`, it also feels a bit strange to parse the config file that might not exist and then wait for the exception.

I don't want to overstep, but I have tested the following solution (closer to what was done in 8.4)

```typescript
  if (options.allowDefaultProject?.length) {
    log('Enabling default project: %s', options.defaultProject);
    let configFile: ts.ParsedCommandLine;

    try {
      configFile = getParsedConfigFile(
        tsserver,
        options.defaultProject,
        tsconfigRootDir,
      );
    } catch (error) {
      throw new Error(
        `Could not read project service default project '${options.defaultProject}': ${(error as Error).message}`,
      );
    }

    service.setCompilerOptionsForInferredProjects(
      // NOTE: The inferred projects API is not intended for source files when a tsconfig
      // exists.  There is no API that generates an InferredProjectCompilerOptions suggesting
      // it is meant for hard coded options passed in. Hard asserting as a work around.
      // See https://github.com/microsoft/TypeScript/blob/27bcd4cb5a98bce46c9cdd749752703ead021a4b/src/server/protocol.ts#L1904
      configFile.options as ts.server.protocol.InferredProjectCompilerOptions,
    );
  }
```

Which will only try to parse the config if we actually want to use a defaultProject.

bradzacher (approved):

PR #9990, opened 2024-09-14
fix(eslint-plugin): [no-unnecessary-type-parameters] consider type parameter used as rest parameter type to be used multiple times

### LINKED ISSUE #9890: Bug: [no-unnecessary-type-parameters] Generic required for curried function params

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.4&fileType=.tsx&code=MYewdgzgLgBA5gUzAgTgS2AERAORFACzTDhgF4BYAKBhgAoBKcgPmttoB4BBFOCGBAA8oSACb8UCAIajwAGwCeMKWAUBtALrM6AOj1TeEAFwwefBiYBuINKJYwA3m3a0A9ACoYsmGHxESMO6uzrQAvgDc1NSgkLDAUlDABFxycth4hMSklDT0TGSsubS6%2BoYmKgoWMNa29k5F7B5eID5%2BWYHBDRFRVKIIwHIGCDAAZgCuYMBQaOCjICgAolJJHAAqAsJiEtKyYIrKqprazvGpAEbLANYmdGgiALYmq2pgY-dnqBr5zNU2ogA01CqdFECSkT2%2BMFWkSo0XA0BgmC4qy45BgagARAQEKkQBj-jAAIwAJgAzASoCgxggNMp%2BDFoDC4bEYFAFAAHBAAeRGmDBAEkIABhAzoBCiLmWVBokbzJZJOiIZDoLC4NokRgMOhIlEMGGuVwuAB6AH4egzYGzOTy%2BVApIKADIgBFkOaLZYEOjxRLJVLpdVwTXa5FcPXUA3Gs1UIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
const genericDoNothing =
  () =>
    <Args extends readonly any[]>(...args: Args): void => {
      /* do nothing */
    };

const catchAllDoNothing =
  () =>
    (...args: any): void => {
      /* do nothing */
    };

declare function forEach<T extends readonly any[]>(
  callback: (item: T[number]) => void,
): (data: T) => T;

const DATA = ["hello", 123, true] as const;

const typeOfDataIsCarriedOver = forEach(genericDoNothing())(DATA);
//    ^? const typeOfDataIsCarriedOver: readonly ["hello", 123, true]

const typeOfDataIsLost = forEach(catchAllDoNothing())(DATA);
//    ^? const typeOfDataIsLost: readonly any[]
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### tsconfig

_No response_

### Expected Result

I think the rule works correctly here, but I don't think there's another way to type this without losing the implicit inference property. Maybe this is a bug in TypeScript? or maybe it's a limitation of the typing system?

### Actual Result

I can't make the typing work without the rule firing too.

### Additional Info

This case is vaguer to me, and I'm not sure the rule is incorrect here; there's just no way around it. But if there is no way around it, maybe the rule shouldn't surface it as an error?

To understand the issue, let's assume that `forEach` is not a function in my code base/project/that I control, and I'm only providing the `doNothing` function.

The issue is solvable by adding `NoInfer` around the param `T` in `forEach`'s callback. But if I don't control that function, shipping a "fixed" version of `doNothing` without the `Args` generic will break our user's code. 

More context here: https://github.com/remeda/remeda/pull/801 and https://github.com/remeda/remeda/blob/main/src/doNothing.ts

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9890

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
@@ -167,6 +167,9 @@ ruleTester.run('no-unnecessary-type-parameters', rule, {
         };
       }
     `,
+    'function fn<T extends unknown[]>(...args: T) {}',
+    'function fn<T extends unknown[]>(cb: (...args: T) => void) {}',
+    'function fn<U, T extends U[]>(firstArg: U, ...restArgs: T) {}',
     `
       function lengthyIdentity<T extends { length: number }>(x: T) {
         return x;
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
@@ -5,7 +5,12 @@ import * as tsutils from 'ts-api-utils';
 import * as ts from 'typescript';
 
 import type { MakeRequired } from '../util';
-import { createRule, getParserServices, nullThrows } from '../util';
+import {
+  createRule,
+  getParserServices,
+  isRestParameterDeclaration,
+  nullThrows,
+} from '../util';
 
 type NodeWithTypeParameters = MakeRequired<
   ts.SignatureDeclaration | ts.ClassLikeDeclaration,
@@ -373,7 +378,10 @@ function collectTypeParameterUsageCounts(
     }
 
     for (const parameter of signature.parameters) {
-      visitType(checker.getTypeOfSymbol(parameter), false);
+      const isRestParameter = (parameter.getDeclarations() ?? []).every(
+        declaration => isRestParameterDeclaration(declaration),
+      );
+      visitType(checker.getTypeOfSymbol(parameter), isRestParameter);
     }
 
     for (const typeParameter of signature.getTypeParameters() ?? []) {
--- a/packages/website/src/globals.d.ts
+++ b/packages/website/src/globals.d.ts
@@ -3,9 +3,6 @@ import type * as ts from 'typescript';
 
 declare global {
   interface WindowRequire {
-    // We know it's an unsafe assertion. It's for window.require usage, so we
-    // don't have to use verbose type assertions on every call.
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-parameters
     <T extends unknown[]>(
       files: string[],
       success?: (...arg: T) => void,


packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
```diff
@@ -167,6 +167,9 @@ ruleTester.run('no-unnecessary-type-parameters', rule, {
         };
       }
     `,
+    'function fn<T extends unknown[]>(...args: T) {}',
```
JoshuaKGoldberg, line 170: Oh 😕 maybe I was wrong to suggest _any_ type parameter in https://github.com/typescript-eslint/typescript-eslint/issues/9890#issuecomment-2321307729... I think this might be pretty directly a case where the rule should report.

Perhaps a better delineation would be rest type parameters that are used in a return statement?

* Not ok: `declare function fn<T extends unknown[]>(...args: T): void`
* Ok: `declare function fn<T extends unknown[]>(): (...args: T) => void`

auvred: Yeah, this rule is very controversial! :slightly_smiling_face: 

> rest type parameters that are used in a return statement?

Hmmm, could you please explain why

```ts
declare function fn<T extends unknown[]>(): (...args: T) => void
```

should be considered valid? Maybe I'm missing something


The type `T` cannot be inferred from the type returned by the function, so this function is intended to be used with explicitly specified type arguments?

Looks like a continuation of the https://github.com/typescript-eslint/typescript-eslint/pull/9536#discussion_r1713663117 thread

JoshuaKGoldberg:
Switching to draft per discussion in #9890. Sorry @auvred 😅

kirkwaiblinger:
Gonna close this, since we ended up closing the backing issue #9890

PR #9994, opened 2024-09-16
fix(eslint-plugin): [no-confusing-non-null-assertion] check !in and !instanceof

### LINKED ISSUE #8795: Enhancement: [no-confusing-non-null-assertion] should check !instanceof

https://typescript-eslint.io/rules/no-confusing-non-null-assertion/

### Description

I propose that the rule should flag on `x !instanceof y`, (and, also, IMO on `x! instanceof y`, but this part is more debatable).

### Fail

```TypeScript
declare const x: any;
// logs false
console.log(x !instanceof SyntaxError);
// also logs false, same AST
console.log(x! instanceof SyntaxError);
```

### Pass

```TypeScript
declare const x: any;
// logs true
console.log(!(x instanceof SyntaxError));
```

### Additional Info

I have encountered this twice independently in the wild at work.
Perhaps part of the confusion comes from the fact that in some other languages, boolean operators can behave more fluently, for example, in Python:
```python
zero = 0
zero is not False # True, equivalent to not (zero is False)
zero is (not False) # False
```
so it's intuitive to think that one could write 
```typescript
x !instanceof y
```
to mean 
```typescript
!(x instanceof y)
```
and not realize that TS is doing you dirty by permitting that syntax at all (which is not valid JS).

Note that prettier autocorrects to the `x! instanceof` form instead of `x !instanceof` form. But either way it's yikes.

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8795 

## Overview

Basically just shares the logic from the other operators. The only difference is that, since it's not _always_ safe to remove the non-null assertion before an `in` or `instanceof`, I additionally added the wrapping suggestion for `a! in b` -> `(a!) in b` and `a! instanceof b` -> `(a!) instanceof b`.

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-confusing-non-null-assertion.mdx
+++ b/packages/eslint-plugin/docs/rules/no-confusing-non-null-assertion.mdx
@@ -9,13 +9,26 @@ import TabItem from '@theme/TabItem';
 >
 > See **https://typescript-eslint.io/rules/no-confusing-non-null-assertion** for documentation.
 
-Using a non-null assertion (`!`) next to an assign or equals check (`=` or `==` or `===`) creates code that is confusing as it looks similar to a not equals check (`!=` `!==`).
+Using a non-null assertion (`!`) next to an assignment or equality check (`=` or `==` or `===`) creates code that is confusing as it looks similar to an inequality check (`!=` `!==`).
 
 ```typescript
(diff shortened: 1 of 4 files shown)


JoshuaKGoldberg (dismissed):
Swell! 🙌

PR #10003, opened 2024-09-16
chore(plugin): [prefer-readonly-parameter-types] remove null from comparison when null does not exist

Zamiell:
## PR Checklist

## Overview

This line is written in a confusing way because the type of `actualParam.typeAnnotation` is `TSESTree.TSTypeAnnotation | undefined`, and thus does not contain `null` at all.

I think an audit of all such confusing cases in the monorepo should be done in the future. The root of the issue is the usage of the "==" operator, which in my opinion should just be banned, but maybe that is too controversial.

### DIFF
--- a/packages/eslint-plugin/src/rules/prefer-readonly-parameter-types.ts
+++ b/packages/eslint-plugin/src/rules/prefer-readonly-parameter-types.ts
@@ -117,7 +117,7 @@ export default createRule<Options, MessageIds>({
               ? param.parameter
               : param;
 
-          if (ignoreInferredTypes && actualParam.typeAnnotation == null) {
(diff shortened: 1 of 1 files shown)


kirkwaiblinger:
This is intentional. Generally speaking, `== null` is a general-purpose nullishness check, without giving specific meaning or importance to the difference between `null` vs `undefined`. We more or less use `=== null`/`=== undefined` only when there is an important difference between the two (which is rare).

eqeqeq enforces that `===` be used in all other cases than comparison to `null`/`undefined`.

Josh-Cena:
I know Brad has strong opinions that you should always conflate null and undefined, but I have strong opposite opinions :)

kirkwaiblinger:
In my book, the only good null check is `typeof x === 'object' && !x` 😆

PR #10004, opened 2024-09-17
fix(eslint-plugin): properly coerce all types to string in `getStaticMemberAccessValue`

### LINKED ISSUE #9999: Bug: [prefer-promise-reject-errors (and other rules too)] Cannot convert Symbol to string (8.6.0 regression)

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.2&fileType=.tsx&code=KYDwDg9gTgLgBAYwDYEMDOa4FkVgErADmAlmjFAJ4A8A0gDRwAqAfHAN4BQccYArgEZJiCOMRjAAtpgC8cAHbAA7tly0GLABQBKANxceAoSIDaAZQoT%2BEJADoxwKChjQAutvb7uUYDF5Q5cDAAFqR24lJmFla29o7OUG66%2BgC%2BHKlAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wlo4D2AW0rJEtTgCtEZOr2gDoqDJHmLI4MAF8QWoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
export class MapRegistry<K, T> {
  public items = new Map<K, T>();
  public [Symbol.iterator]() {
    return this.items[Symbol.iterator]();
  }
}
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/prefer-promise-reject-errors": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

lint should pass

### Actual Result

`TypeError: Cannot convert a Symbol value to a string`

### Additional Info

8.5.0 works fine

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9999 

## Overview

#9836 had some bugs with how various data types are coerced to strings when used in a computed `MemberAccess`; this PR fixes those bugs.

### DIFF
--- a/packages/eslint-plugin/src/util/isArrayMethodCallWithPredicate.ts
+++ b/packages/eslint-plugin/src/util/isArrayMethodCallWithPredicate.ts
@@ -30,7 +30,10 @@ export function isArrayMethodCallWithPredicate(
 
   const staticAccessValue = getStaticMemberAccessValue(node.callee, context);
 
-  if (!staticAccessValue || !ARRAY_PREDICATE_FUNCTIONS.has(staticAccessValue)) {
+  if (
+    typeof staticAccessValue !== 'string' ||
+    !ARRAY_PREDICATE_FUNCTIONS.has(staticAccessValue)
+  ) {
     return false;
   }
 
--- a/packages/eslint-plugin/src/util/isArrayMethodCallWithPredicate.ts
+++ b/packages/eslint-plugin/src/util/isArrayMethodCallWithPredicate.ts
@@ -9,7 +9,7 @@ import * as tsutils from 'ts-api-utils';
 
 import { getStaticMemberAccessValue } from './misc';
 
-const ARRAY_PREDICATE_FUNCTIONS = new Set([
+const ARRAY_PREDICATE_FUNCTIONS = new Set<unknown>([
   'filter',
   'find',
   'findIndex',
@@ -30,10 +30,7 @@ export function isArrayMethodCallWithPredicate(
 
   const staticAccessValue = getStaticMemberAccessValue(node.callee, context);
 
-  if (
-    typeof staticAccessValue !== 'string' ||
-    !ARRAY_PREDICATE_FUNCTIONS.has(staticAccessValue)
-  ) {
+  if (!ARRAY_PREDICATE_FUNCTIONS.has(staticAccessValue)) {
     return false;
   }
 
--- a/packages/eslint-plugin/src/util/misc.ts
+++ b/packages/eslint-plugin/src/util/misc.ts
@@ -292,9 +292,6 @@ function getStaticMemberAccessValue(
   const key =
     node.type === AST_NODE_TYPES.MemberExpression ? node.property : node.key;
   const { type } = key;
-  if (type === AST_NODE_TYPES.Literal) {
-    return String(key.value);
-  }
   if (
     !node.computed &&
     (type === AST_NODE_TYPES.Identifier ||
--- a/packages/eslint-plugin/src/util/misc.ts
+++ b/packages/eslint-plugin/src/util/misc.ts
@@ -253,20 +253,15 @@ function getStaticMemberAccessValue(
     node.type === AST_NODE_TYPES.MemberExpression ? node.property : node.key;
   if (!node.computed) {
     return key.type === AST_NODE_TYPES.Literal
-      ? `${key.value}`
+      ? String(key.value)
       : (key as TSESTree.Identifier | TSESTree.PrivateIdentifier).name;
   }
   const result = getStaticValue(key, sourceCode.getScope(node));
   if (!result) {
-    return;
+    return undefined;
   }
   const { value } = result;
-  return typeof value === 'symbol'
-    ? value
-    : /* intentional interpolation of an `unknown` value, because any non-Symbol value is coerced to a string by
-         JavaScript when in a `MemberAccess`. */
-      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-      `${value}`;
+  return typeof value === 'symbol' ? value : String(value);
 }
 
 /**
--- a/packages/eslint-plugin/src/util/misc.ts
+++ b/packages/eslint-plugin/src/util/misc.ts
@@ -2,8 +2,7 @@
  * @fileoverview Really small utility functions that didn't deserve their own files
  */
 import { requiresQuoting } from '@typescript-eslint/type-utils';
-import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
-import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES, TSESLint, TSESTree } from '@typescript-eslint/utils';
 import type { RuleContext } from '@typescript-eslint/utils/ts-eslint';
 import * as ts from 'typescript';
 
@@ -251,10 +250,15 @@ function getStaticMemberAccessValue(
 ): string | symbol | undefined {
   const key =
     node.type === AST_NODE_TYPES.MemberExpression ? node.property : node.key;
-  if (!node.computed) {
-    return key.type === AST_NODE_TYPES.Literal
-      ? String(key.value)
-      : (key as TSESTree.Identifier | TSESTree.PrivateIdentifier).name;
+  const { type } = key;
+  if (type === AST_NODE_TYPES.Literal) {
+    return String(key.value);
+  }
+  if (
+    type === AST_NODE_TYPES.Identifier ||
+    type === AST_NODE_TYPES.PrivateIdentifier
+  ) {
+    return key.name;
   }
   const result = getStaticValue(key, sourceCode.getScope(node));
   if (!result) {
--- a/packages/eslint-plugin/src/util/misc.ts
+++ b/packages/eslint-plugin/src/util/misc.ts
@@ -242,7 +242,7 @@ type NodeWithKey =
 function getStaticMemberAccessValue(
   node: NodeWithKey,
   { sourceCode }: RuleContext<string, unknown[]>,
-): string | undefined {
+): string | null {
   const key =
     node.type === AST_NODE_TYPES.MemberExpression ? node.property : node.key;
   if (!node.computed) {
@@ -250,12 +250,9 @@ function getStaticMemberAccessValue(
       ? `${key.value}`
       : (key as TSESTree.Identifier | TSESTree.PrivateIdentifier).name;
   }
-  const value = getStaticValue(key, sourceCode.getScope(node))?.value as
-    | string
-    | number
-    | null
-    | undefined;
-  return value == null ? undefined : `${value}`;
+  const result = getStaticValue(key, sourceCode.getScope(node));
+  // we must use `String(...)` rather than template literal interpolation, because interpolation throws a runtime error if `value` is a `symbol`
+  return result && String(result.value);
 }
 
 /**
@@ -268,7 +265,7 @@ const isStaticMemberAccessOfValue = (
   context: RuleContext<string, unknown[]>,
   ...values: string[]
 ): boolean =>
-  (values as (string | undefined)[]).includes(
+  (values as (string | null)[]).includes(
     getStaticMemberAccessValue(memberExpression, context),
   );
 
--- a/packages/eslint-plugin/src/util/misc.ts
+++ b/packages/eslint-plugin/src/util/misc.ts
@@ -251,7 +251,8 @@ function getStaticMemberAccessValue(
       : (key as TSESTree.Identifier | TSESTree.PrivateIdentifier).name;
   }
   const result = getStaticValue(key, sourceCode.getScope(node));
-  // we must use `String(...)` rather than template literal interpolation, because interpolation throws a runtime error if `value` is a `symbol`
+  /* we must use `String(...)` rather than template literal interpolation, because interpolation throws a runtime error
+     if `value` is a `symbol` */
   return result && String(result.value);
 }
 
--- a/packages/eslint-plugin/src/util/misc.ts
+++ b/packages/eslint-plugin/src/util/misc.ts
@@ -240,9 +240,49 @@ type NodeWithKey =
   | TSESTree.TSAbstractPropertyDefinition;
 
 /**
- * Gets the key being accessed or declared, such as `value` in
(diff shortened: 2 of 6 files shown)


packages/eslint-plugin/src/util/misc.ts
```diff
@@ -242,20 +242,18 @@ type NodeWithKey =
...
   if (!node.computed) {
     return key.type === AST_NODE_TYPES.Literal
       ? `${key.value}`
       : (key as TSESTree.Identifier | TSESTree.PrivateIdentifier).name;
   }
-  const value = getStaticValue(key, sourceCode.getScope(node))?.value as
-    | string
-    | number
-    | null
-    | undefined;
```
abrahamguo, line 257: this type assertion was not correct. `getStaticValue` can return _anything_ — `symbol`, `bigint`, `array`, `Date`, etc.

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
```diff
@@ -266,6 +266,7 @@ ruleTester.run('prefer-promise-reject-errors', rule, {
       declare const foo: PromiseConstructor;
       foo.reject(new Error());
     `,
+    'console[Symbol.iterator]();',
```
JoshuaKGoldberg, line 269: [Testing] This is just a bit different from the `public [Symbol.iterator]() { ... }` from #9999. Could you include that too, please?

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -242,20 +242,19 @@ type NodeWithKey =
...
-): string | undefined {
+): string | null {
   const key =
     node.type === AST_NODE_TYPES.MemberExpression ? node.property : node.key;
   if (!node.computed) {
     return key.type === AST_NODE_TYPES.Literal
       ? `${key.value}`
       : (key as TSESTree.Identifier | TSESTree.PrivateIdentifier).name;
   }
-  const value = getStaticValue(key, sourceCode.getScope(node))?.value as
```
abrahamguo, line 253: using optional chaining was not correct. we need to distinguish between a return value of `null`, which means "no static value", versus a return value of `{ value: null }`, which means "access the property called `'null'`.

JoshuaKGoldberg (dismissed):
> [abrahamguo](https://github.com/abrahamguo) commented [5 hours ago](https://github.com/typescript-eslint/typescript-eslint/issues/9999#issuecomment-2353986652):
>
> ... I will plan on opening a PR with the fix in the next 12 hours  ...

You work fast! 😄

...and, thanks for this! We can ship a quick patch release to get this to users.

packages/eslint-plugin/tests/rules/explicit-module-boundary-types.test.ts
```diff
@@ -416,7 +417,7 @@ export class Test {
       `,
       options: [
         {
-          allowedNames: ['prop', 'method', 'foo'],
+          allowedNames: ['prop', 'method', 'null', 'foo'],
```
abrahamguo, line 420: note that adding `'Symbol(a)'` to `allowedNames` allows both
`'Symbol(a)'() {}`
and
`[Symbol('a')]() {}`

Not sure if this is a feature or a bug, or if it's worth documenting? 🤷‍♂️

packages/eslint-plugin/tests/rules/prefer-promise-reject-errors.test.ts
abrahamguo: > You work fast! 😄

Under promise, over deliver, etc etc 😉 

Additional test added!

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -242,20 +242,19 @@ type NodeWithKey =
...
-    | string
-    | number
-    | null
-    | undefined;
-  return value == null ? undefined : `${value}`;
```
abrahamguo, line 258: seems to be
- a strange inconsistency of JavaScript that `` `${Symbol('a')}` `` throws while `String(Symbol('a'))` and `Symbol('a').toString()` work fine
- a minor type hole in TypeScript that `tsc` _does_ warn for interpolating a `Symbol`, but does _not_ warn for interpolating an `unknown`

abrahamguo:
Lint failed due to CI flake.

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -242,20 +242,19 @@ type NodeWithKey =
...
+  const result = getStaticValue(key, sourceCode.getScope(node));
+  /* If a value is returned, coerce it to a string, because that's what JavaScript does with any data type in a
```
kirkwaiblinger, line 254: Well, JS doesn't do this with symbols, though (which is kind of the whole point of symbols). Symbols are keys unto themselves. So, I wouldn't think it would make sense to stringify symbols ever (their string representation is only a debugging tool). Shouldn't we just return the symbol itself?

abrahamguo: Yes, I agree that is more correct. The only downside is that Symbol-based methods will not be able to be allow-listed using things like `explicit-module-boundary-types` => `allowedNames` or `class-methods-use-this` => `exceptMethods`.

packages/eslint-plugin/tests/rules/explicit-module-boundary-types.test.ts
abrahamguo: This is no longer true with the latest change requested by @kirkwaiblinger.

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -242,20 +242,25 @@ type NodeWithKey =
...
   node: NodeWithKey,
   { sourceCode }: RuleContext<string, unknown[]>,
-): string | undefined {
+): string | symbol | undefined {
   const key =
     node.type === AST_NODE_TYPES.MemberExpression ? node.property : node.key;
   if (!node.computed) {
     return key.type === AST_NODE_TYPES.Literal
       ? `${key.value}`
       : (key as TSESTree.Identifier | TSESTree.PrivateIdentifier).name;
```
kirkwaiblinger, line 251: This type assertion isn't necessary if the `key` is checked within the condition, i.e.
```suggestion
  if (!node.computed) {
    const key =
      node.type === AST_NODE_TYPES.MemberExpression ? node.property : node.key;
    return key.type === AST_NODE_TYPES.Literal
      ? `${key.value}`
      : key.name;
```

Taking this suggestion literally will incur duplication, though, since `key` is used in the else branch too. Is there a way to make this nice?

Granted another type assertion in this very function was part of what burned us on this bug, it would be nice to avoid this one too.

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -242,20 +242,25 @@ type NodeWithKey =
 function getStaticMemberAccessValue(
   node: NodeWithKey,
   { sourceCode }: RuleContext<string, unknown[]>,
-): string | undefined {
+): string | symbol | undefined {
   const key =
     node.type === AST_NODE_TYPES.MemberExpression ? node.property : node.key;
   if (!node.computed) {
     return key.type === AST_NODE_TYPES.Literal
       ? `${key.value}`
```
kirkwaiblinger, line 250:
```suggestion
      ? String(key.value)
```

packages/eslint-plugin/src/util/isArrayMethodCallWithPredicate.ts
```diff
@@ -30,7 +30,10 @@ export function isArrayMethodCallWithPredicate(
 
   const staticAccessValue = getStaticMemberAccessValue(node.callee, context);
 
-  if (!staticAccessValue || !ARRAY_PREDICATE_FUNCTIONS.has(staticAccessValue)) {
+  if (
+    typeof staticAccessValue !== 'string' ||
+    !ARRAY_PREDICATE_FUNCTIONS.has(staticAccessValue)
+  ) {
```
kirkwaiblinger, line 36: (nit) I think applying DeMorgan on these makes the intention much more clear, but up to you! 🙂 
```suggestion
  if (
    !(typeof staticAccessValue === 'string' &&
       ARRAY_PREDICATE_FUNCTIONS.has(staticAccessValue)
    )
  ) {
```

packages/eslint-plugin/tests/rules/explicit-module-boundary-types.test.ts
kirkwaiblinger: Since `Symbol('a')` is always a unique value, I don't think that it would make sense for the `allowedNames` to be able to specify it in this way.

Whether `Symbol.for('a')` should have a way of specifying it.... 🤷‍♂️. That could be sorted out in a feature request. In any case, I think it would belong in the rule logic and/or specifier logic to stringify the symbol, rather than in the static member access utility.

kirkwaiblinger (changes requested):
Some syntax changes requested, but I think the logic is good

packages/eslint-plugin/src/util/isArrayMethodCallWithPredicate.ts
kirkwaiblinger: (side note) - I _really_ wish that `Set#has` were typed to allow `unknown` rather than the type of the set . Having to narrow the type to the set is often counterproductive. 

I personally actually really wouldn't mind doing
```suggestion
  if (
    ARRAY_PREDICATE_FUNCTIONS.has(staticAccessValue as any)
  ) {
```
but i have a feeling that might not be so popular

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -242,20 +242,25 @@ type NodeWithKey =
...
+  if (!result) {
+    return;
+  }
+  const { value } = result;
+  return typeof value === 'symbol'
+    ? value
+    : /* intentional interpolation of an `unknown` value, because any non-Symbol value is coerced to a string by
+         JavaScript when in a `MemberAccess`. */
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      `${value}`;
```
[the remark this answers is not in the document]
abrahamguo: We should really be checking `String(...)` with `restrict-template-expressions`, meaning that this will come back 😛

packages/eslint-plugin/src/util/isArrayMethodCallWithPredicate.ts
abrahamguo: I 200% agree with you about the type of `Set#has` (and `Array#includes`) haha...
https://github.com/mattpocock/ts-reset

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -239,36 +239,42 @@ type NodeWithKey =
   | TSESTree.PropertyDefinition
   | TSESTree.TSAbstractMethodDefinition
   | TSESTree.TSAbstractPropertyDefinition;
+
+/**
+ * Gets the key being accessed or declared, such as `value` in
+ * `x.value`, `x['value']`,
+ * or even `const v = 'value'; x[v]` (or optional variants thereof).
+ */
```
kirkwaiblinger, line 247:
```suggestion
/**
 * Gets a member being accessed or declared if its value can be statically, and
 * resolves it to the string or symbol value that will be used as the actual member
 * access key at runtime. Otherwise, returns `undefined`.
 *
 * ```ts
 * x.member // returns 'member'
 * ^^^^^^^^
 * 
 * x?.member // returns 'member' (optional chaining is treated the same)
 * ^^^^^^^^^
 *
 * x['value'] // returns 'value'
 * ^^^^^^^^^^
 *
 * x[Math.random()] // returns undefined (not a static value)
 * ^^^^^^^^^^^^^^^^
 *
 * arr[0] // returns '0' (NOT 0)
 * ^^^^^^
 *
 * arr[0n] // returns '0' (NOT 0n)
 * ^^^^^^^
 *
 * const s = Symbol.for('symbolName')
 * x[s] // returns `Symbol.for('symbolName')` (since it's a static/global symbol)
 * ^^^^
 *
 * const us = Symbol.('symbolName')
 * x[us] // returns undefined (since it's a unique symbol, so not statically analyzable)
 * ^^^^^
 *
 * var object = {
 *     1234: '4567', // returns '1234' (NOT 1234)
 *     ^^^^^^^^^^^^
 *     method() { } // returns 'method'
 *     ^^^^^^^^^^^^
 * }
 *
 * class WithMembers {
 *     foo: string // returns 'foo'
 *     ^^^^^^^^^^^
 * }
 * ```
 */
```

packages/eslint-plugin/src/util/misc.ts
abrahamguo: Totally agree! Refactored ✅

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -241,20 +241,66 @@ type NodeWithKey =
...
   { sourceCode }: RuleContext<string, unknown[]>,
 ): string | symbol | undefined {
   const key =
     node.type === AST_NODE_TYPES.MemberExpression ? node.property : node.key;
-  if (!node.computed) {
-    return key.type === AST_NODE_TYPES.Literal
-      ? String(key.value)
-      : (key as TSESTree.Identifier | TSESTree.PrivateIdentifier).name;
+  const { type } = key;
+  if (type === AST_NODE_TYPES.Literal) {
```
kirkwaiblinger, line 295: (question, no specific action/resolution implied) is this conditional necessary? Or will it just be caught by the `getStaticValue`?

kirkwaiblinger (approved):
I think this looks good! Thanks!

packages/eslint-plugin/src/util/misc.ts
kirkwaiblinger: checked locally and all the tests pass if I comment it out 🤷

abrahamguo: good catch, removed!

kirkwaiblinger (approved):

Josh-Cena (approved):

PR #10008, opened 2024-09-17
feat(eslint-plugin): [return-await] check for-await loop iteree

### LINKED ISSUE #8858: Enhancement: [await-thenable] should check that `for-await` loop is used on an async iterable

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.3&showAST=es&fileType=.tsx&code=GYewTgBAhg7lCWAXCAKAxiAdgZ2diIwEA2gOSYikA0EpiAFgKaZQBGANo9tbU2I6QC6ASggBvAL4AoRgA8ADuGRiIEoA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AQwHd3K78ALRE3YAjJOiiJo0APbRI4MAF8QSoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
for await (const s of ['no', 'thenables', 'here']) {}
export { }
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expect that the first line should flag due to `for await` being used on something that does not have a `Symbol.asyncIterator`.

### Actual Result

No error.

### Additional Info

Note that it's unambiguous IMO that the example given should flag. However, I, personally, would also like for the rule to flag on misused for-await-of, where the iterable is a _synchronous_ iterator over promises. That should be written as an ordinary for loop. It's only a true async iterable that should be allowed to be iterated over in a for-await-of loop.

Refresher on for-await-of: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8858 

## Overview

Just ask the type checker if a `Symbol.asyncIterator` is present, ezpz. Add tests and docs (less ezpz).

### DIFF
--- a/packages/eslint-plugin/docs/rules/await-thenable.mdx
+++ b/packages/eslint-plugin/docs/rules/await-thenable.mdx
@@ -50,8 +50,9 @@ While it is valid JavaScript to use `for await...of` with synchronous iterables 
 Synchronous iteration over thenables and asynchronous iteration are fundamentally different things, and the `for await...of` has several behaviors that are specifically designed for operating on async-iterables, that do not apply to sync-iterables.
 
 For one, the `for await...of` loop accesses its fulfilled values sequentially, which is an appropriate behavior for async-iterables, but is not always desirable for sync-iterables of Thenables.
+For instance, sync-iterables may need to manipulate their Thenable entries directly, without accessing their fulfillment values at all (that is to say, without awaiting them).
 
-Because of this, the error-handling of the `for await...of` loop works properly for async-iterables, but subtly permits unclosed generators and unhandled promise rejections when operating on sync-iterables of Thenables. For detailed examples of this, see the [MDN documentation on using `for await...of` with sync-iterables](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of#iterating_over_sync_iterables_and_generators).
+Because of this, the error-handling protocol of the `for await...of` loop works properly for async-iterables, but subtly permits unclosed generators and unhandled promise rejections when operating on sync-iterables of Thenables. For detailed examples of this, see the [MDN documentation on using `for await...of` with sync-iterables](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of#iterating_over_sync_iterables_and_generators).
 
 Instead of using the `for await...of` loop with sync-iterables of Thenables, or using `await` on promises within the body of a `for...of` (which also permits floating promise rejections), consider instead using one of the [promise concurrency methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#promise_concurrency) for robust behavior.
 
--- a/packages/eslint-plugin/docs/rules/await-thenable.mdx
+++ b/packages/eslint-plugin/docs/rules/await-thenable.mdx
@@ -10,9 +10,9 @@ import TabItem from '@theme/TabItem';
 > See **https://typescript-eslint.io/rules/await-thenable** for documentation.
 
 A "Thenable" value is an object which has a `then` method, such as a Promise.
-The `await` keyword is generally used to retrieve the result of calling a Thenable's `then` method.
+The [`await` keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) is generally used to retrieve the result of calling a Thenable's `then` method.
 
-If the `await` keyword is used on a value that is not a Thenable, the value is directly resolved immediately.
+If the `await` keyword is used on a value that is not a Thenable, the value is directly resolved, but will still pause execution until the next microtask.
 While doing so is valid JavaScript, it is often a programmer error, such as forgetting to add parenthesis to call a function that returns a Promise.
 
 ## Examples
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/docs/rules/await-thenable.mdx
```diff
@@ -40,6 +40,76 @@ await createValue();
 </TabItem>
 </Tabs>
 
+## Async Iteration (`for await...of` Loops)
+
+This rule also inspects [`for await...of` statements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of), which are designed for iterating over async-iterable objects.
+If the value being iterated over is not async-iterable, an ordinary `for...of` statement is preferable, even if the value is an iterable of Thenables.
```
Josh-Cena, line 46: This is debatable. Banning iterables of thenables should be an option.

kirkwaiblinger: Eh, kind of... in any case, I was going off of https://github.com/typescript-eslint/typescript-eslint/issues/8858#issuecomment-2148977352. It's easy enough to add an option to allow iterables of thenables, though I'm -0.5 on doing so.

Josh-Cena: In lieu of an option, can we add more motivation to the docs (as what I said in that comment)?

kirkwaiblinger: Good point! Made some changes around this 👍

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -62,6 +72,55 @@ export default createRule({
...
+        // symbol exists.
+        if (asyncIteratorSymbol == null) {
+          context.report({
+            loc: getForStatementHeadLoc(context.sourceCode, node),
+            messageId: 'forAwaitOfNonThenable',
+            suggest: [
+              // This suggestion causes broken code for sync iterables of promises, since
+              // the loop variable will not be awaited.
+              //
+              // Ideally, if the iterable yields promises, we would offer a suggestion to
```
kirkwaiblinger, line 103: bleh - I guess this isn't always a good idea, particularly with possibly infinite iterables. But, then again, why would you have an infinite sync iterable of promises? Thinking out loud.

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -62,6 +72,44 @@ export default createRule({
           });
         }
       },
+
+      ForOfStatement(node): void {
```
Josh-Cena, line 76:
```suggestion
      'ForOfStatement[await=true]'(node): void {
```

And you can remove the condition below

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -62,6 +72,44 @@ export default createRule({
...
+          type,
+          'asyncIterator',
+          checker,
+        );
+
+        if (asyncIteratorSymbol == null) {
+          context.report({
+            loc: getForStatementHeadLoc(context.sourceCode, node),
+            messageId: 'forAwaitOfNonThenable',
+            suggest: [
```
Josh-Cena, line 96: Can we only offer this suggestion if the iterator result is not thenable?

packages/eslint-plugin/docs/rules/await-thenable.mdx
```diff
@@ -40,6 +40,87 @@ await createValue();
...
+For instance, sync-iterables may need to manipulate their Thenable entries directly, without accessing their fulfillment values at all (that is to say, without awaiting them).
+
+Because of this, the error-handling protocol of the `for await...of` loop works properly for async-iterables, but subtly permits unclosed generators and unhandled promise rejections when operating on sync-iterables of Thenables. For detailed examples of this, see the [MDN documentation on using `for await...of` with sync-iterables](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of#iterating_over_sync_iterables_and_generators).
+
+Instead of using the `for await...of` loop with sync-iterables of Thenables, or using `await` on promises within the body of a `for...of` (which also permits floating promise rejections), consider instead using one of the [promise concurrency methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#promise_concurrency) for robust behavior.
```
Josh-Cena, line 57: This is really long and I couldn't grasp the point well, so suggesting a rewrite.

```suggestion
While `for await...of` can be used with synchronous iterables, and it will await each promise produced by the itereable, it is inadvisable to do so.
There are some tiny nuances that you may want to consider.

The biggest difference between using `for await...of` and using `for...of` (plus awaiting each result yourself) is error handling.
When an error occurs within the loop body, `for await...of` does _not_ close the original sync iterable, while `for...of` does.
For detailed examples of this, see the [MDN documentation on using `for await...of` with sync-iterables](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of#iterating_over_sync_iterables_and_generators).

Also consider whether you need sequential awaiting at all. Using `for await...of` may obscure potential opportunities for concurrent processing, such as those reported by [`no-await-in-loop`](https://eslint.org/docs/latest/rules/no-await-in-loop). Consider instead using one of the [promise concurrency methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#promise_concurrency) for better performance.
```

Is there a point I missed? Honestly I don't understand the point about floating promises. Those are tangential to this issue.

packages/eslint-plugin/src/rules/await-thenable.ts
kirkwaiblinger: I don't think so because we'd need `getIterationTypesOfIterable()` or `getIteratedTypeOrElementType` or similar from TS none of which is exposed on the `checker`.

If we check arrays and tuples specifically, i.e. known iterables, this is silly since 
```ts
for (const promisedValue of await Promise.all(promises)) {
}
```
is a much better suggestion in those cases (since `await`ing in the loop is generally problematic for exception handling given a non-lazy array of promises)

Whereas with lazy sync-iterables of promises, which I don't know if we can detect, it may be better to do 
```ts
for (const promise of yieldPromises()) {
    const promisedValue = await promise;
}
```

I guess the only thing that's unambiguous here is that if you have an array or tuple of non-thenables, this suggestion is good. Maybe that's the only situation where we give a suggestion?

packages/eslint-plugin/docs/rules/await-thenable.mdx
kirkwaiblinger: Yeah, your version is better 😁

> Honestly I don't understand the point about floating promises. Those are tangential to this issue.

Fair - I agree that they're tangential, so I think it makes sense to leave it off.

If you were curious, I was just trying to get at the fact that this causes avoidable rejections:
```ts
for (const promise of [Promise.reject(1), Promise.reject(2)] {
    const value = await promise;
    console.log(value);
}
```

See also - https://github.com/eslint/eslint/issues/18918 lol

JoshuaKGoldberg (approved):
🚢

phaux:
wrong rule in title lol

kirkwaiblinger:
> wrong rule in title lol

smh 😞. Changing the title and release notes, thanks for pointing this out!

haines:
I think this can result in false positives: https://github.com/typescript-eslint/typescript-eslint/issues/10111

Issue #9268, opened 2024-06-04
Bug: [no-unnecessary-type-assertion] Internal error cannot read properties of undefined (reading 'parent')


kasium:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.5&fileType=.tsx&code=KYDwDg9gTgLgBAYwDYEMDOa4GUIFthbBQBuAlgsHAN4CwAUHHGFKcSjJegJ4B2CcaPMABiAVz4AKAJQAuOAAUoeUmmAAeYhFIATAHzV6jRggg808QfgAi7FHAC8cFAHcUpeGBQBzYADpgbEii7MAS0g76AFJYAPIAcr6eUKoS2hAIovg8MACEUlIA3IZwAL70ZXRAA&eslintrc=N4KABBYEQA4IYCcDOBTBUBc0ACAXAnjCkgMYICWMuAtMQDbkB2uA9PMmlADTiSyKoEAeSrkA9oySYwoSHOgwEYgFYoSuaVFykJAM3IBzAHTKkE7r3nQzAVwQkUAFUIpNAWzEATG3VeWwAL48crB0NgZMUlgA2v4QUHgupBRUtEgMzCz0TDQwYRGMUP4AusF8KAAeuCiMnlFgsVZQ2cwYCGpibm41niieFk154UwYiUTJlDQtrO0knd21fQPyocOMowTjZJNpGTMdXT191O0AjjbkFIwG1Jso1CQAFmoA1kwGRXLFIAFAA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
// this code is shorted so that it triggers the internal error
// the error also shows up the real productive code

export class SomeService {
  private async someFunc(): Promise<void> {
    const someData = await page.evaluate(() => JSON.parse(document!));
[...]

PR #10009, opened 2024-09-18
feat(eslint-plugin): [no-unnecessary-condition] add checkTypePredicates

### LINKED ISSUE #9076: Enhancement: [no-unnecessary-condition] Check truthiness assertion functions

### Description

This is spinoff from [the corresponding issue for strict-boolean-expressions](https://github.com/typescript-eslint/typescript-eslint/issues/9009). 

Following the conversation from https://github.com/typescript-eslint/typescript-eslint/issues/9009#issuecomment-2105447714, it is proposed that no-unnecessary-condition also check the asserting argument of an assertion function as a boolean context for an unnecessary condition. 

However, for this rule, that behavior should definitely be behind an option, since assertion functions can often be used to assert that something which appears to be true according to the types actually is true at runtime. Codebases that do so regularly will probably want to disable this behavior.

### Fail

```TypeScript
function assert(x: unknown): asserts x {
  if (!x) {
    throw new Error('assertion');
  }
}

assert(true)
```

### Pass

```TypeScript
function assert(x: unknown): asserts x {
  if (!x) {
    throw new Error('assertion');
  }
}

assert(Math.random() > 0.5)
```

### Additional Info

_No response_

### LINKED ISSUE #10007: Enhancement: [no-unnecessary-condition] report on type guards usages where the type is already equal

### Description

Report when checking type guards, where the type of the variable is already _exactly_ equal to the type guard return type.

### Fail

```TypeScript
function isBool(x: unknown): x is boolean {
    return typeof x === 'boolean';
}

declare const b: boolean;

if (isBool(b)) { // this is unnecessary, since b is already boolean.
}
```

### Pass

```TypeScript
declare const maybeBoolean: boolean | string;

if (isBool(maybeBoolean)) { // necessary - might not a boolean.
}

declare const t: true;

if (isBool(t)) { // debatable - t is a subtype of boolean, not exactly equal. this has some funny effects in the type system.
    console.log(t); // t gets widened to boolean here, even though it was typed as true.
} else {
    console.error(t); // t also widened to boolean here.
}
```

### Additional Info

https://github.com/typescript-eslint/typescript-eslint/issues/9076 is in a similar domain, checking unnecessary assertion functions.

kirkwaiblinger:
## PR Checklist

   - fixes #9076, 
   - fixes #10007

## Overview

- Basically completely rewrite the existing method `findAssertedArgument` in SBE to use `checker.getResolvedSignature()` and `checker.getTypePredicateOfSignature()`, and move it into shared location
- Put new behavior behind option "checkTypePredicates" - see [this discord conversation](https://discord.com/channels/508357248330760243/1287101496269213778) for the thought process on naming
- Add docs
- Enable internally and suppress violations

### DIFF
--- a/packages/eslint-plugin/src/util/assertionFunctionUtils.ts
+++ b/packages/eslint-plugin/src/util/assertionFunctionUtils.ts
@@ -39,50 +39,41 @@ export function findTruthinessAssertedArgument(
   const calleeType = getConstrainedTypeAtLocation(services, node.callee);
   const callSignatures = tsutils.getCallSignaturesOfType(calleeType);
 
-  const checker = services.program.getTypeChecker();
-
-  let assertedParameterIndex: number | undefined = undefined;
-  for (const signature of callSignatures) {
-    const predicateInfo = checker.getTypePredicateOfSignature(signature);
-    if (predicateInfo == null) {
-      return undefined;
-    }
-
-    // Be sure we're dealing with a truthiness assertion function, in other words,
-    // `asserts x` (but not `asserts x is T`, and also not `asserts this is T`).
-    if (
-      !(
-        predicateInfo.kind === ts.TypePredicateKind.AssertsIdentifier &&
-        predicateInfo.type == null
-      )
-    ) {
-      return undefined;
-    }
-
-    const assertedParameterIndexForThisSignature = predicateInfo.parameterIndex;
-
-    if (
-      assertedParameterIndex != null &&
-      assertedParameterIndex !== assertedParameterIndexForThisSignature
-    ) {
-      // The asserted parameter we found for this signature didn't match
-      // previous signatures.
-      return undefined;
-    }
-
-    assertedParameterIndex = assertedParameterIndexForThisSignature;
-
-    if (assertedParameterIndex >= checkableArguments.length) {
-      return undefined;
-    }
-  }
-
-  // Didn't find a unique assertion index.
-  if (assertedParameterIndex == null) {
+  if (callSignatures.length === 0) {
     return undefined;
   }
 
-  return checkableArguments[assertedParameterIndex];
+  const checker = services.program.getTypeChecker();
+
+  const typePredicates = callSignatures.map(signature =>
+    checker.getTypePredicateOfSignature(signature),
+  );
+
+  const [firstTypePredicateResult, ...otherTypePredicateResults] =
+    typePredicates;
+  if (firstTypePredicateResult == null) {
+    return undefined;
+  }
+
+  // Ensure all call signatures are asserting the same thing.
+  const { parameterIndex, kind, type } = firstTypePredicateResult;
+  if (!(kind === ts.TypePredicateKind.AssertsIdentifier && type == null)) {
+    return undefined;
+  }
+
+  if (
+    otherTypePredicateResults.some(
+      otherResult =>
+        otherResult == null ||
+        otherResult.parameterIndex !== parameterIndex ||
+        otherResult.kind !== kind ||
+        otherResult.type != null,
+    )
+  ) {
+    return undefined;
+  }
+
+  return checkableArguments.at(parameterIndex);
 }
 
 /**
@@ -94,11 +85,9 @@ export function findTypeGuardAssertedArgument(
   node: TSESTree.CallExpression,
 ):
   | {
-      predicateKind:
-        | ts.TypePredicateKind.AssertsIdentifier
-        | ts.TypePredicateKind.Identifier;
+      asserts: boolean;
       argument: TSESTree.Expression;
-      type: ts.Type | undefined;
+      type: ts.Type;
     }
   | undefined {
   // If the call looks like `assert(expr1, expr2, ...c, d, e, f)`, then we can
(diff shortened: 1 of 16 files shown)


packages/typescript-estree/tests/lib/semanticInfo.test.ts
```diff
@@ -172,6 +172,7 @@ describe('semanticInfo', () => {
     ).declarations[0].init!;
     const tsBinaryExpression =
       parseResult.services.esTreeNodeToTSNodeMap.get(binaryExpression);
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- intentionally asserting that the runtime value matches the types.
```
JoshuaKGoldberg, line 175: 🤔 I feel like we could just use an `as` for this file's asserts... they're roughly doing the same thing.

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
JoshuaKGoldberg, line 1: [Testing] Lots of logic in `assertionFunctionUtils.ts` around different parameter types, but no examples here. Could you add some? First examples that come to mind:

* `this` parameters
* `declare function assert(a: unknown, b: unknown, c: unknown): asserts b;`
* Spreads/rests

JoshuaKGoldberg (changes requested):
Seems very reasonable to me! The comments were quite helpful, thanks for that. 🙌 

Requesting changes on added test coverage.

packages/typescript-estree/tests/lib/semanticInfo.test.ts
kirkwaiblinger: well... no. These don't change the types, _that's why they're flagged_ 😀 

And this is exactly one of the cases where asserting that the runtime values match the types is warranted; the `esTreeNodeToTSNodeMap` and its inverse can well have unmapped values, see also https://github.com/typescript-eslint/typescript-eslint/issues/9268#issuecomment-2148437137 for a real bug caused by this.

However, I did notice that some of the assertions were actually redundant, and others could be replaced with an ordinary `expect(x).toBeDefined()`, which really just hides the unnecessariness from the rule. 

Made some changes 👍

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
kirkwaiblinger: Added some coverage... Much more extensive tests of the `findAsssertedArgument` logic itself were written for SBE in https://github.com/typescript-eslint/typescript-eslint/pull/9074

LMK if you want me to go up to that level of coverage here as well, or whether the amount there is now is ok

JoshuaKGoldberg: 👍 lgtm!

packages/eslint-plugin/src/util/assertionFunctionUtils.ts
kirkwaiblinger, line 1: FYI - I realized while working on this that we can use `checker.getTypePredicateOfSignature(signature)`, which.... does all the heavy lifting for us 🤦‍♂️ 

So this is now _much_ simpler. 🥳

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
```diff
@@ -516,29 +516,6 @@ declare const nullableString: string | null;
 assert(3 as any, nullableString);
       `,
     },
-    // Intentional TS error - A rest parameter must be last in a parameter list.
```
kirkwaiblinger, line 519: These tests were for coverage reasons that no longer make sense with the new algorithm.
None of them crashes. 2 of them now give a false positive, but they match how TS thinks about the situation 🤷‍♂️

packages/eslint-plugin/src/util/assertionFunctionUtils.ts
kirkwaiblinger: See https://github.com/typescript-eslint/typescript-eslint/pull/10009/commits/4ac67657e1d48eea9f0fad75ffa2e6023e1b8e53 for just the diff of the algorithm change

kirkwaiblinger:
Per discord conversation, we'll want to bring https://github.com/typescript-eslint/typescript-eslint/issues/10007 into this in order to put it behind the same option. Marking as draft until then

kirkwaiblinger:
This has changed a _lot_ since its initial version. I would recommend reviewing it more or less from scratch

packages/typescript-estree/tests/test-utils/test-utils.ts
```diff
@@ -139,6 +140,7 @@ export function omitDeep(
             value.push(visit(el, node));
           }
           node[prop] = value;
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- I don't know if it's safe to change this.
```
JoshuaKGoldberg, line 143: Seems odd to me to leave an _"I don't know if ..."_ in code. I tried removing them locally and some `semanticInfo.test.ts` snapshots broke. I got around it by:

* Changing the type of `oNode` and the return types of `function visit` & `function omitDeep` to `UnknownObject | undefined`
* Removing this `if (isObjectLike(child)) {` from the `else`

WDYT?

JoshuaKGoldberg (approved):
LGTM, very nice refactor!

Just one nit suggestion on `test-utils.ts`, otherwise 🚀 from me.

packages/typescript-estree/tests/test-utils/test-utils.ts
kirkwaiblinger: Hmm, well I'm scared of that since the one usage of `omitDeep` is asserted to be not null https://github.com/typescript-eslint/typescript-eslint/blob/d6c67fe0854ffdbe7ee2623d7ea05be7a04cd1c0/packages/typescript-estree/tests/test-utils/test-utils.ts#L75-L82

But I'm realizing now that this is "just" test code so I feel a little less scared of messing around with it

kirkwaiblinger: An easier fix for now seems like just turning `isObjectLike` into an ordinary `boolean`-returning function, since its type predicate isn't actually needed in the two places it's used.  🤷 Plus, no runtime changes, yay!

JoshuaKGoldberg: 🤝 perfect!

PR #10010, opened 2024-09-18
feat(eslint-plugin): [no-unsafe-call] check calls of Function

### LINKED ISSUE #9108: no-unsafe-call: ban calling Function

### Description

Unintuitively, TS narrows `typeof x === 'function'` to the unsafe `Function` type if x is not already constrained. However, it is safe to use on a constrained type. We should report on unconstrained callables.

### Fail Cases

```TypeScript
declare const maybeFunction: unknown;
if (typeof maybeFunction === 'function') {
    // TS allows this.
    maybeFunction('call', 'with', 'any', 'args');
}
```

### Pass Cases

```TypeScript
declare const maybeFunction: string | (() => string);
if (typeof maybeFunction === 'function') {
    // TS errors if more args are provided than the declared call signature, so this is safe.
    maybeFunction();
}
```

### Additional Info

it's probably fine/good to allow `typeof x === 'function` as long as `x` isn't called. I'm not sure exactly where the line should be drawn between reporting and not reporting. It seems like it should be reported if we _pass_ `x` to a function that might expect a specific call signature. Maybe a heuristic could be `typeof x === 'function` is ok as long as `x` isn't referenced in the `true` branch of the conditional?

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9108 

## Overview

- Simply check for `Function` type in existing rule.
- Minor cleanup of some sloppiness in the report messages, to ensure that they can say "`Function` typed" rather than "`any` typed"

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-unsafe-call.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unsafe-call.mdx
@@ -59,6 +59,34 @@ String.raw`foo`;
 </TabItem>
 </Tabs>
 
+## The Unsafe `Function` Type
+
+The `Function` type is behaves almost identically to `any` when called, so this rule also disallows calling values of type `Function`.
+
+<Tabs>
+<TabItem value="❌ Incorrect">
+
+```ts
+const f: Function = () => {};
+f();
+```
+
+</TabItem>
+</Tabs>
+
+Note that whereas [no-unsafe-function-type](./no-unsafe-function-type.mdx) helps prevent the _creation_ of `Function` types, this rule helps prevent the unsafe _use_ of `Function` types, which may creep into your codebase without explicitly referencing the `Function` type at all.
+See, for example, the following code:
+
+```ts
+function unsafe(maybeFunction: unknown): string {
+  if (typeof maybeFunction === 'function') {
+    // TypeScript allows this, but it's completely unsound.
+    return maybeFunction('call', 'with', 'any', 'args');
+  }
+  // etc
+}
+```
+
 ## When Not To Use It
 
 If your codebase has many existing `any`s or areas of unsafe code, it may be difficult to enable this rule.
--- a/packages/eslint-plugin/src/rules/no-unsafe-call.ts
+++ b/packages/eslint-plugin/src/rules/no-unsafe-call.ts
@@ -107,7 +107,7 @@ export default createRule<[], MessageIds>({
             return;
           }
         } else {
-          // eslint-disable-next-line no-lonely-if -- readability
+          // eslint-disable-next-line no-lonely-if
           if (callSignatures.length > 0) {
             return;
           }
--- a/packages/eslint-plugin/src/rules/no-unsafe-call.ts
+++ b/packages/eslint-plugin/src/rules/no-unsafe-call.ts
@@ -75,7 +75,44 @@ export default createRule<[], MessageIds>({
             type: isErrorType ? '`error` type' : '`any`',
           },
         });
-      } else if (isBuiltinSymbolLike(services.program, type, 'Function')) {
+        return;
+      }
+
+      if (isBuiltinSymbolLike(services.program, type, 'Function')) {
+        // this also matches subtypes of `Function`, like `interface Foo extends Function {}`.
+        //
+        // For weird TS reasons that I don't understand, these are
+        //
+        // safe to construct if:
+        // - they have at least one call signature _that is not void-returning_,
+        // - OR they have at least one construct signature.
+        //
+        // safe to call (including as template) if:
+        // - they have at least one call signature
+        // - OR they have at least one construct signature.
+
+        const constructSignatures = type.getConstructSignatures();
+        if (constructSignatures.length > 0) {
+          return;
+        }
+
+        const callSignatures = type.getCallSignatures();
+        if (messageId === 'unsafeNew') {
+          const nonVoidReturnCallSignatures = callSignatures.filter(
+            signature =>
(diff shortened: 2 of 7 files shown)


packages/eslint-plugin/src/rules/no-unsafe-call.ts
```diff
@@ -74,6 +75,14 @@ export default createRule<[], MessageIds>({
             type: isErrorType ? '`error` type' : '`any`',
           },
         });
+      } else if (isBuiltinSymbolLike(services.program, type, 'Function')) {
```
kirkwaiblinger, line 78: Drafting in order to address trouble with 
```ts
interface AlsoBad extends Function {
}

interface NotBad extends Function {
   (x: string) => string;
}
```

packages/website/src/components/lib/parseConfig.ts
```diff
@@ -55,7 +55,7 @@ export function parseTSConfig(code?: string): TSConfig {
 const moduleRegexp = /(module\.exports\s*=)/g;
 
 function constrainedScopeEval(obj: string): unknown {
-  // eslint-disable-next-line @typescript-eslint/no-implied-eval
+  // eslint-disable-next-line @typescript-eslint/no-implied-eval, @typescript-eslint/no-unsafe-call
```
kirkwaiblinger, line 58: To be clear, 
```ts
// this is the no-implied-eval violation
const yoloFunction = Function('inline function');
// this is the no-unsafe-call violation
yoloFunction();
```
They just happen to be on the same line here. So, this isn't a case of these rules stepping on each others' toes IMO.

packages/eslint-plugin/docs/rules/no-unsafe-call.mdx
```diff
@@ -59,6 +59,34 @@ String.raw`foo`;
...
+
+```ts
+function unsafe(maybeFunction: unknown): string {
+  if (typeof maybeFunction === 'function') {
+    // TypeScript allows this, but it's completely unsound.
+    return maybeFunction('call', 'with', 'any', 'args');
+  }
+  // etc
+}
+```
```
JoshuaKGoldberg, line 88: [Praise] Great example, really clear and concise.

packages/eslint-plugin/src/rules/no-unsafe-call.ts
```diff
@@ -74,6 +75,52 @@ export default createRule<[], MessageIds>({
...
+          const nonVoidReturnCallSignatures = callSignatures.filter(
+            signature =>
+              !tsutils.isIntrinsicVoidType(signature.getReturnType()),
+          );
+
+          if (nonVoidReturnCallSignatures.length > 0) {
+            return;
+          }
+        } else {
+          // eslint-disable-next-line no-lonely-if
```
JoshuaKGoldberg, line 110: y?

packages/eslint-plugin/src/rules/no-unsafe-call.ts
```diff
@@ -74,6 +75,52 @@ export default createRule<[], MessageIds>({
...
+        }
+
+        const callSignatures = type.getCallSignatures();
+        if (messageId === 'unsafeNew') {
+          const nonVoidReturnCallSignatures = callSignatures.filter(
+            signature =>
+              !tsutils.isIntrinsicVoidType(signature.getReturnType()),
+          );
+
+          if (nonVoidReturnCallSignatures.length > 0) {
```
JoshuaKGoldberg, line 106: [Style] Nit: no need to loop through all of the signatures, even if there are never that many...

```suggestion
          if (
            callSignatures.some(
              signature =>
                !tsutils.isIntrinsicVoidType(signature.getReturnType()),
            )
          ) {
```

packages/eslint-plugin/src/rules/no-unsafe-call.ts
```diff
@@ -74,6 +75,52 @@ export default createRule<[], MessageIds>({
...
+        // - they have at least one call signature _that is not void-returning_,
+        // - OR they have at least one construct signature.
+        //
+        // safe to call (including as template) if:
+        // - they have at least one call signature
+        // - OR they have at least one construct signature.
+
+        const constructSignatures = type.getConstructSignatures();
+        if (constructSignatures.length > 0) {
+          return;
```
JoshuaKGoldberg, line 96: [Testing] Missing coverage here. I got it with:

```ts
      code: `
interface StillUnsafe extends Function {
  property: string;
  new(): StillUnsafe;
}
declare const unsafe: StillUnsafe;
unsafe();
      `,
```

...but haven't checked if there's a simpler case.

JoshuaKGoldberg (changes requested):
_Awesome_. Just some comments on testing/style, nothing major.

packages/eslint-plugin/src/rules/no-unsafe-call.ts
kirkwaiblinger: oh. I'll revert that.

This is just my personal vendetta; I _really_, _really_ dislike no-lonely-if. IMO if/elseif should only be used for conditions which can be safely reorded, which is incompatible with no-lonely-if

packages/eslint-plugin/src/rules/no-unsafe-call.ts
kirkwaiblinger: Ah, yes. One of issues that is super obvious when reading code, but not when coming up with the rule logic 😆

JoshuaKGoldberg (approved):
Sweet!

PR #10019, opened 2024-09-19
docs(website): Open in Playground button should not cover text

### LINKED ISSUE #9998: Bug(website): Open in Playground button should not cover text
### Relevant Package

website

### Description

Not sure if we have any cases of this currently live but I saw this on a deploy preview of #9707 

image of button covering text

### Additional Info

_No response_

mdm317:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9998 

## Overview

I increased the padding-bottom value when there was a possibility of the 'Open in Playground' button overlapping with the code text.

Before
Screenshot 2024-09-19 at 7 36 20 PM

After
Screenshot 2024-09-19 at 7 35 53 PM

### DIFF
--- a/packages/website/src/theme/CodeBlock/Content/styles.module.css
+++ b/packages/website/src/theme/CodeBlock/Content/styles.module.css
@@ -43,7 +43,8 @@
   /* rtl:ignore */
   float: left;
   min-width: 100%;
-  padding: var(--ifm-pre-padding);
+  padding: var(--ifm-pre-padding) var(--ifm-pre-padding)
+    calc(var(--ifm-pre-padding) * 2.5);
 }
 
 .codeBlockLinesWithNumbering {
@@ -102,5 +103,9 @@
   .playgroundButton {
     position: static;
     margin: calc(var(--ifm-pre-padding) / 2);
+  }
+
+  .codeBlockLines {
+    padding: var(--ifm-pre-padding);
   }
 }


packages/website/src/theme/CodeBlock/Content/styles.module.css
```diff
@@ -43,7 +43,8 @@
   /* rtl:ignore */
   float: left;
   min-width: 100%;
-  padding: var(--ifm-pre-padding);
+  padding: var(--ifm-pre-padding) var(--ifm-pre-padding)
+    calc(var(--ifm-pre-padding) * 2.5);
```
JoshuaKGoldberg, line 47: [Bug] This _always_ adds the extra padding. Which means one-line codeblocks get it:

Screenshot of the Getting Started page's Step 1: Installation code block with unnecessary extra bottom padding

We shouldn't add the extra padding if the code block doesn't need it.

Proposal: how about adding a class to the block if the last line of code goes over a certain character length threshold?

JoshuaKGoldberg (approved):
Thanks for starting on this! It's not so straightforward, I'm sad to say 🙂

JoshuaKGoldberg (changes requested):
wrong button :(

mdm317:
Thank you for taking the time to review my code! 
I’ll make the necessary changes and push the updates shortly. Your insights are very helpful.

packages/website/src/theme/CodeBlock/Content/styles.module.css
mdm317: > Proposal: how about adding a class to the block if the last line of code goes over a certain character length threshold?

I updated it to add padding only when the button appears because I think it's difficult to detect when the length exceeds the threshold

JoshuaKGoldberg: > I think it's difficult to detect when the length exceeds the threshold

Why is that? We always know the content of the code block. Is doing something like this not possible?

```ts
const linesOfCode = code.split("\n");
const lastLineOfCode = code.at(-1);
const lastLineLength = lastLineOfCode.length;
const showButton = lastLineLength > 50;
```

_(using a lot of variables for clarity; I'd probably be a bit more pithy in the actual code)_

mdm317: **My initial thought.**

To detect length overflow, I thought I should calculate the width of the `code` element rather than its length. Because the width of the `code` element varies depending on the viewport. For example, the width of the `code` element is 643px when the viewport width is 1260px, and 718px when the viewport width is 1300px.
Additionally, since each character has a different width, the width of the last code line's element also needs to be calculated.
So I thought I would need two refs: one for the `code` container and one for the last code line, and need to add a resize event for both refs.
I think this behavior is a bit complicated.
For a simple solution, so i just add padding when playgroun button has appeard
**Your suggestion**
> const showButton = lastLineLength > 50;

But the approach you suggested of limiting the number of characters to 50 also seems like a simple solution.
Let me know your thoughts on whether I should limit by the length of the string, consider its width, or look for another approach.

JoshuaKGoldberg: > since each character has a different width

That shouldn't be the case, the code is all in monospace font.

But, yeah, I think just going by the length of the string as a rough heuristic should be "good enough". Most code blocks end with one or two lines of `}s` or other <10 character things.

mdm317: Thank you for sharing your thoughts!
> That shouldn't be the case, the code is all in monospace font.

That's right, I was wrong.


I’ll change it as you suggested.

packages/website/src/theme/CodeBlock/Content/String.tsx
```diff
@@ -90,6 +92,7 @@ export default function CodeBlockString({
               <code
                 className={clsx(
                   styles.codeBlockLines,
+                  needsMorePadding && styles.codeBlockLinesMorePadding,
```
JoshuaKGoldberg, line 95: Ahh, this is still adding more padding even when there's not a playground button. Example from https://deploy-preview-10019--typescript-eslint.netlify.app/getting-started/#step-1-installation:

Screenshot of the linked 'Step 1: Installation' section. The 'npm install' codeblock has too extra vertical padding, but no 'Show in Playground' button.

I think we should only add the padding for codeblocks that have the playground button.

JoshuaKGoldberg (changes requested):
Getting there 😎

packages/website/src/theme/CodeBlock/Content/String.tsx
mdm317: I missed this part again. Thanks for reviewing it

JoshuaKGoldberg (approved):
Looks great to me, thanks for iterating on this @mdm317! 🙌

PR #10040, opened 2024-09-23
fix(eslint-plugin): [no-deprecated] max callstack exceeded when class implements itself

### LINKED ISSUE #10031: Bug: [no-deprecated] Maximum call stack size exceeded when class implements itself

### Issue Description

If you have a class in TypeScript that implements itself (i.e. `class Foo implements Foo`), inside which is a getter method (i.e. `get bar(): number { return 42; }`) that is used by another method (i.e. `baz(): number { return this.bar; }`), then upon running `npx eslint Foo.ts` I get the error "RangeError: Maximum call stack size exceeded" while running rule `@typescript-eslint/no-deprecated`, which occurs while linting the line containing `return this.bar`.

Now, why exactly you would want a class to implement itself is quite beyond me, but I accidentally had that in my code and it took me quite a while to figure out what was going on. TypeScript itself seems to have no problem with it at all. I have no problem at all if the resolution is to detect this and then give an error, but that error must be descriptive.

My reproduction repo is quite minimal, so I'm not sure why I was unable to reproduce it in the playground. I'm not too good with node-ish setups, I guess.

https://github.com/FWDekker/typescript-eslint-no-deprecated-bug

### Repro Steps

1) clone the repo
2) `npm install`
3) `npx eslint Foo.ts`

### Versions

| `@typescript-eslint/eslint-plugin`     | `8.6.0` |
| `@typescript-eslint/parser`            | `8.6.0` |
| `@typescript-eslint/scope-manager`     | `8.6.0` |
| `@typescript-eslint/typescript-estree` | `8.6.0` |
| `@typescript-eslint/type-utils`        | `8.6.0` |
| `@typescript-eslint/utils`             | `8.6.0` |
| `TypeScript`                           | `5.5.4` |
| `ESLint`                               | `9.11.0` |
| `node` (Debian package)                                 | `20.17.0` |

tarunrajput:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10031

## Overview

wraps `getJsDocTags` call in a try-catch

### DIFF
--- a/packages/eslint-plugin/src/rules/no-deprecated.ts
+++ b/packages/eslint-plugin/src/rules/no-deprecated.ts
@@ -152,9 +152,14 @@ export default createRule({
     function getJsDocDeprecation(
       symbol: ts.Signature | ts.Symbol | undefined,
     ): string | undefined {
-      const tag = symbol
-        ?.getJsDocTags(checker)
-        .find(tag => tag.name === 'deprecated');
+      let jsDocTags: ts.JSDocTagInfo[] | undefined;
+      try {
+        jsDocTags = symbol?.getJsDocTags(checker);
+      } catch {
+        // workaround for https://github.com/microsoft/TypeScript/issues/60024
+        return;
+      }
+      const tag = jsDocTags?.find(tag => tag.name === 'deprecated');
 
       if (!tag) {
         return undefined;
--- a/packages/eslint-plugin/tests/rules/no-deprecated.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-deprecated.test.ts
@@ -253,6 +253,20 @@ ruleTester.run('no-deprecated', rule, {
       },
     },
     'call();',
+
+    // this test is to ensure the rule doesn't crash when class implements itself
+    // https://github.com/typescript-eslint/typescript-eslint/issues/10031
+    `
+      class Foo implements Foo {
+        get bar(): number {
+          return 42;
+        }
+
+        baz(): number {
+          return this.bar;
+        }
+      }
+    `,
   ],
   invalid: [
     {


packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -116,17 +116,21 @@ export default createRule({
     function getJsDocDeprecation(
       symbol: ts.Signature | ts.Symbol | undefined,
     ): string | undefined {
-      const tag = symbol
-        ?.getJsDocTags(checker)
-        .find(tag => tag.name === 'deprecated');
+      try {
```
kirkwaiblinger, line 119: Let's try to be as surgical about the try/catch as we can - can we make it so that essentially _only_ the `getJsDocTags()` call is wrapped?

packages/eslint-plugin/src/rules/no-deprecated.ts
kirkwaiblinger, line 1: let's add the test case from the issue to the `valid` cases in packages/eslint-plugin/tests/rules/no-deprecated.test.ts with a note that it's just there to ensure it doesn't crash and link the github issue https://github.com/typescript-eslint/typescript-eslint/issues/10031.

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -116,17 +116,21 @@ export default createRule({
...
+          return undefined;
+        }
 
-      const displayParts = tag.text;
+        const displayParts = tag.text;
 
-      return displayParts ? ts.displayPartsToString(displayParts) : '';
+        return displayParts ? ts.displayPartsToString(displayParts) : '';
+      } catch {
+        return;
```
kirkwaiblinger, line 132: could you add a comment that says something like `// workaround for https://github.com/microsoft/TypeScript/issues/60024`?

kirkwaiblinger (changes requested):

kirkwaiblinger:
awesome, thanks! 🙂 

gif about infinite loop

kirkwaiblinger (approved):

auvred (approved):
Thanks!

PR #10046, opened 2024-09-23
fix(eslint-plugin): stop warning on @ts-nocheck comments which aren't at the beginning of the file

### LINKED ISSUE #8753: Bug: [ban-ts-comment] false positive when @ts-nocheck is not at the beginning of the file

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.2&fileType=.ts&code=FBA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AIwEMnb9layAewC2wxC3RRE0aIOiRwYAL4glQA&tsconfig=&tokens=false

### Repro Code

```TypeScript
const a = 1

// @ts-nocheck - should not be reported

// TS error is not actually suppressed
const b: string = a
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/ban-ts-comment": "error"
  }
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

`// @ts-nocheck` should not be reported because valid TS pragmas should be placed at the beginning of the file

### Actual Result

`// @ts-nocheck` and `// @ts-check` are reported even if they are not valid TS pragmas

### Additional Info

Follow up for https://github.com/typescript-eslint/typescript-eslint/pull/8416

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8753

## Overview

This PR resolves #8753 and stops warning on `// @ts-nocheck` comments which are not at the beginning of the file:

```ts
const a = 1;

// @ts-nocheck - should not be reported

// TS error is not actually suppressed
const b: string = a;
```

### DIFF
--- a/packages/eslint-plugin/src/rules/ban-ts-comment.ts
+++ b/packages/eslint-plugin/src/rules/ban-ts-comment.ts
@@ -181,13 +181,6 @@ export default createRule<[Options], MessageIds>({
       );
     }
 
-    function isPositionEarlierThan(
-      a: TSESTree.Position,
-      b: TSESTree.Position,
-    ): boolean {
-      return a.line <= b.line || a.column < b.column;
-    }
-
     return {
       Program(node): void {
         const firstStatement = node.body.at(0);
@@ -204,7 +197,7 @@ export default createRule<[Options], MessageIds>({
           if (
             directive === 'nocheck' &&
             firstStatement &&
-            isPositionEarlierThan(firstStatement.loc.start, comment.loc.start)
+            firstStatement.loc.start.line <= comment.loc.start.line
           ) {
             return;
           }
--- a/packages/eslint-plugin/src/rules/ban-ts-comment.ts
+++ b/packages/eslint-plugin/src/rules/ban-ts-comment.ts
@@ -190,9 +190,7 @@ export default createRule<[Options], MessageIds>({
 
     return {
       Program(node): void {
-        const firstStatement = node.body[0] as
-          | TSESTree.ProgramStatement
-          | undefined;
+        const firstStatement = node.body.at(0);
 
         const comments = context.sourceCode.getAllComments();
 
--- a/packages/eslint-plugin/src/rules/ban-ts-comment.ts
+++ b/packages/eslint-plugin/src/rules/ban-ts-comment.ts
@@ -181,8 +181,19 @@ export default createRule<[Options], MessageIds>({
       );
     }
 
+    function isPositionEarlierThan(
+      a: TSESTree.Position,
+      b: TSESTree.Position,
+    ): boolean {
+      return a.line <= b.line || a.column < b.column;
+    }
+
     return {
-      Program(): void {
+      Program(node): void {
+        const firstStatement = node.body[0] as
+          | TSESTree.ProgramStatement
+          | undefined;
+
         const comments = context.sourceCode.getAllComments();
 
         comments.forEach(comment => {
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/ban-ts-comment.ts
```diff
@@ -192,6 +201,14 @@ export default createRule<[Options], MessageIds>({
           }
           const { directive, description } = match;
 
+          if (
+            directive === 'nocheck' &&
+            firstStatement &&
+            isPositionEarlierThan(firstStatement.loc.start, comment.loc.start)
```
auvred, line 207:
```suggestion
            firstStatement.loc.start.line <= comment.loc.start.line
```

[Bug] If `// @ts-nocheck` comment is placed on the line before actual code, but its `loc.start.column` is greater than `firstStatement.loc.start.column`, the `nocheck` directive is not reported, even though this is a valid TS pragma and it effectively suppresses all TS errors.

```ts
 /// @ts-nocheck
const a: true = false
```

[playground link](https://deploy-preview-10046--typescript-eslint.netlify.app/play#ts=5.5.2&fileType=.tsx&code=AQejwAQFwZwWgHYHsDGALApig1gKBUgjFMAIYBcwUATgK4bAC8wAZqQDYwZA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AIwEMnb9layAewC2wxC3RRE0aIOiRwYAL4glQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

I don't think we need to compare `loc` columns here. Since singleline comments should be placed either on their own line or at the end of a line.
And if `// @ts-nocheck` is placed at the end of a line, it's no longer a valid TS pragma.

auvred (changes requested):

packages/eslint-plugin/src/rules/ban-ts-comment.ts
ronami: Oh oops. 🙈 

I understand that checking columns alone is a bug, my intention was to not report the following:

```ts
/* @ts-nocheck - should not be reported */ const a = 1;
```

The function I was going to write was something like:

```ts
function isPositionEarlierThan(
  a: TSESTree.Position,
  b: TSESTree.Position,
): boolean {
  if (a.line > b.line) {
    return false;
  }
  
  return a.line === b.line && a.column < b.column;
}
```

But I can see the block comment with `ts-nocheck` doesn't disable type-checking [and the lint rule doesn't report this already](https://github.com/typescript-eslint/typescript-eslint/blob/896c7312dfe8374312df9404cb11db8378cd091c/packages/eslint-plugin/src/rules/ban-ts-comment.ts#L109-L110).

I've changed my code accordingly, thanks 🙏

auvred: I think it's worth including my example in the test suite as well. That way we can prevent regressions in the future.

packages/eslint-plugin/tests/rules/ban-ts-comment.test.ts
```diff
@@ -1127,6 +1119,22 @@ if (false) {
         },
       ],
     },
+    {
+      // comment's column > first statement's column
+      // eslint-disable-next-line @typescript-eslint/internal/plugin-test-formatting
```
ronami, line 1124: I used an `eslint-disable` comment over `noFormat()` [since it doesn't handle disabling indentation checks](https://github.com/typescript-eslint/typescript-eslint/blob/94c5484f747be901f1610fc37192714a619bb355/packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts#L392-L398).

auvred (approved):
:100:

PR #10049, opened 2024-09-24
fix(eslint-plugin): correct use-at-your-own-risk type definitions

### LINKED ISSUE #9967: Bug: type definitions for use-at-your-own-risk/eslint-recommended-raw are missing

### Relevant Package

typescript-eslint

### Playground Link

_No response_

### Repro Code

```TypeScript
// @ts-check
import { default as tseslintRecommendedRaw } from "@typescript-eslint/eslint-plugin/use-at-your-own-risk/eslint-recommended-raw"; // ts7016
import eslintPluginSvelte from "eslint-plugin-svelte";
import svelteEslintParser from "svelte-eslint-parser";
import tseslint from "typescript-eslint";

// This is a part of ESLint flat config.
export const svelteConfig = tseslint.config(
  ...eslintPluginSvelte.configs["flat/recommended"],
  {
    files: ["*.svelte", "**/*.svelte"],
    languageOptions: {
      parser: svelteEslintParser,
      parserOptions: {
        extraFileExtensions: [".svelte"],
      },
    },
    rules: {
      // Re-apply TSLint rules to svelte code blocks
      // to avoid ESLint false-positives on, e.g., no-undef.
      ...tseslintRecommendedRaw.default("minimatch").rules,
    },
  },
  {
    rules: {
      // override/add rules settings here, such as:
      // 'svelte/rule-name': 'error'
    },
  },
);
```

### ESLint Config

```javascript
N/A
```

### tsconfig

```jsonc
N/A
```

### Expected Result

I should be able to import `@typescript-eslint/eslint-plugin/use-at-your-own-risk/eslint-recommended-raw` without any typing error.

### Actual Result

I see TS7016.

> Could not find a declaration file for module '@typescript-eslint/eslint-plugin/use-at-your-own-risk/eslint-recommended-raw'. '.../node_modules/@typescript-eslint/eslint-plugin/dist/configs/eslint-recommended-raw.js' implicitly has an 'any' type.
>  Try `npm i --save-dev @types/typescript-eslint__eslint-plugin` if it exists or add a new declaration (.d.ts) file containing `declare module '@typescript-eslint/eslint-plugin/use-at-your-own-risk/eslint-recommended-raw';`ts(7016)

### Additional Info

Discord thread: https://discord.com/channels/1026804805894672454/1283088823298818079

I'm trying to set up eslint with typescript support for script blocks in svelte files. However, I saw eslint complaining about `no-undef` violations inside svelte files (ts files works well), which should have been disabled by `tseslint.configs.recommendedTypeChecked` per my configuration. After checking the definition of `recommendedTypeChecked`, it seems [these presets are only applied to .js/.ts files](https://github.com/typescript-eslint/typescript-eslint/blob/4d31ebe71ac26612a129ac4be98692b11387145e/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts#L21).

So I guess this could be worked around if I can somehow re-apply the rules once more, with `*.svelte` as `files`, and that's why I'm resorting to the raw rules exported here.

By checking its package.json file, I believe the specified .d.ts files are actually missing from the package:

package.json: https://unpkg.com/browse/@typescript-eslint/eslint-plugin@8.5.0/package.json
root folder: https://unpkg.com/browse/@typescript-eslint/eslint-plugin@8.5.0/

As this is js file, and eslint won't check for typings when importing the file, there is no actual production impact, but there will be noise when in IDE.

Another possible workaround mentioned by Josh is
```ts
const tseslintPlugin = require("@typescript-eslint/eslint-plugin");

console.log(tseslintPlugin.configs['eslint-recommended'].overrides[0].rules);
```

### Versions

| `@typescript-eslint/eslint-plugin`     | `8.5.0` |

aryaemami59:
## PR Checklist

- [X] Addresses an existing open issue: fixes #9967.

### **Before:**

tseslint-eslint-plugin-before

### **After**:

tseslint-eslint-plugin-after

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -5,6 +5,7 @@
   "files": [
     "dist",
     "docs",
+    "eslint-recommended-raw.d.ts",
     "index.d.ts",
     "rules.d.ts",
     "package.json",
--- a/packages/eslint-plugin/rules.d.ts
+++ b/packages/eslint-plugin/rules.d.ts
@@ -41,7 +41,7 @@ import type {
   RuleRecommendationAcrossConfigs,
 } from '@typescript-eslint/utils/ts-eslint';
 
-export interface ESLintPluginDocs {
+interface ESLintPluginDocs {
   /**
    * Does the rule extend (or is it based off of) an ESLint code rule?
    * Alternately accepts the name of the base rule, in case the rule has been renamed.
@@ -62,16 +62,25 @@ export interface ESLintPluginDocs {
   requiresTypeChecking?: boolean;
 }
 
-export type ESLintPluginRuleModule = RuleModuleWithMetaDocs<
+type ESLintPluginRuleModule = RuleModuleWithMetaDocs<
   string,
   readonly unknown[],
   ESLintPluginDocs
 >;
 
-export type TypeScriptESLintRules = Record<
+type TypeScriptESLintRules = Record<
   string,
   RuleModuleWithMetaDocs<string, unknown[], ESLintPluginDocs>
 >;
 
 declare const rules: TypeScriptESLintRules;
-export default rules;
+
+declare namespace rules {
+  export type {
+    ESLintPluginDocs,
+    ESLintPluginRuleModule,
+    TypeScriptESLintRules,
+  };
+}
+
+export = rules;
--- a/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts
+++ b/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts
@@ -7,7 +7,7 @@
  * - disables rules from eslint:recommended which are already handled by TypeScript.
  * - enables rules that make sense due to TS's typechecking / transpilation.
  */
-export default (
+const config = (
   style: 'glob' | 'minimatch',
 ): {
   files: string[];
@@ -44,3 +44,5 @@ export default (
     'prefer-spread': 'error', // ts transpiles spread to apply, so no need for manual apply
   },
 });
+
+export = config;
--- a/packages/eslint-plugin/src/rules/index.ts
+++ b/packages/eslint-plugin/src/rules/index.ts
@@ -128,7 +128,7 @@ import unboundMethod from './unbound-method';
 import unifiedSignatures from './unified-signatures';
 import useUnknownInCatchCallbackVariable from './use-unknown-in-catch-callback-variable';
 
-export default {
+const rules = {
   'adjacent-overload-signatures': adjacentOverloadSignatures,
   'array-type': arrayType,
   'await-thenable': awaitThenable,
@@ -258,3 +258,5 @@ export default {
   'unified-signatures': unifiedSignatures,
   'use-unknown-in-catch-callback-variable': useUnknownInCatchCallbackVariable,
 } satisfies Linter.PluginRules;
+
+export = rules;
--- a/packages/eslint-plugin/eslint-recommended-raw.d.ts
+++ b/packages/eslint-plugin/eslint-recommended-raw.d.ts
@@ -2,4 +2,4 @@ declare const config: (style: 'glob' | 'minimatch') => {
   files: string[];
   rules: Record<string, 'off' | 'warn' | 'error'>;
 };
-export default config;
+export = config;
--- a/packages/eslint-plugin/src/index.ts
+++ b/packages/eslint-plugin/src/index.ts
@@ -21,27 +21,31 @@ const { name, version } = require('../package.json') as {
   version: string;
 };
 
+const configs = {
+  all,
+  base,
+  'disable-type-checked': disableTypeChecked,
+  'eslint-recommended': eslintRecommended,
+  recommended,
+  /** @deprecated - please use "recommended-type-checked" instead. */
+  'recommended-requiring-type-checking': recommendedTypeChecked,
+  'recommended-type-checked': recommendedTypeChecked,
+  'recommended-type-checked-only': recommendedTypeCheckedOnly,
+  strict,
+  'strict-type-checked': strictTypeChecked,
+  'strict-type-checked-only': strictTypeCheckedOnly,
+  stylistic,
+  'stylistic-type-checked': stylisticTypeChecked,
+  'stylistic-type-checked-only': stylisticTypeCheckedOnly,
+};
+
+const meta = {
+  name,
+  version,
+};
+
 export = {
-  configs: {
-    all,
-    base,
-    'disable-type-checked': disableTypeChecked,
-    'eslint-recommended': eslintRecommended,
-    recommended,
-    /** @deprecated - please use "recommended-type-checked" instead. */
-    'recommended-requiring-type-checking': recommendedTypeChecked,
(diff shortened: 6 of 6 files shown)


packages/eslint-plugin/package.json
```diff
@@ -27,6 +28,18 @@
...
+      "use-at-your-own-risk/rules": [
+        "./rules.d.ts"
+      ],
+      "use-at-your-own-risk/eslint-recommended-raw": [
+        "./eslint-recommended-raw.d.ts"
+      ]
+    }
+  },
+  "types": "./index.d.ts",
+  "main": "./dist/index.js",
```
bradzacher, line 42: We purposely do not list these keys. Please remove them.

packages/eslint-plugin/package.json
```diff
@@ -27,6 +28,18 @@
...
+  "typesVersions": {
+    "*": {
+      "use-at-your-own-risk/rules": [
+        "./rules.d.ts"
+      ],
+      "use-at-your-own-risk/eslint-recommended-raw": [
+        "./eslint-recommended-raw.d.ts"
+      ]
+    }
+  },
```
bradzacher, line 40: Why do we need a duplicate definition here? These are already defined via the `exports` key.

packages/eslint-plugin/rules.d.ts
```diff
@@ -62,16 +62,25 @@ export interface ESLintPluginDocs {
...
 >;
 
-export type TypeScriptESLintRules = Record<
+type TypeScriptESLintRules = Record<
   string,
   RuleModuleWithMetaDocs<string, unknown[], ESLintPluginDocs>
 >;
 
 declare const rules: TypeScriptESLintRules;
-export default rules;
```
bradzacher, line 77: Why did you switch this from an `export default`?

packages/eslint-plugin/src/configs/eslint-recommended-raw.ts
```diff
@@ -7,7 +7,7 @@
  * - disables rules from eslint:recommended which are already handled by TypeScript.
  * - enables rules that make sense due to TS's typechecking / transpilation.
  */
-export default (
```
bradzacher, line 10: Why did you switch this from an `export default`?

packages/eslint-plugin/src/rules/index.ts
```diff
@@ -128,7 +128,7 @@ import unboundMethod from './unbound-method';
 import unifiedSignatures from './unified-signatures';
 import useUnknownInCatchCallbackVariable from './use-unknown-in-catch-callback-variable';
 
-export default {
```
bradzacher, line 131: Why did you switch this from an `export default`?

bradzacher (changes requested):

packages/eslint-plugin/package.json
aryaemami59: Done!

packages/eslint-plugin/package.json
aryaemami59: `Node10` module resolution would fail without them

tseslint-without-typesVersions

packages/eslint-plugin/rules.d.ts
aryaemami59: With a `default` export we get:

tseslint-rules-with-default-export

Which basically means in ESM files we will have an extra `.default` property, so for exmaple if you have an ESM file like `foo.mjs` and you try to import `@typescript-eslint/eslint-plugin/use-at-your-own-risk/rules`:

```js
import rules from '@typescript-eslint/eslint-plugin/use-at-your-own-risk/rules'
console.log(Object.keys(rules)) // We will get [ 'default' ]
```

The `console.log(Object.keys(rules))` should give us all the rules but instead we get `[ 'default' ]` which is not correct.

packages/eslint-plugin/src/rules/index.ts
aryaemami59: See [this comment](https://github.com/typescript-eslint/typescript-eslint/pull/10049#discussion_r1775961226).

packages/eslint-plugin/src/configs/eslint-recommended-raw.ts
aryaemami59: Basically the same answer as [this comment](https://github.com/typescript-eslint/typescript-eslint/pull/10049#discussion_r1775961226), with a `default` export we get:

tseslint-eslint-recommended-raw-with-default-export

packages/eslint-plugin/package.json
bradzacher: > Node10 module resolution would fail without them

We do not support Node10.
Node10 as a version of node is LONG deprecated.
Node10 as a module resolution is also bad because it means your module resolution doesn't match reality.

aryaemami59: That's understandable. However, from my perspective, the benefits of supporting `Node10` module resolution still outweigh the cons for now. While `Node10` itself is deprecated, there are still a significant number of codebases that rely on it. A quick search on GitHub reveals that there are more projects using `Node10` module resolution than `Node16`, `NodeNext`, and `Bundler` combined [36k using `Node10`](https://github.com/search?q=path%3A%2F%5Etsconfig.*.json%24%2F+language%3Ajson+%2F%22moduleResolution%22%3A+%22node%22%2F+OR+%2F%22moduleResolution%22%3A+%22node10%22%2F&type=code) versus [25k using other resolutions](https://github.com/search?q=path%3A%2F%5Etsconfig.*.json%24%2F+language%3Ajson+%2F%22moduleResolution%22%3A+%22node16%22%2F+OR+%2F%22moduleResolution%22%3A+%22nodenext%22%2F+OR+%2F%22moduleResolution%22%3A+%22bundler%22%2F&type=code).

Given that supporting it is relatively simple using the [`typesVersions`](https://github.com/andrewbranch/example-subpath-exports-ts-compat/tree/main/examples/node_modules/types-versions-wildcards) strategy, it seems like a low-effort way to ensure compatibility for a larger portion of users, at least in the short term.

bradzacher: The better search would be to separate `node` from `node10`.
You'll find that there's just 298 hits for `node10` -- i.e. 298 projects that have INTENTIONALLY opted to use node v10 resolution.

The remaining 35.8k hits for `node` are more than likely people that setup their config a long time ago and haven't touched it because it looks like it's the right thing to use. People don't realise that it's an alias for `node10` and that it's the wrong thing to use.

And you'll find that almost all of these codebases could migrate to `Node16` or `Bundler` without any bugs.

We went through this already when we dropped support for `Node10` resolution. Many users complained because they thought we had broken types, but we educated them about their incorrect configs, they updated it and people improved their setups.

We do not want to add it back because it is incorrect and misrepresents the types and accessibility of our package.

bradzacher (changes requested):
Everything looks good.
Just need to remove `typesVersions` then we're good to go!

packages/eslint-plugin/package.json
aryaemami59: Fair enough, I'll remove it.

bradzacher (approved):
brilliant - thanks for your work on this!

Issue #2616, opened 2020-09-30
Rule suggestion: use of `never`


NWilson:
**Repro**

The following code is clearly bad, but compiles just fine, because the `typeof` condition narrows the type of `a` to `never`, allowing it to be passed to a function that clearly can't accept it. (This is distilled from a real example in a Microsoft codebase.)

```TS
function useTheValue(a: { foo: number }) {
  return a.foo;
}

export function testFunc(a: string | number) {
  if (typeof a === 'boolean') {
    return useTheValue(a);
  }
  return 0;
}
```

I apologize if it exists - but I can't find a `typescript-eslint` rule that will detect this.

I think the rule I want is "disallow any use of a `never`-typed value except as an immediate operand to `return`".

bradzacher:
this could definitely be added to our `no-unsafe-*` class of rules.
Assigning `never` to something is very unsafe!

CreativeTechGuy:
I found this when looking for something similar myself and want to try my hand at implementing it. I just want to confirm what I'm thinking of implementing and get suggestions ahead of time.

I think the rule could be very simple and error on the use of `never` in any case. This would disable the use of never as a return value but I believe there'd be overlap in the people who want this rule and also want to prohibit a function that never returns. (I honestly haven't seen a real example of needing to explicitly type `never`.)

Maybe if we want to get fancy the rule could have an option to allow never return values if that's something a lot of people might want. But that could also be a follow up if the need is there.

For a name, this seems like it'd be similar to `no-namespace` so `no-never` would probably make sense although I admit the name does read a bit weird.

bradzacher:
Instead of creating a new rule entirely, IMO it'd be much better to add handling for `never` to our existing `no-unsafe-*` rules.

The rules are already setup to prevent unsafe operations on an `any`, so it's a natural extension to also look for `never`.

CreativeTechGuy:
I played around with prototyping it last night and I think the functionality I have is a bit different. It identifies any variable which is of type `never`. Either through being scoped down to `never` or being explicitly typed as `never`.

In this example below:
```ts
const obj = "abc";

if (typeof obj === "undefined") {
    const test = obj; // Both test and obj on this line are errors as they have the never type
    console.log(test); // test is an error here as it has the never type
}

function neverReturn (): never { // This is fine as it's a return type and not a variable type
    throw new Error("abc");
}

let num: never; // num is an error as it explicitly has the never type
```

[...]

PR #10051, opened 2024-09-24
feat(eslint-plugin): new rule `no-unsafe-type-assertion`

### LINKED ISSUE #7173: Rule proposal: no-unsafe-type-assertion

### Description

One of the hardest thing to explain to people not experienced in TS is when it's ok to use `as`. I default to tell people "don't use `as`", and I constantly have to deal with that in code reviews. I'd love to have a rule that only allows type assertions that widen a type. For example, `as unknown` is always totally ok. `as string | number` is safe on a `string`, but unsafe in a `string | boolean`.

This rule should only allow doing `x as Type` in the cases where the type of `x` is assignable to `type`. So, the rule of thumb would be: if `const a: Type = x` is fine, then `x as Type` is fine. Otherwise it's not.

### Fail Cases

```typescript
const x = 'something';
const y = x as unknown as number; // Type `unknown` is not assignable to type `number` 🚫

//-----------

function f() {
  return Math.random() < 0.5 ? 42 : 'meh';
}
const z = f() as number; // Type `number | string` is not assignable to type `number` 🚫
```

### Pass Cases

```typescript
const a = 'something';
const b = a as unknown; // Type `string` is assignable to type `unknown` ✅

//-----------

function g() {
  return Math.random() < 0.5 ? 42 : 'meh';
}
const h = f() as number | string | boolean; // Type `number | string` is assignable to type `number | string | boolean` ✅

//-----------

const person = {
    name: 'Chuck',
    lastName: 'Norris' as string | undefined // Type `Norris` is assignable to type `string | undefined`  ✅
}
person.lastName = undefined; // this is fine because of the type assertion (just showing why we'd want the `as string | undefined`)
```

### Additional Info

Here's a[ link to a TS playground](https://www.typescriptlang.org/play?#code/PQKgtBldkAQCoAsCmBnZtWIPYFcA2AJrAOYCWAbhgIYB2syATo9o7DB9CMAFA8DG2WqgAusAB6wAvLADkqbAFtkIxGVolZAbgFDRsAJ7SJsaqli5aAa1rYA7vTOxauRQCMmW2MGAIDABwwAA0sbe1og2DJzWzEzVDISWmo3fAwRbFgRAOCXdyZIwF4NwGq9vh9OSD4AM0t+ETIhWCqACgBKWABvHlhYRhVcRnoAWWpVADpGOkIlNtgAHlgABjGAVlgAflgAFgAmWAAuOWVEbR4AX11hMQAvYxb2pzyPRi8fP0DYIKemWAAfTBEjHUJEi0Wc2DiqASSRSaSymWyHy+rmeRVKPFAFQqCBQ6FMfSa6gwWM43D4giupmM8iUKjUGlOFP0bmM1FM5lCtgcr188Byn1EQI0oPM8USyVS6QR-JC1i5EVggFByMrAEnVWr1RokWZdHp9EQDYajRATKYzdoLZZrTa7A5HZAnHQXJliRB3WaPFE-f6C4F-WBubDYNJ0HnvXJetg+wHAkXs6ESuEZLIy75RgFCkj+wPB5B0SLKjGqrHkvRiQKMBT0GS6nrOajKQ6yADCiFw-CssgANN06-gzCIAHIN5BNwesIGoWTsjN+-6WQjIKpE4hvPlI8fMaJxsUwyXwlNI30af0LpcryJK848CtVsb90TD5TGM-L2jIQhh1RgsFvjAefhqFwPFsCqLIUEPGgoSYTV6GaAArYCxCwew-TsRAjDsZBZGIOw6DEVRgicY8s3nWhFz-QgglaHggA) with the passing and (proposed) failing cases.

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7173

## Overview

This PR addresses #7173 and introduces the new `no-unsafe-type-assertion` rule for using type assertions in a type-safe way.

The rule forbids using type assertions to narrow a type:

```ts
function f() {
  return Math.random() < 0.5 ? 42 : 'oops';
}

// Type `number | string` is not assignable to type `number` 🚫
const z = f() as number;
```

Instead, it only supports type casting for widening a type:

```ts
const items = [1, 2, 3, 4];

// Type `number` is assignable to type `number | undefined` ✅
const number = items[0] as number | undefined;
```

### Non null assertions

Non-null assertions are a way to narrow down a type, as long as the type includes `null` or `undefined`:

```ts
declare const foo: () => string | undefined;
foo()!;
```

Reporting non-null assertions can create a lot of noise, especially on nested member expressions. Since there's already a rule [that forbids non-null assertions](https://typescript-eslint.io/rules/no-non-null-assertion/), I've decided not to report these a second time.

Non-null assertions can be written differently as a type-cast (the two examples are semantically the same, and [`non-nullable-type-assertion-style`](https://typescript-eslint.io/rules/non-nullable-type-assertion-style) will auto-fix between the two):

```ts
declare const foo: () => string | undefined;
foo() as string;
```

Since [`no-non-null-assertion`](https://typescript-eslint.io/rules/no-non-null-assertion) doesn't cover the type-cast way, this rule will report it as unsafe, even though it's essentially the same as the non-null operator (which won't be reported).

This does mean, though, that usage like this will be reported by both this rule and [`non-nullable-type-assertion-style`](https://typescript-eslint.io/rules/non-nullable-type-assertion-style):

```ts
declare const foo: () => string | undefined;
foo() as string;
```

### Possibly long lint errors

The current implementation displays the full names of both the type and the asserted type. If any of them is a big union, it can create long error messages:

```
Unsafe type assertion: type `string` is not assignable to type `"no-redundant-type-constituents" | "no-unsafe-type-assertion" | "no-unsafe-member-access" | "no-unsafe-assignment" | "naming-convention" | "adjacent-overload-signatures" | ... 121 more ... | "use-unknown-in-catch-callback-variable"`
```

While some of them can be reduced (by mimicking TypeScript's errors and only showing the most specific type that doesn't match), some error messages may be truncated.

In some cases, this can get way out of hand. For example, [this type assertion](https://github.com/typescript-eslint/typescript-eslint/blob/6dbf2341751bbafcd37b5af51cd87184eb9853bb/packages/eslint-plugin/src/rules/no-unused-vars.ts#L576-L579) has the following lint error:

```
Unsafe type assertion: type `NonNullable<ClassDeclarationWithName | ClassDeclarationWithOptionalName | FunctionDeclarationWithName | ... 199 more ... | undefined>` is not assignable to type `TSModuleDeclaration`
```

I think this is too verbose, but I see many other rules using similar logic to display type names.

What's your opinion on this?

### Testing this on a large project

I've run the rule locally on `@typescript-eslint`'s repo. There are 98 failures; most seem valid, but I found some issues. Most issues I've encountered are related to how straightforward (or unclear) the error message is rather than its correctness.

I wrote about [long error messages](#possibly-long-lint-errors), but I think it's worth discussing how types are displayed.

The following example:

```ts
interface Foo {
  type: 'foo';
}
interface Bar {
  type: 'bar';
}

declare const foo: Foo | Bar;
const bar = foo as Bar;
```

Will fail with:

```
Unsafe type assertion: type `Foo | Bar` is not assignable to type `Bar`
```

Types are referenced by their name (same for type constraints) rather than explaining why these types don't match.

I'll be happy to hear your opinions on this.

### Forbids a common pattern

A common (though unsafe) pattern is to cast a value to `unknown,` followed by whatever type it should be:

```ts
foo(bar as unknown as number);
```

This escape hatch lets developers cast a value explicitly to something else, knowing it's unsafe.

I think it's reasonable to have an option not to report such cases so the developer doesn't have to use an eslint-ignore comment in addition to a very verbose assertion.

### Relies on TypeScript 5.4 or above

This rule relies on https://github.com/microsoft/TypeScript/pull/56448, which made `isTypeAssignableTo()` public and can't be used with earlier versions of TypeScript.

I added a note about this in the docs, but I don't know if there should be a runtime check along with it. I couldn't find anything on similar rules.


packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
```diff
@@ -0,0 +1,62 @@
...
+        assertedType,
+      );
+
+      if (!isAssertionSafe) {
+        context.report({
+          node,
+          messageId: 'unsafeTypeAssertion',
+          data: {
+            type: checker.typeToString(nodeType),
+            asserted: checker.typeToString(assertedType),
```
ronami, line 48: I used `checker.typeToString()` over `getTypeName()` because with `getTypeName()` string literals get transformed into `string`. This can cause errors like ``type `string` is not assignable to type `string`.`` if a string is casted to a string literal.

packages/eslint-plugin/tests/rules/no-unsafe-type-assertion.test.ts
```diff
@@ -0,0 +1,559 @@
...
+declare const a: () => boolean;
+const b = a() as boolean | number | string;
+    `,
+    `
+declare const a: string;
+const b = a as string;
+    `,
+    `
+declare const a: string;
+const b = <string | number>a;
```
ronami, line 36: Each test is here on two forms: `a as string` and `<string>a`. I did this manually rather than provide a function to do this. Do you think it's necessary to test each case both ways? Should I do this differently?

packages/eslint-plugin/docs/rules/no-unsafe-type-assertion.mdx
```diff
@@ -0,0 +1,61 @@
...
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+If your codebase has many unsafe type assertions, then it may be difficult to enable this rule.
+It may be easier to skip the `no-unsafe-*` rules pending increasing type safety in unsafe areas of your project.
+You might consider using [ESLint disable comments](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) for those specific situations instead of completely disabling this rule.
+
+This rule requires TypeScript v5.4 or above; you should not use it if your TypeScript version is lower.
```
kirkwaiblinger, line 57: This isn't correct - the `isTypeAssignableTo` method is only _publicly documented_ on the API starting in v5.4, but it's available at runtime in all our supported versions. Indeed, that's why we've waited until now to use it! 🙂 

The public visibility of the `isTypeAssignableTo` is nothing but an implementation detail on our end, and not something that affects users, so this adominition about the rule being only available in TS v5.4 and up should be removed. Feel free to [try the rule in the playground using our oldest supported TS version](https://deploy-preview-10051--typescript-eslint.netlify.app/play/#ts=4.8.4&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gDwFz0xgEsUBzeAH3hQFcBbAIxBgG4AoN7eKNAjYsiyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tieQEMAZolpFStPsmSJo%2BShybooAdz7QF4MAF8QmoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false) 🙂

packages/eslint-plugin/tests/rules/no-unsafe-type-assertion.test.ts
```diff
@@ -0,0 +1,918 @@
...
+        },
+      ],
+    },
+    {
+      code: `
+interface Foo {
+  bar: number;
+}
+
+// no additional properties are allowed
```
kirkwaiblinger, line 753: Oh..... Yeah, this may be annoying.

At first glance, at least, I think that this code _should_ be considered valid. I'm guessing that the assignability API surprisingly encodes [excess property checks](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks) into the type system here 🫤.

Note that the following is allowed by the rule.
```ts
interface Foo {
  bar: number;
}

const hmmm = { bar: 1, bazz: 1 }

// no complaint despite additional properties
export const foo = <Foo>hmmm;
```

I wonder if this is a behavior we can control? Maybe we can get the non-object-literal version of the type or something to feed into the assignability check? Not familiar with the details here, but I'd be interested in further investigation and discussion

kirkwaiblinger (changes requested):
Great work on a first pass!  And, I'm personally _super_ excited for this to be getting worked on. 🙂 

Looking forward to seeing where the discussions on all the complicated edge cases go! Hopefully I've given you plenty to think about 😁

packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Agreed in general. It'd be good to see interactions with `satisfies`, tuples, and classes too.

packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
```diff
@@ -0,0 +1,62 @@
...
+  name: 'no-unsafe-type-assertion',
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow type assertions that widen a type',
+      requiresTypeChecking: true,
+    },
+    messages: {
+      unsafeTypeAssertion:
+        'Unsafe type assertion: type `{{type}}` is not assignable to type `{{asserted}}`',
```
JoshuaKGoldberg, line 19: Thinking on the long error messages: we haven't had to really solve this problem before, which is why we don't have good helpers. Most of our error messages either don't mention types or only refer to small names such as identifiers.

For this rule, I think we can generally assume that the explicitly written type should be simple enough to report with. What do you think about this?

> Unsafe type assertion: type '{{type}}' is more narrow than the original type.

I don't love "original type" but 🤷 can't think of anything nicer...

packages/eslint-plugin/tests/rules/no-unsafe-type-assertion.test.ts
JoshuaKGoldberg: This almost feels like a wholly separate kind of complaint: an _"unnecessary"_ rather than an _"unsafe"_ one. Why would someone want to assert an object typed value to a strictly more narrow object type?

Agreed this rule shouldn't report in this case 😬. It's not really _"unsafe"_ per normal TypeScript type theology IMO. The type system would allow passing `foo` to a location that requires `Foo`, e.g.:

```ts
interface Foo {
  bar: number;
}

const hmmm = { bar: 1, bazz: 1 }

let foo: Foo;
foo = hmmm; // no complaint
```

JoshuaKGoldberg (approved):
I agree with Kirk 🙂 

Love how straightforward this implementation is... and it pains me to say we might need to make it more complicated.

packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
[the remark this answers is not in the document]
ronami: This is interesting; I definitely missed it on my initial PR. Some thoughts after looking into some of these:

Adding custom logic on top of `isAssignableTo()`, like not allowing casting to or from an `any`, may require additional checks. For example, should the following also be invalid (I think they should)?

```ts
// promises
declare const a: Promise<string>;
a as Promise<any>;

// arrays
declare const a: string[];
a as any[];

// other box values such as tuples, objects, etc, recursively.
```

Additionally, if they should be invalid, should casts like the following also be (I think they should)?

```ts
// unions 
declare const a: 'hello' | string[];
a as string | any[];
```

I'll be happy to address these if they make sense.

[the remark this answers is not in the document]
kirkwaiblinger: Hmm, these are good questions. Note that some of this logic also exists in [no-unsafe-assignment](https://typescript-eslint.io/rules/no-unsafe-assignment/). I guess ideally we would have consistency between the two in what is considered an unsafe assignment.

My inclination is that in a perfect world we'd have `no-unsafe-type-assertion` ban assignments that are banned by TS (which is basically exactly `checker.isTypeAssignableTo()`) _and_ whatever would be additionally banned by `no-unsafe-assignment` (under a shared implementation if feasible).

I also think that it's ok if the first implementation doesn't include the full logic of `no-unsafe-assignment` and that comes as a followup, to avoid scope creep in implementation, and because what we've posited here is actually somewhat inconsistent with the no-unsafe-assignment rules (for instance no-unsafe-assignment allows `const foo: any = 3` even though we're suggesting here that `3 as any` be disallowed 🤔), so we may want to give further thought to the checks in no-unsafe-assignment or what's being asked here 🤔 🤔.

Curious what others think, too.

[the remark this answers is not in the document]
kirkwaiblinger: @JoshuaKGoldberg @typescript-eslint/triage-team

packages/eslint-plugin/tests/rules/no-unsafe-type-assertion.test.ts
kirkwaiblinger: FWIW - Having kicked this around in the back of my head I'm actually starting to think it's probably fine to ban this as long as we have a special case error message. But either way that requires us to be able to detect or infer that that's _why_ it's flagging. If we can achieve that, then we can bikeshed about special-case error messaging, suppressing, or providing an option to control it 🤷

ronami: Yes! I agree with this. I looked into this a bit more, and I can tell the following works/fails too:

```ts
interface Foo {
  bar: number;
}

// fails
const bar = { bar: 1, bazz: 1 } as Foo;

// works
const foo = { bar: 1, bazz: 1 };
const bar = foo as Foo;
```

The same behavior can be seen on the example on the link @kirkwaiblinger sent ([excess property checks](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks)), as well as assignments with a type annotation (which I assume run `isAssignableTo()` or similar):

```ts
interface Foo {
  bar: number;
}

// fails
const bar: Foo = { bar: 1, bazz: 1 };

// works
const z = { bar: 1, bazz: 1 };
const bazz: Foo = z;
```

I think that, as @kirkwaiblinger explained, TypeScript is making [excess property checks](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks) specifically on object literals. One way around this (as mentioned in the linked article) is assigning the value to a variable.

I'm not sure on this yet, but I think TypeScript may be widening the type as it's assigned to a variable (I'm not sure yet what it means [to widen an object literal](https://github.com/microsoft/TypeScript/blob/e99e6e2e87a6d7f152eb7cc491ce11da1926662b/src/compiler/checker.ts#L25212-L25228)). I can verify that `isTypeAssignableTo()` with the widened type (via `getWidenedType()`, specifically if the type is an object literal), works as you'd expect, and doesn't fails on excess properties.

packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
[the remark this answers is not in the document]
ronami: While I'm at it, I've updated the PR to somewhat match the functionality of `no-unsafe-assignment`:

The rule checks both the type of the expression and the asserted type for `any`s. This means both of the following show up as failures (`any => unknown` is still allowed):

```ts
3 as any;
_any_ as number;
```
I've re-used [the same utility function](https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/type-utils/src/isUnsafeAssignment.ts#L18) the other `no-unsafe-*` rule uses, which has some boxed value checks (promises, arrays).

If the type is unchanged, the rule won't fail, regardless if the type is `any` or `never`. There's no added logic for the unsafe `Function` type.

If an object literal is used as the expression of a type assertion, its widened type is used instead of its original type. Otherwise, TypeScript will include [excess property checks](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks) for the object literal. I'm not sure if this is the correct way to tackle this; [I've described my insights on this here](https://github.com/typescript-eslint/typescript-eslint/pull/10051#discussion_r1800019322).

I also updated the tests and grouped them in a somewhat meaningful way. I didn't see many test files that do this. Please let me know if I should change this to be more like the norm.

packages/eslint-plugin/src/util/isTypeUnchanged.ts
```diff
@@ -0,0 +1,39 @@
...
+      .unionTypeParts(cast)
+      .filter(part => !isTypeFlagSet(part, ts.TypeFlags.Undefined));
+
+    if (uncastParts.length !== castParts.length) {
+      return false;
+    }
+
+    const uncastPartsSet = new Set(uncastParts);
+    return castParts.every(part => uncastPartsSet.has(part));
+  }
```
ronami, line 36: I'm not sure if this is actually necessary for what I'm using this for, and this edge case currently isn't tested on `no-unsafe-type-assertion.test.ts`. I'll be happy to either update this or cover this with a test if the recent changes make sense.

ronami:
This is ready for a re-review. Please note these two comments: https://github.com/typescript-eslint/typescript-eslint/pull/10051#discussion_r1801983188, https://github.com/typescript-eslint/typescript-eslint/pull/10051#discussion_r1800019322.

kirkwaiblinger (changes requested):
This is really outstanding work. I appreciate how neat and organized you've managed to keep the implementation. Very reviewer-friendly 🙂.

Implementation might change still based on the resolution to the last question in https://github.com/typescript-eslint/typescript-eslint/pull/10051/files#r1774180854. But I have confidence that if it does, it'll be easy to change, thanks to your good work!

Only specific change requested is the small bug in the error message.

Also, for what it's worth, I tried this out on a real repo and all the reports I saw made sense.

packages/eslint-plugin/docs/rules/no-unsafe-type-assertion.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+const items = [1, '2', 3, '4'];
+
+const number = items[0] as number | string | undefined;
+```
+
+</TabItem>
+</Tabs>
+
+## When Not To Use It
```
kirkwaiblinger, line 51: could be worth noting that this will cause lots of reports in test code if you stub objects out, too

kirkwaiblinger (commented):
Oh, also noting that checking surrounding `never` doesn't seem to have been implemented. 

However, there is also already an issue, https://github.com/typescript-eslint/typescript-eslint/issues/2616, for including `never`-checking in _all_ our no-unsafe-* rules, and, since you've outsourced much of the logic here to a shared utility, you could argue that we should defer to handling `never` in separate work, in the shared code. 🤔

packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
```diff
@@ -0,0 +1,156 @@
...
+        nodeWidenedType,
+        assertedType,
+      );
+
+      if (!isAssertionSafe) {
+        context.report({
+          node,
+          messageId: 'unsafeTypeAssertion',
+          data: {
+            type: checker.typeToString(expressionType),
```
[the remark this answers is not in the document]
ronami: Oh, thanks! I fixed this along with related tests.

ronami:
> Oh, also noting that checking surrounding `never` doesn't seem to have been implemented.
> 
> However, there is also already an issue, #2616, for including `never`-checking in _all_ our no-unsafe-* rules, and, since you've outsourced much of the logic here to a shared utility, you could argue that we should defer to handling `never` in separate work, in the shared code. 🤔

That was my thought process too, and I started looking into #2616 a couple of days ago. Having similar behavior to `no-unsafe-assignment` made a lot of sense, and both assigning and casting from/to an unsafe type (be it `any` but I think also `never`) seem to have a lot of similarities.

I'll be happy to implement `never` checks as part of this rule if you think it should be included from the start.

packages/eslint-plugin/docs/rules/no-unsafe-type-assertion.mdx
ronami: Oh, this can definitely create some noise. I added a couple of sentences to explain this.

packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
```diff
@@ -0,0 +1,156 @@
...
+      if (!isAssertionSafe) {
+        context.report({
+          node,
+          messageId: 'unsafeTypeAssertion',
+          data: {
+            type: checker.typeToString(assertedType),
+          },
+        });
+      }
+    }
```
kirkwaiblinger, line 146: Following from https://github.com/typescript-eslint/typescript-eslint/pull/10051#discussion_r1794398363, my proposal would be: 

```suggestion
      // fallback to checking assignability
      const isAssertionSafe = checker.isTypeAssignableTo(
        expressionType,
        assertedType,
      );

      if (!isAssertionSafe) {
        const nodeWidenedType = isObjectLiteralType(expressionType)
          ? checker.getWidenedType(expressionType)
          : expressionType;

        const isUnsafeDueToExcessPropertyChecks =
          nodeWidenedType !== expressionType &&
          checker.isTypeAssignableTo(nodeWidenedType, assertedType);

        context.report({
          node,
          messageId: isUnsafeDueToExcessPropertyChecks
            ? 'unsafeTypeAssertionExcessProperties'
            : 'unsafeTypeAssertion',
          data: {
            type: checker.typeToString(assertedType),
          },
        });
      }
```

WDYT?

ronami: Some of my thoughts:

I think we should definitely have a dedicated message if we are to report excess properties as unsafe.

While excess properties can be the cause of unsafe code (see #5955 and https://github.com/typescript-eslint/typescript-eslint/issues/5677#issuecomment-1251772990 for examples), the only example I can think of where this is unsafe with this rule is:

```ts
type X = { foo: string };

function foo(x: X) {
  // ...
}

// calling a function
foo({ foo: '', bar: '' } as X);

// assigning to a variable
const b: X = { foo: '', bar: '' } as X;
```

But since even assigning the object literal to a variable suppresses the issue (and removes the need for a type assertion), I'm not sure how helpful it would be to have the rule warn here (even though this is an unsafe assertion that can lead to bugs).

I personally think this may be a better fit in a separate rule (potentially `no-unsafe-excess-properties`, though this is a different take than what's suggested in #5955) or be baked into the other `no-unsafe-*` rules, so invalid code like the linked examples is warned.

Let me know if you prefer this to be part of the current rule - I'll be happy to update the PR accordingly.

kirkwaiblinger: Ok, no, this is definitely fair. Let's leave it the way you have it!

packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: I agree with Kirk 🙂. IMO the highest priority to start is keeping it conceptually as simple as possible. This resonates with me:

> My inclination is that in a perfect world we'd have no-unsafe-type-assertion ban assignments that are banned by TS (which is basically exactly checker.isTypeAssignableTo(), with the possible exception of object literal excess property checks) and whatever would be additionally banned by no-unsafe-assignment (under a shared implementation if feasible).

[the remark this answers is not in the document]
JoshuaKGoldberg: I agree with @kirkwaiblinger 🙂. IMO the highest priority to start is keeping it conceptually as simple as possible. This resonates with me:

> My inclination is that in a perfect world we'd have no-unsafe-type-assertion ban assignments that are banned by TS (which is basically exactly checker.isTypeAssignableTo(), with the possible exception of object literal excess property checks) and whatever would be additionally banned by no-unsafe-assignment (under a shared implementation if feasible).

> But, that raises the question - why doesn't `no-unsafe-assignment` flag that case? Why doesn't `no-unsafe-assignment` even ban this case?
> 
> ```ts
> declare const s: string;
> const a: any = s;
> ```

I believe it's b), that the rule doesn't look at that direction of type flow. The rule is for _assigning_ `any`, not assigning _to_ `any`.

> then maybe we should be allowing `as any` here too 🤔.

I think that kind of defeats the purpose of the rule a lot more. Maybe as a followup opt-in `allowAsAny` kind of option? But not for the initial implementation IMO.

packages/eslint-plugin/docs/rules/no-unsafe-type-assertion.mdx
```diff
@@ -0,0 +1,61 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-unsafe-type-assertion** for documentation.
+
+This rule forbids using type assertions to narrow a type, as it can lead to unsafe assumptions that bypass TypeScript’s type-checking.
```
JoshuaKGoldberg, line 12: > unsafe assumptions

_What_ unsafe assumptions?

Someone reading this who isn't already familiar with the concepts wouldn't likely understand what it means. I think it'd even be easy to misinterpret this as saying _all_ type assertions are unsafe.

Could you add a section to this page explaining this for someone who doesn't already understand?

JoshuaKGoldberg (commented):
Just posting a thought while https://github.com/typescript-eslint/typescript-eslint/pull/10051/files#r1828085145 is pending.

ronami:
> Just posting a thought while https://github.com/typescript-eslint/typescript-eslint/pull/10051/files#r1828085145 is pending.

Thanks @JoshuaKGoldberg; I'll update this section in a bit. To make sure, is https://github.com/typescript-eslint/typescript-eslint/pull/10051/files#r1828085145 pending for something from me?

JoshuaKGoldberg:
Last I checked it was requesting more test coverage. If that's done then we can re-review. 👍

ronami:
> Last I checked it was requesting more test coverage. If that's done then we can re-review. 👍

The PR has gone through several changes since then (with the help of @kirkwaiblinger ❤️). Other than [your recent request](https://github.com/typescript-eslint/typescript-eslint/pull/10051#discussion_r1835739165) I think the PR is ready for a re-review.

packages/eslint-plugin/src/util/isTypeUnchanged.ts
JoshuaKGoldberg: I think if there's a test case that passes only with this logic, then yeah, let's use it. Otherwise just a `===` is sufficient IMO.

Great extract+reuse!

packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
```diff
@@ -0,0 +1,156 @@
...
+      );
+
+      // consider unchanged type as safe
+      const typeIsUnchanged = isTypeUnchanged(
+        compilerOptions,
+        expressionType,
+        assertedType,
+      );
+
+      if (typeIsUnchanged) {
```
JoshuaKGoldberg, line 70: [Style] Nit: several of these checks both have an explanatory comment _and_ a variable of a similar name. I'd personally remove either or both to be a bit more succinct. I think the code is pretty readable without more explanations.

```suggestion
      if (isTypeUnchanged(compilerOptions, expressionType, assertedType)) {
        return;
      }
```

Just a style nit, not a problem if you ignore it 🙂. The later-on ones around `isUnsafeAssignment` are a little less clear, so I personally might keep one of the two strategies for them in particular...

JoshuaKGoldberg (changes requested):
🙌

packages/eslint-plugin/src/util/isTypeUnchanged.ts
ronami: Thanks! I've reverted the extract and reuse to stick with a simple `===` check. If I see an edge case in which it's necessary I'll add it back 👍

packages/eslint-plugin/src/rules/no-unsafe-type-assertion.ts
ronami: Yeah, I agree that the PR was a bit over-explained with comments. I tried to simplify this and keep the ones I think are important (added one regarding why type-widening is performed which was missing).

Please let me know if this is OK or if I should change anything.

packages/eslint-plugin/docs/rules/no-unsafe-type-assertion.mdx
ronami: I definitely agree! I've changed/expanded the opening paragraph to explain this better. Wdyt?

packages/eslint-plugin/docs/rules/no-unsafe-type-assertion.mdx
```diff
@@ -0,0 +1,63 @@
+---
+description: 'Disallow type assertions that widen a type.'
```
kirkwaiblinger, line 2: (typo)
```suggestion
description: 'Disallow type assertions that narrow a type.'
```

kirkwaiblinger (changes requested):
brilliant, brilliant, brilliant! ❤️ 

brilliant

(this is an approval except for the typo)

kirkwaiblinger (dismissed):

JoshuaKGoldberg (dismissed):
Whoohoo! Thanks!

Really excited to finally have a meaningful rule around unsafe assertions in. This has been a sore spot for _so long_.

PR #10058, opened 2024-09-25
docs: [no-unsafe-call] clarify that you can never safely narrow Function

kirkwaiblinger:
## PR Checklist

## Overview

Followup to https://github.com/typescript-eslint/typescript-eslint/issues/9108#issuecomment-2373995560

cc @controversial feel free to chime in with your thoughts!

Co-authored-by: @controversial

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-unsafe-call.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unsafe-call.mdx
@@ -78,12 +78,58 @@ Note that whereas [no-unsafe-function-type](./no-unsafe-function-type.mdx) helps
 See, for example, the following code:
 
 ```ts
-function unsafe(maybeFunction: unknown): string {
+function callUnsafe(maybeFunction: unknown): string {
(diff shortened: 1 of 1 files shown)


controversial:
This looks great!

I think the key clarification for me would be something explicit along the lines of “there is no safe way to narrow a value whose type you don’t know into something that’s safe to call” 

Maybe it would also be valuable to suggest type assertion within the try/catch as a preferred workaround to suppressing the lint rule? like:
```ts
try {
  (maybeFunction as () => void)();
} catch (e) {
  console.error(
    'Function either could not be called or threw an error when called:',
    e,
  );
}
```
since this way, you still let `no-unsafe-call` check that it *would* be safe to call the function if it had the type you expect it to have

but I think as-is, the PR is a great improvement to the docs for this new case

kirkwaiblinger:
@controversial Good suggestions! I've updated it, and pared down the extraneous information 👍

JoshuaKGoldberg (approved):
🆒

PR #10067, opened 2024-09-27
feat(eslint-plugin): [no-confusing-void-expression] add an option to ignore void<->void

### LINKED ISSUE #8538: Enhancement: [no-confusing-void-expression] Add option to ignore void<->void

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8538

## Overview

This PR addresses #8538 and adds an option (namely `ignoreVoidReturningFunctions`) that allows returning `void` expressions from either:

- Functions explicitly annotated with `void` as one of their return types.
- Their contextual type includes `void`.

I enabled the rule back on `@typescript-eslint`'s monorepo (since it's currently disabled) with the option and had to fix one error that looks valid. Running the rule on `@typescript-eslint`'s monorepo helped me figure out the various edge cases of the rule (hopefully, I didn't miss a whole lot).

I added a "game plan" comment after seeing @kirkwaiblinger do this on several other rules, and it has been helpful to read through.

One open question I have is whether or not this should apply to async functions. I mean, if the following should be valid or not:

```ts
function returnsVoid(): void {}

async function foo(): Promise<void> {
  return returnsVoid();
}
```

Please let me know if this makes sense or if I missed anything.

Edit: Only when I finished most of the work I noticed the linked PR of https://github.com/typescript-eslint/typescript-eslint/pull/8632. I noticed I was missing a bunch of tests and added the ones I was missing, thanks @developer-bandi 🙏


packages/eslint-plugin/docs/rules/no-confusing-void-expression.mdx
```diff
@@ -119,6 +119,24 @@ function doSomething() {
 console.log(void alert('Hello, world!'));
 ```
 
+### `ignoreVoidReturningFunctions`
+
+Whether to ignore returns from functions with explicit `void` return types and functions with contextual `void` return types.
```
JoshuaKGoldberg, line 124: [Docs] I think most users won't understand what _"contextual"_  means. How does this strike you?

```suggestion
Whether to ignore returns from functions with `void` return types when inside a function with a `void` return type.
```

packages/eslint-plugin/docs/rules/no-confusing-void-expression.mdx
```diff
@@ -119,6 +119,24 @@ function doSomething() {
...
+
+It may be preferable to allow functions that explicitly return `void` to return `void` expressions. This exception will also apply to functions with an implicit contextual return type of `void`.
```
JoshuaKGoldberg, line 126: [Docs] It's good to be explicit on _why_ something is or isn't preferable:

```suggestion
Some projects prefer allowing functions that explicitly return `void` to return `void` expressions. Doing so allows more writing more succinct functions.

:::note
This is technically risky as the `void`-returning function might actually be returning a value not seen by the type system.
:::
```

packages/eslint-plugin/src/util/getParentFunctionNode.ts
JoshuaKGoldberg, line 1: [Praise] I like the reuse here 🙂 nice.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -388,5 +415,56 @@ export default createRule<Options, MessageId>({
...
+
+        return tsutils
+          .unionTypeParts(returnType)
+          .some(tsutils.isIntrinsicVoidType);
+      }
+
+      if (
+        ts.isFunctionExpression(functionTSNode) ||
+        ts.isArrowFunction(functionTSNode)
+      ) {
```
JoshuaKGoldberg, line 457: [Style] Nit: a little less code to read & run here, as `getContextualType` is just looking for expressions:

```suggestion
      if (ts.isExpression(functionTSNode)) {
```

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -388,5 +415,56 @@ export default createRule<Options, MessageId>({
...
+      //   - Return based on whether the contextual type includes `void` or not
+
+      const functionTSNode = services.esTreeNodeToTSNodeMap.get(functionNode);
+
+      if (functionTSNode.type) {
+        const returnType = checker.getTypeFromTypeNode(functionTSNode.type);
+
+        return tsutils
+          .unionTypeParts(returnType)
+          .some(tsutils.isIntrinsicVoidType);
```
JoshuaKGoldberg, line 451: [Testing] No tests fail if this is changed to `.every`. I think that means there's some logic being missed?

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
```diff
@@ -388,5 +415,56 @@ export default createRule<Options, MessageId>({
       const type = getConstrainedTypeAtLocation(services, targetNode);
       return tsutils.isTypeFlagSet(type, ts.TypeFlags.VoidLike);
     }
+
+    function isFunctionReturnTypeIncludesVoid(functionType: ts.Type): boolean {
+      const callSignatures = tsutils.getCallSignaturesOfType(functionType);
+
+      return callSignatures.some(signature => {
```
JoshuaKGoldberg, line 422: [Testing] No tests fail if this is changed to `.every`. I think that means there's some logic being missed?

JoshuaKGoldberg (changes requested):
Looks like a great start to me! Mostly requesting changes on docs & testing, the game plan seems very reasonable. 🏈

packages/eslint-plugin/docs/rules/no-confusing-void-expression.mdx
ronami: Yes! This is a lot better, thanks ❤️

packages/eslint-plugin/docs/rules/no-confusing-void-expression.mdx
ronami: Can't agree more on this, thanks 👍

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
ronami: Nice.

ronami: I've seen [this done here](https://github.com/typescript-eslint/typescript-eslint/blob/ef5df2f83bc20810f5b33ce31f691a8766990977/packages/eslint-plugin/src/rules/no-unsafe-return.ts#L104-L105), would it make sense to send a tiny PR to shorten it in a similar way? Looks like it works out locally.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
ronami: @JoshuaKGoldberg is it possible for a function expression (arrow or standard) to have multiple call signatures (other than function overloading)? I've currently changed this part of the code to assume only one call signature is available.

ronami: Okay, I figured this out and added the necessary test. 👍

JoshuaKGoldberg:
@ronami is this ready for re-review?

We normally wait to be explicitly re-requested, so we're holding off posting more comments for now.

ronami:
> @ronami is this ready for re-review?
> 
> We normally wait to be explicitly re-requested, so we're holding off posting more comments for now.

Oh, I wasn't aware of this. Yes! It is ready for review.

JoshuaKGoldberg:
Great thanks!

In the future, you can use the actual GitHub feature: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews#re-requesting-a-review. That'll put it back on our queue. We have automations around it to remove the `awaiting response` label.

packages/eslint-plugin/src/rules/no-confusing-void-expression.ts
JoshuaKGoldberg: If you want, I'd 👍 it!

JoshuaKGoldberg (approved):
Looks great, thanks for all the changes!

Renee Young hand-waving saying "Glorious!"

PR #10069, opened 2024-09-28
fix(eslint-plugin): [return-await] sync the behavior with await-thenable

### LINKED ISSUE #9903: Bug: [return-await] Gives false positive with unknown values

### Playground Link

[link](https://typescript-eslint.io/play/#ts=5.3.3&fileType=.tsx&code=IYZwngdgxgBAZgV2gFwJYHsI2QUxMgCiQGsJ0B3CABWACdgBbALhhLMoEoYBvAKBmy0wPfgJi0cyBLSzBywVMlYRSFanUYBuUQF94qCMAA2R4XzEwomEOiM4AdEfQBzAgHIAkljgHjpmABGTlDEbhy6vDpAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6aRfWaJ2gQwDufSvnRRE0aAHtokcGAC%2BIBUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

### Repro Code

```TypeScript
async function test(unknownParam: unknown) {
  try {
    return await unknownParam;
  } finally {
    console.log('In finally block')
  }
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/return-await": "error",
  },
};
```

### Expected Result

Since `unknownParam` could be a Promise (we don't know), the correct code is `return await unknownParam`, so that the `finally` block will execute after the promise completes, not before.

### Actual Result

The rule does not allow that, and forces you to write incorrect code.

mdm317:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9903

## Overview

sync the behavior with await-thenable.
If the type is any or unknown, we don't determine whether to add await or not.

I moved the below error test case to a valid test case.
```ts
const fn = (): unknown => null;
async function test() {
  return await fn();
}
const fn = (): any => null;
async function test() {
  return await fn();
}
```
And add valid test case 
```ts
async function test(unknownParam: unknown) {
  try {
    return await unknownParam;
  } finally {
    console.log('In finally block');
  }
}
```


packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -307,13 +307,14 @@ export default createRule({
...
+            return;
+          }
           context.report({
             messageId: 'nonPromiseAwait',
             node,
-            ...fixOrSuggest(useAutoFix, {
+            ...fixOrSuggest(true, {
               messageId: 'nonPromiseAwait',
               fix: fixer => removeAwait(fixer, node),
             }),
```
auvred, line 320: [Refactor] Since we don't report suggestions anymore, we don't need to use `fixOrSuggest` here. How about replacing the whole `...fixOrSuggest(...)` call with just `fix: fixer => removeAwait(fixer, node),`?

auvred (changes requested):

packages/eslint-plugin/src/rules/return-await.ts
mdm317: Thanks for reviewing it, and i fixed this code the way you showed me.

packages/eslint-plugin/src/rules/return-await.ts
```diff
@@ -307,16 +307,14 @@ export default createRule({
 
       if (!isThenable) {
         if (isAwait) {
-          // any/unknown could be thenable; do not auto-fix
-          const useAutoFix = !(isTypeAnyType(type) || isTypeUnknownType(type));
-
+          // sync behavior with return-await
```
kirkwaiblinger, line 310:
```suggestion
          // any/unknown could be thenable; do not enforce whether they are `await`ed.
```

kirkwaiblinger:
Looking at this closer, I'm wondering whether we should enforce that `any` and `unknown` _must_ be awaited in an error-handling context, instead of ignoring them 🤔 Can't make up my mind.

kirkwaiblinger (approved):
Looks good to me!

JoshuaKGoldberg (approved):
🆒

Issue #10088, opened 2024-10-02
Bug: @typescript-eslint/no-unnecessary-type-parameters: ESLint crashes on invalid syntax instead of reporting error


beijixiaohu:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

[Playground](https://typescript-eslint.io/play/#ts=5.5.4&fileType=.ts&code=KYDwDg9gTgLgBASwHY2FAZgQwMbDgZRmkwHNh80A3BXOAbwFgAoOOAZ2BgElUBbAHgAqAPgAUdOAGtgATwBc7GFGQkANHEqYANgFdgCwXAC%2BASgUAFKBF4IO-ShAQATYQG5mR5kA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

### Repro Code

```TypeScript
export interface StorageService {
  setItem<T>({ key: string, value: T }): Promise<void>;
}
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/no-unnecessary-type-parameters": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expected ESLint to report a syntax error for the invalid method parameter syntax, such as "Unexpected token, expected )" or a similar parsing error.

### Actual Result
```
Error: Non-null Assertion Failed: Type parameter should be present in scope's variables.
    at …
```

### Additional Info

[...]

PR #9900, opened 2024-08-28
fix(eslint-plugin): [no-unnecessary-type-parameters] fix AST quick path scope analysis

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9889 
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

The existing code was inspecting the `references` on the `Scope` itself, but we want to be inspecting the `references` on the corresponding `Variable`.

 * [docs for `references` property on `Scope`](https://eslint.org/docs/latest/extend/scope-manager-interface#references)
 * [docs for `references` property on `Variable`](https://eslint.org/docs/latest/extend/scope-manager-interface#references-1)

Coauthor credit to Josh for the breadcrumbs in https://github.com/typescript-eslint/typescript-eslint/issues/9889#issuecomment-2313374807

Co-authored-by: @JoshuaKGoldberg

PR #10093, opened 2024-10-02
fix(eslint-plugin): [no-unnecessary-type-parameters] cannot assume variables are either type or value

Josh-Cena:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/10088

## Overview

I thought it was something in the scope manager but it turned out to be really simple: in `setItem<T>(T): T`, the variable `T` has two declarations, and it's both a type and a value. Therefore we can't assume that the type parameters declared aren't also available as values.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
@@ -50,9 +50,7 @@ export default createRule({
         const smTypeParameterVariable = nullThrows(
           (() => {
             const variable = scope.set.get(esTypeParameter.name.name);
-            return variable != null &&
-              variable.isTypeVariable &&
-              !variable.isValueVariable
+            return variable != null && variable.isTypeVariable
               ? variable
               : undefined;
           })(),
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
@@ -971,6 +971,30 @@ const f = <T,>(
       ],
     },
     {
+      code: `
+declare function setItem<T>(T): T;
+      `,
+      errors: [
+        {
+          messageId: 'sole',
+          data: { descriptor: 'function', name: 'T', uses: 'used only once' },
+        },
+      ],
+    },
+    {
+      code: `
+interface StorageService {
+  setItem<T>({ key: string, value: T }): Promise<void>;
+}
+      `,
+      errors: [
+        {
+          messageId: 'sole',
+          data: { descriptor: 'function', name: 'T', uses: 'never used' },
+        },
+      ],
+    },
+    {
       // This isn't actually an important test case.
       // However, we use it as an example in the docs of code that is flagged,
       // but shouldn't necessarily be. So, if you make a change to the rule logic


bradzacher:
> Therefore we can't assume that the type parameters declared aren't also available as values

I think that the rule should probably just bail out of this usecase altogether, rather than trying to make it work.

I think this for two reasons:

(1) realistically it's a SUPER rare case that shouldn't happen in any sane code!
Declaring a parameter that's of the same name as the type parameter is unclear and barely working code.
Additionally the generally accepted naming conventions for parameters is lower case, but the generally accepted naming convention for type parameters is pascal case -- which does mean that it's an even rarer case that you actually use the exact same name even accidentally.

(2) value vs type usages are not the simplest thing to deal with and I doubt that the rule has the logic to handle the distinction.

For example consider the following signatures:
```ts
declare function foo<T>(T: string, a: T): typeof T;
declare function foo<T>(T: string, a: typeof T): typeof T;
declare function foo<T>(T: string, a: typeof T): T;
declare function foo<T>(T: string, a: T): T;
```

The first signature has just 1 usage of the type parameter `T` and so the rule should report.
Only it wouldn't report unless it has the correct handling to distinguish between type and value references.
The second signature is similarly invalid and should report because there's no usages of the type parameter `T`.
The third signature is similarly invalid and should report because there's only 1 usage of the type parameter `T`.

The fourth signature is valid and should be valid in the rule as there is a usage of `T` as a type in two distinct locations.

However there is also another complicated case for the rule to handle:
```ts
declare function foo<T>(T: T, a: string): typeof T;
declare function foo<T>(T: string, a: T): typeof T;
```

Technically the rule shouldn't report on the first signature here because `T` is actually used in 2 distinct locations because `typeof T === T`. However I don't believe that the rule will be able to resolve that relationship properly?

We also can't just add handling or `typeof T` either because the second signature is valid.

---

So with all that said - I think that we should just opt out of handling any type parameter that's both a type AND a value.
There's so much complexity and the rule should just ignore it as a super rare edge-case that won't happen in real / sane code.

Josh-Cena:
By "opt out", do you mean to assume that they are used more than once? Currently the rule's behavior is completely valid—it only collects usage of the variable in type positions. Check out the deploy preview: https://deploy-preview-10093--typescript-eslint.netlify.app/play#ts=5.5.2&fileType=.tsx&code=CYUwxgNghgTiAEAzArgOzAFwJYHtVJxwB4AVAPgAoSAueAZwxi1QHMAaeKWkgSlowCeABxA5E8EgG4AUKEiwEKdNjwFi5KrQZNWHLvEEixEvgeGjxU2eGhwkaTLnyJCpSjXqNm7Tv3PHebhk5W0UHFWdXDQ9tbz1uUytreTslR1UXdXdueM8dFlNDCwlgmwV7ZSc1N008uN8TPyNLGSA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

kirkwaiblinger (approved):

kirkwaiblinger:
@bradzacher 

> I doubt that the rule has the logic to handle the distinction.

Looks like it is explicitly checked for....

https://github.com/typescript-eslint/typescript-eslint/blob/3f02687d964c1e7888e5db62256d56cd72c4f281/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts#L140-L148

I think this seems like it was just a straightforward bug on my part in https://github.com/typescript-eslint/typescript-eslint/pull/9900/files, and therefore it makes sense to me to move forward with this fix as is, granted that the rule already passes the proposed edge cases without modification, as @Josh-Cena showed.

PR #10094, opened 2024-10-03
fix(typescript-eslint): propagate `name` field to extended configs in `config` helper

### LINKED ISSUE #10086: Bug: The `name` value is not propagated in the extended configs

### Relevant Package

typescript-eslint

### Playground Link

_No response_

### Repro Code

```TypeScript
No code needed. Just run eslint with the below config and it will error as expected.
```

### ESLint Config

```javascript
import eslint from '@eslint/js'
import tseslint from 'typescript-eslint'

	{
		name: "TypeScript",
		files: ['**/*.{ts,tsx}'],
		extends: [
                        // intentional bug here, these items should use the spread oporator because they are arrays
			tseslint.configs.strictTypeChecked,
			tseslint.configs.stylisticTypeChecked,
		],
		languageOptions: {
			parserOptions: {
				projectService: true,
				tsconfigRootDir: __dirname, // use import.meta.dirname after updating to Node 20+
			},
		},
	},
)
```

### tsconfig

_No response_

### Expected Result

When an error occurs with one of the configs, I expect the output to include the `name` of the config, but when using `extends`, the configs are all "unnamed".

### Actual Result

In the example above I have added the name `TypeScript`. But when it errors I see 
> ConfigError: Config (**unnamed**): Unexpected key "0" found.

### Additional Info

The issue appears to be [here](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-eslint/src/config-helper.ts#L100-L103). It only adds `files` and `ignores` to the extended objects. This should also included the name appended with an index. For example

```js
...(config.name && { name: `${config.name}-${index}` }),
```

### Versions

| `@typescript-eslint/eslint-plugin`     | `X.Y.Z` |
| `@typescript-eslint/parser`            | `X.Y.Z` |
| `@typescript-eslint/rule-tester`       | `X.Y.Z` |
| `@typescript-eslint/scope-manager`     | `X.Y.Z` |
| `@typescript-eslint/typescript-estree` | `X.Y.Z` |
| `@typescript-eslint/type-utils`        | `X.Y.Z` |
| `@typescript-eslint/utils`             | `X.Y.Z` |
| `TypeScript`                           | `X.Y.Z` |
| `ESLint`                               | `X.Y.Z` |
| `node`                                 | `X.Y.Z` |

auvred:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10086

### DIFF
--- a/packages/typescript-eslint/src/config-helper.ts
+++ b/packages/typescript-eslint/src/config-helper.ts
@@ -94,6 +94,7 @@ export function config(
     const extension = {
       ...(config.files && { files: config.files }),
       ...(config.ignores && { ignores: config.ignores }),
+      ...(config.name && { name: config.name }),
     };
 
     return [
--- a/packages/typescript-eslint/src/config-helper.ts
+++ b/packages/typescript-eslint/src/config-helper.ts
@@ -91,17 +91,16 @@ export function config(
       return config;
     }
 
-    const extension = {
-      ...(config.files && { files: config.files }),
-      ...(config.ignores && { ignores: config.ignores }),
-      ...(config.name && { name: config.name }),
-    };
-
     return [
-      ...extendsArr.map(conf => ({
-        ...conf,
-        ...extension,
-      })),
+      ...extendsArr.map(extension => {
+        const name = [config.name, extension.name].filter(Boolean).join('__');
+        return {
+          ...extension,
+          ...(config.files && { files: config.files }),
+          ...(config.ignores && { ignores: config.ignores }),
+          ...(name && { name }),
+        };
+      }),
       config,
     ];
   });
--- a/packages/typescript-eslint/tests/configs.test.ts
+++ b/packages/typescript-eslint/tests/configs.test.ts
@@ -398,4 +398,69 @@ describe('config helper', () => {
       },
     ]);
   });
+
+  it('flattens extended configs with names if base config is unnamed', () => {
+    expect(
+      plugin.config({
+        extends: [
+          { name: 'extension-1', rules: { rule1: 'error' } },
+          { rules: { rule2: 'error' } },
+        ],
+        files: ['common-file'],
+        ignores: ['common-ignored'],
+        rules: { rule: 'error' },
+      }),
+    ).toEqual([
+      {
+        files: ['common-file'],
+        ignores: ['common-ignored'],
+        name: 'extension-1',
+        rules: { rule1: 'error' },
+      },
+      {
+        files: ['common-file'],
+        ignores: ['common-ignored'],
+        rules: { rule2: 'error' },
+      },
+      {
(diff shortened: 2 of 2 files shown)


packages/typescript-eslint/src/config-helper.ts
```diff
@@ -94,6 +94,7 @@ export function config(
     const extension = {
```
kirkwaiblinger, line 94: This will work of course to solve the issue, but wondering if this should be 
`Object.from(Object.entries(config).filter([key, value] => value != null))`
or somesuch instead in order to pass along any other unenumerated properties?

kirkwaiblinger (approved):

packages/typescript-eslint/src/config-helper.ts
bradzacher: Nah -- we're not trying to include _every_ property. This is here to declare a set of overrides that the "base" config applies to the extended config. All other properties come from the extended config.

For example:
```ts
const extended = {
  file: ['a'],
  ignores: ['b'],
  name: 'c',
  rules: { foo: 1 },
};
const base = config({
  extends: [extended],
  file: ['x'],
  ignores: ['y'],
  name: 'z',
  rules: { bar: 1 },
});

// result

const base = [
  {
    file: ['x'],
    ignores: ['y'],
    name: 'z',
    rules: { foo: 1 },
  },
  {
    file: ['x'],
    ignores: ['y'],
    name: 'z',
    rules: { bar: 1 },
  }
];
```

Anything not in this object will just derive from the extended config

packages/typescript-eslint/src/config-helper.ts
```diff
@@ -94,6 +94,7 @@ export function config(
...
       ...(config.files && { files: config.files }),
       ...(config.ignores && { ignores: config.ignores }),
+      ...(config.name && { name: config.name }),
```
bradzacher, line 97: I think we need to be a little smarter here and merge the names to help make things more understandable.

For example -- instead of _overriding_ the name, do like

```ts
name = (() => {
  if (extension.name) {
    if (conf.name) {
      return `${extension.name}__${conf.name}`;
    }
    return extension.name;
  }
  return undefined;
})(); 
```

WDYT?

bradzacher (approved):
nice! thanks!

ntnyq:
ESLint officially recommended using `/` as the config name separator.

Refs: 
- [ESLint docs - configuration-naming-conventions](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-naming-conventions)

Would you mind to change `__` to `/` or make it configurable or left it to be.

JoshuaKGoldberg:
Oh, TIL. Yes that'd probably be good - up for filing an issue @ntnyq? 🙂

Issue #9764, opened 2024-08-09
Rule proposal: prevent an explicit `: boolean` return type annotation if a precidate return type could be inferred


bradzacher:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+new+plugin+rule%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### My proposal is suitable for this project

- [X] My proposal specifically checks TypeScript syntax, or it proposes a check that requires type information to be accurate.
- [X] My proposal is not a "formatting rule"; meaning it does not just enforce how code is formatted (whitespace, brace placement, etc).
- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Description

TS 5.5 introduced [Inferred Type Predicates](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates).
TS will try to infer a predicate return type if:
- The function does not have an explicit return type or type predicate annotation.
- The function has a single return statement and no implicit returns.
- The function does not mutate its parameter.
- The function returns a boolean expression that’s tied to a refinement on the parameter.

The first point is the kicker -- a lot of people (us included) annotate return types explicitly.
Or they'll annotate their own predicates.

We should introduce a rule that warns if:
- the function matches the other 3 conditions
- the function has an explicit return type

This would allow people to lean on TS's inference for much safer code -- explicit type predicates are horridly unsound! TS only enforces that the predicate is assignable to the variable's original type -- not that the predicate is correct.

For example this code is TS valid but it is WILDLY incorrect and will crash at runtime:
```ts
const x = [1, 2, null].filter((x): x is number => x == null);
```

### Fail Cases

```TypeScript
const x = [1, 2, null].filter((x): boolean => x != null);
```

### Pass Cases

```TypeScript
const x = [1, 2, null].filter(x => x != null);

// perhaps configurable to allow this?
// by default inferring the predicate would probably be better as it's more type safe
const x = [1, 2, null].filter((x): x is NonNullable<typeof x> => x != null);
```

kirkwaiblinger:
[...]

PR #10106, opened 2024-10-05
feat(eslint-plugin): [strict-boolean-expressions] check array predicate functions' return statements

### LINKED ISSUE #8016: Enhancement: [strict-boolean-expressions] Check array filter/find predicate functions as boolean locations.

### Description

The `strict-boolean-expressions` does not prevent the following bug:

```ts
const numbers: number[] = [];
const filteredNumbers = numbers.filter((element) => {
  return element; // BUG - SHOULD BE SOME KIND OF BOOLEAN COMPARISON!!!
});
```

This is because it does not examine function return values, it only examines comparisons.

Note that the [`no-unnecessary-condition`](https://typescript-eslint.io/rules/no-unnecessary-condition/) rule is a similar rule, and it does support function return values. Thus, the function-examining logic needs to also be copy-pasted to this rule, so that this rule can catch the bug above.

For reference, the rule should be able to handle the following `Array` methods:

- Array.every
- Array.filter
- Array.find
- Array.findIndex
- Array.findLast
- Array.findLastIndex
- Array.some

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8016

## Overview

This PR addresses #8016 and updates `strict-boolean-expressions` to check for array predicate return statements.

Two things to note:

### Doesn't handle implicit returns

I didn't handle implicit return types from array predicates since I'm not sure if the rule should cover this:

```ts
[1, null].filter(x => {
  if (x) {
    return true;
  }

  // implicitly returns `undefined`
});
```

Please let me know if you think it should.

### Somewhat of a breaking change

This is a breaking change, though the issue itself isn't marked as one, and according to the docs:

> A change to the plugins shall not be considered breaking if it:
> - Adds additional checks to an existing rule that causes new reports in a small-to-medium set of cases in an average codebase.

Alternatively, I can put this behind a flag, which can be removed in a later version.


packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
```diff
@@ -114,6 +158,19 @@ if (x) {
...
+      options: [{ allowNullableBoolean: true }],
+      code: `
+        declare const x: (boolean | null)[];
+        x.filter(function (t) {
+          return t;
+        });
+        x.filter(t => {
+          return t ? 1 : 0;
+        });
+        x.filter(t => !t);
```
JoshuaKGoldberg, line 171: [Testing] We generally try to stick to one thing under test for each test case. For this one, it looks like two things are being tested:

* That the `allowNullableBoolean` option still works...
* ...and that general logic holds true for all three variations of something passed to `.filter`

I don't think the latter is necessary. We can trim the tests down a bit:

```suggestion
```

Here and the other ones with several variations of the same, such as `allowNullableString: true`.

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
```diff
@@ -1303,6 +1439,7 @@ if (((Boolean('')) && {}) || (foo && void 0)) { }
         declare const x: string; if (x) {}
         (x: string) => (!x);
         <T extends string>(x: T) => x ? 1 : 0;
+        ['foo', 'bar'].filter(x => x);
```
JoshuaKGoldberg, line 1442: [Testing] Blurgh, `batchedSingleLineTests`. That old deprecated thing. Let's not add more uses of it in this PR.

If there's anything you want to specifically test in here that isn't tested elsewhere, please add it as a standalone test.

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
JoshuaKGoldberg, line 1: [Testing] Can you add tests for `return;`, `return undefined;`, and implicit undefined return? Regardless of what the end behavior ends up being, since they're a case explicitly discussed, I think there should be tests for them.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
JoshuaKGoldberg, line 1: Threading:

> I didn't handle implicit return types from array predicates since I'm not sure if the rule should cover this:

It would be nice if the rule did cover implicit undefined returns. [It does handle _explicit_ ones](https://deploy-preview-10106--typescript-eslint.netlify.app/play/#ts=5.5.2&fileType=.ts&code=Nocg9gdgpiA0AEECuAbFBdAdAMwJYoBcoAnACgA94BeAPngG8AoeeXbeC%2BAQiqsVRQBKBsxbxiUAkmIR4BYkigBuUQF9GoiVJnwkEACZQ80fStWCVQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZfKsugIwHs%2BSAIZN6AD2LQUySnyaoMkRNGh9okcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false), so missing the implicit ones is a bit of a functionality hole / bug.

How difficult would it be to add in? I'm not aware of good ways to detect whether a function has an implicit return undefined in TypeScript 😬

JoshuaKGoldberg (changes requested):
Nice! One discussion on implicit undefined return, and some requests around testing. Direction seems great to me! 🚀

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kirkwaiblinger: Driving by to say there is discussion on this on the issue too, see https://github.com/typescript-eslint/typescript-eslint/issues/8016#issuecomment-2404170958

Makes sense to move the conversation here though

JoshuaKGoldberg: Oh ha sorry I missed those. Looks like you and I are on the same page Kirk - we should check the inferred types if possible.

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
ronami: @JoshuaKGoldberg I can see both `return;` and `return undefined;` are covered by @kirkwaiblinger's [recent addition](https://github.com/typescript-eslint/typescript-eslint/pull/10009) to `@typescript-eslint/no-unnecessary-condition`.

Do you think they should be checked by this rule as well?

Regardless, I'll add test cases for both 👍

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
ronami: I've pushed a few commits to address this. I did something similar to https://github.com/typescript-eslint/typescript-eslint/pull/9955 by checking the function's inferred/explicit return type.

I used a heuristic to report a function as implicitly returning `undefined` if both conditions are met:
- No issues have been found on any of its return statements by this rule.
- It doesn't have empty return statements (e.g. `return;`).

Otherwise, the rule may have two reports for instances such as the following:

```ts
['one', null].filter(x => {
  return undefined;
});
```

For this I had to make some adjustments to the existing utility functions the rule uses. Please let me know if it makes sense.

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
JoshuaKGoldberg: I think they both should be, yes. And great spot! Glad this isn't totally without precedent.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -260,19 +268,124 @@ export default createRule<Options, MessageId>({
...
+          const { hasReported, hasReturnStatement } =
+            traverseArrayPredicateReturnStatements(predicate);
+
+          // Don't report a function as implicitly returning `undefined` only
+          // when no other issues were found with it.
+          if (hasReported) {
+            return;
+          }
+
+          // Empty functions has their return type as `void` rather than `undefined`
```
ronami, line 348: e.g. `[1, 2].filter(x => {});` would have its inferred return value as `void`.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
ronami: One edge case I haven't got to do yet, is the following (which should be valid, currently flagged as returning `undefined`, even though the function returns `never`):

```ts
[1, 2].filter(x => {
  if (typeof x === 'number') {
    return true;
  }
});
```

ronami:
This is ready for a re-review.

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
```diff
@@ -521,6 +642,19 @@ assert(nullableString);
         break;
       }
     `,
+    `
+      const x = (t: string | null) => t;
```
kirkwaiblinger, line 646: Two thoughts...
1. This appears to be the only added test case with a handler that is not a function literal. Maybe add a few more with varied return types?
2. I think this should be invalid, no? The handler returns `string | null`, and if we inline it, it does report.

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
kirkwaiblinger, line 1: [Some test cases I was playing around with....](https://deploy-preview-10106--typescript-eslint.netlify.app/play/#ts=5.5.2&fileType=.ts&code=MYewdgzgLgBAZjAvDAFFAXDaAnAlmAcxgB8YwBXAG0oEokA%2BGKAbgChWB6DrACxCoAmMAEYBTGPgBuAQ0q4BzGNlFRy2SGSqVpwyuJz4CrANoByUwBpN1ALoA6OLkpRR2FHBptO3CH0EjxKVl5RWVVdQhrbV19KDxCKwhpAFtxaUiAB2VJXH4IOxNzKwpbBycXNwAPBhhKz3ZvGABJAFkAeV5%2BSiExCTAZOQUYcEoATxEQED1pME7-XtlKEAB3UQE7GEaAIXJYKB5XUVNImZhpbAJyVLA9nmlb8WwqcRRl3H2YAVE4aSpYEAyUFykDojQEIFEJ2oKywcUMfTOEymolOS2A92BiRAzRgl0hkX2h2OjVO50u11u9wkUGOwwA1mQQHtsXBtEYzJYovZHM5XChqohGJUAIT1RoAZT83QCfQGISUKjUGhK0T0sPiBAKHOKWm55T5-JomAMhBIURqdS8XBgkq6PUC-WCQzCSsiKp0apNRBQokqGTkwHeCvCsygowy4mS9wqkVabRoWqKXLKvLchuNcNNpBVFtFbCAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZfKsugIwHs%2BSAIZN6AD2LQUySnyaoMkAO5DoTSODABfEFqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)
```ts
const f = (t: string | null) => t;

// should be invalid; returns nullable string
['', null].filter(f);

// should be invalid; returns nullable string, same as previous.
['', null].filter(x => x);


// IMO should be invalid; only boolean should be allowed. 
// But there's an argument that the rule (with default options) 
// does allow string in a boolean location, so I guess there's
// an argument that it's ok not to flag
['', null].filter(x => x!);

// Should be invalid; returns nullable string.
['', null].filter((x): string | null => x);

// Should be invalid; returns nullable string (explicit return type matters IMO).
['', null].filter((x): string | null => x!);
```

On a high level, I think we want to flag based on the return type of the handler, regardless of where it comes from, and regardless of whether the implementation of the function actually can return the full width of its declared return type. And, I personally think that this is a special case for the rule where we should demand a true boolean, not just whatever expressions the rule allows in boolean expressions based on its configuration, but we can go either way on this I suppose.

In short, though, I'm thinking
1. Check return type of passed in function. If it's not an allowed type, report it.
2. (this part is optional and could , maybe should(?), be omitted) if a function literal is provided, check individual returns for their return type, and report them individually. 

The thing is, if we implement number 2, and say "well, if we're reporting a return statement, we don't need to bother reporting the handler as a whole", I think this has a problem still, because the user can disable the report at a specific return statement and it will still allow a bad return type.

```ts
// I still think this should report due to its explicitly prohibited return type.
arr.filter((x): string | null => {
    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
    return Math.random() > 0.5 ? null : string
}
```

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -260,19 +268,124 @@ export default createRule<Options, MessageId>({
...
+          if (
+            !hasReturnStatement &&
+            (predicate.type === AST_NODE_TYPES.FunctionExpression ||
+              predicate.type === AST_NODE_TYPES.ArrowFunctionExpression)
+          ) {
+            return context.report({
```
kirkwaiblinger, line 354: In cases where we know it's a function literal, what do you think of adding a suggestion to put an explicit `: boolean` return type? Then the user can let the type errors guide them in resolving which return statements are illegal, rather than us having to do a best effort to underline them?

packages/eslint-plugin/tests/rules/strict-boolean-expressions.test.ts
ronami: I read this several times and went over the PR again, and I couldn't agree more. I think not checking each return statement of function literals is also correct (2). This is a somewhat significant change for the PR; I'll push an update in a bit and ask for a re-review.

Thanks you, @kirkwaiblinger, this has been very helpful 🙏

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
ronami: Nice! I'll update the PR with this shortly.

ronami: @kirkwaiblinger this may collide with #9764 (at least for array's`.filter()`), in which having an explicit boolean return type may prevent TypeScript from inferring the predicate return type.

Wdyt? I'm not sure if this is correct (at least for `.filter()`).

kirkwaiblinger: Hmmmmmm. Well, it's not a hard collision, in that _if_ an explicit type predicate is inferred, then we wouldn't be reporting the function in the first place, since, to be an inferred type predicate, the function must be returning a boolean.

I see what you're saying that it's a bit of an awkward workflow, in that one rule might ask you to put `: boolean`, just for another to ask you to take it away.
```ts
declare const arr: Array<string | null>;

arr.filter(x => x); // flagged by SBE

// suggested to
arr.filter((x): boolean => x); // TS ERROR, x not assignable to boolean, _and_ lint error, remove explicit `: boolean`

// manually fix to
arr.filter((x): boolean => x != null); // lint error, remove explicit `: boolean`

// OR manually fix to
arr.filter((x): boolean => x != null && x !== ''); // lint error, remove explicit `: boolean`
// _but_ note that this would be a false positive since a type predicate is _not_ inferred here.
```

But... since it's a _suggestion_ TBH I think it's ok. Suggestions can always suggest things that another rule may disapprove of, and they should not generally be understood to be the only way to resolve a report. The goal is to guide you to finding a solution, which, IMO, this would do (by surfacing TS type errors), even if it's a little awkward. In any case that rule doesn't exist yet, so 🤷 I don't think we should stress about it too much.

***

That does remind me, though, if there isn't already, could you ensure there's a (valid) test case for explicit and inferred type predicates, to be sure we don't flag them as non-boolean (it did work as expected when I tried the playground 👍)

kirkwaiblinger: definitely a good observation, though! Really appreciate the deep attention to detail.

ronami:
Please note that this feature is not behind a flag, which means it will be a breaking change. I don't know how many lint failures this adds to an average project, but I think it may be safer to release this under an option (defaults to `false`) and change it in a future version.

Wdyt? I'll be happy to adjust the PR accordingly.

JoshuaKGoldberg:
> breaking change

From https://typescript-eslint.io/users/versioning/#eslint-plugin:

> A change to the plugins **_shall not_** be considered breaking if it:
> ...
> * Adds additional checks to an existing rule that causes new reports in a small-to-medium set of cases in an average codebase

I think this qualifies. Honestly, I personally consider this closer to a bugfix than a feature. Having it necessitate minor version rather than patch is a nicety for users.

[Obligatory xkcd spacebar comic](https://xkcd.com/1172)

ronami:
> > breaking change
> 
> From https://typescript-eslint.io/users/versioning/#eslint-plugin:
> 
> > A change to the plugins **_shall not_** be considered breaking if it:
> > ...
> > 
> > * Adds additional checks to an existing rule that causes new reports in a small-to-medium set of cases in an average codebase
> 
> I think this qualifies. Honestly, I personally consider this closer to a bugfix than a feature. Having it necessitate minor version rather than patch is a nicety for users.
> 
> [Obligatory xkcd spacebar comic](https://xkcd.com/1172)

Yes, I think so too (mentioned exactly this on the PR description when I opened it 👍 ).

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -275,6 +315,87 @@ export default createRule<Options, MessageId>({
...
+    ): void {
+      // custom message for accidental `async` function expressions
+      if (isFunctionExpressionNode(node) && node.async) {
+        return context.report({
+          node,
+          messageId: 'predicateCannotBeAsync',
+        });
+      }
+
+      const type = checker.getApparentType(services.getTypeAtLocation(node));
```
kirkwaiblinger, line 344: wondering if there's a way to leverage `checker.getResolvedSignature` or similar to deal with some edge cases around generic/overload resolution 🤔. If not, I think it's probably ok, since the arrow function form does work correctly, and is much more common in real world use cases. 

```ts
declare function f(x: number): string;
declare function f(x: string | null): boolean;


// shouldn't report; doesn't.
['one', null].filter(f);
['one', null].filter(x => f(x));


// (technically) should report; doesn't.
[35].filter(f);
// correct with arrow function form
[35].filter(x => f(x));


declare function generic<T>(x: T): T;
// (technically) shouldn't report; does.
[true, false].filter(generic);
// correct with arrow function form
[true, false].filter(x => generic(x));

// important test case with generic type predicate (important for code that predates inferred type predicates)
declare function notNullish<T>(x: T): x is NonNullable<T>;
// should not report; doesn't
['one', null].filter(notNullish);
```

***

if we can't figure out which overload will be used, I think the only change I would make is that we should probably error if _any_ of the overloads doesn't return a boolean, rather than only if _all_ don't return a boolean. Maybe that's a bit overbearing, but I think it's preferable to have always false positives here, rather than sometimes false positives _and_ sometimes false negatives

kirkwaiblinger (changes requested):

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
ronami: That's an interesting problem. I looked around, but I don't think `checker.getResolvedSignature` can help here, potentially the currently not exposed [`isSignatureAssignableTo`](https://github.com/microsoft/TypeScript/blob/b3c67d32020a3015fe068fbcb1dcfb18aedcaad3/src/compiler/checker.ts/#L21273-L21275) or [`compareSignaturesRelated`](https://github.com/microsoft/TypeScript/blob/b3c67d32020a3015fe068fbcb1dcfb18aedcaad3/src/compiler/checker.ts/#L21294-L21430).

The best I can think of is to mimic that somewhat: Since we know for a fact [every array predicate method](https://github.com/typescript-eslint/typescript-eslint/blob/244435126619afb9497ace04cbf4819012e27330/packages/eslint-plugin/src/util/isArrayMethodCallWithPredicate.ts#L13-L21) takes the type of the array as its first argument (and a `number` as the second one), we can try and rule out which signature we should check.

I did manage to implement the above, so these both correctly report:

```ts
// (technically) should report; doesn't.
[35].filter(f);
// correct with arrow function form
[35].filter(x => f(x));
```

I'm not sure, however how to tackle the generic one:

```ts
declare function generic<T>(x: T): T;
// (technically) shouldn't report; does.
[true, false].filter(generic);
// correct with arrow function form
[true, false].filter(x => generic(x));
```

Wdyt? I'll add the missing test cases regardless of how this is tackled.

kirkwaiblinger: > That's an interesting problem. I looked around, but I don't think checker.getResolvedSignature can help here, potentially the currently not exposed [isSignatureAssignableTo](https://github.com/microsoft/TypeScript/blob/b3c67d32020a3015fe068fbcb1dcfb18aedcaad3/src/compiler/checker.ts/#L21273-L21275) or [compareSignaturesRelated](https://github.com/microsoft/TypeScript/blob/b3c67d32020a3015fe068fbcb1dcfb18aedcaad3/src/compiler/checker.ts/#L21294-L21430).

Oh well. Thanks for looking! 

I think that I would prioritize the test cases as follows
```ts
declare function notNullish<T>(x: T): x is NonNullable<T>;

declare function f(x: number): string;
declare function f(x: string | null): boolean;

declare function generic<T>(x: T): T;

// Priority 1. The only one that is truly important.
// should not report
['one', null].filter(notNullish);

// Priority 2
// should report
[35].filter(f);

// Priority 10
// shouldn't report (but erring on the side of reporting is fine)
['one', null].filter(f);

// Priority 11
// shouldn't report (but erring on the side of reporting is fine)
[true, false].filter(generic);
```

If the heuristic implementation is simple, we can do it, but otherwise I'm not sure its worth extra code complexity. 

Does that help?

ronami: Yes, thanks!

I've implemented this as simple as I could, and this does solve every case you mentioned but the last (which also shows as an incorrect report on `no-unnecessary-condition`), wdyt?

If you think this adds redundant complexity I'll be happy to revert this and ask all signatures to return a `boolean`.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -232,6 +242,111 @@ export default createRule<Options, MessageId>({
...
+      const parameterType = checker.getTypeOfSymbol(parameter);
+
+      const constrained = isTypeParameter(parameterType)
+        ? checker.getBaseConstraintOfType(parameterType)
+        : parameterType;
+
+      // treat a type parameter without constraint as `unknown`
+      if (!constrained) {
+        return true;
+      }
```
ronami, line 282: [This comment](https://github.com/typescript-eslint/typescript-eslint/pull/10314#discussion_r1836097699) is probably relevant here as well.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
ronami: Notice I've pushed one additional change just now that simplifies things a bit further.

ronami: @kirkwaiblinger note that a similar behavior can be seen on `no-unnecessary-condition` ([playground link](https://typescript-eslint.io/play/#ts=4.8.4&fileType=.tsx&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMBGACgA8AueVZAWwCMQZ4AfO5CCASlowxkIANwAoUJFgIU6bHgLEyVWiRhZUAc37w1GzeLEB6I7oAWObsCRQsEADTwoRAmYRgcMOJl1ZNqKAxkOHgsFww3eDwEOCCYVH14AANBYSSxAG0AZgBWR3oeAF0AOkQ7QhhyQhJSXkMTGzsXDy9wDAgATycXKCQ0TFx8EEoABzgiIkHM3PzuCBKyiAqqeABeAD5FGqpeOrEgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK%2BSpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```ts
declare function test1(x: number | null): true;
declare function test1(x: string): string;

// should fail, as the correct signature is the one returning `true`
[35, null].filter(test1);

// fails correctly as a function expression
[35, null].filter(x => test1(x));
```

And also on `no-misused-promises` ([playground link](https://typescript-eslint.io/play/#ts=4.8.4&fileType=.tsx&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMBGACgA8AueVZAWwCMQZ4AfO5CCASlqY4cEEFFQBuAFChIsBCnTY8BYmSq0SMLKgDm-eAAUYOBliIgAPPWasAfFMkB6R-CIALHN2DwADlCJEADTw-gRuCGA4MHCYrlg6qFAYyHDwZmEIeAhwyTCo2jrwAAaCwqKoRZIA2gDMAKzB9DwAugB0iFgQhDDkhCSkvA7Ovv7mRPCR0eAYEACeIeNQSGiYuPgglD5wAWvV9Y3cEG0dXaxU8AC8tir9VLyDkkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oFtLlZkiACa1i0Dr0GoMkRNHHRI4MAF8QKoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```ts
declare function test1(x: number | null): boolean;
declare function test1(x: string): Promise<number>;

// should pass, as the correct signature is the one returning `boolean`
[35, null].filter(test1);

// passes correctly as a function expression
[35, null].filter(x => test1(x));
```

Since all of the examples above check call signatures of array predicates, they can reuse the heuristic I wrote to fix these bugs. If this makes sense, I'll be happy to open issues (followed by PRs) for this.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -275,6 +367,105 @@ export default createRule<Options, MessageId>({
...
+          node: predicateNode,
+          messageId: 'predicateCannotBeAsync',
+        });
+      }
+
+      const type = checker.getApparentType(
+        services.getTypeAtLocation(predicateNode),
+      );
+
+      const signatures = isFunctionExpression
```
kirkwaiblinger, line 404: I find it odd that changing this to unconditionally use the `getMatchingArrayPredicateSignatures()` heuristic can make _invalid_ tests fail. I'd think it should only really be able to make _valid_ tests fail. Looking at the failures I get, they all involve omitted parameters. So - is there a bug in the heuristic around omitted parameters?

Surely this should be flagged?

([playground](https://deploy-preview-10106--typescript-eslint.netlify.app/play/#ts=5.5.2&fileType=.tsx&code=GYVwdgxgLglg9mABMOcAUAPAXIsIC2ARgKYBOAlIgN4C%2BAUANoCMANLiADYcC6AdMQDcyATzQo45ANxA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZfKsugIwHs%2BSAIZN6AD2LQUySnyaoMkAO5DoTSODABfEFqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false))
```ts
function foo(x: number) {}
[1, null].every(foo);
```

kirkwaiblinger: Oh, hmmmmmmmmm. Upon closer inspection, I think we're running into a `strictFunctionTypes` problem impacting the assignability API. Idk, this kind of makes me want to avoid the heuristics altogether and stick with simply: if it ain't clearly a boolean according to the `getCallSignatures()`, just report it. I just don't think we want it to be our problem to reimplement what TS should have been doing from the start. WDYT? Also maybe would be good to tag in other team members for opinions on this.

kirkwaiblinger:
btw - sorry for taking so long to get around to a review on this!

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
ronami: That's interesting!

I didn't consider reporting this case, as it's normally a type error:

```ts
function foo(x: number) {}
[1, null].every(foo);
```

However, as you pointed out, this code is valid (typescript-wise) with `strict` mode being turned off. I'm so used to having `strict` mode enabled that I didn't even consider this an option.

I can think of ways to fix this "bug" (checking if the parameter is assignable to the value or the other way around. This will check if they have some overlap), but I'm not sure it's worth the maintenance burden and the potential bugs.

As you suggested, I've updated the PR to be a lot simpler and report on any function that may return a non-boolean.

Thanks for your help with this ❤️

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -232,6 +242,88 @@ export default createRule<Options, MessageId>({
...
+          continue;
+        }
+
+        signatures.push(signature);
+      }
+
+      return signatures;
+    }
+
+    function isFunctionExpressionNode(
```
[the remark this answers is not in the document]
ronami: Nice! I've already got it implemented for me in https://github.com/typescript-eslint/typescript-eslint/blob/d24a82854d06089cbd2a8801f2982fd4781f3701/packages/utils/src/ast-utils/predicates.ts#L69 😄

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -232,6 +242,111 @@ export default createRule<Options, MessageId>({
       | TSESTree.IfStatement
       | TSESTree.WhileStatement;
 
+    function isBooleanType(expressionType: ts.Type): boolean {
```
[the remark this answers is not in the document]
ronami: Nice! 👍

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -275,6 +367,105 @@ export default createRule<Options, MessageId>({
...
+      ) {
+        return;
+      }
+
+      const canFix = isFunctionExpression && !predicateNode.returnType;
+
+      return context.report({
+        node: predicateNode,
+        messageId: 'predicateReturnsNonBoolean',
+        suggest: canFix
```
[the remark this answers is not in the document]
ronami: Nice! I've tried to use this here, but I think it's a bit more tricky: The fixer relies on the `canFix` check, which narrows down the type of `predicateNode` to a function expression. Using `getFixOrSuggest` removes that and will require a type assertion, which I'm not sure is isn't worth it.

Wdyt?

ronami:
> btw - sorry for taking so long to get around to a review on this!

No problem, I appreciate your assistance with the PR, it's been very helpful 👍

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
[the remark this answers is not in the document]
kirkwaiblinger: Oh, that's fun! I'm honestly surprised the narrowing even applies within the closure (it can get reset on mutable things). Anyways, that's reasonable, let's just leave it as is 👍

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -275,6 +285,103 @@ export default createRule<Options, MessageId>({
...
+                    );
+                  }
+
+                  const closingBracket = nullThrows(
+                    context.sourceCode.getTokenAfter(lastParam),
+                    'missing token following function parameters.',
+                  );
+
+                  // one or more parameters wrapped with parens
+                  if (closingBracket.value === ')') {
```
kirkwaiblinger, line 366: There is this utility also if you wanted, but it's fine as-is also

https://github.com/typescript-eslint/typescript-eslint/blob/3f02687d964c1e7888e5db62256d56cd72c4f281/packages/eslint-plugin/src/util/misc.ts#L227-L234

kirkwaiblinger (changes requested):
home stretch!

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -275,6 +285,103 @@ export default createRule<Options, MessageId>({
...
+      if (isFunctionExpression && predicateNode.async) {
+        return context.report({
+          node: predicateNode,
+          messageId: 'predicateCannotBeAsync',
+        });
+      }
+
+      const type = checker.getApparentType(
+        services.getTypeAtLocation(predicateNode),
+      );
```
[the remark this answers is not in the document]
ronami: Nice catch! This is completely unnecessary, I'm not sure what I was thinking 😆

I've updated the code to use `services.getTypeAtLocation()`, but I also reviewed the PR as it's somewhat old, and I've learned a lot since then. I found a bug of not checking type constraints, which I've addressed (and added relevant tests), so this code looks slightly different than the suggestion above.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -275,6 +285,103 @@ export default createRule<Options, MessageId>({
...
+
+      if (
+        signatures.every(signature =>
+          tsutils
+            .unionTypeParts(signature.getReturnType())
+            .every(isBooleanType),
+        )
+      ) {
+        return;
+      }
```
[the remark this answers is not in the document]
ronami: Thanks, I think this is unnecessary 👍

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
ronami: Nice! I've used this (though I had to make slight changes to it) along with refactoring my old code for the fixer a bit too.

packages/eslint-plugin/src/util/misc.ts
```diff
@@ -225,8 +225,8 @@ function isRestParameterDeclaration(decl: ts.Declaration): boolean {
   return ts.isParameter(decl) && decl.dotDotDotToken != null;
 }
 
-function isParenlessArrowFunction(
-  node: TSESTree.ArrowFunctionExpression,
+function isParenlessFunctionExpression(
```
kirkwaiblinger, line 228: For me, `isParenlessFunctionExpression` evokes this:
```ts
(function iifeWithParens() {})();
// vs
+function iifeNoParens() {}();

// or

arr.map(function (x) { return x; });
// vs
arr.map((function (x) { return x; }));
```

rather than

```ts
(x) => y;
// vs
x => y;
```

Let's do one or multiple of
- Keep the name specific to arrow functions
- Keep the `node` type unchanged and just check the type of the node in SBE before calling it (I would definitely recommend this, since otherwise we should probably short-circuit false if the node is not an arrow function expression in the implementation, at which point... why even let it be called on non- arrow functions)
- Add JSDoc to be more clear what this function is even checking for.

kirkwaiblinger (changes requested):

packages/eslint-plugin/src/util/misc.ts
ronami: I see your point, it's definitely confusing.

It's also completely redundant (like you mentioned). I've reverted the changes to `isParenlessArrowFunction` and added a check verifying the node is an arrow function expression.

kirkwaiblinger (approved):
I think we're there! Great work on this! I'm excited to get to use this feature ❤️

JoshuaKGoldberg (approved):
So thorough, love it! 🚀

PR #10114, opened 2024-10-09
feat(eslint-plugin): [no-base-to-string] prevent stringifying arrays with elements that cannot be stringified

### LINKED ISSUE #3388: [no-base-to-string] add option to prevent usage of Array .join on non-toString things
(Note: I couldn’t find any issue template/info about feature requests, I hope this is ok! 🙏 )

Summary: Disallow `.join()` calls on `Array`s that aren’t `Array<string>`.

Motivation:

I had a function like this:

```ts
function noCommonRoot(paths: Array<string>): string {
  return `
I could not find a common ancestor for these paths:

${paths.join("\n")}

Files on different drives is not supported.
  `.trim();
}
```

I had trouble with using too much `string` types an mixing stuff up, so I introduced an `AbsolutePath` type to help me, and replaced `string` with `AbsolutePath` in many places.

```ts
type AbsolutePath = { tag: "AbsolutePath"; absolutePath: string };

function noCommonRoot(paths: Array<AbsolutePath>): string {
  return `
I could not find a common ancestor for these paths:

${paths.join("\n")}

Files on different drives is not supported.
  `.trim();
}
```

When I had fixed all compile errors, I had accidentally introduced a bug! `noCommonRoot` now outputs:

```
I could not find a common ancestor for these paths:

[object Object]
[object Object]
[object Object]

Files on different drives is not supported.
```

Oops!

I searched for `.join` in the entire project and switched to using this function instead, to avoid this problem in the future:

```ts
function join(array: Array<string>, separator: string): string {
  return array.join(separator);
}
```

That way I could fix the `noCommonRoot` function:

```ts
function noCommonRoot(paths: Array<AbsolutePath>): string {
  return `
I could not find a common ancestor for these paths:

${join(
  paths.map((path) => path.absolutePath),
  "\n"
)}

Files on different drives is not supported.
  `.trim();
}
```

Bonus: my `join` function _requires_ the separator argument. That would be nice to lint too. The default `","` is never what I want.

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3388

## Overview

This PR addresses #3388 and adds checks calling `.join()` on an array that not all of its elements can be stringified correctly:

```ts
function test(type: {}[]) {
  return type.join();
}
```

I've also included failing in the case of[ trying to stringify an array](https://github.com/typescript-eslint/typescript-eslint/issues/3388#issuecomment-2403349701) (without `.join()`):

```ts
function test(type: {}[]) {
  return `${type}`;
}

function test(type: {}[]) {
  return type.toString();
}
```

Please note that I haven't put this behind an option, so this is technically a breaking change, [although according to the docs](https://typescript-eslint.io/users/versioning#eslint-plugin):

> A change to the plugins shall be considered breaking if it will require the user to change their config. More specifically:
>  - Changes the default behavior of a rule in such a way that causes new reports in a large set of cases in an average codebase.

Please let me know if I should put this behind an option.


ronami:
Closing as it takes me a bit longer than I thought to finish this. I'll re-open when the PR is ready for review.

PR #9848, opened 2024-08-20
chore: enable eslint-plugin-perfectionist on scope-manager package

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: continues #8479
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Enables sorting on `packages/scope-manager`.

💖

PR #10128, opened 2024-10-10
chore: update to nx 20

JamesHenry:


nx.json
```diff
@@ -3,7 +3,6 @@
   "nxCloudAccessToken": "YjIzMmMxMWItMjhiMS00NWY2LTk1NWYtYWU3YWQ0YjE4YjBlfHJlYWQ=",
   "release": {
     "projects": [
-      "*",
```
JamesHenry, line 6: This is no longer needed, we can use a list of only negative matchers and the ALL EXCEPT is implied

tools/release/changelog-renderer.js
```diff
@@ -2,34 +2,15 @@
...
-    entryWhenNoChanges,
-    changelogRenderOptions,
-    repoSlug,
-    conventionalCommitsConfig,
-    changes,
-  });
-
-  // Append our custom messaging to the generated changelog entry
-  return `${defaultChangelog}\n\nYou can read about our [versioning strategy](https://main--typescript-eslint.netlify.app/users/versioning) and [releases](https://main--typescript-eslint.netlify.app/users/releases) on our website.`;
+module.exports = class CustomChangelogRenderer extends (
```
JamesHenry, line 8: The changelog renderer API is now class based where individual elements of the render can be easily patched by overwriting base methods. In our case we are just bolting text onto the full default render, so we only need to amend the main render() method.

tools/release/release.mts
```diff
@@ -59,9 +59,13 @@ if (workspaceVersion === null) {
   process.exit(0);
 }
 
-const publishStatus = await releasePublish({
+const publishProjectsResult = await releasePublish({
```
JamesHenry, line 62: `releasePublish` now always returns a mapping of project names to data (which right now only contains exit codes), rather than one simplistic exit code for the whole set of projects

jest.config.js
```diff
@@ -1,9 +1,9 @@
 'use strict';
 
 // @ts-check
-const { getJestProjects } = require('@nx/jest');
+const { getJestProjectsAsync } = require('@nx/jest');
```
JoshuaKGoldberg, line 4: I'm getting flashbacks to C#/.NET 😛

JamesHenry: The sync version is only deprecated and still there so sadly we have to have another name for this one!

knip.ts
```diff
@@ -1,5 +1,6 @@
 import type { KnipConfig } from 'knip' with { 'resolution-mode': 'import' };
 
+// eslint-disable-next-line import/no-default-export
```
JoshuaKGoldberg, line 3: Oh yeah, I fixed this in https://github.com/typescript-eslint/typescript-eslint/pull/9848 too - by just excluding `knip.ts` from `import/no-default-export`... let me merge that now.

packages/typescript-eslint/src/index.ts
```diff
@@ -178,6 +178,7 @@ module.exports = config(
 );
 ```
 */
+// eslint-disable-next-line import/no-default-export
```
JoshuaKGoldberg, line 181: Same with https://github.com/typescript-eslint/typescript-eslint/pull/9848 here

JoshuaKGoldberg (approved):
🍰

PR #10134, opened 2024-10-11
chore(rule-tester): rethrow exceptions in RuleTester to avoid circular JSON issue

### LINKED ISSUE #8942: Repo: handle exceptions that occur while testing a rule better
### Suggestion

If you throw an exception in a rule (within the RuleListener object handlers, not at the top level of the `create()` function), the whole test suite fails with an unhelpful error message, rather than failing just the individual test run and printing a useful error message. For instance, here's what happens when I throw an exception in a line affected by just one test case: 

```
 FAIL  tests/rules/no-unnecessary-type-arguments.test.ts
  ● Test suite failed to run

    TypeError: Converting circular structure to JSON
        --> starting at object with constructor 'Object'
        |     property 'params' -> object with constructor 'Array'
        |     index 0 -> object with constructor 'Object'
        --- property 'parent' closes the circle
        at stringify (<anonymous>)
```

@JoshuaKGoldberg has a hacky workaround. Feel free to share in case that makes it easier to pinpoint the root cause?

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8942 

## Overview

I don't really get the circular JSON problem but looks like we can suppress it by just rethrowing the exception 🤷‍♂️

### DIFF
--- a/packages/rule-tester/src/RuleTester.ts
+++ b/packages/rule-tester/src/RuleTester.ts
@@ -184,10 +184,37 @@ export class RuleTester extends TestFramework {
       rules: { [`${RULE_TESTER_PLUGIN_PREFIX}validate-ast`]: 'error' },
     });
 
-    this.#linter = new Linter({
-      configType: 'flat',
-      cwd: this.#testerConfig.languageOptions.parserOptions?.tsconfigRootDir,
-    });
+    // This proxy nonsense is a workaround for https://github.com/jestjs/jest/issues/14840
+    // see also https://github.com/typescript-eslint/typescript-eslint/issues/8942
+    //
+    // For some reason rethrowing exceptions skirts around the circular JSON error.
+    this.#linter = new Proxy(
+      new Linter({
+        configType: 'flat',
+        cwd: this.#testerConfig.languageOptions.parserOptions?.tsconfigRootDir,
+      }),
+      {
+        // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+        get: (target, prop, receiver) => {
+          if (prop === 'verify') {
+            return (
+              ...args: Parameters<Linter['verify']>
+            ): ReturnType<Linter['verify']> => {
+              try {
+                return target.verify(...args);
+              } catch (error) {
+                throw new Error('Caught an error while linting', {
+                  cause: error,
+                });
+              }
+            };
+          }
+
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+          return Reflect.get(target, prop, receiver);
+        },
+      },
+    );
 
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (approved):
_Sigh_. I really hate going with bandaids like this... but given how slow Jest maintenance is, I don't see a way around it. Blagh.

kirkwaiblinger:
> _Sigh_. I really hate going with bandaids like this... 

Yeah, it really is yucky IMO 🙁 but what can you do 🤷

packages/rule-tester/src/RuleTester.ts
```diff
@@ -184,10 +184,37 @@ export class RuleTester extends TestFramework {
...
 
-    this.#linter = new Linter({
-      configType: 'flat',
-      cwd: this.#testerConfig.languageOptions.parserOptions?.tsconfigRootDir,
-    });
+    // This proxy nonsense is a workaround for https://github.com/jestjs/jest/issues/14840
+    // see also https://github.com/typescript-eslint/typescript-eslint/issues/8942
+    //
+    // For some reason rethrowing exceptions skirts around the circular JSON error.
+    this.#linter = new Proxy(
```
bradzacher, line 191: Can't we just wrap the `verify` call directly?

https://github.com/typescript-eslint/typescript-eslint/blob/0884b6a43c26caafb6ea7247290ca966da99cd36/packages/rule-tester/src/RuleTester.ts#L1228-L1235

Or failing that - directly override `verify`, eg
```ts
const linter = new Linter(...);
const oldVerify = linter.verify;
linter.verify = (...args) => { ... oldVerify(...args) ... };
```

kirkwaiblinger: > Can't we just wrap the verify call directly?

There are three `this.#linter.verify` calls so I at least wanted to put them all in one codepath.

> Or failing that - directly override `verify`, eg
> 
> ```ts
> const linter = new Linter(...);
> const oldVerify = linter.verify;
> linter.verify = (...args) => { ... oldVerify(...args) ... };
> ```

Yeah, that works too. I just wasn't sure if `Linter` was some arcane class object that depended on crazy `this`-context and/or getter binding stuff to work correctly (like, for example, the `context` variable is within the rules) or if we were safe to do that. Can switch it to that if it seems safe.

JoshuaKGoldberg: oh yeah now that Brad mentions it, +1 to that more direct strategy

bradzacher (approved):

JoshuaKGoldberg (approved):

bradzacher (approved):

PR #10135, opened 2024-10-11
fix(eslint-plugin): [class-literal-property-style] don't report nodes with `override` keyword

### LINKED ISSUE #3602: [class-literal-property-style] Literal property style does not respect inherited types

**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/class-literal-property-style": "warn"
  }
}
```

```TS
abstract class BaseClass
{
    // class-literal-property-style error; this one is fair, because it isn't entirely aware
    // that I'm going to override it, so this one I can see just being an eslint disable
    get cursor(): string|string[] { return 'crosshair'; }

    readonly hidden: boolean = false;

    abstract get icon(): string;
}

class ChildClass extends BaseClass
{
    // class-literal-property-style error
    // this one, however, is interesting because it should know that it is
    // overriding an existing getter and can't be made a literal 
    override get cursor() { return 'square'; }

    readonly hidden: boolean = true

    readonly icon = 'circle'

    something: boolean
}
```

**Expected Result**

ESLint shouldn't suggest `class-literal-property-style` when overriding a getter, because it is not possible to turn an inherited getter into a readonly property (ts(2610))

**Actual Result**

It suggests breaking the code by attempting to convert the children into a literal when it is not possible. Conversely, it suggests converting the base class one into a literal, which will not work for children that need to use a getter for conditional return values (ts(2611)).

**Additional Info**

As a side-issue, this rule is not enum-aware and does not trigger if the literal returned from a getter is an enum value:

```ts
enum Test
{
    Value
}

class TestClass
{
    // should trigger here
    get test() { return Test.Value; }
}
```

**Versions**

| `@typescript-eslint/eslint-plugin` | `8.2.1` |
| `@typescript-eslint/parser`        | `8.2.1` |
| `TypeScript`                       | `4.3.4` |
| `ESLint`                           | `7.28.0` |
| `node`                             | `14.8.0` |

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3602 

## Overview

Note that this affects both the 'getters' and 'fields' style.

### DIFF
--- a/packages/eslint-plugin/src/rules/class-literal-property-style.ts
+++ b/packages/eslint-plugin/src/rules/class-literal-property-style.ts
@@ -145,6 +145,7 @@ export default createRule<Options, MessageIds>({
         MethodDefinition(node): void {
           if (
             node.kind !== 'get' ||
+            node.override ||
             !node.value.body ||
             node.value.body.body.length === 0
           ) {
@@ -222,7 +223,7 @@ export default createRule<Options, MessageIds>({
           }
         },
         PropertyDefinition(node): void {
-          if (!node.readonly || node.declare) {
+          if (!node.readonly || node.declare || node.override) {
             return;
           }
           const { properties } =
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (approved):
💯

PR #10137, opened 2024-10-12
feat(eslint-plugin): [prefer-optional-chain] support if statement as part of chain

### LINKED ISSUE #6309: Enhancement: [prefer-optional-chain] Also rewrite `if (x) x.something()`

https://typescript-eslint.io/rules/prefer-optional-chain/

### Description

The rule apparently only tries to prevent `a && a.b`, but another way to write that is `if (a) a.b`

Could the rule be updated to include `if` conditions?

### Fail

```typescript
if (callback) {
  callback();
}

if (list) {
  list.push(item);
}
```

### Pass

```typescript
callback?.();

list?.push(item);

// Optional chaining doesn't apply here
if (callback) {
  callbacks.push(callback);
}
```

### Additional Info

_No response_

omril1:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6309

## Overview

Add an `allowSuggestingOnIfStatements` option that would look at `if` statements with a single expression in their body and try to report a suggestion for a bigger chain that includes the `if`'s `test` node and `body` node


packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1917,6 +2059,35 @@ describe('hand-crafted cases', () => {
         !x || x.a;
       `,
       "typeof globalThis !== 'undefined' && globalThis.Array();",
+      // https://github.com/typescript-eslint/typescript-eslint/issues/8380
```
omril1, line 2062: I moved these tests to here since I think they are more closely related

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -157,6 +262,7 @@ export default createRule<
         }
       },
 
+      // specific handling for `(foo ?? {}).bar` / `(foo || {}).bar`
```
omril1, line 265: I think this comment got misplaced in #9843

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -665,35 +665,189 @@ describe('|| {}', () => {
...
+  ruleTester.run('prefer-optional-chain', rule, {
+    invalid: [
+      {
+        code: noFormat`if (foo) { foo.bar(); }`,
+        errors: [{ messageId: 'preferOptionalChain', suggestions: null }],
+        options: [{ allowSuggestingOnIfStatements: true }],
+        output: 'foo?.bar();',
+      },
+      {
+        code: noFormat`if (foo) { foo.bar() }`, // Missing semi-colon
```
JoshuaKGoldberg, line 682: [Style] Personally, I don't find value in a lot of the comments & separation here.
* Rule unit test cases tend to grow organically over time. The comments don't generally age well.
* You can infer from the code anyway what these comments are trying to convey

```suggestion
        code: noFormat`if (foo) { foo.bar() }`,
```

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -665,35 +665,189 @@ describe('|| {}', () => {
...
+        code: noFormat`if (foo) { foo.bar; }`,
+        options: [{ allowSuggestingOnIfStatements: true }],
+      },
+      {
+        code: noFormat`if (foo) { foo.bar?.baz; }`,
+        options: [{ allowSuggestingOnIfStatements: true }],
+      },
+      // Ignore when comment exists before or after statement inside consequent
+      {
+        code: noFormat`if (foo) { /* comment */ foo.bar(); }`,
```
JoshuaKGoldberg, line 769: [Bug] Why ignore this? I would think we'd still want to report, and have the fixer/suggestion move the comment. This `/* comment */` doesn't change the fact that the code violates the rule's new preference.

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
JoshuaKGoldberg, line 1: [Bug] A couple of bugs that should have tests added:

* [Comments around the `if` are removed](https://deploy-preview-10137--typescript-eslint.netlify.app/play/#ts=5.6.2&fileType=.ts&code=DYUwLgBAbghsCuIBcEDOYBOBLAdgcwG0BdCAHwnhwBMQAzXEKgbgCgWtaIB6AKghgg8uEABTc%2BAI0HDYCEOIgBjaQEoFVaRADe0OIgB0GEFBAZUIESqYQAvkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5ujABtcNhy9o-aJAA0ylVhXZI4%2BPwDuAZVgBzKynzMrAeSYBJLufyj8iALaIWqBj4cIi62AC%2BYQC6ypHhQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)
* [The tested node should be type-checked: `if` checks for truthiness while `?.` is the more strict `== null`](https://deploy-preview-10137--typescript-eslint.netlify.app/play/#ts=5.6.2&fileType=.ts&code=DYUwLgBAbghsCuIBcEDOYBOBLAdgcwgB8J4cATEAM1xDIG4AoBrSiAClgRAEoIBvBhGhxEAOjAB7AMqZceNt0YBfBkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5ujABtcNhy9o-aJAA0ylVhXZI4%2BPwDuAZVgBzKynzMrAeSYBJLufyj8iALaIWqBj4cIi62AC%2BYQC6ypHhQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)
  * I might be wrong here, the `checkString` option & co. can be tricky...
  * I know a lot of existing tests violate this, but that's old & not how we'd prefer (#8298). Anything for this PR's new code should be properly typed.

packages/eslint-plugin/docs/rules/prefer-optional-chain.mdx
```diff
@@ -89,6 +89,32 @@ This style of code isn't super common - which means having this option set to `t
...
 
+### `allowSuggestingOnIfStatements`
+
+{/* insert option description */}
+
+Examples of code for this rule with `{ allowSuggestingOnIfStatements: true }`:
+
+<Tabs>
+
+<TabItem value="❌ Incorrect for `allowSuggestingOnIfStatements: true`">
```
JoshuaKGoldberg, line 100: [Docs] We've standardized the tabs a bit:

```suggestion
<TabItem value="❌ Incorrect">
```

packages/eslint-plugin/docs/rules/prefer-optional-chain.mdx
```diff
@@ -89,6 +89,32 @@ This style of code isn't super common - which means having this option set to `t
...
+
+```ts option='{ "allowSuggestingOnIfStatements": true }'
+if (foo) {
+  foo.bar();
+}
+```
+
+</TabItem>
+<TabItem value="✅ Correct for `allowSuggestingOnIfStatements: false`">
```
JoshuaKGoldberg, line 109:
```suggestion
<TabItem value="✅ Correct">
```

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -94,6 +100,7 @@ export default createRule<
   defaultOptions: [
     {
       allowPotentiallyUnsafeFixesThatModifyTheReturnTypeIKnowWhatImDoing: false,
+      allowSuggestingOnIfStatements: false,
```
JoshuaKGoldberg, line 103: For this major version (v8) I think it makes sense to leave this off by default. But in general, this is a good change that makes the rule more comprehensive. I suspect we'll want to enable it by default in the next major version that changes defaults (presumably, v10).

_"allow suggesting X"_ is also a bit of a double negative. Most `allow*` options make rules more _lenient_.

Suggestion: let's have this be `true` by default, with the intent to make it more strict later?

```suggestion
      allowIfStatements: true,
```

(this'll also need a docs update)

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -109,7 +116,105 @@ export default createRule<
...
+            return;
+          }
+
+          const hasCommentsInIfBody =
+            context.sourceCode.getCommentsBefore(ifBodyStatement).length ||
+            context.sourceCode.getCommentsAfter(ifBodyStatement).length;
+
+          if (hasCommentsInIfBody) {
+            return;
+          }
```
JoshuaKGoldberg, line 151: [Bug] I think we should still report in this case. Folks put comments all over the place sometimes; on their own they don't indicate a lint rule should be ignored.

```suggestion
```

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -109,7 +116,105 @@ export default createRule<
     const seenLogicals = new Set<TSESTree.LogicalExpression>();
 
     return {
-      // specific handling for `(foo ?? {}).bar` / `(foo || {}).bar`
+      // specific handling for `if (foo) { foo.bar(); }` / `if (foo) foo.bar();`
+      'IfStatement[consequent.type=BlockStatement][consequent.body.length=1], IfStatement[consequent.type=ExpressionStatement]':
+        (
```
JoshuaKGoldberg, line 121: [Style] Nit: no need to make these `: () =>` arrow function properties. They can be direct methods. That'd give it one less level of indentation.

```suggestion
      'IfStatement[consequent.type=BlockStatement][consequent.body.length=1], IfStatement[consequent.type=ExpressionStatement]'(
```

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -109,7 +116,105 @@ export default createRule<
...
+          }
+          analyzeChain(
+            context,
+            parserServices,
+            options,
+            node,
+            '&&',
+            currentChain,
+          );
+          return;
```
JoshuaKGoldberg, line 215: [Style] Nit: no need; it's at the end of a function anyway.

```suggestion
```

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -109,7 +116,105 @@ export default createRule<
...
+      // specific handling for `if (foo) { foo.bar(); }` / `if (foo) foo.bar();`
+      'IfStatement[consequent.type=BlockStatement][consequent.body.length=1], IfStatement[consequent.type=ExpressionStatement]':
+        (
+          node: {
+            consequent: {
+              type:
+                | AST_NODE_TYPES.BlockStatement
+                | AST_NODE_TYPES.ExpressionStatement;
+            };
+          } & TSESTree.IfStatement,
```
JoshuaKGoldberg, line 128: [Style] Nit: as nice as these types are, there's no need for the fancy `consequent` specificity. Removing it doesn't cause any new type errors.

```suggestion
          node: TSESTree.IfStatement,
```

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -109,7 +116,105 @@ export default createRule<
...
+        ): void => {
+          if (!options.allowSuggestingOnIfStatements) {
+            return;
+          }
+          if (node.alternate) {
```
JoshuaKGoldberg, line 133: [Style] Nit: we can save a bit of space...

```suggestion
          if (!options.allowSuggestingOnIfStatements || node.alternate) {
```

omril1: I usually nitpick about this as well in PRs, but here I wanted the option toggling to be separate from the rule logic, but I don't mind that much.

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
omril1: @JoshuaKGoldberg Talking about this rule moving comments while the comments inside it where moved by other rule's auto fix 😄

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
omril1: The rule as it is (when `requireNullish` is false/default) is not completely type-safe, and I was wondering if it should continue this way with only suggesting instead of fixing ([example for both](https://deploy-preview-10137--typescript-eslint.netlify.app/play/#ts=5.5.2&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3gMyUgXDlBGggD7wDe8ARrIQGpICWwAYgK4pgbOrwBfANwAoEc2zwAFLiQA6OjACUVEfBx4FsKUtECxsrTHgAyExvmKA-HJ1CgA&eslintrc=N4KABGBECmAeAu0B2ATAzpAXGA2uCYA9IVAA4A2ArgOYCWSmAAvAJ6nRoDGATrafAFoO5evELdonAPYBbGchTQUAiQEdKtXkmoDW7AZwAWkgNb1qkfAF0ANPkjdK5DljCgCUZmw48%2Bg4aKEpBIAZtDcAlL8tFJIAIbkBoZx9K44MNzcUtyQNm74HpAJ5FIA7gDKNNQc8OYA8kgAkiHl8HGI8kjwGNjwjtB2HhAO0Oqa0AByTiJohq4hCWjQBWAAvlb4qyCrQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false)) I also messed it up and made it an autofix here, I could use some advice on this.

I also wanted to convey that the rule doesn't apply to any optional inside if, but only to single statement call expressions but it's too much.

Naming things is hard :)

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
omril1: `requireNullish` is also quite annoying in real-life code since typescript can lie about array access being nonNullable.

```ts
// with requireNullish: true
declare const foo: { baz: VoidFunction }[];
const bar = foo[999999];

bar && bar.baz();
```
This won't report even though it could be undefined and optional chaining could make sense.

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
JoshuaKGoldberg: Looking through https://typescript-eslint.io/rules/?=fixable, the first few that jump to mind are:

* [@typescript-eslint/consistent-type-definitions](https://typescript-eslint.io/rules/consistent-type-definitions)
* [@typescript-eslint/method-signature-style](https://typescript-eslint.io/rules/method-signature-style)
* [@typescript-eslint/prefer-function-type](https://typescript-eslint.io/rules/prefer-function-type)

I wouldn't trust any existing rule to be honest. It's entirely possible there are existing bugs with comments. You might end up filing separate issues on other rules out of this 😄

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
JoshuaKGoldberg: Haha, not that I know of.

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
JoshuaKGoldberg: Haha, it really is...

I think the new if statements should follow the same logic as existing code. `if (abc) { abc.def }` to me is mentally the same model as `abc && abc.def`. So as long as they're consistent, and the new option just stops the rule from reporting on the `if` case, I'm happy.

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
omril1: Is there a reason `no-useless-return` is not enabled?

JoshuaKGoldberg: Not that I know of 😄. It might just be that we never thought to look at it. I'd be in favor if you want to file a separate issue 😄

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -47,6 +48,11 @@ export default createRule<
         type: 'object',
         additionalProperties: false,
         properties: {
+          allowIfStatements: {
+            type: 'boolean',
+            description:
+              'Allow autofix to optional chain on if statements with a single call expression in the consequent.',
```
JoshuaKGoldberg, line 54: [Docs] This option name reads to me like it's allowlisting `if` statements. As in, turning this on tells the rule to ignore them. The opposite of what it does! Going off the current functionality, I think it'd be more appropriately named as `checkIfStatements`.

However, long-term I think we'll want to have the checking of `if` statements on by default. I expect most folks will prefer to have the rule check them as well, and only a minority will want to opt out using the option.

Proposal: let's have the option's functionality be inverted so that it's an opt-out for `if` statements. And, with that, a slightly more precise name:

```suggestion
          ignoreIfStatements: {
```

```suggestion
          allowIfStatements: {
            type: 'boolean',
            description:
              'Whether to ignore `if` statements with a single expression in the consequent.',
```

packages/eslint-plugin/src/rules/prefer-optional-chain.ts
```diff
@@ -109,7 +116,87 @@ export default createRule<
...
+          for (const operand of operands) {
+            if (operand.type === OperandValidity.Invalid) {
+              return;
+            }
+            currentChain.push(operand);
+          }
+        } else {
+          if (ifBodyExpression.type !== AST_NODE_TYPES.CallExpression) {
+            return;
+          }
```
JoshuaKGoldberg, line 181: [Bug] That's not for this rule to say. Ignoring unused property accesses is in the realm of [`@typescript-eslint/no-unused-expressions`](https://typescript-eslint.io/rules/no-unused-expressions). Folks should be able get this optional chaining complaint on code like:

```ts
if (a) {
  a.b;
}
```

There are codebases that use getters for control flow. I hate it, almost everyone hates it, but it's a legitimate choice that this rule shouldn't judge.  For example, `.to.be.*;` in Chai: https://www.chaijs.com/api/bdd/#method_null / https://www.chaijs.com/api/bdd/#extensible / etc.


```suggestion
```

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
```diff
@@ -242,13 +246,17 @@ function getReportRange(
 }
 
 function getReportDescriptor(
-  sourceCode: SourceCode,
+  context: RuleContext<
+    PreferOptionalChainMessageIds,
+    [PreferOptionalChainOptions]
+  >,
```
JoshuaKGoldberg, line 252: [Question] Why the change? Unrelated/unnecessary?

Unless there's a need to switch this, could you revert the change please?

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
```diff
@@ -401,7 +409,42 @@ function getReportDescriptor(
...
+      sourceCode.getTokenAfter(lastOperand.node)?.value === ';';
+    if (chainEndedWithSemicolon) {
+      newCode += ';';
+    }
+
+    const nodeBeforeTheComment = chainEndedWithSemicolon
+      ? lastOperand.node.parent
+      : lastOperand.node;
+    const commentsBefore = sourceCode.getCommentsBefore(chain[1].node);
+    const commentsAfter = sourceCode.getCommentsAfter(nodeBeforeTheComment);
```
JoshuaKGoldberg, line 425: [Bug] It's hard to write comments handling code that's specific to types of comments & indentation. This very specific, which makes it brittle. For example:

* Other comments such as `#! env` exist
* Non-`' '` indentation exists, i.e. `\t`

I think you'll have a much better time using `sourceCode.getText`. As a reference, `prefer-function-type` slices text to capture everything in it:

https://github.com/typescript-eslint/typescript-eslint/blob/7ea2074ec74de1a4a9699ad96602b129c6f66762/packages/eslint-plugin/src/rules/prefer-function-type.ts#L109-L111

Does that kind of more general approach not work here?

JoshuaKGoldberg (changes requested):
Progress! 🚀

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
omril1: Yeah, mistake, the change was from an older approach I tried, I'll revert

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
omril1: I kept a side branch where I tried something like that, but it got ugly real fast it almost doubled the code to handle comments and it still had issues with indentation.

It feels like the hardest part of lint fixers is dealing with comments because the tooling for text manipulation on AST nodes is so much better.

JoshuaKGoldberg: Yeah 🥲 I really wish ASTs treated them the same...

JoshuaKGoldberg:
👋 Hey @omril1! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

omril1:
> 👋 Hey @omril1! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

Hey, I'll try to advance it a bit more in the weekend, I still need to handle all the moving comments edge cases.

packages/eslint-plugin/docs/rules/prefer-optional-chain.mdx
```diff
@@ -265,6 +265,32 @@ thing && thing.toString();
...
+{/* insert option description */}
+
+Examples of code for this rule with `{ ignoreIfStatements: false }`:
+
+<Tabs>
+
+<TabItem value="❌ Incorrect">
+
+```ts option='{ "ignoreIfStatements": false }'
+if (foo) {
```
kirkwaiblinger, line 279: One thing I'm noticing is that this handles the `if (foo)` form but not the `if (foo != null)` which is syntactically guaranteed to be equivalent. Do we plan to add handling for that case into this PR? ([playground examples](https://deploy-preview-10137--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF0VC0BGEIAXPABQCU8AvAHzwBuSAlsPAD7woCuuA3AFD8A9EPhwADkhgY0iKXDAYIAT35MAZmTA58hSgG9%2B8bLigEQFAQF9BI%2BGgAWSHqxxokYkJOlrNpbabm8ACEVJw8EAZGJroW5NaCoJCwCMjoWGAOUEwoZoQk%2BvBQJJgwOQDm8Fbs4Xy2ohJSMnIwCkqqGlpZOXkgUcaZ2bnmAHRQI4MwAIIYpAAM8fw2wqKOzhCuEO6e3hi%2BXUO9IWHcuP2I3cOEYxNZ07MLS0A&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uigB3UdCaRwYAL4gNQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false))

Recent PRs in prefer-nullish-coalescing have handled these types of cases, see https://github.com/typescript-eslint/typescript-eslint/pull/10708, https://github.com/typescript-eslint/typescript-eslint/pull/10517 in case there are helpful code examples

omril1: Hi, are you asking me or the other typescript-eslint maintainers?
I haven't considered this case at all here.

omril1: From playing around with trying to support this it doesn't seem like a trivial change.
It seems to me that `compareNodesUncached` needs to be modified to recurse the nodes inside the `if`'s test node, but it doesn't work.

I'd prefer if we open a followup, this rule has a lot of logic already and I can think of many more missing cases it can match for.

kirkwaiblinger: Filed https://github.com/typescript-eslint/typescript-eslint/issues/10842 👍

omril1:
@JoshuaKGoldberg ping, I saw a lot of your other work in the typescript/eslint/typescript-eslint repos, and I get that you are busy, but can we please advance this?

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -665,35 +665,393 @@ describe('|| {}', () => {
...
+            // comment 2
+          }
+        `,
+        errors: [{ messageId: 'preferOptionalChain' }],
+        options: [{ ignoreIfStatements: false }],
+        output: `
+          declare const foo: undefined | { bar: () => void };
+          foo?.bar();
+          // comment 1
+            // comment 2
```
omril1, line 922: I took the suggestion from here: https://github.com/typescript-eslint/typescript-eslint/pull/10137#discussion_r1854400776
This preserves the range between the first and last comments, so the indent before the first comment is modified.

I feel like it's impossible to get 100% correctness when moving comments. Do we want to change the indent of all of them?

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
```diff
@@ -401,7 +405,85 @@ function getReportDescriptor(
...
+    if (chainEndedWithSemicolon) {
+      newCode += ';';
+    }
+
+    const nodeBeforeTheComment = chainEndedWithSemicolon
+      ? lastOperand.node.parent
+      : lastOperand.node;
+
+    const commentsBefore = sourceCode.getCommentsBefore(chain[1].node);
+    const commentsAfter = sourceCode.getCommentsAfter(nodeBeforeTheComment);
```
JoshuaKGoldberg, line 422: Continuing https://github.com/typescript-eslint/typescript-eslint/pull/10137/files#r1855228237:

> handle comments and it still had issues with indentation.

I'm not convinced this current approach of manually handling comments is the right one. If you do a full search on rules you'll see that it's very rare we have to dig into `getCommentsBefore` or other comment APIs.

Example: see the top comment I added in `test.ts` with the single-letter test cases. A few get missed even with all this added comment & token logic.


Instead of relocating comments, I think it'd less code and more reliable behavior to have it add and delete specific nodes and tokens. No need to manage comment locations if the fixer isn't removing ranges that include them. As in, it would:

* Delete the `if`
* Delete the `(`
* Delete the `)`
* Delete the `{` if it exists
* Add a `?.`
* Delete the `}` if it exists

Could you please try that out? If I'm wrong and this strategy is way more code & also brittle, we'll at least want to know for sure.

Note, we don't generally have to deal with indentation - that's a formatter's concern. As long as the comment exists in roughly a similar location we're happy.

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
JoshuaKGoldberg: We don't care deeply about indentation or spacing in general, that's a formatter's concern. As long as the comment is in a reasonably similar location that seems to match up we're happy.

If we can do some very basic work to make the comments a little prettier that's fine. But we don't want to add a lot of formatting-specific code that makes the file harder to read.

JoshuaKGoldberg:
👋 Just checking in @omril1, is this still something you have time for?

omril1:
> 👋 Just checking in @omril1, is this still something you have time for?

@JoshuaKGoldberg I think I'm giving up on this. The rule as it is currently doesn't handle all of the comments between the optional chaining nodes, and this PR is reusing all of the existing logic from `analyzeChain`.

I have no interest in expanding the scope of the PR, as it was already hard enough for me to integrate `IfStatement` into the current logic.

JoshuaKGoldberg:
Ah, I'm sorry to hear that - especially since I think this is largely my fault for trying to sling a lot of annoying comments work on you. I think what was missing on my end was I'd thought there was existing comment handling that was being regressed here. But I think it'd be better to get this in and file a followup issue for the comments. Do you think you'd have time to just wontfix my comments there?

omril1:
> Ah, I'm sorry to hear that - especially since I think this is largely my fault for trying to sling a lot of annoying comments work on you. I think what was missing on my end was I'd thought there was existing comment handling that was being regressed here. But I think it'd be better to get this in and file a followup issue for the comments. Do you think you'd have time to just wontfix my comments there?

I'm not sure what 'wontfix' mean, do you want me to just acknowledge it as out of scope in each discussion?

I'll probably try to pick easier PRs to work on next time, I just liked how this rule helped simplify the awful legacy codebases I worked on 😄

JoshuaKGoldberg:
Exactly: whatever you don't have scope to do, mention that in the conversation. 

And yeah, this is one of my favorite rules too. Really a pity what a nightmare working with its implementation is. 🥲

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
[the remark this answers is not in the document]
omril1: I tested on the playground how the current rule works in main, and this is what I got (The return type of the declaration should be `undefined` instead of `void` as well for some reason)
```ts
declare const foo: undefined | { bar: () => undefined };
/* h */ foo /* i */ /* j */ && /* k */ /* l */ foo /* m */ /* n */. /* o */ /* p */ bar /* q */ /* r */() /* s */ /* t */;
```
Has a suggestion fixer to
```ts
declare const foo: undefined | { bar: () => undefined };
/* h */ foo?.bar() /* s */ /* t */;
```

I think this is out of the scope of the PR, as the code reuses this logic only to support `if` statements.

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
omril1: I'd rather this be considered out of the PR's scope.

As I mentioned above, the rule currently doesn't handle comments well between the optionally chained nodes, and the logic that I wanted to add tried to reuse it as much as possible.

I looked around the codebase for different ways comments are handled, and this rule is one of the hardest ones, as the comments are not easy to relate between the transformed parts.

kirkwaiblinger:
Closing this PR as it's been stale for a long time without activity. Feel free to reopen if you have time - but no worries if not! If anybody wants to drive it forward, please do post your own PR. If you use this as a start, please add a co-author attribution at the end of your PR description. Thanks! 😊

PR #10139, opened 2024-10-12
fix(visitor-keys): reorder `TSSatisfiesExpression` and `TSTypeAssertion`

### LINKED ISSUE #10112: Bug: [no-useless-assignment] False positive in code using `satisfies`

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.4&showAST=es&fileType=.ts&code=MYewdgzgLgBAZiEMC8MDkCRoNwChcCWcMAFFAE4CuApgJQwDeAvviQNqYC6MEAhlAQhwC1CDACC5crwCeAHmjkCYAOYA%2BWgDoE5AKK9gACxKKUaxrhgxQkEABtqmuyBUmKtPEw%2B4gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaWZRJZZQgQwoEsBzXAW0VwBd0pFpp9pJwYAL4ghQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
const foo = 'foo';

if (true) {}

([foo] satisfies Array<string>).forEach(str => {
  console.log(str);
});
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "no-useless-assignment": "error",
  },
};
```

### tsconfig

_No response_

### Expected Result

The variable `foo` should not violate the `no-useless-assignment` rule.

### Actual Result

The `no-useless-assignment` rule is reported for `foo`.

### Additional Info

I'm sorry if I used an incorrect issue template or if this should be reported in the ESLint repository, but I had some trouble classifying this issue (for example whether this should be a new extension rule request) since I'm not familiar with ESLint internals. I decided on reporting a rule bug here because I wasn't able to reproduce this without the `satisfies` operator.

Removing either the `satisfies` operator or the preceding `if` statement makes the error report disappear. Furthermore, the `if` statement can be replaced with some other statements (e.g., `while (false) {}`) and the issue persists. I also tried to debug this problem by comparing what happens with and without the `satisfies` operator, and it appears that node ranges differ, making ESLint believe that `foo` is not used. Specifically, the third comparison in the [`isIdentifierUsedInSegment()`](https://github.com/eslint/eslint/blob/5dcbc519b729698be651bdbddb39da774cb70bf1/lib/rules/no-useless-assignment.js#L212-L221) returns `false` when `satisfies` is present.

abrahamguo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10112

### DIFF
--- a/packages/visitor-keys/src/visitor-keys.ts
+++ b/packages/visitor-keys/src/visitor-keys.ts
@@ -253,8 +253,6 @@ const additionalKeys: AdditionalKeys = {
   TSTupleType: ['elementTypes'],
   TSTypeAliasDeclaration: ['id', 'typeParameters', 'typeAnnotation'],
   TSTypeAnnotation: ['typeAnnotation'],
-  // this is intentionally different to SharedVisitorKeys.AsExpression because
-  // the type annotation comes first in the source code
   TSTypeAssertion: ['typeAnnotation', 'expression'],
   TSTypeLiteral: ['members'],
   TSTypeOperator: ['typeAnnotation'],
--- a/packages/visitor-keys/src/visitor-keys.ts
+++ b/packages/visitor-keys/src/visitor-keys.ts
@@ -143,6 +143,7 @@ const SharedVisitorKeys = (() => {
   return {
     AbstractPropertyDefinition: ['decorators', 'key', 'typeAnnotation'],
     AnonymousFunction,
+    AsExpression: ['expression', 'typeAnnotation'],
     ClassDeclaration: [
       'decorators',
       'id',
@@ -155,7 +156,6 @@ const SharedVisitorKeys = (() => {
     Function: ['id', ...AnonymousFunction],
     FunctionType,
     PropertyDefinition: [...AbstractPropertyDefinition, 'value'],
-    TypeAssertion: ['expression', 'typeAnnotation'],
   } as const;
 })();
 
@@ -193,7 +193,7 @@ const additionalKeys: AdditionalKeys = {
   TSAbstractPropertyDefinition: SharedVisitorKeys.AbstractPropertyDefinition,
   TSAnyKeyword: [],
   TSArrayType: ['elementType'],
-  TSAsExpression: SharedVisitorKeys.TypeAssertion,
+  TSAsExpression: SharedVisitorKeys.AsExpression,
   TSAsyncKeyword: [],
   TSBigIntKeyword: [],
   TSBooleanKeyword: [],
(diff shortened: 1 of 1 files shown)


packages/visitor-keys/src/visitor-keys.ts
```diff
@@ -244,12 +244,7 @@ const additionalKeys: AdditionalKeys = {
   TSQualifiedName: ['left', 'right'],
   TSReadonlyKeyword: [],
   TSRestType: ['typeAnnotation'],
-  TSSatisfiesExpression: [
```
kirkwaiblinger, line 247: Just noting for the record that it looks like this was set up this way from the start in https://github.com/typescript-eslint/typescript-eslint/pull/5717. The erroneous/misleading comment was added in https://github.com/typescript-eslint/typescript-eslint/pull/5926/files with no particular comment.

Just saying this all to do due diligence to investigate whether there was any particular reason in the paper trail that it was set up that way in the first place, but it doesn't seem there is (see @bradzacher's comment https://github.com/eslint/eslint/issues/19002#issuecomment-2402139894)

packages/visitor-keys/src/visitor-keys.ts
```diff
@@ -258,7 +253,9 @@ const additionalKeys: AdditionalKeys = {
   TSTupleType: ['elementTypes'],
   TSTypeAliasDeclaration: ['id', 'typeParameters', 'typeAnnotation'],
   TSTypeAnnotation: ['typeAnnotation'],
-  TSTypeAssertion: SharedVisitorKeys.TypeAssertion,
+  // this is intentionally different to SharedVisitorKeys.AsExpression because
```
kirkwaiblinger, line 256: Personally I don't think this comment is necessary, and would be better omitted (especially considering the previous incarnation of it was wrong). But it's not important one way or another 🤷‍♂️

kirkwaiblinger (dismissed):

packages/visitor-keys/src/visitor-keys.ts
```diff
@@ -258,7 +253,9 @@ const additionalKeys: AdditionalKeys = {
   TSTupleType: ['elementTypes'],
   TSTypeAliasDeclaration: ['id', 'typeParameters', 'typeAnnotation'],
   TSTypeAnnotation: ['typeAnnotation'],
-  TSTypeAssertion: SharedVisitorKeys.TypeAssertion,
```
kirkwaiblinger, line 261: Also noting for the record that this was previously, apparently unintentionally, flipped in https://github.com/typescript-eslint/typescript-eslint/pull/5926/files#diff-04c8c6315c623a82f3004fbfe093162ab6811b1f207cf1832d495b2979466f5eL139

packages/visitor-keys/src/visitor-keys.ts
bradzacher: I don't remember why I did this.
It was probably a cock up and I copied `TSTypeAssertion` rather than `TSAsExpression`.

If no tests fail then I assume it was defs just a bug 🤷

packages/visitor-keys/src/visitor-keys.ts
bradzacher: let's remove it

bradzacher (approved):

bradzacher:
Thanks heaps!

kirkwaiblinger (approved):
👍👍

PR #10149, opened 2024-10-15
feat(eslint-plugin): [no-unnecessary-type-parameters] add suggestion fixer

### LINKED ISSUE #9663: Enhancement: [no-unnecessary-type-parameters] Add suggestions for simple cases.

https://typescript-eslint.io/rules/no-unnecessary-type-parameters/

### Description

Add suggestions with the intended fix.

### Fail

```TypeScript
function toUnknown<T>(x: T): unknown {
    return x;
}

function withExtends<T extends number | string>(x: T): unknown {
    return Math.random();
}
```

### Pass

```TypeScript
// replace unconstrained input generic with unknown
function toUnknown(x: unknown): unknown {
   return x;
}

// replace constrained input generic with its constraint type
function withExtends(x: number | string): unknown {
    return Math.random();
}
```

### Additional Info

There are probably more cases where we could make suggestions, too. These are just the first things that came to mind. I think that adding these suggestions would have high value to help the user understood why the rule is reporting, since this rule is quite subtle

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9663 

## Overview

Offers to replace the unnecessary type parameter with its constraint (and remove the type parameter declaration).

Special case is that `T extends any` gets replaced with `unknown` instead since that's how it behaves.


packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -84,6 +92,113 @@ export default createRule({
...
+                ).params;
+
+                const index = typeParams.indexOf(esTypeParameter);
+
+                /* istanbul ignore if: this is an assertion that should never happen */
+                if (index === -1) {
+                  throw new Error(
+                    "type parameter should be in node's type parameters",
+                  );
+                } else if (typeParams.length === 1) {
```
JoshuaKGoldberg, line 129: Why include this, if it's not something that can even be unit tested?

```suggestion
                if (typeParams.length === 1) {
```

JoshuaKGoldberg (approved):
Nice!

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
kirkwaiblinger: My paranoia is: the rule fixer _assumes_ the type parameter to be removed/replaced is _in_ the method/function/class's declared list of type parameters of the node being inspected. But the way we get the type parameter to be removed is indirect. 
1. ESTree selector inspects function/method: https://github.com/typescript-eslint/typescript-eslint/blob/9a4274095907d652d2abf07ac4f733a4eea6c121/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts#L206-L226
2. We switch into the ts AST node world with a type assertion: https://github.com/typescript-eslint/typescript-eslint/blob/9a4274095907d652d2abf07ac4f733a4eea6c121/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts#L43-L46
3. We traverse to its type parameters and back out to the ESTree type parameter node https://github.com/typescript-eslint/typescript-eslint/blob/9a4274095907d652d2abf07ac4f733a4eea6c121/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts#L54-L58
4. We _assume_ that the ESTree node from step 1. is the _grandparent_ of the type parameter from 3, in order to get its type parameter declaration (which we have to assert even exists): https://github.com/typescript-eslint/typescript-eslint/blob/9a4274095907d652d2abf07ac4f733a4eea6c121/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts#L117-L120
5. We need to figure out what position the reported ESTree type parameter is within that asserted-to-exist type parameter declaration's list of type parameters. My goal was to _double-check_ here that the reported type parameter really _is_ in that list, since, we have come here by two different paths completely, and it doesn't seem that far-fetched that a change to the rule logic could break some of our assumptions.

For reference, the ASTs look like
```text
ESTREE: node that we inspect ---(maybe has) ---> TSTypeParameterDeclaration -----> Array<TypeParameter>
TS AST: node that we inspect -----> Array<TypeParameterDeclaration>
```

kirkwaiblinger: Anyways, I can go ahead and remove it if you prefer. Just one of those things that I prefer to do so that we fail faster  and more understandably when a nonlocal code change breaks some of our assumptions.

JoshuaKGoldberg: Yeah this kind of strategy used to be a lot more common in the project. But it practically never paid off 😞. Our test coverage is so high & users are so active in reporting that it just ended up being a lot of code bloat. I'm still a -1 on it, but if you want to advocate for it I wouldn't be opposed to hearing more inputs from the team.

kirkwaiblinger: i chop 🔪

JoshuaKGoldberg (approved):
Since this only adds to the linting side of things, rather than changing functionality, I say let's merge for today's release. 🚀

PR #10160, opened 2024-10-16
feat(eslint-plugin): [consistent-indexed-object-style] report mapped types

### LINKED ISSUE #6598: Enhancement: [consistent-indexed-object-style] forbidden from using the "in" keyword in index signature if `Record` is preferred

https://typescript-eslint.io/rules/consistent-indexed-object-style/

### Description

When `"record"` option is chosen in `@typescript-eslint/consistent-indexed-object-style`, we should also be forbidden from using the `"in"` keyword in index signature. It's particularly useful for literal types.

Here's the justification:
- we should prefer `Record` versus `Index Signature` whenever we want to use literal types, as index signature accepts only `string`, `number` or `symbol` as key type. 
- And when `Record` is preferred, it also brings an advantage of ensuring no missing property in literal types -> which further ensures type safety, and avoid getting unexpected `undefined` case from non-existing key.
- However, such advantage will not be kept since the current `@typescript-eslint/consistent-indexed-object-style` rule allows people using `{ [key in Option] : string}` 

### Fail

```typescript
/* eslint @typescript-eslint/consistent-indexed-object-style: ["error", "record"] */

type Option = "red" | "yellow";

const optionToString: { [key in Option]: string} = {
   "red": "this is red" // won't be warned that "yellow" is missing
}
```

### Pass

```typescript
/* eslint @typescript-eslint/consistent-indexed-object-style: ["error", "record"] */

type Option = "red" | "yellow";

// if we use Record, it will prompt error if any property is missing; 
// E.g. it should say "Property 'yellow' is missing in type '{ red: string; }' but required in type 'Record<Option, string>'." 
// if we just have "red" as the only property.
const optionToString: Record<Option, string> = {
   "red": "this is red",
   "yellow": "this is yellow",
}


// Or, if we indeed want to get undefined for some keys; making them explicit when defining
// the type and value is better and clearer.
const optionToString: Record<Option, string | undefined> = {
   "red": "this is red",
   "yellow": undefined,
}
```

### Additional Info

_No response_

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6598 

## Overview

This reports on mapped types that can be converted to `Record`s. _However_, note that not all mapped types can be converted to `Record`s, so the rule only reports when they can be converted.

```ts
// can't be converted to record, because `K` is used in the computed type. Not reported.
type Foo1 = {
  [K in T]: K
};

// Can be converted to record. Reported.
type Foo2 = {
  [K in T]: T
}
// fixes to 
type Foo2 = Record<T, T>

// Can be converted to record. Reported.
type WithModifiers = {
  +readonly [key in T]+?: T
}
// fixes to 
type WithModifiers = Readonly<Partial<Record<T, T>>>


// WEIRD CASE: Cannot be converted to Record due to keyof (mapped type syntax preserves readonly modifiers in the keys of T)
type WithKeyof = {
  [K in keyof T]: T
}

// WEIRD CASE: Can be converted to Record due to keyof being parenthesized
type WithKeyof = {
  [K in (keyof T)]: T
}
```

For more detail on the "weird case", study the following example. (Thanks to @Retsam for the helpful discord conversation in https://discord.com/channels/508357248330760243/508357638677856287/1296364542498177046!)
```ts
type MyObject = {
  required: string,
  optional?: number,
}

type BooleanValuesNoParens = {
  [key in keyof MyObject]: boolean
}

// valid
const bv: BooleanValuesNoParens = {
  required: true,
}

type BooleanValuesWithParens = {
  [key in (keyof MyObject)]: boolean
}

// TS ERROR: Property 'optional' is missing
const bv: BooleanValuesWithParens = {
  required: true,
}
```

This actually has practical application inside our repo!

https://github.com/typescript-eslint/typescript-eslint/blob/6250dabd3048eab32d7624ab8d697a6e09bb6c9c/packages/utils/src/ts-eslint/Parser.ts#L38-L41

***

There is **one open question**: should we report on cases where the `-readonly` modifier is used? I've gone with the approach for now that we should report it even though we can't provide a fix. Open to differing opinions on this, though.

```ts
type PlusReadonly = {
  +readonly [K in T]: T
}
// equivalent to 
type PlusReadonly = Readonly<Record<T, T>>

type PlusOptional = {
  [K in T]+?: T
}
// equivalent to
type PlusOptional = Partial<Record<T, T>>

type MinusOptional = {
  [K in T]-?: T
}
// equivalent to
type MinusOptional = Required<Record<T, T>>


type MinusReadonly = {
  -readonly [K in T]: T
}
// no built-in equivalent :( You can write your own easily though.
type Mutable<T> = {
  -readonly [K in keyof T]: T[K]
}
// then,
type MinusReadonly = Mutable<Record<T, T>>
```


kirkwaiblinger:
## IGNORE THIS COMMENT
### But please don't delete it

This is a draft of a note to myself and I may followup on some things in it

***

<s>
I got this working in #10160, but I'm unsure whether we can actually do it safely.

***

TL;DR:
Seems like 
- index signature is applicable to infinite types only
- record is applicable to any types, but can't distinguish between required and optional keys
- mapped type is applicable to any types, but is more powerful than records when acting on finite types

Therefore,
| from | to | ok or not? |
| -- | -- | -- |
| index | record | ok |
| index | mapped type | ok |
| record | index | TS Error if key type is not infinite |
| record | mapped type | ok (but would have to parenthesize `keyof` types to exactly preserve behavior) |
| mapped type | record | broken if key type is not infinite |
| mapped type | record | broken if key type is finite |

(where it is assumed that the mapped type does not use its key in the computed type).

***

While doing this I learned that while a `Record` is trivially replaceable with a mapped type, a mapped type is more than a record, so the reverse is not always true. 

In fact, the modifiers, i.e. `+readonly`/`-?`, aren't the problem; as (except for `-readonly`) they can be replaced by built-in `Readonly<T>`/`Partial<T>`/`Required<T>`.

The problem is that mapped types behave differently when the constraint is a _non-infinite type_.

I learned this in our repo via this line, which would be reported by the new check: 

https://github.com/typescript-eslint/typescript-eslint/blob/6250dabd3048eab32d7624ab8d697a6e09bb6c9c/packages/utils/src/ts-eslint/Parser.ts#L35-L41

Here's a minimal demo:

[Playground](https://typescript-eslint.io/play/#ts=5.3.3&showAST=es&fileType=.tsx&code=C4TwDgpgBAsiDyAjAVhAxsKBeKBvAUFFAE4QCOArgJakAmAXFAM7DFUB2A5gDSFQD2YYFX7sAhgBsA-I3YUAtogjFeAX3z5QkKACF%2B-CRDHsAapIoQm2PHwDaAawggoHKI5D8AZrAQp0wAF1GRH1DY3x1fAB6KKgAN0kqWnw0URYoRDjGPQMjU3NLawIiUkoaCAYoVgs1DU1waByw-IkLJgB1KmAACwAldH5iWmt%2B1KGAHncvHyRUDG4M0LyAPgBuDRioABUAZSgAUV7e%2BF7GAAViQWVQKAByQWFRSVuXK3kqJiYOThS0zEziNklsYzK1LJ0eqNBsMcMUSORqHRGNUIBEgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

```ts
type MyObject = {
  required: string,
  optional?: number,
}

type BooleanValues = {
  [key in keyof MyObject]: boolean
}

// valid
const bv: BooleanValues = {
  required: true,
}


type BooleanValuesWithRecord = Record<keyof MyObject, boolean>;

// TS ERROR: Property 'optional' is missing
const bvr: BooleanValuesWithRecord = {
  required: true
}
```

Note that you can even trigger this by putting parens around `keyof MyObject`, like so:
```ts
type MyObject = {
  required: string,
  optional?: number,
}

type BooleanValues = {
  [key in (keyof MyObject)]: boolean
}

// TS ERROR: Property 'optional' is missing
const bv: BooleanValues = {
  required: true,
}
```

In other words, the `keyof` operator gets special treatment in the constraint of a mapped type that can't be replicated by converting to a record. And I think we'd need the type checker if we wanted to try to figure out when that behavior is being used.

***

NOTE, though, that this also makes me realize that not all records can be converted to indexed access, for example:

[Playground](https://typescript-eslint.io/play/#ts=5.3.3&fileType=.tsx&code=C4TwDgpgBASlC8sIGMD2AnAJgHgOQDNVVcoAfKXAIwEN1cAaKAZ2HQEsA7AcwD4BuIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6MgeyeUuX0Ra1mAE0QAPRMNocARgCtEZOn0JJ0YANqQA7gENoTSABooI8bR4BzJrvyxoiSAF1wYAL4g3QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

### Code

```ts
type R = Record<'foo' | 'bar', string>;
```

### Eslint config

```json
{
  "rules": {
    "@typescript-eslint/consistent-indexed-object-style": ["warn", "index-signature"]
  }
}
```

The above example autofixes to 

```ts
type R = { [key: 'foo' | 'bar']: string };
```

which is illegal TS....

... so maybe we aren't too worried about these edge cases and we should just use disable comments for these cases?

***
</s>

packages/eslint-plugin/src/rules/consistent-indexed-object-style.ts
```diff
@@ -142,6 +143,69 @@ export default createRule<Options, MessageIds>({
...
+                const valueType = context.sourceCode.getText(
+                  node.typeAnnotation,
+                );
+
+                let recordText = `Record<${keyType}, ${valueType}>`;
+
+                if (node.optional === '+' || node.optional === true) {
+                  recordText = `Partial<${recordText}>`;
+                } else if (node.optional === '-') {
+                  recordText = `Required<${recordText}>`;
```
kirkwaiblinger, line 197: I'm realizing that `Required<Record<K, V>>` is probably always redundant, and I think can always just be `Record<K, V>`

JoshuaKGoldberg: Talked 1:1 - tricky and maybe not determinable. Sticking with it as-is.

JoshuaKGoldberg (approved):
Looks great to me, thanks! 🚀 

Excited to have this extra bit of consistency.

Thompson1985 (approved):

PR #10162, opened 2024-10-17
feat(eslint-plugin): add `no-inferred-any` rule

butchler:
## PR Checklist

## Overview

I apologize in advance for creating a pull request without first creating a ticket. I considered creating a ticket to explain the linting rule first, but because this rule was already fully implemented, I figured the best way to explain the rule might be to share implementation and tests themselves. Please let me know if you would like me to create a ticket anyway.

This linting rule is one that I created at work about 4 years ago (I have permission from my bosses to share this code), and we've been using it in our relatively large mixed TypeScript/JavaScript project to help catch accidental usages of `any`. I wanted to share it when I first created it, but completely forgot to and just remembered it recently. 😅

In a nutshell, the rule tries to catch cases where an `any` type is accidentally used due to type inference. For example:

```ts
const response = await fetch('/');
const data = await response.json();
// Oops, I forgot that TypeScript's built-in DOM API return types often use `any` instead of `unknown`!
// Ideally I should be forced to assert that the data matches the shape I expect,
// but `data` is inferred as `any` so there's essentially no type checking here.
data.nested.property;
```

It's particularly useful in projects with a mix of untyped JavaScript code and typed TypeScript code that use `allowJs`, because importing from JS modules will implicitly use the `any` type. But because it's relatively common for NPM packages or even TypeScript's built-in DOM types to use `any`, I think it could be useful in any TypeScript project.

Does this seem like a rule that would be worth adding to `typescript-eslint`?

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-inferred-any.mdx
+++ b/packages/eslint-plugin/docs/rules/no-inferred-any.mdx
@@ -10,7 +10,7 @@ import TabItem from '@theme/TabItem';
 > See **https://typescript-eslint.io/rules/no-inferred-any** for documentation.
 
 Even if you never explicitly use the `any` type in your own code, you might
-unknownigly write code that variables with the type `any` or calls functions
+unknowingly write code that variables with the type `any` or calls functions
 which return `any`. These `any` types might come from:
 
 - Untyped JavaScript code in the same project, in a mixed TypeScript/JavaScript project
@@ -68,7 +68,7 @@ console.log(data2.property);
 
 By default, variables, exports, and function return types are all checked by
 this rule. But specific kinds of checks can be disabled, mainly to make it
-easier to incrementically introduce the rule into an existing codebase.
+easier to incrementally introduce the rule into an existing codebase.
 
 ### `checkExports`
 
--- a/packages/eslint-plugin/docs/rules/no-inferred-any.mdx
+++ b/packages/eslint-plugin/docs/rules/no-inferred-any.mdx
@@ -1,0 +1,163 @@
+---
+description: 'Disallow implicitly inferred `any` type.'
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-inferred-any** for documentation.
+
+Even if you never explicitly use the `any` type in your own code, you might
+unknownigly write code that variables with the type `any` or calls functions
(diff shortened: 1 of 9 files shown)


abrahamguo:
This is already built-in to `typescript-eslint` via the `no-unsafe-*` family of rules.

In this specific case, this specific error is handled by `@typescript-eslint/no-unsafe-member-access`: [playground link](https://typescript-eslint.io/play#ts=5.5.4&showAST=es&fileType=.tsx&code=MYewdgzgLgBATgUwgB3BBMC8MCGB3HAS1gDMEpgALACgHIB6WgSgG4AoUSWAExyhyy4CxeElSQEAOgBWEcNVZt69GAHkQyCABoYASRgkQcAOYhYUSnxgAVAJ7IEAZWBxCyKLQgwARgFdCADZQALSEYDAAIqoAsjAAggAK%2BohQvnDhUPZIMCAkUAjhvugwAAY4YLYlMGHQCDjcOSSlvmAA1mAgeGAlAIRKKrrcdQEBtnowEJQgvgEN3hiGcMAIDVAguBDocOaWOxi8-DAAtnxU2RYYkzgO4wgAHg7AUFr9Pr6wJQc4VYReYWRwRANHBeMoVKpyGAXRCeGBIdBgKCEHAjMYdKFZGBnYCtMLGGCUBCISRsL6SMBIfLcSTIOAaImZdhse6obYwADeAF8WEA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tieQEMAZoloBbRCIBGiaLT5kyKVBkjToHaJHBgAviG1A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

bradzacher:
We already have this - see the `no-unsafe-*` rules.

https://typescript-eslint.io/rules/no-unsafe-argument
https://typescript-eslint.io/rules/no-unsafe-assignment
https://typescript-eslint.io/rules/no-unsafe-call
https://typescript-eslint.io/rules/no-unsafe-member-access
https://typescript-eslint.io/rules/no-unsafe-return

These rules combine to cover all cases where you can operate on an `any`.

butchler:
I looked back at my commits and it seemed like I started working on this linting rule in late 2019, just before the `no-unsafe-*` rules were added, so that explains why I didn't notice those rules at the time. 😅 Glad to know that they're built into `typescript-eslint` now! 🙂

PR #10182, opened 2024-10-20
feat(eslint-plugin): [no-unnecessary-type-conversion] add rule

### LINKED ISSUE #8515: Rule proposal: no-unnecessary-cast for strings (maybe numbers too)

### Description

I'd like `String(something)`, `something.toString()`, and perhaps `something as string` to show an error if something is a `string` already, but to be allowed if it's a `string | number` or even a `string | number | undefined`. This is similar to the proposal at https://github.com/typescript-eslint/typescript-eslint/issues/6385 but it appears there're some issues with doing this with numbers that I don't think applies to strings.

In my case, I would like to do this because I am working with a library that exports a `string | number | undefined`, but I believe their type is incorrect, and I would like to essentially be notified that I can remove my casts if the type changes and becomes a `string` instead, and I believe similar use cases would not be exceeding rare, where one might be able to catch useless casts after updating one type somewhere, whether in a library or in their own type definitions. My feeling is that if this were turned on it would trigger errors in more than a couple of codebases. In reference to the similar issue for numbers, I do support the implementation of that as well for similar reasons.

### Fail Cases

```TypeScript
let replace = 'me';
replace = String(replace)
```

### Pass Cases

```TypeScript
let replace: string | number = 'me';
replace = String(replace)
```

### Additional Info

This is also similar to eslint's https://eslint.org/docs/latest/rules/no-extra-boolean-cast, which I'm actually not sure how they achieve without typescript, but that rule makes a lot of sense to me as well.

skondrashov:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8515

## Overview

Shows autofixable errors for:

```typescript
// setup
let str = "string";
const num = 1;
const bool = true;
const big = BigInt(1);

// begin invalid cases

String("hello");
"hello".toString();
"" + "hello";
"hello" + "";

String(1 + "");
(1 + "").toString();
"" + (1 + "");
1 + "" + "";

String(str);
str.toString();
"" + str;
str + "";
str += "";

Number(+"1");
+(+"1");
~~+"1";

Number(num);
+num;
~~num;

Boolean(bool);
!!bool;

BigInt(big);
```

with the following error messaging:

```
   9:2  error  Passing a string to String() does not change the type or value of the string            @typescript-eslint/no-unnecessary-coercion
  10:9  error  Using .toString() on a string does not change the type or value of the string           @typescript-eslint/no-unnecessary-coercion
  11:2  error  Concatenating '' with a string does not change the type or value of the string          @typescript-eslint/no-unnecessary-coercion
  12:9  error  Concatenating a string with '' does not change the type or value of the string          @typescript-eslint/no-unnecessary-coercion
  14:2  error  Passing a string to String() does not change the type or value of the string            @typescript-eslint/no-unnecessary-coercion
  15:9  error  Using .toString() on a string does not change the type or value of the string           @typescript-eslint/no-unnecessary-coercion
  16:2  error  Concatenating '' with a string does not change the type or value of the string          @typescript-eslint/no-unnecessary-coercion
  17:8  error  Concatenating a string with '' does not change the type or value of the string          @typescript-eslint/no-unnecessary-coercion
  19:2  error  Passing a string to String() does not change the type or value of the string            @typescript-eslint/no-unnecessary-coercion
  20:5  error  Using .toString() on a string does not change the type or value of the string           @typescript-eslint/no-unnecessary-coercion
  21:2  error  Concatenating '' with a string does not change the type or value of the string          @typescript-eslint/no-unnecessary-coercion
  22:5  error  Concatenating a string with '' does not change the type or value of the string          @typescript-eslint/no-unnecessary-coercion
  23:5  error  Concatenating a string with '' does not change the type or value of the string          @typescript-eslint/no-unnecessary-coercion
  25:2  error  Passing a number to Number() does not change the type or value of the number            @typescript-eslint/no-unnecessary-coercion
  26:2  error  Using the unary + operator on a number does not change the type or value of the number  @typescript-eslint/no-unnecessary-coercion
  27:2  error  Using ~~ on a number does not change the type or value of the number                    @typescript-eslint/no-unnecessary-coercion
  29:2  error  Passing a number to Number() does not change the type or value of the number            @typescript-eslint/no-unnecessary-coercion
  30:2  error  Using the unary + operator on a number does not change the type or value of the number  @typescript-eslint/no-unnecessary-coercion
  31:2  error  Using ~~ on a number does not change the type or value of the number                    @typescript-eslint/no-unnecessary-coercion
  33:2  error  Passing a boolean to Boolean() does not change the type or value of the boolean         @typescript-eslint/no-unnecessary-coercion
  34:2  error  Using !! on a boolean does not change the type or value of the boolean                  @typescript-eslint/no-unnecessary-coercion
  36:2  error  Passing a bigint to BigInt() does not change the type or value of the bigint            @typescript-eslint/no-unnecessary-coercion
```

Does not (and should not) show errors for:

```typescript
// setup
let str = Math.random() > 0.5 ? "string" : 1;
const num = Math.random() > 0.5 ? 1 : "1";
const bool = Math.random() > 0.5 ? true : 1;
const big = Math.random() > 0.5 ? BigInt(1) : 1;

// begin valid cases

String(str);
str.toString();
`${str}`;
"" + str;
str + "";
str += "";

new String("asdf");

Number(num);
+num;
~~num;
~1;

Boolean(bool);
!!bool;
!false;

BigInt(big);
```

Does not do anything with Symbol() because everything weird I tried to do with symbols was already a typescript error.

Does not do anything for some conversions to number which are already forbidden by typescript ("The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type."), including:

```typescript
"123" * 1;
"123" / 1;
"123" - 0;
"123" | 0;
```

Does not cover `${"123"}`, because that overlaps with the `no-unnecessary-template-expression` rule. Need feedback on what to do about that.

I went with the word "coercion" because it was chosen by @Josh-Cena in the original issue but I think it's a poor choice. I think "conversion" or even "type conversion idiom" might be the most precise, because `.toString()` for example doesn't rely on coercion, and while `!!value` and `!value` both coerce the type of `value` to a boolean, it's the property of the idiom `!!` that it can be used for type conversion which puts it under the umbrella of this rule. So I think `no-unnecessary-type-conversion` is a better name for this rule, given the things that I'm currently having it check for.

Catches the following problems in this repository (separate commit):
```
typescript-eslint\packages\eslint-plugin\src\rules\adjacent-overload-signatures.ts
  79:21  error  Using !! on a boolean does not change the type or value of the boolean  @typescript-eslint/no-unnecessary-coercion
  97:21  error  Using !! on a boolean does not change the type or value of the boolean  @typescript-eslint/no-unnecessary-coercion

typescript-eslint\packages\eslint-plugin\src\rules\explicit-function-return-type.ts
  170:27  error  Using !! on a boolean does not change the type or value of the boolean  @typescript-eslint/no-unnecessary-coercion
  177:9   error  Using !! on a boolean does not change the type or value of the boolean  @typescript-eslint/no-unnecessary-coercion

typescript-eslint\packages\eslint-plugin\src\rules\member-ordering.ts
  492:14  error  Using !! on a boolean does not change the type or value of the boolean  @typescript-eslint/no-unnecessary-coercion

typescript-eslint\packages\eslint-plugin\src\rules\no-duplicate-enum-values.ts
  50:21  error  Passing a string to String() does not change the type or value of the string  @typescript-eslint/no-unnecessary-coercion
  52:21  error  Passing a number to Number() does not change the type or value of the number  @typescript-eslint/no-unnecessary-coercion

typescript-eslint\packages\eslint-plugin\src\rules\no-empty-interface.ts
  84:44  error  Using !! on a boolean does not change the type or value of the boolean  @typescript-eslint/no-unnecessary-coercion

typescript-eslint\packages\eslint-plugin\src\rules\no-redundant-type-constituents.ts
  176:10  error  Using !! on a boolean does not change the type or value of the boolean  @typescript-eslint/no-unnecessary-coercion

typescript-eslint\packages\eslint-plugin\src\rules\no-unnecessary-template-expression.ts
   22:24  error  Passing a string to String() does not change the type or value of the string  @typescript-eslint/no-unnecessary-coercion
  174:22  error  Passing a string to String() does not change the type or value of the string  @typescript-eslint/no-unnecessary-coercion

typescript-eslint\packages\eslint-plugin\src\rules\no-unsafe-assignment.ts
  202:17  error  Passing a string to String() does not change the type or value of the string  @typescript-eslint/no-unnecessary-coercion

typescript-eslint\packages\eslint-plugin\src\rules\prefer-return-this-type.ts
  64:14  error  Using !! on a boolean does not change the type or value of the boolean  @typescript-eslint/no-unnecessary-coercion
```

Might not have the best placement for the error scope, and might not have the best formatting for autofixes, could use feedback on those too.


skondrashov:
Sorry for force-push, wasn't sure how else to resolve the merge conflicts other than redoing the commit cause they didn't show up locally. Hopefully they go away now.

kirkwaiblinger:
> ```ts
> // setup
> let str = Math.random() > 0.5 ? "string" : 1;
> const num = Math.random() > 0.5 ? 1 : "1";
> const bool = Math.random() > 0.5 ? true : 1;
> const big = Math.random() > 0.5 ? BigInt(1) : 1;

(hint) - We often use `declare` to do this sort of setup for demo purposes, i.e.
```ts
// setup
declare let str: string | number;
declare const num: number | string;
// etc
```

It's a useful pattern for testing sometimes too, since you may run into surprising behavior when `const t = true;` has type `true` rather than type `boolean`. Nothing specifically requested, just letting you know about another tool for your kit in case it comes in handy during discussion on this PR 🙂 

> Does not do anything with Symbol() because everything weird I tried to do with symbols was already a typescript error.

Makes sense 👍 

> Does not do anything for some conversions to number which are already forbidden by typescript ("The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type."), including:
> 
> ```ts
> "123" * 1;
> "123" / 1;
> "123" - 0;
> "123" | 0;
> ```

Yep, also makes sense. And, even if TS did allow them, we don't need to stress about identifying every possible way for a value to be coerced, just the common patterns, and others can always be added in the future, too. Appreciate your close attention to detail on all these, though!

> Does not cover `${"123"}`, because that overlaps with the `no-unnecessary-template-expression` rule. Need feedback on what to do about that.

This is an astute observation. Is there anything this rule would flag that that rule doesn't flag? If not, I'd lean towards letting that separate rule handle this case. If you want you could even put a note in the docs that tells the user that template expressions intentionally aren't handled because they're already dealt with more thoroughly in the `no-unnecessary-template-expression` rule. 
 
> I went with the word "coercion" because it was chosen by @Josh-Cena in the original issue but I think it's a poor choice. I think "conversion" or even "type conversion idiom" might be the most precise, because `.toString()` for example doesn't rely on coercion, and while `!!value` and `!value` both coerce the type of `value` to a boolean, it's the property of the idiom `!!` that it can be used for type conversion which puts it under the umbrella of this rule. So I think `no-unnecessary-type-conversion` is a better name for this rule, given the things that I'm currently having it check for.

Hmm, I hear you. I personally think either one is fine. Maybe someone else will care more strongly 🤷 

> Might not have the best placement for the error scope, and might not have the best formatting for autofixes, could use feedback on those too.

I think for these reports, reporting the whole node is totally fine
```ts
declare const s: string;
s + '';
~~~~~~

!!(longExpressionThatsABoolean);
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
But, I really like what you've done with the more granular approach.
```ts
!!(longExpressionThatsABoolean);
~~
```
So I'd say go for it! But if it becomes an implementation hassle, no stress either if you end up falling back to the ordinary strategy.

kirkwaiblinger:
> Sorry for force-push...

No worries! The real pain is just force pushing between reviews, not before a reviewer has looked at it at all 🙂.


>..., wasn't sure how else to resolve the merge conflicts other than redoing the commit cause they didn't show up locally. 

Oh, so I think the trouble that you're having is that you've used the `main` branch on your fork. The "usual" vanilla workflow would be

1. fork typescript-eslint. The main branch of your fork points to the HEAD of main of the original typescript-eslint _at the time the fork was made_.
2. Create a branch for your feature, work on there.
3. Whenever you need to synchronize upstream changes, use the github UI or CLI to synchronize the changes on typescript-eslint's main with your fork's main. Then merge those into your branch.

But you're in a slightly awkward scenario since your branch is on main. What I personally do (which differs from the above) is to set a second remote on your local repo, i.e. `git remote add upstream git@github.com:typescript-eslint/typescript-eslint.git`. Then you can `git fetch --all && git checkout upstream/main` (or similar) to make a branch from there, to merge changes into your PR branch. There's... (too) many ways to address this, good ol git, haha. 

LMK if this helps or if you'd like some more help. (I can also push things directly to your branch to bail you out if needed 🤣). And no stress if you end up doing a force push again this time to get yourself in a clean state here!

kirkwaiblinger (changes requested):
This is great work! And especially impressive considering it's your first PR to typescript-eslint. Well done! 

Requesting a few changes, but nothing major.

🙂

packages/eslint-plugin/tests/rules/no-unnecessary-coercion.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    `,
+    `
+!false;
+    `,
+    `
+~256;
+    `,
+  ],
+
+  invalid: [
```
kirkwaiblinger, line 70: Couple general nits/info:
1. No need for the `type` in the `error` report. That asserts that the `node` provided in the `context.report()` is the provided type. However, that's useless here, since `node`'s only effect is to provide a report location if `loc` is not provided. And `loc` _is_ provided. So we're just testing anything useful by that 😆 . 
2. Since you're using custom `loc` for the report locations, it's best to specify the full `line`, `endLine`, `column`, endColumn` for the error. We're less careful about that when we just do the `node`'s location but here it makes sense to be more granular.

packages/eslint-plugin/tests/rules/no-unnecessary-coercion.test.ts
```diff
@@ -0,0 +1,212 @@
...
+const ruleTester = new RuleTester({
+  languageOptions: {
+    parserOptions: {
+      tsconfigRootDir: rootDir,
+      project: './tsconfig.json',
+    },
+  },
+});
+
+const messageId = 'unnecessaryCoercion';
```
kirkwaiblinger, line 18: (nit) Eh, i'd inline this

Josh-Cena:
> I went with the word "coercion" because it was chosen by @Josh-Cena in the original issue but I think it's a poor choice. I think "conversion" or even "type conversion idiom" might be the most precise, because `.toString()` for example doesn't rely on coercion, and while `!!value` and `!value` both coerce the type of `value` to a boolean, it's the property of the idiom `!!` that it can be used for type conversion which puts it under the umbrella of this rule. So I think `no-unnecessary-type-conversion` is a better name for this rule, given the things that I'm currently having it check for.

Agreed; let's go with `no-unnecessary-type-conversion` instead.

packages/eslint-plugin/src/rules/no-unnecessary-coercion.ts
```diff
@@ -0,0 +1,242 @@
...
+              end: node.right.loc.start,
+            },
+            data: {
+              violation: "Concatenating '' with a string",
+              type: 'string',
+            },
+          });
+        }
+      },
+      'UnaryExpression[operator = "+"]'(node: TSESTree.UnaryExpression): void {
```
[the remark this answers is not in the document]
skondrashov: I see what you mean. Not really sure. The error message I landed on makes me feel like I'm ok with still flagging these:

```
Using the unary + operator on a number does not change the type or value of the number
```

I'd assume the reason people would want to verbosely declare positive numbers would commonly have to do with having positive and negative counterparts for a number, eg `+100` and `-100`. Imagine then:

```
const PI = 3.14159
const positive_PI = +PI
const negative_PI = -PI
```

Now they're not using a number literal, but your reasoning still applies, they're just being verbose. Feels like a bit of a slippery slope? I would personally be using my rule (in line with what I generally expect from a linter) to tell me if I'm writing symbols that don't do anything. But it could be hard for a user in that context to see what that has to do with type conversion.

packages/eslint-plugin/src/rules/no-unnecessary-coercion.ts
```diff
@@ -0,0 +1,242 @@
...
+          context.report({
+            node,
+            messageId: 'unnecessaryCoercion',
+            fix: (fixer): RuleFix[] => [
+              fixer.removeRange([node.range[0], node.argument.range[0]]),
+              fixer.removeRange([node.argument.range[1], node.range[1]]),
+            ],
+            loc: {
+              start: node.loc.start,
+              end: node.argument.loc.start,
```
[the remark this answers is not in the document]
skondrashov: This wonderfully addresses the concerns I had with some of the red underlines I was seeing, with a couple of caveats.

1) You can't add numbers to objects, though I figured out the structure of the object so I'm modifying the column field, and assuming the operator ends on the same line it begins (seems reasonable enough).
2) I opted not to treat !! and ~~ as operators of length 2, because nothing stops a developer from writing something like:
```
! 
    !
              false
```
but I was able to get this underline for a similar case, and all parenthesis/whitespace concerns I had I think are now assuaged:
image

See the commit titled "improve loc generation for !!, +, and ~~ operators" for details, keeping in mind that I use `node.parent.loc` for the start and `node.loc` for the end for the `!!` and `~~` operators to achieve the desired effect.

packages/eslint-plugin/tests/rules/no-unnecessary-coercion.test.ts
skondrashov: changed this since it appears to only be done in a minority of test files, but strongly disagree for the record, as magic constants should be easy to change in one place

JoshuaKGoldberg:
👋 Just checking in @skondrashov - is this still something you have time and energy for?

skondrashov:
I'll have time to finish up the changes probably early in the new year!

JoshuaKGoldberg:
Awesome! I'll move this to draft then in the meantime, so we don't bug you. 🚀

packages/eslint-plugin/src/rules/no-unnecessary-coercion.ts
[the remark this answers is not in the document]
skondrashov: 1. the rule already has `type: 'suggestion'` (unless I did that in response to this comment and forgot), and it doesn't seem like the API encourages me to move the fix into the `suggest` array
2. in implementing this, it seemed this only makes sense if the thing being type converted is an identifier, so I implemented it only for that case. eg `'asdf' satisfies string` seems besides the point

packages/eslint-plugin/src/rules/no-unnecessary-coercion.ts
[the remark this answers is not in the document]
skondrashov: there's one case that sucks still: `!(!false)` autofix removes the left parenthesis

packages/eslint-plugin/src/rules/no-unnecessary-coercion.ts
```diff
@@ -0,0 +1,242 @@
...
+        if (
+          node.callee.type === AST_NODE_TYPES.Identifier &&
+          node.arguments.length === 1
+        ) {
+          const type = getConstrainedTypeAtLocation(
+            services,
+            node.arguments[0],
+          );
+
+          if (
```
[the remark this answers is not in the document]
skondrashov: fixed for `String`/`Boolean` etc, still needs a fix for the case where `.toString()` is overwritten

skondrashov:
I addressed all of @kirkwaiblinger's feedback I think. Poor commit hygiene on a couple of the things but mostly separated into atomic commits. There's still 2 things I need to do:

1. The case where .toString() is overwritten should not trigger the rule, in the vein of the other scope analysis comment.
2. I'd like to have different autofixes for the case where `str += ''` is used as a statement vs as an expression. Probably not that important but not that difficult either.

Might need help with the first, will implement the second tomorrow, but essentially ready for review again.

packages/eslint-plugin/src/rules/no-unnecessary-coercion.ts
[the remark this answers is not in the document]
kirkwaiblinger: If we went forward with this, I would certainly draw a hard line at single `+` applied to numeric literal... So, still reporting on the outer `+` of `+(+3)`/`+ +3`, for example. IMO the slope isn't really slippery since something like `+PI` is ambiguous from static analysis as to whether it's a positive or negative number and instead looks more like a mistaken attempt at `Math.abs(PI)` than anything. That said, something like `1-+2` is a big yikes that would be permitted by the naive implementation I have in my head, so... 

i'm fine with leaving it as is 👍. I do agree that your error message wording (and squiggly placement!) makes it totally understandable _why_ the rule is reporting, and it's easy enough for someone to create a feature request for it in the future if there's a want.

Thanks for the discussion!

packages/eslint-plugin/src/rules/no-unnecessary-coercion.ts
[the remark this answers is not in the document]
kirkwaiblinger: > 1. the rule already has `type: 'suggestion'` (unless I did that in response to this comment and forgot)...

This is actually something entirely separate with an unfortunate naming clash, see [the eslint core docs](https://eslint.org/docs/latest/extend/custom-rules#rule-structure). It doesn't affect whether fixes are applied as "fixes" or "suggestion fixes". 

> ...and it doesn't seem like the API encourages me to move the fix into the `suggest` array

Could you clarify what you mean by this? I don't follow. But... yes, we do put suggestion fixers in the report function calls. For a random example, see

https://github.com/typescript-eslint/typescript-eslint/blob/b17c7f25af620c22ab9a27c270c964a6fd29de7c/packages/eslint-plugin/src/rules/no-floating-promises.ts#L192-L204

> 2. in implementing this, it seemed this only makes sense if the thing being type converted is an identifier, so I implemented it only for that case. eg `'asdf' satisfies string` seems besides the point

Especially since this is posited as a suggestion fixer that the user would manually apply, I think we are better off not trying to be clever here. The user can simply disregard the suggestion when it doesn't make sense. 

Identifier is definitely too restrictive since member expressions like `a.b.c` are equally sensible... And if we support those, we need to consider chain expressions `a?.b.c`.

Furthermore, In principle, code like the following might be useful sometimes
```ts
declare function f(): string;

f() satisfies string + "some other string";
```
That way if someone changes `f()` to return a `number` (maybe declared in a whole nother file), they're notified of the issue, instead of getting something like `'42some other string'`. 

In conclusion, while it's true that literals specifically are a little silly to suggest a `satisfies` expression, I think any other syntax is totally fair game, so let's just allow any syntax to trigger this suggestion.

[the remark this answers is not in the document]
kirkwaiblinger: Looking a bit more at the implementation, I'm assuming you're referring to the fact that, this code is rather awkward:

https://github.com/typescript-eslint/typescript-eslint/blob/606556e809b98decfa190a8ed3890876a5293725/packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts#L389-L419

If it helps, feel free to study our util `getFixOrSuggest` and use and/or modify it to be helpful to this use case

https://github.com/typescript-eslint/typescript-eslint/blob/5b2ebcdea9011a6de7e657118874ae85c08d5a0e/packages/eslint-plugin/src/util/getFixOrSuggest.ts#L3-L21

But, i think it's moot since we probably want them all to be suggestions anyway rather than ever applying an autofix

packages/eslint-plugin/tests/rules/no-unnecessary-coercion.test.ts
kirkwaiblinger: > changed this since it appears to only be done in a minority of test files, but strongly disagree for the record, as magic constants should be easy to change in one place

Thanks 🙏. It's true about the magic constants, _but_ this is sort of an idiomatic approach in TS (for better or worse) since the literal types enforce correctness. In other words - this forces you to type the strings out correctly:

https://github.com/typescript-eslint/typescript-eslint/blob/606556e809b98decfa190a8ed3890876a5293725/packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts#L19-L21

kirkwaiblinger (commented):
Responded to some comments for now, but don't have time to finish a thorough review at the moment, will come back to this. Once again thanks for your detailed work and discussions!

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
```diff
@@ -0,0 +1,461 @@
...
+
+    const shouldHaveSatisfiesSuggestion = (type: AST_NODE_TYPES) =>
+      [AST_NODE_TYPES.Identifier, AST_NODE_TYPES.MemberExpression].includes(
+        type,
+      );
+
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    const surroundWithParentheses = (
```
kirkwaiblinger, line 80: I'm wondering if this should instead use `getWrappingFixer` or one of the related functions? Or does that not work here?

https://github.com/typescript-eslint/typescript-eslint/blob/3f02687d964c1e7888e5db62256d56cd72c4f281/packages/eslint-plugin/src/util/getWrappingFixer.ts#L29-L78

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
```diff
@@ -0,0 +1,461 @@
...
+      const matchesType = (t: ts.Type): boolean => {
+        return isTypeFlagSet(t, typeFlag);
+      };
+
+      if (type.isUnion()) {
+        return type.types.every(matchesType);
+      }
+
+      if (type.isIntersection()) {
+        return type.types.some(matchesType);
```
kirkwaiblinger, line 57: Could you pls ensure that this and other lines lacking coverage are hit by test cases? 🙏

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
```diff
@@ -0,0 +1,461 @@
...
+          messageId: 'unnecessaryTypeConversion',
+          data: reportDescriptorMessageData,
+          fix: fixFunction,
+          suggest: shouldHaveSatisfiesSuggestion(node.argument.type)
+            ? [
+                {
+                  messageId: 'unnecessaryTypeConversionSuggestion',
+                  data: { type: reportDescriptorMessageData.type },
+                  fix(fixer): RuleFix[] {
+                    return [
```
kirkwaiblinger, line 130: I'm guessing this is being reported as missing coverage because the `fix` takes precedence if present. Let's ensure that we don't provide both (instead giving both options as suggestions)

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
skondrashov: For the `satisfies` suggestions, this seems to work really well. For the regular fix, it works, but it's weird cause you have to pass a no-op as the `wrap` parameter, and it does some redundant logic internally, and puts some parentheses in places they don't need to be, in some cases double sets of parentheses.

The code is a whole lot cleaner using `getWrappingFixer` for everything, but it's strange for the case where nothing is actually wrapping the code. I'm tempted to modify `getWrappingFixer` to make the `wrap` parameter optional, but it's kind of weird because then it's not really a "wrapping" fixer at all, more an unwrapping fixer. Would like your thoughts, but I've pushed up the code that uses `getWrappingFixer` for everything as well as a very rudimentary change to the API that makes `wrap` optional.

packages/eslint-plugin/src/rules/no-unnecessary-coercion.ts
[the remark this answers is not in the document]
skondrashov: The awkwardness in the API that I see is that you're forced to provide `messageId` and `data` even if there's no fix to be applied, so you have to duplicate those fields in the suggestion fix.

So, you get something like:
```typescript
context.report({
  node,
  messageId,
  data,
  suggest: [
    {
      messageId,
      data,
      fix: someFixer,
    },
  ],
});
```

rather than the cleaner:

```typescript
context.report({
  node,
  messageId,
  data,
  fix someFixer,
});
```

To be honest, I don't understand the motivation for this, though the only thing stopping me otherwise is how verbose it is. Is the point that the fix can be dangerous when working with types that are inaccurate, so it's reckless to autofix it?

skondrashov:
After latest round of fixes, I believe there are 3 things I still need to do:
1. Do scope analysis to see if `.toString()` has been overwritten before applying the rule in cases involving it.
2. Move autofixes to suggestion fixes. I would like to understand the motivation for this change, but I just didn't have time to get to it today.
3. Depending on feedback, do something with the fixes that use `getWrappingFixer` but don't need to pass a `wrap` parameter.

I left some comments in the appropriate threads. @kirkwaiblinger do you mind resolving the threads that have been resolved to your satisfaction? I don't want to do it myself in case I'm wrong in resolving something.

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
```diff
@@ -0,0 +1,369 @@
...
+                    ...wrappingFixerParams,
+                    wrap: expr => `${expr} satisfies ${typeString}`,
+                  }),
+                },
+              ],
+            });
+          }
+        }
+      },
+      'CallExpression > MemberExpression.callee > Identifier[name = "toString"].property'(
```
kirkwaiblinger, line 294: Today I learned that you could use ESQuery with the `.callee` like this. Cool!

Question, though, does the `.property` at the end do anything? I can't think what it would do and no tests fail if I change it to 

```suggestion
      'CallExpression > MemberExpression.callee > Identifier[name = "toString"]'(
```

kirkwaiblinger: To make sure the selector is correct could you also add the following (valid) test case?
```ts
export {};
declare const toString: string;
toString.toUpperCase();
```

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
kirkwaiblinger: Would you mind providing a few examples? Would save me quite a bit of time 🙏 In general, we're not too stressed about some extra parens, which most users are likely to fix with a formatter anyway, unlike too few parens, which will actually break people's code.

kirkwaiblinger:
> After latest round of fixes, I believe there are 3 things I still need to do:
> 
> 1. Do scope analysis to see if `.toString()` has been overwritten before applying the rule in cases involving it.

Let's not do this, actually. It's complex to detect, a rare situation, the only situation in which it would affect correctness the rule's report is if someone overrides `String.prototype.toString` in a way that `"foo".toString()` does not return `"foo"` - not a situation we're interested in accommodating 🙃

> 2. Move autofixes to suggestion fixes. I would like to understand the motivation for this change, but I just didn't have time to get to it today.

Sounds good 👍 On the motivation, for me:
- Removing conversions to primitives is a pretty meaningful code change and can cause breakage if the TS types are not accurate at runtime. Can unpack this more, but I'd rather be cautious and demand human intervention rather than run an autofix, which many users have configured to fix on save.
- We are providing two different approaches (with and without `satisfies`) for the fix, and the only way to express multiple fix options is with two suggestions, rather than an autofix (or providing a rule option to configure the fix - [example](https://typescript-eslint.io/rules/consistent-type-imports/#fixstyle) - like but that's totally overkill here)

> 3. Depending on feedback, do something with the fixes that use `getWrappingFixer` but don't need to pass a `wrap` parameter.

responded asking for more detail 👍 

> I left some comments in the appropriate threads. @kirkwaiblinger do you mind resolving the threads that have been resolved to your satisfaction? I don't want to do it myself in case I'm wrong in resolving something.

Will do! 👍

packages/eslint-plugin/tests/rules/no-unnecessary-type-conversion.test.ts
```diff
@@ -0,0 +1,467 @@
...
+  languageOptions: {
+    parserOptions: {
+      project: './tsconfig.json',
+      tsconfigRootDir: rootDir,
+    },
+  },
+});
+
+ruleTester.run('no-unnecessary-type-conversion', rule, {
+  valid: [
```
kirkwaiblinger, line 18: (testing)
Could you please add some valid cases around ensuring that only _primitives_ cause the rule to flag? In particular, I'm thinking about the boxed primitives, so, things like
```ts
String(new String('foo'));

function f(x: Number) {
    return +x;
}
```
etc

Also it would be good to add some tests around stringy and numbery `enum`s. 

Also let's add some coverage around generics, i.e.
```ts
function f<T extends string>(x: T) {
    return String(x);
}
```

kirkwaiblinger (changes requested):
Another good iteration! Looking forward to the next round! 🙂

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
```diff
@@ -0,0 +1,369 @@
...
+      CallExpression(node: TSESTree.CallExpression): void {
+        if (
+          node.callee.type === AST_NODE_TYPES.Identifier &&
+          node.arguments.length === 1
+        ) {
+          const getType = () =>
+            getConstrainedTypeAtLocation(services, node.arguments[0]);
+
+          const isBuiltInCall = (name: string) => {
+            if ((node.callee as TSESTree.Identifier).name === name) {
```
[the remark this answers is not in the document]
kirkwaiblinger: Fun fact, I think it's a bug in TS that you need to alias `nodeCallee` to preserve the narrowing in the closure, see https://github.com/microsoft/TypeScript/issues/61158

jdufresne:
I'd like to throw out another potential case of an unnecessary type converstion: `Object.values(array)`. This is a construct that I have found in my own projects. Once found, I simplify it to be the array (or a copy of the array). If you think this is better suited for an enhancement after this version lands, I can create a new ticket to document it.

```ts
// Bad
Object.values(['a', 'b', 'c'])  // simplify to: ['a', 'b', 'c']

const array = ['a', 'b', 'c']
Object.values(array)  // simplify to: array


// Good
type MyObject = { a: string }

let arrayOrObject: array | MyObject
arrayOrObject = ['a', 'b', 'c']
Object.values(arrayOrObject)  // could be an object, can't simplify
```

kirkwaiblinger:
@jdufresne fun suggestion! Yeah, let's scope this as an enhancement rather than part of the initial implementation 👍

jdufresne:
> @jdufresne fun suggestion! Yeah, let's scope this as an enhancement rather than part of the initial implementation 👍

Great. Thanks. Done as https://github.com/typescript-eslint/typescript-eslint/issues/10941

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
skondrashov: That test fails without the `.property`, which I guess is what you had in mind? Adding the test, leaving the selector for that reason.

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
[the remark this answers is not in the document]
skondrashov: Is there an advantage of using `isNodeOfType` over `nodeCallee.type === AST_NODE_TYPES.Identifier`? Aliasing `nodeCallee` seemed to solve the problem on its own.

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
```diff
@@ -0,0 +1,369 @@
...
+      isDoubleOperator: boolean, // !! or ~~
+    ) {
+      const type = services.getTypeAtLocation(node.argument);
+      if (doesUnderlyingTypeMatchFlag(type, typeFlag)) {
+        const wrappingFixerParams = {
+          node: isDoubleOperator ? node.parent : node,
+          innerNode: [node.argument],
+          sourceCode: context.sourceCode,
+        };
+        const typeString =
```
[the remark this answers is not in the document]
skondrashov: lovely, did this almost exactly but with `typeString: string,` changed to `typeString: 'boolean' | 'number',` and the order of parameters changed so that `isDoubleOperator` is last cause it feels like it should be. I feel like you probably don't care about the order of the parameters, but let me know if you think changing the type is a bad move.

skondrashov:
@kirkwaiblinger I think I'm ready for another look.

I left some replies, but I thought I'd reply to the `getWrappingFixer` questions here. I can only think of one case where there are extra parentheses added by the current code: `'a' + 'b' + '';` -> `('a' + 'b');`. Which is annoying but harmless.

I said before in one of my comments that there are cases with double parentheses, but I couldn't find any now. Not sure if I was wrong before, but everything else I tried this time made sense. Nothing with too few parentheses and just one with too many. I found some sets of patterns that might be enlightening to play with anyway -

These patterns (and similar for other types) strip all parentheses, no matter how many you put:

```typescript
String(((('a'))));
((('a'))).toString();
((('a'))) + '';
```

If the thing being operated on isn't unary, the output gets a single set of parentheses, again regardless of how many there are. The last case here is the one that feels like it's adding extra parentheses:

```typescript
String('a' + 'b');
('a' + 'b').toString();
'a' + 'b' + '';
```

Using the `satisfies` fix this is enlightening because the fix to use `satisfies` instead of the `String` constructor places 2 pairs of parentheses in the proper places (`(('a' + 'b') satisfies string).toString()`). The other shows that everything somehow works fine with weird double operators and whitespace:

```typescript
String('a' + 'b').toString();
( ~ ( ~ ( 5 ) ) );
```

I think my main concern is whether what I did to the code in `getWrappingFixer` is acceptable, and what I can do to make it better if not.

Still to do:

1. Depending on feedback, do something with the fixes that use getWrappingFixer but don't need to pass a wrap parameter.
1. Figure out what to do about the incorrect fixes for enums where all the fields are the same primitive type.

packages/eslint-plugin/tests/rules/no-unnecessary-type-conversion.test.ts
kirkwaiblinger: Hmm. Yeah the `Foo` vs `string` thing is a bit of an annoying judgement call... let's keep the logic as-is (with `String(foo)` flagging and suggesting `foo`).

kirkwaiblinger (approved):
I think I'm happy with this! Thanks for your very detailed work on this. Outstanding first PR!

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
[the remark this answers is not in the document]
kirkwaiblinger: Nope! checking the `.type` explicitly is totally fine too. It tends to be more useful to use the helper when checking a list of `.type`s.

packages/eslint-plugin/docs/rules/no-unnecessary-type-conversion.mdx
```diff
@@ -11,7 +11,7 @@ import TabItem from '@theme/TabItem';
 
 JavaScript has several idioms used to convert values to certain types. With TypeScript, it is possible to see at build time if the value is already of that type, making the conversion unnecessary.
 Performing unnecessary conversions increases visual clutter and harms code readability, so it's generally best practice to remove them if they don't change the type of an expression.
-This rule reports when a type conversion idiom is identified which does not change the type of an expression.
+This rule reports when a type conversion idiom is identified which does not change the runtime type of an expression.
```
skondrashov, line 14: It's about the buildtime type of the expression I think more than the runtime type. The runtime type isn't necessarily the same going in vs out of the expression, and if it is it's most likely because it matches the buildtime types, which is what we are scanning.

kirkwaiblinger: Hmm, I guess it depends how you look at it. I was thinking about an example like this
```ts
function f(x: 1 | 2) {
   const n = x;
//       ^? const n: 1 | 2
}

function f(x: 1 | 2) {
   const n = Number(x);
//       ^? const n: number
}
```

The static TS type has changed as a result of the action of the `Number()` constructor (widened to `number`), but the _runtime_ types they imply are the same; both `1 | 2` and `number` imply `typeof n === 'number'`. Note for context that our docs generally assume the TS types are correct unless otherwise noted (as in the When Not To Use It section).

***

That all said, I think it's tricky enough to get the wording right that I think I'll revert it to "type" rather than make it worse by trying to be more precise 🙃

kirkwaiblinger (dismissed):

packages/eslint-plugin/docs/rules/no-unnecessary-type-conversion.mdx
skondrashov: Yeah I don't remember cause it was a while ago, but I have a feeling I intentionally avoided saying build- or runtime for something related to all of these reasons. Thank you!

skondrashov:
@kirkwaiblinger Sorry to bother, but I think resetting your last review also reset the review request for triage-team.

JoshuaKGoldberg:
I am _salivating_ at how many existing violations were removed in this PR. 🤤

packages/eslint-plugin/docs/rules/no-unnecessary-type-conversion.mdx
```diff
@@ -0,0 +1,74 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-unnecessary-type-conversion** for documentation.
+
+JavaScript has several idioms used to convert values to certain types. With TypeScript, it is possible to see at build time if the value is already of that type, making the conversion unnecessary.
```
JoshuaKGoldberg, line 12: [Docs] I think we'll want to explicitly call out what the things are in plain text, before getting to the explanation of why we don't want them -> examples of them.

```suggestion
JavaScript provides several commonly used idioms to convert values to a specific types:

* Primitive coercion (e.g. `Boolean(value)`, `String(value)`): using a built-in primitive function
* String concatenation (e.g. `value + ""`): turning a value into a string
* Unary coercion (e.g. `+value`, `!!value`): using a built-in operator

These conversions are unnecessary if the value is already that type.
```

packages/eslint-plugin/src/rules/no-unnecessary-type-conversion.ts
```diff
@@ -0,0 +1,376 @@
...
+              doesUnderlyingTypeMatchFlag(
+                getTypeLazy(),
+                ts.TypeFlags.BooleanLike,
+              )) ||
+            (isBuiltInCall('BigInt') &&
+              doesUnderlyingTypeMatchFlag(
+                getTypeLazy(),
+                ts.TypeFlags.BigIntLike,
+              ))
+          ) {
```
JoshuaKGoldberg, line 271: [Refactor] If I'm understanding, this only needs to check the type against a built-in call at most once, right? When the name === matches?

If so: we can get rid of a good amount of the manual `... && ... && ...` checks by using a direct mapping from the name to type flags. Roughly:

```suggestion
        const nodeCallee = node.callee;
        const builtInTypeFlags = {
          BigInt: ts.TypeFlags.BigIntLike,
          Boolean: ts.TypeFlags.BooleanLike,
          Number: ts.TypeFlags.NumberLike,
          String: ts.TypeFlags.StringLike,
        };

        if (
          nodeCallee.type !== AST_NODE_TYPES.Identifier ||
          node.arguments.length !== 1 ||
          !(nodeCallee.name in builtInTypeFlags)
        ) {
          return;
        }

        const typeFlag =
          builtInTypeFlags[nodeCallee.name as keyof typeof builtInTypeFlags];
        const scope = context.sourceCode.getScope(node);
        const variable = scope.set.get(nodeCallee.name);
        if (
          !!variable?.defs.length ||
          !doesUnderlyingTypeMatchFlag(
            getConstrainedTypeAtLocation(services, node.arguments[0]),
            typeFlag,
          )
        ) {
          return;
        }
```

Now it doesn't need to keep checking `isBuiltInCall('...')` up to 4x.

A nice little bonus: even with the extra lines for early `{ return; }`s, this cuts down the total file line count from 376 to 363. 😎

JoshuaKGoldberg (approved):
Awesome, this is looking very good! A stellar first PR indeed 🏆. Thanks for pushing on this!

Just a few small things from me - nothing we can't just do before merge if you don't have the time.

kirkwaiblinger (approved):

PR #10193, opened 2024-10-22
chore: switch from explicit-function-return-type to explicit-module-boundary-types internally

### LINKED ISSUE #9419: Repo: Switch from explicit-function-return-type to isolatedDeclarations and/or explicit-module-boundary-type
### Suggestion

Forking out of https://github.com/typescript-eslint/typescript-eslint/pull/8173#discussion_r1649974627:

> > @joshuakgoldberg: ... we keep the [@typescript-eslint/explicit-function-return-type](https://typescript-eslint.io/rules/explicit-function-return-type) rule enabled.
> 
> @bradzacher: We could probably relax that.
> I'd be okay with us doing isolatedDeclarations instead of EFRT.

I'd be in favor of moving to a lighter-weight enforcement of function return types internally. My understanding of the context is that we've traditionally enforced function return types for these reasons:

* To be explicit in types as a stylistic point - holding over from when this was much more common, years ago
* Type performance: since we're a non-trivial sized repo, it saves time in TypeScript type checking
* Type stability: i.e. what [TypeScript 5.5's `isolatedDeclarations`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#isolated-declarations) enforces

Now that `isolatedDeclarations` are stable, it seems reasonable to me that we'd switch our internal linting to only enforcing what it requires. We'd get the type performance and stability, without the more heavy-handed stylistic requirements. Our non-exported functions would now be able to have inferred return types! What does the team think?

> Note that this is only for the _internal linting of this repository_. This issue doesn't impact end-users of typescript-eslint.

💖

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9419

## Overview

Purely switches the ESLint config and some necessary inline comments. Doesn't attempt to remove existing annotations.

💖

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -125,10 +125,7 @@ export default tseslint.config(
         'error',
         { prefer: 'type-imports', disallowTypeAnnotations: true },
       ],
-      '@typescript-eslint/explicit-function-return-type': [
-        'error',
-        { allowIIFEs: true },
-      ],
+      '@typescript-eslint/explicit-module-boundary-types': 'error',
       '@typescript-eslint/no-explicit-any': 'error',
       'no-constant-condition': 'off',
       '@typescript-eslint/no-unnecessary-condition': [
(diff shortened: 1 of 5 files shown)


kirkwaiblinger (approved):
😢 😢

PR #10194, opened 2024-10-22
fix(eslint-plugin): [no-unnecessary-condition] improve error message for literal comparisons

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9643 
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

In addition to changing the wording, I've also had the error message compute whether the condition is actually true or false.

PR #10205, opened 2024-10-24
fix(eslint-plugin): [no-unnecessary-condition] falsey bigint should be falsey

### LINKED ISSUE #10197: Bug: [no-unnecessary-condition] should not say that falsey bigint is truthy

### Playground Link

https://typescript-eslint.io/play/#ts=5.6.3&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3gMyhNIAngEICWA5gJIoYBc8ADCgNwBQp28AFLvkWVRoBKeAG9WAXyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK%2BSpPRQA7iKaRwYAL4hNQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
declare const falseyBigInt: 0n;
if (falseyBigInt) {
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-unnecessary-condition": "warn"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

Unnecessary conditional, value is always falsy.

### Actual Result

Unnecessary conditional, value is always truthy.

### Additional Info

This is a bug with this line. 

https://github.com/typescript-eslint/typescript-eslint/blob/373c63a40ccb6c5306999a0bd4cdb9cd5348988f/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts#L32

The `.value` field returns a `PseudoBigInt` _object_ that looks like `{ negative?: boolean, base10Value: string }`, rather than a `bigint` itself (for legacy TS support reasons). So checking for truthiness is incorrect; we need to convert to a real bigint.

peanutenthusiast:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10197

## Overview

This adds a utility function to check for a falseyBigInt. I implemented functions to convert `type.value` to BigInt when necessary, yet this still required the extra condition in `checkNode` for the specific edge case.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -39,7 +39,7 @@ const getValue = (type: ts.LiteralType): bigint | number | string => {
   return type.value;
 };
 
-const isFalseyBigInt = (type: ts.Type): boolean => {
+const isFalsyBigInt = (type: ts.Type): boolean => {
   return (
     tsutils.isLiteralType(type) &&
     valueIsPseudoBigInt(type.value) &&
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -24,12 +24,25 @@ import {
   findTypeGuardAssertedArgument,
 } from '../util/assertionFunctionUtils';
 
+const valueIsPseudoBigInt = (
+  value: number | string | ts.PseudoBigInt,
+): value is ts.PseudoBigInt => {
+  return typeof value === 'object' && 'base10Value' in value;
+};
+
+const getValue = (type: ts.LiteralType): bigint | number | string => {
+  if (valueIsPseudoBigInt(type.value)) {
+    return BigInt(type.value.base10Value);
+  }
+  return type.value;
+};
+
 // Truthiness utilities
 // #region
 const isTruthyLiteral = (type: ts.Type): boolean =>
   tsutils.isTrueLiteralType(type) ||
   //  || type.
-  (type.isLiteral() && !!type.value);
+  (type.isLiteral() && !!getValue(type));
 
 const isPossiblyFalsy = (type: ts.Type): boolean =>
   tsutils
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -29,7 +29,7 @@ import {
 const valueIsPseudoBigInt = (
   value: number | string | ts.PseudoBigInt,
 ): value is ts.PseudoBigInt => {
-  return typeof value === 'object' && 'base10Value' in value;
+  return typeof value === 'object'
 };
 
 const getValue = (type: ts.LiteralType): bigint | number | string => {
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -29,7 +29,7 @@ import {
 const valueIsPseudoBigInt = (
   value: number | string | ts.PseudoBigInt,
 ): value is ts.PseudoBigInt => {
-  return typeof value === 'object'
+  return typeof value === 'object';
 };
 
 const getValue = (type: ts.LiteralType): bigint | number | string => {
@@ -69,7 +69,13 @@ const isPossiblyTruthy = (type: ts.Type): boolean =>
     .some(intersectionParts =>
       // It is possible to define intersections that are always falsy,
       // like `"" & { __brand: string }`.
-      intersectionParts.every(type => !tsutils.isFalsyType(type)),
+      intersectionParts.every(
+        type =>
+          !tsutils.isFalsyType(type) &&
+          // below is a workaround for ts-api-utils bug
+          // see https://github.com/JoshuaKGoldberg/ts-api-utils/issues/544
+          !isFalseyBigInt(type),
+      ),
     );
 
 // Nullish utilities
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -26,10 +26,30 @@ import {
 
 // Truthiness utilities
 // #region
+const valueIsPseudoBigInt = (
+  value: number | string | ts.PseudoBigInt,
+): value is ts.PseudoBigInt => {
+  return typeof value === 'object' && 'base10Value' in value;
```
kirkwaiblinger, line 32: Granted the input is `string | number | ts.PseudoBigInt`, checking anything but the `typeof` is unnecessary to discriminate between them
```suggestion
  return typeof value === 'object'
```

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -26,10 +26,30 @@ import {
...
+};
+
+const getValue = (type: ts.LiteralType): bigint | number | string => {
+  if (valueIsPseudoBigInt(type.value)) {
+    return BigInt(type.value.base10Value);
```
kirkwaiblinger, line 37: I don't think it will make a difference in correctness for this PR but probably best to be precise about negative numbers. 
```suggestion
    return BigInt((type.value.negative ? '-' : '') + type.value.base10Value);
```
See also https://github.com/typescript-eslint/typescript-eslint/pull/10194/files#diff-3cc43b506517aba406ae3a31d7998b116c4b09db76b632c6cb3c2e54fc71d66cR72-R93 (not merged yet) that we'll probably want to align on

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -306,6 +326,8 @@ export default createRule<Options, MessageId>({
         messageId = !isUnaryNotArgument ? 'alwaysFalsy' : 'alwaysTruthy';
       } else if (!isPossiblyFalsy(type)) {
         messageId = !isUnaryNotArgument ? 'alwaysTruthy' : 'alwaysFalsy';
+      } else if (isFalseyBigInt(type)) {
```
kirkwaiblinger, line 329: oh, fun. This line "_shouldn't_" be necessary, since for `0n`, `!isPossiblyTruthy(type)` _should_ be `true`, since ts-api-utils's `isFalsyType(type)` _should_ be `true`. But it's not, since ts-api-utils has the exact same bug that was in this rule!

https://github.com/JoshuaKGoldberg/ts-api-utils/blob/41cd21d41f445448681dd32308a48a004f4717cb/src/types/utilities.ts#L50

Filed https://github.com/JoshuaKGoldberg/ts-api-utils/issues/544 to get that corrected. I'd say - let's move this workaround to the `isPossiblyTruthy` function instead just after the `isFalsyType(type)` check, and put a note that this is a workaround for the ts-api-utils bug

kirkwaiblinger (changes requested):
Good stuff, thanks for sending this in!

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
peanutenthusiast: [Lines 75-77](https://github.com/typescript-eslint/typescript-eslint/pull/10205/files#diff-3cc43b506517aba406ae3a31d7998b116c4b09db76b632c6cb3c2e54fc71d66cR75-R77) implement the suggested change and note.

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
kirkwaiblinger, line 1: Can you please also add a test with a truthy bigint to ensure that we don't just report all bigints as falsy?

kirkwaiblinger (changes requested):

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -26,10 +26,30 @@ import {
...
+    tsutils.isLiteralType(type) &&
+    valueIsPseudoBigInt(type.value) &&
+    !getValue(type)
+  );
+};
 const isTruthyLiteral = (type: ts.Type): boolean =>
   tsutils.isTrueLiteralType(type) ||
   //  || type.
-  (type.isLiteral() && !!type.value);
+  (type.isLiteral() && !!getValue(type));
```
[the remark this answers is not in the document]
peanutenthusiast: Alright, I've added several test, including the above (valid?), and others for positive, negative bigints. Keep me posted, should we consider any other potential test cases.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -26,10 +26,30 @@ import {
...
+};
+
+const getValue = (type: ts.LiteralType): bigint | number | string => {
+  if (valueIsPseudoBigInt(type.value)) {
+    return BigInt((type.value.negative ? '-' : '') + type.value.base10Value);
+  }
+  return type.value;
+};
+
+const isFalseyBigInt = (type: ts.Type): boolean => {
```
Josh-Cena, line 42:
```suggestion
const isFalsyBigInt = (type: ts.Type): boolean => {
```

:)

kirkwaiblinger (approved):

kirkwaiblinger:
Looks great to me! Thank you for doing this, and handling some pretty surprising edge cases!

gif

Josh-Cena (approved):

Issue #10208, opened 2024-10-25
Enhancement: [await-thenable] should prohibit using a sync disposable with `await using`


kirkwaiblinger:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### My proposal is suitable for this project

- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Link to the rule's documentation

https://typescript-eslint.io/rules/await-thenable/, https://typescript-eslint.io/rules/no-floating-promises, https://typescript-eslint.io/rules/no-misused-promises

### Description

I would like to find a place to lint for providing a sync disposable to an `await using` statement. This relates closely to https://github.com/typescript-eslint/typescript-eslint/issues/8858, so maybe await-thenable is the right place for this check? But it also has concerns very adjacent to no-floating-promises and no-misused-promises, so I'm not sure.

### Fail

```TypeScript
async function unhandledRejection() {             
    await using _ = {
        async [Symbol.dispose]() {
            throw new Error('dispose error');
        }
    }
}
```

### Pass

```TypeScript
async function properlyHandledRejection() {             
    await using _ = {
        async [Symbol.asyncDispose]() {
            throw new Error('dispose error');
        }
    }
}
```

### Additional Info

The promise rejection in `unhandledRejection` cannot be caught, since the `[Symbol.dispose]()` method is called, but _is not awaited_, by an `await using` statement. Only the `[Symbol.asyncDispose]()` method is awaited. 

There's a lot of subtlety here so I've provided a repo to be able to play with variations on this at https://github.com/kirkwaiblinger/typescript-eslint-repro-await-using

kirkwaiblinger:
Note that the fundamental concern here is that a `Promise`-returning `[Symbol.dispose]()` method is a clear vector for unhandled rejections, whether invoked by an `await using` or an ordinary `using` statement.

bradzacher:
probably makes most sense for this to be in `await-thenable`, I think.

[...]

PR #10209, opened 2024-10-26
feat(eslint-plugin): [await-thenable] report unnecessary `await using` statements

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes part of #10208 

## Overview

Basically just checks whether the RHS of an `await using` statement _might_ have a `[Symbol.asyncDispose]()`, otherwise reports. 

Added in docs and a couple minor refactors. Also modified the `getFixOrSuggest()` method to allow providing _no_ fix or suggestion.

### DIFF
--- a/packages/eslint-plugin/src/rules/await-thenable.ts
+++ b/packages/eslint-plugin/src/rules/await-thenable.ts
@@ -127,7 +127,7 @@ export default createRule<[], MessageId>({
           );
           const type = services.getTypeAtLocation(init);
           if (isTypeAnyType(type)) {
-            return;
+            continue;
           }
 
           const hasAsyncDisposeSymbol = tsutils
--- a/packages/eslint-plugin/src/rules/await-thenable.ts
+++ b/packages/eslint-plugin/src/rules/await-thenable.ts
@@ -121,10 +121,10 @@ export default createRule<[], MessageId>({
         node: TSESTree.VariableDeclaration,
       ): void {
         for (const declarator of node.declarations) {
-          const init = nullThrows(
-            declarator.init,
-            'Expected init to be present on an await using variable declarator',
-          );
+          const init = declarator.init;
+          if (init == null) {
+            continue;
+          }
           const type = services.getTypeAtLocation(init);
           if (isTypeAnyType(type)) {
             continue;
--- a/packages/eslint-plugin/src/rules/await-thenable.ts
+++ b/packages/eslint-plugin/src/rules/await-thenable.ts
@@ -4,6 +4,7 @@ import * as tsutils from 'ts-api-utils';
 
 import {
   createRule,
+  getFixOrSuggest,
   getParserServices,
   isAwaitKeyword,
   isTypeAnyType,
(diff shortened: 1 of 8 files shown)


packages/eslint-plugin/tests/rules/await-thenable.test.ts
```diff
@@ -475,5 +518,99 @@ for  (const value of yieldNumberPromises()) {
...
+  };
+}
+      `,
+            },
+          ],
+        },
+      ],
+    },
+    {
+      code: `
```
kirkwaiblinger, line 580: Honestly I'm kind of torn on this one. In principle, the user could split it up to be 
```ts
  using a = disposable,
  await using b = asyncDisposable,
  using c = disposable,
  await using d = asyncDisposable,
  using e = disposable;
```

But it doesn't really have any benefit (unlike how the `for await` case can be actively harmful to use `for await` unnecessarily). So we could change the logic to make this a valid case unless _none_ of the declarators are possibly-async-iterable 🤷

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -112,6 +116,57 @@ export default createRule<[], MessageId>({
           });
         }
       },
+
+      'VariableDeclaration[kind="await using"]'(
+        node: TSESTree.VariableDeclaration,
+      ): void {
+        for (const declarator of node.declarations) {
+          const init = nullThrows(
```
kirkwaiblinger, line 124: Turns out this is not sound, see 
```ts
async function awaitUsingLoopInitializer() {
  for (await using x of []) {}
}
```

packages/eslint-plugin/tests/docs.test.ts
```diff
@@ -252,9 +252,17 @@ describe('Validating rule docs', () => {
...
+          // Hard-code exceptions for code in headings.
+          if (
+            heading.text ===
+            'Explicit Resource Management (`await using` Statements)'
+          ) {
+            return;
+          }
+
+          expect(heading.text).toBe(titleCase(heading.text));
+        });
```
JoshuaKGoldberg, line 265: Ehh, this is super specific to the heading. I think we'd be better off generalizing. How about removing anything inside code ticks with a regex?

```suggestion
        headings.forEach(heading => {
          const nonCodeText = heading.text.replace(/`[^`]*`/g, '');
          expect(nonCodeText).toBe(titleCase(nonCodeText));
        });
```

packages/eslint-plugin/tests/rules/await-thenable.test.ts
JoshuaKGoldberg: No strong preference from me. I think it's fine as-is 🤷.

JoshuaKGoldberg (changes requested):
Clever, I like how the `getFixOrSuggest` change fits in!

Just requesting changes on the docs, WDYT?

Hugh Grant as Samuel Faulkner saying "oh, that's clever"

packages/eslint-plugin/tests/docs.test.ts
kirkwaiblinger: oh, yeah, much better 👍

JoshuaKGoldberg (approved):
🙌

PR #10216, opened 2024-10-27
fix(eslint-plugin): [no-misused-promises] improve report loc for methods

### LINKED ISSUE #10212: Bug: [no-misused-promises] bad report loc for methods in object literals

### Playground Link

https://typescript-eslint.io/play/#ts=5.6.2&showAST=ts&fileType=.ts&code=C4TwDgpgBAEghgZwGoHsCWATAshYALFDKAXigG8AoKKAMwAoBKALigDd0MKBfCigYxQA7BMCgoW8ZBxz5CJclSiIQgvrUbkoPLkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oFtLlZkiACa1i0Dr0GoMkAO4BDaE0jgwAXxBqgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
type HasVoidMethod = {
  f(): void
}

const o: HasVoidMethod = {
  async f() { }
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-misused-promises": "warn"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

```ts
const o: HasVoidMethod = {
  async f() { }
//~~~~~~~
}
```

### Actual Result

```ts
const o: HasVoidMethod = {
  async f() { }
//       ~~~~~~
}
```

### Additional Info

For methods, should probably underline an explicitly provided return type if present. _Otherwise_, use the getFunctionHeadLoc utility.

Gyumong:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10212 

## Overview
This PR fixes an issue in the no-misused-promises rule where async methods within object literals that return a Promise were not correctly reported when a void return was expected.


packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -490,10 +491,27 @@ export default createRule<Options, MessageId>({
...
+                } else {
+                  context.report({
+                    loc: getFunctionHeadLoc(functionNode, context.sourceCode),
+                    node: functionNode,
+                    messageId: 'voidReturnProperty',
+                  });
+                }
+              }
+            }
+          }
```
JoshuaKGoldberg, line 514: [Logic] This change is a little misleading to read: it looks like `context.report` is only called in some specific cases. Specifically, if all three `if`s are truthy: `signature`, ``tsutils.isThenableType(...)`, and `isFunction(functionNode)`. But what's really supposed to be happening is changing the reported `loc` or `node` based on some logic.

If we're guaranteed that these cases are happening, then there's no need for `if` statements. I'm inferring they're just here so to help TypeScript narrow down some values, right?

Change request: don't add new, redundant runtime checks.

Hint: I deleted all the added variables except `functionNode` and the tests still passed. So I bet you can remove most of the new `if`s. 🙂

JoshuaKGoldberg (changes requested):
Good start! Let's streamline this a bit?

packages/eslint-plugin/src/rules/no-misused-promises.ts
Gyumong: Thank you for your feedback! You’re correct; the extra if statements were initially added to assist TypeScript with type narrowing.

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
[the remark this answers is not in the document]
Gyumong: so kind, thx!

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -490,10 +504,20 @@ export default createRule<Options, MessageId>({
...
         if (isVoidReturningFunctionType(checker, tsNode.name, contextualType)) {
-          context.report({
-            node: node.value,
-            messageId: 'voidReturnProperty',
-          });
+          const functionNode = findFunctionNode(node.value);
+
+          if (functionNode.returnType) {
+            context.report({
+              node: functionNode.returnType,
```
kirkwaiblinger, line 511: Looks like the actual `typeAnnotation` is what we want; seemingly the `returnType` AST node includes the `: T`, not just `T` 🤷 
```suggestion
              node: functionNode.returnType.typeAnnotation,
```

WDYT?

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -490,10 +504,20 @@ export default createRule<Options, MessageId>({
         );
 
         if (isVoidReturningFunctionType(checker, tsNode.name, contextualType)) {
-          context.report({
-            node: node.value,
-            messageId: 'voidReturnProperty',
-          });
+          const functionNode = findFunctionNode(node.value);
```
kirkwaiblinger, line 507: It doesn't seem to me that the `findFunctionNode` function should be necessary here... We _know_ that the ESTree `node` is a property that, according to the TS AST, is a method declaration. So, we shouldn't need to recurse to the enclosing function node; we should be infer directly where the function node would be. Seemingly, 

```ts
const x = {
    f() { } // <-- functionNode _is_ `node.value`
}
```
not sure if there other cases to be aware of.

I note that all the tests pass if do
```suggestion
          const functionNode = node.value as TSESTree.FunctionExpression;
```

I'm not sure if that's due to missing test coverage or that the function call is actually redundant. Could you investigate? Thanks!

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
kirkwaiblinger, line 1: (testing) 

1. Could please ensure that all _existing_ tests that hit the changed branch have their full loc (`line`, `endLine`, `column`, `endColumn`) specified too? (Hint - one way to identify which tests are in the changed branch is to just throw an exception it in and see which tests failed - when I do this, I count 2 existing, and the 2 that you added 🙂)
2. For thoroughness, could you add an `invalid` test or two that have a promise-returning method without the async keyword, too? i.e. `f() { return Promise.resolve('foo')`, `f(): Promise<void> { throw new Error() }`.

kirkwaiblinger (changes requested):
Couple more changes, but I think we're in pretty good shape 🙂

packages/eslint-plugin/src/rules/no-misused-promises.ts
Gyumong: I’ve reviewed the code and tested various scenarios, and it seems safe to use type assertion (node.value as TSESTree.FunctionExpression) here. I couldn’t identify any edge cases where findFunctionNode would be necessary. Therefore, I went ahead with the suggested change. Let me know if you have any further concerns! 😊

kirkwaiblinger: Awesome! Thank you for looking into it! 🙂

packages/eslint-plugin/src/rules/no-misused-promises.ts
kirkwaiblinger, line 1: actually, one more thing, sorry! Would you mind making these same changes in the `if (ts.isPropertyAssignment(tsNode))` branch, if the property being assigned is a function literal?

That way we get things like
```ts
const foo: Foo = {
    f: async () => { }
//     ~~~~~~~~ 
}
```

kirkwaiblinger (changes requested):

packages/eslint-plugin/src/rules/no-misused-promises.ts
Gyumong: Isn't it uncommon to use arrow functions as object methods, since arrow functions don't bind their own this?
However, looking at the current implementation of getFunctionHeadLoc, even when we have an arrow function `(node.type === AST_NODE_TYPES.ArrowFunctionExpression)`, it first hits the Property branch if the parent node is a Property. This prevents proper location handling for arrow functions.
What are your thoughts on this? Should we change the order of conditionals to handle arrow function locations more accurately?

kirkwaiblinger: > Isn't it uncommon to use arrow functions as object methods, since arrow functions don't bind their own this? 

Speaking for myself only here, I use them commonly for exactly that reason 🤷 The lack of `this`-binding decreases my cognitive load since I don't have to think about any of JS's crazy `this`-context stuff. The only time I'd ever voluntarily use method syntax is for instance methods on a `class` (but not an object literal, where I'll always use arrow fns). 

> However, looking at the current implementation of getFunctionHeadLoc, even when we have an arrow function `(node.type === AST_NODE_TYPES.ArrowFunctionExpression)`, it first hits the Property branch if the parent node is a Property. This prevents proper location handling for arrow functions. What are your thoughts on this? Should we change the order of conditionals to handle arrow function locations more accurately?

Hmm, when I try it locally, I get:

```ts
type Foo = {
  f: () => void
}

const foo: Foo = {
  f: async () => { }
//~~~~~~~~~
}
```

which isn't verbatim what I had put in my comment, but I think it's good, too, and definitely better than just underlining the whole function.

What do you think? From your comment, it sounds like you might be describing something different?

Gyumong: > Speaking for myself only here, I use them commonly for exactly that reason 🤷 The lack of this-binding decreases my cognitive load since I don't have to think about any of JS's crazy this-context stuff. The only time I'd ever voluntarily use method syntax is for instance methods on a class (but not an object literal, where I'll always use arrow fns)

Ah, I see. I missed that. Agreed, this is a pretty common pattern for state libraries such as React.

>actually, one more thing, sorry! Would you mind making these same changes in the if (ts.isPropertyAssignment(tsNode)) branch, if the property being assigned is a function literal?

Given that we're specifically checking for function literals (both arrow functions and function expressions) in this context, would it make sense to extract this check into a dedicated utility function? Something like:
```typescript
const isFunctionLiteral = isNodeOfTypes([
AST_NODE_TYPES.FunctionExpression,
AST_NODE_TYPES.ArrowFunctionExpression,
] as const);
```

What are your thoughts on this approach?

kirkwaiblinger: > Given that we're specifically checking for function literals (both arrow functions and function expressions) in this context, would it make sense to extract this check into a dedicated utility function? Something like:
> 
> ```ts
> const isFunctionLiteral = isNodeOfTypes([
> AST_NODE_TYPES.FunctionExpression,
> AST_NODE_TYPES.ArrowFunctionExpression,
> ] as const);
> ```
> 
> What are your thoughts on this approach?

Definitely! I kinda think we should just export a function from `getFunctionHeadLoc.ts` like so
```ts
const isFunctionNode = isNodeOfTypes([
  AST_NODE_TYPES.ArrowFunctionExpression,
  AST_NODE_TYPES.FunctionDeclaration,
  AST_NODE_TYPES.FunctionExpression,
]);
```
That way it exactly matches the allowed input type of `FunctionNode` for `getFunctionHeadLoc`

Gyumong: I see your point about using the existing isFunction utility. I was actually debating whether we need FunctionDeclaration here.

kirkwaiblinger: > I was actually debating whether we need FunctionDeclaration here.

If we add a helper declared and used in the `no-misused-promises` rule file, then no, since the place we'd use it can't be a `FunctionDeclaration`.

If we add a reusable helper in getFunctionHeadLoc.ts, then yes, since it needs to match the `getFunctionHeadLoc` inputs. 

Either of these approaches would solve our needs for this PR, of course. I'm suggesting we go with the second in this PR just because I think it's beneficial for the repo overall 🙂

Gyumong: In the previous discussion, there was a suggestion to add a reusable helper function in getFunctionHeadLoc.ts to benefit the overall repository. Since the isFunction utility already exists, it makes sense to utilize it to reduce code duplication and enhance readability.

kirkwaiblinger (approved):
Love it!

JoshuaKGoldberg (approved):
🙌 yay!

PR #10218, opened 2024-10-27
feat(eslint-plugin): [switch-exhaustiveness-check] add support for "no default" comment

### LINKED ISSUE #2631: [switch-exhaustiveness-check] add support for "no default" comment

**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/switch-exhaustiveness-check": "warn"
  }
}
```

```TS
enum SomeKind {
    Kind1,
    Kind2,
    Kind3,
    Kind4,
}

const foo: SomeKind = determineKind();

// Kind1 and Kind3 need some extra logic for whatever reason
switch (foo) {
    case SomeKind.Kind1:
        // some special logic for this kind
        break;

    case SomeKind.Kind3:
        // some special logic for this kind
        break;

    /* no default */
}

```

**Expected Result**

Similar to the core ESLint rule `default-case`, I would expect the linter to see that I have explicitly stated that no more cases are needed.

**Actual Result**

I get a warning that the switch/case doesn't cover the full enum.

**Versions**

| `@typescript-eslint/eslint-plugin` | `4.3.0` |
| `@typescript-eslint/parser`        | `4.3.0` |
| `ESLint`                           | `7.10.0` |
| `node`                             | `14.13.0` |

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2631

## Overview
similar to eslint's default-case rule, i modified to have the same effect as writing a default clause by specifying the `no default` comment. In the allowDefaultCaseForExhaustiveSwitch option, if no default is specified, an error may occur in the annotation in some cases, so it is necessary to review whether it is okay.

one additional work is left The option considerDefaultExhaustiveForUnions in PR #9954 is in preparation for merging, so you will need to write a test case for that option once this PR is merged.


packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -110,6 +109,28 @@ export default createRule<Options, MessageIds>({
     const checker = services.program.getTypeChecker();
     const compilerOptions = services.program.getCompilerOptions();
 
+    function getCommentDefaultCase(
+      node: TSESTree.SwitchStatement,
+    ): TSESTree.Comment | undefined {
+      const defaultCaseCommentConstants = ['no default', 'No Default'];
```
yeonjuan, line 115: What do you think about using the same pattern as eslint `default-case`? ("i" flag for case-insensitive)

https://github.com/eslint/eslint/blob/35a8858d62cb050fa0b56702e55c94ffaaf6956d/lib/rules/default-case.js#L6-L8
```TS
const DEFAULT_COMMENT_PATTERN = /^no default$/iu;
```

developer-bandi:
I will resolve the conflict by merging the changes to the considerDefaultExhaustiveForUnions option currently reflected in the main branch and request a review again.

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -14,9 +14,11 @@ import {
   requiresQuoting,
 } from '../util';
 
+const DEFAULT_COMMENT_PATTERN = /^no default$/iu;
```
JoshuaKGoldberg, line 17: [Feature] This should be configurable with an option, the way https://eslint.org/docs/latest/rules/default-case#commentpattern is.

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -161,7 +181,7 @@ export default createRule<Options, MessageIds>({
 
       return {
         containsNonLiteralType,
-        defaultCase,
+        defaultCase: defaultCase ? defaultCase : getCommentDefaultCase(node),
```
JoshuaKGoldberg, line 184: [Style] Nit: I'm surprised this isn't caught by https://typescript-eslint.io/rules/prefer-nullish-coalescing ...

```suggestion
        defaultCase: defaultCase ?? getCommentDefaultCase(node),
```

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -176,7 +196,7 @@ export default createRule<Options, MessageIds>({
 
       // If considerDefaultExhaustiveForUnions is enabled, the presence of a default case
       // always makes the switch exhaustive.
-      if (considerDefaultExhaustiveForUnions && defaultCase != null) {
+      if (considerDefaultExhaustiveForUnions && defaultCase) {
```
JoshuaKGoldberg, line 199: Nit: unnecessary unrelated change, let's not increase PR noise if we don't have to:

```suggestion
      if (considerDefaultExhaustiveForUnions && defaultCase != null) {
```

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -218,7 +238,9 @@ export default createRule<Options, MessageIds>({
     ): TSESLint.RuleFix {
       const lastCase =
         node.cases.length > 0 ? node.cases[node.cases.length - 1] : null;
-      const defaultCase = node.cases.find(caseEl => caseEl.test == null);
+      const defaultCase =
+        node.cases.find(caseEl => caseEl.test == null) ??
+        getCommentDefaultCase(node);
```
JoshuaKGoldberg, line 243: [Refactor] This should have already been found by `getSwitchMetadata`. Now that the logic around finding a default case is getting more complex, let's take this in as a `fixSwitch` parameter instead of re-finding it.

JoshuaKGoldberg (changes requested):
I took a quick initial pass and think there should be an option added to maintain parity with ESLint core. I can take a deeper look once that's in (or if I'm wrong about that). Thanks!

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -38,6 +40,11 @@ type Options = [
      */
     requireDefaultForNonUnion?: boolean;
 
+    /**
+     * Regular expression to evaluate comments to mean default case
+     */
+    defalutCaseCommentPattern?: string;
```
JoshuaKGoldberg, line 46: [Docs] Typo, and also this should have an entry in `packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx`.

```suggestion
    defaultCaseCommentPattern?: string;
```

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -81,6 +88,10 @@ export default createRule<Options, MessageIds>({
             type: 'boolean',
             description: `If 'true', the 'default' clause is used to determine whether the switch statement is exhaustive for union type`,
           },
+          defalutCaseCommentPattern: {
+            type: 'string',
+            description: `Regular expression to evaluate comments to mean default case`,
```
JoshuaKGoldberg, line 93: [Docs] WDYT about this phrasing, to make it a bit more clear what the comment does?

```suggestion
            description: `Regular expression for a comment that can indicate an intentionally omitted default case.`,
```

packages/eslint-plugin/tests/rules/switch-exhaustiveness-check.test.ts
JoshuaKGoldberg, line 1: [Testing] Missing test cases for `defaultCaseCommentPattern`.

JoshuaKGoldberg (changes requested):
A few necessities aren't filled in: docs and tests for the option. If you fill those in we can review more deeply. Thanks!

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -78,6 +80,25 @@ switch (literal) {
 }
 ```
 
+### defaultCaseCommentPattern
+
+
+Basically, there is logic to treat annotations matching /^no default$/iu as default case.
+If you want to change this pattern, just enter the pattern in this option.
```
JoshuaKGoldberg, line 88: [Docs] Cleanup for clarity:

```suggestion
It can sometimes be preferable to omit the default case for only some switch statements.
For those situations, this rule can be given a pattern for a comment that's allowed to take the place of a `default:`.
```

https://jameshfisher.com/2017/02/22/dont-use-simply - same with _"just"_.

As of #9925, the option's default value is explicitly stated in the automatically added description.

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -254,6 +275,37 @@ switch (fruit) {
...
+switch (value) {
+  case 'a':
+    break;
+  // no default
+}
+```
+
+</TabItem>
+</Tabs>
+
```
JoshuaKGoldberg, line 308: [Docs] I don't think this is necessary, given that `defaultCaseCommentPattern` already documents itself and automatically mentions its default value.

```suggestion
```

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
JoshuaKGoldberg, line 17: [Feature] Actually, now that I'm re-reading this, I'm not sure we should have a default here. The core ESLint rule does it but it seems kind of arbitrary to me. Maybe we should just have no default at all?

cc @typescript-eslint/triage-team

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
```diff
@@ -110,6 +122,26 @@ export default createRule<Options, MessageIds>({
...
+    ): TSESTree.Comment | undefined {
+      const lastCase = node.cases.at(-1);
+      const commentsAfterLastCase = lastCase
+        ? context.sourceCode.getCommentsAfter(lastCase)
+        : [];
+      const defaultCaseComment = commentsAfterLastCase.at(-1);
+      const commentRegExp =
+        defaultCaseCommentPattern != null
+          ? new RegExp(defaultCaseCommentPattern, 'u')
+          : DEFAULT_COMMENT_PATTERN;
```
JoshuaKGoldberg, line 136: [Refactor] This re-runs the `new RegExp` on each `switch` statement. Let's move it to a root statement inside the `create()`, not inside a nested function, so it only happens once.

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
kirkwaiblinger: I think 
- "no default" is a sensible default ignore comment pattern. 
- Sensible defaults are good things since they help standardize across projects (and in this case even between rules)
- I probably wouldn't have any comment ignore pattern on-by-default though
- Making a separate on/off toggle for the comment ignore pattern and a regexp option for it is overly verbose
- Alignment with [default-case](https://eslint.org/docs/latest/rules/default-case) is good for user experience

Therefore probably having the single regexp option that defaults to "no default" is the best compromise option here. But no strong opinion.

JoshuaKGoldberg: 👍 I'm ok with this. Your reasoning is convincing to me, in particular the alignment with the core rule.

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
developer-bandi: The above link helped me a lot. Thank you!

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
FloEdelmann, line 17: If there's a default value, please specify it in `defaultOptions`.

developer-bandi: In order to put a default value in defaultOptions, the defaultCaseCommentPattern option must be able to receive a flag. Therefore, this option should be in the form of an array, with an expression at the first index and a flag at the second index. 

Previously, the default value was treated as a separate variable because of equivalence with eslint rules. Would there be any problem with implementing it like this?

JoshuaKGoldberg: As much as I would love to keep things to the `defaultOptions` plumbing, I don't think this is doable. I think this PR is fine as-is.

packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts
JoshuaKGoldberg: https://github.com/typescript-eslint/typescript-eslint/issues/10470

JoshuaKGoldberg (dismissed):
LGTM, thanks for all the iteration on this! Your hard work is paying off 🔥

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -78,6 +78,25 @@ switch (literal) {
 }
 ```
 
+### defaultCaseCommentPattern
```
JoshuaKGoldberg, line 81: The website build is failing because these should be backticked:

```suggestion
### `defaultCaseCommentPattern`
```

packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.mdx
```diff
@@ -78,6 +78,25 @@ switch (literal) {
 }
 ```
 
+### `defaultCaseCommentPattern`
+
```
JoshuaKGoldberg, line 83: Per @FloEdelmann's https://github.com/typescript-eslint/typescript-eslint/pull/10218#discussion_r1868991801, not being able to use `defaultOptions` means the default isn't being put into docs. So for now:

```suggestion
{/* insert option description */}

Default: `/^no default$/iu`.
```

PR #10235, opened 2024-10-30
fix(eslint-plugin): [consistent-type-definitions] don't leave trailing parens when fixing type to interface

### LINKED ISSUE #10233: Bug: consistent-type-definitions leaves trailing parenthesis when converting a paren-wrapped type to an interface

### Issue Description

This isn't actually a complex issue, but I didn't immediately see how to set eslint version 9 in the playground and I wasn't interested in learning, since I already had a reproduction repo set up.

The issue is that when you use the --fix argument with the `@typescript-eslint/consistent-type-definitions` rule, it will leave a trailing parenthesis if the type it converts was wrapped in parens.

For example, the following code

```typescript
export type Interface = ({
    [key: string]: unknown;
});
```

becomes

```typescript
export interface Interface {
    [key: string]: unknown;
});
```

triggering a syntax error.

I tried this on the latest version of typescript-eslint, including 8.12.2 and 8.12.3-alpha.1

https://github.com/rdecoito/repro-consistent-type-definitions

### Repro Steps

Copy-pasted from README in repo:

1. Clone this repo
2. Run `npm i`
3. Run `npm run lint`
	- You'll see an error reported that the type defined in src/index.ts should be an interface instead
4. Run `npm run lint -- --fix`
	- You'll see a new syntax error reported. Open src/index.ts to see the problem: the rule left behind a parenthesis that it shouldn't have

### Versions

| `typescript-eslint`             | `~8.11.0` |
| `TypeScript`                           | `5.6.3` |
| `ESLint`                               | `~9.13.0` |
| `node`                                 | `20.18.0` |

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10233 

## Overview

rewrite the fixer to be more robust

### DIFF
--- a/packages/eslint-plugin/src/rules/consistent-type-definitions.ts
+++ b/packages/eslint-plugin/src/rules/consistent-type-definitions.ts
@@ -1,8 +1,9 @@
 import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
 
 import { AST_NODE_TYPES, AST_TOKEN_TYPES } from '@typescript-eslint/utils';
+import * as ts from 'typescript';
 
-import { createRule } from '../util';
+import { createRule, nullThrows, NullThrowsReasons } from '../util';
 
 export default createRule({
   name: 'consistent-type-definitions',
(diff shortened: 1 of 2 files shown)


auvred (approved):
Nice!

bradzacher (dismissed):

packages/eslint-plugin/src/rules/consistent-type-definitions.ts
```diff
@@ -68,15 +71,27 @@ export default createRule({
...
+                    afterToken.type === AST_TOKEN_TYPES.Punctuator &&
+                    afterToken.value === ';'
+                  ) {
+                    fixes.push(fixer.remove(afterToken));
+                    break;
+                  }
+                }
+
+                break;
               }
```
bradzacher, line 95: The better way to do this might be to do (psuedocode):
1) `getTokenBefore(typeAnnotation, '=')`
2) `getTokenAfter(typeAnnotation, ';')`
3) replace range from (1) to (2) with `typEAnnotation`

Josh-Cena: ^Agreed; this feels a bit hacky

kirkwaiblinger: yeah, this was a good call. Result is much cleaner now 👍

packages/eslint-plugin/tests/rules/consistent-type-definitions.test.ts
```diff
@@ -347,6 +359,61 @@ export declare interface Test {
...
+  a: string;
+}
+      `,
+    },
+    {
+      code: noFormat`
+type Foo = ((((((((({
+  a: string;
+})))))))));
+      `,
```
bradzacher, line 388: Could you add a test case like

```ts
type Foo = ((({
  a: string;
})))

// next statement has a semi
const bar = 1;
```

To ensure we are handling the no-semi case?

kirkwaiblinger: sure! added 👍

kirkwaiblinger: Note that the risk of messing up a semi/no-semi issue is also lessened now that the fixer isn't querying for it; it's just deleting up to `node.range[1]`.

bradzacher (dismissed):

bradzacher (approved):

PR #10299, opened 2024-11-07
feat(eslint-plugin): [prefer-nullish-coalescing] fix detection of `ignoreConditionalTests` involving boolean `!` operator

### LINKED ISSUE #10284: Bug: [prefer-nullish-coalescing] `ignoreConditionalTests` doesn't ignore the boolean `!` operator in a condition

### Playground Link

https://typescript-eslint.io/play/#ts=5.6.3&fileType=.ts&code=DYUwLgBAZg9jBcEDOYBOBLAdgcwgHwgFdMATEKLEEiAXggHJYZ6BuAKFEgCMBDVRFBhz4ipcpWp16vVKzZsQADwAOMVJCjEAxmHQxMEMCBQBhHkhAAKAJQQA3mwCQ6KBEtN8BGbYePHW-SQYUAA6YBhsS0Y4CDUIGXprdkcAXzY0hRU1DW1dfUNjMDMLACYbeycXNwBCdxi8Lz5rHyd-QOCQMIiozBgNGLiEpKc0jI86YjIKTCp2GVpRKYl2IA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6MAG1IvaOOiQAuuDABfEPqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkAhqqrAO4BSZlemArugDTgaMABbpoAa3YUwXXgJpR0JALKwAJtwwBJIniyxE02f0G0AZrEzR0AYWIlkJPbpt0HRAOY6AYmnQA5OnglIx4TBUhMJVhUADd0dmJVDQxQuVMoEnFkRAAZZAAjG1EJNPChElhuK3RlOkNOMPkKrmRoPDLm2iJYAFUiaBLxdDUdNXQADxGAQWhrEg4ZJozIHq0kVDbkPAAldDxq0k6Vnu8GVDxhTCqPYVcSJR0AZWZtweOInv7uB7Vc2GgDEWxi6UC%2BRB%2BIwACnRMEF9lhgctPrB1ohNtBtth4phMMhxh8hPB1Jp0HtKqhuARiNJIAVuEQ1BhMJBQZAcQU6AR4MlSU8AJ66OgTaRmIHoMCCAC%2BIClQA&tokens=false

### Repro Code

```TypeScript
let foo: string | undefined = 'foo';
let bar: string | undefined = 'bar';

export function testCase() {
	if (foo || bar) {
		console.log('foo or bar');
	}
}

export function testCase2() {
	if (!(foo || bar)) {
		console.log('not foo or bar');
	}
}

foo = undefined;
bar = undefined;
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/prefer-nullish-coalescing": ["error"]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "allowJs": true,
    "checkJs": true,
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "resolveJsonModule": true,
    "skipLibCheck": true,
    "sourceMap": true,
    "strict": true,
    "noUncheckedIndexedAccess": true,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitOverride": true,
    "moduleResolution": "bundler",
    "verbatimModuleSyntax": false 
  }
}
```

### Expected Result

I would have expected that the if-statement inside `testCase2` also is allowed, but after updating from `8.12.2` to `8.13.0` it results in an error suggesting to use `??` instead referring to the `prefer-nullish-coalescing` rule.

I think it should to be allowed to do this check to see whether one of two strings are truthy inside an if-statement as shown in `testCase2`.

### Actual Result

In the example provided `testCase2` will  **not** run the `console.log` statement when `foo=""` and `bar="bar"`, however, the `testCase1` **runs** the if-statement in the same case. This is due to the difference between `||` and `??` when it comes to the empty string.  

Meaning the rule now suggests something that is logically not the same, and would change the behavior of code similar to the one shown in the example 

### Additional Info

_No response_

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10284

## Overview

Since we maintain the same logic as the ignoreBooleanCoercion option, should we add the same logic to this option as well?

```typescript
let a: string | undefined;
let b: string | undefined;

Boolean(!!(a || b)) 
Boolean(!(a || b)) 
```


kirkwaiblinger:
Hey! merging from main for ya to fix the build errors

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -476,6 +476,10 @@ function isConditionalTest(node: TSESTree.Node): boolean {
     return isConditionalTest(parent);
   }
 
+  if (parent.type === AST_NODE_TYPES.UnaryExpression) {
```
kirkwaiblinger, line 479: be sure to check that it's really a `!` operator! Or you will get bugs on code like this 🙂 
```ts
declare const a: string | undefined;
declare const b: string;
if (+(a || b)) {
}
```

developer-bandi: I made a mistake because I was not familiar with the type. I've also added a corresponding test case just in case.

kirkwaiblinger (approved):
thanks!

kirkwaiblinger:
> Since we maintain the same logic as the ignoreBooleanCoercion option, should we add the same logic to this option as well?
> 
> ```ts
> let a: string | undefined;
> let b: string | undefined;
> 
> Boolean(!!(a || b)) 
> Boolean(!(a || b)) 
> ```

Yeah, probably a good idea to do this tbh.

kirkwaiblinger (changes requested):
What you have looks good, but I think it would be good to implement your suggestion in  https://github.com/typescript-eslint/typescript-eslint/pull/10299#issuecomment-2467616526

kirkwaiblinger (approved):
salute gif

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
```diff
@@ -583,6 +583,60 @@ if (((a = b), b || c)) {
...
+          ignoreBooleanCoercion: true,
+        },
+      ],
+    },
+    {
+      code: `
+let a: string | undefined;
+let b: string | undefined;
+
+Boolean(!(a || b));
```
kirkwaiblinger, line 632: oh, hmm. I hate to do this. But actually, now that I'm thinking harder about this, I don't think we want this. This would be unambiguously better to write as `!Boolean(a || b)`, so the `!` really is the sort of implicit coercion that `ignoreBooleanCoercion` constructor is trying to force you to use `Boolean` on instead.

So sorry to change request changes one way and then request they be undone, but would you mind reverting the `ignoreBooleanCoercion` changes?

kirkwaiblinger (changes requested):
Changed my mind about https://github.com/typescript-eslint/typescript-eslint/pull/10299#pullrequestreview-2426485025, sorry! 🫣

developer-bandi:
Thank you. I have reflected this.

kirkwaiblinger (approved):
Great! And sorry again for my previous mistaken review!

bradzacher (approved):

PR #10310, opened 2024-11-08
fix(eslint-plugin): [no-misused-promises] `inheritedMethods` and `properties` to check all statically analyzable declarations

### LINKED ISSUE #10211: Bug: [no-misused-promises] `inheritedMethods` check should flag all statically analyzable declarations, not just identifiers and numbers.

### Playground Link

https://typescript-eslint.io/play/#ts=5.6.2&fileType=.ts&code=MYewdgzgLgBNCGUCWwDKBPAtgIxAGxgF4YMd8A6AMxACcAKAcgWWDi1zwYEoBuAKD5IwUAKY1K8YCJgBJYWIlSYAbz4wYSACYjhSSkjEAuGHS5EAfDABuILf3UBGY6YvXbm%2BzABMzs4Us2dmpwUDRCAOYAMkiiNPB4vq6BHsGgmAAOAK6imqihEdGx8Yn%2BbkHqAPQVMADuIngEANZgIDVgbGR4wQDapBzkMWKItAC6JQHu-MFVMHgiEBBwmTTSejBQABZIi%2BCri%2BkgC0jYc%2Bsgs0Kw1DQ9zCh9%2BGMmfhNBAL4CfMB48AswAMI-ABeQI0GTmmB0UEWcliimkqnUv3QYFYWihegM9DMyhgHyREBRrAcLlx%2BJgyNRMG6XhGpLxwUprCY%2BTAUUGcU49PJTJgDDSWRyeTCbMKQy5OIZBKJ1IeeAGRSgo25nl53TuaHYjxVfDeQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oFtLlZkiACa1i0Dr0GoMkAO4BDaE0jgwAXxBqgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false

### Repro Code

```TypeScript
const staticSymbol = Symbol.for('static symbol');

interface Interface {
  identifier: () => void;
  1: () => void;
  2: () => void;
  stringLiteral: () => void;
  computedStringLiteral: () => void;
  // well known symbol
  [Symbol.iterator]: () => void;

  // less sure if this one is possible to lint for
  [staticSymbol]: () => void;
}


class Clazz implements Interface {
  async identifier() { }
  async 1() { }
  async [2]() { }
  async 'stringLiteral'() { }
  async 'computedStringLiteral'() { }
  async [Symbol.iterator]() { };
  async [staticSymbol]() { };
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-misused-promises": "warn"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true
  }
}
```

### Expected Result

All method declarations should flag

### Actual Result

Only `async identifier() {}` and `async 3() {}` flagged.

### Additional Info

_No response_

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10211

## Overview

This PR resolves #10211 and uses the existing `getStaticMemberAccessValue()` utility instead of `tsutils.getPropertyOfType()` to handle most statically analyzable declarations/properties.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-misused-promises.ts
+++ b/packages/eslint-plugin/src/rules/no-misused-promises.ts
@@ -7,6 +7,7 @@ import * as ts from 'typescript';
 import {
   createRule,
   getParserServices,
+  getStaticMemberAccessValue,
   isArrayMethodCallWithPredicate,
   isFunction,
   isRestParameterDeclaration,
@@ -568,11 +569,26 @@ export default createRule<Options, MessageId>({
           continue;
         }
 
+        if (
+          node.type !== AST_NODE_TYPES.MethodDefinition &&
+          node.type !== AST_NODE_TYPES.TSAbstractMethodDefinition &&
+          node.type !== AST_NODE_TYPES.TSMethodSignature &&
+          node.type !== AST_NODE_TYPES.TSPropertySignature
+        ) {
+          continue;
+        }
+
+        const staticAccessValue = getStaticMemberAccessValue(node, context);
+
+        if (!staticAccessValue) {
+          continue;
+        }
+
         for (const heritageType of heritageTypes) {
           checkHeritageTypeForMemberReturningVoid(
             nodeMember,
             heritageType,
-            memberName,
+            staticAccessValue,
           );
         }
       }
@@ -588,9 +604,13 @@ export default createRule<Options, MessageId>({
     function checkHeritageTypeForMemberReturningVoid(
       nodeMember: ts.Node,
       heritageType: ts.Type,
-      memberName: string,
+      staticAccessValue: string | symbol,
     ): void {
-      const heritageMember = getMemberIfExists(heritageType, memberName);
+      const heritageMember = getMemberIfExists(
+        heritageType,
+        staticAccessValue,
+        checker,
+      );
       if (heritageMember === undefined) {
         return;
       }
@@ -944,18 +964,44 @@ function getHeritageTypes(
     .map(typeExpression => checker.getTypeAtLocation(typeExpression));
 }
 
+function getWellKnownStringOfSymbol(symbol: symbol): string | null {
+  switch (symbol) {
+    case Symbol.iterator:
+      return 'iterator';
+    case Symbol.asyncIterator:
+      return 'asyncIterator';
+  }
+
+  return null;
+}
+
 /**
- * @returns The member with the given name in `type`, if it exists.
+ * @returns The member with the given name or known-symbol in `type`, if it exists.
  */
 function getMemberIfExists(
   type: ts.Type,
-  memberName: string,
+  staticAccessValue: string | symbol,
+  checker: ts.TypeChecker,
 ): ts.Symbol | undefined {
-  const escapedMemberName = ts.escapeLeadingUnderscores(memberName);
-  const symbolMemberMatch = type.getSymbol()?.members?.get(escapedMemberName);
-  return (
-    symbolMemberMatch ?? tsutils.getPropertyOfType(type, escapedMemberName)
-  );
+  if (typeof staticAccessValue === 'string') {
+    const escapedMemberName = ts.escapeLeadingUnderscores(staticAccessValue);
+    const symbolMemberMatch = type.getSymbol()?.members?.get(escapedMemberName);
+    return (
+      symbolMemberMatch ?? tsutils.getPropertyOfType(type, escapedMemberName)
+    );
+  }
+
+  const wellKnownSymbolName = getWellKnownStringOfSymbol(staticAccessValue);
+
+  if (wellKnownSymbolName) {
+    return tsutils.getWellKnownSymbolPropertyOfType(
+      type,
+      wellKnownSymbolName,
+      checker,
+    );
+  }
+
+  return undefined;
 }
 
 function isStaticMember(node: TSESTree.Node): boolean {
--- a/packages/eslint-plugin/src/rules/no-misused-promises.ts
+++ b/packages/eslint-plugin/src/rules/no-misused-promises.ts
@@ -455,9 +455,6 @@ export default createRule<Options, MessageId>({
           });
         }
       } else if (ts.isMethodDeclaration(tsNode)) {
-        if (ts.isComputedPropertyName(tsNode.name)) {
-          return;
-        }
         const obj = tsNode.parent;
 
         // Below condition isn't satisfied unless something goes wrong,
@@ -477,9 +474,14 @@ export default createRule<Options, MessageId>({
         if (objType === undefined) {
           return;
         }
-        const propertySymbol = checker.getPropertyOfType(
+        const staticAccessValue = getStaticMemberAccessValue(node, context);
+        if (!staticAccessValue) {
+          return;
+        }
+        const propertySymbol = getMemberIfExists(
           objType,
-          tsNode.name.text,
+          staticAccessValue,
+          checker,
         );
         if (propertySymbol === undefined) {
           return;
--- a/packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
@@ -1,4 +1,4 @@
-import { RuleTester } from '@typescript-eslint/rule-tester';
+import { noFormat, RuleTester } from '@typescript-eslint/rule-tester';
 
 import rule from '../../src/rules/no-misused-promises';
 import { getFixturesRootDir } from '../RuleTester';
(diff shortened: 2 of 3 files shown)


packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -944,18 +966,44 @@ function getHeritageTypes(
...
+function getWellKnownStringOfSymbol(symbol: symbol): string | null {
+  switch (symbol) {
+    case Symbol.iterator:
+      return 'iterator';
+    case Symbol.asyncIterator:
+      return 'asyncIterator';
+  }
+
+  return null;
+}
```
ronami, line 978: I'm not sure if there's a better way to map a symbol to its string counterpart `tsutils.getWellKnownSymbolPropertyOfType` is expecting. The list for [well-known symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#well-known_symbols) isn't long, but this will still require maintaining this mapping.

ronami: If this approach makes sense, this list can be expanded to support other well-known symbols.

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -568,11 +571,26 @@ export default createRule<Options, MessageId>({
           continue;
         }
 
+        if (
+          node.type !== AST_NODE_TYPES.MethodDefinition &&
+          node.type !== AST_NODE_TYPES.TSAbstractMethodDefinition &&
+          node.type !== AST_NODE_TYPES.TSMethodSignature &&
+          node.type !== AST_NODE_TYPES.TSPropertySignature
```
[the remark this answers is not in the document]
ronami: This is necessary mainly for type-checking, and I don't know if this can realistically have a runtime scenario. Please let me know if this should be handled differently (currently, there is no coverage for this).

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -568,11 +571,27 @@ export default createRule<Options, MessageId>({
           continue;
         }
 
+        if (
+          node.type !== AST_NODE_TYPES.MethodDefinition &&
+          node.type !== AST_NODE_TYPES.TSAbstractMethodDefinition &&
+          node.type !== AST_NODE_TYPES.TSMethodSignature &&
+          node.type !== AST_NODE_TYPES.TSPropertySignature &&
+          node.type !== AST_NODE_TYPES.PropertyDefinition
```
ronami, line 579: This is enough to pass the test suit but there are additional types available to satisfy `getStaticMemberAccessValue`.

ronami: This is necessary mainly for type-checking, and I don't know if this can realistically have a runtime scenario. Please let me know if this should be handled differently (currently, there is no coverage for this).

packages/eslint-plugin/tests/rules/no-misused-promises.test.ts
JoshuaKGoldberg, line 1: [Testing] One thing under test per case please (https://typescript-eslint.io/contributing/pull-requests#granular-unit-tests)

packages/eslint-plugin/src/rules/no-misused-promises.ts
JoshuaKGoldberg: I don't believe this can realistically be a runtime scenario. I think the real issue here is that although `nodeMember` is type `ts.TypeElement | ts.ClassElement`, `node` is type `TSESTree.Node`. 

The two strategies I'd think could be good would be:

* Simpler: use an `as` to correct TypeScript's understanding of `node`'s type to be `NodeWithKey` (which will need to be exported from `misc.ts`)
* More aligned with how most of our typed lint rules go: instead of iterating over the `tsNode.members`, iterate over the members of `node`

The latter is probably a lot more work and changes the existing code. So I think the `as` strategy would be reasonable.

packages/eslint-plugin/src/rules/no-misused-promises.ts
JoshuaKGoldberg: Is there a reason we need to stick to well-known symbols? I'd think we should go with any detected symbol, no?

JoshuaKGoldberg (changes requested):
I like where this is going!

packages/eslint-plugin/src/rules/no-misused-promises.ts
ronami: Yes, I've started with a type-assertion on my initial implementation but I wasn't sure if there's a realistic case where this would explode.

I updated the code to use a type assertion, as the second option seems to be a lot more work.

packages/eslint-plugin/src/rules/no-misused-promises.ts
ronami: Hmm, I might be missing something, but I'm not entirely sure how to check if a JavaScript `symbol` exists on another `ts.Type` 🤔 

Here's how existing code checks if an inherited type has a `string` symbol:

```ts
const escapedMemberName = ts.escapeLeadingUnderscores(staticAccessValue);
const symbolMemberMatch = type.getSymbol()?.members?.get(escapedMemberName);

return (
  symbolMemberMatch ?? tsutils.getPropertyOfType(type, escapedMemberName)
);
```

I've looked into the `SymbolObject` but couldn't find a way to check against an arbitrary `symbol` (not to be confused with `ts.Symbol`).

Because of this, I've created a function to map between known symbols and their string representation, which are passed to `tsutils.getWellKnownSymbolPropertyOfType` to check if they exist on the inherited type.

JoshuaKGoldberg:
👋 Hey @ronami, did you want to re-request review on this?

No pressure, just asking because it's been a bit and I've seen other things from you in the meantime. 🙂

ronami:
> 👋 Hey @ronami, did you want to re-request review on this?
> 
> No pressure, just asking because it's been a bit and I've seen other things from you in the meantime. 🙂

Hey @JoshuaKGoldberg, thanks for checking up on this ❤️ 

I've [commented here](https://github.com/typescript-eslint/typescript-eslint/pull/10310#discussion_r1842679683) on something I'm not sure how to implement. I didn't want to ask for a re-review since it's still an open discussion.

packages/eslint-plugin/src/rules/no-misused-promises.ts
JoshuaKGoldberg: Hmm gotcha thanks. cc @kirkwaiblinger - I'm fine with this as a patchwork helper, with as little hardcoding as possible.

```ts
function getWellKnownStringOfSymbol(symbol: symbol): string | null {
  for (const [key, value] of Object.entries(Symbol)) {
    if (symbol === value) {
      return key;
    }
  }

  return null;
}
```

And, sorry for dropping the ball on this again 🤦 there's something about symbols that makes me instinctively forget the conversation...

ronami: I've updated the PR to not hard-code well-known symbols, thanks!

The only alternative I had in mind for not hard-coding was to extract the value from [`Symbol.prototype.toString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toString) with a regex (which I really didn't want to do), but I didn't think of iterating the global `Symbol`'s keys.

This means that arbitrary symbols (e.g. `Symbol.for('foo')`) won't be reported by the rule though.

JoshuaKGoldberg: Yeah I don't see a way to work for arbitrary symbols 😞

kirkwaiblinger: cross-linking discord thread https://discord.com/channels/1026804805894672454/1322853884292497418

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -492,9 +492,14 @@ export default createRule<Options, MessageId>({
         if (objType == null) {
           return;
         }
-        const propertySymbol = checker.getPropertyOfType(
+        const staticAccessValue = getStaticMemberAccessValue(node, context);
+        if (!staticAccessValue) {
```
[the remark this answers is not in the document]
ronami: Oh, nice catch! Is there any reason we don't have `strict-boolean-expressions` enabled?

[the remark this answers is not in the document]
kirkwaiblinger: Because IIRC both Joshs don't like it hehe. Brad and I both do like it. 🤷‍♂️

kirkwaiblinger:
Hey @ronami, do you plan to make further changes regarding https://github.com/typescript-eslint/typescript-eslint/pull/10310#discussion_r1899103025 or is this intended to be reviewed as is?

ronami:
> Hey @ronami, do you plan to make further changes regarding [#10310 (comment)](https://github.com/typescript-eslint/typescript-eslint/pull/10310#discussion_r1899103025) or is this intended to be reviewed as is?

Oh, I'm sorry; I've looked into it but completely forgot to get back to it.

---

I didn't consider the issue with the current approach, which could result in different values on different machines and NodeJS versions (though looking into [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#browser_compatibility), it seems pretty much all well-known symbols are supported on rather old Node versions already).

It does seem like `getProperties()[0].escapedName` matches the child property and has a similar form to the one [`getWellKnownSymbolPropertyOfType`](https://github.com/JoshuaKGoldberg/ts-api-utils/blob/b61005709aec3090ea9e93c231055cf2cfa2715f/src/types/getters.ts#L88) uses (`__@staticSymbol@13`), though as you said, it's not part of the public API.

I haven't noticed `heritageType.getProperties().map(sym => sym.links.nameType)`, which, while not part of the API, does seem to work in various edge cases (interesting 🙃).

Having a way to get the key type of a property `ts.Symbol` would be handy. I'll be happy to check if this can be added to the public API or if there's a reasonable hack we can currently use. Should this be filed as a GitHub issue or discussed on the TypeScript Discord channel?

Thanks for your help with this ❤️

kirkwaiblinger:
> I didn't consider the issue with the current approach, which could result in different values on different machines and NodeJS versions (though looking into [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#browser_compatibility), it seems pretty much all well-known symbols are supported on rather old Node versions already).

Ok, that definitely alleviates some of the concern 👍 Though I guess there's maybe still a forwards-compatibility question. 

> I haven't noticed `heritageType.getProperties().map(sym => sym.links.nameType)`, which, while not part of the API, does seem to work in various edge cases, and types do seem `===` to each other, interesting 🙃.
> 
> Having a way to get the key type of a property `Symbol` would be useful. I'll be happy to check if this can be added to the public API or if there's a reasonable hack we can currently use. Should this be filed as a GitHub issue or discussed on the TypeScript Discord channel?

What I usually do is just ask questions on the TS discord and copy down the link on the issue/PR I'm working on - partly for visibility and partly just so I can find it again myself! 😅 

> Thanks for your help with this ❤️

❤️

JoshuaKGoldberg:
👋 @ronami just checking in, how is this going?

ronami:

PR #10314, opened 2024-11-09
fix(eslint-plugin): [await-thenable, return-await] don't flag awaiting unconstrained type parameter as unnecessary

### LINKED ISSUE #10311: Bug: [await-thenable] Don't flag values of an unconstrained type parameter

### Playground Link

https://typescript-eslint.io/play/#ts=5.6.2&fileType=.ts&code=IYZwngdgxgBAZgV2gFwJYHsIwO4CdgAOBAprgDwAqAfABQBuwANgsQFwwUCUMA3gFAwYuYsgS4swbMFTIYDZsQDcfAL5A&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AQwHd3K78ALRE3YAjJOiiJo0APbRI4MAF8QSoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
async function wrapper<T>(value: T) {
  return await value;
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

The `await value` shouldn't be reported on. We don't know that `T` is or isn't a Promise/Thenable. I'd like to be able to use the `await` just in case it is.

### Actual Result

On the `await value`:

```plaintext
Unexpected `await` of a non-Promise (non-"Thenable") value.
```

### Additional Info

For a slightly more realistic reproduction, consider a function that `await`s the result of calling a function for a generic value:

```ts
type GetSomething<T> = (timestamp: number) => T;

async function wrapper<T>(getter: GetSomething<T>) {
  const timestamp = performance.now();

  // I'd like to await in here to preserve the call stack
  return await getter(timestamp);
}
```

Somewhat related: #8301, where it was noted that the rule intentionally ignores `any` and `unknown` values because there's no guarantee they aren't a Promise.

💖

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10311

## Overview

This PR resolves #10311 and stops reporting on some type constraints (see [playground link](https://typescript-eslint.io/play/#ts=5.6.2&fileType=.ts&code=PTAEBcE8AcFNWgQwE6ILa3LZoDuBLcAC1ADsB7UAY3NIGdxV9TwAoROyUq0AMwFdu4fLQiwGARgA8AFQB8ACgBuiADb9YALlAyAlKADerUKGSZ%2ByUqES5EhUCvWwA3KwC%2BrViAgx4SVBhYOATEoAAGiKSQYaDkOGGCANYUuKQxNPSMdizsnNx8glTColgMAEyyoLAAHlikACZ0oEkppIqOGtp6hsam5pbWtvYdLu6eHFw8AkIiVqXgAMyVNXWN1lHtap06%2BkYmZuAWVjZ24A5box5eYPX4ZkXhkdGx8S3kqTFQcAgcdOK5kwKMxK4nAABZlBdtG9UrtegcjoNTucnK4rhN8tMirMxAwAKyQpzaJ5w-b9Y5DM4jNGsIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AQwHd3K78ALRE3YAjJOiiJo0APbRI4MAF8QSoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)).

As per @kirkwaiblinger's [comment](https://github.com/typescript-eslint/typescript-eslint/issues/10311#issuecomment-2466115554), I shared common logic between `await-thenable` and `return-await`, so this fixes similar issues with both ([playground link for `return-await`](https://typescript-eslint.io/play/#ts=5.6.2&fileType=.ts&code=PTAEBcE8AcFNWgQwE6ILa3LZoDuBLcAC1ADsB7UAY3NIGdxV9TwAoROyUq0AMwFdu4fLQiwGARgA8AFQB8ACgBuiADb9YALlAyAlKADerUKGSZ%2ByUqES5EhUCvWwA3KwC%2BrViAgx4SVBhYOATEoAAGiKSQYaDkOGGCANYUuKQxNPSMdizsnNx8glTColgMAEyyoLAAHlikACZ0oEkppIqOGtp6hsam5pbWtvYdLu6eHFw8AkIiVqXgAMyVNXWN1lHtap06%2BkYmZuAWVjZ24A5box5eYPX4ZkXhkdGx8S3kqTFQcAgcdOK5kwKMxK4nAABZlBdtG9UrtegcjoNTucnK4rhN8tMirMxAwAKyQpzaJ5w-b9Y5DM4jNGsIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEIHodQAEAXATwAcUBjaASzKIFoV4aA7InAQwHdObGiAC0StOAIyToovaK0gAacNkjFyVWvSbIW7HNERFYshjz5EpkRNGgB7aJCUBfEI6A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)).

On top of not reporting on unconstrained type parameters, this also stops reporting on type parameters with valid constraints like `unknown` or `any` (not part of the original issue); please let me know if this is OK.

```ts
// type parameter with no constraint
async function test1<T>(value: T) {
  return await value;
}

// type parameter with `any` or `unknown` constraint
async function test2<T extends unknown>(value: T) {
  return await value;
}

async function test3<T extends any>(value: T) {
  return await value;
}

// direct `any` or `unknown` type pass
async function test4(value: unknown) {
  return await value;
}

async function test5(value: any) {
  return await value;
}
```

### DIFF
--- a/packages/eslint-plugin/src/util/needsToBeAwaited.ts
+++ b/packages/eslint-plugin/src/util/needsToBeAwaited.ts
@@ -17,26 +17,25 @@ export function needsToBeAwaited(
   node: ts.Node,
   type: ts.Type,
 ): Awaitable {
+  // can't use `getConstrainedTypeAtLocation` directly since it's bugged for
+  // unconstrained generics.
+  const constrainedType = !tsutils.isTypeParameter(type)
+    ? type
+    : checker.getBaseConstraintOfType(type);
+
+  // unconstrained generic types should be treated as unknown
+  if (constrainedType == null) {
+    return Awaitable.May;
+  }
+
   // `any` and `unknown` types may need to be awaited
-  if (isTypeAnyType(type) || isTypeUnknownType(type)) {
+  if (isTypeAnyType(constrainedType) || isTypeUnknownType(constrainedType)) {
     return Awaitable.May;
   }
 
   // 'thenable' values should always be be awaited
-  if (tsutils.isThenableType(checker, node, type)) {
+  if (tsutils.isThenableType(checker, node, constrainedType)) {
     return Awaitable.Always;
-  }
-
-  // recurse into a type parameter constraint
-  if (tsutils.isTypeParameter(type)) {
-    const constraint = type.getConstraint();
-
-    // if there's no constraint, this may need to be awaited
-    if (!constraint) {
-      return Awaitable.May;
-    }
-
-    return needsToBeAwaited(checker, node, constraint);
   }
 
   // anything else should not be awaited
--- a/packages/eslint-plugin/src/util/needsToBeAwaited.ts
+++ b/packages/eslint-plugin/src/util/needsToBeAwaited.ts
@@ -1,0 +1,44 @@
+import type * as ts from 'typescript';
+
+import {
+  isTypeAnyType,
+  isTypeUnknownType,
+} from '@typescript-eslint/type-utils';
+import * as tsutils from 'ts-api-utils';
+
+export enum Awaitable {
+  Always,
+  Never,
+  May,
+}
+
+export function needsToBeAwaited(
+  checker: ts.TypeChecker,
+  node: ts.Node,
+  type: ts.Type,
+): Awaitable {
+  // `any` and `unknown` types may need to be awaited
+  if (isTypeAnyType(type) || isTypeUnknownType(type)) {
+    return Awaitable.May;
+  }
+
+  // 'thenable' values should always be be awaited
+  if (tsutils.isThenableType(checker, node, type)) {
+    return Awaitable.Always;
+  }
+
+  // recurse into a type parameter constraint
+  if (tsutils.isTypeParameter(type)) {
+    const constraint = type.getConstraint();
+
+    // if there's no constraint, this may need to be awaited
+    if (!constraint) {
+      return Awaitable.May;
+    }
+
+    return needsToBeAwaited(checker, node, constraint);
+  }
+
+  // anything else should not be awaited
+  return Awaitable.Never;
+}
--- a/packages/eslint-plugin/src/rules/await-thenable.ts
+++ b/packages/eslint-plugin/src/rules/await-thenable.ts
@@ -58,8 +58,9 @@ export default createRule<[], MessageId>({
         }
 
         const originalNode = services.esTreeNodeToTSNodeMap.get(node);
+        const certainty = needsToBeAwaited(checker, originalNode, type);
 
-        if (needsToBeAwaited(checker, originalNode, type) === Awaitable.Never) {
+        if (certainty === Awaitable.Never) {
           context.report({
             node,
             messageId: 'await',
--- a/packages/eslint-plugin/src/rules/await-thenable.ts
+++ b/packages/eslint-plugin/src/rules/await-thenable.ts
@@ -3,12 +3,14 @@ import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
 import * as tsutils from 'ts-api-utils';
 
 import {
+  Awaitable,
   createRule,
   getFixOrSuggest,
   getParserServices,
   isAwaitKeyword,
   isTypeAnyType,
   isTypeUnknownType,
+  needsToBeAwaited,
   nullThrows,
   NullThrowsReasons,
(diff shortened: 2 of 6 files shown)


JoshuaKGoldberg (dismissed):
Pacha from Emperor's New Groove giving a perfection sign with a hand. Caption: "PERFECTION!"

packages/eslint-plugin/src/util/needsToBeAwaited.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (!constraint) {
+      return Awaitable.May;
+    }
+
+    return needsToBeAwaited(checker, node, constraint);
+  }
+
+  // anything else should not be awaited
+  return Awaitable.Never;
+}
```
kirkwaiblinger, line 44: hmm. We _should_ be able to use `getConstrainedTypeAtLocation` in the caller to solve this, rather than dealing with the type parameter constraint here, but unfortunately that function... doesn't really work on unconstrained generics, since it doesn't return an unknown type. Filed https://github.com/microsoft/TypeScript/issues/60475.

That said, I think let's try to match the implementation here as closely as possible to `getConstrainedTypeAtLocation()`, though... I'm thinking like this?

```suggestion
export function needsToBeAwaited(
  checker: ts.TypeChecker,
  node: ts.Node,
  type: ts.Type,
): Awaitable {
  // can't use `getConstrainedTypeAtLocation` directly since it's bugged for
  // unconstrained generics.
  const constrainedType = !tsutils.isTypeParameter(type)
    ? type
    : checker.getBaseConstraintOfType(type);

  // unconstrained generic types should be treated as unknown
  if (constrainedType == null) {
    return Awaitable.May;
  }

  // `any` and `unknown` types may need to be awaited
  if (isTypeAnyType(constrainedType) || isTypeUnknownType(constrainedType)) {
    return Awaitable.May;
  }

  // 'thenable' values should always be be awaited
  if (tsutils.isThenableType(checker, node, constrainedType)) {
    return Awaitable.Always;
  }

  // anything else should not be awaited
  return Awaitable.Never;
}
```

Note that writing it without recursion (like `getConstrainedTypeAtLocation`) assumes that descending into a constraint is idempotent.... and, if it's not, then we better fix that for `getConstrainedTypeAtLocation` too! (I'm wondering if that's indeed the case for generics whose constraint is another generic 🤔 maybe a bugged test case can be devised for this and we can file a new issue to fix it in both places if so)

kirkwaiblinger (changes requested):
Just minor changes requested. Looks great!

packages/eslint-plugin/src/util/needsToBeAwaited.ts
ronami: Should I add a TODO for refactor this to use `getConstrainedTypeAtLocation()` once the `unknown` type is available on `ts.TypeChecker`?

kirkwaiblinger: I'd say no TODO comment, but just file a repo maintenance followup issue that's a catch-all for investigating edge cases around unconstrained generics in both `getConstrainedTypeAtLocation` and mention this discussion/`needsToBeAwaited`. There's a bit of bug-hunting I want to do with the current impl of `getConstrainedTypeAtLocation` too. 

Unfortunately even if TS adds `checker.getUnknownType()` we still have to wait until we drop support for the latest TS version that doesn't have it yet in order to actually use it, which is >= 2 years from now 🫤

bradzacher (approved):

PR #10315, opened 2024-11-09
fix(eslint-plugin): [explicit-module-boundary-types] and [explicit-function-return-type] don't report on `as const satisfies`

### LINKED ISSUE #10231: Bug: [explicit-module-boundary-types] `allowDirectConstAssertionInArrowFunctions` should not error when also using `satisfies`

### Playground Link

https://typescript-eslint.io/play/#ts=5.6.3&fileType=.ts&code=FAFwngDgpgBAYgewTAvDA3sGMBmSBcMAziAE4CWAdgOYA0WMARgIamGUCuAto1KfQF8A3MGBQAHhASkQMAMYJKJGM1QwAFAEpUAPg2ZseBIQBEcAPKWT9bCzYwAjACYAzIOZF5ikpqEwA9P4wADwAtKEwACoAFuSeAO7SANaeHjAS0HIgUAAmAHSiGdKyCkqyjGpauvoMRqYWVjZMrITObsACKp6lykTMIHE45FCeiAi%2BAUFhETFxMCDRpAjxqZTppEukBUA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6RAD2MbMroFsA9gBMEiWgCNBsJsICG0QrSKlUGANrhsORNGiDokADRbtwKHPjxBAdwAilaIjL4AwoKbJ8AQWTJd%2BJQeAJJM3nq2AGIyLkGe6GD4cIhgAL6mYAC6WumpQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
type Foo = {
  foo: string,
  bar: number,
};

export const a = () => ({
  foo: "FOOO",
  bar: 123,
}as const); // <-- This works as expected.

export const b = () => ({
  foo: "FOOO",
  bar: 123,
} as const satisfies Foo); // <-- This throws an error.
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/explicit-module-boundary-types": [
      "error",
      { "allowDirectConstAssertionInArrowFunctions": true }
    ]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

This concerns both `explicit-module-boundary-types` and `explicit-function-return-type` rules.

I have a situation where I want to use both `as const` followed by `satisfies`. With the rule currently as written as soon as you add a `satisfies` after an `as const` the rule will error. Leaving off the `satisfies` works as expected. 

In this scenario, the `satisfies Foo` provides extra type safety ensuring that `a satisfies Foo` but still keeping the ability to have explicit type literals for `foo: "FOOO"` and `bar: 123`. 

### Actual Result

The rule should not throw an error when using `as const satisfies Foo`

### Additional Info

_No response_

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10231

## Overview

This PR resolves #10231 and adjusts both `explicit-module-boundary-types` and `explicit-function-return-type` to not fail on the following when `allowDirectConstAssertionInArrowFunctions` is enabled:

```ts
type Foo = {
  foo: string,
  bar: number,
};

export const a = () => ({
  foo: "FOOO",
  bar: 123,
}as const); // <-- This works as expected.

export const b = () => ({
  foo: "FOOO",
  bar: 123,
} as const satisfies Foo); // <-- This unexpectedly causes the rules to report.
```

### DIFF
--- a/packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
+++ b/packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
@@ -176,15 +176,12 @@ function doesImmediatelyReturnFunctionExpression({
 /**
  * Checks if a function belongs to:
  * ```
- * () => ({ action: 'xxx' } as const)
+ * ({ action: 'xxx' } as const)
  * ```
  */
-function returnsConstAssertionDirectly(
-  node: TSESTree.ArrowFunctionExpression,
-): boolean {
-  const { body } = node;
-  if (isTypeAssertion(body)) {
-    const { typeAnnotation } = body;
+function isConstAssertion(node: TSESTree.Node): boolean {
+  if (isTypeAssertion(node)) {
+    const { typeAnnotation } = node;
     if (typeAnnotation.type === AST_NODE_TYPES.TSTypeReference) {
       const { typeName } = typeAnnotation;
       if (
@@ -256,11 +253,20 @@ function isValidFunctionExpressionReturnType(
   }
 
   // https://github.com/typescript-eslint/typescript-eslint/issues/653
-  return (
-    options.allowDirectConstAssertionInArrowFunctions === true &&
-    node.type === AST_NODE_TYPES.ArrowFunctionExpression &&
-    returnsConstAssertionDirectly(node)
-  );
+  if (
+    !options.allowDirectConstAssertionInArrowFunctions ||
+    node.type !== AST_NODE_TYPES.ArrowFunctionExpression
+  ) {
+    return false;
+  }
+
+  const { body } = node;
+
+  if (body.type === AST_NODE_TYPES.TSSatisfiesExpression) {
+    return isConstAssertion(body.expression);
+  }
+
+  return isConstAssertion(body);
 }
 
 /**
--- a/packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
+++ b/packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
@@ -260,11 +260,12 @@ function isValidFunctionExpressionReturnType(
     return false;
   }
 
-  const { body } = node;
+  let body = node.body;
+  while (body.type === AST_NODE_TYPES.TSSatisfiesExpression) {
+    body = body.expression;
+  }
 
-  return isConstAssertion(
-    body.type === AST_NODE_TYPES.TSSatisfiesExpression ? body.expression : body,
-  );
+  return isConstAssertion(body);
 }
 
 /**
--- a/packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
+++ b/packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
@@ -262,11 +262,9 @@ function isValidFunctionExpressionReturnType(
 
   const { body } = node;
 
-  if (body.type === AST_NODE_TYPES.TSSatisfiesExpression) {
-    return isConstAssertion(body.expression);
-  }
-
-  return isConstAssertion(body);
+  return isConstAssertion(
+    body.type === AST_NODE_TYPES.TSSatisfiesExpression ? body.expression : body,
+  );
 }
 
 /**
--- a/packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts
+++ b/packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts
@@ -449,9 +449,26 @@ interface R {
 }
 
 const func = (value: number) => ({ type: 'X', value }) as const satisfies R;
-const func = (value: number) => ({ type: 'X', value }) as const satisfies R;
-const func = (value: number) => x as const satisfies R;
-const func = (value: number) => x as const satisfies R;
+      `,
+      options: [
+        {
+          allowDirectConstAssertionInArrowFunctions: true,
+        },
+      ],
+    },
+    {
+      code: `
+const func = (value: number) => x as const satisfies number;
+      `,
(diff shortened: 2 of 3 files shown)


packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
```diff
@@ -256,11 +253,20 @@ function isValidFunctionExpressionReturnType(
...
+    return false;
+  }
+
+  const { body } = node;
+
+  if (body.type === AST_NODE_TYPES.TSSatisfiesExpression) {
+    return isConstAssertion(body.expression);
+  }
+
+  return isConstAssertion(body);
```
JoshuaKGoldberg, line 269: [Style] This is super nitpicky, but I found collapsing down to one statement to be a bit more readable:

```suggestion
  return isConstAssertion(
    body.type === AST_NODE_TYPES.TSSatisfiesExpression ? body.expression : body,
  );
```

No worries at all if you hate this 😄

packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts
JoshuaKGoldberg, line 1: [Testing] One thing under test per case please (https://typescript-eslint.io/contributing/pull-requests#granular-unit-tests)

JoshuaKGoldberg (changes requested):
Great! Just requesting changes on the tests, not the teeny nitpick.

Alec Baldwin as Jack Donaghy in 30 Rock enthusiastically giving two thumbs up. Caption: "GREAT!"

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
ronami: Not my cup of tea (🍵) but is is more concise 👍

packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts
ronami: I was relying on older tests that checked `as const`, I changed the ones I added to only test a single case each 👍

JoshuaKGoldberg (dismissed):
❤️‍🔥

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
```diff
@@ -256,10 +253,17 @@ function isValidFunctionExpressionReturnType(
...
+    node.type !== AST_NODE_TYPES.ArrowFunctionExpression
+  ) {
+    return false;
+  }
+
+  const { body } = node;
+
+  return isConstAssertion(
+    body.type === AST_NODE_TYPES.TSSatisfiesExpression ? body.expression : body,
   );
```
kirkwaiblinger, line 267: You might wanna use a `while` loop here, like we do in other rules for recursive syntaxes.
```suggestion
  let body = node.body;
  while (body.type === AST_NODE_TYPES.TSSatisfiesExpression) {
    body = body.expression;
  }

  return isConstAssertion(body);
```
Not sure if there's any real world use case but here's a test case 🤷 
```ts
type FooBar = {
  foo: string,
  bar: number,
};

type EmptyObjectType = {}

export const b = () => ({
  foo: "FOOO",
  bar: 123,
} as const satisfies FooBar satisfies EmptyObjectType); 
```

kirkwaiblinger (dismissed):

packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts
ronami: That's interesting, I didn't consider this edge case, thanks 👍

ronami: I added the necessary tests to cover this.

packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts
```diff
@@ -468,6 +468,27 @@ const func = (value: number) => x as const satisfies number;
     },
     {
       code: `
+const func = (value: number) => x as const satisfies 10 satisfies number;
```
kirkwaiblinger, line 471: fun fact, this code is illegal TS, since `as const` cannot be used on identifiers. Sometimes we intentionally write test cases with illegal TS, but lately we've generally been trying to avoid doing so (or we try to provide a `@ts-expect-error` comment to indicate we're intentionally testing illegal TS). Let's use an alternative here (and in the explicit-module-boundary test case too) that isn't an error since we can easily swap it out with one? See https://github.com/typescript-eslint/typescript-eslint/pull/10315#discussion_r1836146131 for a (surprisingly tedious to craft) example of a legal nested-`satisfies` `as const` pattern.

kirkwaiblinger (changes requested):

packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts
ronami: Sure, I've updated the tests I added to be valid TypeScript 👍

kirkwaiblinger (dismissed):

bradzacher (dismissed):

bradzacher (approved):

PR #10341, opened 2024-11-16
chore: update contributors

bradzacher:
## Overview

This hasn't been updated in over 8 months.
Fixes the script so it does pagination properly.
This likely still won't work as our CI job because we make un-authenticated requests and the rate limit for that is 60/hour.

We'll have to figure out a way around that at some point.
For now I just used a personal token to bypass the limit so we can update this.


JoshuaKGoldberg (approved):
Yikes, did not realize this was broken at all. Thank you Brad!!

Samatta20 (commented):
Help

PR #5847, opened 2022-10-18
[WIP] feat(eslint-plugin): [no-unused-type-properties] add rule

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1328
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/typescript-eslint/typescript-eslint/blob/main/CONTRIBUTING.md) were taken

Issue #6442, opened 2023-02-08
Rule proposal: disallow unused return types


KhafraDev:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+new+plugin+rule%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#supported-rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### My proposal is suitable for this project

- [X] My proposal specifically checks TypeScript syntax, or it proposes a check that requires type information to be accurate.
- [X] My proposal is not a "formatting rule"; meaning it does not just enforce how code is formatted (whitespace, brace placement, etc).
- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Description

Paired with [explicit-function-return-types](https://typescript-eslint.io/rules/explicit-function-return-type), it's possible to include extraneous return types that are never used. This can happen when refactoring code meaning that some code branches might become impossible to reach.

For example you might have this function:
```ts
function random (param?: string): string | null {
  if (typeof param !== 'string') return null
  
  return 'has param'
}
```

where the return types are correct, but then it's refactored to

```ts
// note that param is no longer optional but the return type is the same union
function random (param: string): string | null {
  return 'has param'
}
```

and let's assume it's being used in a codebase before param is required:

```ts
const rand = random(param)

if (rand === null) {
  throw new Error('...')
}
```

since the return type is unchanged, the null check branch might not be removed during refactoring, and neither the linter or tsc will complain about it.

### Fail Cases

```typescript
function test (): string | null {
  return 'string'
}

class A {
  static test (): string | null {
    return 'string'
  }

  test (): string | null {
    return 'string'
  }
}
```

### Pass Cases

```typescript
function test (): string {
  return 'string'
}

class A {
  static test (): string | null {
    if (Math.random() < 0.5) return null

    return 'string'
  }

  test (): string {
    return 'string'
  }
}
```

### Additional Info

It's totally possible I missed an issue that matched this proposal, but I couldn't find any.

JoshuaKGoldberg:
[...]

PR #10344, opened 2024-11-16
feat(eslint-plugin): new rule `no-unused-destructure-type-properties`

### LINKED ISSUE #1328: New rule suggestion: complete destructuring of parameters
Consider this code:

```ts
function Button({ text }: {
  text: string;
  onClick: () => void;
}) {
  return <button>{text}</button>;
}
```

The `onClick` prop was accidentally omitted, and no current rule is able to catch it (that I know of). I suggest a rule, called maybe `fully-destructure-arguments`, which would trigger if a property is present on an object type, but is not destructured.

It (probably) shouldn't trigger for interfaces et al, only immediate inline object types. An interface may be used elsewhere, but an inline object type is only used in one place, so if it's destructured and is missing a prop, the forgotten prop is guaranteed never to be used.

Bikeshedding is welcome on the rule name or the suggestion in general.

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1328

## Overview

This PR attempts to tackle #1328 and add a new rule to flag unused destructuring of inline objects and tuple types:

```ts
function test({
  used,
}: {
  // Unused property 'unused' in destructuring.
  unused: boolean;
  used: boolean;
}) {
  // ...
}
```

Some thoughts/things to note:

- I've drawn inspiration from @JoshuaKGoldberg's #5847 and used some of its functionality. I used code directly rather than attributing the original commits; please let me know if I should add them and push forcefully (I didn't want to attribute the commits unless it's OK with the author).
- The rule warns on tuple types in addition to object types. This isn't part of the original issue, but it made sense. Let me know if I should remove it:

  ```ts
  function test([a]: [
    boolean,
    // Unused key '1' in destructuring.
    number,
  ]) {
    // ...
  }
  ```

- The implementation is recursive, so nested inline object/tuple types are also flagged:

  ```ts
  function test({
    used1: { used2 },
  }: {
    used1: {
      used2: string;
      // Unused property 'unused' in destructuring.
      unused: number;
    };
  }) {
    // ...
  }
  ```

- I've used some heuristics to determine which fields of an object/array a dynamic property (one which `getStaticMemberAccessValue` failed to analyze statically) uses. I've relied on its type to narrow down which object keys it may access. For example, the following will flag `bazz` as unused:

  _I'll add that I'm not sure if it's important to cover this or bail if it happens. I'll be happy to hear some opinions on this._

  ```ts
  declare const c: 'bar' | 'foo';

  function test({
    [c]: a,
  }: {
    foo: string;
    bar: number;
    // Unused property 'bazz' in destructuring.
    bazz: string;
  }) {
    // ...
  }
  ```

- I've used a different heuristic to determine which index signatures the destructuring object uses. The following will flag `[i: number]: boolean;` as being unused:

  ```ts
  function test({
    used,
  }: {
    // Unused index signature '[number]' in destructuring.
    [i: number]: boolean;
    [i: string]: boolean;
  }) {
    // ...
  }
  ```

- I've used even another heuristic (🙈) to try and warn on unused template literal index signatures. The rule will be able to catch the following as unused index signature:

  ```ts
  function test({
    used,
  }: {
    used: string;
    // Unused index signature '[`a${string}`]' in destructuring.
    [i: `a${string}`]: string;
  }) {
    // ...
  }
  ```

  But will fail to catch this one as being unused:

  ```ts
  function test({
    a,
  }: {
    // Should report this as an unused index signature.
    [i: `b${string}`]: string;
    [i: `a${string}`]: string;
  }) {
    // ...
  }
  ```

- I changed `getStaticMemberAccessValue`'s return type to differentiate between a `number` key and a `string` one (currently, both are normalized to `string`). This is necessary to match a key to the correct index signature. I've updated the relevant rules that use it.
- I included a suggestion to remove an unused property/element. Having this as an auto-fix seems inconvenient, as you sometimes write the type before you destructure over it.
- I decided not to warn on edge cases like this, though it's possible to do so:

  ```ts
  function test({ 1: a }: [
    // Should report this as unused.
    string,
    number,
    // Should report this as unused.
    boolean,
  ]) {
    // ...
  }
  ```

- I also decided to avoid warning if an object destructure has a key that failed to be used as an index. TypeScript already shows an error for this, and I didn't want to add warnings that can confuse unnecessarily. For example, the following will not show lint errors from this rule (even though `world` is unused):

  ```ts
  function test1({
    // Property 'hello' does not exist on type '{ world: string; }'.
    hello,
  }: {
    world: string;
  }) {
    // ...
  }
  ```

  _What do you think? Should the rule warn on these cases?_

- Bike-shedding: I don't know the correct name for this (currently `no-partial-destructuring`), another option I was considering is `no-unused-destructure-elements`, will be happy to hear opinions on this.

I'll add that there were a lot of edge cases to cover with this rule, and I'm sure I missed some; please let me know if any of this makes sense 🚀


packages/eslint-plugin/docs/rules/no-partial-destructuring.mdx
JoshuaKGoldberg, line 1: > I used code directly rather than attributing the original commits; please let me know if I should add them and push forcefully (I didn't want to attribute the commits unless it's OK with the author).

No worries on preserving attribution for me. Whatever's easiest on your end is 👍 from me. My name is on enough git blames for this project as it is. 😄 Thanks for asking!

packages/eslint-plugin/docs/rules/no-partial-destructuring.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+const [hello, world]: [string, number] = arr;
+```
+
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+If your project uses destructuring where only partial values are relevant, this rule may not suit your needs. Consider turning it off.
```
JoshuaKGoldberg, line 39: > where only partial values are relevant

[Docs] Why? In what use case would that be reasonable? (asking to put in the docs, not just rhetorically)

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+      });
+    }
+
+    return {
+      "ArrayPattern[typeAnnotation.typeAnnotation.type='TSTupleType']"(
+        node: TSESTree.ArrayPattern & {
+          typeAnnotation: TSESTree.TSTypeAnnotation;
+        },
+      ): void {
+        checkParam(node, node.typeAnnotation.typeAnnotation);
```
JoshuaKGoldberg, line 436: [Refactor] The types will need to be updated, but this more direct call just works otherwise:

```suggestion
        checkTuple(node, node.typeAnnotation.typeAnnotation);
```

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+      },
+      "ObjectPattern[typeAnnotation.typeAnnotation.type='TSTypeLiteral']"(
+        node: TSESTree.ObjectPattern & {
+          typeAnnotation: TSESTree.TSTypeAnnotation;
+        },
+      ): void {
+        checkParam(node, node.typeAnnotation.typeAnnotation);
```
JoshuaKGoldberg, line 443: [Refactor] This one too:

```suggestion
        checkObject(node, node.typeAnnotation.typeAnnotation);
```

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+
+    function checkTuple(
+      param: TSESTree.ArrayPattern,
+      typeAnnotation: TSESTree.TSTupleType,
+    ): void {
+      for (const [index, member] of typeAnnotation.elementTypes.entries()) {
+        // bail on rest type `[...string[]]`
+        if (member.type === AST_NODE_TYPES.TSRestType) {
+          return;
+        }
```
JoshuaKGoldberg, line 130: [Bug] It's possible for a rest type to be unused. This rule should handle that.

```ts
function receive([value]: [string, ...string[]][]) {
```

https://deploy-preview-10344--typescript-eslint.netlify.app/play/#ts=5.6.2&fileType=.ts&code=GYVwdgxgLglg9mABAJwKYVTAbqgFAbSwEMAbEVAXQC5F8BnKZGMAcwBpEA6bhp1-ChQEBKRAG8AUIkTEyqANwSAvkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1uIENp9KPeLQAmKfHDL5YVJgHN0URNGgdokcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+
+        const property = param.elements.at(index);
+
+        if (property === undefined) {
+          reportOnMember(member, { type: 'key', key: String(index) });
+          continue;
+        }
+
+        if (
+          // bail in case of `[, ...]`
```
JoshuaKGoldberg, line 140: [Docs] Nit: the `...` looking like a rest/spread confused me 😅. Can a different placeholder be used? Maybe:

  ```suggestion
          // bail in case of `[,]` empty slots
```

File: packages/eslint-plugin/src/rules/no-partial-destructuring.ts [hunk 6]
```diff
@@ -0,0 +1,447 @@
...
+        const valueType = services.getTypeAtLocation(property.value);
+
+        // bail if misses; this is a type-error
+        if (tsutils.isIntrinsicErrorType(valueType)) {
+          return;
+        }
+
+        const memberKey = getStaticMemberAccessValue(property, context);
+
+        // collect dynamic keys which we failed to statically analyzed
```
JoshuaKGoldberg, line 96: [Docs] Nit:

```suggestion
        // collect dynamic keys which we failed to statically analyze
```

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+      // collect used properties
+      for (const property of node.properties) {
+        // bail on a rest element
+        if (property.type === AST_NODE_TYPES.RestElement) {
+          return;
+        }
+
+        const valueType = services.getTypeAtLocation(property.value);
+
+        // bail if misses; this is a type-error
```
JoshuaKGoldberg, line 89: [Docs] I got confused by this and had to play with the test case.

```suggestion
        // bail if the property doesn't have a known type (that's a type error)
```

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
JoshuaKGoldberg, line 1: [Feature] What do you think about also handling constituents of `Record` union keys? 

For example, I might expect the `"unused"` here to be reported on:

```ts
function test({ used }: Record<"used" | "unused", boolean>) {
  return used;
}
```

https://deploy-preview-10344--typescript-eslint.netlify.app/play/#ts=5.6.2&fileType=.ts&code=GYVwdgxgLglg9mABFApgZygCgN6JGlAE0QF8AuRAJRQjgCdCAeAInyOcQB9FWw3DmAGkQAjOHAA2KAIZgAfAEpE2AFCJEdFFBB0k-ANwqSQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1uIENp9KPeLQAmKfHDL5YVJgHN0URNGgdokcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+          indexParameterType,
+          dynamicProperties,
+          remainingProperties,
+        )
+      ) {
+        return;
+      }
+
+      // Template literal index signatures count as used if any remaining property's type is the
+      // same, this may miss some unused template literal index signatures
```
JoshuaKGoldberg, line 178: > this may miss some unused template literal index signatures

[Docs] What do you mean by this? (asking not just rhetorically, this should be in the docs if it's relevant)

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+        );
+
+        for (const type of tsutils.unionTypeParts(destructure.type)) {
+          if (tsutils.isStringLiteralType(type) && type.value === memberKey) {
+            return destructure.property;
+          }
+
+          if (tsutils.isNumberLiteralType(type) && type.value === memberKey) {
+            return destructure.property;
+          }
```
JoshuaKGoldberg, line 273: [Refactor] This could be condensed into a more dense `if((string || number) && ===`, or even better, switched to the more general API for less code:

```suggestion
          if (tsutils.isLiteralType(type) && type.value === memberKey) {
            return destructure.property;
          }
```

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
JoshuaKGoldberg, line 1: > I decided not to warn on edge cases like this, though it's possible to do so:
> 
> ```ts
> function test({ 1: a }: [
>   // Should report this as unused.
>   string,
>   number,
>   // Should report this as unused.
>   boolean,
> ]) {
>   // ...
> }
> ```

[Bug] Bad news, this is a legit pattern people occasionally use. It's super weird and I personally wouldn't be surprised if someone eventually suggested a lint rule against it. But it's valid TS and therefore this rule should support it IMO.

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+
+    function reportOnMember(
+      member: TSESTree.Node,
+      data: { key: string; type: string },
+    ): void {
+      context.report({
+        node: member,
+        messageId: 'partialDestructuring',
+        data,
+        suggest: [
```
JoshuaKGoldberg, line 412: > I included a suggestion to remove an unused property/element. Having this as an auto-fix seems inconvenient, as you sometimes write the type before you destructure over it.

[Feature] I'd personally prefer an autofix. There already exist a lot of rules that can inconveniently change WIP code. For example, [`@typescript-eslint/consistent-indexed-object-style`](https://typescript-eslint.io/rules/consistent-indexed-object-style) can autofix an index signature to a `Record` if you run it before adding another property to the index signature. Rules aren't intended to be run on WIP code.

Also: when I'm newly enabling this rule on a codebase with many violations, I would sure love an autofixer!

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
JoshuaKGoldberg, line 140: [Bug] There can still be an unused property even in the presence of empty slots:

```ts
function test([a, , b]: [boolean, number, object, string]) { }
//                                                ~~~~~~
```

https://deploy-preview-10344--typescript-eslint.netlify.app/play/#ts=5.6.2&fileType=.ts&code=GYVwdgxgLglg9mABFApgZygCgNoEMA0ihARgLoBci2xccANirmIWCALbEoBOhcxAVimiEMXGGADmpAJSIA3ogC%2BAKAD0qxJq3aduvfoNaAfidNA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1uIENp9KPeLQAmKfHDL5YVJgHN0URNGgdokcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+     */
+    function checkStaticMember(
+      member: TSESTree.TSMethodSignature | TSESTree.TSPropertySignature,
+      remainingProperties: RemainingProperties,
+      dynamicProperties: DynamicProperties,
+    ): void {
+      const memberKey = getStaticMemberAccessValue(member, context);
+
+      if (memberKey === undefined) {
+        return;
```
JoshuaKGoldberg, line 203: [Testing] Uncovered line, as reported by codecov. Either this is reachable and should have a test covering it, or it's not and either a refactor or type assertion should be used to eliminate the dead code.

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+     * @returns `true` if an index signature's value type matches an existing
+     * property's value type, `false` otherwise
+     */
+    function indexSignatureHasMatchingValue(
+      remainingProperties: RemainingProperties,
+      indexSignature: TSESTree.TSIndexSignature,
+    ): boolean {
+      // `{ used }: { [i: `_${string}`] }`
+      if (!indexSignature.typeAnnotation) {
+        return true;
```
JoshuaKGoldberg, line 355: [Testing] Uncovered line, as reported by codecov. Either this is reachable and should have a test covering it, or it's not and either a refactor or type assertion should be used to eliminate the dead code.

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
```diff
@@ -0,0 +1,447 @@
...
+    ): boolean {
+      if (isNoUncheckedIndexedAccess) {
+        const indexSignatureParts = tsutils
+          .unionTypeParts(indexSignatureType)
+          .filter(part => !isTypeFlagSet(part, ts.TypeFlags.Undefined));
+
+        const comparedTypeParts = tsutils.unionTypeParts(comparedType);
+
+        if (indexSignatureParts.length !== comparedTypeParts.length) {
+          return false;
```
JoshuaKGoldberg, line 391: [Testing] Uncovered line, as reported by codecov. Either this is reachable and should have a test covering it, or it's not and either a refactor or type assertion should be used to eliminate the dead code.

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
JoshuaKGoldberg, line 1: > Bike-shedding: I don't know the correct name for this (currently `no-partial-destructuring`), another option I was considering is `no-unused-destructure-elements`, will be happy to hear opinions on this.

[Docs] Good question... Maybe `no-unused-destructure-type-properties`? Kind of verbose, but more technically precise to what the rule does?

`no-partial-destructuring` makes it sound not specific to types. I think we'll want to avoid misconceptions like that. A lot of newcomers have a hard time mentally mapping between _runtime_ destructures and _type system_ destructures.

cc @typescript-eslint/triage-team, I'm out of ideas.

JoshuaKGoldberg, line 1: > heuristic (🙈) to try and warn on unused template literal index signatures ... But will fail to catch this one as being unused:
>
> ```ts
> function test({
>   a,
> }: {
>   // Should report this as an unused index signature.
>   [i: `b${string}`]: string;
>   [i: `a${string}`]: string;
> }) {
>   // ...
> }
>

I can see why this is difficult 😬 ... and also think this is a bug that should be avoided if possible. IIRC someone (you?) had worked with the `checker.isTypeAssignableTo` API recently for something like this, but don't quite recall. Do you think that might help?

JoshuaKGoldberg (changes requested):
Whoo! What a great rule. Really happy to see you pick this up and make it much much more comprehensive than my old feeble attempt 😂. Really great work!

Lots of comments, many of which are docs - the implementation generally looks great to me.

Cute cat going evil and saying 'power' with flames arising behind

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
kirkwaiblinger: Agreed that `no-partial-destructuring` is problematic since it makes me think it's gonna report on `const { a } = { a: 'foo', b: 'bar' }`.

I'm down with `no-unused-destructure-type-properties`.

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
ronami: Nice catch!

I think you've meant the following, in which the rest type is unused:

```diff
- function receive([value]: [string, ...string[]][]) {
+ function receive([value]: [string, ...string[]]) {
```

Please correct me if I misunderstood anything about this.

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
ronami: Interesting, I like it.

I think there is some similarity between this and my suggestion on https://github.com/typescript-eslint/typescript-eslint/issues/6442#issuecomment-2437185952. In my opinion, the rule should only warn on inline union types and skip referenced ones:

```ts
// should report
function test({ used }: Record<"used" | "unused", boolean>) {
  return used;
}

// potentially used elsewhere on the project
type K = "used" | "unused";

// shouldn't report
function test({ used }: Record<K, boolean>) {
  return used;
}
```

Does this make sense? Wdyt?

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
ronami: I'm convinced on this, thanks 👍

packages/eslint-plugin/src/rules/no-partial-destructuring.ts
JoshuaKGoldberg: Rick and Morty side character guy in business suit saying "YES!" confidently and snapping his fingers

ronami:
> 👋 @ronami are you waiting on anything from anyone on this PR?

Oh, I completely forgot about to get back to this. I'm moving this back to draft since it's taking me longer to finalize the PR.

Thanks 👍

ronami:

PR #10361, opened 2024-11-20
chore: update to latest nx and use project references first approach

### LINKED ISSUE #10581: Repo: Refactor TSConfigs to use project references
### Suggestion

As implemented in #9038 -> #10361: we're a monorepo and [TypeScript project references](https://www.typescriptlang.org/docs/handbook/project-references.html) are the canonical way to manage monorepo TS type checking.

### Additional Info

💖

JamesHenry:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10581


knip.ts
JoshuaKGoldberg, line 1: I suspect a lot of the includes in this file are due to https://github.com/webpro-nl/knip/issues/873. Once that's resolved & released, if they're still there, we should file an issue over on Knip.

tsconfig.base.json
```diff
@@ -1,21 +1,32 @@
...
+    "removeComments": false,
+    "resolveJsonModule": false,
+    "skipDefaultLibCheck": false,
     "skipLibCheck": true,
-    "sourceMap": true,
+    "sourceMap": false,
     "strict": true,
-    "target": "ES2022"
+    "target": "es2022",
+    "verbatimModuleSyntax": false
```
maschwenk, line 30: Just as an outside observer who is considering a similar change, curious around the rationale of a bunch of these changes. Can make some guesses:

`allowJs` -- presume this is just some standardization
`composite` -- just pulling this up into defaults
`emitDeclarationOnly` -- we don't need to emit JS because were now using a single tsc command at the top of the repo to build all the javascript. need the `.d.ts` to be able to do references though
`module`/`moduleResolution` -- curious what necessitated this?
`skipDefaultLibCheck` -- doesn't this mean you will be checking Typescript's own `d.ts.` files? I always thought pretty safe to turn on both `skipLibCheck` and  `skipDefaultLibCheck`
`sourceMap` -- presume this is because the composite packages do not need source mappings because they are not emitting js
`verbatimModuleSyntax` -- unsure about this one?

packages/ast-spec/tsconfig.build.json
```diff
@@ -1,12 +1,19 @@
...
   "compilerOptions": {
-    "composite": true,
-    "jsx": "preserve",
-    "outDir": "./dist",
-    "rootDir": "./src",
-    "resolveJsonModule": true
+    "baseUrl": ".",
+    "rootDir": "src",
+    "outDir": "dist",
+    "tsBuildInfoFile": "dist/tsconfig.build.tsbuildinfo",
```
maschwenk, line 7:
```suggestion
```
could possibly reduce some boilerplate by putting this:

```json
    "outDir": "${configDir}/dist",
    "tsBuildInfoFile": "${configDir}/dist/tsconfig.build.tsbuildinfo",
```
into

`../../tsconfig.base.json`

.github/workflows/ci.yml
```diff
@@ -171,7 +171,6 @@ jobs:
             'eslint-plugin-internal',
             'parser',
             'rule-tester',
-            'rule-schema-to-typescript-types',
```
leosvelperez, line 174: This project doesn't have a `test` target anymore. It didn't contain any spec files, so the `jest.config.js` file was removed.

JamesHenry:
Ready to go 🎉 @typescript-eslint/triage-team

JoshuaKGoldberg (approved):
LGTM! I don't feel confident enough in these doings to respond to the open community questions, but did try it out a bunch locally and enjoy it. Thanks for the overhaul! 🚀

eslint.config.mjs
```diff
@@ -36,7 +36,6 @@ export default tseslint.config(
     plugins: {
       ['@typescript-eslint']: tseslint.plugin,
       ['@typescript-eslint/internal']: tseslintInternalPlugin,
-      ['eslint-comments']: eslintCommentsPlugin,
```
kirkwaiblinger, line 39: why remove this declaration? We generally have all the plugins "forward-declared", even if not strictly necessary (including our own `@typescript-eslint`).

packages/website-eslint/tsconfig.build.json
kirkwaiblinger, line 1: What's the purpose of this file? It's a `tsconfig.build.json`, but it isn't used in the esbuild build? Also it doesn't have `checkJs: true`, but all the files in this package are JS files... Is that intentional? I'm not super knowledgeable in this area so just asking some questions 🙏

packages/utils/tsconfig.json
kirkwaiblinger, line 1: Curious, here and elsewhere, why do we have project references to other packages for these package-root "solution" tsconfigs that don't include any files themselves? Are the transitive references in the `*.build.json`/`*.spec.json` not sufficient?

tsconfig.base.json
kirkwaiblinger, line 30: I'd think we might as well remove most of the options that we explicitly set to their defaults? Like `verbatimModuleSyntax: false`, `removeComments: false`, `experimentalDecorators: false`, for example, and other niche options?

tsconfig.eslint.json
kirkwaiblinger, line 1: should this be named more like `tsconfig.repo-config-files.json` or something? Seems to apply to more than just our `eslint.config.m?js`...

kirkwaiblinger (commented):
Left a few questions - I'm not terribly familiar with nx or ts-in-the-large structure, so I'm largely asking to have a record of the answer to some questions, not at all implying a question indicates a problem. 

It would also be lovely if we could get a PR description that gives an overview of the changes 🙏

eslint.config.mjs
JamesHenry: I recommend you potentially looking into this as a follow up. There is no typed plugin exposed by the up to date community version of the package (which this PR introduces), so the only way to even attempt to wire up something like this is to add a bunch of lint and type assertion comments.

See https://eslint-community.github.io/eslint-plugin-eslint-comments/ for the up to date docs

packages/website-eslint/tsconfig.build.json
JamesHenry: It's just keeps things consistent and avoids a more complex Nx plugin configuration. You are totally right about missing checkJs though, thanks, I've added that in and fixed up the missing esquery types that it unearthed (meaning we've never been type checking this stuff correctly!)

JamesHenry: (If we got rid of the `tsconfig.build.json` we'd have to put everything in the `tsconfig.json` and have an excludes setup up on the inference of the build target, and a separate plugin entry to only infer the typecheck target for website-eslint. All doable, but just adds extra stuff)

eslint.config.mjs
JamesHenry: Oh just noticed we can remove the `eslint.config.js` workaround now too :) we only need the `eslint.config.mjs` file at the root

tsconfig.eslint.json
JamesHenry: This is a long-standing convention/recommendation in typescript-eslint - it's to do with the fact that the config is there for linting purposes, rather than to imply it is only checking linting related files

JamesHenry: I've updated it

packages/utils/tsconfig.json
JamesHenry: This was what came out of our discussions with the TS team and various POCs. Nx automatically keeps these references in sync based on actual usage as well.

However, if you want to do some experimenting post merge to show that they are no needed, then that would be interesting to focus on in isolation for sure

JamesHenry:
Let's go 🎉 

We can address any remaining minor tweaks in small follow ups.

Thanks again to @leosvelperez for his help on this, and to @jakebailey @bradzacher @JoshuaKGoldberg @kirkwaiblinger for reviews

PR #10372, opened 2024-11-23
feat: support TypeScript 5.7

### LINKED ISSUE #10115: TypeScript 5.7 Support
https://devblogs.microsoft.com/typescript/announcing-typescript-5-7-beta/

_Please be patient._

---

# 🔲 Support for --target es2024 and --lib es2024

We'll need to update the [`getLib` function](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/parser/src/parser.ts#L46-L77).

# 🔲 Search Ancestor Configuration Files for Project Ownership

# 🔲 lib.d.ts Updates

- Checks for Never-Initialized Variables
- Path Rewriting for Relative Paths
- Faster Project Ownership Checks in Editors for Composite Projects
- Validated JSON Imports in --module nodenext
- Support for V8 Compile Caching in Node.js
- TypedArrays Are Now Generic Over ArrayBufferLike
- Creating Index Signatures from Non-Literal Method Names in Classes
- More Implicit any Errors on Functions Returning null and undefined

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10115

## Overview

Updates lib and scope definitions.

Doesn't update `project: true` to match the new ancestor configuration file searching, as that should already be covered by `projectService: true`.

💖


bradzacher (approved):

packages/types/src/lib.ts
```diff
@@ -71,13 +72,21 @@ type Lib =
   | 'es2022.intl'
   | 'es2022.object'
   | 'es2022.regexp'
-  | 'es2022.sharedmemory'
```
burtek, line 74: Newest typescript might have removed this, but the plugin should be compatible with previous TS versions (as per range in package.json) and those previous versions do know what `es2022.sharedmemory` is. This change makes this eslint plugin incompatible with previous TS versions in projects declaring use of `es2022.sharedmemory`

JoshuaKGoldberg: Interesting. Could you please file an issue with a reproduction?

burtek: Sure, on it

burtek: https://github.com/typescript-eslint/typescript-eslint/issues/10395

burtek: Actually, it's not this change, it's the rename of the lib file. Still caused by this PR

PR #10374, opened 2024-11-23
fix(eslint-plugin): [no-deprecated] check if a JSX attribute is deprecated

### LINKED ISSUE #10275: Bug: [no-deprecated] No warnings about usage of deprecated props in JSX components

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.4&fileType=.tsx&code=JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwCwAUI8AHYxZQF5ZwAKxYdAG9GcOAHoECUWKQABACZYwlXMnYKZicTKUqsajfwhgAXHFQworAOb0GAX0aNcEFpbgBhEmcxUYAHQAYp4APMaCAHxwALxwABRCcHqq6lgKEXAOAJSx0fEyoQrAAG4o1sgAtDZQyABGdemRWnAAEsCF4sUlzQzZ9i5uHsgsEDAAFhwAogAeyOAANjxxod7gycqpRgIxAEQQ43CjuxKRNEA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNFjpEZAIb5E3dFADuQ6E0jgwAXxCKgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
import * as React from 'react';

interface Props {
  /**
   * @deprecated
   */
  deprecatedProp: string;
}

const Comp: React.FC<Props> = ({ deprecatedProp }) => (
  <div aria-grabbed>
    Hi
  </div>
);

const anotherExample = <Comp deprecatedProp="oh no" />;
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/no-deprecated": "warn"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expect to see a warning for the line that uses 'aria-dragged' and for the last line.

### Actual Result

No warnings, though TypeScript does recognize the deprecation.

### Additional Info

fwiw, seems like this was an issue in eslint-plugin-deprecated too. Someone tried to solve this separately (I haven't checked if it works): https://github.com/sebakerckhof/eslint-plugin-deprecated-jsx-props

mdm317:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10275

## Overview

Added a logic to check deprecated attributes from JSDoc when used in JSX.


JoshuaKGoldberg (changes requested):
Codecov is letting us know that a couple areas of code aren't covered by unit tests. For each, one of the following is true:

* The code is reachable, so a unit test should exercise it
* The code can be refactored to not include that case (e.g. passing a narrowed value around instead of re-checking something)
* This is some edge case in TypeScript-land and a type assertion should be used (instead of a runtime check)

I can re-review once the test coverage gaps are gone. Cheers!

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -271,11 +271,32 @@ export default createRule({
       );
     }
 
+    function getJSXAttributeDeprecation(
+      openingElement: TSESTree.JSXOpeningElement,
+      node: IdentifierLike,
+    ): string | undefined {
+      if (openingElement.name.type !== AST_NODE_TYPES.JSXIdentifier) {
```
Josh-Cena, line 278: Why this? Do we not want to check `<Tab.Item>`?

mdm317: I didn't consider the cases where <Tab.Item> is used. Thank you for pointing that out!

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -271,11 +271,32 @@ export default createRule({
...
+    function getJSXAttributeDeprecation(
+      openingElement: TSESTree.JSXOpeningElement,
+      propertyName: string,
+    ): string | undefined {
+      const tsNode = services.esTreeNodeToTSNodeMap.get(openingElement.name);
+
+      const contextualType = nullThrows(
+        checker.getContextualType(tsNode as ts.Expression),
+        'Expected jsx opening element name to have contextualType',
+      );
```
JoshuaKGoldberg, line 283: [Testing] _Most_ of the time, when a `nullThrows` is added for a type retrieval, it's masking a situation where the type legitimately could be nullish. But I tried locally and couldn't break this one! 🤷 

In this case, if the contextual type can't be found, then TypeScript gives back the intrinsic error type: `{ flags: 1, intrinsicName: "error", ... }`. So that's fine.

To be thorough, could you please add some tests for this case?

* The namespace being `any` `unknown`, and instrinsic error
* The attribute being `any` `unknown`, and instrinsic error

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -271,11 +271,32 @@ export default createRule({
...
 
+    function getJSXAttributeDeprecation(
+      openingElement: TSESTree.JSXOpeningElement,
+      propertyName: string,
+    ): string | undefined {
+      const tsNode = services.esTreeNodeToTSNodeMap.get(openingElement.name);
+
+      const contextualType = nullThrows(
+        checker.getContextualType(tsNode as ts.Expression),
+        'Expected jsx opening element name to have contextualType',
```
JoshuaKGoldberg, line 282: [Style] Nit:

```suggestion
        'Expected JSX opening element name to have contextualType',
```

JoshuaKGoldberg (changes requested):
Implementation looks great to me, very nicely done! 👏 

Just requesting changes on a bit more tests, since there's the tricky `nullThrows`. Thanks!

JoshuaKGoldberg (approved):
Glorious! Thank you! 🙌

PR #10378, opened 2024-11-23
POC - using assignability APIs in no-unnecessary-condition

kirkwaiblinger:
## PR Checklist

## Overview

This fundamentally changes the truthiness/falsiness algorithm used in no-unnecessary-condition. Up until now, the rule has checked various heuristics for possible falsiness: essentially whether the type is a known falsy literal and whether the type has TS's (inconsistent) PossiblyFalsy flag set. The proposed change is to use the assignability API to simply check whether one of JS's few falsy types is assignable to the type being investigated.

This has a few interesting consequences that improve correctness....
```ts
// was flagged, no longer flagged
function usesEmptyObject(x: {}) {
  if (x) {
    console.log('truthy case')
  } else {
    console.log('truthy case')
  }
}
// because 
usesEmptyObject(false);

// was flagged, no longer flagged
function usesNumberyObject(x: { toFixed: () => string }) {
  if (x) {
    console.log('truthy case')
  } else {
    console.log('truthy case')
  }
}
// because 
usesNumberyObject(0);
```
Annoyingly, this also has the drawback that the following no longer flags.
```ts
const neverNull = {};
if (neverNull) {
}
```
That's due to TS's (questionable) decision to give `neverNull` type `{}` instead of type `object`, even when initializing a `const` variable.

The following does still flag 
```ts
const neverNull: object = {};
if (neverNull) {
}
```

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
@@ -107,7 +107,7 @@ function assert(condition: unknown): asserts condition {
 
 assert(false); // Unnecessary; condition is always falsy.
 
-const neverNull = {};
+const neverNull = { someProperty: 'someValue' };
 assert(neverNull); // Unnecessary; condition is always truthy.
 
 function isString(value: unknown): value is string {
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -41,41 +41,32 @@ const getValueOfLiteralType = (
   return type.value;
 };
 
-const isFalsyBigInt = (type: ts.Type): boolean => {
-  return (
-    tsutils.isLiteralType(type) &&
-    valueIsPseudoBigInt(type.value) &&
-    !getValueOfLiteralType(type)
-  );
-};
-const isTruthyLiteral = (type: ts.Type): boolean =>
-  tsutils.isTrueLiteralType(type) ||
-  (type.isLiteral() && !!getValueOfLiteralType(type));
-
-const isPossiblyFalsy = (type: ts.Type): boolean =>
+const isPossiblyFalsy = (
+  checker: ts.TypeChecker,
+  type: ts.Type,
+  falsyTypes: ts.Type[],
+): boolean =>
   tsutils
     .unionTypeParts(type)
-    // Intersections like `string & {}` can also be possibly falsy,
-    // requiring us to look into the intersection.
     .flatMap(type => tsutils.intersectionTypeParts(type))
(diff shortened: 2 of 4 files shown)


packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
```diff
@@ -107,7 +107,7 @@ function assert(condition: unknown): asserts condition {
 
 assert(false); // Unnecessary; condition is always falsy.
 
-const neverNull = {};
+const neverNull = { someProperty: 'someValue' };
```
JoshuaKGoldberg, line 110: > That's due to TS's (questionable?) decision to give `neverNull` type `{}` instead of type `object`, even when initializing a `const` variable.

😩 this again... is there a TypeScript issue filed that we can reference? It feels like a bug to me.

I also don't mind it very much. It's not often that folks use `{}`. This docs change kind of makes it more realistic to me.

JoshuaKGoldberg (commented):
I like this direction! Nicely done 🙂

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
kirkwaiblinger: > 😩 this again... is there a TypeScript issue filed that we can reference? It feels like a bug to me.

That's a good question. I'll take an action item to investigate this.

> I also don't mind it very much. It's not often that folks use `{}`. This docs change kind of makes it more realistic to me.

Agreed that the docs change is probably an improvement. But, the false negative here makes the rule a little less understandable to me, so I'd lean on the side of wanting it to be correct if we can get it to be.

kirkwaiblinger: Oh, dear, the moment I start playing around with this, I immediately find infuriating behavior around `{}`....
```ts
function takesObject(x: object) {
  if (x) {
    console.log('should always be true')
  } else {
    console.error('this should be unreachable');
  }
}
const o: {} = 0;
// no error here!?!?!?!?
takesObject(o);
```

kirkwaiblinger: Regarding `const o = {}; // has type {}`, started discord conversation at https://discord.com/channels/508357248330760243/508357638677856287/1310419206260654081.

Regarding `{}` -> `object` assignments, filed https://github.com/microsoft/TypeScript/issues/60582... and immediately learned this is a duplicate of https://github.com/microsoft/TypeScript/issues/56205, in which this behavior is deemed intentional 🫤.

Reminder to self: I'm pretty sure we'll have some good info to add to the motivation for https://typescript-eslint.io/rules/no-empty-object-type/ as an outcome of these discussions 😆

kirkwaiblinger:
This POC has served its purpose as a POC, so closing.

PR #10385, opened 2024-11-25
fix(typescript-estree): update conditions for unsupported version warning

### LINKED ISSUE #10115: TypeScript 5.7 Support

inga-lovinde:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10115 

## Overview

I don't really know what I'm doing (this is my first PR to this repository), but I think that in https://github.com/typescript-eslint/typescript-eslint/pull/10372 , the update to version list in `typescript-estree` might have slipped through, `typescript-estree` was still declaring that it only supports TS versions below 5.7.0 (so users of TS 5.7 are currently still getting warnings about unsupported TS version even on the latest typescript-eslint at 8.15.1-alpha.6).

Seeing how for previous TS versions, const SUPPORTED_TYPESCRIPT_VERSIONS was updated in the same "feat: support TypeScript X.Y" pull requests as everything else, it seems that it should have went there, but it didn't, so I'm creating a separate PR for it.

I also took a liberty of upgrading the lowest supported version from 4.7.4 to 4.8.4, because the comment says "This needs to be kept in sync with package.json in the typescript-eslint monorepo", and lowest supported `typescript` version in package.json in the typescript-eslint monorepo was updated from 4.7.4 to 4.8.4 in #8973 but didn't make its way to `warnAboutTSVersion.ts`.

### DIFF
--- a/packages/typescript-estree/src/parseSettings/warnAboutTSVersion.ts
+++ b/packages/typescript-estree/src/parseSettings/warnAboutTSVersion.ts
@@ -6,7 +6,7 @@ import type { ParseSettings } from './index';
 /**
  * This needs to be kept in sync with package.json in the typescript-eslint monorepo
  */
-const SUPPORTED_TYPESCRIPT_VERSIONS = '>=4.7.4 <5.7.0';
+const SUPPORTED_TYPESCRIPT_VERSIONS = '>=4.8.4 <5.8.0';
 
 /*
  * The semver package will ignore prerelease ranges, and we don't want to explicitly document every one


JoshuaKGoldberg (approved):
💡 Aha! Thank you for the catch. Clearly this file keeps flying under the radar. Much appreciated @inga-lovinde!

inga-lovinde:
@JoshuaKGoldberg , since this file has to be maintained manually, maybe there could be a test catching the inconsistency between the main package.json and the range defined in this file?

That would require updating `warnAboutTSVersion.ts` to expose `SUPPORTED_TYPESCRIPT_VERSIONS` though (right now it only exposes `warnAboutTSVersion` which is not very useful for testing)...

JoshuaKGoldberg:
That's a good idea, I like it. Would you be open to filing a followup issue for that please? 🙏 

I ask because we're going to publish our weekly major in about 15 hours and I don't know that we'd be able to get a nontrivial test like that approved in time.

inga-lovinde:
I can try to open an issue tomorrow (it's 3 AM here), maybe also create a PR, after this one is merged :)

bradzacher:
goodcatch

bradzacher:
test for this: #10386

PR #10419, opened 2024-11-29
chore: enforce repo nullish check style

### LINKED ISSUE #10186: Repo: internal lint rule for our code standards around `=== nullishLiteral`
### Suggestion

We usually (intentionally) conflate `null` and `undefined` in this repo, and use `== null` or `== undefined` instead of `=== null` or  `=== undefined`. This pattern is generally adhered to pretty consistently, but it's surprising for some contributors, (ref, eg, https://github.com/typescript-eslint/typescript-eslint/pull/10003), and there are plenty of examples of deviations.

I propose let's lint for equality comparisons to nullish literals to enforce that only `== null`/`!= null` are used. We can autofix `== undefined` to `== null` and _suggest_ when a strict equality comparison is used, which might sometimes need to be disabled in rare cases.

Presumably it would be preferable if [eqeqeq](https://eslint.org/docs/latest/rules/eqeqeq) had a configuration to support this, but... it doesn't. And it's super trivial to write one ourself in this case.

### Additional Info

_No response_

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10186 

## Overview

adds internal lint rule eqeq-null to enforce that you do `== null` instead of `=== null` or `=== undefined`.

erred on the side of disabling reports in the repo if it seemed suspicious whether it could safely be disabled

### DIFF
--- a/packages/ast-spec/tests/util/serializers/Node.ts
+++ b/packages/ast-spec/tests/util/serializers/Node.ts
@@ -63,6 +63,7 @@ const serializer: NewPlugin = {
 
     for (const key of keys) {
       const value = node[key];
+      // eslint-disable-next-line @typescript-eslint/internal/eqeq-nullish -- I don't know whether this is safe to fix
       if (value === undefined) {
         continue;
       }
--- a/packages/ast-spec/tests/util/serializers/Node.ts
+++ b/packages/ast-spec/tests/util/serializers/Node.ts
@@ -63,7 +63,7 @@ const serializer: NewPlugin = {
 
     for (const key of keys) {
       const value = node[key];
-      // eslint-disable-next-line @typescript-eslint/internal/eqeq-nullish -- I don't know whether this is safe to fix
+      // eslint-disable-next-line @typescript-eslint/internal/eqeq-nullish -- intentional strict equality
       if (value === undefined) {
         continue;
       }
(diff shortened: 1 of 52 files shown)


packages/ast-spec/tests/util/serializers/Node.ts
```diff
@@ -63,6 +63,7 @@ const serializer: NewPlugin = {
 
     for (const key of keys) {
       const value = node[key];
+      // eslint-disable-next-line @typescript-eslint/internal/eqeq-nullish -- I don't know whether this is safe to fix
```
bradzacher, line 66: It's not - it would change the snaps
You can try it to confirm!

packages/eslint-plugin/src/rules/no-restricted-types.ts
```diff
@@ -145,6 +145,7 @@ export default createRule<Options, MessageIds>({
     ): void {
       const bannedType = bannedTypes.get(name);
 
+      // eslint-disable-next-line @typescript-eslint/internal/eqeq-nullish -- unsure if this is safe to fix
```
bradzacher, line 148: This is safe as the schema disallows null

bradzacher (dismissed):

Issue #10165, opened 2024-10-17
Configs: [return-await] Change default option from in-try-catch to always


Zamiell:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22preset+config+change%22+) and found none that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Description

In the documentation for the `@typescript-eslint/return-await` rule, it says that using "in-try-catch" as opposed to "always" is purely a stylistic concern.

However, on [this mdn page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#improving_stack_trace), it explains in the "Improving stack trace" section that using the "in-try-catch" style results in a worse stack trace.

Is this document accurate? If so, this is a compelling reason to use "always" instead of "in-try-catch", and I propose that the default option should be updated for all users of TSESLint.

kirkwaiblinger:
I'm +1, but note

- the reason for it being `in-try-catch` in the first place was historically due to the rule's association with the eslint rule `no-return-await`. See also https://github.com/typescript-eslint/typescript-eslint/issues/9517, https://github.com/typescript-eslint/typescript-eslint/issues/9030, https://github.com/eslint/eslint/issues/17345, etc.

- last time we chatted about this, not all typescript-eslint team members agreed on `always` being preferable to `in-try-catch`. This partially contributed to `error-handling-correctness-only` being created as a compromise to put the rule in the strict config, see https://github.com/typescript-eslint/typescript-eslint/pull/9364.

- even if we find alignment that `always` is preferable, we'd have to agree it's preferable enough to justify a(n) (autofixable) breaking change (I do hold this opinion, but much less strongly now that actual unsafety problems with `in-try-catch` have been resolved, see https://github.com/typescript-eslint/typescript-eslint/pull/9031 and https://github.com/typescript-eslint/typescript-eslint/pull/9044).

JoshuaKGoldberg:
🤔 It is kind of unfortunate that the rule has two different areas of use:

* Logical: error-handling contexts
* Stylistic: ordinary contexts

I wonder if should split the rule in two, so that the logical use could go into `recommendedTypeChecked` or `strictTypeChecked` and the stylistic use in `stylisticTypeChecked`...

[...]

PR #10421, opened 2024-11-29
docs: [return-await] make the rule no longer an extension of ESLint no-return-await

### LINKED ISSUE #9517: Docs: treat return-await as a standalone rule, not an extension

### Suggested Changes

1. First paragraph should no longer say that this is an extension rule. 
2. Instead, an info box or similar should be created to mention that this _used to be_ an extension rule, and link an issue or two on context (as well as a link to the base rule)
3. Options should _not_ say "See [eslint/no-return-await options](https://eslint.org/docs/rules/no-return-await#options)". There are no options for the base rule and (I'm pretty sure) there never were any.

### Affected URL(s)

http://typescript-eslint.io/rules/return-await

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9517


ArnaudBarre:
Do you plan to also update the documentation to advice on using always and remove the fact that in non error context this is only stylistic?

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger, line 1: responding to @ArnaudBarre 

> Do you plan to also update the documentation to advice on using always and remove the fact that in non error context this is only stylistic?

Not within this PR, unfortunately. I'm fully on board with that goal, but last I recall the team were not all in agreement on this. Therefore, my goal with this PR is pretty constrained to just removing the connection to `no-return-await`, which is agreed-upon.

I would invite you to check if there is an open issue already addressing your question, and, if so, engage on it or give it a 👍, or to open one up if there isn't one already!

ArnaudBarre: Yep there is: https://github.com/typescript-eslint/typescript-eslint/issues/10165!

packages/eslint-plugin/docs/rules/return-await.mdx
```diff
@@ -9,15 +9,46 @@ import TabItem from '@theme/TabItem';
...
+  return Promise.resolve(42);
+}
+bar().then(x => console.log(x)); // logs 42
+
+async function baz(): Promise<number> {
+  // value, since `await Promise.resolve(42)` is 42
+  return await Promise.resolve(42);
+}
+baz().then(x => console.log(x)); // logs 42
+```
```
JoshuaKGoldberg, line 32: Here's how the new page looks on my monitor:

Screenshot of the new docs page with the Netlify Drawer. The last fully visible sentence is the 'Returning a value rather than ...' one just after the code block.

Not much explanation above the fold, but a lot of code block. I think that's a tl;dr for many users.

Suggestion: what do you think about moving the plain text descriptions above the code block?

JoshuaKGoldberg (changes requested):
The phrasing all looks great to me! Just requesting changes on optimizing the 'above-the-fold' part of the docs.

Zamiell:
> Do you plan to also update the documentation to advice on using always and remove the fact that in non error context this is only stylistic?

Here's the related issue for changing the default value: https://github.com/typescript-eslint/typescript-eslint/issues/10165

packages/eslint-plugin/docs/rules/return-await.mdx
kirkwaiblinger: I'm not sure exactly what you mean by "moving the plain text descriptions above the code block"... 

However, I removed the code block for now, and I think it reads ok. What do you think?

JoshuaKGoldberg: That's what I meant yeah - moving more of the text explanations above the code block. 👍

JoshuaKGoldberg (approved):
Yay, happy to see this rule finally get the standalone emphasis it deserves! ❤️

PR #9560, opened 2024-07-14
feat: stricter parent types for the AST

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6225
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Adds all the parent types specified in #6225 and fixes the resulting type and linting errors.
Pretty straightforward here!

Issue #4586, opened 2022-02-22
[explicit-module-boundary-types] New option to allow unspecified return type if there are overload signatures


Josh-Cena:
- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.
- [x] I have [read the FAQ](https://typescript-eslint.io/docs/linting/troubleshooting) and my problem is not listed.

**Repro**

[Playground](https://typescript-eslint.io/play/#ts=4.5.2&sourceType=module&code=KYDwDg9gTgLgBAMwK4DsDGMCWEWIhACgEMAuOAZxikxQHMBKMy6ugbgChRJZFUNtcCfMTIokAWwBGwKIzhipMjl2jxk6LDjyFScVAGsUEAO4p6cAN7s4NuFGAwkUXEQ4BfdkA&rules=N4IgAgLgngDgpgZwMYCcCWMIFpEBs0B2EA9HAB4z5JrYC2A9gCYCuucWARvcwYwIYooWaPAQgAXCDgoU9FCAC+QA&tsConfig=N4XyA)

```JSON
{
  "rules": {
    "@typescript-eslint/explicit-module-boundary-types": "error"
  }
}
```

```TS
export function foo(a: string): string;
export function foo(a: number): number;
export function foo(a: unknown) {
  return a;
}
```

**Expected Result**

I'd like a new option that makes this code pass, since the implementation signature is not visible to the outside, while the _boundary_, defined by the overload signatures, is well-typed.

**Actual Result**

This errors, even with all the `allow*` options turned on.

**Versions**

| package                            | version |
| ---------------------------------- | ------- |
| `@typescript-eslint/eslint-plugin` | `5.12.1` |
| `@typescript-eslint/parser`        | `5.12.1` |
| `TypeScript`                       | `4.5.5` |
| `ESLint`                           | `8.9.0` |
| `node`                             | `17.5.0` |

bradzacher:
I don't think that this is a good idea because it can lead you to writing a bad implementation.

Your example shows one of the major type-safety holes - an implementation that returns `unknown` is considered compatible with all strictly typed signatures.

But we know that isn't safe or correct because `unknown` is not compatible. For example

```ts
export function foo(a: string): string;
export function foo(a: number): number;
export function foo(a: unknown) {
  return (true as unknown);
}
```
[...]

PR #10422, opened 2024-11-29
fix(eslint-plugin): [no-invalid-void-type] don't flag function overloading with `void` in return type annotation

### LINKED ISSUE #5752: Bug: [no-invalid-void-type] Can't use `void` in return type union for overloaded function

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.4&sourceType=module&code=GYVwdgxgLglg9mABMAFASgFyIG5xgEwG4AoUSWBZFADywGcoAnGMAc00QebZLOniSpqAfnpMW7LLgKIAPp3FtEAb2KJEMYIhqIAhAF59icPgCmwFqfxoVa9YkamoIRkmol1AX2LfS4fpSs6FgACoxwALYwdKYAPNL4AHy8-hRIQbQK3JKIYZHRcVwSycQAhnQAnpDIqQKIGaJZEhx5UTHxePhyTWyJtuqa2tR6hsZgZhZgVjaq9g5OLm4eiN6eQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1uYDcBDRgBNaPDpWFFS6KImjQO0SODABfECqA&tsconfig=N4XyA

### Repro Code

```typescript
function f(): void;
function f(x: string): string;
function f(x?: string): void | string {
  if (x !== undefined) {
    return x;
  }
}

function g(): Promise<void>;
function g(x: string): Promise<string>;
async function g(x?: string): Promise<void | string> {
  if (x !== undefined) {
    return x;
  }
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-invalid-void-type": "error"
  }
}
```

### tsconfig

```jsonc
{}
```

### Expected Result

No errors.

### Actual Result

void is only valid as a return type or generic type variable. 3:25 - 3:29
void is only valid as a return type or generic type variable. 11:39 - 11:43

### Additional Info

It's possible to work around the linter in the first example by changing the implementation return type from `void | ...` to `undefined | ...`, although this may require introducing an explicit `return undefined` statement to the function body if the compiler option `noImplicitReturns` is set.

In the second example, changing `Promise<void | ...>` to `Promise<undefined | ...>` is not possible because it introduces a compiler error

>This overload signature is not compatible with its implementation signature. (2394)

### Versions

| package | version |
| -- | -- |
| `@typescript-eslint/eslint-plugin` | `5.39.0` |
| `@typescript-eslint/parser` | `5.39.0` |
| `TypeScript` | `4.8.4` |
| `ESLint` | `8.15.0` |
| `node` | `web` |

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5752

## Overview

This PR addresses #5752 and adjusts the rule so it doesn't report on `void` that's used as part of the return type of a function overload implementation:

```ts
function f(): void;
function f(x: string): string;
function f(x?: string): void | string {
  if (x !== undefined) {
    return x;
  }
}
```


packages/eslint-plugin/src/rules/no-invalid-void-type.ts
```diff
@@ -174,6 +174,135 @@ export default createRule<[Options], MessageIds>({
...
+        | TSESTree.TSDeclareFunction,
+    ): string | symbol | undefined {
+      if (
+        node.type === AST_NODE_TYPES.FunctionDeclaration ||
+        node.type === AST_NODE_TYPES.TSDeclareFunction
+      ) {
+        const id = nullThrows(
+          node.id,
+          'This may be null if and only if the parent is an `ExportDefaultDeclaration`.',
+        );
```
ronami, line 230: This _currently_ only gets called for nodes whose parents aren't `ExportDefaultDeclaration,` and so this is never `null`. If this code gets refactored, it may change, and this can throw.

I can change this to an optional change-expression, though this will show up as a partially covered line by codecov.

packages/eslint-plugin/src/rules/no-invalid-void-type.ts
```diff
@@ -174,6 +174,135 @@ export default createRule<[Options], MessageIds>({
...
+      switch (node.type) {
+        case AST_NODE_TYPES.ClassBody:
+        case AST_NODE_TYPES.Program:
+        case AST_NODE_TYPES.TSModuleBlock:
+        case AST_NODE_TYPES.BlockStatement:
+          return node.body;
+      }
+
+      /* istanbul ignore next */
+      return [];
```
ronami, line 192: Is there a preferred way to tackle this? I think this is currently an unreachable line.

packages/eslint-plugin/src/rules/no-invalid-void-type.ts
```diff
@@ -174,6 +174,135 @@ export default createRule<[Options], MessageIds>({
...
+      if (
+        node.type === AST_NODE_TYPES.FunctionDeclaration ||
+        node.type === AST_NODE_TYPES.TSDeclareFunction
+      ) {
+        const id = nullThrows(
+          node.id,
+          'This may be null if and only if the parent is an `ExportDefaultDeclaration`.',
+        );
+
+        return id.name;
```
ronami, line 232: This _currently_ only gets called for nodes whose parents aren't `ExportDefaultDeclaration,` and so this is never `null`. If this code gets refactored, it may change, and this can throw.

I can change this to an optional change-expression, though this will show up as a partially covered line by codecov.

```suggestion
        return node.id?.name;
```

JoshuaKGoldberg: Similar to the other comment, my fallback would be to use a `!` with a disable comment explaining why the types are wrong. If possible, fixing up the types to be correct would be even better.

You might find the existing uses of the `MakeRequired` utils type interesting. Some places in code use for node types in similar situations. E.g. `no-confusing-void-expression` has handling for return statements that must have an `argument`. Maybe there's a similar flow here for functions that must have an `id`?

I'm not worried about refactors. Our test coverage is so high that the cost of unnecessary safety precautions in code is IMO higher than the cost of the brittleness.

JoshuaKGoldberg (changes requested):
LGTM overall! Just requesting touchups around the types / `istanbul ignore`s. 🚀

packages/eslint-plugin/src/rules/no-invalid-void-type.ts
ronami: Interesting! I didn't consider proposing changes/improvements to the AST types. I've found #9560, and I think it would be a good option if I could be sure the changes are correct.

To my understanding, TypeScript's AST doesn't even have an actual AST node for `ExportDefaultDeclaration` or `ExportNamedDeclaration` as they're expressed as [`modifiers`](https://github.com/microsoft/TypeScript/blob/f69580f82146bebfb2bee8c7b8666af0e04c7e34/src/compiler/types.ts#L2083).

I've looked into [`@babel/parser`'s types](https://github.com/babel/babel/blob/87a5e2eef28a1f5a3783da5ed25a163236541b8d/packages/babel-parser/src/types.ts), [`@swc/types`](https://github.com/swc-project/swc/blob/9eceec44dffdfef7eb4079bf35e82301383528c8/packages/types/index.ts), and [`@oxc-project/types`](https://github.com/oxc-project/oxc/blob/main/npm/oxc-types/types.d.ts), but I didn't find anything that makes me confident this is valid in every case. I did verify `@typescript-eslint` itself is type-checked successfully with these assumptions (also the more technical [1](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#sec-function-definitions), [2](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#sec-exports)).

The stricter parent types for the AST I'd like to propose:
- `ExportDefaultDeclaration` and `ExportNamedDeclaration` should have the stricter `parent` type of `TSESTree.BlockStatement | TSESTree.Program | TSESTree. TSModuleBlock` (I'm not sure if `TSESTree.BlockStatement` is an option.
- `FunctionDeclaration` should have the stricter `parent` type of `TSESTree.BlockStatement | TSESTree.ExportDefaultDeclaration | TSESTree.ExportNamedDeclaration | TSESTree.Program`.
  
  TypeScript itself doesn't have a specific `parent` for `FunctionDeclaration` (it falls back to the generic `ts.Node`), although they do have stricter parent types for some other nodes.

How did we verify the correctness of the proposals in #6225? Also, would this type of change be considered a breaking change? I noticed that #9560 was released in v8, but it wasn't marked as a [breaking change](https://github.com/typescript-eslint/typescript-eslint/blob/main/CHANGELOG.md#-features-18).

I'm a bit hesitant to change the AST's types, as it can potentially lead to bugs, and I would love to hear opinions on this.

I updated the PR with these changes, so it's easier to discuss them.

packages/eslint-plugin/tests/rules/no-invalid-void-type.test.ts
```diff
@@ -419,6 +579,54 @@ ruleTester.run('allowInGenericTypeArguments: true', rule, {
...
+    {
+      code: 'type invalidVoidUnion = void | Map;',
+      errors: [
+        {
+          column: 25,
+          line: 1,
+          messageId: 'invalidVoidUnionConstituent',
+        },
+      ],
+    },
```
ronami, line 591: This test is unrelated to the PR, but it adds coverage to part of the code that is currently uncovered (it seemed small enough to be part of the PR, even if it's unrelated).

packages/eslint-plugin/src/rules/no-invalid-void-type.ts
```diff
@@ -174,6 +174,120 @@ export default createRule<[Options], MessageIds>({
...
+        // and only if the parent is an `ExportDefaultDeclaration`.
+        //
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        return node.id!.name;
+      }
+
+      return getStaticMemberAccessValue(node, context);
+    }
+
+    function hasOverloadMethods(
```
ronami, line 238: I think this may be useful as a shared utility function for different rules (might be better named `isOverloadImplementation`?). Re-using this function simplifies tackling https://github.com/typescript-eslint/typescript-eslint/issues/4586, for example.

ronami: Adding that this attempts to emulate `checker.isImplementationOfOverload()` for rules which aren't type-aware.

JoshuaKGoldberg (changes requested):
Comments inline. Exciting! 🚀

packages/eslint-plugin/src/rules/no-invalid-void-type.ts
ronami: That sounds good; I'm not sure why I mixed both on the same PR; thanks! 👍 

I've opened https://github.com/typescript-eslint/typescript-eslint/issues/10682, drafted this PR, and marked it as blocked.

If the issue is accepted, I'll create a new PR with the stricter parent node changes and eventually re-open this PR for review with only its relevant changes.

packages/eslint-plugin/src/util/hasOverloadSignatures.ts
```diff
@@ -0,0 +1,90 @@
...
+
+import { getStaticMemberAccessValue } from './misc';
+
+/**
+ * A type unaware variation of `checker.isImplementationOfOverload()`.
+ *
+ * @return true if the function or method node has overload signatures.
+ */
+export function hasOverloadSignatures(
+  node: MethodDefinitionWithBody | TSESTree.FunctionDeclaration,
```
ronami, line 16: As opposed to `FunctionDeclaration`, which has a different node type for its definition (`TSDeclareFunction`), `MethodDefinition` doesn't (`MethodDefinition` seem to either have its `value` be a `FunctionExpression` for implementation or `TSEmptyBodyFunctionExpression` in case of overloading, ignoring type/interface's `TSMethodSignature` which seems like it cannot be part of a class).

I've replaced the previous type of `TSESTree.MethodDefinition` with `MethodDefinitionWithBody` (even though the `.value` key isn't necessary).

packages/eslint-plugin/src/util/hasOverloadSignatures.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import type { MethodDefinitionWithBody } from './isMethodDefinitionWithBody';
+
+import { getStaticMemberAccessValue } from './misc';
+
+/**
+ * @return `true` if the function or method node has overload signatures.
+ */
+export function hasOverloadSignatures(
+  node: MethodDefinitionWithBody | TSESTree.FunctionDeclaration,
```
ronami, line 14: Unlike `FunctionDeclaration`, which has a different type of node for its definition (`TSDeclareFunction`), `MethodDefinition` doesn't. `MethodDefinition` seems to either have its `value` be a `FunctionExpression` for implementations or `TSEmptyBodyFunctionExpression` in case of overloading, ignoring type/interface's `TSMethodSignature`, which seems like it cannot be part of a class.

I've replaced the previous type of `TSESTree.MethodDefinition` with `MethodDefinitionWithBody` (even though the `.value` key isn't necessary) to make sure only function/method implementations can be passed. Otherwise, this may wrongly return `true` for a method overload node:

```ts
class T {
  // compiles, even though this is a type error
  foo(): string;
}
```

Please let me know if this check is unnecessary (or not worth the complexity).

packages/eslint-plugin/src/util/hasOverloadSignatures.ts
```diff
@@ -0,0 +1,88 @@
...
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+
+import type { MethodDefinitionWithBody } from './isMethodDefinitionWithBody';
+
+import { getStaticMemberAccessValue } from './misc';
+
+/**
+ * @return `true` if the function or method node has overload signatures.
+ */
+export function hasOverloadSignatures(
```
ronami, line 13: I've extracted this a separate utility because I think it may be useful in other places (e.g. #4586).

ronami:
Opening for re-review, please note:

- I've renamed and extracted `hasOverloadSignatures` to a separate utility because I think it may be useful in other places (e.g. #4586).

- Unlike `FunctionDeclaration`, which has a different type of node for its definition (`TSDeclareFunction`), `MethodDefinition` doesn't. `MethodDefinition` seems to either have its `value` be a `FunctionExpression` for implementations or `TSEmptyBodyFunctionExpression` in case of overloading (ignoring type/interface's `TSMethodSignature`, which seems like it cannot be part of a class's body).

  `hasOverloadSignatures` currently requests a node of type `TSESTree.FunctionDeclaration | TSESTree.MethodDefinition`, which means that it's possible to pass an overload method definition, for which the function will wrongly return `true`:
  
  ```ts
  class T {
    // compiles, even though this is a type error
    foo(): string;
  }
  ```

  This doesn't affect the `no-invalid-void-type` rule, as it [only passes function/method implementations](https://github.com/typescript-eslint/typescript-eslint/pull/10422/files#diff-938d0610e79914950d260fe15bef27f4a789b31221900e599e6aa92eb53f3ba9R274-R275), but it may be an issue for future rules that use this. This can have a stricter type, but it does add a bit of complexity (see https://github.com/typescript-eslint/typescript-eslint/pull/10422/commits/350e82f5e52eff98955b8e8d7235dac42bd01be7).

  I'm not sure if this is worth the extra complexity, would love to hear another opinion on this 🙃

JoshuaKGoldberg (approved):
Yeah I think this is great as-is. The utility has no known issues as used right now so we can cross that bridge when we get to it IMO. Awesome stuff! 🙌

PR #10433, opened 2024-12-01
fix(eslint-plugin): [no-base-to-string] short circuit on unions

kirkwaiblinger:
I haven't measured or anything, but I'm hoping this is slightly helpful for performance in some cases. I wonder if it might also prevent infinite recursion on some recursive types 🤷 

## PR Checklist


packages/eslint-plugin/src/rules/no-base-to-string.ts
JoshuaKGoldberg, line 1: @kirkwaiblinger maybe https://github.com/typescript-eslint/performance would be useful to you on this? I bet you could make a set of cases in it that generate `no-base-to-string`-heavy code and test the perf?

kirkwaiblinger:
Closing since I'm feeling lazy. But if perf issues come up we can revisit 🤷‍♂️

PR #10436, opened 2024-12-01
fix(eslint-plugin): [use-unknown-in-catch-callback-variable] only flag function literals

### LINKED ISSUE #9057: Enhancement: [use-unknown-in-catch-callback-variable] Option to only check inline functions

https://typescript-eslint.io/rules/use-unknown-in-catch-callback-variable/

### Description

I propose an option to only check on inline functions, i.e. `.catch((foo) => bar)` and not `.catch(baz)`. If `baz` accepts `any`, it's either from an external library or will be flagged by `--noImplicitAny`/`no-explicit-any`.

Use case: I have a few instances of `.catch(console.error)` in my codebase to log errors without throwing. `console.error` is typed as `(...data: any[]): void`, and as such this rule triggers an error.

### Fail

```TypeScript
Promise.resolve().catch((error) => error) // Preserve old behavior
```

### Pass

```TypeScript
Promise.resolve().catch((error: unknown) => error) // Preserve old behavior
Promise.resolve().catch(console.log) // Would previously error
Promise.resolve().catch(foo) // Would previously error

function foo(bar: any) { }
```

### Additional Info

_No response_

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9057 

## Overview

Since we're no longer generally flagging expressions that have type `(e: any) => whatever`, I added in recursion to catch function literals not passed directly, but whose type is still inferred. So, `promise.catch(e => condition ? (e) => {} : (e) => {});` still can and does trigger the rule, since `e` is inferred as `any` in both locations. Plus, the rule reports them each individually and consequently offers a suggestion now.

### DIFF
--- a/packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
+++ b/packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
@@ -113,7 +113,7 @@ export default createRule<[], MessageIds>({
           return collectFlaggedNodes(
             nullThrows(
               node.expressions.at(-1),
-              'seuence expression must have multiple expressions',
+              'sequence expression must have multiple expressions',
             ),
           );
         case AST_NODE_TYPES.ConditionalExpression:
--- a/packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
+++ b/packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
@@ -102,7 +102,7 @@ export default createRule<[], MessageIds>({
 
     function collectFlaggedNodes(
       node: Exclude<TSESTree.Node, TSESTree.SpreadElement>,
-    ): Exclude<TSESTree.Node, TSESTree.SpreadElement>[] {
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/use-unknown-in-catch-callback-variable.ts
```diff
@@ -100,10 +100,41 @@ export default createRule<[], MessageIds>({
...
+        case AST_NODE_TYPES.LogicalExpression:
+          return [
+            ...collectFlaggedNodes(node.left),
+            ...collectFlaggedNodes(node.right),
+          ];
+        case AST_NODE_TYPES.SequenceExpression:
+          return collectFlaggedNodes(
+            nullThrows(
+              node.expressions.at(-1),
+              'seuence expression must have multiple expressions',
```
JoshuaKGoldberg, line 116:
```suggestion
              'sequence expression must have multiple expressions',
```

JoshuaKGoldberg (dismissed):
Wow this is so much more effort than I thought would happen - and it's a great improvement! 🔥 

Pitch Meetings guy on YouTube saying "Wow wow wow ... wow"

PR #10454, opened 2024-12-05
fix(eslint-plugin): [no-unnecessary-condition] better message when comparing between literal types

### LINKED ISSUE #10455: Enhancement: [no-unnecessary-condition] improve error message for statically analyzable literal type comparisons again

### Description

I've thought more about the error message change in https://github.com/typescript-eslint/typescript-eslint/issues/9643 / https://github.com/typescript-eslint/typescript-eslint/pull/10194, and I'm not happy with the current error message (changed in #10194). 

'Unnecessary conditional, comparison is always {{trueOrFalse}}. Both sides of the comparison always have a literal type.' is more helpful than the prior message was, yes, but it doesn't fully explain what we're flagging to the user. 

I propose we change it to 'Unnecessary conditional, comparison is always {{trueOrFalse}}, since `{{left}} {{operator}} {{right}}` is {{trueOrFalse}}.', so that something like `if (expressionResultingInMinus55 < expressionResultingIn23) {}` will report an error as 'Unnecessary conditional, comparison is always true, since `-55 < 23` is true.'. 

This would IMO make it clearer _what_ types we're seeing and _why_ that indicates the expression has a constant value, helping the user assess whether the types are correct before resolving the complaint.

I have successfully implemented this in https://github.com/typescript-eslint/typescript-eslint/pull/10454 and I find the reports to be a big improvement when experimenting with it in the playground.

### Fail

```TypeScript
// unnecessary condition; 1 === 2 is false
if (1 === 2) {}


declare const expressionResultingIn23: 23;
declare const expressionResultingInMinus55: -55;

// unnecessary condition; -55 < 23 is true.
if (expressionResultingInMinus55 < expressionResultingIn23) {}
```

### Pass

```TypeScript
n/a
```

### Additional Info

cc @StyleShit

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10455


packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -491,6 +491,9 @@ export default createRule<Options, MessageId>({
           node,
           messageId: 'literalBooleanExpression',
           data: {
+            left: valueOfLiteralTypeToExpression(leftStaticValue.value),
```
kirkwaiblinger, line 494: I also investigated using `checker.typeToString()`. The only difference I observed was that that message for enums would look like 
`Unnecessary conditional, comparison is always true, since Foo.a === Foo.a is true` 
rather than 
`Unnecessary conditional, comparison is always true, since 1 === 1 is true`.

Not sure which is better. 🤷‍♂️ Slightly leaning towards the `checker.typeToString()` tbh.

JoshuaKGoldberg: 🤷 No preference from me, I could see either being nice.

JoshuaKGoldberg (approved):
Awkward seal meme. Caption: "SEAL OF APPROVAL"

I'd like to merge this before our Monday release. @kirkwaiblinger if you want to switch it over before then, feel free to do it and merge yourself (or if you're in a rush, just ping me and I can).

kirkwaiblinger:
should be good to merge now 👍

JoshuaKGoldberg (approved):
Yay!

PR #10461, opened 2024-12-05
fix(eslint-plugin): [no-unsafe-type-assertion] fix for unsafe assertion to a constrained type parameter

### LINKED ISSUE #10453: Bug: [no-unsafe-type-assertion] Fails to report unsafe assertions involving constrained type parameters

### Playground Link
[Playground link](https://typescript-eslint.io/play/#ts=5.5.4&fileType=.tsx&code=C4TwDgpgBA8gRgKwgY2FAvFA3gKClAMwHsiAuKAZ2ACcBLAOwHMBuHAXx04HoupGSAJlDgQAFgEMAbrSLVC1IgFsooSBWR0wwUjggAPMLLQEArvVQz6hM8gA8AYRNUl8JKgAqUfcAj0BFWEQUYAA%2BAAoiRHJHZ0VXYPcASminYBcgj2w8KB4oMIEiSiUIYFEGRigAd1pSlVFoBjATYETsySJaIUiEVnxcjQhxHyhxKHoISqhu4MJZOugiZqbgbOQieioxifjUch20TCxZsigAcnqQU6g2XpzeAHEYGAARclLaAIJxWgAbAJFkOInNB9p41iYfkIRFBFLJoBRIMhaARaMg6uIrJUJGhKtABkMIAJVutNuNKjE0nEMtooBT0m5gJ5MGT9rdqCUTNQrGS6VSGawONxeHBxFCxFIZHICAplPQiABaMwUcQECDy1Rq8QUCgQajASw6fSGPXWcz69am5AAJgcqXpCS8eh8fgC%2B3C3RSsVByVpdr5DtwfV4%2BUKFGK7yYVRqonmUEazVa%2BHanSmiFu-XZBJGW0m01QszkpQWS2axI2aBZ1L21Iw2GO5HOEEu13TvAAQgBBV51D5bQkBNJQcH0NBaqCAHAIlSroBqRtrdeb6IBcAjjBBG9BAUEk4h%2BJgWa6LKnA0DA4mo4kUJV1VUWkJh4gA1g00ErFn4y6SJrz9rXKwy576XrUu42TssAnLcl%2BfqsuwOBAA&eslintrc=N4KABGBECmAeAu0B2ATAzpAXGA2uCUADgDYCuA5gJZKYAC8AnodGgMYBOlh8AtC8dXgB6dtFYB7ALaTkKaCh6NmPVgAsxAa3mQANPgiQSFanSUsOXXv0FC0jAXcqtFTaCvWstKSPgC6eg3ZSYhYsMFACKHpXNk5uPjQBJGEJJDRKO2ReMx45ADNqSnhKcTSwnBh2dnF2XSgzSH99KLNYywSk4SRxHlI0gEM8txz%2BtDRodmLSsMrq2vwAXxAFoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDArugDTg2RGwAqkWgALdNADW6ACYBJIjPQAPWQEFo0dCTKUO6XgF8QhoA&tokens=false)

### Repro Code

```TypeScript
type Object = {
  foo: string;
}

export function myfunc<CustomObjectT extends Object>(input: CustomObjectT): CustomObjectT {
  const newObject: Object = { foo: 'hey' };
  // This is UNSAFE because CustomObjectT can/will be instantiated with a narrower subtype of Object
  // no-unsafe-type-assertion should (but doesn’t) catch this
  const newCustomObject = newObject as CustomObjectT;
  return newCustomObject;
}
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/no-unsafe-type-assertion": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true
  }
}
```

### Expected Result

The line with the type assertion involving the generic parameter (`newObject as CustomObjectT`) should report `@typescript-eslint/no-unsafe-type-assertion` by the same logic wherein Typescript considers such a conversion unsafe:
> ```
> 2322: Type 'Object' is not assignable to type 'CustomObjectT'.
>   'Object' is assignable to the constraint of type 'CustomObjectT', but 'CustomObjectT' could be instantiated with a different subtype of constraint 'Object'.
> ```

The first example in the playground shows that this case is not an allowable *assignment* according to TS, which, to my understanding, should be the heuristic used by `no-unsafe-type-assertion`

### Actual Result

No error is reported from `@typescript-eslint/no-unsafe-type-assertion`

### Additional Info

cc @ronami

thanks for your hard work on this rule!

controversial:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10453 

## Overview

Attempts to fix the issue described in #10453:
```ts
type Obj = { foo: string };
function func<T extends Obj>() {
  const o: Obj = { foo: 'hi' };
  // this type assertion is unsafe because T could be instantiated with a more specific subtype of Obj
  // no-unsafe-type-assertion should (but didn’t) report this
  const myObj = o as T;
}
```

After experimenting with the logic in `no-unsafe-type-assertion`, I found that obtaining the “types to check” using `services.getTypeAtLocation` in place of `getConstrainedTypeAtLocation` (i.e. omitting a call to `getBaseConstraintOfType`) allows the subsequent `isTypeAssignableTo` check to correctly catch this case.

Additionally, I added a new error message for this case: if the “expression type” is *not* assignable to the “asserted type”, but the asserted type has a constraint to which the expression type is assignable, the rule will report the following:
> Unsafe type assertion: the original type is assignable to the constraint of type '{{type}}', but '{{type}}' could be instantiated with a different subtype of its constraint.

---

This is my first contribution to this project and I’m broadly unfamiliar with both ESLint plugin development and TypeScript internals, so it’s very possible that I’m missing a hidden implication of this change—i.e. I’m not sure if there was an important reason to use `getConstrainedTypeAtLocation` in the original implementation. However, the existing tests for this rule are still passing.


controversial:
With respect to a custom error message, @kirkwaiblinger suggested:
> As a first pass, I think we could just model it after the TS error message?
> 
> ```
> Unsafe type assertion: 'Object' is assignable to the constraint of type 'CustomObjectT', but 'CustomObjectT' could be instantiated with a different subtype of constraint 'Object'.
> ```

In the original PR of this rule, #10051, a discussion about the appropriate verbosity of the error message [resolved](https://github.com/typescript-eslint/typescript-eslint/pull/10051#discussion_r1795639908) to omit any types other than the *explicitly written type* from the error message in order to prevent excessively verbose errors.

With that in mind, I’m starting out with the following error message:

> Unsafe type assertion: **the original type** is assignable to the constraint of type 'CustomObjectT', but 'CustomObjectT' could be instantiated with a different subtype of **the constraint**.

I’m happy to amend this message to be more verbose if we decide to go that way, though!

kirkwaiblinger (changes requested):
What a great first PR to typescript-eslint! This is very much on the right track, and in a tricky area of the codebase, too. 

Left a few comments and questions.

controversial:
Thanks so much for the thoughtful review @kirkwaiblinger !!

I added the test cases from your playground and addressed the missing cases. For the missing case of *unconstrained* type parameters, I added another message to match the one Typescript uses:
> Unsafe type assertion: '{{type}}' could be instantiated with an arbitrary type which could be unrelated to the original type.

---

<details>
<summary>
The only test case that tripped me up a bit was the one you had called <code>parameterExtendsOtherParameter</code>:

</summary>

```ts
function parameterExtendsOtherParameter<T, V extends T>(x: T, y: V) {
  x = y;
  y as T; // allowed

  y = x;
  x as V; // banned; generic arbitrary subtype <-- currently has wrong error message.
}
```

My first impression was that this case should raise the “x is assignable to the constraint of V” message. However, my current implementation instead says “V could be instantiated with an arbitrary type.” **Notably**, though (and somewhat to my surprise) this behavior is consistent with the message that Typescript gives for this case, so I think it’s probably ok as-is?

I [added](https://github.com/typescript-eslint/typescript-eslint/pull/10461/commits/c39c4b72645a86028b46a6837c7c06744eafb54e) a new `parameterExtendsOtherParameter` test where T is *not* unconstrained; here, both TypeScript (for assignment) and my implementation raise “is assignable to the constraint of V” like I expected.

I renamed the original test to <code>parameterExtends<strong>Unconstrained</strong>Parameter</code>; I’m curious to hear your thoughts about whether we should try to produce “assignable to the constraint” for this edge case?

</details>

controversial:
Merged the latest `main` to resolve merge conflicts from #10472

When you get a chance, let me know if there’s anything else you need from me @kirkwaiblinger :)

kirkwaiblinger:
> My first impression was that this case should raise the “x is assignable to the constraint of V” message. However, my current implementation instead says “V could be instantiated with an arbitrary type.” **Notably**, though (and somewhat to my surprise) this behavior is consistent with the message that Typescript gives for this case, so I think it’s probably ok as-is?

I feel strongly that TS is wrong here, since `V` plainly cannot be _unrelated_ to `T`. It's guaranteed to be a subtype of `T`, since... `V extends T` 🤷 . However, I agree that it makes sense to match their behavior here, and we can follow up by filing an issue to TS if we like. (Feel free to do this if you're interested! Otherwise I probably will at some point 🙂)

> I [added](https://github.com/typescript-eslint/typescript-eslint/pull/10461/commits/c39c4b72645a86028b46a6837c7c06744eafb54e) a new `parameterExtendsOtherParameter` test where T is _not_ unconstrained; here, both TypeScript (for assignment) and my implementation raise “is assignable to the constraint of V” like I expected. I renamed the original test to `parameterExtends**Unconstrained**Parameter`.

🎉

packages/eslint-plugin/tests/rules/no-unsafe-type-assertion.test.ts
```diff
@@ -1110,6 +1133,247 @@ export function myfunc<CustomObjectT extends string>(
...
+        ],
+      },
+      // 2. parameter -> parameter assertions
+      {
+        code: `
+function parameterExtendsOtherParameter<T extends string | number, V extends T>(
+  x: T,
+  y: V,
+) {
+  y as T; // allowed
```
kirkwaiblinger, line 1265:
```suggestion
```

packages/eslint-plugin/tests/rules/no-unsafe-type-assertion.test.ts
```diff
@@ -1110,6 +1133,247 @@ export function myfunc<CustomObjectT extends string>(
...
+            endLine: 7,
+            line: 7,
+            messageId: 'unsafeTypeAssertionAssignableToConstraint',
+          },
+        ],
+      },
+      {
+        code: `
+function parameterExtendsUnconstrainedParameter<T, V extends T>(x: T, y: V) {
+  y as T; // allowed (as above)
```
kirkwaiblinger, line 1282:
```suggestion
```

kirkwaiblinger (changes requested):
Really outstanding work! Great attention to detail on noticing the additional error message, nicely done on pushing back where it made sense to do so!

Trivial cleanup requested but this is otherwise an enthusiastic approval from me! 🙂 We'll probably leave this open for a little bit as well for another team member to take a look if they like, since it's nuanced stuff. 

Thanks!

cuzco

controversial:
cleanup completed! thanks again for your thoughtful reviews, excited to have this merged at some point :)

kirkwaiblinger (approved):

packages/eslint-plugin/tests/rules/no-unsafe-type-assertion.test.ts
```diff
@@ -1269,8 +1269,8 @@ function parameterExtendsOtherParameter<T extends string | number, V extends T>(
           {
             column: 3,
             endColumn: 9,
-            endLine: 7,
-            line: 7,
+            endLine: 6,
+            line: 6,
```
controversial, line 1273: my bad

PR #3805, opened 2021-08-28
fix(eslint-plugin): false-positive/negative with array index in no-unnecessary-condition

ypresto:
This PR fixes several false-positive/negatives in no-unnecessary-condition.
- False-positive at nullish coalescing after array index access: `arr[42]?.foo ?? 0`
- False-negative at `arr[42] ?? 0` where `arr[42]` is always nullish or never.
- False-negative at `arr[0].foo?.bar` where `foo` is not nullable.

PR #1534, opened 2020-01-28
feat(eslint-plugin): [no-unnecessary-condition] ignore basic array indexing false positives

Retsam:
[As discussed here](https://github.com/typescript-eslint/typescript-eslint/issues/1423#issuecomment-573874462), a current pain-point with no-unnecessary-condition is that array indexes are often a source of false positives because the types don't represent the possibility of an out-of-bounds error.

This adds a check to ignore code like:

```ts
function example(arr: number[]) {
  if(arr[0]) {}
}
```

This version only handles the "obvious" case where the index signature is directly inside the condition - it will still raise errors on code like:

```ts
function example(arr: number[]) {
  const head = array[0];
  if(head) {}
}
```

I believe fixing this would be possible, but I'm not sure it's worth the complexity of implementation.

---

I considered making this configurable, but I just can't come up with any cases in which you'd actually want the previous behavior.  Unlike the object case, I don't think anyone is realistically going to model their arrays as `Array<T | undefined>` for the sake of safe index checks.

This I don't think this is considered a breaking-change - all previously valid code is still valid - but if so, I'd rather point at 3.0 than put it behind a temporary option flag.

Fixes #1544

PR #10464, opened 2024-12-06
feat(eslint-plugin): add new rule `no-extraneous-return-types`

### LINKED ISSUE #6442: Rule proposal: disallow unused return types

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6442

## Overview

This PR attempts to tackle #6442 and add a new rule to report on unused return type annotations:

```ts
function test (): string | null {
  return 'string'
}
```

Some of my thoughts:

- The implementation takes [the approach described here](https://github.com/typescript-eslint/typescript-eslint/issues/6442#issuecomment-2437185952), and the rule warns on inline union types but not on referenced ones. This is important since referenced union types may be harder to fix and maintain.

  Consider the following example:

   ```ts
  interface A {};
  interface B {};
  interface C {};
  interface D {};
 
  // used throughout the project
  export type Base = A | B | C | D;
 
  function getNode(): Base {
    // ...
  }
   ```

  The function may only be returning a couple of the possible types in the union type and not all of them. Returning a base type is convenient, and having the function to declare each type it may return is tedious, especially in cases with large union types (e.g. [`TSESTree.Node`](https://github.com/typescript-eslint/typescript-eslint/blob/0d0722c079cb8ad4f0b32b97063c9aa405d2a391/packages/ast-spec/src/unions/Node.ts#L175-L343)).

  Another example is using a type exposed by a third-party library that is usually used contextually, but if the function is extracted, it may have an explicit return type annotation ([vite plugins](https://github.com/vitejs/vite/blob/1bfe21b9440f318c940f90e425a18588595225fd/packages/vite/src/node/plugin.ts#L336) come to mind):

  ```ts
  export type VitePlugin = { kind: 'one' } | { kind: 'two' };

  function createVitePlugin(): VitePlugin {
    return {
      kind: 'one',
    };
  }
  ```

  In my opinion, asking functions to declare exactly the type they're returning can be tedious, especially if some of these types come from a third party and aren't exposed (an example that comes to mind is [`React.ReactNode`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/04c93eabdf4ccdf47ae2e4e45ec01c5069e7e190/types/react/index.d.ts#L426-L439)).

- The implementation includes special treatment for opening promises (also `async` functions), arrays, and tuples. Additional ones can be added (e.g. `Set`, `Map`, and potentially object-literal fields):

  ```ts
  // Unused or redundant union type 'null' in return type.
  function test(): Promise<string | null> {
    return Promise.resolve('hello');
  }
  ```

- The rule doesn't include a suggestion or an auto-fix. I think an auto-fix is correct here, but I'll be happy to hear some opinions.

- Sometimes, not all code paths return a value on a function. Because of this, the rule will not report `void` or `undefined`  being potentially unnecessary on the return type annotation:

  ```ts
  function test(a: boolean): number | undefined {
    if (a) {
      return 1;
    }
  }
  ```

- Running this on `typescript-eslint`'s code base shows what seems like 7 valid reports (some show wrong types, some highlight a problematic implementation): [1](https://github.com/typescript-eslint/typescript-eslint/blob/0d0722c079cb8ad4f0b32b97063c9aa405d2a391/packages/eslint-plugin/src/rules/consistent-return.ts#L57-L59), [2](https://github.com/typescript-eslint/typescript-eslint/blob/0d0722c079cb8ad4f0b32b97063c9aa405d2a391/packages/eslint-plugin/src/rules/explicit-member-accessibility.ts#L247-L260), [3](https://github.com/typescript-eslint/typescript-eslint/blob/0d0722c079cb8ad4f0b32b97063c9aa405d2a391/packages/eslint-plugin/src/rules/no-array-delete.ts#L73-L104), [4](https://github.com/typescript-eslint/typescript-eslint/blob/0d0722c079cb8ad4f0b32b97063c9aa405d2a391/packages/eslint-plugin/src/rules/no-mixed-enums.ts#L129-L187), [5](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-unnecessary-template-expression.ts/#L115-L123), [6](https://github.com/typescript-eslint/typescript-eslint/blob/0d0722c079cb8ad4f0b32b97063c9aa405d2a391/packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts#L93-L98), [7](https://github.com/typescript-eslint/typescript-eslint/blob/0d0722c079cb8ad4f0b32b97063c9aa405d2a391/packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts#L208-L215).

- I think it may be helpful to have the option to ignore module boundaries (e.g. exported functions or public methods of exported classes/objects). It's convenient to sometimes have unused return types, so calling code will need to handle potential future return types.

- When the rule reports a failure, it stringifies the name of the redundant type: `Unused or redundant union type '{{type}}' in return type.`. While this works well for flat unions (e.g. `string | number | boolean`), I'm not sure how clear this is when the rule fails on a nested box like a promise or an array: `Array<number | string | boolean>`. Wdyt? Should the message be different for box (and potentially nested box) values?

🚀

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
+++ b/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
@@ -9,11 +9,9 @@ import TabItem from '@theme/TabItem';
 >
 > See **https://typescript-eslint.io/rules/no-extraneous-return-types** for documentation.
 
-Explicit return type annotations are optional in TypeScript. However, explicit return types do make it visually more clear what type is returned by a function.
-
-A function can include return type annotations that are never used. This can happen when refactoring code or by mistake. Having unused return types adds unnecessary complexity because it needs to be handled by code that calls the function.
-
-This rule reports cases where a function has an unused return type annotation.
+Explicit function return type annotations can sometimes be helpful in TypeScript to describe a more precise return type than what the type checker would infer.
+Functions that can return one of multiple types of values generally use union types to describe their return type.
+This lint rule checks for "extraneous", or unnecessary, constituents of union type returns: those that don't describe a type the function can return.
 
 ## Examples
 
--- a/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
+++ b/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
@@ -1,0 +1,67 @@
+---
+description: 'Disallow extraneous union types in return type annotations.'
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-extraneous-return-types** for documentation.
+
+Explicit return type annotations are optional in TypeScript. However, explicit return types do make it visually more clear what type is returned by a function.
+
+A function can include return type annotations that are never used. This can happen when refactoring code or by mistake. Unused return types also need to be handled by code that calls such a function.
+
+This rule reports cases where a function has an unused return type annotation.
+
+## Examples
+
+<Tabs>
+<TabItem value="❌ Incorrect">
+
+```ts
+function test1(): string | null {
+  return 'string';
+}
+
+function test2(): Promise<string | null> {
+  return Promise.resolve('string');
+}
+
+class A {
+  test(): string | null {
+    return 'string';
+  }
+}
+```
+
+</TabItem>
+<TabItem value="✅ Correct">
+
+```ts
+function test1(): string {
+  return 'string';
+}
+
+function test2(): Promise<string> {
+  return Promise.resolve('string');
+}
+
+class A {
+  test(): string {
+    return 'string';
+  }
+}
+```
+
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+This rule can be completely redundant for projects without explicit return type annotations on functions. It works best when coupled with other rules that enforce explicit return type annotations such as [`explicit-function-return-type`](https://typescript-eslint.io/rules/explicit-function-return-type) or [`explicit-module-boundary-types`](https://typescript-eslint.io/rules/explicit-module-boundary-types).
+
+```
+
+```
--- a/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
+++ b/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
@@ -58,4 +58,5 @@ class A {
 
 ## When Not To Use It
 
-This rule can be completely redundant for projects without explicit return type annotations on functions. It works best when coupled with other rules that enforce explicit return type annotations such as [`explicit-function-return-type`](https://typescript-eslint.io/rules/explicit-function-return-type) or [`explicit-module-boundary-types`](https://typescript-eslint.io/rules/explicit-module-boundary-types).
+Some projects have early-stage functions, which are likely to evolve later on to match their declared return types.
+You might consider using [ESLint disable comments](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) for those specific situations instead of completely disabling this rule.
--- a/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
+++ b/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
@@ -11,7 +11,7 @@ import TabItem from '@theme/TabItem';
 
 Explicit return type annotations are optional in TypeScript. However, explicit return types do make it visually more clear what type is returned by a function.
 
-A function can include return type annotations that are never used. This can happen when refactoring code or by mistake. Unused return types also need to be handled by code that calls such a function.
+A function can include return type annotations that are never used. This can happen when refactoring code or by mistake. Having unused return types adds unnecessary complexity because it needs to be handled by code that calls the function.
 
 This rule reports cases where a function has an unused return type annotation.
 
--- a/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
+++ b/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
@@ -62,6 +62,3 @@ class A {
 
 This rule can be completely redundant for projects without explicit return type annotations on functions. It works best when coupled with other rules that enforce explicit return type annotations such as [`explicit-function-return-type`](https://typescript-eslint.io/rules/explicit-function-return-type) or [`explicit-module-boundary-types`](https://typescript-eslint.io/rules/explicit-module-boundary-types).
 
-```
-
-```
--- a/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
+++ b/packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
@@ -61,4 +61,3 @@ class A {
 ## When Not To Use It
 
 This rule can be completely redundant for projects without explicit return type annotations on functions. It works best when coupled with other rules that enforce explicit return type annotations such as [`explicit-function-return-type`](https://typescript-eslint.io/rules/explicit-function-return-type) or [`explicit-module-boundary-types`](https://typescript-eslint.io/rules/explicit-module-boundary-types).
-
--- a/packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
+++ b/packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
@@ -1,0 +1,278 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+import type * as ts from 'typescript';
+
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+import * as tsutils from 'ts-api-utils';
+
+import { createRule, forEachReturnStatement, getParserServices } from '../util';
+
+function getTypeArgumentIfPromiseNode(
+  node: TSESTree.TypeNode,
+): TSESTree.TypeNode | null {
+  if (
+    // Check for Promise<...> type reference
+    node.type === AST_NODE_TYPES.TSTypeReference &&
+    node.typeName.type === AST_NODE_TYPES.Identifier &&
+    node.typeName.name === 'Promise' &&
+    node.typeArguments?.params.length === 1
+  ) {
+    return node.typeArguments.params[0];
+  }
+
+  return null;
+}
+
+function getTypeArgumentIfArrayNode(
+  node: TSESTree.TypeNode,
+): TSESTree.TypeNode | null {
+  if (
+    // Check for Array<...> type reference
+    node.type === AST_NODE_TYPES.TSTypeReference &&
+    node.typeName.type === AST_NODE_TYPES.Identifier &&
+    node.typeName.name === 'Array' &&
+    node.typeArguments?.params.length === 1
+  ) {
+    return node.typeArguments.params[0];
+  }
+
+  if (
+    // Check for (...)[]
+    node.type === AST_NODE_TYPES.TSArrayType
+  ) {
+    return node.elementType;
+  }
+
+  return null;
+}
+
+function getTypeArgumentIfTupleNode(
+  node: TSESTree.TypeNode,
+): TSESTree.TypeNode[] | null {
+  if (
+    // Check for [...]
+    node.type === AST_NODE_TYPES.TSTupleType &&
+    node.elementTypes.length > 0
+  ) {
+    return node.elementTypes;
+  }
+
+  return null;
+}
+
+export default createRule({
+  name: 'no-extraneous-return-types',
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Disallow extraneous union types in return type annotations',
+      requiresTypeChecking: true,
+    },
+    fixable: 'code',
+    messages: {
+      unusedReturnTypes:
+        "Unused or redundant union type '{{type}}' in return type.",
+    },
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    function getRelevantReturnTypes(
+      node: TSESTree.TypeNode,
+      returnTypes: ts.Type[],
+    ): ts.Type[] | null {
+      const type = services.getTypeAtLocation(node);
+
+      const relevantReturnTypes = returnTypes.filter(returnType =>
+        checker.isTypeAssignableTo(returnType, type),
+      );
+
+      if (relevantReturnTypes.length === 0) {
+        context.report({
+          node,
+          messageId: 'unusedReturnTypes',
+          data: {
+            type: checker.typeToString(type),
+          },
+        });
+
+        return null;
+      }
+
+      return relevantReturnTypes;
+    }
+
+    function checkReturnType(
+      node: TSESTree.TypeNode,
+      returnTypes: ts.Type[],
+    ): void {
+      const tupleElementTypes = getTypeArgumentIfTupleNode(node);
+
+      if (tupleElementTypes) {
+        const relevantReturnTypes = getRelevantReturnTypes(node, returnTypes);
+
+        if (relevantReturnTypes) {
+          for (const [index, elementType] of tupleElementTypes.entries()) {
+            const returnTypes = relevantReturnTypes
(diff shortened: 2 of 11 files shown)


packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    return {
+      'FunctionDeclaration,FunctionExpression,ArrowFunctionExpression'(
+        node:
+          | TSESTree.ArrowFunctionExpression
+          | TSESTree.FunctionDeclaration
+          | TSESTree.FunctionExpression,
+      ): void {
+        if (!node.returnType || node.generator) {
```
JoshuaKGoldberg, line 210: > `|| node.generator`

[Bug] Generators are indeed wacky at times, but not always. As a user I'd think it a bug that the rule doesn't flag these:

```ts
function* test(): Generator<number | string> {
  return 1;
}
```

```ts
function* test(): Generator<never, number | string, unknown> {
  return 1;
}
```

packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
JoshuaKGoldberg, line 1: > The rule doesn't include a suggestion or an auto-fix. I think an auto-fix is correct here, but I'll be happy to hear some opinions.

[Feature] I'd start with an auto-fix. Thinking out loud:

* This is purely in the type system so we're not going to cause runtime error
* Any actual necessary type shouldn't be reported on, so an invalid fix would be a bug in the rule already

packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
```diff
@@ -0,0 +1,311 @@
...
+          return;
+        }
+
+        const actualReturnTypes = getPotentialReturnTypes(node);
```
JoshuaKGoldberg, line 214: [Style] This confused me at first 😄. Is there a reason to name one _"actual"_ and the other _"potential"_?

If not, request: use the same term for both? 

After reading through it, _"actual"_ matches how I'm understanding it. So +1 from me on that one.

packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
```diff
@@ -0,0 +1,311 @@
...
+      }
+
+      return relevantReturnTypes;
+    }
+
+    function checkReturnType(
+      node: TSESTree.TypeNode,
+      returnTypes: ts.Type[],
+      allowUnusedVoidOrUndefinedValues = false,
+    ): void {
```
JoshuaKGoldberg, line 58: [Style] This is totally optional and if you don't like it I won't be upset at all 😄, but: I found this function to be a lot to read through. It's essentially a set of individual cases to check: tuple types, promise type argument, array type argument, miscellaneous. It might be clearer to split it up?

```ts
function checkReturnType(...) {
  const tupleElementTypes = getTypeArgumentIfTupleNode(node);
  if (tupleElementTypes) {
    checkTupleElementReturnTypes(...);
    return;
  }

  const promiseTypeArgument = getTypeArgumentIfPromiseNode(node);
  if (promiseTypeArgument) {
    checkPromiseTypeArgumentReturnTypes(...);
    return;
  }

  // (similar const/if/check/return for array type arguments)

  // (similar if/check/return for union types)

  // ...miscellaneous logic
}
``` 

If you don't like this, not a problem at all - just a personal preference of mine.

packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
```diff
@@ -0,0 +1,311 @@
...
+        "Unused or redundant union type '{{type}}' in return type.",
+    },
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    const services = getParserServices(context);
+    const checker = services.program.getTypeChecker();
+
+    function getRelevantReturnTypes(
```
JoshuaKGoldberg, line 29: [Style] Nit: this function both _gets_ and _checks_. It took me a re-read to get that it does both. WDYT about expanding its name to mention both?

```suggestion
    function getRelevantReturnTypesOrReport(
```

packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
```diff
@@ -0,0 +1,67 @@
...
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+This rule can be completely redundant for projects without explicit return type annotations on functions. It works best when coupled with other rules that enforce explicit return type annotations such as [`explicit-function-return-type`](https://typescript-eslint.io/rules/explicit-function-return-type) or [`explicit-module-boundary-types`](https://typescript-eslint.io/rules/explicit-module-boundary-types).
+
+```
+
+```
```
JoshuaKGoldberg, line 67: [Docs] Unused?

```suggestion
```

packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+```
+
+</TabItem>
+</Tabs>
+
+## When Not To Use It
+
+This rule can be completely redundant for projects without explicit return type annotations on functions. It works best when coupled with other rules that enforce explicit return type annotations such as [`explicit-function-return-type`](https://typescript-eslint.io/rules/explicit-function-return-type) or [`explicit-module-boundary-types`](https://typescript-eslint.io/rules/explicit-module-boundary-types).
```
JoshuaKGoldberg, line 63: [Docs] I think this accidentally implies that rule is _only_ useful with those functions. Which is not true, I've found myself occasionally wanting it in projects that don't use those rules or isolatedDeclarations. Even though I hate unnecessary types I still sometimes use them to make functions more clear or get better return types.

I think this should instead note the actual edge cases that make the rule less useful. Which are ... not that many that I can think of. Maybe:

* If the functions are early stage and will likely be evolved/widened later
* If you're explicitly trying to match some external, more-permissive interface

...along with the standard notes about considering using inline disables in those cases.

packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-extraneous-return-types** for documentation.
+
+Explicit return type annotations are optional in TypeScript. However, explicit return types do make it visually more clear what type is returned by a function.
+
+A function can include return type annotations that are never used. This can happen when refactoring code or by mistake. Having unused return types adds unnecessary complexity because it needs to be handled by code that calls the function.
+
+This rule reports cases where a function has an unused return type annotation.
```
JoshuaKGoldberg, line 16: [Docs] Well, they're _often_ optional, but sometimes required - e.g. for well-typed recursive functions. 

Also, this never actually mentions union types. Which makes it seem like the rule checks entire return type annotations for being necessary - not just individual unions. I think the explanation would be most helpful if it centered around that.

WDYT about something roughly like this?

```suggestion
Explicit function return type annotations can sometimes be helpful in TypeScript to describe a more precise return type than what the type checker would infer.
Functions that can return one of multiple types of values generally use union types to describe their return type.
This lint rule checks for "extraneous", or unnecessary, constituents of union type returns: those that don't describe a type the function can return.
```

...that's kind of clunky and I don't love it. Someone like @Josh-Cena who's good at describing things could probably do a better job. 😄

packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
JoshuaKGoldberg, line 1: > The implementation includes special treatment for opening promises (also async functions), arrays, and tuples. Additional ones can be added (e.g. `Set`, `Map`, and potentially object-literal fields)

[Feature] How about ... any generic shape altogether? It's kind of limiting to not be able to check user-defined classes:

```ts
class Box<T> {
  constructor(public readonly value: T) {}
}

function makeBox(): Box<number | string> {
  return new Box("abc");
}
```

How difficult / edge-case-y would that be?

JoshuaKGoldberg (changes requested):
This is an awesome start! 👏 

I generally agree with most of the decisions you made. I commented a few feature additions I think should land with the rule inline. Looking forward to the discussion, I think this one is going to be very nuanced.

packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
ronami: That's brilliant! It was rather simple to do, thanks!

ronami: Okay, this is significantly more complex than I initially considered it. Summarizing my thoughts about this.

---

While the example above can be tackled easily enough, for some boxed values, it would be difficult to tell which return statements match:

```ts
// matching an object literal against a class type
class BoxClass<T> {
  constructor(public readonly value: T) {}
}

function test1(): BoxClass<number | string> {
  return { value: 'abc' };
}

// matching against a non-class boxed type
interface BoxType<T> {
  value: T;
}

function test2(): BoxType<number | string> {
  return { value: 'abc' };
}

// type that creates nested boxes
type NestedBox<T> = BoxClass<BoxClass<T>>;

function test3(): NestedBox<number | string> {
  return new BoxClass(new BoxClass(10));
}

// types that make computations rather than create boxed values
function test4(): ReturnType<(() => number) | (() => string)> {
  return 10;
}
```

packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
```diff
@@ -0,0 +1,311 @@
...
+import { createRule, forEachReturnStatement, getParserServices } from '../util';
+
+export default createRule({
+  name: 'no-extraneous-return-types',
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Disallow extraneous union types in return type annotations',
+      requiresTypeChecking: true,
+    },
```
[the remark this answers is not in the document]
ronami: Yes, I'd even consider having such a rule in `recommended`. Won't this be a breaking change though (and should be done in a new major)?

packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
ronami: I've added support Generators (and also AsyncGenerators 🙈) 👍

packages/eslint-plugin/src/rules/no-extraneous-return-types.ts
ronami: I've added an auto-fix 👍 

One place I'm not entirely sure how to handle though, is the following:

```ts
function test1(): string {
  return 1;
}

function test2(): boolean | number {
  return 1;
}
```

In both of those cases, the return type is, technically, unused. There's a type error, but it's unused. I chose to report these return types as unused, but not apply an auto-fix. Wdyt about this? Should these cases be reported? Potentially with a different error message?

packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
ronami: I really like the way you've explained this 👍

packages/eslint-plugin/docs/rules/no-extraneous-return-types.mdx
ronami: Yes, I agree.

The current explanation of `This rule can be completely redundant for projects...` is just wrong (and may make incoming developers miss a lot of the value of such a rule). I usually omit return type annotation in projects, and I still have dozens of functions I choose to annotate.

I updated this to match this, let me know what you think 👍

ronami:
One issue I noticed, which I'm not sure if the rule should be a special case, is the following (I noticed this on tseslint's code-base):

```ts
// noUncheckedIndexedAccess turned off

function foo(a: string[]): string | null {
  return a[0] ?? null;
}
```

The type of this expression is `string` rather than `string | null`, and it looks like this has been special-cased by `no-unnecessary-condition` (in #3805). The `null` return type then pops up as unused, even though it is possible and even allowed by `no-unnecessary-condition`.

Should this be special-cased by this rule too?

kirkwaiblinger:
> One issue I noticed, which I'm not sure if the rule should special case, is the following (I noticed this on tseslint's code-base):
> 
> ```ts
> // noUncheckedIndexedAccess turned off
> 
> function foo(a: string[]): string | null {
>   return a[0] ?? null;
> }
> ```
> 
> The type of this expression is `string` rather than `string | null`, and it looks like this has been special-cased by `no-unnecessary-condition` (in #1534). The `null` return type then shows up as unused, even though it is possible and even allowed by `no-unnecessary-condition`.

I feel like every 3 months or so I have an eye-opening revelation about how bizarrely TS can break my assumptions, and this is one such case. Coincidentally I was just looking at a very similar situation in https://github.com/microsoft/TypeScript/issues/60813

> Should it be special-cased by this rule too?

I'm begrudgingly in favor of trying to special case this... One could imagine recursing on ternary, nullish coalescing, possibly even logical `&&`/`||` operands to look for possibly returned types rather than just the overall expression 🤷‍♂️

ehoogeveen-medweb:
> 🤔 I feel like the allowing basic array indexing false positives isn't as good as it used to be. #1534 (January-March 2020) predates [TypeScript 4.1](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html) (November 2020). Now that we have `noUncheckedIndexAccesses`, indexed types have the option to be better handled by the compiler...

Anecdotal and it's been a while since I tried it, but I found `noUncheckedIndexedAccess` pretty awkward to use. For arrays in particular, the problem is that TS has no "array length inference" - either you have tuples with an immutable length, or you have arrays with no guarantees about the length where you have to check every time. I played around with workarounds like intersecting with tuples like `[T, ...T]`, but that's very unsafe when combined with functions that modify arrays in place (like `.pop()` or `.push()`). I had similar problems with objects, although I think some of those may be fixed now.

JoshuaKGoldberg:
Ha, yeah, same. I've tried the flag repeatedly and never ended up liking it.

kirkwaiblinger:
The corollary though is that then we "should" flag this code too (depending on tsconfig settings)
```ts
function foo(): string | null {
    if (true) {
        return ""
    } else {
        return null;
    }
}
```
because this is valid TS
```ts
function foo(): string  {
    if (true) {
        return ""
    } else {
        return null;
    }
}
```
([playground](https://www.typescriptlang.org/play/?target=99&moduleResolution=99&module=199&ts=5.7.2&noUncheckedIndexedAccess=true&allowUnreachableCode=true#code/GYVwdgxgLglg9mABMOcAUBKAXIgzlAJxjAHNFEBvAKHPJmETUJAFMNKbauCWoQCkAIkGdyAX0QsANrhYcu3XvyRgQUqQG5RiMVTFA) - sometimes this doesn't work right... make sure `target` is set to something recent)

kirkwaiblinger:
I think it makes more sense for us to check what we can syntactically determine may be returned, even if TS doesn't actually use that strategy (yet?) in its actual return type validation.

ronami:

PR #10472, opened 2024-12-07
fix(eslint-plugin): use consistent naming for asserting types and casting values

### LINKED ISSUE #10458: Bug: [no-unsafe-type-assertion] Remove the term "cast" from error message report

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3lFIQC5sUBPAbgCgc9s14AmUoA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tieQEMAZolpFStPsmSJo%2BShybooAdz7QF4MAF8QmoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
declare const anee: any;
anee as 2;
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-unsafe-type-assertion": "warn"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

Unsafe **type assertion** from `any` detected: consider using type guards or a safer **type assertion**. 2:1 - 2:10

### Actual Result

Unsafe **cast** from `any` detected: consider using type guards or a safer **cast**. 2:1 - 2:10

### Additional Info

The [docs for no-unsafe-type-assertion](https://typescript-eslint.io/rules/no-unsafe-type-assertion/), as well as [the TS docs on type assertions](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions), are careful not to use the word "cast" since, in other programming languages, that generally can imply runtime behavior. Note that this idea is reinforced by the naming of the core rule [no-extra-boolean-cast](https://eslint.org/docs/latest/rules/no-extra-boolean-cast).

I propose: let's avoid using the word "cast" entirely.

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10458

## Overview

This PR addresses #10458 and makes consistent use of asserting vs casting.

There is some inconsistency for casting values: to cast ([`no-extra-boolean-cast`](https://eslint.org/docs/latest/rules/no-extra-boolean-cast)), to coerce (e.g. [`no-implicit-coercion`](https://eslint.org/docs/latest/rules/no-implicit-coercion), [MDN](https://developer.mozilla.org/en-US/docs/Glossary/Type_coercion)), to convert (upcoming [`no-unnecessary-type-conversion`](https://github.com/typescript-eslint/typescript-eslint/pull/10182#issuecomment-2453282643)). To my understanding, these all mean the same thing.

I've used consistent language for casting a value (currently using `to cast`). Please let me know if this sounds OK or if I should change it back.

### DIFF
--- a/packages/eslint-plugin/docs/rules/explicit-function-return-type.mdx
+++ b/packages/eslint-plugin/docs/rules/explicit-function-return-type.mdx
@@ -161,7 +161,7 @@ let funcExpr: FuncType = function () {
 };
 
 let asTyped = (() => '') as () => string;
-let castTyped = <() => string>(() => '');
+let assertTyped = <() => string>(() => '');
 
 interface ObjectType {
   foo(): number;
@@ -172,7 +172,7 @@ let objectProp: ObjectType = {
 let objectPropAs = {
   foo: () => 1,
 } as ObjectType;
-let objectPropCast = <ObjectType>{
+let objectPropAssertion = <ObjectType>{
   foo: () => 1,
 };
 
--- a/packages/eslint-plugin/docs/rules/explicit-function-return-type.mdx
+++ b/packages/eslint-plugin/docs/rules/explicit-function-return-type.mdx
@@ -161,7 +161,6 @@ let funcExpr: FuncType = function () {
 };
 
 let asTyped = (() => '') as () => string;
-let assertTyped = <() => string>(() => '');
 
 interface ObjectType {
   foo(): number;
@@ -172,9 +171,6 @@ let objectProp: ObjectType = {
 let objectPropAs = {
   foo: () => 1,
 } as ObjectType;
-let objectPropAssert = <ObjectType>{
-  foo: () => 1,
-};
 
 declare function functionWithArg(arg: () => number);
 functionWithArg(() => 1);
--- a/packages/eslint-plugin/docs/rules/explicit-function-return-type.mdx
+++ b/packages/eslint-plugin/docs/rules/explicit-function-return-type.mdx
@@ -172,7 +172,7 @@ let objectProp: ObjectType = {
 let objectPropAs = {
   foo: () => 1,
 } as ObjectType;
-let objectPropAssertion = <ObjectType>{
+let objectPropAssert = <ObjectType>{
   foo: () => 1,
 };
 
--- a/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.mdx
+++ b/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.mdx
@@ -176,7 +176,7 @@ Also, if you would like to ignore all primitives types, you can set `ignorePrimi
 
 {/* insert option description */}
 
-Whether to ignore expressions that cast a value into a boolean: `Boolean(...)`.
+Whether to ignore expressions that coerce a value into a boolean: `Boolean(...)`.
 
 Incorrect code for `ignoreBooleanCoercion: false`, and correct code for `ignoreBooleanCoercion: true`:
 
--- a/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.mdx
+++ b/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.mdx
@@ -176,7 +176,7 @@ Also, if you would like to ignore all primitives types, you can set `ignorePrimi
 
 {/* insert option description */}
 
-Whether to ignore expressions that coerce a value into a boolean: `Boolean(...)`.
+Whether to ignore expressions that cast a value into a boolean: `Boolean(...)`.
 
 Incorrect code for `ignoreBooleanCoercion: false`, and correct code for `ignoreBooleanCoercion: true`:
 
--- a/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
+++ b/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
@@ -253,7 +253,7 @@ export let objectProp: ObjectType = {
 export let objectPropAs = {
   foo: () => 1,
 } as ObjectType;
-export let objectPropAssertion = <ObjectType>{
+export let objectPropAssert = <ObjectType>{
   foo: () => 1,
 };
 
--- a/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
+++ b/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
@@ -242,7 +242,7 @@ export let funcExpr: FuncType = function () {
 };
 
 export let asTyped = (() => '') as () => string;
-export let castTyped = <() => string>(() => '');
+export let assertTyped = <() => string>(() => '');
 
 interface ObjectType {
   foo(): number;
@@ -253,7 +253,7 @@ export let objectProp: ObjectType = {
 export let objectPropAs = {
   foo: () => 1,
 } as ObjectType;
-export let objectPropCast = <ObjectType>{
+export let objectPropAssertion = <ObjectType>{
   foo: () => 1,
 };
 
--- a/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
+++ b/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
@@ -242,7 +242,6 @@ export let funcExpr: FuncType = function () {
 };
 
 export let asTyped = (() => '') as () => string;
-export let assertTyped = <() => string>(() => '');
 
 interface ObjectType {
   foo(): number;
@@ -253,9 +252,6 @@ export let objectProp: ObjectType = {
 export let objectPropAs = {
   foo: () => 1,
 } as ObjectType;
-export let objectPropAssert = <ObjectType>{
-  foo: () => 1,
-};
 
 type FooType = (bar: string) => void;
 export const foo: FooType = bar => {};
--- a/packages/eslint-plugin/docs/rules/non-nullable-type-assertion-style.mdx
+++ b/packages/eslint-plugin/docs/rules/non-nullable-type-assertion-style.mdx
@@ -15,7 +15,7 @@ There are two common ways to assert to TypeScript that a value is its type witho
 - `as`: Traditional type assertion with a coincidentally equivalent type
 
 `!` non-null assertions are generally preferred for requiring less code and being harder to fall out of sync as types change.
-This rule reports when an `as` cast is doing the same job as a `!` would, and suggests fixing the code to be an `!`.
+This rule reports when an `as` assertion is doing the same job as a `!` would, and suggests fixing the code to be an `!`.
 
 ## Examples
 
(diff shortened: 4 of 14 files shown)


packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.mdx
```diff
@@ -176,7 +176,7 @@ Also, if you would like to ignore all primitives types, you can set `ignorePrimi
 
 {/* insert option description */}
 
-Whether to ignore expressions that coerce a value into a boolean: `Boolean(...)`.
+Whether to ignore expressions that cast a value into a boolean: `Boolean(...)`.
```
kirkwaiblinger, line 179: Let's leave this this as "coerce".

kirkwaiblinger (changes requested):
Noting that there are a fair number of problematic usages of "cast" remaining that we'll want to address still, either here or in a followup

ronami:
> Noting that there are a fair number of problematic usages of "cast" remaining that we'll want to address still, either here or in a followup

I've just realized my VSCode had match-case and match-whole-word locked in (🙈). I think I've managed to update every usage of `cast` that refers to a type assertion.

packages/eslint-plugin/docs/rules/explicit-function-return-type.mdx
```diff
@@ -161,7 +161,7 @@ let funcExpr: FuncType = function () {
 };
 
 let asTyped = (() => '') as () => string;
-let castTyped = <() => string>(() => '');
+let assertTyped = <() => string>(() => '');
```
ronami, line 164: I'm unsure about these documentation variable changes.

ronami, line 164: I've tried to somewhat match the wording on the `asTyped` and `objectPropAs` variable names. Since the AST-node name for angular bracket assertion is `TSTypeAssertion`, I think it makes sense to use the assert wording here.

kirkwaiblinger:
Looks like there's still two more places where "cast" is used in user-facing strings...

https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L71


https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L103

... and then also here, though I'm honestly not sure why we even have this:

https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/docs/rules/strict-boolean-expressions.mdx#L171-L201

packages/eslint-plugin/docs/rules/explicit-function-return-type.mdx
kirkwaiblinger: Honestly, what if we just remove those examples? 🤔 Angle-bracket type assertions are yucky, and we don't really want to encourage them. Even with the name "castTyped"/"assertTyped" it took me longer than I care to admit to make sense of those examples 😆

kirkwaiblinger (commented):
couple minor comments

ronami:
> Looks like there's still two more places where "cast" is used in user-facing strings...
> 
> https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L71
> 
> https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L103
> 
> ... and then also here, though I'm honestly not sure why we even have this:
> 
> https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/docs/rules/strict-boolean-expressions.mdx#L171-L201

I've noticed these, too, but to my understanding, the meaning in those cases is actually to cast rather than to assert.

They all require the developer to make a runtime change rather than a type-only assertion. For example, these two reports ([1](https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L103), [2](https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L71)) include a suggestion to "cast" a value to a boolean with `Boolean(value)`.

I think these are similar to the coerce meaning in https://github.com/typescript-eslint/typescript-eslint/pull/10472#discussion_r1874544777.

kirkwaiblinger:
> I've noticed these, too, but to my understanding, the meaning in those cases is actually to cast rather than to assert.
> 
> They all require the developer to make a runtime change rather than a type-only assertion. For example, these two reports ([1](https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L103), [2](https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L71)) include a suggestion to "cast" a value to a boolean with `Boolean(value)`.
> 
> I think these are similar to the coerce meaning in [#10472 (comment)](https://github.com/typescript-eslint/typescript-eslint/pull/10472#discussion_r1874544777), and also to the meaning of casting/coercing in [no-extra-boolean-cast](https://eslint.org/docs/latest/rules/no-extra-boolean-cast).
> 
> I'm not sure which terminology to use in those cases. There are some places we use the term coercing, some casting, and here ([open PR](https://github.com/typescript-eslint/typescript-eslint/pull/10182#issuecomment-2453282643)) we use converting.

Gotcha, yeah, I see where you're coming from. My spiel on this is: within a _pure JS context_, one can (just barely) get away with using "cast" as in [no-extra-boolean-cast](https://eslint.org/docs/latest/rules/no-extra-boolean-cast). IMO it's still not a good term for what it's describing, but it is at least unambiguous, since "cast" has no other possible meaning within JS. But, throw _TS_ into the mix, and "cast" means different things to different people, and even sometimes different things to the same people in different contexts 😆. Therefore, I don't think we should use it for either the "type assertion" or "type conversion/coersion" meanings at all in a TS context.

So, then, to your specific question, I think we should change those terms in SBE, and we should change them to "convert"/"conversion". My vague idea of "coercion" vs "conversion" is that "coercion" is what happens when the language performs a type conversion _implicitly_, usually because objects of incompatible types are interacting, like `true + ''`. So I'd think of `Boolean(x)` as a conversion but not a coercion (but all coercions are conversions). Don't know if that's quite right, though; @Josh-Cena is generally our expert on these sorts of things!

Incidentally, regarding https://github.com/typescript-eslint/typescript-eslint/pull/10472#discussion_r1874544777, that does imply that the word "convert" would be most in line with my previous paragraph. But, the option name is `ignoreBooleanCoercion`, so I think our hands are tied to use "coerce" in the description.

https://github.com/typescript-eslint/typescript-eslint/pull/10472/files/34610c6a07b3edc3acde8ae158a0f7472ced0f53#diff-a1ede3fb0351c7f188d7e9fcc7b931a865ebb2a0197b2323b3e104623812f76dR175-R179

ronami:
> > I've noticed these, too, but to my understanding, the meaning in those cases is actually to cast rather than to assert.
> > They all require the developer to make a runtime change rather than a type-only assertion. For example, these two reports ([1](https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L103), [2](https://github.com/typescript-eslint/typescript-eslint/blob/d09d36de88dbef86575252a0542e92801fb6a890/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L71)) include a suggestion to "cast" a value to a boolean with `Boolean(value)`.
> > I think these are similar to the coerce meaning in [#10472 (comment)](https://github.com/typescript-eslint/typescript-eslint/pull/10472#discussion_r1874544777), and also to the meaning of casting/coercing in [no-extra-boolean-cast](https://eslint.org/docs/latest/rules/no-extra-boolean-cast).
> > I'm not sure which terminology to use in those cases. There are some places we use the term coercing, some casting, and here ([open PR](https://github.com/typescript-eslint/typescript-eslint/pull/10182#issuecomment-2453282643)) we use converting.
> 
> Gotcha, yeah, I see where you're coming from. My spiel on this is: within a _pure JS context_, one can (just barely) get away with using "cast" as in [no-extra-boolean-cast](https://eslint.org/docs/latest/rules/no-extra-boolean-cast). IMO it's still not a good term for what it's describing, but it is at least unambiguous, since "cast" has no other possible meaning within JS. But, throw _TS_ into the mix, and "cast" means different things to different people, and even sometimes different things to the same people in different contexts 😆. Therefore, I don't think we should use it for either the "type assertion" or "type conversion/coersion" meanings at all in a TS context.
> 
> So, then, to your specific question, I think we should change those terms in SBE, and we should change them to "convert"/"conversion". My vague idea of "coercion" vs "conversion" is that "coercion" is what happens when the language performs a type conversion _implicitly_, usually because objects of incompatible types are interacting, like `true + ''`. So I'd think of `Boolean(x)` as a conversion but not a coercion (but all coercions are conversions). Don't know if that's quite right, though; @Josh-Cena is generally our expert on these sorts of things!
> 
> Incidentally, regarding [#10472 (comment)](https://github.com/typescript-eslint/typescript-eslint/pull/10472#discussion_r1874544777), that does imply that the word "convert" would be most in line with my previous paragraph. But, the option name is `ignoreBooleanCoercion`, so I think our hands are tied to use "coerce" in the description.
> 
> https://github.com/typescript-eslint/typescript-eslint/pull/10472/files/34610c6a07b3edc3acde8ae158a0f7472ced0f53#diff-a1ede3fb0351c7f188d7e9fcc7b931a865ebb2a0197b2323b3e104623812f76dR175-R179

Yes! I can't agree more about how confusing "to cast" is, especially with TS, as it can have so many meanings.

I've updated the PR and adjusted the "to cast" usage in SBE, too; thanks!

packages/eslint-plugin/docs/rules/explicit-function-return-type.mdx
ronami: Yeah, I was confused by some of these examples too. I don't think the additional angle bracket assertion adds any value, either.

I'll be happy to remove them. Should I push it as part of this PR?

ronami: Done 👍

kirkwaiblinger (approved):
Love it! Thank you for doing this!! This had been bugging me for a while 😆

JoshuaKGoldberg (approved):
Lovely!

Issue #10438, opened 2024-12-01
Repo: Investigate better handling of unconstrained generics


kirkwaiblinger:
### Suggestion

Sadly, right now `getConstrainedTypeOfLocation` is bugged for unconstrained generics. The code just gives back the type parameter type if getBaseConstraintOfType(nodeType) is nullish, but that's not correct. Unconstrained generics should behave like `unknown`. Here's what it is right now:

https://github.com/typescript-eslint/typescript-eslint/blob/62d57559564fb08512eafe03a2c1b167c4377601/packages/type-utils/src/getConstrainedTypeAtLocation.ts#L7-L20

And here's what it "should" be: 

```ts
/**
 * Resolves the given node's type. Will resolve to the type's generic constraint, if it has one.
 */
export function getConstrainedTypeAtLocation(
  services: ParserServicesWithTypeInformation,
  node: TSESTree.Node,
): ts.Type {
  const nodeType = services.getTypeAtLocation(node);
  if (tsutils.isTypeParameter(nodeType)) {
    const checker = services.program.getTypeChecker();

    return (
      checker.getBaseConstraintOfType(nodeType) ?? checker.getUnknownType()
    );
  }
  return nodeType;
}
```

Unfortunately, due to https://github.com/microsoft/TypeScript/issues/60475, there is no `checker.getUnknownType()` yet or for a while.

It kind of sucks, but I propose that we instead modify the implementation to something more like
```ts
export function getConstrainedTypeAtLocation(
  services: ParserServicesWithTypeInformation,
  node: TSESTree.Node,
): ts.Type | undefined { // <-- note the `| undefined`
  const nodeType = services.getTypeAtLocation(node);
  if (tsutils.isTypeParameter(nodeType)) {
    const checker = services.program.getTypeChecker();
    return checker.getBaseConstraintOfType(nodeType) 
    /* in the future this can have ?? checker.getUnknownType() */;
  }
  return nodeType;
}
```
and force callers to handle the unconstrained case separately.

***

Note - I think we should also double-check (and add tests) for what happens when the constraint is `any`. I'm pretty sure `function f<T extends any>(x: T) {}` behaves as though `T` were `unknown` rather than `any`. So maybe we should really be doing

```ts
export function getConstrainedTypeAtLocation(
  services: ParserServicesWithTypeInformation,
  node: TSESTree.Node,
): ts.Type | undefined {
  const nodeType = services.getTypeAtLocation(node);
[...]

PR #10473, opened 2024-12-07
fix(eslint-plugin): [no-unnecessary-condition] don't flag values of an unconstrained or valid type parameter

### LINKED ISSUE #10442: Bug: [no-unnecessary-condition] don't flag values of an unconstrained or valid type parameter

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3hkmCMAXPABQCU8AvAHzwBGSSEIUKA3AFCiSwLLpY5MAJiJlKNAK4oA1iiQB3Np3DQ4iVJmy4MAZiIAeACpUx1eIY5dVvDQO0AWA4fggAHjhTA0dBkxYnyMwt2dgB6UPgAByg0NFx2AG08ABp4FAkICABdADoAMwBLCBwYYkEMPFIOMIjo2Pik1PTM3MLikFLyoSqQ8Pg0AAskDOAomLRElLSM7PyikrLtHR6a-qGRsdjJppnW%2BY7FzHseoA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK%2BSpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
declare const test1: () => boolean;
declare const test2: () => unknown;
declare const test3: <T>() => T;
declare const test4: <T extends boolean>() => T;

// passes
[1, null].filter(test1);

// passes
[1, null].filter(test2);

// should pass
[1, null].filter(test3);

// should pass
[1, null].filter(test4);
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expect all use-cases to be consistent and not report as errors.

### Actual Result

The use-cases with a type constraint report (unnecessarily) as an error.

### Additional Info

This seems similar to #10311 and looks like the fix should be similar.

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10442

## Overview

This PR addresses #10442 and handles a potential return type being a type constraint.

The PR suffers from the same issues around `getConstrainedTypeAtLocation` described by @kirkwaiblinger on #10438 and requires a small workaround.

Note that `no-unnecessary-condition` makes a lot of use of `getConstrainedTypeAtLocation`, but going over them roughly, I couldn't find additional bugs around unconstrained type parameters.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -648,30 +648,41 @@ export default createRule<Options, MessageId>({
           )
           .map(sig => sig.getReturnType())
           .map(t => {
+            // TODO: use `getConstraintTypeInfoAtLocation` once it's merged
+            // https://github.com/typescript-eslint/typescript-eslint/pull/10496
             if (tsutils.isTypeParameter(t)) {
               return checker.getBaseConstraintOfType(t);
             }
 
             return t;
           });
-        /* istanbul ignore if */ if (returnTypes.length === 0) {
-          // Not a callable function
-          return;
+
+        let hasFalsyReturnTypes = false;
+        let hasTruthyReturnTypes = false;
+
+        for (const type of returnTypes) {
+          // Predicate is always necessary if it involves `any` or `unknown`
+          if (!type || isTypeAnyType(type) || isTypeUnknownType(type)) {
+            return;
+          }
+
+          if (isPossiblyFalsy(type)) {
+            hasFalsyReturnTypes = true;
+          }
+
+          if (isPossiblyTruthy(type)) {
+            hasTruthyReturnTypes = true;
+          }
         }
-        // Predicate is always necessary if it involves `any` or `unknown`
-        if (
-          // https://github.com/typescript-eslint/typescript-eslint/issues/10438
-          returnTypes.some(t => !t || isTypeAnyType(t) || isTypeUnknownType(t))
-        ) {
-          return;
-        }
-        if (!returnTypes.some(t => t && isPossiblyFalsy(t))) {
+
+        if (!hasFalsyReturnTypes) {
           return context.report({
             node: callback,
             messageId: 'alwaysTruthyFunc',
           });
         }
-        if (!returnTypes.some(t => t && isPossiblyTruthy(t))) {
+
+        if (!hasTruthyReturnTypes) {
           return context.report({
             node: callback,
             messageId: 'alwaysFalsyFunc',
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -673,6 +673,11 @@ export default createRule<Options, MessageId>({
           if (isPossiblyTruthy(type)) {
             hasTruthyReturnTypes = true;
           }
+
+          // bail early if both a possibly-truthy and a possibly-falsy have been detected
+          if (hasFalsyReturnTypes && hasTruthyReturnTypes) {
+            return;
+          }
         }
 
         if (!hasFalsyReturnTypes) {
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -657,6 +657,11 @@ export default createRule<Options, MessageId>({
             return t;
           });
 
+        if (returnTypes.length === 0) {
+          // Not a callable function, e.g. `any`
+          return;
+        }
+
         let hasFalsyReturnTypes = false;
         let hasTruthyReturnTypes = false;
 
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -646,22 +646,32 @@ export default createRule<Options, MessageId>({
           .getCallSignaturesOfType(
             getConstrainedTypeAtLocation(services, callback),
           )
-          .map(sig => sig.getReturnType());
+          .map(sig => sig.getReturnType())
+          .map(t => {
+            if (tsutils.isTypeParameter(t)) {
+              return checker.getBaseConstraintOfType(t);
+            }
+
+            return t;
+          });
         /* istanbul ignore if */ if (returnTypes.length === 0) {
           // Not a callable function
           return;
         }
         // Predicate is always necessary if it involves `any` or `unknown`
-        if (returnTypes.some(t => isTypeAnyType(t) || isTypeUnknownType(t))) {
+        if (
+          // https://github.com/typescript-eslint/typescript-eslint/issues/10438
+          returnTypes.some(t => !t || isTypeAnyType(t) || isTypeUnknownType(t))
+        ) {
           return;
         }
-        if (!returnTypes.some(isPossiblyFalsy)) {
+        if (!returnTypes.some(t => t && isPossiblyFalsy(t))) {
           return context.report({
             node: callback,
             messageId: 'alwaysTruthyFunc',
           });
         }
-        if (!returnTypes.some(isPossiblyTruthy)) {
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -646,22 +646,32 @@ export default createRule<Options, MessageId>({
...
         // Predicate is always necessary if it involves `any` or `unknown`
-        if (returnTypes.some(t => isTypeAnyType(t) || isTypeUnknownType(t))) {
+        if (
+          // https://github.com/typescript-eslint/typescript-eslint/issues/10438
+          returnTypes.some(t => !t || isTypeAnyType(t) || isTypeUnknownType(t))
+        ) {
           return;
         }
-        if (!returnTypes.some(isPossiblyFalsy)) {
+        if (!returnTypes.some(t => t && isPossiblyFalsy(t))) {
```
ronami, line 668: Checking `t` here is unnecessary (as the `returnTypes.some` check above would return), but I preferred having this over a type assertion.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -646,22 +646,32 @@ export default createRule<Options, MessageId>({
...
             getConstrainedTypeAtLocation(services, callback),
           )
-          .map(sig => sig.getReturnType());
+          .map(sig => sig.getReturnType())
+          .map(t => {
+            if (tsutils.isTypeParameter(t)) {
+              return checker.getBaseConstraintOfType(t);
+            }
+
+            return t;
```
ronami, line 655: Optimally, we could use `getConstrainedTypeAtLocation` here once https://github.com/typescript-eslint/typescript-eslint/issues/10438 is tackled.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
JoshuaKGoldberg: -1 from me, since we know `t` is truthy the extra runtime `t &&` check is confusing.

I also find it kind of clunky that there are three `.some`s on the same array in rapid succession... and also a five-year-old `/* istanbul ignore if */` just before them. 

WDYT about something like this?

```ts
let reportMessage: MessageId | undefined;

for (const type of returnTypes) {
  if (!type || isTypeAnyType(type) || isTypeUnknownType(type)) {
    return;
  }

  if (isPossiblyFalsy(type)) {
    reportMessage = 'alwaysTruthyFunc';
  } else if (isPossiblyTruthy(type)) {
    reportMessage = 'alwaysFalsyFunc';
  }
}

if (reportMessage) {
  context.report({ node: callback, messageId: reportMessage });
}
```

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
JoshuaKGoldberg: 👍 agreed, that'd be nice. Suggestion: I'd put a TODO comment here so it's not lost.

JoshuaKGoldberg (changes requested):
Nice! Just requesting changes on a bit of cleanup for the older code.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
ronami: Yes! That's significantly easier to understand 👍 

I've made some changes since it doesn't exactly match the current implementation (a couple of failing tests). The existing code is somewhat confusing with: https://github.com/typescript-eslint/typescript-eslint/blob/ce5d3e3b5b1c2d85482c9ee6f66fed59fe04d584/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts#L658-L669

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -646,22 +646,43 @@ export default createRule<Options, MessageId>({
...
+            // https://github.com/typescript-eslint/typescript-eslint/pull/10496
+            if (tsutils.isTypeParameter(t)) {
+              return checker.getBaseConstraintOfType(t);
+            }
+
+            return t;
+          });
+
+        let hasFalsyReturnTypes = false;
+        let hasTruthyReturnTypes = false;
```
JoshuaKGoldberg, line 661: One more optimization: we know that we won't care about possibly truthy types if there was a possibly falsy type. So this can become:

```suggestion
        let knownReturnType: 'falsy' | 'truthy' | undefined;
```

...and the truthiness check below could be:

```ts
} else if (knownReturnType !== 'falsy' && isPossiblyTruthy(type)) {
  knownReturnType = 'truthy';
}
```

JoshuaKGoldberg (changes requested):
One last touchup, since you're humoring me 😄

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
ronami: I like the suggested optimization, but I don't think the proposed change matches the current implementation (some tests fail unless I've missed something).

To my understanding, the current code can be changed to the following so that as soon as both a truthy and a falsy values exist, the function bails:

```diff
for (const type of returnTypes) {
  // Predicate is always necessary if it involves `any` or `unknown`
  if (!type || isTypeAnyType(type) || isTypeUnknownType(type)) {
    return;
  }

  if (isPossiblyFalsy(type)) {
    hasFalsyReturnTypes = true;
  }

  if (isPossiblyTruthy(type)) {
    hasTruthyReturnTypes = true;
  }

+  if (hasFalsyReturnTypes && hasTruthyReturnTypes) {
+    return;
+  }
}
```

I think the same value can be possibly-truthy and possibly-falsy at the same time (e.g. `boolean`), so both checks have to run for every return type.

I think matching the current implementation with a single value state (`let knownReturnType: 'falsy' | 'truthy' | undefined`) will require potentially nested if statements, as even something like the following fails to catch some cases (like the returned `boolean` type):

```ts
if (isPossiblyFalsy(type) && knownReturnType !== 'truthy') {
  knownReturnType = 'falsy';
}

if (isPossiblyTruthy(type) && knownReturnType !== 'falsy') {
  knownReturnType = 'truthy';
}
```

I'll be happy to update with the above diff if it makes sense or something different if I'm missing something.

JoshuaKGoldberg: Gotcha - that works!

JoshuaKGoldberg (approved):
LGTM! I think this is ready if you are ✨

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
```diff
@@ -646,22 +646,48 @@ export default createRule<Options, MessageId>({
           .getCallSignaturesOfType(
             getConstrainedTypeAtLocation(services, callback),
           )
-          .map(sig => sig.getReturnType());
-        /* istanbul ignore if */ if (returnTypes.length === 0) {
-          // Not a callable function
-          return;
```
kirkwaiblinger, line 652: minor regression.... which I _think_ hit this line previously, but I haven't validated that.
```ts
declare const lol: any;
[1, 2, 3].filter(lol); // should not flag
```

kirkwaiblinger: Sweet! Thanks! Yeah, IIRC `any` and `Function` are the two types I know of that are permitted to be called without any TS errors, but don't have any call signatures. Some obscure linting trivia 🤷‍♂️

kirkwaiblinger (approved):
🚀 

Bug is solved and the existing code is better for it, too. 
yay

PR #10477, opened 2024-12-09
fix: ignore when constructor is  typed array

### LINKED ISSUE #10445: Bug: [consistent-generic-constructors] <short description of the issue>

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=DYUwLgBAhgXBCqBLAdmAHAQQE5agTwB5tc8AhAVwDNKQsAZRAaxAD4IBeCZEAdwRXTF8ACgCUAbgBQANwD2iACbQpQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6MgeyeUuX0Ra0A5gMRUytTt3xwy%2BDtFQZIY6AsjgwAXxBagA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkqyARhWANqTokBMADJwMyQAugBpwNSPFgATAK4Zm7Lr06RRNKCTyZk0PADk5qAMIALdNADWZSlpno14vAENMAc3R4FHHv0hiAviD%2BQA&tokens=false

### Repro Code

```TypeScript
let a: Uint8Array<ArrayBufferLike> = new Uint8Array();
void a;
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/consistent-generic-constructors": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "lib": ["es2023"],
    "module": "es2022",
    "strictNullChecks": true,
    "target": "es2023"
  }
}
```

### Expected Result

Either don't report the error, or don't generate invalid code upon fix.

### Actual Result

Rule reports error:
> The generic type arguments should be specified as part of the constructor type arguments. 1:5 - 1:54

If I tap **Fix**, the code becomes:

```ts
let a = new Uint8Array<ArrayBufferLike>();
void a;
```

And then TypeScript reports error:

> 2554: Expected 1-3 arguments, but got 0. 1:9 - 1:42

### Additional Info

If I enable two rules consistent-generic-constructors and no-unnecessary-type-arguments, [playground](https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=DYUwLgBAhgXBCqBLAdmAHAQQE5agTwB5tc8AhAVwDNKQsAZRAaxAD4IBeCZEAdwRXTF8ACgCUAbgBQANwD2iACbQpQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6MgeyeUuX0Ra0A5gMRUytTt3xwy%2BDtFQZIY6AsgAacNjxFSyCtToNmbJh1qwmTRGRTIAhtEK09iWk6GwAtgPxKoVXVtAF8QEKA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkqyARhWANqTokBMADJwMyQAugBpwNSPFgATAK4Zm7Lr06RRNKCTyZk0PADk5qAMIALdNADWZSlpno14vAENMAc3R4FHHv0hiAviD%2BQA&tokens=false), both rules would report errors.
The outcome depends on the order of fix.

* If consistent-generic-constructors is fixed first, the above problem occurs.
* If no-unnecessary-type-arguments is fixed first, the resulting code is well-formed.

mdm317:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10445

## Overview

Hard-code the typed arrays and added code to skip check if the type is a typed array.

### DIFF
--- a/packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
+++ b/packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
@@ -33,6 +33,30 @@ export default createRule<Options, MessageIds>({
   },
   defaultOptions: ['constructor'],
   create(context, [mode]) {
+    const isTypedArrayReference = (typeNode: TSESTree.TypeNode) => {
+      const typedArrayNames = [
+        'Int8Array',
+        'Uint8Array',
+        'Uint8ClampedArray',
+        'Int16Array',
+        'Uint16Array',
+        'Int32Array',
+        'Uint32Array',
+        'BigInt64Array',
+        'BigUint64Array',
+        'Float32Array',
+        'Float64Array',
+      ];
+      if (
+        typeNode.type === AST_NODE_TYPES.TSTypeReference &&
+        typeNode.typeName.type === AST_NODE_TYPES.Identifier &&
+        typedArrayNames.includes(typeNode.typeName.name)
+      ) {
+        return true;
+      }
+      return false;
+    };
+
     return {
       'VariableDeclarator,PropertyDefinition,:matches(FunctionDeclaration,FunctionExpression) > AssignmentPattern'(
         node:
@@ -59,6 +83,12 @@ export default createRule<Options, MessageIds>({
         }
         const [lhsName, rhs] = getLHSRHS();
         const lhs = lhsName.typeAnnotation?.typeAnnotation;
+
+        // If it's a typed array, we will ignore.
+        // https://github.com/typescript-eslint/typescript-eslint/issues/10445
+        if (lhs && isTypedArrayReference(lhs)) {
+          return;
+        }
 
         if (
           !rhs ||
--- a/packages/eslint-plugin/tests/rules/consistent-generic-constructors.test.ts
+++ b/packages/eslint-plugin/tests/rules/consistent-generic-constructors.test.ts
@@ -40,6 +40,16 @@ class A {
     `
 const a = function (a: Foo = new Foo<string>()) {};
     `,
+    'let a: Uint8Array<ArrayBufferLike> = new Uint8Array();',
+    'let a: Uint8ClampedArray<ArrayBufferLike> = new Uint8ClampedArray();',
+    'let a: Int16Array<ArrayBufferLike> = new Int16Array();',
+    'let a: Uint16Array<ArrayBufferLike> = new Uint16Array();',
+    'let a: Int32Array<ArrayBufferLike> = new Int32Array();',
+    'let a: Uint32Array<ArrayBufferLike> = new Uint32Array();',
+    'let a: BigInt64Array<ArrayBufferLike> = new BigInt64Array();',
+    'let a: BigUint64Array<ArrayBufferLike> = new BigUint64Array();',
+    'let a: Float32Array<ArrayBufferLike> = new Float32Array();',
+    'let a: Float64Array<ArrayBufferLike> = new Float64Array();',
     // type-annotation
     {
       code: 'const a = new Foo();',


packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
```diff
@@ -33,6 +33,30 @@ export default createRule<Options, MessageIds>({
...
+      ];
+      if (
+        typeNode.type === AST_NODE_TYPES.TSTypeReference &&
+        typeNode.typeName.type === AST_NODE_TYPES.Identifier &&
+        typedArrayNames.includes(typeNode.typeName.name)
+      ) {
+        return true;
+      }
+      return false;
+    };
```
JoshuaKGoldberg, line 58: [Bug] It's not enough to just look at the name. If someone happens to redeclare their own class, such as `class Uint8Array<T>`, the rule should know to check that:

https://deploy-preview-10477--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.ts&code=MYGwhgzhAECqCWA7ALgDgIICdNgJ4B4AVAPmgG8AoaaAehugDomKBfCikAU2WjAC44SNFhwEReAEIBXAGYzOmADLwA1p1IBeaIk4B3QSgzY8ACgCUAbnacAHgAcA9ph5kWQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6MgeyeUuX0Ra0A5gMRUytTt3xwy%2BDtFQZIY6AsjgwAXxBagA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkqyARhWANqTokBMADJwMyQAugBpwNSPFgATAK4Zm7Lr06RRNKCTyZk0PADk5qAMIALdNADWZSlpno14vAENMAc3R4FHHv0hiAviD%2BQA&tokens=false

```ts
class Uint8Array<T> {
  // ...
}

let a: Uint8Array<ArrayBufferLike> = new Uint8Array();

export {}
```

You'll want to use the scope manager to see whether the identifier is referencing something declared or imported in the file vs. a global.

packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
```diff
@@ -33,6 +33,30 @@ export default createRule<Options, MessageIds>({
   },
   defaultOptions: ['constructor'],
   create(context, [mode]) {
+    const isTypedArrayReference = (typeNode: TSESTree.TypeNode) => {
```
JoshuaKGoldberg, line 36: [Style] It's probably optimized away by engines, but to make the code more clear, a few suggestions:

* Move the `typedArrayNames` to a single declared array outside the `createRule`, so it doesn't look like it's declared differently on each check
* Use a `Set` for O(1) `.has`, which is really what's being done
* There only needs to be one `return`

I.e. let's turn this function into:

```ts
function isTypedArrayReference(typeNode: TSESTree.TypeNode) {
  return (
    typeNode.type === AST_NODE_TYPES.TSTypeReference &&
    typeNode.typeName.type === AST_NODE_TYPES.Identifier &&
    typedArrayNames.has(typeNode.typeName.name)
  );
}
```

JoshuaKGoldberg (changes requested):
A good start! 🚀

Josh-Cena:
FWIW, my personal preference is still "we should not special case built-ins". These types of declarations can legitimately happen with user-defined types, so we should just create a generic allowlist of types.

mdm317:
Should we avoid specifically handling the built-in array-likes?

JoshuaKGoldberg:
@Josh-Cena just confirming, what do you suggest we do here?

Josh-Cena:
This:

> we should just create a generic allowlist of types.

Like a new option called `ignore` or `allow` where the types would just be ignored for consistency checks.

mdm317:
> Like a new option called ignore or allow where the types would just be ignored for consistency checks.

I added an allow option to ignore consistency checks.

Should I add TypedArray to the default option?

packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
JoshuaKGoldberg: This is still open, no?

packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
```diff
@@ -29,10 +34,24 @@ export default createRule<Options, MessageIds>({
...
         enum: ['type-annotation', 'constructor'],
       },
+      {
+        type: 'object',
+        additionalProperties: false,
+        properties: {
+          ignore: {
+            type: 'array',
+            description:
+              'A list of constructor names to ignore when enforcing the rule.',
```
JoshuaKGoldberg, line 44: IMO if we do go with a list then we'd want to use [TypeOrValueSpecifier](https://typescript-eslint.io/packages/type-utils/type-or-value-specifier). Otherwise it'll have the same issue as https://github.com/typescript-eslint/typescript-eslint/pull/10477#discussion_r1885099562 with coincidentally matching names.

But, I'm not sure this is the right step. I started a thread later in this file about it.

packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
```diff
@@ -77,7 +96,8 @@ export default createRule<Options, MessageIds>({
           lhs &&
           (lhs.type !== AST_NODE_TYPES.TSTypeReference ||
             lhs.typeName.type !== AST_NODE_TYPES.Identifier ||
-            lhs.typeName.name !== rhs.callee.name)
+            lhs.typeName.name !== rhs.callee.name ||
+            options?.ignore?.includes(lhs.typeName.name))
```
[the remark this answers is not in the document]
Josh-Cena: > * What benefit is there to reporting on `Uint8Array` and similar? I.e.: in what situations would one want the rule to report on them?

There's none; I would expect them to be ignored by default.

> If there are none, why not hardcode the rule to always ignore them?

The _rule_ should be generic. Any library-specific logic (even those related to native objects) should be injected via options. Built-in objects should not be distinguishable from user-defined objects.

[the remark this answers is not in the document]
JoshuaKGoldberg: Ok great I think we're aligned? Just confirming, I'm of the opinion that:

* The rule should not report those things by default
* The list should be empty (`[]`) by default

...so the rule should not provide a way to now report on those built-ins?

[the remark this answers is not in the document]
JoshuaKGoldberg: Following up: for each of those points, generally I think the team is either neutral or on board. 👍

JoshuaKGoldberg:
@mdm317 sorry for the delay on this! I think there's consensus on what to do in https://github.com/typescript-eslint/typescript-eslint/pull/10477#discussion_r2095694203. Is there anything else blocking you on this PR?

mdm317:
@JoshuaKGoldberg 
All good, no blockers. Appreciate you driving the consensus, I’ll resolve the merge conflicts on my side.

packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
```diff
@@ -29,10 +34,24 @@ export default createRule<Options, MessageIds>({
...
+          ignore: {
+            type: 'array',
+            description:
+              'A list of constructor names to ignore when enforcing the rule.',
+            items: {
+              type: 'string',
+            },
+          },
+        },
+      },
```
JoshuaKGoldberg: [Refactor] Oh, sorry, I should have noted this earlier: we've been moving away from array-style rule options (`['error', 'some-string', ...]`) and towards full objects (`['error', { option: 'some-string', ... }]`). For rules like this one that already have an array-style option, I think we'd want to switch to an `anyOf` union.

```js
{
  anyOf: [
    {
      type: "array",
      items: [
        {
          type: "string",
          description: "Which constructor call syntax to prefer.",
          enum: ["type-annotation", "constructor"],
        },
      ],
    },
    {
      type: "object",
      properties: {
        allow: {
          description: `Types to allow in template expressions.`,
          ...typeOrValueSpecifiersSchema,
        },
        style: {
          description: "Which constructor call syntax to prefer.",
          enum: ["constructor", "type-annotation"],
        },
      },
      additionalProperties: false,
    },
  ];
}
```

_(roughly, you might need to clean that syntax up a bit - I haven't verified it's exactly this)_

I realize we've already taken a while to discuss other big changes on this PR, and this request is probably annoying. I can go ahead and do it myself if you want to focus on other things. Feel free to ignore this comment to opt into that. 😅

JoshuaKGoldberg (changes requested):
Current status is waiting on:

* Having the known _"can't be fixed"_ built-in lib types always be allowed
* Implementing the `TypeOrValueSpecifier` change

packages/eslint-plugin/src/rules/consistent-generic-constructors.ts
mdm317: @JoshuaKGoldberg 
it’s not annoying at all! I really appreciate the offer. 
If it’s not too much trouble, I’d be grateful if you could handle that part, since I’m not very familiar with the `TypeOrValueSpecifier` part yet.

JoshuaKGoldberg:
Ok, looking back at this, I figured out the discrepancy - we'd only 👍d hardcoding the known built-in array classes as ignored. Adding a TypeOrValueSpecifier ignore option is out-of-scope because that format requires type information. This rule isn't typed right now.

Sorry for the confusion. I updated the code to just hardcode allowing the types mentioned in https://github.com/microsoft/TypeScript/pull/59417/files.

bradzacher (approved):

connorshea:
Was it intentional that this list omits BigInt64Array and BigUint64Array?

PR #10481, opened 2024-12-10
chore: fix incorrect licenses

alan-agius4:
`MITClause` is not valid, it should be `MIT` instead.

This is blocking us to update due to invalid license.


JamesHenry (approved):

JamesHenry:
Thanks a lot @alan-agius4!

PR #10490, opened 2024-12-12
fix(eslint-plugin): [consistent-indexed-object-style] use a suggestion over an auto-fix if can't reliably determine that produced index signature is valid

### LINKED ISSUE #4730: [consistent-indexed-object-style] fixer produces invalid code and deletes comments

**Repro**

[Playground](https://typescript-eslint.io/play#ts=5.3.2&fileType=.tsx&code=C4TwDgpgBAcghgNwEZwE4ElgQLYGcoC8UAShAMYD2qAJgDwBQUUuwqAlgHYDmANI1AHoBUStmwQOwfi3bd6APgDc9eqEixEKDFjyES5KnX4BRAB5kANgFdqEWgCI49qAB8o9pM7f2y9nu6d5PiYZTi5%2BJXogA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6MgeyeUuX0Ra1mAE0QAPRMNocARgCtEZOn0JJ0YANrhsORNGgdokADTadkEeNo8A5kwCG%2BWNESQzYALraAviG9A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

```JSON
{
  "rules": {
    "@typescript-eslint/consistent-indexed-object-style": ["error", "index-signature"]
  }
}
```

```TS
type NavbarItems = Record<
  string,
  // lazy-load the component
  () => (props: any) => JSX.Element
>;

type NavbarItems = Record<
  Exclude<"a" | "b" | "c", "a">,
  string
>;
```

**Expected Result**

I expect the first to be fixed with comment preserved, and the second to be fixed to valid code:

```TS
// lazy-load the component
type NavbarItems = { [key: string]: () => (props: any) => JSX.Element };

type NavbarItems = { [key in Exclude<"a" | "b" | "c", "a">]: string };
```

**Actual Result**

```ts
// This has the comment swallowed
type NavbarItems = { [key: string]: () => (props: any) => JSX.Element };

// This should use mapped type instead of index signature
type NavbarItems = { [key: Exclude<"a" | "b" | "c", "a">]: string };
```

**Additional Info**

**Versions**

| `@typescript-eslint/eslint-plugin` | `5.16.0` |
| `@typescript-eslint/parser`        | `5.16.0` |
| `TypeScript`                       | `4.6.3` |
| `ESLint`                           | `8.11.0` |
| `node`                             | `17.5.0` |

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4730

## Overview

This PR addresses #4730 and adjusts the rule so it uses a suggestion instead of an auto-fix if the change may create an invalid index signature.

Some thoughts:

_Even if the rule uses a suggestion, it can still flag records that can't be fixed to a valid index signature. Developers may need to add `eslint-disable` comments for these these. As long as the rule [isn't type aware](https://github.com/typescript-eslint/typescript-eslint/issues/4730#issuecomment-1078753745), I think that's the best alternative._

### DIFF
--- a/packages/eslint-plugin/src/rules/consistent-indexed-object-style.ts
+++ b/packages/eslint-plugin/src/rules/consistent-indexed-object-style.ts
@@ -3,7 +3,12 @@ import type { ReportFixFunction } from '@typescript-eslint/utils/ts-eslint';
 
 import { AST_NODE_TYPES, ASTUtils } from '@typescript-eslint/utils';
 
-import { createRule, isParenthesized, nullThrows } from '../util';
+import {
+  createRule,
+  getFixOrSuggest,
+  isParenthesized,
+  nullThrows,
+} from '../util';
 
 type MessageIds =
   | 'preferIndexSignature'
@@ -20,6 +25,7 @@ export default createRule<Options, MessageIds>({
       recommended: 'stylistic',
     },
     fixable: 'code',
+    // eslint-disable-next-line eslint-plugin/require-meta-has-suggestions -- suggestions are exposed through a helper.
     hasSuggestions: true,
     messages: {
       preferIndexSignature: 'An index signature is preferred over a record.',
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (approved):
Awesome!

@Josh-Cena I was just thinking about #4730, it keeps coming up. This is great.

packages/eslint-plugin/src/rules/consistent-indexed-object-style.ts
```diff
@@ -113,14 +119,37 @@ export default createRule<Options, MessageIds>({
...
+
           context.report({
             node,
             messageId: 'preferIndexSignature',
-            fix(fixer) {
-              const key = context.sourceCode.getText(params[0]);
-              const type = context.sourceCode.getText(params[1]);
-              return fixer.replaceText(node, `{ [key: ${key}]: ${type} }`);
-            },
+            fix: shouldAutoFix
```
kirkwaiblinger, line 138: (not a blocker) wondering whether getOrSuggest can be used here for some visual simplification?

https://github.com/typescript-eslint/typescript-eslint/blob/5b2ebcdea9011a6de7e657118874ae85c08d5a0e/packages/eslint-plugin/src/util/getFixOrSuggest.ts#L1-L21

kirkwaiblinger (commented):
Haven't looked too closely, just dropped one drive-by question. Feel free to merge, no blockers from me

packages/eslint-plugin/src/rules/consistent-indexed-object-style.ts
ronami: Nice, it can, though it would require an `eslint-disable` line similar to this one, as it's the only suggestion the rule has: https://github.com/typescript-eslint/typescript-eslint/blob/d24a82854d06089cbd2a8801f2982fd4781f3701/packages/eslint-plugin/src/rules/return-await.ts#L47-L48

I've updated the PR to use `getOrSuggest`, and I think that even though it requires the disable comment, it helps simplify the logic.

kirkwaiblinger (approved):

PR #10494, opened 2024-12-13
fix(eslint-plugin): [prefer-reduce-type-parameter] don't report cases in which the fix results in a type error

### LINKED ISSUE #3440: Bug: [prefer-reduce-type-parameter] unfixable reporting

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3440

## Overview

This PR tackles #3440 and makes the rule skip cases where the type assertion is necessary and auto-fixing to using `reduce()`'s type parameter will result in a type error.

Alternatively, as code that does this isn't exactly safe, the rule can still report such cases but use a suggestion instead of an auto-fix (although `no-unsafe-type-assertion` will report the assertion as unsafe).

### DIFF
--- a/packages/eslint-plugin/tests/rules/prefer-reduce-type-parameter.test.ts
+++ b/packages/eslint-plugin/tests/rules/prefer-reduce-type-parameter.test.ts
@@ -101,6 +101,9 @@ ruleTester.run('prefer-reduce-type-parameter', rule, {
         );
       }
     `,
+    `
+      ['a', 'b'].reduce((accum, name) => \`\${accum} | hello \${name}!\`);
+    `,
   ],
   invalid: [
     {
--- a/packages/eslint-plugin/tests/rules/prefer-reduce-type-parameter.test.ts
+++ b/packages/eslint-plugin/tests/rules/prefer-reduce-type-parameter.test.ts
@@ -57,23 +57,67 @@ ruleTester.run('prefer-reduce-type-parameter', rule, {
         return a.concat(1);
       }, [] as number[]);
     `,
+    `
+      ['a', 'b'].reduce(
+        (accum, name) => ({
+          ...accum,
+          [name]: true,
+        }),
+        {} as Record<'a' | 'b', boolean>,
+      );
+    `,
+    // Object literal may only specify known properties, and 'c' does not exist in
+    // type 'Record<"a" | "b", boolean>'.
+    `
+      ['a', 'b'].reduce(
+        (accum, name) => ({
+          ...accum,
+          [name]: true,
+        }),
+        { a: true, b: false, c: true } as Record<'a' | 'b', boolean>,
+      );
+    `,
+    // '{}' is assignable to the constraint of type 'T', but 'T' could be
+    // instantiated with a different subtype of constraint 'Record<string, boolean>'.
+    `
+      function f<T extends Record<string, boolean>>() {
+        ['a', 'b'].reduce(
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/tests/rules/prefer-reduce-type-parameter.test.ts
```diff
@@ -57,23 +57,67 @@ ruleTester.run('prefer-reduce-type-parameter', rule, {
...
+    `
+      function f<T extends Record<string, boolean>>() {
+        ['a', 'b'].reduce(
+          (accum, name) => ({
+            ...accum,
+            [name]: true,
+          }),
+          {} as T,
+        );
+      }
```
ronami, line 91: This was a missing edge case in https://github.com/typescript-eslint/typescript-eslint/issues/10453.

packages/eslint-plugin/tests/rules/prefer-reduce-type-parameter.test.ts
```diff
@@ -57,23 +57,70 @@ ruleTester.run('prefer-reduce-type-parameter', rule, {
...
+    `,
+    `
+      ['a', 'b'].reduce((accum, name) => \`\${accum} | hello \${name}!\`);
```
ronami, line 105: This isn't directly related to the change, but it was previously missing coverage, and because it is on its own line, it fails codecov.

packages/eslint-plugin/tests/rules/prefer-reduce-type-parameter.test.ts
```diff
@@ -57,23 +57,67 @@ ruleTester.run('prefer-reduce-type-parameter', rule, {
...
+    // Object literal may only specify known properties, and 'c' does not exist in
+    // type 'Record<"a" | "b", boolean>'.
+    `
+      ['a', 'b'].reduce(
+        (accum, name) => ({
+          ...accum,
+          [name]: true,
+        }),
+        { a: true, b: false, c: true } as Record<'a' | 'b', boolean>,
+      );
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Heh, yeah, I say let's cover it.

JoshuaKGoldberg (approved):
YES FINALLY!!!!

Steve Buscemi in a brown messy wig saying "Whew!" and shaking/vibrating his head

PR #10474, opened 2024-12-07
fix(eslint-plugin): [no-unnecessary-boolean-literal-compare] flag values of a type parameter with boolean type constraints

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10443
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

This PR follows-up #10443 and reports on values with a boolean type parameter that are being compared to a boolean literal:

```ts
// fails
const test1: (someCondition: boolean) => void = someCondition => {
  if (someCondition === true) {}
};

// should fail
const test2: <T extends boolean>(someCondition: T) => void = someCondition => {
  if (someCondition === true) {}
};
```

Issue #9899, opened 2024-08-28
Enhancement: [no-deprecated] Add 'allow' option for specific types or values


BurningEnlightenment:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [X] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### My proposal is suitable for this project

- [X] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Link to the rule's documentation

https://typescript-eslint.io/rules/no-deprecated/

### Description

I find it very useful to flag and address deprecated API usage during the dependency upgrade cycle. Whence I'd like to enforce the rule for the usage of my dependencies.
However, for my own code base I want to follow a gradual deprecation cycle, i.e. I want to introduce an updated API and inform users writing new code about the updated API via a `@deprecated` JSDoc comment. Afterwards I'd phase the deprecated API out and remove it in the end.
The current rule is unsuited for the combined requirements, so I propose a boolean option `ignoreUserDeprecations`.

### Fail

```TypeScript
/** @deprecated Use apiV2 instead. */
async function apiV1(): Promise<void> { /*…*/ }

async function apiV2(): Promise<void> { /*…*/ }

await apiV1();

/* >===================================< */

import { parse } from 'node:url';

// 'parse' is deprecated. Use the WHATWG URL API instead.
const url = parse('/foo');
```

### Pass

```TypeScript
/** @deprecated Use apiV2 instead. */
async function apiV1(): Promise<void> { /*…*/ }

async function apiV2(): Promise<void> { /*…*/ }

// apiV1 has been defined in the current code base and is therefore not flagged
await apiV1();

/* >===================================< */

// Modern Node.js API, uses `new URL()`
const url2 = new URL('/foo', 'http://www.example.com');
```

### Additional Info

Originally proposed in the [no-deprecated] issue https://github.com/typescript-eslint/typescript-eslint/issues/8988#issuecomment-2313459356

ViniciusSCG:
to fix your trouble check this fix, i see it in another issue,
[...]

PR #10496, opened 2024-12-13
fix(eslint-plugin): add getConstraintInfo to handle generic constraints better

### LINKED ISSUE #10438: Repo: Investigate better handling of unconstrained generics

### LINKED ISSUE #10443: Bug: [no-unnecessary-boolean-literal-compare] doesn't flag values of a type parameter with boolean type constraint

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=PTAEDMEMEsBsGcBQBjA9gO3gF1FgptgIwBcoAFPKgLZ4DCGAJtFtBqQEaqqx6ToCUoALwA%2BUADdU0BsNCUa9dExYZhYgN6JQoaOHLy6jZq3TChQ3ACcArnkHqAvogcBuRIhByAFqmuwZUHAoGNi4BFgATKQAPAAqoHgAHvhK8KCc3LzoIhTUhkrGbKCxgqISUjIWBorKJmqgmtq6%2Bnk1habmFlg2dg1OrohAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ARhw5IhTWo3yJoQ%2BLTIcAtsRGJ0UJdA7RI4MAF8QZoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
// fails
const test1: (someCondition: boolean) => void = someCondition => {
  if (someCondition === true) {}
};

// should fail
const test2: <T extends boolean>(someCondition: T) => void = someCondition => {
  if (someCondition === true) {}
};
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-unnecessary-boolean-literal-compare": "error"
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expect all use-cases to be consistent and report as errors.

### Actual Result

The use-cases with a type constraint doesn't report as an error (even though it should).

### Additional Info

This seems similar to #10311, and it looks like the fix should be similar.

Note that this issue is also similar to #10442, and I wasn't sure if I should create one issue for each rule or aggregate them under a single issue. I apologize if this was a bit spammy. I checked existing rules for similar problems after resolving #10311.

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue
   - fixes #10438
   - <s>fixes (not anymore) #10443</s> EDIT - this was already solved in the meantime, just refactoring

## Overview

Since removing or modifying `getConstrainedTypeAtLocation` would be a breaking change (see https://typescript-eslint.io/users/versioning#parser-typescript-estree-scope-manager-types-type-utils-utils), <s>I've deprecated it and added a new utility in its place</s> I've left it in place for now, and just added a new utility. Since we may change our implementation in the future, I've made the new utility internal, and not deprecated the existing `getConstrainedTypeAtLocation`. Deprecating it and replacing internal usages can be handled in followup work. See https://github.com/typescript-eslint/typescript-eslint/pull/10496#discussion_r1898269806

I've redone the testing so that the tests for `getConstrainedTypeAtLocation` actually test something useful, and tested several edge cases for both the old and new APIs.

I've used the new API for a few rules for good measure. We'll want to gradually replace the rest of our usages in followup work.

Retouched the fix in #10314 <s>and used the new strategy to solve #10443</s> and in #10474

***
<s>TODO - I'll have to find a way to manage the deprecation linting failures 🙁  https://github.com/typescript-eslint/typescript-eslint/issues/9899 would be extremely clutch. Not really sure how to handle this otherwise, other than to alias the function with a non-deprecated version to replace the existing usages, with the intention of cleaning up the alias later. </s> Not deprecating anything yet.

<s>UPDATE - I chose to alias `getConstrainedTypeAtLocation` as the ironically not-deprecated `DEPRECATED_getConstrainedTypeAtLocation` in order to allow existing usages not to fail CI.</s>


packages/type-utils/tests/getConstrainedTypeAtLocation.test.ts
```diff
@@ -1,57 +1,309 @@
...
+      mockServices(typeChecker, parameterNode),
+      mockEstreeNode(),
+    );
+
+    expect(tsutils.isTypeParameter(constraintAtLocation)).toBe(false);
+    expect(tsutils.isIntrinsicStringType(constraintAtLocation)).toBe(true);
+  });
+});
+
+describe('getConstraintTypeInfoAtLocation', () => {
```
kirkwaiblinger, line 183: Add tests for generic whose constraint is generic

packages/type-utils/src/getConstrainedTypeAtLocation.ts
```diff
@@ -4,8 +4,18 @@ import type {
 } from '@typescript-eslint/typescript-estree';
 import type * as ts from 'typescript';
 
+import * as tsApiUtils from 'ts-api-utils';
```
JoshuaKGoldberg, line 7: [Style] Since this is its name everywhere else:

```suggestion
import * as tsutils from 'ts-api-utils';
```

packages/type-utils/package.json
```diff
@@ -58,6 +58,7 @@
   "devDependencies": {
     "@jest/types": "29.6.3",
     "@typescript-eslint/parser": "8.18.0",
+    "@typescript/vfs": "*",
```
JoshuaKGoldberg, line 61: y?

```suggestion
```

packages/type-utils/src/getConstrainedTypeAtLocation.ts
```diff
@@ -4,8 +4,18 @@ import type {
...
 /**
  * Resolves the given node's type. Will resolve to the type's generic constraint, if it has one.
+ *
+ * If the type is generic and does _not_ have a constraint, the type will be
+ * returned as-is, rather than returning an `unknown` type. This can be checked
+ * for by checking for the type flag ts.TypeFlags.TypeParameter.
+ *
+ * @see https://github.com/typescript-eslint/typescript-eslint/issues/10438
+ *
+ * @deprecated Use getConstraintTypeInfoAtLocation instead.
```
JoshuaKGoldberg, line 18: [Docs] Fancy linkage:

```suggestion
 * @deprecated Use {@link getConstraintTypeInfoAtLocation} instead.
```

packages/type-utils/src/getConstrainedTypeAtLocation.ts
```diff
@@ -18,3 +28,59 @@ export function getConstrainedTypeAtLocation(
 
   return constrained ?? nodeType;
 }
+
+export interface ConstraintTypeInfoBase {
+  constraintType: ts.Type | undefined;
+  isTypeParameter: boolean;
+  type: ts.Type;
+}
+
```
JoshuaKGoldberg, line 37: [Style] This doesn't enforce much and isn't used anywhere outside the `extends`:

```suggestion
```

JoshuaKGoldberg (approved):
LGTM, nice to see the cleanup & related fixes!

Just requesting a collection of touchups, nothing major.

packages/type-utils/src/getConstrainedTypeAtLocation.ts
kirkwaiblinger: fair enough. i thought this was sort of a required part of the idiom but yeah it can be left off 👍

packages/type-utils/package.json
kirkwaiblinger: 🤷‍♂️ 

https://github.com/typescript-eslint/typescript-eslint/blob/bc12e04be86f7c7dd42d4f2a02f1cf51f35bf067/packages/type-utils/tests/getConstrainedTypeAtLocation.test.ts#L46-L76

kirkwaiblinger: ❤️

kirkwaiblinger: Actually, I may be able to copy testing patterns from within this repo, and avoid that, lemme poke around.

JoshuaKGoldberg (approved):
I like this strategy 😄

packages/type-utils/package.json
kirkwaiblinger: Yeah, removed all that vfs stuff. Turns out, we already infra for parsing and using type information. _Who would have guessed?_ 🤦‍♂️

packages/type-utils/src/getConstrainedTypeAtLocation.ts
```diff
@@ -18,3 +28,52 @@ export function getConstrainedTypeAtLocation(
...
+  | ConstraintTypeInfoConstrained
+  | ConstraintTypeInfoNonGeneric
+  | ConstraintTypeInfoUnconstrained;
+
+/**
+ * Resolves the given node's type, and returns info about whether it is a generic or ordinary type.
+ *
+ * Successor to {@link getConstrainedTypeAtLocation} due to https://github.com/typescript-eslint/typescript-eslint/issues/10438
+ */
+export function getConstraintTypeInfoAtLocation(
```
kirkwaiblinger, line 60: @JoshuaKGoldberg I've just had the thought - I think this might be the wrong API to write. It might be better to just accept a type as input, like
```ts
export declare function getConstraintInfo(type: ts.Type): { 
  isTypeParameter: boolean, 
  constraint: ts.Type | undefined
};
```
This maintains a separation between _getting_ types and _analyzing_ types.

JoshuaKGoldberg: @kirkwaiblinger just confirming, are you planning on making changes? And/or, is this a question?

kirkwaiblinger: Lol. In my head, I was asking you for your opinion, yeah, but I forgot to write that out 😆 

Coming back to this, I currently lean towards: let's go with this PR in its current state, and if there's a need to extract just the constraint info part of this function that's quite easy to do in a followup when the need arises.

Feel free to offer your opinion! 😁

kirkwaiblinger: If you're on board with this plan, then this PR is up to date and ready to be merged 👍

JoshuaKGoldberg: Oh awesome! I don't have a strong preference here.

Proposal: whatever strategy you want to go with, how about marking the final creation as `@internal` so it's not part of the public API? That way we can play with it as we want.

kirkwaiblinger: Made lots of changes in this arena. New utility is just internal to eslint-plugin for now, no deprecation yet. Should be much easier to review!

JoshuaKGoldberg (approved):
🎉 This is great!

packages/type-utils/src/getConstrainedTypeAtLocation.ts
JoshuaKGoldberg: I like this strategy. The union of possible constraint info points seems reasonable to me.

PR #10504, opened 2024-12-15
fix(typescript-estree): align TS module nodes to namespaces

### LINKED ISSUE #10486: Bug: Align `module X.Y {}` AST with `namespace X.Y {}` AST

### Relevant Package

ast-spec

### Description

https://github.com/typescript-eslint/typescript-eslint/pull/6272 changed the AST of `namespace X.Y {}` from `TSModuleDeclaration { name: Identifier("X"), body: TSModuleDeclaration { name: Identifier("Y"), body: [] }` to `TSModuleDeclaration { name: TSQualifiedName("X", "Y"), body: [] }`.

The same change was _not_ done for `module X.Y {}` even if it's semantically equivalent to `namespace X.Y {}`. There is significant value on keeping their AST aligned, since most rules you'd write for one are likely also relevant for the other.

https://github.com/typescript-eslint/typescript-eslint/issues/6440 already proposed in the past to change the AST of `module X.Y {}`, but it was closed as a duplicate of https://github.com/typescript-eslint/typescript-eslint/issues/4966 (tracking `namespace X.Y {}`) and then it never happened.

I tried looking at all the linked discussions, but I couldn't find any reason for the difference between `module X.Y {}` and `namespace X.Y {}`.

### Additional Info

I noticed this in https://github.com/typescript-eslint/typescript-eslint/pull/10452: Babel is changing its AST to align with what typescript-eslint does for `namespace X.Y {}`, and I noticed that the change just _moved_ the differences rather than removing them.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10486

## Overview

Switches `convert.ts`'s checks to be on whether the name is a string literal, not whether `ts.NodeFlags.Namespace` is set.

In theory, this is a breaking change for the AST - even though it's a direct bug fix. Not sure if we can feasibly release it as in a minor version. 😬 

💖


Josh-Cena:
I doubt parser users are actually handling legacy `module` declarations, and parser users' users writing `module` code would be exceptionally rare too.

JoshuaKGoldberg:
We had a chat internally and 👍d it. Nobody's raised a strong indication of why we'd want to keep the old buggy behavior. Merging now for today's release. 🤞

Issue #4905, opened 2022-05-04
[prefer-nullish-coalescing] request for ternary support


jguddas:
The following cases can all be converted to `foo ?? 'a string'`.

```ts
const foo: any = 'bar';
foo !== undefined && foo !== null ? foo : 'a string';
foo === undefined || foo === null ? 'a string' : foo;

const foo:? string = 'bar';
foo !== undefined ? foo : 'a string';
foo === undefined ? 'a string' : foo;

const foo: string | null = 'bar';
foo !== null ? foo : 'a string';
foo === null ? 'a string' : foo;
```

This is even hinted at in the docs, but sadly currently not covered.

https://github.com/typescript-eslint/typescript-eslint/blob/da485279f13cb95db1ee131a4f9c5367d54020fe/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md?plain=1#L9-L17

Supporting `foo ? foo : 'a string'` would create conflicts with `no-unneeded-ternary` which is why I'm against doing so.

I can have a look at creating a PR if there is interest for this.

PR #10517, opened 2024-12-18
fix(eslint-plugin): [prefer-nullish-coalescing] doesn't report on ternary but on equivalent ||

### LINKED ISSUE #10470: Bug: [prefer-nullish-coalescing] Doesn't report on ternary, but does on equivalent ||

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=JYOwLgpgTgZghgYwgAgEIHsAeyDeAoZZANzgBsBXCALmQGcwpQBzAbjwF888ATCBUuFBQxyIBGGDoQyJhDAAxMqQBGiANYZMACgCUNTWx58BQ5Ain1kveOVJhN%2BrMgA%2ByUddARuh63Fv2nZ1dZBSVVBA0sXTYAehjCBMSAPyS8OOQABSEYaDdaZmQQW1JgWgALM3QyCFoEAvQAB2g4MHQoZAA6Lq5ff01kAH4rCBs7fpoQxVIVdU1otPiAOXRkaCg25CotIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6KL2jjokcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
interface Box {
  value: string;
}

declare function getFallbackBox(): Box;

declare const defaultBox: Box | undefined;

defaultBox || getFallbackBox();
//         ~~
// Prefer using nullish coalescing operator ...

defaultBox ? defaultBox : getFallbackBox();
// No error :(
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

Both the `||` or and the `?:` ternary should report.

### Actual Result

Only the `||` or reports.

### Additional Info

Ternary support was added in https://github.com/typescript-eslint/typescript-eslint/issues/4905, but seems limited to more strict cases of both `undefined` and `null` being checked?

Split out of https://github.com/typescript-eslint/typescript-eslint/pull/10218#discussion_r1829827873.

💖

OlivierZal:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10470


OlivierZal:
Hi @JoshuaKGoldberg, @kirkwaiblinger, the lint failure probably comes from an autofix after I saved the files from my IDE, but I can easily revert.

I think I still need to make sure that the following is excluded from raising an error:
```ts
declare const x: string | undefined // and the same with null
declare const y: string

x ? x : y
```

to avoid any conflict with `no-unneeded-ternary` (see https://github.com/typescript-eslint/typescript-eslint/issues/4905).

Don't hesitate to start sharing any comment you would have on my PR.

OlivierZal:
Hi @JoshuaKGoldberg,

I think the 2 lint errors are false negatives because they succeeded before I rebased and made a squash.

Anyway, I would just like to check with the typescript-eslint team the scope of this fix, so I can adapt the code and tests accordingly.

Do we agree that we should suggest a fix for the following:
```ts
const foo: NonNullable<object> | null; // same with `| undefined` and `| null | undefined`null
const bar: NonNullable<object>
foo ? foo : bar;
!foo ? bar : foo;
```

but not for the following (also for `number`, `boolean`, `string`, `unknown`, `any`):
```ts
const foo: string | null; // same with `| undefined` and `| null | undefined`
const bar: string;
foo ? foo : bar;
!foo ? bar : foo;
```

OlivierZal:
Hi @JoshuaKGoldberg, @kirkwaiblinger,

I think the 2 lint errors are false negatives because they weren't raised before I rebased and squashed.

Anyway, to start with, I wrote tests directly based on the example provided in the issue but there's still work to be done, and I'd like to check with the typescript-eslint team about the scope of this fix so I can adapt the code and tests accordingly.

Do we agree that we should suggest a fix for the following:
```ts
type A = NonNullable<...>; // I write this only for the understanding purpose

const foo: A | null; // same with `| undefined` and `| null | undefined`
const bar: A;
foo ? foo : bar;
!foo ? bar : foo;
```

but not for `string`, `number`, `boolean`, `unknown`, `any`?

OlivierZal:
Hi,

I know the branch is outdated, but maybe I’ll wait for a first feedback before to merge upstream in it.

OlivierZal:
@JoshuaKGoldberg, @kirkwaiblinger, https://github.com/typescript-eslint/typescript-eslint/runs/34947251166 check seems to be stuck. 

However, tests passed so I guess the PR is ready for a first review.

OlivierZal:
I've just noticed that `ignorePrimitive` has not been implemented for ternary expressions. It probably should be part of this PR, since the fix introduces the need, shouldn't it?

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -343,33 +343,48 @@ export default createRule<Options, MessageIds>({
...
+              isNodeEqual(testNode, node.consequent)
+            ) {
+              identifier = testNode;
+            } else if (
+              (operator === '===' || operator === '==') &&
+              isNodeEqual(testNode, node.alternate)
+            ) {
+              identifier = testNode;
+            } else {
+              return;
```
JoshuaKGoldberg, line 385: [Style] Nit: since `if (!identifier) { return; }` later will be run anyway:

```suggestion
```

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
JoshuaKGoldberg, line 1: [Tests] Some missing cases to be added:

* Fix/suggestion logic, since the `isFixable` function was modified
* [The original issue's `||` repro](https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=JYOwLgpgTgZghgYwgAgEIHsAeyDeAoZZANzgBsBXCALmQGcwpQBzAbjwF888ATCBUuFBQxyIBGGDoQyJhDAAxMqQBGiANYZMACgCUNTWx58BQ5Ain1kveOVJhN%2BrMgA%2ByUddARuh63Fv2nZ1dZBSVVBA0sXTYAehjCBMSAPyS8OOQABSEYaDdaZmQQW1JgWgALM3QyCFoEAvQAB2g4MHQoZAA6Lq5ff01kAH4rCBs7fpoQxVIVdU1otPiAOXRkaCg25CotIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6KL2jjokcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

JoshuaKGoldberg (changes requested):
A good start - nice job working with such tricky existing logic! 

The tests for `valid` cases and primitives are very nice and thorough. Requesting changes on more thorough coverage of `invalid` please. Thanks!

OlivierZal:
Hi @JoshuaKGoldberg, thanks for the review, I've made 2 of the requested changes (nit + original issue tests).

Regarding the fix/suggestion logic tests following the `isFixable` modification, the only change is the introduction of an "`implicitEquality`" case (I made it a constant in https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/20db420f72efe51332e29b0c471b151bcf3f7eda for more clarity):
- the [first occurrence](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R396) is actually the negation of this case in order to keep the pre-existing logic, so it is already covered by the existing tests;
- the [next condition](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R401 excludes the "`implicitEquality`" case (so it is also covered by the existing tests);
- the [next one](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R409) leads to always valid expressions, so is covered by the tests I've added [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR225-R239) (so no fix/suggestion);
- then [the "`implicitEquality`" case occurs](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R413): fix/suggestion logic is explicitly covered [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR948)
- the last conditions also exclude the "`implicitEquality`" case since a `return` already happened (so they are also covered by the existing tests);
- [this addition](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R443) is also tested by the "`!x ? y : x` tests which I've added, which all falls [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR948).

Do you see any other cases I’ve missed that I could add?

OlivierZal:
Hi @JoshuaKGoldberg, thanks for the review, I've made at least 2 of the requested changes:
- handling [nit](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/1fbf63bfb633fe9039bd65e2e5327b961db2f95a);
- adding [original issue tests](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/74a40592bb2ed9afa3e0a57ab508a4b8f6ea638c);
- adding more tests [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/0dc6e4e361460829a45fc0ef742ef4fd6b5431d7) and [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/c19897afa58c84249b764a45ff9753cf33d19a7c).

Regarding the fix/suggestion logic tests following the `isFixable` modification, the only change is the introduction of an "`implicitEquality`" case (I made it a constant in https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/20db420f72efe51332e29b0c471b151bcf3f7eda for more clarity):
- the [first occurrence](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R396) is actually the negation of this case in order to keep the pre-existing logic, so it is already covered by the existing tests;
- the [next condition](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R401) (handling `==` / `!=` operators) excludes this "`implicitEquality`" case (so it is also covered by the existing tests);
- the [next one](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R409) (handling `any` / `unknown` types) leads to always valid expressions (no fix/suggestion), covered by the tests I've added [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR289-R575);
- then [the "`implicitEquality`" case occurs](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R413): the fix/suggestion logic for invalid expressions is covered [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR1372) (`x ?? y` – you need to expand to see it), tests for valid expressions (no fix/suggestion) have been added [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR128-R288);
- the following excludes the "`implicitEquality`" case since a `return` already happened [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R415) in such a case (so they are also covered by the existing tests);
- the fix/suggestion logic related to [this addition](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R443) is tested by the "invalid" tests I've added – `!x ? y : x` for the "consequent" case and `x ? x : y` for the "alternate" case (handles the new `!operator` condition) –, which all fall [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR1372) (`x ?? y` – you need to expand to see it).

Do you see any other cases I’ve missed that I could add? Or do you have something else in mind?

OlivierZal:
Hi @JoshuaKGoldberg, thanks for the review, I've made at least 2 of the requested changes:
- handling [nit](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/1fbf63bfb633fe9039bd65e2e5327b961db2f95a);
- adding [original issue tests](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/74a40592bb2ed9afa3e0a57ab508a4b8f6ea638c);
- adding more tests [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/0dc6e4e361460829a45fc0ef742ef4fd6b5431d7) and [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/c19897afa58c84249b764a45ff9753cf33d19a7c).

Regarding the fix/suggestion logic tests following the `isFixable` modification, the only change is the introduction of an "`implicitEquality`" case (I made it a constant in https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/20db420f72efe51332e29b0c471b151bcf3f7eda for more clarity):
- the [first occurrence](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R396) is actually the negation of this case in order to keep the pre-existing logic, so it is already covered by the existing tests;
- the [next condition](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R401) (handling `==` / `!=` operators) excludes this "`implicitEquality`" case (so it is also covered by the existing tests);
- the [next one](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R409) (handling `any` / `unknown` types) leads to always valid expressions (no fix/suggestion), covered by the tests I've added [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR289-R575);
- then [the "`implicitEquality`" case occurs](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R413): the fix/suggestion logic for invalid expressions is covered [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR1372) (`x ?? y` – you need to expand to see it), tests for valid expressions (no fix/suggestion) have been added [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR128-R288);
- the following excludes the "`implicitEquality`" case since a `return` already happened [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R415) in such a case (so they are also covered by the existing tests);
- the fix/suggestion logic related to [this addition](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R443) is tested by the "invalid" tests I've added – `!x ? y : x` for the "consequent" case and `x ? x : y` for the "alternate" case (handles the new `!operator` condition) –, which all fall [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR1372) (`x ?? y` – you need to expand to see it).

Do you see any other cases I’ve missed that I could add? Or do you have something else in mind?

OlivierZal:
Hi @JoshuaKGoldberg, thanks for the review, I've made at least 2 of the requested changes:
- handling [nit](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/1fbf63bfb633fe9039bd65e2e5327b961db2f95a);
- adding [original issue tests](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/74a40592bb2ed9afa3e0a57ab508a4b8f6ea638c);
- adding more tests [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/0dc6e4e361460829a45fc0ef742ef4fd6b5431d7) and [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/c19897afa58c84249b764a45ff9753cf33d19a7c).

Regarding the fix/suggestion logic tests following the `isFixable` modification, the only change is the introduction of an "`implicitEquality`" case (I made it a constant in https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/20db420f72efe51332e29b0c471b151bcf3f7eda for more clarity):
- the [first occurrence](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R396) is actually the negation of this case in order to keep the pre-existing logic, so it is already covered by the existing tests;
- the [next condition](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R401) (handling `==` / `!=` operators) excludes this "`implicitEquality`" case (so it is also covered by the existing tests);
- the [next one](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R409) (handling `any` / `unknown` types) leads to always valid expressions (no fix/suggestion), covered by the tests I've added [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR289-R575);
- then [the "`implicitEquality`" case occurs](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R413): the fix/suggestion logic for invalid expressions is covered [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR1372) (`x ?? y` – you need to expand to see it), tests for valid expressions (no fix/suggestion) have been added [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR128-R288);
- the following excludes the "`implicitEquality`" case since a `return` already happened [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R415) in such a case (so they are also covered by the existing tests);
- the fix/suggestion logic related to [this addition](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R443) is tested by the "invalid" tests I've added – `!x ? y : x` for the "consequent" case and `x ? x : y` for the "alternate" case (handles the new `!operator` condition) –, which all fall [here](https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-c8c88d5eba96c9a3dd571eea038b005d4f0f5433b805aa34d3f31e08376cb1caR1372) (`x ?? y` – you need to expand to see it).

Do you see any other cases I’ve missed that I could add? Or do you have something else in mind?

JoshuaKGoldberg (approved):
I think this is plenty thorough enough for me 😄 but I've also kind of burned out on reviewing `prefer-nullish-coalescing`. We should get review from another @typescript-eslint/triage-team member.

OlivierZal:
@JoshuaKGoldberg, @kirkwaiblinger, waiting for this PR having a second review, I keep updating the branch each time a new PR is merged. Is it the right thing to do, or can I leave it as is until the review?

JoshuaKGoldberg:
Oh! You don't need to keep doing that. Thank you for the effort though!

https://typescript-eslint.io/contributing/pull-requests/#updating-over-time

OlivierZal:
Oh sorry I missed that! And won’t mention anyone explicitly anymore 😜

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -343,42 +343,57 @@ export default createRule<Options, MessageIds>({
...
-          if (isNullLiteral(testNode)) {
-            hasNullCheck = true;
-          } else if (isUndefinedIdentifier(testNode)) {
-            hasUndefinedCheck = true;
-          } else if (
-            (operator === '!==' || operator === '!=') &&
-            isNodeEqual(testNode, node.consequent)
+        if (!operator) {
+          if (
+            node.test.type === AST_NODE_TYPES.Identifier &&
```
kirkwaiblinger, line 352: Why only identifiers? I'd think member expressions at least are fair game.

([playground](https://deploy-preview-10517--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3lAXPA3vAEZ6YwCWKA5vAD7wCuKoAZpSMPAL4DcAUKEiwEydFhZIkeKCgCe-PlAB0ReAEIAvPBQMIEeAH5sK%2BHglJ%2By1bXrnLJo1dPw7QA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6KAHdJ0JpHBgAviG1A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false))
```ts
a.b != null ? a.b : foo; // reports
a.b || foo; // reports
a.b ? a.b : foo; // doesn't report - ??
```

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -343,42 +343,57 @@ export default createRule<Options, MessageIds>({
...
+            }
+          }
         }
 
         if (!identifier) {
           return;
         }
 
         const isFixable = ((): boolean => {
+          const implicitEquality = !operator || operator === '!';
```
kirkwaiblinger, line 393: I'm not sure I understand what `implicitEquality` means in this context. If I understand the code correctly, though, would `hasTruthinessCheck` be a better term?

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -395,6 +410,11 @@ export default createRule<Options, MessageIds>({
             return false;
           }
 
+          if (implicitEquality) {
+            const nullishFlags = ts.TypeFlags.Null | ts.TypeFlags.Undefined;
+            return (flags & ~nullishFlags & ts.TypeFlags.PossiblyFalsy) === 0;
```
kirkwaiblinger, line 415: I think it would be reuse our truthiness/falsiness/nullishness utils from no-unnecessary-condition here. I pass all test cases (and understand the code much better) with
```ts
          if (hasTruthinessCheck) {
            const nonNullishType = checker.getNonNullableType(type);
            return !isPossiblyFalsy(nonNullishType);
          }
```

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -343,42 +343,57 @@ export default createRule<Options, MessageIds>({
...
-          } else if (
-            (operator === '!==' || operator === '!=') &&
-            isNodeEqual(testNode, node.consequent)
+        if (!operator) {
+          if (
+            node.test.type === AST_NODE_TYPES.Identifier &&
+            isNodeEqual(node.test, node.consequent)
           ) {
-            identifier = testNode;
+            identifier = node.test;
```
kirkwaiblinger, line 355: Surely these branches imply both 
```ts
hasUndefinedCheck = true;
hasNullCheck = true;
```
?

kirkwaiblinger (changes requested):

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
```diff
@@ -125,6 +125,454 @@ x === null ? x : y;
...
+!x ? y : x;
+      `,
+      `
+declare let x: any | null | undefined;
+x ? x : y;
+      `,
+      `
+declare let x: any | null | undefined;
+!x ? y : x;
+      `,
```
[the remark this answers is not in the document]
OlivierZal: Sure I can remove them. Originally I didn't have them but since the code checks separately and in different ways null, undefined, unknown and any types, I wanted to make sure that a condition wouldn't override another. 

Will remove in a next commit.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: Right, I'll add it. 

By the way it reminds me that optional chains need better handling: https://github.com/typescript-eslint/typescript-eslint/issues/10531 (but it's in the binary case).

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: I meant that no `==` etc. is explicitly written but despite that a check is done. Indeed I can rename it: the simple fact you have issues to understand what it means demonstrates it's not a good name 😅

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: Thanks for the tip, I'm not familiar enough (at least, yet) with existing utils, but it's exactly what it's needed.

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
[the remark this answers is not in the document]
kirkwaiblinger: understood, but in case we're not 100% on the same page, what i mean is that the rule codepaths are exactly identical anyway because the type objects are exactly identical (`any` is just `any` - not a union of `any` and other stuff)... Try playing with the "types" tab of our playground to see what I mean? [playground](https://typescript-eslint.io/play/#ts=5.7.2&showAST=types&fileType=.tsx&code=CYUwxgNghgTiAEEQBd5QIwC40DsCeA3AFBGiSwJKpQBM2U%2B8APvDgK4QTPxs6gBmASxwhgxUuGhxEKVtnaduvAcNEEgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: Yes, that's the case but even if I set them the values wouldn't be read in such cases because it wouldn't enter in any of the condition where it's used: it wouldn't enter in https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R396, and then will return in https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R413

For more clarity, maybe should I default these to `undefined`, and set them only the cases where it's used? Or as you suggest: I do set them to `true` every time it is, even if the value is read later.

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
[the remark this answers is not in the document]
OlivierZal: Oh ok, I get what you mean now.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: But agreed that leave it as `false` although it is not is not a good thing.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: @kirkwaiblinger, what do you think about creating a `truthiness.ts` file under `/util`?

OlivierZal: what would mean creating dedicated tests by the way

kirkwaiblinger: Yep, makes sense to me!

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: Added: https://github.com/typescript-eslint/typescript-eslint/pull/10517/files#diff-dd40826685069133bdddab944ed6d6e71438ba0847484dd9a8ded0faba704283R354

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: Done in https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/1a8909d94a9df0e59dec559ce43a7b03cd47dc11 (I can add more tests for safety, but all the code is covered).

Do you advise to handle https://github.com/typescript-eslint/typescript-eslint/issues/10531 in this PR, or do I keep it for another PR?

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: Done in https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/99c0d026b6fd26238ce1bbd2de09b1f53ae30399

I'll add the new utils test in a next commit.

OlivierZal: I rolled back new utils update in rule `no-unnecessary-condition` (https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/aedc66b3a86a933c977a3557a83954dc3f3e2c44) since I guess it's cleaner not to change a rule when updating another rule. Do you agree? If so I'll create an issue for the utils update.

OlivierZal:
I didn't handle `x?.y` on purpose (at least for now, in the perimeter of this review), because the code doesn't handle optional chains in ternaries yet (https://github.com/typescript-eslint/typescript-eslint/issues/10531).

kirkwaiblinger (changes requested):
I've had a look at this, and there one big wrinkle that I'm just noticing (apologies if this significantly changes the direction of the PR) - we don't have have parity with the `a || b` case now because the checks there are different. 

https://github.com/typescript-eslint/typescript-eslint/blob/ec87f639e44767901122511d17da50d214fad161/packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts#L187-L279

For example, 
[Playground](https://deploy-preview-10517--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3lAXPJARgFbgYDcAUKJLAsulgXlCgJ6UVTwD828eBMvAD0w%2BHAAOSGBjSd4AHwXxBIscCQg0KAORZJ0jBQogAHlJnwA3vAC%2BQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6KAHdJ0JpHBgAviG1A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false)
```ts
declare const a: object;
declare const b: any;

a ? a : b; // reports
a || b; // doesn't report

export { }
```

***

There seem to be two fundamental discrepancies:
1. The `a || b` case currently ignores cases when the tested expression isn't possibly-nullish. I don't have strong feelings on whether that's the "right" behavior, but I think we should probably match that logic for the `a ? a : b` case for the purposes of this PR.
2. The `a || b` case has logic around the `ignorePrimitives` option. I think we need to respect that option for the `a ? a : b` case as well. 

Would you kindly look into those? Thanks! And great work on this so far!

OlivierZal:
@kirkwaiblinger, thanks for your review.

Indeed, I didn’t notice your first point because, when I started working on the PR, I expected everything to be factorized. However, it turns out that’s not the case. I also admit I didn’t check all the existing `||` cases, as the logic there is different.

Regarding your second point, I did notice it, but I believe it wasn’t raised until now because only explicit `null`/`undefined` checks were considered (unlike `||`, which doesn’t handle them). Therefore, `a`’s type wasn’t relevant before, but it is now with this PR.

What would you advise for the following?
- Option 1: Factorize the `||` and ternary logic. This is generally better for maintenance and might avoid the current issues with the rule, though the checks on the identifier differ in some cases.
- Option 2: Simply align the behavior while keeping both logics as they are.

kirkwaiblinger:
> Indeed, I didn’t notice your first point because, when I started working on the PR, I expected everything to be factorized. However, it turns out that’s not the case. I also admit I didn’t check all the existing `||` cases, as the logic there is different.

It kind of is, though, right? There's the `checkAssignmentOrLogicalExpression` check?

> Regarding your second point, I did notice it, but I believe it wasn’t raised until now because only explicit `null`/`undefined` checks were considered (unlike `||`, which doesn’t handle them). Therefore, `a`’s type wasn’t relevant before, but it is now with this PR.
> 
> What would you advise for the following?
> 
> * Option 1: Factorize the `||` and ternary logic. This is generally better for maintenance and might avoid the current issues with the rule, though the checks on the identifier differ in some cases; IMO given that both logics are very different, it wouldn't be a simple merge but a big reshape.
> * Option 2: Simply align the behavior while keeping both logics as they are.

My mental picture of this is as follows. There are two fundamentally different types of checks that this rule does:

- null checks via truthiness/falsiness checks that could be refactored to a `a ?? b`. This includes:

    - `a || b`
    - `a ||= b` (to become `a ??= b`)
    - (being added in this PR) `a ? a : b`/`!a ? b : a` (both of which become `a ?? b`.
    
    The common preconditions for these checks being reported are:
    
    - The tested expression is possibly `null` or `undefined`
    - The `ignorePrimitives` option is respected. (Note `ignorePrimitives: true` is very similar conceptually to the logic that you currently have in this PR to only fix if the nonnullable portion of the type is not possibly falsy)

    AFAICT, this logic is included in `checkAssignmentOrLogicalExpression` currently.
    
- _explicit_ null(ish) checks that could be refactored to `a ?? b`. This just includes ternaries that look like
    - `a != null ? a : b` (and all the strict/loose null/undefined possibly-negated combinations)
    
    The preconditions for this type of check being reported are 
     
    - The null(ish) check includes testing all possible nullish values that the tested expression could result in. In other words, 
       `objectOrNull !== null ? objectOrNull : somethingElse` should be reported but `objectOrNull !== undefined ? objectOrUndefined : somethingElse` should not.
       
     This logic is currently _inline_ in the `ConditionalExpression` selector, since that's the only place where it's used. 
     
So my thought is that you'd want to adapt `checkAssignmentOrLogicalExpression` as needed in order to reuse logic from it in the newly-added check. Then, all the "truthiness null checks" will go through that function (which may well be renamed) and all the "explicit null checks" will be handled in a single code path (which may stay inline since it's only used once, and its algorithm shouldn't change as part of this PR - though it may be convenient to factorize it for implementation reasons). 

No need to overthink keeping the diff as tiny as possible! Correctness trumps diff size! And also the test cases help give us confidence in changes 🙂 

Does this approach seem good to you? Entirely possible I've missed something.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -343,32 +347,48 @@ export default createRule<Options, MessageIds>({
           }
         }
 
-        if (!operator) {
-          return;
-        }
-
         let identifier: TSESTree.Node | undefined;
```
kirkwaiblinger, line 350: Not your code originally, but could you rename this, since it's not only `identifier`s? It's identifiers or member expressions (in other words, most-likely-side-effect-free-or-at-least-idempotent-identifer-like-expressions 😁)

OlivierZal:
Sounds pretty, I'll try to do my best!

OlivierZal:
@kirkwaiblinger, I think I addressed your comments.

I still need to add some tests (ignorePrimitives for ternaries) but I have a remark and a question:

- [Remark] I didn't make `checkAssignmentOrLogicalExpression` straight forward the "passage" of all the three use cases (assignment / logical / ternary) because it contains a logic shared only by assignment and logical – including a proper reporting with suggest/fix. On the other hand, I thought it wouldn't be very consistent to mix it (within a same function) with only _part_ of ternary's reporting – while many variables (`operator`, `node.alternate`...) are necessary and located with the _inline_ logic.
 
That's why I rather chose to extract the common logic in a separate function and renamed accordingly, but feel free to comment.

- [Question] Regarding options:
  - `checkAssignmentOrLogicalExpression` didn't contain only the reporting common to assignment and logical, but also some option checks:
    - `ignoreMixedLogicalExpressions`, which applies only to assignment and logical
    - `ignoreConditionalTests`, which could also apply to ternaries since the related check accepts any `ESTree.Node` as argument (and not only `ESTree.AssignmentExpression` or `ESTree.LogicalExpression`)
  - `ignoreBooleanCoercion` appears only in the assignment use case, but it could also to the two other use cases for the same than for `ignoreConditionalTests`

I chose to leave them as is to avoid any breaking changes outside the scope of the reported issue. However, if we consider it a bug and believe all the options should behave the same way for as many use cases as possible, I can make the necessary changes.

kirkwaiblinger (changes requested):
I think the current algorithm and factorization looks good! There's one minor complaint, where I think you've added a workaround which can be avoided - I made a little suggestion PR in order to illustrate this, see https://github.com/OlivierZal/typescript-eslint/pull/2. Fortunately the test cases you've written validated this well when I was playing around 👍 

I think if we make those few changes we'll be close to done iterating! 🙂

OlivierZal:
@kirkwaiblinger, I’m delighted I wasn’t too far off the mark 😅. Ultimately, I realized that the inconsistency mentioned in the corresponding issue went beyond the cited example, and since I’m still quite new to the project, I wasn’t entirely sure if taking a few liberties was appropriate.

Thank you so much for your time, help, and guidance.

The [last commit](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/80f3a2015669d8e09a98a056dd203d8579ce8655) only fixes a lint issue in [this commit](https://github.com/typescript-eslint/typescript-eslint/pull/10517/commits/5bd6b4e346a037302c3eb00c6ba577bd087ab8cd); all the other checks passed successfully.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -464,7 +536,9 @@ function isConditionalTest(node: TSESTree.Node): boolean {
 
   if (
     parent.type === AST_NODE_TYPES.ConditionalExpression &&
-    (parent.consequent === node || parent.alternate === node)
+    (parent.consequent === node ||
+      parent.alternate === node ||
+      node.type === AST_NODE_TYPES.UnaryExpression)
   ) {
```
kirkwaiblinger, line 542: Oh, I'm realizing that these `UnaryExpression` recursions are unnecessary as well, I think, right?
```suggestion
    (parent.consequent === node || parent.alternate === node)
  ) {
```

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -512,7 +586,9 @@ function isBooleanConstructorContext(
 
   if (
     parent.type === AST_NODE_TYPES.ConditionalExpression &&
-    (parent.consequent === node || parent.alternate === node)
+    (parent.consequent === node ||
+      parent.alternate === node ||
+      node.type === AST_NODE_TYPES.UnaryExpression)
```
kirkwaiblinger, line 591:
```suggestion
    (parent.consequent === node || parent.alternate === node)
```

packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.mdx
kirkwaiblinger, line 1: I think this maybe deserves an additional mention at the top of the page... what do you think of changing the bullet list to 
```
This rule reports when you may consider replacing:
- An `||` operator with `??`
- An `||=` operator with `??=`
- Ternary expressions (`?:`) that are equivalent to `||` or `??` with `??`
```
Or some such. 

Trying to find some way of technically including both the `a ? a : b` and the fundamentally-different `a != null ? a : b` (which is currently already lacking mention) without making a big deal out of it to the user

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -12,12 +12,15 @@ import {
...
   isNullLiteral,
-  isTypeFlagSet,
+  isPossiblyNullish,
   isUndefinedIdentifier,
   nullThrows,
   NullThrowsReasons,
 } from '../util';
 
+const isIdentifierOrMemberExpressionType = (type: TSESTree.AST_NODE_TYPES) =>
+  [AST_NODE_TYPES.Identifier, AST_NODE_TYPES.MemberExpression].includes(type);
```
kirkwaiblinger, line 22: We have a utility we use for this (that makes a fancy type predicate)
```suggestion
const isIdentifierOrMemberExpression = isNodeOfTypes([
  AST_NODE_TYPES.Identifier,
  AST_NODE_TYPES.MemberExpression,
] as const);
```
(note that you'll need to trivially adjust the call sites)

kirkwaiblinger (changes requested):
Few little nits but this is otherwise basically an approval! Great work on this!! It's delightful work on a real tricky part of the codebase 👍 Thanks for working on this!

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: indeed, it was the second part of my "workaround", that your suggestion fixed.

OlivierZal:
Fixed!

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
JoshuaKGoldberg, line 1: [Praise] Really nice refactors here. Love to see a -62 line count on a rule 😄

JoshuaKGoldberg (approved):
🚀 thanks for pushing through on this!

PR #10535, opened 2024-12-21
fix(eslint-plugin): [no-for-in-array] report on any type which may be an array or array-like

### LINKED ISSUE #10534: Bug: [no-for-in-array] fails to report on `Array<T> | undefined`

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gWygTwCMQBBGGALnkxgEsUBzAbQF14AfeAVxVADM6QwANwAoUJFgJk6LLApUMtBi1EiA9Gvi8oNCGkRIy4DBDwjeh%2BAAppmeDQwhs9lPDkBKeAG8AviPWaaAAWSJwQwFo6EAA0VFDYCADuDkHwAAYAKizs8ChhEGnmljaodg5OLjj4RKQwnr7%2BGvAADkiOKBg0UBCmVJz09CCYalCcGEj8AB7w4zNBCBY9SMkMAPxFMNa2WOXOE1WEJGTwq6vwLPV%2BQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDMPpbmtAIbRoQwuiiJRfSODABfEAqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
declare const maybeArr: string[] | undefined;
declare const arr: string[];

// fails correctly
for (const item in arr) {}

// should fail, same with `T[] | null`
for (const item in maybeArr) {}

// potentially suggest/autofix to the following?
for (const item of maybeArr ?? []) {}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-for-in-array": "error",
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expected the rule to report, as it doesn't make sense to iterate `null` or `undefined`. A fixer would be nice (in general, not just for this special case) since most of the time, the intent is to use `for-of`.

### Actual Result

The rule fails to catch these cases.

### Additional Info

Currently, the rule checks specifically for an array or union of arrays. I think this can be expanded to a union with an array and potentially `null` and `undefined`.

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10534

## Overview

See https://github.com/typescript-eslint/typescript-eslint/pull/10535#issuecomment-2564044282 for the updated contents of this PR.

~~This PR tackles #10534 and reports on cases such as:~~

```ts
declare const maybeArr: string[] | undefined;
declare const arr: string[];

// fails correctly
for (const item in arr) {}

// should fail, same with `T[] | null`
for (const item in maybeArr) {}
```

### DIFF
--- a/packages/eslint-plugin/src/rules/no-for-in-array.ts
+++ b/packages/eslint-plugin/src/rules/no-for-in-array.ts
@@ -1,10 +1,10 @@
+import * as tsutils from 'ts-api-utils';
 import * as ts from 'typescript';
 
 import {
   createRule,
   getConstrainedTypeAtLocation,
   getParserServices,
-  isTypeArrayTypeOrUnionOfArrayTypes,
 } from '../util';
 import { getForStatementHeadLoc } from '../util/getForStatementHeadLoc';
 
@@ -45,3 +45,20 @@ export default createRule({
     };
   },
 });
+
+function isTypeArrayTypeOrUnionOfArrayTypes(
+  type: ts.Type,
+  checker: ts.TypeChecker,
+): boolean {
+  for (const t of tsutils.unionTypeParts(type)) {
+    if (tsutils.isTypeFlagSet(t, ts.TypeFlags.Undefined | ts.TypeFlags.Null)) {
+      continue;
+    }
+
+    if (!checker.isArrayType(t)) {
+      return false;
+    }
+  }
+
+  return true;
+}
--- a/packages/eslint-plugin/src/rules/no-for-in-array.ts
+++ b/packages/eslint-plugin/src/rules/no-for-in-array.ts
@@ -1,10 +1,10 @@
-import * as tsutils from 'ts-api-utils';
 import * as ts from 'typescript';
 
 import {
   createRule,
   getConstrainedTypeAtLocation,
   getParserServices,
+  isBuiltinSymbolLike,
 } from '../util';
 import { getForStatementHeadLoc } from '../util/getForStatementHeadLoc';
 
@@ -33,8 +33,9 @@ export default createRule({
         const type = getConstrainedTypeAtLocation(services, node.right);
 
         if (
-          isTypeArrayTypeOrUnionOfArrayTypes(type, checker) ||
-          (type.flags & ts.TypeFlags.StringLike) !== 0
+          isArray(checker, type) ||
+          isRegExpExecArrayLike(services.program, type) ||
+          isArgumentsObjectType(type)
         ) {
           context.report({
             loc: getForStatementHeadLoc(context.sourceCode, node),
@@ -46,19 +47,32 @@ export default createRule({
   },
 });
 
-function isTypeArrayTypeOrUnionOfArrayTypes(
-  type: ts.Type,
-  checker: ts.TypeChecker,
-): boolean {
-  for (const t of tsutils.unionTypeParts(type)) {
-    if (tsutils.isTypeFlagSet(t, ts.TypeFlags.Undefined | ts.TypeFlags.Null)) {
-      continue;
-    }
+function isArgumentsObjectType(type: ts.Type): boolean {
+  return (
+    type.getSymbol()?.escapedName === ts.escapeLeadingUnderscores('IArguments')
+  );
+}
 
-    if (!checker.isArrayType(t)) {
-      return false;
-    }
+function isRegExpExecArrayLike(program: ts.Program, type: ts.Type): boolean {
+  return isTypeRecurser(type, t =>
+    isBuiltinSymbolLike(program, t, 'RegExpExecArray'),
+  );
+}
+
+function isArray(checker: ts.TypeChecker, type: ts.Type): boolean {
+  return isTypeRecurser(
+    type,
+    t => checker.isArrayType(t) || checker.isTupleType(t),
+  );
+}
+
+function isTypeRecurser(
+  type: ts.Type,
+  predicate: (t: ts.Type) => boolean,
+): boolean {
+  if (type.isUnionOrIntersection()) {
+    return type.types.some(t => isTypeRecurser(t, predicate));
   }
 
-  return true;
+  return predicate(type);
 }
--- a/packages/eslint-plugin/src/rules/no-for-in-array.ts
+++ b/packages/eslint-plugin/src/rules/no-for-in-array.ts
@@ -1,10 +1,10 @@
-import type * as ts from 'typescript';
+import * as tsutils from 'ts-api-utils';
+import * as ts from 'typescript';
 
 import {
   createRule,
   getConstrainedTypeAtLocation,
   getParserServices,
-  isBuiltinSymbolLike,
 } from '../util';
 import { getForStatementHeadLoc } from '../util/getForStatementHeadLoc';
 
@@ -32,7 +32,7 @@ export default createRule({
 
         const type = getConstrainedTypeAtLocation(services, node.right);
 
-        if (isArray(checker, type) || isArrayLike(services.program, type)) {
+        if (isArrayLike(checker, type)) {
           context.report({
             loc: getForStatementHeadLoc(context.sourceCode, node),
             messageId: 'forInViolation',
@@ -43,21 +43,23 @@ export default createRule({
   },
 });
 
-function isArrayLike(program: ts.Program, type: ts.Type): boolean {
-  return isTypeRecurser(type, t =>
-    isBuiltinSymbolLike(program, t, [
-      'IArguments',
-      'HTMLCollection',
-      'RegExpExecArray',
-      'NodeList',
-    ]),
+function isArrayLike(checker: ts.TypeChecker, type: ts.Type): boolean {
+  return isTypeRecurser(
+    type,
+    t => t.getNumberIndexType() != null && hasArrayishLength(checker, t),
   );
 }
 
-function isArray(checker: ts.TypeChecker, type: ts.Type): boolean {
-  return isTypeRecurser(
-    type,
-    t => checker.isArrayType(t) || checker.isTupleType(t),
+function hasArrayishLength(checker: ts.TypeChecker, type: ts.Type): boolean {
+  const lengthProperty = type.getProperty('length');
+
+  if (lengthProperty == null) {
+    return false;
+  }
+
+  return tsutils.isTypeFlagSet(
+    checker.getTypeOfSymbol(lengthProperty),
+    ts.TypeFlags.NumberLike,
   );
 }
 
--- a/packages/eslint-plugin/src/rules/no-for-in-array.ts
+++ b/packages/eslint-plugin/src/rules/no-for-in-array.ts
@@ -1,4 +1,4 @@
-import * as ts from 'typescript';
+import type * as ts from 'typescript';
 
 import {
   createRule,
@@ -35,7 +35,7 @@ export default createRule({
         if (
           isArray(checker, type) ||
           isRegExpExecArrayLike(services.program, type) ||
-          isArgumentsObjectType(type)
+          isArgumentsObjectType(services.program, type)
         ) {
           context.report({
             loc: getForStatementHeadLoc(context.sourceCode, node),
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/no-for-in-array.ts
```diff
@@ -45,3 +45,20 @@ export default createRule({
     };
   },
 });
+
+function isTypeArrayTypeOrUnionOfArrayTypes(
```
kirkwaiblinger, line 49: (style) avoid having the identical name function but with different behavior compared to the `isTypeArrayTypeOrUnionOfArrayTypes` in utils?

packages/eslint-plugin/src/rules/no-for-in-array.ts
```diff
@@ -45,3 +45,20 @@ export default createRule({
...
+    if (tsutils.isTypeFlagSet(t, ts.TypeFlags.Undefined | ts.TypeFlags.Null)) {
+      continue;
+    }
+
+    if (!checker.isArrayType(t)) {
+      return false;
+    }
+  }
+
+  return true;
```
kirkwaiblinger, line 63: (optional) This fixes a sort-of bug where a purely nullish type is considered an array type... "sort-of", since TS reports an error for the cases where the rule would reach this bug. 

```suggestion
  const nonNullishParts = tsutils
    .unionTypeParts(type)
    .filter(
      t =>
        !tsutils.isTypeFlagSet(t, ts.TypeFlags.Undefined | ts.TypeFlags.Null),
    );
  return (
    nonNullishParts.length > 0 &&
    nonNullishParts.every(t => checker.isArrayType(t))
  );
```

kirkwaiblinger: (test case)

[Playground](https://deploy-preview-10535--typescript-eslint.netlify.app/play#ts=5.7.2&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3igrhCAlmgBYBc2eE8APvDiqAGYEojADcAUAPTfwACGNAFoQADwAO4DKJgwkMTowXwAFMnRYA1vBYV8RYgEp4Ab3gBfIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDMPpbmtAIbRoQwuigB3EU0jgwAXxCKgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

```ts
declare const nullish: null | undefined;
// @ts-expect-error
for (const k in nullish) { }
```

kirkwaiblinger (changes requested):
Requesting changes here just to account for conversation on the backing issue https://github.com/typescript-eslint/typescript-eslint/issues/10534#issuecomment-2562339211

ronami:
I've updated the rule to report on anything that may be an array (along with some array-likes). Some notes:

- Array-likes currently include regex array's, as mentioned on https://github.com/typescript-eslint/typescript-eslint/issues/10558, along with the `arguments` object. There may be more that I've missed.
- I've added tuples so the rule warns on these too, as mentioned on https://github.com/typescript-eslint/typescript-eslint/issues/10558#issuecomment-2564004636.
- I fixed the bug mentioned on https://github.com/typescript-eslint/typescript-eslint/pull/10535#discussion_r1897763098, in which the rule reports in addition to TypeScript (along with a test case).

Remaining things that I think may benefit the rule (I'll consider opening issues for these):
- No suggestion fixer, even though most of the time (or so I think) it's a mistake and should instead be `for..of`.

packages/eslint-plugin/src/rules/no-for-in-array.ts
```diff
@@ -45,3 +46,33 @@ export default createRule({
     };
   },
 });
+
+function isArgumentsObjectType(type: ts.Type): boolean {
+  return (
+    type.getSymbol()?.escapedName === ts.escapeLeadingUnderscores('IArguments')
```
ronami, line 52: I'm not sure if there's a better way to recognize the type of the `arguments` object. Alternatively, I can rely on AST-alone and not have this use the types at all.

packages/eslint-plugin/src/rules/no-for-in-array.ts
```diff
@@ -45,3 +46,33 @@ export default createRule({
...
+function isTypeRecurser(
+  type: ts.Type,
+  predicate: (t: ts.Type) => boolean,
+): boolean {
+  if (type.isUnionOrIntersection()) {
+    return type.types.some(t => isTypeRecurser(t, predicate));
+  }
+
+  return predicate(type);
+}
```
ronami, line 78: I've used this part from https://github.com/typescript-eslint/typescript-eslint/pull/10551.

packages/eslint-plugin/src/rules/no-for-in-array.ts
```diff
@@ -33,8 +33,9 @@ export default createRule({
         const type = getConstrainedTypeAtLocation(services, node.right);
 
         if (
-          isTypeArrayTypeOrUnionOfArrayTypes(type, checker) ||
-          (type.flags & ts.TypeFlags.StringLike) !== 0
```
ronami, line 37: I'm not sure if this is necessary, as there's no test that fails if this is removed. TypeScript already reports this as an error, so I don't think the rule has to report this twice ([playground link](https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3mgXNjMAligOYDcAUBQGZIzwAUy6WU8x2AlPAN4UC%2BFIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDMPpbmtAIbRoQwuiiJRfSODABfEAqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)).

packages/eslint-plugin/src/rules/no-for-in-array.ts
```diff
@@ -45,3 +42,32 @@ export default createRule({
     };
   },
 });
+
+function isArrayLike(program: ts.Program, type: ts.Type): boolean {
+  return isTypeRecurser(type, t =>
+    isBuiltinSymbolLike(program, t, [
+      'IArguments',
+      'HTMLCollection',
```
ronami, line 50: https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection

kirkwaiblinger:
> Remaining things that I think may benefit the rule (I'll consider opening issues for these):
> 
> * No suggestion fixer, even though most of the time (or so I think) it's a mistake and should instead be `for..of`.

For this, the complexity is that 
```ts
for (const a in x) {
  console.log(a);
}
```
can't be directly fixed to a for-of loop equivalent since the for of would iterate array values and not string keys. So we'd have to be vigilant about keys only used for indexing or something. Note that [prefer-for-of](https://typescript-eslint.io/rules/prefer-for-of/) is in a similar area and may have design and logic for that that we could reuse. In any case punting to a separate issue is a good idea 👍

packages/eslint-plugin/src/rules/no-for-in-array.ts
```diff
@@ -45,3 +42,32 @@ export default createRule({
     };
   },
 });
+
+function isArrayLike(program: ts.Program, type: ts.Type): boolean {
```
kirkwaiblinger, line 46: I wonder how much hard-coding we can cut down on... 
for `RegExpExecArray` (and Array subclasses in general I think) `checker.isArrayLikeType(type)` will take care of it. 

The rest, though, require a bit more thinking. I was able to get all test cases passing with this approach:

```ts

import {
  createRule,
  getConstrainedTypeAtLocation,
  getParserServices,
} from '../util';
import { getForStatementHeadLoc } from '../util/getForStatementHeadLoc';

export default createRule({
  name: 'no-for-in-array',
  meta: {
    type: 'problem',
    docs: {
      description: 'Disallow iterating over an array with a for-in loop',
      recommended: 'recommended',
      requiresTypeChecking: true,
    },
    messages: {
      forInViolation:
        'For-in loops over arrays skips holes, returns indices as strings, and may visit the prototype chain or other enumerable properties. Use a more robust iteration method such as for-of or array.forEach instead.',
    },
    schema: [],
  },
  defaultOptions: [],
  create(context) {
    return {
      ForInStatement(node): void {
        const services = getParserServices(context);

        const type = getConstrainedTypeAtLocation(services, node.right);

        if (isArrayLike(services.program, type)) {
          context.report({
            loc: getForStatementHeadLoc(context.sourceCode, node),
            messageId: 'forInViolation',
          });
        }
      },
    };
  },
});

function isArrayLike(program: ts.Program, type: ts.Type): boolean {
  return isTypeRecurser(type, t => {

    const hasNumberyLength = (() => {
      const lengthProperty = t.getProperty('length');
      if (lengthProperty != null) {
        const lengthType = checker.getTypeOfSymbol(lengthProperty);
        if (lengthType.flags & ts.TypeFlags.NumberLike) {
          return true;
        }
      }
      return false;
    })();

    const hasNumberyIndex = (() => {
      const indexSignature = t.getNumberIndexType();
      return indexSignature != null;
    })();

    return hasNumberyIndex && hasNumberyLength;
  });
}

function isTypeRecurser(
  type: ts.Type,
  predicate: (t: ts.Type) => boolean,
): boolean {
  if (type.isUnionOrIntersection()) {
    return type.types.some(t => isTypeRecurser(t, predicate));
  }

  return predicate(type);
}

```

***

But I haven't thought deeply whether this is a good idea. What do you think?

packages/eslint-plugin/tests/fixtures/tsconfig.lib-dom.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "lib": ["es2015", "es2017", "esnext", "dom"]
+  }
+}
```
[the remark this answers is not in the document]
kirkwaiblinger: Yep, this is good!

packages/eslint-plugin/src/rules/no-for-in-array.ts
kirkwaiblinger: Yep, makes sense to me. Noting that neither TS nor the rule reports on 
```ts
declare const s: String;
for (const a in s) {
}
```
so yeah it doesn't seem like there's much use for those lines

kirkwaiblinger: Fun fact, this goes all the way back to 2016 in tslint: https://github.com/palantir/tslint/pull/1394/files#diff-f26fa47b522c451ad67d905bc6b48c8a6d99a24f01d8616f76abeabe800f807cR65

packages/eslint-plugin/src/rules/no-for-in-array.ts
ronami: Interesting approach, and it seems TypeScript does [some similar checks internally](https://github.com/microsoft/TypeScript/blob/cbac1ddfc73ca3b9d8741c1b51b74663a0f24695/src/compiler/checker.ts/#L24921-L24926) too. My intuition is to be careful with a check like that, in case it may flag false positives, but testing this on various cases, it seems to work nicely (though it is possible I've missed something). It does report on objects such as the following, which I think is intended (in addition to the `arguments` and several other array-likes):

```ts
declare const obj: {
  [key: number]: number;
  length: 1;
};

for (const a in obj) {
  //
}
```

I agree that hard-coding specific array-likes isn't very optimal. In my opinion, we should either use the plain `checker.isArrayLikeType()` and avoid reporting array-likes that it doesn't detect or use this custom way to check array-likes. I think I'm +1 to using this over the plain `checker.isArrayLikeType()`.

I've updated the PR to match this, thanks 👍

kirkwaiblinger (approved):
I'm happy!

JoshuaKGoldberg (approved):
+1, this is very sweet.

PR #10552, opened 2024-12-26
fix(eslint-plugin): [prefer-readonly] autofixer doesn't add type to property that is mutated in the constructor

### LINKED ISSUE #1056: Bug: [prefer-readonly] autofixer doesn't add type to property that is mutated in the constructor
**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/prefer-readonly": 2
  }
}
```

```TS
// your repro code case
class X {
  private _isValid? = true;
  get isValid(): boolean {
    return this._isValid;
  }

  constructor(data?: {}) {
    if (!data) {
      this._isValid = false;
    }
  }
}
```

**Expected Result**

No report

**Actual Result**

```sh
@typescript-eslint/prefer-readonly

Member '_isValid' is never reassigned; mark it as `readonly`
```

**Versions**

| `@typescript-eslint/eslint-plugin` | `2.3.3` |
| `@typescript-eslint/parser`        | `2.3.3` |
| `node`                             | `12.11.1` |
| `yarn`                             | `1.19.0` |

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1056

## Overview

This is a draft PR that attempts to tackle #1056. I've mentioned my thoughts about this at https://github.com/typescript-eslint/typescript-eslint/issues/1056#issuecomment-2562979820.


packages/eslint-plugin/src/rules/prefer-readonly.ts
```diff
@@ -229,13 +229,48 @@ export default createRule<Options, MessageIds>({
...
+            const violatingType = checker.getTypeAtLocation(violatingNode);
+
+            // if the RHS is a literal, its type would be narrowed, while the
+            // type of the initializer (which isn't `readonly`) would be the
+            // widened type
+            if (initializerType === violatingType) {
+              return null;
+            }
+
+            if (!tsutils.isLiteralType(initializerType)) {
```
JoshuaKGoldberg, line 254: [Bug] It's not just literal types though, it could also be [unions of literals (playground)](https://deploy-preview-10552--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.ts&code=MYGwhgzhAEAqCmEAu0DeAoa0AOAnAlgG5hLw64D220AvNAORj0Dc6m0wFAdsrgK7AkFXAAoAlGnZYkAC3wQAdHiq0GAIxZTos%2BUsrU69YJqwBfdOaA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wls4CGAEwD2TeIXRRe0EdEjgwAXxBKgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false):  

```ts
class Test {
  private prop = 'a';

  constructor() {
    this.prop = 'b';
    this.prop = 'c';
  }
}
```

...or a [template literal (playground)](https://deploy-preview-10552--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.ts&code=C4TwDgpgBAMhzAgJwAoEMkQHYGcoF4oADACgBIBvAcjSqgB8oqAjKgXwEoiAobgEwgBjADYZoggPa5gUMGNwAuWPESp5OANy8RaHHgAqEHDIrcospAEsAbmkQWJYAkxJoOVLWaiTpSAK6CwBJIJBxQpubmwAAWljgAdGBIjs5ymLha5mzc2UA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wls4CGAEwD2TeIXRRe0EdEjgwAXxBKgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false):

```ts
type LetterParens = `(${'a' | 'b'})`

declare const parens: LetterParens;

class Test {
  private prop = '(a)';

  constructor() {
    this.prop = parens;
  }
}
```

I think this'll have to use the assignability API.

kirkwaiblinger: I'm confused. Don't these examples currently have the desired behavior?

JoshuaKGoldberg: I suppose from a certain point of view they do. Hmm. I was thinking that the right fixed type here would be a union like `'a' | 'b' | 'c'` or `LetterParens`. But I suppose they do work fine with the more broad `| string`. Especially if it means we don't have to use the more nuanced assignability API.

Ok 👍 never mind from me then!

kirkwaiblinger: Yeah, my thinking was that it positively should be `string` and not `'a' | 'b' | 'c'` since that's the type that the member had before the fix (due to literal widening). So putting the widened type in the fix, and leaving it up to the user to explicitly narrow it down further if they want seems certainly least surprising to me. That being said, we could still turn it into a suggestion instead of a fix if we wanted to draw extra attention to it, since it might be something the user might wish to further refine? 🤷‍♂️

JoshuaKGoldberg: Heh I don't think I've ever seen a rule that provides a fix and a suggestion in the same place. Kind of a weird thing to do. I think it's fine as is.

packages/eslint-plugin/src/rules/prefer-readonly.ts
```diff
@@ -173,6 +173,38 @@ export default createRule<Options, MessageIds>({
...
 
+    function getTypeAnnotationForViolatingNode(
+      node: TSESTree.Node,
+      type: ts.Type,
+      initializerType: ts.Type,
+    ) {
+      const annotation = checker.typeToString(type);
+
+      // verify the about-to-be-added type annotation is in-scope
+      if (tsutils.isTypeFlagSet(initializerType, ts.TypeFlags.EnumLiteral)) {
```
ronami, line 184: I wrote about this on https://github.com/typescript-eslint/typescript-eslint/issues/1056#issuecomment-2562979820, but on some occasions (as far as I know only enums), the rule won't be able to add a type annotation:

```ts
// foo.ts
enum Foo {
  Bar,
  Bazz,
}

export const bar = Foo.Bar;

// index.ts
import { bar } from './foo';

export class Test {
  private foo = bar;
}
```

I chose not to add a type annotation on these cases, but an alternative I think is also valid is to have this be a suggestion instead of an auto-fix.

Would be happy to to hear opinions on this.

JoshuaKGoldberg: I have a vague memory of this coming up once in a while... I think it's fine for now. We don't have any kind of shared/standard plumbing for importing constructs in fixes or suggestions.

JoshuaKGoldberg (approved):
Animated text saying 'happy new year 2025' with cartoon fireworks

ronami:
Thinking some more about this: The rule will add a type annotation whenever literal widening happens, regardless of whether the property is mutated in the constructor or not.

I'm slightly leaning towards adding a type annotation *only* when literal widening happens *and* the property is mutated in the constructor (in addition to an initializer).

kirkwaiblinger:
> Thinking some more about this: The rule will add a type annotation whenever literal widening happens, regardless of whether the property is mutated in the constructor or not.
> 
> I'm slightly leaning towards adding a type annotation _only_ when literal widening happens _and_ the property is mutated in the constructor (in addition to an initializer).

I'm not sold on this. My concerns are
- It increases the level of surprise for the user by adding a special case within a special case
- It changes the public API of the class. This may absolutely be a _desired_ consequence in many cases! But it may also be breaking. This tilts the scales for more towards the conservative option of conspicuously preserving the original type, and letting the user easily opt in explicitly to the narrower one by removing the type annotation.

My order of most to least preferred approaches for this case would be 
1. Don't autofix. Offer suggestions for both possibilities.
2. Autofix with original, widened type (very close second, could probably be argued either way between 1 and 2)
3. Autofix with narrowed type.

But I'll defer to you guys if those points don't sway opinions. It's not a big thing.

ronami:
> > Thinking some more about this: The rule will add a type annotation whenever literal widening happens, regardless of whether the property is mutated in the constructor or not.
> > I'm slightly leaning towards adding a type annotation _only_ when literal widening happens _and_ the property is mutated in the constructor (in addition to an initializer).
> 
> I'm not sold on this. My concerns are
> 
> * It increases the level of surprise for the user by adding a special case within a special case
> * It changes the public API of the class. This may absolutely be a _desired_ consequence in many cases! But it may also be breaking. This tilts the scales for me more towards the conservative option of conspicuously preserving the original type, and letting the user easily opt in explicitly to the narrower one by removing the type annotation.
> 
> My order of most to least preferred approaches for this case would be
> 
> 1. (currently implemented) Autofix with original, widened type
> 2. Don't autofix. Offer suggestions for both possibilities (very close second, could probably be argued either way between 1 and 2)
> 3. Autofix with narrowed type.
> 
> But I'll defer to you guys if those points don't sway opinions. It's not a big thing.

Interesting take!

I didn't put much thought into it, but I agree that it can be confusing. Adding a type annotation whenever literals are involved simplifies how the rule behaves.

I considered the type changing a desired consequence, and I think it will be a breaking change pretty rarely, but it is possible.

For me, the increase in the level of surprise is a big reason to keep it simple and always auto-fix with the widened type.

kirkwaiblinger:
Oh, I've misunderstood, sorry! - this rule _only_ applies to `private`/`#private` propreties. So there isn't a public API concern! In that case I'm on board with only adding the annotation if they actually are mutated 👍

ronami:
> Oh, I've misunderstood, sorry! - this rule _only_ applies to `private`/`#private` propreties. So there isn't a public API concern! In that case I'm on board with only adding the annotation if they actually are mutated 👍

@kirkwaiblinger I agree with your argument that this change increases the level of surprise for the user by adding a special case within a special case.

I think both options are viable (always adding a type annotation for literals or only when literals are modified), and the difference between the two is subtle. Still, I'm leaning toward not adding this special case. I think it makes the auto-fix a lot more straightforward. I don't want the auto-fix to create confusion regarding when or why it adds a type annotation.

I'm OK with going either way, but I would love to hear another opinion on this.

ronami:
> Oh, I've misunderstood, sorry! - this rule _only_ applies to `private`/`#private` propreties. So there isn't a public API concern! In that case I'm on board with only adding the annotation if they actually are mutated 👍

@kirkwaiblinger I agree with your point that this change increases the level of surprise for the user by adding a special case within a special case.

I think both options are viable (always adding a type annotation for literals or only when literals are modified in the constructor), and the difference between the two is subtle. I was going back and forth between what I consider the optimal solution here, but I think I'm +1 for adding a type annotation only when necessary, even though it adds some surprise factor.

I've updated the PR to match this 👍

packages/eslint-plugin/src/rules/prefer-readonly.ts
```diff
@@ -229,13 +261,62 @@ export default createRule<Options, MessageIds>({
...
+              return null;
+            }
+
+            if (esNode.typeAnnotation || !esNode.value) {
+              return null;
+            }
+
+            if (nameNode.type !== AST_NODE_TYPES.Identifier) {
+              return null;
+            }
```
ronami, line 275: To my understanding, this can realistically only be an identifier. This means some cases that should have a type annotation added won't receive it, for example:

```ts
class TestOverlappingClassVariable {
  private 'overlappingClassVariable' = 7;

  constructor() {
    this.overlappingClassVariable = 10;
  }
}
```

This seems like an existing bug that's not necessarily related to this addition. I've opened #10655 as a separate issue to address this.

JoshuaKGoldberg (approved):
LGTM - and such comprehensive unit tests! 😄 

~~Will let you or @kirkwaiblinger merge if you're happy.~~ Edit: it's been a few days and we're about to run a release - merging myself. :shipit:

kirkwaiblinger (approved):

Issue #7112, opened 2023-06-17
Repo: Investigate moving from Jest to Vitest


JoshuaKGoldberg:
### Suggestion

Right now, we use [Jest](https://jestjs.io) in this repository for unit tests. It works well. But, I've noticed unit tests starting to take quite a while. It'd be nice if we could reduce that speed.

[Vitest](https://vitest.dev) purports to be much faster than Jest (and in my personal experience, is). If someone could send a draft PR with our repo migrated over to Vitest and all tests passing, it'd be nice to see:

* What would the actual code changes look like?
* How much faster would Vitest be?

Bonus points if someone can find or write codemods to do the migration automatically...!

rielAsh24:
Hey @JoshuaKGoldberg! I'd be glad to look into this

wojtekmaj:
I looked into this and I've managed to get tests fully green in 4 workspaces, and fell short by just one suite in 2. Main challenges and thoughts:

- Vitest fails to run any test file without test suites defined, and there are some test suites created dynamically.
- process.chwd() can't be used in threads, so unless removed, you'll need to run Vitest without threads, losing most if not all of the performance of the benefits.
- jest-specific-snapshot is not necessary, as there is .toMatchFileSnapshot() in Vitest, which is a huge win, but will require updating thousands of the "specific snapshots" as in Vitest, they don't have metadata, just contents.
- Mocking typescript module doesn't seem to work for me
- Removing require() to package.json of non-existing virtual packages is necessary - it's not possible to mock non-existing packages using Vitest at the moment.

Regarding performance...

Perhaps most interesting, because one of the slowest, is typescript-estree test, one of the ones that "almost pass", which I ran with parse.test.ts deleted.

Here's Jest:

```
> nx run typescript-estree:test

$ jest --coverage
 PASS  tests/lib/getProjectConfigFiles.test.ts
 PASS  tests/lib/node-utils.test.ts
 PASS  tests/lib/describeFilePath.test.ts
 PASS  tests/lib/convert.test.ts
 PASS  tests/lib/warn-on-unsupported-ts.test.ts
 PASS  tests/lib/semanticInfo-singleRun.test.ts
 PASS  tests/lib/parse.project-true.test.ts
 PASS  tests/lib/createParseSettings.test.ts
 PASS  tests/lib/parse.moduleResolver.placeholder-error.test.ts (6.313 s)
[...]

PR #10333, opened 2024-11-15
feat(typescript-eslint): allow infinitely deep array nesting in config function and extends

bradzacher:
## PR Checklist

- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

This has been a pet peeve of mine for a long while now and I've been meaning to action it as a big DevX improvement.

This PR simply allows users to infinitely nest arrays in the `tseslint.config` utility input AND in the `extends` key.

The docs updates are proof enough of how much this simplifies things.

Note that [ESLint upstream is planning on allowing nested arrays](https://github.com/eslint/rfcs/pull/126) which is what spurred me to build it right now.

PR #10579, opened 2024-12-31
chore(eslint-plugin): migrate to vitest

43081j:
## PR Checklist

## Overview

this is just one possible PR for migrating to vitest

we need to avoid an unreviewable, giant PR, so the following has been done in this one:

- only migrate eslint-plugin for now
- for docs snapshots, retain a similar format rather than moving to the standard one-snapshot-per-file
- for all other snapshots, just update the snapshot (since they only had one per file already)

Depends on #10582


JoshuaKGoldberg:
This is awesome, thanks for sending @43081j! I'd really like to dive in soon and see if we can get this merged. But we're a little swamped catching up on issues after the holiday & new year break, and this is a lot of changes. So it'll probably be a little while till we can take a deep look. 😬 

In the meantime, some notes:

* `eslint-plugin` is one of our biggest packages. Is there a reason you wanted to start with this, and not a smaller one like `eslint-plugin-internal` or `type-utils`?
* Codecov is reporting a significant % drop, which means coverage is probably not set up right with it - could you take a look please?
* Although we all would prefer to move away from implicit globals, adding in the `import { describe, it, ... } from 'vitest'` is adding a lot to the diff. Is there a reason not to enable [Vitest `globals`](https://vitest.dev/config/#globals) for now - with a followup to remove it later?

43081j:
> eslint-plugin is one of our biggest packages. Is there a reason you wanted to start with this, and not a smaller one like eslint-plugin-internal or type-utils?

no particular reason, i just chose the package i was looking at at the time

> Codecov is reporting a significant % drop, which means coverage is probably not set up right with it - could you take a look please?

ill have a look

> Although we all would prefer to move away from implicit globals, adding in the import { describe, it, ... } from 'vitest' is adding a lot to the diff. Is there a reason not to enable [Vitest globals](https://vitest.dev/config/#globals) for now - with a followup to remove it later?

we could use `globals` but i would say the diff is actually mostly from snapshot changes than it is from imports.

i think without the snapshot changes, it'd be a pretty small diff. so i'd prefer to keep them, but let me know if you don't

JoshuaKGoldberg:
> i think without the snapshot changes, it'd be a pretty small diff. so i'd prefer to keep them, but let me know if you don't

+1 from me on the small diff 😄. I'd prefer not to keep them. The smaller the diff the better!

> if you could have a look at some point too, that'd be helpful

👍 ACK, we can help as part of the review if it's not figured out

43081j:
> +1 from me on the small diff 😄. I'd prefer not to keep them. The smaller the diff the better!

what i meant was that the imports don't add much to the diff, since the only reason this is a "big PR" is the snapshot changes rather than those imports.

i'd rather keep the imports seeing as its where we want to be

if you diff without snapshots, its small (but of course we need those snapshot updates in this PR)

as for the coverage change, im pretty sure it is just vite doing a much better job by using v8 coverage. that and the added `tools/` directory seem to account for the difference

ronami:
+1 from me on having the smallest diff possible. Since this is already a big change, having it be gradual is a big plus. I think that moving from [Vitest `globals`](https://vitest.dev/config/#globals) to using imports can be done separately on a future PR, which would help not miss anything on this PR.

43081j:
i've now switched to using globals

we need to re-open #10582 and land #10680 before this

kirkwaiblinger (changes requested):
As mentioned in https://github.com/typescript-eslint/typescript-eslint/pull/10680#issuecomment-2601969922 would you mind please removing the unrelated `test()` -> `it()` changes?

packages/typescript-eslint/package.json
```diff
@@ -48,7 +48,8 @@
     "postclean": "rimraf dist && rimraf _ts4.3 && rimraf coverage",
     "format": "prettier --write \"./**/*.{ts,mts,cts,tsx,js,mjs,cjs,jsx,json,md,css}\" --ignore-path ../../.prettierignore",
     "lint": "nx lint",
-    "test": "jest --coverage --passWithNoTests",
+    "test": "jest --passWithNoTests",
```
JoshuaKGoldberg, line 51: Oh, we don't need `--passWithNoTests` anymore. The package actually has tests now as of #10333. I forgot this was here 😄

JoshuaKGoldberg (changes requested):
OK! This is looking great - just a few changes left to iron out.

Despite the large file count, this really didn't impact all that many different types of files. 251 of them were schema snapshots. I looked at around a half dozen of each folder of those and then ran this to mark them as viewed locally:

```js
const fileInfos = Array.from(document.querySelectorAll(`.file-info`)).filter(
  (node) => node.textContent.trim().endsWith(".shot")
);

console.log(fileInfos.length, "schema snapshots");

for (const fileInfo of fileInfos) {
  const input = fileInfo.parentElement.parentElement.querySelector(
    "input[type=checkbox]"
  );

  if (!input.checked) {
    input.click();
  }
}
```

43081j:
have caught up from main and applied the suggestions 👍

JoshuaKGoldberg (dismissed):
🚀 This looks great to me, thanks a million for sticking with it! Love to see the +2.3k 🟩 -4.3k 🟥. 

We should definitely get review from another team member such as @kirkwaiblinger given how they've been looking at it & it's a pretty important change.

Also marking this blocked as resolving the `test-coverage` / `test --coverage` discussion #10582.

packages/eslint-plugin/vitest.setup.ts
aryaemami59, line 1: This doesn't seem to be necessary.

packages/eslint-plugin/package.json
```diff
@@ -56,8 +56,8 @@
     "generate:breaking-changes": "tsx tools/generate-breaking-changes.mts",
     "generate:configs": "npx nx generate-configs repo",
     "lint": "npx nx lint",
-    "test": "cross-env NODE_OPTIONS=\"--experimental-vm-modules\" jest --logHeapUsage",
-    "test-single": "cross-env NODE_OPTIONS=\"--experimental-vm-modules\" jest --no-coverage",
+    "test": "vitest run",
```
kirkwaiblinger, line 59: Let's align these with the other vitest PRs?
```suggestion
    "test": "vitest --run --config=$INIT_CWD/vitest.config.mts",
```

packages/eslint-plugin/package.json
```diff
@@ -56,8 +56,8 @@
...
+    "test-single": "vitest run",
```
kirkwaiblinger, line 60:
```suggestion
    "test-single": "vitest --run --config=$INIT_CWD/vitest.config.mts",
```

kirkwaiblinger (approved):
Looks good to me!

eslint.config.mjs
```diff
@@ -403,6 +403,18 @@ export default tseslint.config(
       '@typescript-eslint/no-unsafe-call': 'off',
       '@typescript-eslint/no-unsafe-member-access': 'off',
       '@typescript-eslint/no-unsafe-return': 'off',
+    },
```
kirkwaiblinger, line 406: refer to https://github.com/typescript-eslint/typescript-eslint/pull/10978#discussion_r2007453476

packages/eslint-plugin/package.json
43081j: Do we know why? 

We should be able to run `vitest run` in all of them and auto discover the config

Is vitest not discovering mts config?

JoshuaKGoldberg:
Starting to see this once in a while:

```plaintext
Error: Test timed out in 5000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
```

https://cloud.nx.app/runs/EisURTJuXm/task/eslint-plugin%3Atest

Something to keep an eye on...

PR #10593, opened 2025-01-02
fix(eslint-plugin): [no-shadow] ignore ordering of type declarations

### LINKED ISSUE #5935: Bug: [no-shadow] Doesn't catch shadowing of global types in generics

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.4&sourceType=module&code=MYGwhgzhAECyCeBhcUA8AJAKrAMgURAFMBbQgOwBcA+aAbwFgAoJgXyaYvgAdDpEBXCBQD2xTN14BeaGX7EARoQBOAbnaNQkGAmRbMAd2GoBQ0eJ40GzRm2uce0LLgIlyFaNKFKAlmQDmKkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1uQAsBDACYcA7uiiJo0DtEjgwAXxDygA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

*See Playground link above*

### ESLint Config

*See Playground link above*

### tsconfig

*See Playground link above*

### Expected Result

I expected both the `HTMLElement` generic and the `HTMLElement` type declaration to both be an error as they both shadow the global type of the same name. The type declaration does show an error (actually a TypeScript compiler error) but there is no error for shadowing the type via a generic. I'd expect `no-shadow` to catch shadowing with generics.

### Actual Result

Only types which are defined in the module scope are identified as shadowing when used in generic declarations.

### Additional Info

_No response_

### Versions

| `@typescript-eslint/eslint-plugin`     | `5.42.1` |
| `@typescript-eslint/parser`            | `5.42.1` |

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5935

## Overview

This PR attempts to tackle #5935.

Some thoughts/questions:

- I've expanded the existing `hoist` configuration to include `types` and `functions-and-types`. I don't think this is optimal, as permutations need to be manually set (e.g. `functions-and-types`). It also adds some complexity to this configuration, but I'm not sure if there's a non-breaking alternative.

  Since this is an extension rule, this somewhat changes the [base rule](https://eslint.org/docs/latest/rules/no-shadow#hoist)'s option, though I did see other rules like [`no-empty-function`](https://github.com/typescript-eslint/typescript-eslint/blob/4dbf48b8eec0321babb4293a1be71b54dd00c018/packages/eslint-plugin/src/rules/no-empty-function.ts#L37-L52) which extend an enum type of their base rule).

  That said, I'm not sure if someone will ever want to configure this to just `functions` or just `types`. It may be better to have the current `functions` options include `types` behavior and rename it in a future major (although this would mean the option would diverge from the base rule).
  
  I'm a bit unsure about this, so I'll be happy to hear other opinions.
- I've changed the default from `functions` to `functions-and-types`. This seems like the most sensible option, as for both, ordering doesn't matter.
- I've included documentation for the added option, similar to how other extension tests do this. I wasn't sure whether to use the correct/incorrect `<Tabs>` or not. I have decided not to use `<Tabs>` to match the rest of the rule's documentation.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-shadow.ts
+++ b/packages/eslint-plugin/src/rules/no-shadow.ts
@@ -11,7 +11,7 @@ type Options = [
   {
     allow?: string[];
     builtinGlobals?: boolean;
-    hoist?: 'all' | 'functions' | 'never';
+    hoist?: 'all' | 'functions' | 'functions-and-types' | 'never' | 'types';
     ignoreFunctionTypeParameterNameValueShadow?: boolean;
     ignoreOnInitialization?: boolean;
     ignoreTypeValueShadow?: boolean;
@@ -26,6 +26,13 @@ const allowedFunctionVariableDefTypes = new Set([
   AST_NODE_TYPES.TSDeclareFunction,
   AST_NODE_TYPES.TSConstructSignatureDeclaration,
   AST_NODE_TYPES.TSConstructorType,
+]);
+
+const functionsHoistedNodes = new Set([AST_NODE_TYPES.FunctionDeclaration]);
+
+const typesHoistedNodes = new Set([
+  AST_NODE_TYPES.TSInterfaceDeclaration,
+  AST_NODE_TYPES.TSTypeAliasDeclaration,
 ]);
 
 export default createRule<Options, MessageIds>({
@@ -63,7 +70,7 @@ export default createRule<Options, MessageIds>({
             type: 'string',
             description:
               'Whether to report shadowing before outer functions or variables are defined.',
-            enum: ['all', 'functions', 'never'],
+            enum: ['all', 'functions', 'functions-and-types', 'never', 'types'],
           },
           ignoreFunctionTypeParameterNameValueShadow: {
             type: 'boolean',
@@ -88,7 +95,7 @@ export default createRule<Options, MessageIds>({
     {
       allow: [],
       builtinGlobals: false,
-      hoist: 'functions',
+      hoist: 'functions-and-types',
       ignoreFunctionTypeParameterNameValueShadow: true,
       ignoreOnInitialization: false,
       ignoreTypeValueShadow: true,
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/tests/rules/no-shadow/no-shadow.test.ts
```diff
@@ -547,6 +547,364 @@ let y;
...
+            shadowedColumn: 5,
+            shadowedLine: 2,
+          },
+          messageId: 'noShadow',
+          type: AST_NODE_TYPES.Identifier,
+        },
+      ],
+      languageOptions: { parserOptions: { ecmaVersion: 6 } },
+      options: [{ hoist: 'functions' }],
+    },
```
ronami, line 568: This isn't directly related to the PR, but it seemed there are very few tests that cover the `functions` option in `hoist` (which now has two extra options).

packages/eslint-plugin/src/rules/no-shadow.ts
```diff
@@ -63,7 +70,7 @@ export default createRule<Options, MessageIds>({
             type: 'string',
             description:
               'Whether to report shadowing before outer functions or variables are defined.',
-            enum: ['all', 'functions', 'never'],
+            enum: ['all', 'functions', 'functions-and-types', 'never', 'types'],
```
JoshuaKGoldberg, line 73: > I've expanded the existing `hoist` configuration to include `types` and `functions-and-types`. I don't think this is optional, as permutations need to be manually set ...

Yeah, I'm thinking this exposes a general learning: don't use `enum` option types for _"when do I X?"_ questions. Eventually folks will want to do something that could be better expressed as an object. I.e. here:

```ts
{ optionName: { functions: true, types: true, variables: true } }
```

...but, to your point, I think it would be even more disruptive to extend the base rule option to a wholly different format in that way. For now I think the `enum` is fine. If this rule gets even more complicated over time then we can always switch to the object later.

tl;dr: agreed as-is 👍

packages/eslint-plugin/tests/rules/no-shadow/no-shadow.test.ts
JoshuaKGoldberg: 👍👍👍 on adding in missing test coverage! Especially if it's at all related to a rule being changed.

JoshuaKGoldberg (dismissed):
🚀

packages/eslint-plugin/src/rules/no-shadow.ts
```diff
@@ -513,15 +520,30 @@ export default createRule<Options, MessageIds>({
...
-          !outerDef ||
-          outerDef.node.type !== AST_NODE_TYPES.FunctionDeclaration)
-      );
+      if (!inner || !outer || inner[1] >= outer[0]) {
+        return false;
+      }
+
+      if (!outerDef) {
+        return true;
+      }
```
ronami, line 529: I went over this again and made one small change. I think I've introduced an unnecessary change to the original implementation (returning `false` instead of `true` if `outerDef` is `false`).

I don't think this code is reachable (all tests seem to pass regardless of this change), and I couldn't reproduce this in a test. Still, just to be sure, I've changed this to match the original implementation.

ronami: Now that it's on its own `if` statement, it fails codecov (even though this wasn't covered previously but was in a single long binary operation). I'm a bit unsure what the best way to tackle this is. I'd normally use `nullThrows` or a type assertion, but I'm not entirely sure this really is an unreachable line.

JoshuaKGoldberg: Yeah I'd just ignore Codecov here. Its reporter isn't perfect.

JoshuaKGoldberg (approved):
👍 from me. Even if the `outerDef` issue is a bug, I suspect it'd be an existing one - and probably not as impactful as the root issue this PR is fixing. Good note!

PR #10599, opened 2025-01-03
feat: standardize debug namespaces to file paths

### LINKED ISSUE #5872: Repo: Standardize debug namespaces per-file

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5872

## Overview

Adds a new `@typescript-eslint/internal/debug-namespace` lint rule that makes sure any `const log = debug('...')` uses a namespace equivalent to the package name, file directory, and file path. 

This caught several discrepancies in existing namespaces. I don't think we need to count this as a breaking change as those namespaces really aren't part of the public API. But this does constitute a `feat:` I think.

💖

### DIFF
--- a/packages/eslint-plugin-internal/src/rules/debug-namespace.ts
+++ b/packages/eslint-plugin-internal/src/rules/debug-namespace.ts
@@ -1,0 +1,66 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+
+import { getStaticValue } from '@typescript-eslint/utils/ast-utils';
+
+import { createRule } from '../util';
+
+function filePathToNamespace(filePath: string) {
+  const matched =
+    /typescript-eslint[\\/]+packages[\\/]+([a-z-]+)[\\/]+(.+)/.exec(filePath);
+  if (!matched) {
+    return;
+  }
+  const [, packageName, relativeFilePath] = matched;
+
+  const relativeFilePathProcessed = relativeFilePath
+    .replace(/^(?:dist|lib|src)\//, '')
+    .replace(/\.\w+$/, '')
+    .replaceAll(/[^a-z0-9-]+/gi, ':');
+
+  return `typescript-eslint:${packageName}:${relativeFilePathProcessed}`;
+}
+
+export default createRule({
+  name: __filename,
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Enforce consistent debug() namespaces based on package name and file path',
+    },
+    fixable: 'code',
+    messages: {
+      mismatched:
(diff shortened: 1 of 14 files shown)


packages/eslint-plugin-internal/src/rules/debug-namespace.ts
```diff
@@ -0,0 +1,66 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+
+import { getStaticValue } from '@typescript-eslint/utils/ast-utils';
+
+import { createRule } from '../util';
+
+function filePathToNamespace(filePath: string) {
+  const matched =
+    /typescript-eslint[\\/]+packages[\\/]+([a-z-]+)[\\/]+(.+)/.exec(filePath);
```
bradzacher, line 9: Probably better to just make this a path relative to this file? Eg `pkgsRoot = path.resolve('..', '..', '..'); relativePath = file.replace(pkgsRoot, '..');`

That way the rule doesn't rely on the folder name.

bradzacher (approved):

packages/eslint-plugin-internal/src/rules/debug-namespace.ts
JoshuaKGoldberg: Ah, those irksome to deal with in unit tests with explicit `filename`s. But +1 that this regex is unseemly. I switched to `.split(/packages[\\/]+/)` and added a bit more testing.

bradzacher (approved):

PR #10633, opened 2025-01-07
fix(eslint-plugin): [no-base-to-string] don't crash for recursive array or tuple types

### LINKED ISSUE #10632: Bug: [no-base-to-string] stack overflow with recursive type

### Playground Link

https://typescript-eslint.io/play#ts=5.7.2&fileType=.tsx&code=KYDwDg9gTgLgBDAnmYcBqBDANgV1QXgCg44AfOHAOwBNgAzAS0uGuLLgCMIItgNK25SjgC2HYFEFwAzjChMA5lIAiGGMCkBVJjAAcAQShQMiKZlzAA2gF0pAZTk4AxjADchQqEiwEyVA6hneHw4AG84SwBrYEQALhk5RWt48zw4AF93QjoqFwYISjg6bgAKADcU7DwASjC2JwLpHmAAOiwIBRKAAwAJBjgAElCy9K7q93SgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oCMBDZRLXxdk%2BKkwDm6KImjQO0SODABfECqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
export type Value =
  | undefined
  | boolean
  | number
  | string
  | Date
  | Uint8Array
  | Value[]
  | Struct;

export type Struct = { [key: string]: Value };

function foo(v: Value) {
  console.log(`Hi ${v}`);  // causes rule to crash
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-base-to-string": "error"
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

No crash

### Actual Result

Rule crashes with stack trace like this:

```
RangeError: Maximum call stack size exceeded
Occurred while linting ...
Rule: "@typescript-eslint/no-base-to-string"
    at getNameOfSymbolAsWritten (.../node_modules/typescript/lib/typescript.js:59075:36)
    at createAccessFromSymbolChain (.../node_modules/typescript/lib/typescript.js:56976:25)
    at symbolToTypeNode (.../node_modules/typescript/lib/typescript.js:56953:26)
    at typeToTypeNodeWorker (.../node_modules/typescript/lib/typescript.js:55397:16)
    at typeToTypeNodeHelper (.../node_modules/typescript/lib/typescript.js:55247:24)
    at mapToTypeNodes (.../node_modules/typescript/lib/typescript.js:56233:28)
    at typeReferenceToTypeNode (.../node_modules/typescript/lib/typescript.js:55913:33)
    at visitAndTransformType (.../node_modules/typescript/lib/typescript.js:55733:24)
    at typeToTypeNodeWorker (.../node_modules/typescript/lib/typescript.js:55402:28)
    at typeToTypeNodeHelper (.../node_modules/typescript/lib/typescript.js:55247:24)
```

### Additional Info

_No response_

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10632

## Overview

This PR addresses #10632 and fixes the rule so it does not fail on recursive array or tuple types. It looks like this regression has been introduced in #10437.

---

Not directly related: It seems not intentional, but since #10437, the rule reports invalid nested array types (which it didn't previously):

```ts
declare const s: { foo: 'bar' }[][];

// [object Object]
s.join();
```

Previously, the array check (which used to run only on `array.join()`) never checked for nested array types. This changed on #10437, and the recursive check also runs as part of `array.join()` (which means the crash reproduces on recursive `array.join()` too, [playground link](https://typescript-eslint.io/play/#ts=5.7.2&fileType=.tsx&code=KYDwDg9gTgLgBDAnmYcBqBDANgV1QXgCg44AfOHAOwBNgAzAS0uGuLLgCMIItgNK25SjgC2HYFEFwAzjChMA5lIAiGGMCkBVJjAAcAQShQMiKZlzAA2gF0pAZTk4AxjADchQqEiwEyVA6hneHw4AG84SwBrYEQALhk5RWt48zw4AF93QjoqFwYISjg6bgAKADcU7DwASjC2Bjo4EsNjRAA6BmkWk3Lq2tC2EjK2gCsIJhLq9xJ0wnSgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oCMBDZRLXxdk%2BKkwDm6KImjQO0SODABfECqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)).

There's no test that checks this behavior (which I think is intended), so I've added a couple of tests for this. Please let me know if this isn't intended behavior or if this should be done on a separate PR.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-base-to-string.ts
+++ b/packages/eslint-plugin/src/rules/no-base-to-string.ts
@@ -74,6 +74,7 @@ export default createRule<Options, MessageIds>({
       }
       const certainty = collectToStringCertainty(
         type ?? services.getTypeAtLocation(node),
+        new Set(),
       );
       if (certainty === Usefulness.Always) {
         return;
@@ -92,8 +93,9 @@ export default createRule<Options, MessageIds>({
     function checkExpressionForArrayJoin(
       node: TSESTree.Node,
       type: ts.Type,
+      visited: Set<ts.Type>,
     ): void {
-      const certainty = collectJoinCertainty(type);
+      const certainty = collectJoinCertainty(type, visited);
 
       if (certainty === Usefulness.Always) {
         return;
@@ -140,9 +142,14 @@ export default createRule<Options, MessageIds>({
       return Usefulness.Never;
     }
 
-    function collectTupleCertainty(type: ts.TypeReference): Usefulness {
+    function collectTupleCertainty(
+      type: ts.TypeReference,
+      visited: Set<ts.Type>,
+    ): Usefulness {
       const typeArgs = checker.getTypeArguments(type);
-      const certainties = typeArgs.map(t => collectToStringCertainty(t));
+      const certainties = typeArgs.map(t =>
+        collectToStringCertainty(t, visited),
+      );
       if (certainties.some(certainty => certainty === Usefulness.Never)) {
         return Usefulness.Never;
       }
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/tests/rules/no-base-to-string.test.ts
```diff
@@ -483,6 +483,28 @@ function foo<T>(x: T) {
...
+String(v);
+    `,
+    `
+type Value = (string | Value)[];
+declare const v: Value;
+
+String(v);
+    `,
+    `
+type Value = [Value, Value];
```
ronami, line 499: To my understanding, types like this should be valid to stringify (also similar ones like `type Value = Value[]`).

packages/eslint-plugin/src/rules/no-base-to-string.ts
```diff
@@ -154,39 +161,57 @@ export default createRule<Options, MessageIds>({
...
-      return collectToStringCertainty(elemType);
+      return collectToStringCertainty(elemType, visited);
     }
 
-    function collectJoinCertainty(type: ts.Type): Usefulness {
+    function collectJoinCertainty(
+      type: ts.Type,
+      visited: Set<ts.Type>,
+    ): Usefulness {
       if (tsutils.isUnionType(type)) {
```
kirkwaiblinger, line 179: I haven't been able to come up with a test case, but should this have a short-circuit as well?
```suggestion
      if (visited.has(type) {
        return Usefulness.Always;
      }
      
      if (tsutils.isUnionType(type)) {
```

packages/eslint-plugin/src/rules/no-base-to-string.ts
```diff
@@ -92,8 +93,9 @@ export default createRule<Options, MessageIds>({
     function checkExpressionForArrayJoin(
```
kirkwaiblinger, line 93: nit - would it make more sense for this function's signature to be unchanged, and just create a new set with the `collectJoinCertainty()` call, i.e. `collectJoinCertainty(type, new Set())`?

packages/eslint-plugin/src/rules/no-base-to-string.ts
ronami: Good point!

I thought about this, too, and I don't think so, but I'm not 100% sure about this.

To my understanding, a recursive type is valid if its recursion is through a box type (e.g. array, promise):

```ts
type Value = Value[];
```

A recursive type without it should show up as an error:

```ts
type Value = Value | Value[];
//   ~~~~~
// Type alias 'Value' circularly references itself.
```

Even with intersections in play, I still don't think there can be a valid recursion that doesn't use a boxed value (currently only an array).

I've tried to come up with such a case but couldn't find anything that would make the rule crash. I don't think the rule can crash without using an array type (which the current implementation covers).

packages/eslint-plugin/src/rules/no-base-to-string.ts
ronami: Thanks!

I don't like the number of times `visited` has to be passed around on the current implementation; simplifying some is really nice!

kirkwaiblinger (approved):
gif

JoshuaKGoldberg (approved):
Ryan Gosling as Ken in Barbie confidently putting on sunglasses atop an existing pair

PR #10639, opened 2025-01-11
feat(eslint-plugin): [no-unnecessary-condition] make `allowConstantLoopConditions` more granular

### LINKED ISSUE #7047: Enhancement: [no-unnecessary-condition] Make allowConstantLoopConditions more granular

https://typescript-eslint.io/rules/no-unnecessary-condition

### Description

The [`allowConstantLoopConditions` option](https://typescript-eslint.io/rules/no-unnecessary-condition/#allowconstantloopconditions) is disabled by default, because cases like `while (shouldRun = true)` are likely errors. But that means the rule option also flags cases like `while (true)`. Which is more of a stylistic preference, I think.

How about making the option more granular:

* `'never'`: same as today's `false`
* `'only-simple'`: allow "simple" cases like `true` _(a better name would be nice!)_
* `'always'`: same as today's `true`

...and in future major version, breaking change to:

* current: `'never'` is be the default value and aliased by `false` _(to not be a breaking change)_
* next: `'only-simple'` becomes the default value

### Fail

```typescript
let shouldRun = true;

while (shouldRun = true) { /* ... */ }
```

### Pass

```typescript
while (true) { /* ... */ }
```

### Additional Info

I stumbled onto this in https://github.com/JoshuaKGoldberg/template-typescript-node-package/pull/469/files#diff-e1ce717e1179a0db14e90ec5374768a206651ca807db58352991eb7895ed7c9cR64

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7047

## Overview

This PR attempts to tackle #7047 and changes `allowConstantLoopConditions` from a `boolean` to the following `enum`:
- `'never'`: same as the current `false`
- `'neverExceptWhileTrue'`: disallows constant conditions in all loops except `while` loops with the literal `true`.
- `'always'`: same as the current `true`

This is a non-breaking change as the existing options map to the new option (minus `neverExceptWhileTrue`).

---

Some thoughts:

I've used a similar option name for `'neverExceptWhileTrue'` as [`no-constant-condition`'s `checkLoops`](https://eslint.org/docs/latest/rules/no-constant-condition#checkloops). The option name is somewhat limiting because it's so specific, it may be better to have an option name that's a bit less specific to allow future changes(?).

The original issue suggests that the default may be changed on a future major. However, since `'neverExceptWhileTrue'` is similar to the current default (`'never'`) but allows for an additional use-case, I think it can be considered to become the new default, especially since the pattern it allows is relatively common (or so I think).

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
@@ -124,21 +124,21 @@ do {
 } while (true);
 ```
 
-#### `'neverExceptWhileTrue'`
+#### `'only-true'`
 
 Disallow constant conditions in all loops except `while` loops with expression `true`.
 
-Example of correct code for `{ allowConstantLoopConditions: 'neverExceptWhileTrue' }`:
+Example of correct code for `{ allowConstantLoopConditions: 'only-true' }`:
 
-```ts option='{ "allowConstantLoopConditions": "neverExceptWhileTrue" }' showPlaygroundButton
+```ts option='{ "allowConstantLoopConditions": "only-true" }' showPlaygroundButton
 while (true) {
   // ...
 }
 ```
 
-Example of incorrect code for `{ allowConstantLoopConditions: 'neverExceptWhileTrue' }`:
+Example of incorrect code for `{ allowConstantLoopConditions: 'only-true' }`:
 
-```ts option='{ "allowConstantLoopConditions": "neverExceptWhileTrue" }' showPlaygroundButton
+```ts option='{ "allowConstantLoopConditions": "only-true" }' showPlaygroundButton
 for (; true; ) {
   // ...
 }
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
@@ -141,13 +141,23 @@ while (true) {
 Example of incorrect code for `{ allowConstantLoopConditions: 'only-allowed-literals' }`:
 
 ```ts option='{ "allowConstantLoopConditions": "only-allowed-literals" }' showPlaygroundButton
-for (; true; ) {
+declare const alwaysTrue: true;
+
+// `alwaysTrue` has the type `true` (which isn't allowed)
+// as only the literal value of `true` is allowed
+
+while (alwaysTrue) {
   // ...
 }
 
-do {
+// same for booleans as only the literal value of `true` or `false`
+// are allowed
+
+declare const bool: boolean;
+
+while (bool) {
   // ...
-} while (true);
+}
 ```
 
 ### `checkTypePredicates`
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
@@ -124,21 +124,23 @@ do {
 } while (true);
 ```
 
-#### `'only-true'`
+#### `'only-allowed-literals'`
 
-Disallow constant conditions in all loops except `while` loops with expression `true`.
+Disallow constant conditions in all loops except `while` loops with certain literal expressions.
 
-Example of correct code for `{ allowConstantLoopConditions: 'only-true' }`:
+Specifically, `true`, `false`, `0`, and `1` are allowed despite always being truthy or falsy, as these are common patterns and are not likely to represent developer errors.
 
-```ts option='{ "allowConstantLoopConditions": "only-true" }' showPlaygroundButton
+Example of correct code for `{ allowConstantLoopConditions: 'only-allowed-literals' }`:
+
+```ts option='{ "allowConstantLoopConditions": "only-allowed-literals" }' showPlaygroundButton
 while (true) {
   // ...
 }
 ```
 
-Example of incorrect code for `{ allowConstantLoopConditions: 'only-true' }`:
+Example of incorrect code for `{ allowConstantLoopConditions: 'only-allowed-literals' }`:
 
-```ts option='{ "allowConstantLoopConditions": "only-true" }' showPlaygroundButton
+```ts option='{ "allowConstantLoopConditions": "only-allowed-literals" }' showPlaygroundButton
 for (; true; ) {
   // ...
 }
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
@@ -84,9 +84,9 @@ function bar(arg?: string | null) {
 
 {/* insert option description */}
 
-#### `'never'` or `false`
+#### `'never'`
 
-Disallow constant conditions in loops.
+Disallow constant conditions in loops. Same as `false`.
 
 Example of incorrect code for `{ allowConstantLoopConditions: 'never' }`:
 
@@ -104,9 +104,9 @@ do {
 } while (true);
 ```
 
-#### `'always'` or `true`
+#### `'always'`
 
-Allow constant conditions in loops.
+Allow constant conditions in loops. Same as `true`.
 
 Example of correct code for `{ allowConstantLoopConditions: 'always' }`:
 
@@ -141,21 +141,21 @@ while (true) {
 Example of incorrect code for `{ allowConstantLoopConditions: 'only-allowed-literals' }`:
 
 ```ts option='{ "allowConstantLoopConditions": "only-allowed-literals" }' showPlaygroundButton
-declare const alwaysTrue: true;
+// `alwaysTrue` has the type of `true` (which isn't allowed)
+// as only the literal value of `true` is allowed.
 
-// `alwaysTrue` has the type `true` (which isn't allowed)
-// as only the literal value of `true` is allowed
+declare const alwaysTrue: true;
 
 while (alwaysTrue) {
   // ...
 }
 
-// same for booleans as only the literal value of `true` or `false`
-// are allowed
+// not even a variable that references the value of `true` is allowed, only
+// the literal value of `true` used directly.
 
-declare const bool: boolean;
+const thisIsTrue = true;
 
-while (bool) {
+while (thisIsTrue) {
   // ...
 }
 ```
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
@@ -126,7 +126,7 @@ do {
 
 #### `'only-allowed-literals'`
 
-Disallow constant conditions in all loops except `while` loops with certain literal expressions.
+Permit idiomatic constant literal conditions in `while` loop conditions.
 
 Specifically, `true`, `false`, `0`, and `1` are allowed despite always being truthy or falsy, as these are common patterns and are not likely to represent developer errors.
 
--- a/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
@@ -84,12 +84,68 @@ function bar(arg?: string | null) {
 
 {/* insert option description */}
 
-Example of correct code for `{ allowConstantLoopConditions: true }`:
+#### `'never'` or `false`
 
-```ts option='{ "allowConstantLoopConditions": true }' showPlaygroundButton
-while (true) {}
-for (; true; ) {}
-do {} while (true);
+Disallow constant conditions in loops.
+
+Example of incorrect code for `{ allowConstantLoopConditions: 'never' }`:
+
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
```diff
@@ -84,12 +84,68 @@ function bar(arg?: string | null) {
...
+for (; true; ) {
+  // ...
+}
+
+do {
+  // ...
+} while (true);
+```
+
+#### `'neverExceptWhileTrue'`
```
JoshuaKGoldberg, line 127: [Naming] Maybe... `onlyTrue` / `'only-true'` per https://github.com/typescript-eslint/typescript-eslint/issues/7047#issuecomment-1902126574?

I don't remember if there's every been a discussion on preferring camelCase vs. kebab-case for property values like this. 🤷

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
JoshuaKGoldberg, line 1: [Non-Actionable] `while (1)` is a thing people sometimes do too. 
* [Sourcegraph search for `"while (1)"`](https://sourcegraph.com/search?q=context:global+lang:typescript+-path:node_modules+%22while+%281%29%22&patternType=keyword&sm=0): 645+ results
* [Sourcegraph search for `"while (true)"`](https://sourcegraph.com/search?q=context:global+lang:typescript+-path:node_modules+%22while+%28true%29%22+count:110000&patternType=keyword&sm=0): some number over 21.8k+ results

...so I'm ok with not doing anything. It does feel weird to use `1` rather than `true` for _"truthiness"_, near-pun intended. 👍 

Just posting for visibility and in case someone wants to discuss...

JoshuaKGoldberg (changes requested):
🚀 Whoohoo! Thanks for taking on this old issue, it's been bugging me for ages!

I agree with the implementation and direction. Keeping the legacy options around to make this non-breaking seems reasonable + necessary to me.

Just requesting changes on the name, since `onlyTrue` is a lot cleaner IMO.

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
kirkwaiblinger: TS does special-case `while (1)`/`while (0)` to be allowed despite being always-truthy/always-falsy:
([playground](https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true&target=99&ts=5.7.3#code/PTAEEMBtIewdwKYBNQGcCWA7Axg06AXAclVACJ0l0YBbcA9bMgKDgAt1I8AKARgEpQAb2YBfZsxCgEAJxkwZoALQRIccAE9SBGQFcCbDaw5dQfAHQBWQSNFA)).

So there'd be precedent for allowing `1` in this rule if we wanted. I don't like it, but there's an argument that it's more of a strict-boolean-expressions concern to ban than a no-unnecessary-condition concern. `while (1)` is also relatively idiomatic in C-like language contexts (where `bool` `false`/`true` and `int` `0`/`1` are semi-interchangeable) so that's probably in part where the JS usage comes from.

Neutral on whether this is supported.

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
ronami: Hmm yes, I agree that this is a lot cleaner 👍

I've done a quick search, and it seems all `enum` options are kebab-case, so I'm going with `only-true` to match this, thanks!

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
ronami: It's interesting that TypeScript special-cased `while(1)`.

It might make sense for `only-true` to also allow this, as it's essentially a different way to write `while(true)`. I'm also neutral on whether this should be supported.

JoshuaKGoldberg: 🌶️ I say let's just go ahead and do it then. It's really not much more difficult to make the option `allowLiterals` and just allow literals.

ronami: Looking some more into it, it seems to have been added [pretty recently in TypeScript 5.6](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#disallowed-nullish-and-truthy-checks). Quoting from the TypeScript 5.6 announcement (https://github.com/microsoft/TypeScript/pull/59217):

> Note that certain expressions are still allowed, even if they are always truthy or nullish. Specifically, true, false, 0, and 1 are all still allowed despite always being truthy or falsy, since code like the following:
> 
> ```ts
> while (true) {
>     doStuff();
> 
>     if (something()) {
>         break;
>     }
> 
>     doOtherStuff();
> }
> ```
> 
> is still idiomatic and useful, and code like the following:
> 
> ```ts
> if (true || inDebuggingOrDevelopmentEnvironment()) {
>     // ...
> }
> ```
> 
> is useful while iterating/debugging code.

I'm not sure if we should special case `1` and `true`, or allow literals in general. If all literals are allowed, projects on older versions of TypeScript may not see a warning at all (from TypeScript or typescript-eslint).

JoshuaKGoldberg: The cross-section of users on older TypeScript versions + the latest versions of typescript-eslint is pretty small. It'll only get smaller over time. I don't think it's a big problem if we have a mild overlap / non-overlap in this way.

Ideating: maybe the config can be `'true' | 'typescript'`, to make TypeScript's behavior opt-in? Or `(boolean | number)[]` so folks can allow a specific list of literals?

🤷 I'm still not passionate about it, and +0.5 on making it support TypeScript's by default. I don't think this is going to be a hot button issue for users. If TypeScript has decided that `0` and `1` can also be constant conditions then it would take up docs space & our energy to explain why we differ.

JoshuaKGoldberg: Thought on it a bit - I think I'm actually +1 on supporting TypeScript's literals now. For the reasons we've mentioned. 💪 

> options' name

I like `'only-allowed-literals'`. There's no real existing term to base it off of and referring to _"the ones TypeScript allows"_ would be kind of vague too. Plus this leaves it open to extension in the future if someone really wants to pass an array in.

ronami: Yes! I agree, thanks 👍

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
```diff
@@ -84,12 +84,70 @@ function bar(arg?: string | null) {
...
+Example of incorrect code for `{ allowConstantLoopConditions: 'only-allowed-literals' }`:
+
+```ts option='{ "allowConstantLoopConditions": "only-allowed-literals" }' showPlaygroundButton
+for (; true; ) {
+  // ...
+}
+
+do {
+  // ...
+} while (true);
```
kirkwaiblinger, line 150:
```ts
do {
  // ...
} while (false);
```
is actually probably a useful example to demonstrate why we'd allow the falsy literals. @Josh-Cena would know the real world use cases, but I think it's basically a pattern similar to an IIFE to define a scope (with `break` available for control flow), but for code that is meant to be side-effect-ful and _not_ return a value.

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
```diff
@@ -84,12 +84,70 @@ function bar(arg?: string | null) {
...
+```
+
+#### `'only-allowed-literals'`
+
+Disallow constant conditions in all loops except `while` loops with certain literal expressions.
```
kirkwaiblinger, line 129: I'm confused by this sentence (double-negative).... Can it just be 
```text
Permit idiomatic constant literal conditions in `while`/`do...while` loop conditions.
```
?

ronami: Yes! That's a lot better 👍

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
ronami: +1, but the rule currently allows literals on a standard `while(false)` loop but not on other types of loops like `do {} while(false)` or `for (;false;) {}` ([discussed here](https://github.com/typescript-eslint/typescript-eslint/pull/10639#discussion_r1925801773)).

I'm neutral on whether this should be expanded to support other types of loops, but as long as only `while()` is supported, I don't think there's an actual use case for `while(false)` outside of debugging.

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
```diff
@@ -84,12 +84,70 @@ function bar(arg?: string | null) {
 
 {/* insert option description */}
 
-Example of correct code for `{ allowConstantLoopConditions: true }`:
+#### `'never'` or `false`
```
kirkwaiblinger, line 87: I'm wondering if we can de-emphasize the `true`/`false` a bit to nudge users towards the string options now... As a random example from the top of my head, the [vscode docs about code actions on save](https://code.visualstudio.com/Docs/languages/javascript#_code-actions-on-save) just give the string name in the header and mention the legacy boolean alias as an afterthought. Probably there are many more examples to draw from in the ecosystem.

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
kirkwaiblinger: Hmm, I see... Well, granted that the option only applies to `while()` loop conditions for now, what would you think of these ideas?
1. Remove the `for`/`do...while` examples from the "incorrect" section. This gives us some more wiggle room to change our minds in the future about whether they should be included by the option.
2. Add (a variant of) the following snippet to the incorrect section, to emphasize that that the option only applies to honest-to-god literals (not literal types, even):
   ```ts
   declare const TRUE: true;

   while (TRUE) {
     // ...
   }
   ```

ronami: Yes for both! Thank you 👍

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
ronami: Interesting! I like this, especially since this will probably (I assume) be removed or deprecated in the next major.

The closet I could find one `typescript-eslint`'s repo is possibly [`TypeOrValueSpecifier`'s universal string specifiers](https://typescript-eslint.io/packages/type-utils/type-or-value-specifier/#universal-string-specifiers) (which isn't too similar). I guess that another option is to stick with `true`, `false`, and `only-allowed-literals` (I've seen this [done here](https://rollupjs.org/configuration-options/#treeshake-propertyreadsideeffects) and [also here](https://vite.dev/config/shared-options.html#json-stringify) for example).

I updated the PR to remove this from the header and down to the description (similar to the linked vscode example).

---

Another option I'm thinking of is the order of these three options: currently `'never'` shows up first, followed by `'always'`, with `'only-allowed-literals'` being last (even though it's probably [I'm assuming] the recommended way to go).

Should it potentially be the first option to promote users using it, at least until it's changed to the default?

kirkwaiblinger (approved):

packages/eslint-plugin/docs/rules/no-unnecessary-condition.mdx
kirkwaiblinger: > Interesting! I like this, especially since this will probably (I assume) be removed or deprecated in the next major.
> 
> The closet I could find one `typescript-eslint`'s repo is possibly [`TypeOrValueSpecifier`'s universal string specifiers](https://typescript-eslint.io/packages/type-utils/type-or-value-specifier/#universal-string-specifiers) (which isn't too similar). If these options would be removed or deprecated, should this have a note on the docs to warn about this potentially?

Eh, I don't think there's any value in adding a pre-deprecated state 🙃 I personally would be happy for the `true`/`false` option types to be deprecated in this PR but maybe others will have opinions. 

> I guess that another option is to stick with `true`, `false`, and `only-allowed-literals` (I've seen this [done here](https://rollupjs.org/configuration-options/#treeshake-propertyreadsideeffects) and [also here](https://vite.dev/config/shared-options.html#json-stringify) for example).

My reaction to this is 🤮, haha

> I updated the PR to remove this from the header and down to the description (similar to the linked vscode example).
> 
> Another option I'm thinking of is the order of these three options: currently `'never'` shows up first, followed by `'always'`, with `'only-allowed-literals'` being last (even though it's probably [I'm assuming] the recommended way to go).
> 
> Should it potentially be the first option to promote users using it, at least until it's changed to the default?

Hmm. I don't have a strong opinion. Probably the default first is best and we can change the order if we change the default in the future? Whatever you think is best!

JoshuaKGoldberg (approved):
🚀 Very nice how this ended up!

PR #10640, opened 2025-01-12
feat(ast-spec): support import attributes in `TSImportType`

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10394
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

This PR addresses #10394 and adds import attributes to the `TSImportType` node.

---

Some thoughts:

I've named this `attributes`, similar to other nodes that use `ImportAttribute[]` ([1](https://github.com/typescript-eslint/typescript-eslint/blob/81dc579e01c6e95bd347725eab4d9a26a5fc724d/packages/ast-spec/src/declaration/ExportNamedDeclaration/spec.ts#L32), [2](https://github.com/typescript-eslint/typescript-eslint/blob/81dc579e01c6e95bd347725eab4d9a26a5fc724d/packages/ast-spec/src/declaration/ImportDeclaration/spec.ts#L26)). This does mean the AST would be a bit different than Babel's, [which uses the name `options`](https://github.com/babel/babel/blob/5d44d275d2b0ea8f2a176315c4ec4d402f894bf8/packages/babel-parser/src/types.ts#L1771). Please let me know if this should be changed.

I didn't include any adjustments to the `scope-manager`, as to my understanding, `ImportAttribute`s can't include any reference to variables (also other nodes don't currently do this).

I also didn't include any changes to [`visitor-keys`](https://github.com/typescript-eslint/typescript-eslint/blob/81dc579e01c6e95bd347725eab4d9a26a5fc724d/packages/visitor-keys/src/visitor-keys.ts#L162-L269) since I didn't see other nodes with `ImportAttribute`s do this.

Since this is currently auto-fixed incorrectly by Prettier (https://github.com/prettier/prettier/issues/16072), I've added the fixture to `.prettierignore`. I've seen this done in similar cases ([see here](https://github.com/typescript-eslint/typescript-eslint/blob/81dc579e01c6e95bd347725eab4d9a26a5fc724d/.prettierignore#L16-L25)).

PR #10649, opened 2025-01-14
fix(visitor-keys): update keys for  `ImportAttribute`

antfu:
## PR Checklist

## Overview

Sorry for a quick PR without an issue - I will fill that later. The key `assertions` on `ImportDeclaration` as renamed to `attributes` but the visitor keys are not updated. This causing the `ImportAttribute` listener in ESLint rule to be called twice with the same node because eslint's visitor keys now include `attributes`:
https://github.com/eslint/js/blob/79f6f9a603e44707a7774ebd3f871da336242fb5/packages/eslint-visitor-keys/lib/visitor-keys.js#L148

Note: I noticed that the keys are actually merged with `eslint-visitor-keys`, as `attributes` is already in `eslint-visitor-keys`, should I remove them instead? If so, there seem to be quite a few keys that are also redundant. -> I made #10650 for the record


JoshuaKGoldberg:
cc @fisker @ronami from #10640 -> #10691: is there more work to be done in #10691?

JoshuaKGoldberg (approved):
Sorry for the delay! I just want to be thorough and do research on this, since it's a tricky change. It looks good to me. Just to be safe I think another review from @typescript-eslint/triage-team would be good.

bradzacher (approved):

PR #10650, opened 2025-01-14
test(visitor-keys): add test to guard redundant visitor keys

antfu:
## PR Checklist

## Overview

From #10649 I found there are quite some redundant visitor keys that mostly might due to the updates in `eslint-vistior-keys`. In order to locate them, I write a simple test to guard them - not sure if you consider them valid testing. With the tests, in this PR, I also removed the keys that are considered redundant.

Feel free to close this PR if it's not desired

### DIFF
--- a/packages/visitor-keys/src/visitor-keys.ts
+++ b/packages/visitor-keys/src/visitor-keys.ts
@@ -132,13 +132,9 @@ type AdditionalKeys = {
  */
 
 const SharedVisitorKeys = (() => {
-  const FunctionType = ['typeParameters', 'params', 'returnType'] as const;
-  const AnonymousFunction = [...FunctionType, 'body'] as const;
-  const AbstractPropertyDefinition = [
-    'decorators',
-    'key',
-    'typeAnnotation',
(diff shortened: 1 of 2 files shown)


bradzacher (changes requested):
the keys aren't redundant!
`unionWith` will preserve our ordering for the keys.
this is VERY important because the ordering defines the order in which the properties are visited. An incorrect order can cause incorrect rule behaviour or incorrect scope analysis because the code is not visited in source order.

```ts
// say this is what eslint-visitor-keys exports
const KEYS = { a: ['a'] };

unionWith({}) == KEYS
unionWith({ a: [] }) == KEYS
unionWith({ a: ['b'] }) == { a: ['b', 'a'] }
unionWith({ a: ['b', 'a', 'c'] }) == { a: ['b', 'a', 'c'] }
```

By removing the keys you will cause `unionWith` to append missing keys to the end of the array -- causing a broken visit order!
This is why the tests are failing.

antfu:
I see. Thanks for pointing out!

PR #10670, opened 2025-01-17
fix(eslint-plugin): [no-deprecated] should allow ignoring of deprecated value

### LINKED ISSUE #10660: Bug: [no-deprecated] should allow ignoring of deprecated value

### Playground Link

https://main--typescript-eslint.netlify.app/play/#ts=5.3.3&fileType=.tsx&code=PQKhAIAEBMFMAcBOsDGBDALra4TAFAoD2AdgM4bhxKqbYBqaANgK6zgC84ALAEwDchUhXAAjNIk5UEydFmiNWsfkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNFjpEZAIb5E3dGADa4bDgDuQ6E0gAaGbKyzskIfHgc5E6Vq2aTsyADNoHALYSrlJKvXmcTIbcQPe-QSLEANV1Yb1cTAF9w7ABdaKjZOIgoiKA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
/** @deprecated */
const deprecatedValue = 42;
const bar = deprecatedValue;
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-deprecated": [
      "warn",
      {
        "allow": [
          {
            "from": "file",
            "name": "deprecatedValue"
          }
        ]
      }
    ]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

No linter report on `deprecatedValue` since it's configured to be allowed

### Actual Result

Linter report on `deprecatedValue`

### Additional Info

This is basically just that TypeOrValueSpecifier is only used for Types right now. Relates to https://github.com/typescript-eslint/typescript-eslint/pull/10585#pullrequestreview-2540761108

y-hsgw:
## PR Checklist

- [ ] Addresses an existing open issue: fixes #10660

### DIFF
--- a/packages/type-utils/src/TypeOrValueSpecifier.ts
+++ b/packages/type-utils/src/TypeOrValueSpecifier.ts
@@ -1,6 +1,8 @@
+import type { TSESTree } from '@typescript-eslint/types';
 import type { JSONSchema4 } from '@typescript-eslint/utils/json-schema';
 import type * as ts from 'typescript';
 
+import { AST_NODE_TYPES } from '@typescript-eslint/types';
 import * as tsutils from 'ts-api-utils';
 
 import { specifierNameMatches } from './typeOrValueSpecifiers/specifierNameMatches';
@@ -202,3 +204,24 @@ export const typeMatchesSomeSpecifier = (
   program: ts.Program,
 ): boolean =>
   specifiers.some(specifier => typeMatchesSpecifier(type, specifier, program));
+
+export function valueMatchesSpecifier(
+  node: TSESTree.Node,
+  specifier: TypeOrValueSpecifier,
+): boolean {
+  if (node.type !== AST_NODE_TYPES.Identifier) {
+    return false;
+  }
+
+  if (typeof specifier === 'string') {
+    return node.name === specifier;
+  }
+
+  return node.name === specifier.name;
+}
+
+export const valueMatchesSomeSpecifier = (
+  node: TSESTree.Node,
+  specifiers: TypeOrValueSpecifier[] = [],
+): boolean =>
+  specifiers.some(specifier => valueMatchesSpecifier(node, specifier));
--- a/packages/eslint-plugin/src/rules/no-deprecated.ts
+++ b/packages/eslint-plugin/src/rules/no-deprecated.ts
@@ -12,6 +12,7 @@ import {
   nullThrows,
   typeOrValueSpecifiersSchema,
   typeMatchesSomeSpecifier,
+  valueMatchesSomeSpecifier,
 } from '../util';
 
 type IdentifierLike =
@@ -372,7 +373,10 @@ export default createRule<Options, MessageIds>({
       }
 
       const type = services.getTypeAtLocation(node);
-      if (typeMatchesSomeSpecifier(type, allow, services.program)) {
+      if (
+        typeMatchesSomeSpecifier(type, allow, services.program) ||
+        valueMatchesSomeSpecifier(node, allow)
+      ) {
         return;
       }
 
--- a/packages/eslint-plugin/tests/rules/no-deprecated.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-deprecated.test.ts
@@ -327,6 +327,30 @@ new A();
     },
     {
       code: `
+/** @deprecated */
+const deprecatedValue = 45;
+const bar = deprecatedValue;
+      `,
+      options: [
+        {
+          allow: [{ from: 'file', name: 'deprecatedValue' }],
+        },
+      ],
+    },
+    {
+      code: `
+/** @deprecated */
+const deprecatedValue = 45;
+const bar = deprecatedValue;
+      `,
+      options: [
+        {
+          allow: ['deprecatedValue'],
+        },
+      ],
+    },
+    {
+      code: `
 import { exists } from 'fs';
 exists('/foo');
       `,


JoshuaKGoldberg (dismissed):
✨

JoshuaKGoldberg (changes requested):
Thanks for sending this in!

Codecov is right to report, we have tests set up in `packages/type-utils` to make sure these utils are fully unit tested. Could you please add tests to `TypeOrValueSpecifier.test.ts`?

All three of the specifier sources formats should be tested: file, lib, and package.

y-hsgw:
I've added the tests!
is the behavior of valueMatchesSpecifier as expected when specifier.from is set to package?

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -202,3 +203,28 @@ export const typeMatchesSomeSpecifier = (
   program: ts.Program,
 ): boolean =>
   specifiers.some(specifier => typeMatchesSpecifier(type, specifier, program));
+
+export function valueMatchesSpecifier(
+  node: TSESTree.Node,
+  specifier: TypeOrValueSpecifier,
+): boolean {
+  if ('name' in node && typeof node.name === 'string') {
```
JoshuaKGoldberg, line 211: [Refactor] We don't generally like to use `in` checks for AST node narrowing. https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/no-property-in-node.md

I see from `no-deprecated` that this is only ever called with a node that is one of:

* `TSESTree.Identifier`
* `TSESTree.JSXIdentifier`
* `TSESTree.Super`

`TSESTree.Super` is always `false`. The other Identifiers have `name: string`. So you should be able to refactor this to restrict the type of `node` and not have to check` typeof node.name`. Or something along those lines. 😄

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -202,3 +203,28 @@ export const typeMatchesSomeSpecifier = (
...
+    if (typeof specifier === 'string') {
+      return node.name === specifier;
+    }
+
+    if (typeof specifier.name === 'string') {
+      return node.name === specifier.name;
+    }
+
+    return specifier.name.includes(node.name);
```
JoshuaKGoldberg, line 220: [Bug] This just checks the specifier name. But it doesn't check where the specifier comes from. This'll need to more closely mirror how `typeMatchesSpecifier` switches on `specifier.from`. Instead of just checking the `specifier.name`, the place the specifier is declared needs to be checked too.

You might need to use the scope manager for this, to see where the value actually comes from locally.

JoshuaKGoldberg (changes requested):
> Do you think it might be necessary to verify whether it is imported from a package?

Indeed I do - the goal of the shared format is to match on both the name _and_ where a type/value is declared.

Thanks!

packages/type-utils/src/TypeOrValueSpecifier.ts
y-hsgw: Thanks! I fixed in [a4463a4](https://github.com/typescript-eslint/typescript-eslint/pull/10670/commits/a4463a4dae6c2c5c619d8ee63db0ac858c75ef36).

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -219,3 +222,46 @@ export const typeMatchesSomeSpecifier = (
...
   specifiers.some(specifier => typeMatchesSpecifier(type, specifier, program));
+
+export function valueMatchesSpecifier(
+  node: TSESTree.Node,
+  specifier: TypeOrValueSpecifier,
+  scopeManager: ScopeManager,
+): boolean {
+  if (
+    node.type === AST_NODE_TYPES.Identifier ||
+    node.type === AST_NODE_TYPES.JSXIdentifier
```
JoshuaKGoldberg, line 233: [Bug] This works for the two specified types of identifiers, but there are other things out there. Private properties, computed keys based on string literals, etc. - which should be tested.

I think you'll want to use `getStaticValue` to get the static value. `getStaticMemberAccessValue` in `eslint-plugin` might be useful as a reference as well, not sure.

Btw, sorry, I realize I suggested these two checks in https://github.com/typescript-eslint/typescript-eslint/pull/10670#discussion_r1967969382 and now am kind of going against that. If this utility were only used internally and by `no-deprecated` that'd be fine. But I'm remembering now that this is also exported publicly under `@typescript-eslint/type-utils`. We'll need it to be more generalized.

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -219,3 +222,46 @@ export const typeMatchesSomeSpecifier = (
...
+  ) {
+    if (typeof specifier === 'string') {
+      return node.name === specifier;
+    }
+
+    if (specifier.from === 'package') {
+      const variable = scopeManager.variables.find(v => v.name === node.name);
+      const targetNode = variable?.defs[0].parent;
+      if (targetNode?.type !== AST_NODE_TYPES.ImportDeclaration) {
```
JoshuaKGoldberg, line 242: [Bug] What if the variable comes from a dynamic package import?

```ts
declare const fs: typeof import("fs");

fs.exists;
// ~~~~~~
```

Using the scope manager to get to the variable's local definition is one step. You might be able to get this to work with more AST+scope work. Or maybe it'd be easier to use TypeScript types and/or symbols? Or some combination of them? YMMV.

JoshuaKGoldberg (changes requested):
A bit more complication! 🚀 

I'm also starting to wonder - maybe it makes sense to unify the `typeMatchesSpecifier` and `valueMatchesSpecifier` logic? If the value checking needs type information anyway...

packages/type-utils/src/TypeOrValueSpecifier.ts
y-hsgw: Thank you for the feedback!
I think Private properties could be supported by 3b3520b.

However, I don't quite understand the support for computed keys. Could you please provide more information?

My understanding is that for cases where `AST_NODE_TYPES.Property` is passed as the first argument to `valueMatchesSpecifier`, and the key is a computed key, we should handle it in that context. Is that correct?
（I was under the impression that an `AST_NODE_TYPES.Identifier` within `AST_NODE_TYPES.Property` would be passed as an argument...）

If possible, it would be very helpful if you could provide some concrete test cases or examples to guide the implementation.🙇‍♂️
I would appreciate it if you could teach me for my learning. (Also for the future!😂)

JoshuaKGoldberg: I appreciate the questions! Here's a test case I was thinking of:

```ts
[
  `
  class MyClass {
    ['computed prop'] = 42;
    value = this['computed prop'];
  }`,
  `'computed prop'`,
],
```

...but now I'm not so sure. We don't have existing code paths for this, and I'm not sure how we'd want to handle differences in quotes (`'computed prop'` vs. `"computed prop"`). cc @typescript-eslint/triage-team - for now I think it's fine to avoid those types of keys, and treat them as a followup issue?

kirkwaiblinger: I think handling the computed properties quotations' on the code side isn't scary due to `node.value.cooked`, but yeah it requires a syntax in the selector. If `privateProp` matches `this.#privateProp` then maybe `multi word prop` should match `this['multi word prop']` without any extra syntax? Similarly that would sort of imply that `aProp` would match `this['aProp']` in addition to `this.aProp`... which seems reasonable?

y-hsgw: Thank you for the advice.
I’ve added support for the literal case in [0e61bcf](https://github.com/...) — I’d appreciate it if you could take a look!

packages/type-utils/src/TypeOrValueSpecifier.ts
y-hsgw: I’ve added support for the dynamic import case as well.[1578137](https://github.com/typescript-eslint/typescript-eslint/pull/10670/commits/15781373b072fcfc641fe28abd7628ea4676a31f)

packages/type-utils/src/TypeOrValueSpecifier.ts
```diff
@@ -219,3 +221,69 @@ export const typeMatchesSomeSpecifier = (
   program: ts.Program,
 ): boolean =>
   specifiers.some(specifier => typeMatchesSpecifier(type, specifier, program));
+
+const getSpecifierNames = (specifierName: string | string[]): string[] => {
+  return typeof specifierName === 'string' ? [specifierName] : specifierName;
+};
+
+const getStaticName = (node: TSESTree.Node): string | undefined => {
```
JoshuaKGoldberg, line 229: [Non-Actionable] I'm surprised we don't already have a function like this somewhere 😄. I spent a bit searching and couldn't find one. Ah well.

JoshuaKGoldberg, line 229: [Non-Actionable] I'm surprised we don't already have a function like this somewhere 😄. I spent a bit searching and couldn't find one. The closest was `getNameFromMember` over in `eslint-plugin` but it's not quite the same. Ah well.

JoshuaKGoldberg:
The CI unit testing failure is unrelated flake.

PR #10631, opened 2025-01-07
feat(eslint-plugin): [no-unnecessary-type-assertion] improve reporting of literal types (POC)

ronami:
## PR Checklist

- [ ] Addresses an existing open issue: fixes #000
- [ ] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [ ] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

The `no-unnecessary-type-assertion` rule has custom logic for handling literal types: https://github.com/typescript-eslint/typescript-eslint/blob/328b7dfe255b2a9c8226db472b21d070f19193af/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts#L227-L229

To my understanding, the current logic is as follows: If the cast type is a literal type, it would only be reported if it's the initializer of a `const` declaration. If it isn't a literal type, it would be reported as long as the assertion isn't `as const` (along with the asserted type and expression having the same type).

Zooming into the logic of handling literals, I think the rule tries to predict when a literal would be widened. To my understanding (quoting [Widening and Narrowing in Typescript](https://shively-sanders.com/manual/Widening-and-Narrowing-in-Typescript.html)), a literal would be widened in the following cases:

> - Literal widening only happens to literal types that originate from expressions. These are called fresh literal types.
> - Literal widening happens whenever a fresh literal type reaches a "mutable" location.

To my understanding, the rule somewhat handles (2) as it only reports literals that are assigned to a `const` variable declaration but ignores (1). (2) is still somewhat ignored, as "fresh" literals are relevant here too.

In my opinion, this logic may be improved so the rule can catch additional cases that it currently misses (and not fail on some false positives), mostly around "fresh literals".

I'm basing this on the [TypeScript Handbook](https://github.com/microsoft/TypeScript-New-Handbook/blob/master/reference/Widening-and-Narrowing.md#fresh-literal-types-and-mutable-locations), [Widening and Narrowing in Typescript](https://shively-sanders.com/manual/Widening-and-Narrowing-in-Typescript.html) and [the literal widening PR](https://github.com/Microsoft/TypeScript/pull/10676), along with testing this on the playground. See this comment https://github.com/typescript-eslint/typescript-eslint/issues/1056#issuecomment-2562979820, as I think they're somewhat related.

This isn't something I have a thorough understanding of, but it seems like the case with the following simplified example:

```ts
[...]

PR #10618, opened 2025-01-06
fix(eslint-plugin): [no-unnecessary-type-assertion] should report readonly class properties with a literal initializer 

y-hsgw:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10617
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Fixed the bug described in the title.

🐛

Issue #8737, opened 2024-03-20
[@typescript-eslint/no-unnecessary-type-assertion] False positives for "as const" assertions with template literals


klassm:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.2&fileType=.tsx&code=KYDwDg9gTgLgBAYwgOwM7zFCAjANsAWwClUUAVATzGAFUoBLABSmADN6Q4BeOAcgAsYMMKgBcAenHAoWKKgB0-CAWABaJMmTAEMeUgLjecAIapEKdAG4AsACgN6ONNkAhU7Qbc4AAwAkAb0wcfGJSZEpqOiYWdhAAXw0YYGQYVGkAN3oEYFRvEzMHGBtbO1BIWDgYKmA4ZmDCEhQvACVtaAATAB50BmQAcwAaOABXZABrZAgAd2QAPjgAMjh-OzhK6tEfAKrqCFY4ILwGsIiPaLYOOICe%2Bn648WuYXr6472K44tLwaHhCuAA1Yy4ejtYwwegoACiMmgmzqR1CTR4K1sax2wE2fn8zmgbjSUXu6SBILBEOQqhxUG8AzsHzsQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&tsconfig=N4KAvkA&tokens=false

### Repro Code

```TypeScript
export const problemJsonTypeUriPrefix = 'https://errors.home-connect.com/';
const errorBaseUri = `${problemJsonTypeUriPrefix}contentservices` as const;

export type ProblemJson = Record<string, unknown> & {
  type: `${typeof problemJsonTypeUriPrefix}${string}/${string}`;
};

export const ValidationError: ProblemJson = {
  type: `${errorBaseUri}/validation-error`,
[...]

PR #10671, opened 2025-01-17
test(eslint-plugin): adjust tests to verify `no-unnecessary-type-assertion` doesn't report template literals with expressions

ronami:
## PR Checklist

## Overview

I've noticed this while reviewing https://github.com/typescript-eslint/typescript-eslint/pull/10618.

To my understanding, these tests should make sure the rule [doesn't report on template literals with expressions](https://github.com/typescript-eslint/typescript-eslint/blob/31be0532065491d9141cbddfe42dd771eadc0a82/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts#L171-L175).

It seems that even if this check is removed, these tests still pass. I think it's related to "fresh literals" ([more on this here](https://github.com/typescript-eslint/typescript-eslint/pull/10631)), as is shown below ([playground link](https://typescript-eslint.io/play/#ts=5.7.3&fileType=.ts&code=MYewdgzgLgBAZiEMC8MDkCRoNwChcD0BMUAFgKYwAGFANrUgCQDemAvlTOQB4AOATuQgQAluBgATEEJigArmFgBDCDCUxaIqOX5LauUJFjboARhTU6DGC3acVs8NDxEYb9wD0A-PlrlYAEZK-BZoQfw4%2BK5klDTk9EzM4RxcfILCYmCS0hBgaLDyimqq6prauvquDlpoqiJgcDqCEsWW8dYs0Pz1AOYcBk7GQlAATBZxCTZJwSkOhs6ExO5u3rhAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```ts
const foo = 'foo';

// the `hello ${foo}` expression does count as a literal
const test1 = `hello ${foo}` as const;
//      ^?

let bar = 'bar';

// the `hello ${bar}` expression doesn't count as a literal
// as it's inferred as `hello ${string}`
const test2 = `hello ${bar}` as const;
//      ^?
```

The expression coming from a widened `let` variable [doesn't count as a literal](https://github.com/typescript-eslint/typescript-eslint/blob/31be0532065491d9141cbddfe42dd771eadc0a82/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts#L227), and it falls back to [being checked as a non-literal](https://github.com/typescript-eslint/typescript-eslint/blob/31be0532065491d9141cbddfe42dd771eadc0a82/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts#L229) (which always allows `as const` assertions).

This PR adjusts the tests to fail if this check is removed or changed incorrectly (this seems correct [according to the original issue](https://github.com/typescript-eslint/typescript-eslint/issues/8737)).

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
@@ -294,12 +294,12 @@ function bar(items: string[]) {
     },
     // https://github.com/typescript-eslint/typescript-eslint/issues/8737
     `
-let myString = 'foo';
+const myString = 'foo';
 const templateLiteral = \`\${myString}-somethingElse\` as const;
     `,
     // https://github.com/typescript-eslint/typescript-eslint/issues/8737
     `
-let myString = 'foo';
+const myString = 'foo';
 const templateLiteral = <const>\`\${myString}-somethingElse\`;
     `,
     'let a = `a` as const;',
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
@@ -294,13 +294,17 @@ function bar(items: string[]) {
     },
     // https://github.com/typescript-eslint/typescript-eslint/issues/8737
     `
-const myString = 'foo';
+declare const myString: 'foo';
 const templateLiteral = \`\${myString}-somethingElse\` as const;
     `,
     // https://github.com/typescript-eslint/typescript-eslint/issues/8737
     `
-const myString = 'foo';
+declare const myString: 'foo';
 const templateLiteral = <const>\`\${myString}-somethingElse\`;
+    `,
+    `
+const myString = 'foo';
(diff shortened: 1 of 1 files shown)


kirkwaiblinger:
should we just use `declare let`/`declare const` to sidestep any complexity about the possible type?

ronami:
> should we just use `declare let`/`declare const` to sidestep any complexity about the possible type?

Hmm since these two are different types (one is a "fresh literal", the other isn't):

```ts
// not a "fresh literal"
declare const a: 'foo';

// is a "fresh literal"
const b = 'foo';
```

Along with seeing how much complexity there is around "fresh literals", I've changed both tests to use `declare const` so they're simpler but also added one test for a standard "fresh literal" `const key = ...` usage.

Wdyt?

kirkwaiblinger (approved):

PR #10679, opened 2025-01-18
chore(typescript-estree): migrate from `jest` to `vitest`

aryaemami59:
## PR Checklist


.gitattributes
```diff
@@ -3,6 +3,8 @@
 *.json eol=lf
 *.md eol=lf
 *.ts eol=lf
+*.mts eol=lf
+*.cts eol=lf
```
JoshuaKGoldberg, line 7: Unrelated - could you splitting these into a separate PR please?

packages/typescript-estree/tests/lib/createParseSettings.test.ts
```diff
@@ -18,36 +18,44 @@ describe('createParseSettings', () => {
       expect(parseSettings.projectService).toBe(projectService);
     });
 
+    vi.unstubAllEnvs();
```
JoshuaKGoldberg, line 21: [Refactor] At first I wrote:

>  https://vitest.dev/config/#unstubenvs - let's move this to the config level so tests don't have to do it. Tests pass for me when I remove these.

But now, I do see it's in `packages/typescript-estree/vitest.config.mts`. So maybe this is just unnecessary?

vitest.config.mts
```diff
@@ -0,0 +1,20 @@
+import { defaultExclude, defineConfig, mergeConfig } from 'vitest/config';
+import vitestBaseConfig from './vitest.config.base.mjs';
+
+const vitestConfig = mergeConfig(
```
JoshuaKGoldberg, line 4: [Style] Nit: here and in other new config files, 

```suggestion
export default mergeConfig(
```

There's an `'import/no-default-export': 'off'` in `eslint.config.mjs`. Let's switch it to apply to all `*.config.*` files, not just ESLint ones.

.gitignore
```diff
@@ -99,3 +99,6 @@ packages/**/.yarn
 
 .nx/cache
 .nx/workspace-data
+
+.vitest-reports
+tsconfig.vitest-temp.json
```
JoshuaKGoldberg, line 104: [Question] I've never seen these be tracked accidentally, and don't see them created when I `yarn test` and `yarn test --coverage` locally. Is this a common mishap?

If not:

```suggestion
```

packages/typescript-estree/package.json
```diff
@@ -44,13 +44,13 @@
     "syntax"
   ],
   "scripts": {
-    "build": "tsc -b tsconfig.build.json",
+    "build": "tsc -b tsconfig.build.json && yarn run postbuild",
```
JoshuaKGoldberg, line 47: Some unrelated changes here? Merge issue?

packages/typescript-estree/tests/lib/describeFilePath.test.ts
```diff
@@ -1,10 +1,10 @@
 import { describeFilePath } from '../../src/create-program/describeFilePath';
 
 describe('describeFilePath', () => {
-  describe.each(['./repos/repo', '/repos/repo', '~/repos/repo'])(
+  describe.for(['./repos/repo', '/repos/repo', '~/repos/repo'] as const)(
```
JoshuaKGoldberg, line 4: TIL about `.for`, I'd never seen it before! 😄 

[Question] Why the switch? The test passes if I revert changes.

Request: switch back to `.each` for this & the other `.for`s, to reduce the diff size.

packages/typescript-estree/tests/lib/parse.test.ts
```diff
@@ -471,8 +468,9 @@ describe('parseAndGenerateServices', () => {
     });
   });
 
-  if (process.env.TYPESCRIPT_ESLINT_PROJECT_SERVICE !== 'true') {
-    describe('invalid file error messages', () => {
+  describe.runIf(process.env.TYPESCRIPT_ESLINT_PROJECT_SERVICE !== 'true')(
```
JoshuaKGoldberg, line 471: [Praise] TIL about `.runIf`, that's a great API!

knip.ts
```diff
@@ -11,6 +11,7 @@ export default {
     types: 'off',
     unresolved: 'off',
   },
+  vitest: { config: ['vitest.config.mts', 'packages/**/vitest.config.mts'] },
```
JoshuaKGoldberg, line 14: We just have the one level of packages:

```suggestion
  vitest: { config: ['vitest.config.mts', 'packages/*/vitest.config.mts'] },
```

nx.json
JoshuaKGoldberg, line 1: [Question] Why would this be different from before? Do Jest and Vitest not have the same behaviors when it comes to dependencies and tasks?

Same with other Nx changes - what's up?

JoshuaKGoldberg (changes requested):
Awesome progress! Really exciting to see a Vitest success report running `yarn test` locally. 🔥 

Left some questions - I think most of the blockers are around _"why?"_s.

vitest.config.base.mts
JoshuaKGoldberg, line 1: [Feature] This'll also need to configure a setup file for [`console-fail-test`](http://npmjs.com/package/console-fail-test). See also how `jest.config.base.js` sets up `setupFilesAfterEnv`.

vitest.config.mts
aryaemami59: @JoshuaKGoldberg 

> [Style] Nit: here and in other new config files,

I only left it as `const vitestConfig = mergeConfig(...)` so we can debug the `vitest` config using simple `console.log(vitestConfig)` calls.

> There's an `'import/no-default-export': 'off'` in `eslint.config.mjs`. Let's switch it to apply to all `*.config.*` files, not just ESLint ones.

That makes sense but we're not currently getting any lint errors even if we have `export default`s in `vitest.config.mts` files.

.gitignore
aryaemami59: `.vitest-reports` is created if we decide to use [sharding](https://vitest.dev/guide/cli.html#shard) and `tsconfig.vitest-temp.json` file is created if we use [`--typecheck`](https://vitest.dev/guide/testing-types.html#testing-types).

vitest.config.base.mts
aryaemami59: > [Feature] This'll also need to configure a setup file for [`console-fail-test`](http://npmjs.com/package/console-fail-test). See also how `jest.config.base.js` sets up `setupFilesAfterEnv`.

Just out of curiosity why are we using [`console-fail-test`](http://npmjs.com/package/console-fail-test) in the first place?

packages/typescript-estree/package.json
aryaemami59: To me it seemed like `postbuild` was never going to run, which is why I added it. Not sure if exclusion of `postbuild` has been intentional or not.

packages/typescript-estree/tests/lib/createParseSettings.test.ts
aryaemami59: That's a good point. I have made the requested adjustments.

.gitattributes
aryaemami59: Absolutely!

knip.ts
aryaemami59: Done!

.gitattributes
aryaemami59: Done #10741.

vitest.config.mts
JoshuaKGoldberg: Well great, the fewer changes the better 😄

.gitignore
JoshuaKGoldberg: 👍 in that case if we end up using either of those features then we can always add them in then.

packages/typescript-estree/package.json
JoshuaKGoldberg: That would be a separate thing from this PR.

vitest.config.base.mts
JoshuaKGoldberg: To make sure we don't accidentally log to the console. I personally have been bit multiple times by leaving in a `console.log` after debugging with it, then shipping that log to production.

packages/typescript-estree/package.json
aryaemami59: That's fair, would it be okay if I open a separate PR for that?

.gitignore
aryaemami59: Sounds good, I'll remove them.

packages/typescript-estree/package.json
JoshuaKGoldberg: I'd suggest an issue first - I'm on mobile now but IIRC this gets run automatically when you run `build`?

packages/typescript-estree/tests/lib/describeFilePath.test.ts
aryaemami59: I used [**`describe.for`**](https://vitest.dev/api/#describe-for) because it simplifies argument types and aligns with [**`test.for`**](https://vitest.dev/api/#test-for).

packages/typescript-estree/package.json
aryaemami59: It actually doesn't because Yarn doesn't support [Lifecycle Scripts](https://v3.yarnpkg.com/advanced/lifecycle-scripts) like NPM does, so neither `postbuild` nor `postclean` get run automatically when `build` or `clean` are run.

aryaemami59: I'll put up a separate PR for this.

nx.json
aryaemami59: @JoshuaKGoldberg Looks like there are some misconfigurations with `nx`, I'll address those in a separate PR.

packages/typescript-estree/tests/lib/describeFilePath.test.ts
aryaemami59: @JoshuaKGoldberg You're right, I will make the switch to `describe.for` and `test.for` in a followup PR. I made the requested changes.

aryaemami59: #10765

packages/typescript-estree/package.json
aryaemami59: #10765

packages/typescript-estree/vitest.config.mts
```diff
@@ -0,0 +1,24 @@
...
+    test: {
+      dir: `${import.meta.dirname}/tests/lib`,
+      name: packageJson.name,
+      root: import.meta.dirname,
+
+      unstubEnvs: true,
+      unstubGlobals: true,
+
+      exclude: process.env.TYPESCRIPT_ESLINT_PROJECT_SERVICE
+        ? [...defaultExclude, '/node_modules/', 'parse.project-true.test.ts']
```
aryaemami59, line 18:
```suggestion
        ? [...defaultExclude, 'parse.project-true.test.ts']
```

Not sure if `'/node_modules/'` should actually be specified here, I tried to stay as close as I possibly could to the original `jest.config.js` file but `'/node_modules/'` is probably not needed here.

PR #10680, opened 2025-01-18
chore(eslint-plugin): consistently use `it` in tests

43081j:
Changes a few occurrences of `test(name, fn)` to use `it(name, fn)` for consistency with the rest of the codebase.

## PR Checklist

### DIFF
--- a/packages/eslint-plugin/tests/areOptionsValid.test.ts
+++ b/packages/eslint-plugin/tests/areOptionsValid.test.ts
@@ -17,16 +17,16 @@ const exampleRule = createRule<['value-a' | 'value-b'], never>({
   name: 'my-example-rule',
 });
 
-test('returns true for valid options', () => {
+it('returns true for valid options', () => {
   expect(areOptionsValid(exampleRule, ['value-a'])).toBe(true);
 });
 
 describe('returns false for invalid options', () => {
-  test('bad enum value', () => {
+  it('bad enum value', () => {
     expect(areOptionsValid(exampleRule, ['value-c'])).toBe(false);
   });
 
-  test('bad type', () => {
+  it('bad type', () => {
     expect(areOptionsValid(exampleRule, [true])).toBe(false);
   });
 });
--- a/packages/eslint-plugin/tests/docs.test.ts
+++ b/packages/eslint-plugin/tests/docs.test.ts
@@ -221,7 +221,7 @@ describe('Validating rule docs', () => {
       const filePath = path.join(docsRoot, `${ruleName}.mdx`);
       const { fullText, tokens } = parseMarkdownFile(filePath);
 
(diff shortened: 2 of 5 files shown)


kirkwaiblinger (approved):
LGTM!

packages/eslint-plugin/tests/areOptionsValid.test.ts
```diff
@@ -17,16 +17,16 @@ const exampleRule = createRule<['value-a' | 'value-b'], never>({
...
 });
 
-test('returns true for valid options', () => {
+it('returns true for valid options', () => {
   expect(areOptionsValid(exampleRule, ['value-a'])).toBe(true);
 });
 
 describe('returns false for invalid options', () => {
-  test('bad enum value', () => {
+  it('bad enum value', () => {
```
JoshuaKGoldberg, line 25: 🤔 I actually like `test()` here over `it()`. The test name doesn't grammatically read out as a full sentence. I've personally kind of gravitated towards:

* `it()`: for most tests, if there's at all a way to phrase it as _"it X when Y"_ or something like that
* `test()`: as a fallback if there's no logic, just a descriptor - like _"test X"_

JoshuaKGoldberg (commented):
-0.5 from me, but if the rest of the team disagrees I don't mind this being merged at all. Thanks for sending!

packages/eslint-plugin/tests/areOptionsValid.test.ts
43081j: I don't really care which one we use as long as it is one, and not two 

If you can get the others to agree, I'm happy to update

JoshuaKGoldberg: Oh don't get me wrong, I think `it()` is the right choice almost all of the time. Just my _personal nitpicky preference_ is to use `test()` in _specific situations_. 😄

43081j: To be honest, my preference is actually suite/test/assert. But the most consistent thing we can do right now is `it` it seems

As a middle ground, id prefer to later move from the now-consistent `it` to `test ` if anything

kirkwaiblinger:
Oh, I apologize, I made a mistake here. I had assumed that this was a _necessary_ change for moving to vitest, because of its inclusion in https://github.com/typescript-eslint/typescript-eslint/pull/10579. However, now that I realize that vitest has `test()` also I don't think we should do this as part of the vitest effort, sorry. It's just noise on unrelated changes. My opinion is that we should
1. Not proceed with this PR
2. Remove the `test()` -> `it()` changes from #10579 

***

Sincere apologies again to @43081j for my misunderstanding and for requesting this PR only to close it 🙁 And thank you to @JoshuaKGoldberg for pointing out my mistake!

PR #10691, opened 2025-01-21
fix(ast-spec): replace `attributes` with `options` property in `TSImportType`

fisker:
## PR Checklist

## Overview

The AST should align with `ImportExpression` which contains `.options` with `Property`

https://github.com/typescript-eslint/typescript-eslint/pull/10640#issuecomment-2603408332


fisker:
Since `ImportExpression` and `TSImportType` are exactly the same code, it will be much easier if they have same  AST  shape, I'd suggest rename `.argument` to `.source` in future.

JoshuaKGoldberg (approved):
Gotcha, thanks - this makes sense. I see why you'd want to align them to the `options` from `ImportExpression`.

Since we only just released support yesterday for this relatively-uncommon AST usage I think it's fine to treat this as a non-breaking-change fix.

cc @ronami as I might have (again) missed something.

ronami (approved):
Oh, I missed this one too, I didn't consider it not having the location information on the entire second "argument" 🙈 

+1 from me to treat this as non-breaking; I didn't even know this was correct TypeScript syntax until a few days ago 😁

---

Some thoughts:

It seems TypeScript's AST uses `ts.ImportAttributes` over `ts.Expression` for this ([see here](https://github.com/microsoft/TypeScript/blob/8da951cbb629b648753454872df4e1754982aef1/src/compiler/types.ts#L2207)), which is also used in [`ts.ImportDeclaration`](https://github.com/microsoft/TypeScript/blob/8da951cbb629b648753454872df4e1754982aef1/src/compiler/types.ts#L3694) and [`ts.ExportDeclaration`](https://github.com/microsoft/TypeScript/blob/8da951cbb629b648753454872df4e1754982aef1/src/compiler/types.ts#L3775).

To my understanding, `typescript-eslint`'s AST coverts these to `ImportAttribute[]` (e.g. [1](https://github.com/typescript-eslint/typescript-eslint/blob/55bed7528decc783284e8b2785f6d2afeb0cfaa0/packages/typescript-estree/src/convert.ts#L2133-L2136), [2](https://github.com/typescript-eslint/typescript-eslint/blob/55bed7528decc783284e8b2785f6d2afeb0cfaa0/packages/typescript-estree/src/convert.ts#L2211-L2214)), so this would be the first time a `ts.ImportAttributes` node is converted into something other than `ImportAttribute[]`.

Still, I think the change is correct, as the existing AST doesn't include location information for the second "argument", making it really hard to use.

Thanks for pushing this @fisker 👍

bradzacher (dismissed):

JoshuaKGoldberg (commented):
Noting #10649: is there more work to be done here?

fisker:
Updated vistorKeys.

packages/visitor-keys/src/visitor-keys.ts
JoshuaKGoldberg, line 1: https://github.com/typescript-eslint/typescript-eslint/pull/10649/files also has updates to `ExportAllDeclaration` and `ExportNamedDeclaration`. Those should also be updated here, right?

JoshuaKGoldberg (changes requested):
One question ~

packages/visitor-keys/src/visitor-keys.ts
fisker: I don't think they are related to this PR. This PR should only focus on `TSImportType`.

JoshuaKGoldberg (approved):
LGTM, thanks. I don't want to keep delaying on this - it's a good change and the longer we wait the more likely an issue will pop up. Merging now to get in today's release.

PR #10698, opened 2025-01-22
fix(eslint-plugin): [strict-boolean-expressions] allowNullableBoolean - support array predicate functions

### LINKED ISSUE #10688: Bug: [strict-boolean-expressions] allowNullableBoolean - support array predicate functions

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.tsx&code=LAKAxg9gdgzgLgAgIYC4EAoCuUAmBTAMwEso8cEAfBAIwggBs8kpKEpN76BKAbQF0EAXgQ84AJ0x4ANAmz5ipHDPac%2BAblBIAdDAgBbPOnQAPLkIB8CUxpChIsRNTTpaDJiyoru-ISPGTlDnp1UGodfUMTM0FLa1A7aHgEMGdXRmZWOUISMl4BYVEJaVlcbMUQ8HCDI1MLKy4bBIcEHFS6dKg830LJPlAcKsjamPqgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUFs6gAQBcBPABxQGNoBLEggWhXioDsCB6ZA68%2BgIwHt%2BSAIbMGADxLQUyKv2aoMAbXC4IkRNGj9okADSq1OYFGHx4-AO4A5BPGG8kAIUEjm6MF1iIwAX0NGALqq-r5AA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
const a: (undefined | boolean | null)[] = [true, undefined, null];
a.some((x) => x);
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
          "@typescript-eslint/strict-boolean-expressions": [
        "error",
            { "allowNullableBoolean": true }
          ]
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expected there to be no errors.

### Actual Result

It seems like the array functions doesn't respect the "allowNullableBoolean" option

### Additional Info

Probably introduced in this PR https://github.com/typescript-eslint/typescript-eslint/pull/10106

y-hsgw:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10688 

## Overview

Fixed the bug described in the title.

🐛


packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -332,7 +332,10 @@ export default createRule<Options, MessageId>({
           return type;
         });
 
-      if (returnTypes.every(returnType => isBooleanType(returnType))) {
+      if (
+        returnTypes.every(returnType => isBooleanType(returnType)) ||
+        options.allowNullableBoolean
```
yeonjuan, line 337: [pr preview playground](https://deploy-preview-10698--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.tsx&code=FAEwpgxgNghgTmABBA9gOwM4BdHzjATwC5Fs4BLNAcwG0BdAbmGD0IDoMUBbMACgA9EAXgB8ifgEoGQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZfKsugIwHs%2BSAIZN6AD2LQUySnyaoMAbUiJo0PtEgAaLNhxD48PgHcAyp2YBzdGABmB5Ii3g9UA0eMA5BPCE8kAEICwkw2nLCILhAAvgC6LtEg0UA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)
```ts
  // "allowString": false,
  // "allowNullableBoolean": true
array.some(x => x); // no error
```
It seems that if `allowNullableBoolean` is true, all return statements are allowed regardless of other options.
Perhaps it is necessary to consider other options?

kirkwaiblinger: Yeah. The check doesn't currently support the "allow" options at all. All that the change here does is disable the check entirely if `allowNullableBoolean` is set. What we'd need to do is implement the logic used elsewhere (see, e.g. the `is()` function and `inspectVariantTypes()`) to this case.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -1106,6 +1114,16 @@ export default createRule<Options, MessageId>({
 
       return variantTypes;
     }
+
+    function hasExactTypes(
```
kirkwaiblinger, line 1118: Granted that this function is curried both places it's used, why not just write it in curried form from the start? (Roughly) something like this?
```ts
    function makeHasExactTypes(
      types: Set<VariantType>,
    ): (...wantedTypes: readonly VariantType[]) => boolean {
      return (...wantedTypes) =>
        types.size === wantedTypes.length &&
        wantedTypes.every(type => types.has(type));
    }
```

Specifically, I'd like to avoid using the `f.bind(null, arg1)` as a form of currying since it's pretty rough syntax. `(...args) => f(arg1, ...args)` is preferred where applicable... 

I'll leave it up to you if you'd like to address this concern at the call sites or at the function definition.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -331,8 +331,18 @@ export default createRule<Options, MessageId>({
...
+      const types = inspectVariantTypes(flattenTypes);
 
-      if (returnTypes.every(returnType => isBooleanType(returnType))) {
+      const is = hasExactTypes.bind(null, types);
+
+      if (is('boolean') || is('truthy boolean')) {
+        return;
+      }
+
+      if (is('nullish', 'boolean') && options.allowNullableBoolean) {
```
kirkwaiblinger, line 345: I don't think we want to hard-code support only for `options.allowNullableBoolean`.... 

Consider:
[Playground](https://deploy-preview-10698--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3rGUCeA8gGYCCMueaAXPABQoCuEE8APvAygNYpIDuKANoBdAJQiA3ACgpAelnwMACwCWaeGiVImweACMEUZvxDAAdBanJ0WNSRR4AcqgCiAWwAOGPGQoBGeABebHJ8Yl98NDM0JDcQWhx8IIA%2BEIoAfjMIEBQAc2VRaStUTHg7B2cUdy8fULwAJiC0sNI6qJi4hLqU%2BABvKXh4eXgLMwMwKAY0BGU1RFRgFQwVVDL1IwgTcwGyojpEvEzsvIK%2BncG4DAYYFEUYBhBpQYBfeBAIabPBi5Arm-giEZpk94M8pM9CkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZfKsugIwHs%2BSAIZN6AD2LQUySnyaoMAbXDYcAdyHQmkADRgVqrKuyQh8eHzUA5BPCE8kNgLY9E0dGE6xEB7AF9fAF0VAL8gA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

### Code

```ts
declare const arrayOfArrays: (null | unknown[])[];

// this should be allowed...
const isAnyNonEmptyArray1 = arrayOfArrays.some(array => array?.length);

const isAnyNonEmptyArray2 = arrayOfArrays.some(array => {
  // ...because this condition is allowed.
  if (array?.length) {
    return true;
  } else {
    return false;
  }
});
```

<details>
<summary>Expand</summary>

### Eslint config

```json
{
  "rules": {
    "@typescript-eslint/strict-boolean-expressions": [
      "warn", 
      {
        "allowNullableNumber": true
      }
    ]
  }
}
```

### TypeScript config

```json
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

| package | version |
| -- | -- |
| `@typescript-eslint/eslint-plugin` | `8.21.0` |
| `@typescript-eslint/parser` | `8.21.0` |
| `TypeScript` | `5.7.2` |
| `ESLint` | `9.18.0` |
| `node` | `web` |

</details>


I think instead we want to abtract and respect essentially the whole decision tree encoded in the `checkNode()` function: 

https://github.com/typescript-eslint/typescript-eslint/blob/4dbf48b8eec0321babb4293a1be71b54dd00c018/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L433-L985

kirkwaiblinger (changes requested):

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
kirkwaiblinger: PS - this comment might be irrelevant depending on the outcome of https://github.com/typescript-eslint/typescript-eslint/pull/10698#discussion_r1938572194

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
y-hsgw: I have developed a function that abstracts the logic previously contained within the `checkNode` function in commit 351fdfa. 
I would appreciate it if you could review it.

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -331,8 +331,48 @@ export default createRule<Options, MessageId>({
...
+      ];
+      const types = inspectVariantTypes(flattenTypes);
+      const category = determineTypeCategory(types);
 
-      if (returnTypes.every(returnType => isBooleanType(returnType))) {
+      if (category === 'boolean' || category === 'never') {
+        return;
+      }
+
+      if (category === 'nullable boolean' && options.allowNullableBoolean) {
```
kirkwaiblinger, line 344: I'm thinking we also want to share all this logic between the cases, right? Noting that there's some sort of bug here, too, see 

(default options)
```ts
declare const nullOrString: string | null;

// no report
['one', null].filter((x) => nullOrString);

// does report
if (nullOrString) {
}
```

But furthermore I'd have expected the report message to be more like 

"Unexpected nullable string value value in array predicate. An explicit comparison or type conversion is required, etc.", basically the same as the other one, but with "in array predicate" instead of "in conditional".

Thoughts?

kirkwaiblinger (changes requested):

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
y-hsgw: I've shared the logic you pointed out across the cases.→ [a532171](https://github.com/typescript-eslint/typescript-eslint/pull/10698/commits/a53217115e2f39041e5954658480916b9f524416)
I'm currently working on tests and the report message.
Please let me know if there's any misunderstanding or misalignment at this stage!

packages/eslint-plugin/src/rules/strict-boolean-expressions.ts
```diff
@@ -76,36 +85,63 @@ export default createRule<Options, MessageId>({
       conditionErrorAny:
         'Unexpected any value in conditional. ' +
```
kirkwaiblinger, line 86: Here and throughout, I think you'll want to use the messaging templating syntax.
```suggestion
        'Unexpected any value in {{context}}. ' +
```
and, usage:
```ts
context.report({
    messageId: 'conditionErrorAny',
    data: { context: 'array predicate' /* or 'conditional' */ }
    // etc
});
```

I think that can simplify the changeset a lot?

kirkwaiblinger (changes requested):

kirkwaiblinger (changes requested):
There's still quite a bit of duplication in this changeset that I'd like to cut down on.... It's hard to enumerate all the details of such a complex change, though, so I've created a branch with my suggestions. In short, I suggest
- modifying the `determineTypeCategory` function to also be responsible for interpreting the options, and return the report type/message.
- extracting the logic mapping report type/message to its corresponding suggestion fixers in order to reuse it for the function literals case (building off your idea!)
- always providing the `(): boolean` fix for function literals
- ensuring that `[].map((x): boolean | null => Math.random() > 0.5)` still triggers a report, due to the explicit return type, even though the one return location wouldn't trigger it.

See my suggested changes here, https://github.com/y-hsgw/typescript-eslint/pull/2, which you're welcome to merge onto your branch wholesale if you like

y-hsgw:
Thank you so much for your tremendous support! I've merged your PR into this branch.

kirkwaiblinger (approved):
Sweet! 👍 

Will want a second review from @typescript-eslint/triage-team

JoshuaKGoldberg (approved):
Whew! This looked bigger and scarier than it ended up being - nice job extracting to the shared data plumbing. 🔥 

Alec Baldwin as Jack Donaghy excitedly smiling and giving two thumbs up. Caption: "GREAT!"

PR #10708, opened 2025-01-24
fix(eslint-plugin): [prefer-nullish-coalescing] report on chain expressions in a ternary

### LINKED ISSUE #10531: Bug: [prefer-nullish-coalescing] doesn't report on chain expressions in a ternary

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&esQuery=N4IgZglgNgLgpgJxALhCAvkA&code=JYOwLgpgTgZghgYwgAgEIHsAeyDeAoZZANzgBsBXCALmQGcwpQBzAbjwF888ATCBUuFBQxyIBGGDoQyJhDAAxMqQBGiANYZMACgCUNTWx58BQ5Ain1kveOVJhNNHMjiPky-VmQAfZKOugIbmR2YMNefkEUcxBLazhbeywAeQAHCSkyVzgAfldlXLRPHz8IGACgkPZDAHpq5CEU9CgwWjMmoXFSAE8jGztNADo4AeVkAEIAXmQQW1JkbKtS%2BP6sIZHkGlkFJVUEDSxdGrraAAt0WyCGprBe5cTMVPSQMiHs9cnp2fnFvvvHyWepFe602ckUpBU6k0hy4tToZwuzlItHQ9QgjWatwSmn%2BGSBOXeUxmEO%2BcWxyTSAJew1GoO2EN2%2B20OhYQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6KL2jjokcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
interface Box {
  value: string;
}

declare function getFallbackBox(): Box;

declare const defaultBox: { a: { b: Box | undefined } };

declare const defaultBoxOptional: { a?: { b?: Box | undefined } };

// reports correctly
defaultBox.a.b != null ? defaultBox.a.b : getFallbackBox();

// should report
defaultBoxOptional.a?.b != null ? defaultBoxOptional.a?.b : getFallbackBox();

// should also report
defaultBoxOptional.a?.b != null ? defaultBoxOptional.a.b : getFallbackBox();
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/prefer-nullish-coalescing": "error",
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expected all cases above to be reported with a suggested fix.

### Actual Result

Only the first case (member-expression) is reported.

### Additional Info

I think the third case (chain-expression with a member-expression as an alternate) is the more common case, as `no-unnecessary-condition` will flag the second case (chain-expression).

OlivierZal:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10531


OlivierZal:
Ready for review

OlivierZal:
Will need a rebase once https://github.com/typescript-eslint/typescript-eslint/pull/10732 merged

OlivierZal:
Hi @kirkwaiblinger, I propose a code refactorization to handle optional chain except the cases I mention in https://github.com/typescript-eslint/typescript-eslint/issues/10733#issuecomment-2630549948: I leave them for a dedicated PR in the perimeter of https://github.com/typescript-eslint/typescript-eslint/issues/10733 since I think the fix will handle these cases too.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -428,22 +438,21 @@ export default createRule<Options, MessageIds>({
...
-              isNodeEqual(testNode, node.alternate)
+              isNodeOrNodeExpressionEqual(
+                testNode,
+                getNonNullishBranchNode(node, operator),
+              )
             ) {
-              identifierOrMemberExpression = testNode;
+              if (!nullishCoalescingLeftNode) {
+                nullishCoalescingLeftNode = testNode;
+              }
```
kirkwaiblinger, line 448: Ironically, this line could use the nullish coalescing operator 😁 

Also the reason that it needs to be only assigned once is extremely subtle IMO and deserves a comment. WDYT?

```suggestion
                // We want only the first expression in a multi-part nullish check, 
                // since subsequent expressions might not need all the ?.s. 
                // Eg a?.b?.c !== undefined && a.b.c !== null ? a.b.c : 'foo';
                nullishCoalescingLeftNode ??= testNode;
```

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -647,3 +652,51 @@ function isMixedLogicalExpression(
...
+  node: TSESTree.ConditionalExpression,
+  operator: NullishCheckOperator,
+): { nonNullish: TSESTree.Expression; nullish: TSESTree.Expression } {
+  if (!operator || ['!=', '!=='].includes(operator)) {
+    return { nonNullish: node.consequent, nullish: node.alternate };
+  }
+  return { nonNullish: node.alternate, nullish: node.consequent };
+}
+
+function getNonNullishBranchNode(
```
kirkwaiblinger, line 690: Not sure that these helpers need to be separate things from the `getBranchNodes()` helper, but it's ok either way

packages/eslint-plugin/src/util/isNodeEqual.ts
```diff
@@ -29,5 +29,11 @@ export function isNodeEqual(a: TSESTree.Node, b: TSESTree.Node): boolean {
       isNodeEqual(a.property, b.property) && isNodeEqual(a.object, b.object)
     );
   }
+  if (
```
kirkwaiblinger, line 32: This is a relatively sensible change, but no `prefer-nullish-coalescing` tests fail if I remove this change. I'm thinking we either want to 
1. Revert this change, letting the `prefer-nullish-coalescing` `isNodeOrNodeExpressionEqul()` helper handle this as it already does, or
2. Include tests for the other places `isNodeEqual()` is used to demonstrate its impact on other rules and that that impact is beneficial, not harmful.

FWIW, if we go with 2., I think that it probably should be changed to recurse on `a` or `b` being a chain expression _separately_ (like `isNodeOrNodeExpressionEqual()` does) since it makes sense to consider `a?.b` to be equal to the `a?.b` portion of `a?.b.c`, if that makes sense. Also we should probably document that the function does not make a distinction between `a?.b.c` and `a.b?.c`, in other words, it doesn't check _which_ member access are optional. That's the right behavior for prefer-nullish-coalescing, due to how the nodes being compared are related, but might not be true in general, idk 🤷‍♂️ .

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
kirkwaiblinger, line 1: I think the one thing I'm missing in the tests is any cases where the RHS has _more_ optional chains than the LHS. So, for example, something like
```ts
a.b?.c != null ? a?.b?.c : 'something else';
```
Should be rewritten `a.b?.c ?? 'something else'`. The extra `?.` in the consequent are syntactically provably unnecessary (assuming well-behaved getters, of course).

The code does this correctly, but let's just add some test coverage.

kirkwaiblinger (changes requested):
Great work on _really_ tricky code. And good job with the thorough test cases! Helped me out quite a bit with reviewing. Just a few minor comments.

kirkwaiblinger:
> Hi @kirkwaiblinger, I propose a code refactorization to handle optional chains except the cases I mention in [#10733 (comment)](https://github.com/typescript-eslint/typescript-eslint/issues/10733#issuecomment-2630549948): I would leave them for a dedicated PR in the perimeter of #10733 since I think the fix could handle these cases as well.

Agreed, let's not worry about those cases for this PR... As I understand it, https://github.com/typescript-eslint/typescript-eslint/issues/10733#issuecomment-2630549948 brings up separate edge cases.

packages/eslint-plugin/tests/rules/prefer-nullish-coalescing.test.ts
OlivierZal: Tests added

packages/eslint-plugin/src/util/isNodeEqual.ts
OlivierZal: I get your point: I've chosen to go with option 1 in order to focus on solving the related issue and avoid unnecessarily handling of potential side-effects on other rules (but I could handle a `isNodeEqual()` update in a separate issue if you feel it's accurate).

kirkwaiblinger (approved):
Lovely! Thank you! 
clappy

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -647,3 +656,40 @@ function isMixedLogicalExpression(
...
+  if (
+    a.type === AST_NODE_TYPES.ChainExpression ||
+    b.type === AST_NODE_TYPES.ChainExpression
+  ) {
+    if (a.type === AST_NODE_TYPES.ChainExpression) {
+      a = a.expression;
+    }
+    if (b.type === AST_NODE_TYPES.ChainExpression) {
+      b = b.expression;
+    }
```
JoshuaKGoldberg, line 673: [Refactor] There's a `skipChainExpression` function in `prefer-promise-reject-errors`. I  tried reusing it here and all tests passed locally. `no-floating-promises`, `no-inferrable-types`, `prefer-find`, and `prefer-string-starts-ends-with` have some similar logics that I think could mostly or completely be switched to using the utility. WDYT?

This is definitely not a blocker, so if you don't see this by release time Monday, I can just merge and file a followup issue.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
```diff
@@ -475,6 +482,12 @@ export default createRule<Options, MessageIds>({
             return true;
           }
 
+          const tsNode = parserServices.esTreeNodeToTSNodeMap.get(
+            nullishCoalescingLeftNode,
+          );
+          const type = checker.getTypeAtLocation(tsNode);
```
JoshuaKGoldberg, line 488: [Refactor] Nit: a nicer new API for this ~

```suggestion
          const type = parserServices.getTypeAtLocation(
            nullishCoalescingLeftNode,
          );
```

This is existing code, so definitely not a blocker 😄

JoshuaKGoldberg (approved):
Looks great to me, nicely done! 🔥 

Just a couple of small non-blocking suggestions. We can always file a followup for the `skipChainExpression` one / do the `typeAtLocation` one ourselves, no worries.

packages/eslint-plugin/src/rules/prefer-nullish-coalescing.ts
OlivierZal: Nice!

OlivierZal:
Hi @JoshuaKGoldberg, thanks for your review!

As advised I've moved `skipChainExpression` to an util, and used it for `no-floating-promises`.

For `no-inferrable-types`, `prefer-find`, and `prefer-string-starts-ends-with`: if you don't mind I prefer leave the migration to this util for another issue (that I can take), because the chain expressions are involved in recursion that IMO makes it less straightforward.

OlivierZal:
@JoshuaKGoldberg, your suggestion to use `skipChainExpression` allowed me to find that this case is not taken into account:

https://deploy-preview-10708--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.tsx&code=CYUwxgNghgTiAEEQBd4E8Bc8oDs0G4BYAKFElgSVQA8sBvbAfnvgCMWwsBWeAXz75FiJABTUAdFEbjWASmlh4jeBKgzxirAGYgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6KL2jjokcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

OlivierZal:
@kirkwaiblinger, @JoshuaKGoldberg, I didn't take into account the possibility to have parenthesis within the chain. Will work on it.

JoshuaKGoldberg:
Good point! Marking as `status: waiting for author` to keep this off the review queue until it's ready. Whenever you re-request review, we can take another look. Thanks for the thoroughness!

OlivierZal:
@JoshuaKGoldberg, @kirkwaiblinger, I think I've fixed the missing use cases

OlivierZal:
@JoshuaKGoldberg, @kirkwaiblinger, I think I've fixed the missing use cases:

https://deploy-preview-10708--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.tsx&code=CYUwxgNghgTiAEEQBd4E8Bc8oDs0G4BYAKFElgSVQA8sBvbAfnvgCMWwsBWeAXz75FiJABTUAdFEbjWASmlh4jeBKkzxirAGYhYydLkKl8MdKiz1m%2BDqA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WlqYSNkAC1pkA9gEMkyMswDm6KL2jjokcGAC%2BILUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

kirkwaiblinger (changes requested):
One pedantic complaint about naming but otherwise thumbs up 👍

OlivierZal:
Fixed!

kirkwaiblinger (approved):
Delightful. Thanks again!

OlivierZal:
I’m pretty sure the PR was green before Kirk’s approval, maybe I messed up when I created a new PR from this branch…

kirkwaiblinger:
maybe try merging main and we'll see? There were some changes to CI recently

OlivierZal:
Indeed, CI issue that merging main didn’t solve. I’m not sure I know how to fix it, I’ll need to investigate.

OlivierZal:
Will be fixed by https://github.com/typescript-eslint/typescript-eslint/pull/10870

OlivierZal:
@JoshuaKGoldberg, you were faster 😉

OlivierZal:
Seems there’s a failing snapshot on main.

JoshuaKGoldberg:
Hoping to fix it with #10880. In the meantime, we're good to go here. ✅

PR #10710, opened 2025-01-24
fix(eslint-plugin): [no-shadow] ignore declare variables in definition files shadowing global variables

### LINKED ISSUE #2654: [no-shadow] declaring a global causes no-shadow to error
Hey, I think I found what's either a bug or a confusing way that globals and ambient types are handled (potentially similarly to the `no-undef` issue [mentioned in the FAQs](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/FAQ.md#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors)). 

I'm currently migrating a large codebase to use TypeScript. We have a global `has()` function that we use to determine what environment we're in. It's handled by our build system, and is set as a read-only global in our existing ESLint config. 

When I try and declare an ambient type for this global, I've noticed that `@typescript-eslint/no-shadow` thinks that I'm shadowing it. Removing `globals: { has: false }` from the config fixes this problem. Is this expected behavior?

Thanks in advance, and sorry if I'm overlooking something.

**Repro**

I set up a fresh project by running the following in an empty directory:
```sh
yarn init -y
yarn add -D eslint typescript @typescript-eslint/parser @typescript-eslint/eslint-plugin
```

I then added these two files
```js
// .eslintrc.js
module.exports = {
  root: true,
  plugins: [
    '@typescript-eslint',
  ],
  extends: [
    'eslint:recommended',
    'plugin:@typescript-eslint/recommended',
  ],
  globals: {
    has: false,
  },
  rules: {
    "no-shadow": "off",
    "@typescript-eslint/no-shadow": ["error", { builtinGlobals: true }],
  }
};
```

```TS
// index.d.ts
declare const has: (environment: "dev" | "prod" | "test") => boolean;
```

**Expected Result**

I'd expect `yarn eslint index.d.ts` to not error.

**Actual Result**

It logged the following error:
```
 1:15  error  'has' is already declared in the upper scope  @typescript-eslint/no-shadow
```

**Versions**

| `@typescript-eslint/eslint-plugin` | `4.4.0` |
| `@typescript-eslint/parser`        | `4.4.0` |
| `ESLint`                           | `7.10.0` |
| `node`                             | `12.15.0` |

sopa301:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2654

## Overview

I added an additional guard clause checking if a variable has a `declare` tag and removing it from consideration under no-shadow if it does. The code may look very clunky because the conditional statements were derived through experimentation, so they may not catch all cases. Apologies for the mess, and do let me know how I may improve it.

There's one thing I couldn't figure out how to handle: The test case on `no-shadow.test.ts`, line `953` demonstrates the current behaviour when there is a `declare` keyword variable being shadowed by another in an inner scope. Should we have the error point to the `declare` variable or the global variable? 🧼

### DIFF
--- a/packages/eslint-plugin/src/rules/no-shadow.ts
+++ b/packages/eslint-plugin/src/rules/no-shadow.ts
@@ -3,7 +3,7 @@ import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
 import { DefinitionType, ScopeType } from '@typescript-eslint/scope-manager';
 import { AST_NODE_TYPES, ASTUtils } from '@typescript-eslint/utils';
 
-import { createRule } from '../util';
+import { createRule, isDefinitionFile } from '../util';
 import { isTypeImport } from '../util/isTypeImport';
 
 export type MessageIds = 'noShadow' | 'noShadowGlobal';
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/no-shadow.ts
```diff
@@ -566,6 +566,28 @@ export default createRule<Options, MessageIds>({
       };
     }
 
+    /**
+     * Checks if the initialization of a variable has the declare modifier in a
+     * definition file.
+     * @param variable The variable to check.
+     * @returns Whether or not the initialization of a variable has the declare
+     * modifier in a definition file.
```
JoshuaKGoldberg, line 574: [Docs] Nit: I don't think these JSDocs add anything. From just the name we know this is checking _"is declare in dts file?"_ for a _"variable"_. The first brief description is the only piece that adds new info IMO.

```suggestion
```

```suggestion
```

JoshuaKGoldberg (dismissed):
🔥 looks great to me, thanks!

PR #10718, opened 2025-01-27
feat(utils): add `reportUnusedInlineConfigs` to LinterOptions

FloEdelmann:
Follows https://github.com/eslint/eslint/pull/19201, released in https://github.com/eslint/eslint/releases/tag/v9.19.0. Currently, that flag is not yet usable with `typescriptEslint.config(…)` because of the missing type.

See also https://eslint.org/docs/latest/use/configure/configuration-files#configuration-objects.

### DIFF
--- a/packages/utils/src/ts-eslint/Config.ts
+++ b/packages/utils/src/ts-eslint/Config.ts
@@ -203,6 +203,8 @@ export namespace FlatConfig {
     /**
      * A severity string indicating if and how unused inline directives
      * should be tracked and reported.
+     *
+     * since ESLint 9.19.0
      * @default "off"
      */
     reportUnusedInlineConfigs?:
--- a/packages/utils/src/ts-eslint/Config.ts
+++ b/packages/utils/src/ts-eslint/Config.ts
@@ -200,6 +200,14 @@ export namespace FlatConfig {
       | boolean
       | SharedConfig.Severity
       | SharedConfig.SeverityString;
+    /**
+     * A severity string indicating if and how unused inline directives
+     * should be tracked and reported.
+     * @default "off"
+     */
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (approved):
👍 thanks!

packages/utils/src/ts-eslint/Config.ts
```diff
@@ -200,6 +200,14 @@ export namespace FlatConfig {
       | boolean
       | SharedConfig.Severity
       | SharedConfig.SeverityString;
+    /**
+     * A severity string indicating if and how unused inline directives
+     * should be tracked and reported.
+     * @default "off"
+     */
+    reportUnusedInlineConfigs?:
```
kirkwaiblinger, line 208: I'm wondering - should this have an indication that the option is available only since eslint 9.19.0? 

We did something similar in previous work, see

https://github.com/typescript-eslint/typescript-eslint/blob/f5e23e28653705fc7021d8c780f0d63656dea518/packages/utils/src/ts-eslint/Rule.ts#L73-L80

JoshuaKGoldberg (changes requested):
Agreed with Kirk, a `since` would be good.

packages/utils/src/ts-eslint/Config.ts
FloEdelmann: Added in [`a899d2f`](https://github.com/typescript-eslint/typescript-eslint/pull/10718/commits/a899d2fce98788aba76002b3c3a4a4a1a0ace047).

kirkwaiblinger (approved):
Thanks!

JoshuaKGoldberg:
Hmm, interesting spot... That probably deserves its own issue.

PR #10725, opened 2025-01-28
docs: refreshed typescript-eslint package docs, especially tseslint.config

### LINKED ISSUE #9733: Docs: Clarity issue on Flat config `extends` description

### Suggested Changes

Coming over from https://stackoverflow.com/questions/78827606/configure-eslint-eslint-config-js-to-only-analyze-typescript-files-ts/78828601?noredirect=1#comment138996204_78828601: re-reading the description in https://typescript-eslint.io/packages/typescript-eslint/#flat-config-extends:

> The `tseslint.config` utility function also adds handling for the `extends` property on flat config objects. This allows you to more easily extend shared configs for specific file patterns whilst also overriding rules/options provided by those configs:

...I do agree with the friendly StackOverflow commenter that the docs aren't super clear on what the thing _does_. They just speak to its purpose.

I'm thinking something like this as a rephrase, maybe?

> `tseslint.config` takes in a config object that includes an `extends` array of configs to extend. It returns the equivalent ESLint config of applying the rest of the settings for each extension.

### Affected URL(s)

https://typescript-eslint.io/packages/typescript-eslint/#flat-config-extends

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9733

## Overview

Primarily applies two refactors on the page:

* Moves `config(...)` docs much higher, so people are more likely to see them
  * The eslintrc migration section is moved to the very end - there are still folks who need it but it's not quite as commonplace as it was half a year ago
* Rewords the description of `extends` to what's suggested in the issue, rather than the now-no-longer-technically-true variadic function concept

Also touches up on some other phrasings:

* Slightly longer descriptions introductory table, to explain more about the exports
* `// ...` instead of the 2nd+ rule in `rules: { ... }`s, for brevity
  * Switched from `no-unsafe-*` to `no-floating-promises` since it's more commonly known on its own

💖

### DIFF
--- a/docs/packages/TypeScript_ESLint.mdx
+++ b/docs/packages/TypeScript_ESLint.mdx
@@ -162,7 +162,7 @@ export default tseslint.config({
 
 ### Manual usage
 
-`tseslint.config` specifies typescript-eslint `parser` and `plugin` options for you, so there is no need to manually provide those.
+[typescript-eslint's recommended and stylistic configurations](../users/configs) specify typescript-eslint `parser` and `plugin` options for you, so there is no need to manually provide those.
 However, in complex ESLint configurations, you may find yourself manually specifying those options yourself.
 
 #### Manually configuring our plugin and parser
@@ -204,7 +204,7 @@ This has been the standard namespace for our plugin for many years and is what u
 You may choose a different namespace - but note that currently [flat configs allow the same plugin to be registered, configured, and have duplicate reports under multiple namespaces](https://github.com/eslint/eslint/discussions/17766).
 :::
 
-#### Usage with other plugins
+### Usage with other plugins
 
 Below is a more complex example of how you might use our tooling with flat configs.
 This config:
--- a/docs/packages/TypeScript_ESLint.mdx
+++ b/docs/packages/TypeScript_ESLint.mdx
@@ -16,12 +16,12 @@ This package is the main entrypoint that you can use to consume our tooling with
 
 This package exports the following:
 
-| name      | description                                                     |
-| --------- | --------------------------------------------------------------- |
-| `config`  | A utility function for creating type-safe flat configs          |
-| `configs` | [Our eslint (flat) configs](../users/Shared_Configurations.mdx) |
-| `parser`  | [Our parser](./Parser.mdx)                                      |
-| `plugin`  | [Our plugin](./ESLint_Plugin.mdx)                               |
+| Name      | Description                                                                            |
+| --------- | -------------------------------------------------------------------------------------- |
+| `config`  | A utility function for creating type-safe flat configs -- see [`config(...)`](#config) |
+| `configs` | [Shared ESLint (flat) configs](../users/Shared_Configurations.mdx)                     |
+| `parser`  | A re-export of [`@typescript-eslint/parser`](./Parser.mdx)                             |
+| `plugin`  | A re-export of [`@typescript-eslint/eslint-plugin`](./ESLint_Plugin.mdx)               |
 
 ## Installation
 
@@ -29,19 +29,9 @@ This package exports the following:
 npm i typescript-eslint
 ```
 
-### Migrating from "legacy" config setups
-
-If you're migrating from a "legacy" `.eslintrc` configuration setup you likely have our plugin and parser installed separately. This package includes these as dependencies so you can freely uninstall your local references:
-
-```bash npm2yarn
-npm un @typescript-eslint/parser @typescript-eslint/eslint-plugin
-```
-
-For more information on migrating from a "legacy" config setup, see [ESLint's Configuration Migration Guide](https://eslint.org/docs/latest/use/configure/migration-guide).
-
 ## Usage
 
-The simplest usage of this package would be:
+We recommend getting started by using `tseslint.config` helper function in your ESLint config:
 
 ```js title="eslint.config.mjs"
 // @ts-check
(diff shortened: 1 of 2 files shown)


docs/packages/TypeScript_ESLint.mdx
```diff
@@ -55,111 +45,16 @@ export default tseslint.config(
...
-);
-```
+This config file exports a flat config that enables both the [core ESLint recommended config](https://www.npmjs.com/package/@eslint/js) and [our recommended config](../users/Shared_Configurations.mdx#recommended).
 
-## `config(...)`
+### `config(...)`
 
-The `config` function is a [variadic](https://en.wikipedia.org/wiki/Variadic_function) [identity function](https://en.wikipedia.org/wiki/Identity_function) which is a fancy way of saying that it's a function with a spread argument that accepts any number flat config objects and returns the objects unchanged. It exists as a way to quickly and easily provide types for your flat config file without the need for JSDoc type comments.
+`tseslint.config` takes in any number of ESLint config objects.
+Each of those config objects may additionally include an `extends` array of configs to extend.
```
kirkwaiblinger, line 53: Consistency with next line:
```suggestion
`tseslint.config(...)` takes in any number of ESLint config objects.
Each of those config objects may additionally include an `extends` array of configs to extend.
```

more natural phrasing? 
```suggestion
`tseslint.config` takes in any number of ESLint config objects, each of which may additionally include an `extends` array of configs to extend.
```

Combined should you choose to accept both:
```suggestion
`tseslint.config(...)` takes in any number of ESLint config objects, each of which may additionally include an `extends` array of configs to extend.
```

docs/packages/TypeScript_ESLint.mdx
```diff
@@ -204,12 +99,15 @@ export default [
 </Tabs>
 
 :::note
-We _**strongly**_ recommend using this utility to improve the config authoring experience — however it is entirely optional. By choosing not to use it you lose editor autocomplete and type checking for config files but otherwise it _will not_ impact your ability to use our tooling.
+We _**strongly**_ recommend using this utility to improve the config authoring experience — however it is entirely optional.
+By choosing not to use it you lose editor autocomplete and type checking for config files.
```
kirkwaiblinger, line 103: and also the extends key 🤷

docs/packages/TypeScript_ESLint.mdx
```diff
@@ -261,3 +159,110 @@ export default tseslint.config({
   },
 });
 ```
+
+### Manual usage
+
+`tseslint.config` specifies typescript-eslint `parser` and `plugin` options for you, so there is no need to manually provide those.
```
kirkwaiblinger, line 165: ... No it doesn't... right? It's the `tseslint.configs.recommended[0]` (or `tseslint.configs.strict[0]`, or the `*-type-checked` variants of those) that does it?

docs/packages/TypeScript_ESLint.mdx
```diff
@@ -261,3 +159,110 @@ export default tseslint.config({
...
+
+:::warning
+We **_strongly_** recommend declaring our plugin with the namespace `@typescript-eslint` as shown above.
+If you use our shared configs this is the namespace that they use.
+This has been the standard namespace for our plugin for many years and is what users are most familiar with.
+
+You may choose a different namespace - but note that currently [flat configs allow the same plugin to be registered, configured, and have duplicate reports under multiple namespaces](https://github.com/eslint/eslint/discussions/17766).
+:::
+
+#### Usage with other plugins
```
kirkwaiblinger, line 207: Not sure this should be a subsection of "manual" as opposed to its own section along the lines of "Full, worked example"?

kirkwaiblinger (changes requested):

docs/packages/TypeScript_ESLint.mdx
JoshuaKGoldberg: For now. https://github.com/eslint/rfcs/pull/126 was accepted and https://github.com/eslint/eslint/issues/19116 is in _Implementing_ status. I think mentioning `extends` won't be accurate for very long.

docs/packages/TypeScript_ESLint.mdx
JoshuaKGoldberg: Right, yes 😅

bradzacher (approved):
lgtm15

docs/packages/TypeScript_ESLint.mdx
kirkwaiblinger: hmm, but what's scary to me about not mentioning it at all is that https://github.com/eslint/eslint/issues/19116 will have `extends` with different semantics. Is there a way to preempt that? I'm ok with whatever you think is best here

kirkwaiblinger (commented):
Oh - the other thing that I think might be good to mention would be the `.flat(Infinity)` call that makes the `...spread` unnecessary. Same convo as https://github.com/typescript-eslint/typescript-eslint/pull/10725#discussion_r1958653484 I suppose

kirkwaiblinger (approved):
Approving since my feedback isn't necessarily requesting changes but curious for your thoughts

docs/packages/TypeScript_ESLint.mdx
JoshuaKGoldberg: Hmm, true... proposal: can we split this out as a followup issue? I don't have a strong gut feeling on the right way to preempt this.

PR #10728, opened 2025-01-28
docs(eslint-plugin): [class-methods-use-this] refresh options docs

### LINKED ISSUE #8729: Docs: [class-methods-use-this]  ignoreClassesThatImplementAnInterface public-fields option's description and playground link is not clear

### Suggested Changes

In [class-methods-use-this's ignoreClassesThatImplementAnInterface options's public-fields docs](https://typescript-eslint.io/rules/class-methods-use-this/#public-fields), I found two errors.

1. description is not clear
the description is `Example of incorrect code when ignoreClassesThatImplementAnInterface is set to 'public-fields':` but, below example code is incorrect and correct code. I think only correct code is enough to explain the option and above ignoreClassesThatImplementAnInterface's true example should fix.

result is in below

ignoreClassesThatImplementAnInterface's value is true
```javascript
class X implements Y {
  method() {};
  property = () => { };

  private privateMethod() { };
  private privateProperty = () => { };

  protected protectedMethod() { } ;
  protected protedtedProperty = () => { };
}
```

ignoreClassesThatImplementAnInterface's value is public-fields
```javascript
class X implements Y {
  method() {}
  property = () => {};
}
```

2. playground link
correct and incorrect error's [playground link](https://typescript-eslint.io/play/#eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQMbwENlkBaAW0QBcALAewBNTZlESaBLVDSWgM18gAacNkgABSgE8ADilzR20yiRTx2AO0oB6fEVIUaDJizbVO6KImjRa0SCIC%2BIB0A&code=MYGwhgzhAEAa0EsC2AHEBTJ6B2AXGAmtAN4BQ00WuAFgPYAmAFAJQkC%2B50KATrSut1wBPaAF5oLMQD52AblKceCAG5hc6LtxVr0AWXQ0Gk4hwpLV6zdvUAFXv0Ejxk0TJPzFvdcHX0rFvQM6JlYTT1pvX38dOz4BYTEJVlc5UjYgA&fileType=.ts) is not include ignoreClassesThatImplementAnInterface option that is public-fields

### Affected URL(s)

https://typescript-eslint.io/rules/class-methods-use-this/#public-fields

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8729

## Overview

Aligns the rule's options' descriptions with how most other `ignore*` ones are described.

💖

### DIFF
--- a/packages/eslint-plugin/docs/rules/class-methods-use-this.mdx
+++ b/packages/eslint-plugin/docs/rules/class-methods-use-this.mdx
@@ -10,7 +10,7 @@ import TabItem from '@theme/TabItem';
 > See **https://typescript-eslint.io/rules/class-methods-use-this** for documentation.
 
 This rule extends the base [`eslint/class-methods-use-this`](https://eslint.org/docs/rules/class-methods-use-this) rule.
-It adds support for ignoring `override` methods or methods on classes that implement an interface.
+It adds support for ignoring `override` methods and/or methods on classes that implement an interface.
 
 ## Options
 
@@ -33,8 +33,6 @@ const defaultOptions: Options = {
 
 {/* insert option description */}
 
-Makes the rule ignore any class member explicitly marked with `override`.
-
 Example of a correct code when `ignoreOverrideMethods` is set to `true`:
 
 ```ts option='{ "ignoreOverrideMethods": true }' showPlaygroundButton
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
Will fix up the unit tests soon, I don't they block review...

packages/eslint-plugin/docs/rules/class-methods-use-this.mdx
```diff
@@ -33,8 +33,6 @@ const defaultOptions: Options = {
 
 {/* insert option description */}
 
-Makes the rule ignore any class member explicitly marked with `override`.
```
ronami, line 36: Nice! I just noticed the resulting doc page had more-or-less the same option description twice, as the `/* insert option description */` comments add the original option description.

I've checked if there are other rules that have this issue which led me to #10748. Some rules have options with the same description (for example: https://typescript-eslint.io/rules/no-confusing-void-expression/#ignorearrowshorthand).

I wonder if/how this can be prevented.

ronami (approved):
This is such a big improvement, thank you 🎉

packages/eslint-plugin/docs/rules/class-methods-use-this.mdx
JoshuaKGoldberg: Yeah we could probably add a lint rule or test to check that there isn't a strict duplicate. Or loose duplication if we remove spaces, lowercase, or even check something like the Levenshtein distance and/or [`are-docs-informative`](https://www.npmjs.com/package/are-docs-informative)-style distance. But there are so many that are very differently phrased -and we aren't adding new violations- I don't think it'd catch much long-term. +1 to #10748 from me as a one-time thing.

PR #10730, opened 2025-01-28
docs: emphasize performance tradeoffs of typed linting

### LINKED ISSUE #10202: Docs: Highlight performance issues with type-aware rules

### Suggested Changes

Right now, there is very little mention of how type-aware linting affects performance. If you go to any type-aware rule page, like this:

https://typescript-eslint.io/rules/await-thenable

We see this:

> This rule requires [type information](https://typescript-eslint.io/getting-started/typed-linting) to run.

But there is no mention of any performance issues.

If you click through to the typed linting docs, the only real mention is way down the page:
https://typescript-eslint.io/getting-started/typed-linting/#how-is-performance

I'd really like to see:

1. The note on each rule page updated to include "This rule requires [type information](https://typescript-eslint.io/getting-started/typed-linting) to run, which comes with performance tradeoffs."
2. An "important" block at the top of https://typescript-eslint.io/getting-started/typed-linting/ noting that there is a performance hit for type-aware rules and then linking to the section with more details.

### Affected URL(s)

https://typescript-eslint.io/getting-started/typed-linting/

And each type-aware rule page.

### Additional Info

_No response_

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10202

## Overview

Adds almost no new content, instead emphasizing or reorganizing existing content:

* Adds the suggested _"which comes with performance tradeoffs"_ note to the auto-added cards on top of rules pages
* Moves FAQs from `/getting-started/typed-linting` to the actual typed linting FAQs, promoting the _Performance_ heading to an h2

💖

### DIFF
--- a/docs/getting-started/Typed_Linting.mdx
+++ b/docs/getting-started/Typed_Linting.mdx
@@ -8,6 +8,7 @@ import TabItem from '@theme/TabItem';
 
 Some typescript-eslint rules utilize TypeScript's type checking APIs to provide much deeper insights into your code.
 This requires TypeScript to analyze your entire project instead of just the file being linted.
+
 As a result, these rules are slower than traditional lint rules but are much more powerful.
 
 To tap into TypeScript's additional powers, there are two small changes you need to make to your config file:
--- a/docs/getting-started/Typed_Linting.mdx
+++ b/docs/getting-started/Typed_Linting.mdx
@@ -6,7 +6,8 @@ title: Linting with Type Information
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-Some typescript-eslint rules utilize the awesome power of TypeScript's type checking APIs to provide much deeper insights into your code.
+Some typescript-eslint rules utilize TypeScript's type checking APIs to provide much deeper insights into your code.
+These rules are slower than traditional lint rules but are much more powerful.
 To tap into TypeScript's additional powers, there are two small changes you need to make to your config file:
 
 <Tabs groupId="eslint-config">
@@ -44,7 +45,7 @@ For CommonJS modules and/or older versions of Node.js, [use `__dirname` or an al
 
 In more detail:
 
-- `tseslint.configs.recommendedTypeChecked` is another [shared configuration](../users/Shared_Configurations.mdx) we provide. This one contains recommended rules that additionally require type information.
+- `tseslint.configs.recommendedTypeChecked` is a [shared configuration](../users/Shared_Configurations.mdx). It contains recommended rules that additionally require type information.
 - `parserOptions.projectService: true` indicates to ask TypeScript's type checking service for each source file's type information (see [Parser#projectService](../packages/Parser.mdx#projectService)).
 - `parserOptions.tsconfigRootDir` tells our parser the absolute path of your project's root directory (see [Parser#tsconfigRootDir](../packages/Parser.mdx#tsconfigrootdir)).
 
@@ -78,7 +79,7 @@ module.exports = {
 
 In more detail:
 
-- `plugin:@typescript-eslint/recommended-type-checked` is another [shared configuration](../users/Shared_Configurations.mdx) we provide. This one contains recommended rules that additionally require type information.
+- `plugin:@typescript-eslint/recommended-type-checked` is a [shared configuration](../users/Shared_Configurations.mdx). It contains recommended rules that additionally require type information.
 - `parserOptions.projectService: true` indicates to ask TypeScript's type checking service for each source file's type information (see [Parser#projectService](../packages/Parser.mdx#projectService)).
 - `parserOptions.tsconfigRootDir` tells our parser the absolute path of your project's root directory (see [Parser#tsconfigRootDir](../packages/Parser.mdx#tsconfigrootdir)).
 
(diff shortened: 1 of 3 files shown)


docs/getting-started/Typed_Linting.mdx
```diff
@@ -6,7 +6,8 @@ title: Linting with Type Information
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-Some typescript-eslint rules utilize the awesome power of TypeScript's type checking APIs to provide much deeper insights into your code.
+Some typescript-eslint rules utilize TypeScript's type checking APIs to provide much deeper insights into your code.
+These rules are slower than traditional lint rules but are much more powerful.
```
nzakas, line 10: Just to give a little more insight. "Slower" is a bit vague (how much slower?). The important information is that it's analyzing the whole project, which gives a better idea of what "slower" might mean.

```suggestion
Some typescript-eslint rules utilize TypeScript's type checking APIs to provide much deeper insights into your code. This requires TypeScript to analyze your entire project instead of just the file being linted.

As a result, these rules are slower than traditional lint rules but are much more powerful.
```

packages/website/src/theme/MDXComponents/RuleAttributes.tsx
```diff
@@ -127,7 +127,7 @@ export function RuleAttributes({ name }: { name: string }): React.ReactNode {
           <Link href="/getting-started/typed-linting" target="_blank">
             type information
           </Link>{' '}
-          to run.
+          to run, which comes with performance tradeoffs.
```
nzakas, line 130: Can this be linked to the description in the Typed_Linting.mdx page?

JoshuaKGoldberg: 🤔 I suppose that would help push people to it. But I'm a little hesitant to throw up yet another link. 
* This area is already pretty link-heavy - and in general I'm not a fan of having multiple in a "card"
  * (the _"extending ... in an ESLint configuration ... "_ one already having two links is already an exception I'm not thrilled about...)
* I think the reading flow for typed linting is best started on the definition. Jumping straight to the performance paragraphs is a lot of text that relies on people to have already read the definition. In other words, we expect them to have read what the first link points to anyway.

docs/getting-started/Typed_Linting.mdx
```diff
@@ -6,7 +6,11 @@ title: Linting with Type Information
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-Some typescript-eslint rules utilize the awesome power of TypeScript's type checking APIs to provide much deeper insights into your code.
+Some typescript-eslint rules utilize TypeScript's type checking APIs to provide much deeper insights into your code.
+This requires TypeScript to analyze your entire project instead of just the file being linted.
+
+As a result, these rules are slower than traditional lint rules but are much more powerful.
+
 To tap into TypeScript's additional powers, there are two small changes you need to make to your config file:
```
kirkwaiblinger, line 14: This sentence reads a little awkwardly now.... The "additional powers" is now kind of a dangling reference which used to point to the "...the awesome power...." in the formerly preceding sentence. The new "...much more powerful" doesn't set up quite the same parallelism. Any way to tweak this wording?

docs/getting-started/Typed_Linting.mdx
```diff
@@ -141,125 +145,7 @@ module.exports = {
 
 You can read more about the rules provided by typescript-eslint in our [rules docs](/rules) and [shared configurations docs](../users/Shared_Configurations.mdx).
 
-## FAQs
-
-### Can I customize the TSConfig used for typed linting?
```
kirkwaiblinger, line 146: I think it's mostly ok to move these around, but it changes the meaning to me around `parserOptions.project`... As is - it's indicated as being part of the gamut of possibilities, just a less-recommended approach. After moving this to FAQs it becomes a hidden option that doesn't really have first class documentation at all 🤔.

For my brain, I tend to work much better with communication that indicates "We're going to demo approach A here, but be advised that there are alternative approaches, which you can read up more on at \[here\]\(link\)." rather than "The way to achieve X is the following. Oh, you had problems with that? There are other ways I didn't tell you about, too.". Feels like sort of a ["comefrom" statement](https://en.wikipedia.org/wiki/COMEFROM)

Do you think there's a good way to address this?

packages/website/src/theme/MDXComponents/RuleAttributes.tsx
kirkwaiblinger: I tend to agree that the way it's currently set up feels a bit more natural to me. I'm the type of user that would open both links in new tab and be confused/annoyed that they've basically pointed me to the same place. But I think either way works.

docs/getting-started/Typed_Linting.mdx
JoshuaKGoldberg: Hmm yeah. I merged the first two groups together and switched the third to saying the more straightforward _"enable type typed linting"_, WDYT?

docs/getting-started/Typed_Linting.mdx
JoshuaKGoldberg: I actually like that it's getting more buried over time 😄. It really is _not_ how we want people to configure things long-term. It's >= the amount of configuration, doesn't support monorepos, is primarily for the _"separate TSConfigs for linting"_ use case that is not generally recommended - basically I don't see any reason why we'd want to encourage folks to use it.

But, I do see your point that there's now _no_ way to get to it easily. I added a small list at the bottom. WDYT?

kirkwaiblinger (approved):

PR #10744, opened 2025-01-29
feat(eslint-plugin): [no-redundant-type-constituents] use assignability checking for redundancy checks

### LINKED ISSUE #7742: Enhancement: [no-redundant-type-constituents] Use assignability checking for redundancy checks

https://typescript-eslint.io/rules/no-redundant-type-constituents

### Description

Right now, `no-redundant-type-constituents` is very limited:

> This rule plays it safe and only works with bottom types, top types, and comparing literal types to primitive types.

But sometimes an intersection or union type will have two constituents where one constituent is a superset of the other. It'd be nice to report on those cases too. Or, heck, even cases where they're deeply equal, but declared as different object types.

### Fail

```typescript
type ShallowEqual = number | number;
type DeepEqual = { a: string } | { a: string };
type Superset = { a: string } | { a: "b" | "c" };
```

### Pass

```typescript
type ShallowEqual = string | number;
type DeepEqual = { a: string } | { a: number };
type Superset = { a: string } | { a: "b" | "c" | 0 };
```

### Additional Info

If we could use a type assignability / relationship API (https://github.com/microsoft/TypeScript/issues/9879), then we could catch those. Filing this long-standing rule feature request for tracking.

mdm317:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7742 

## Overview

> even cases where they're deeply equal, but declared as different object types.

This feature seems to overlap with this rule [no-duplicate-type-constituents](https://typescript-eslint.io/rules/no-duplicate-type-constituents/) so I'm not sure whether to include it. 
For now, I'll remove it and proceed.

**deeply equal**
```
type ShallowEqual = number | number;
type DeepEqual = { a: string } | { a: string };
```

**Union**
If x is assignable to y, then in the union x∣y, x is redundant and can be removed. 
However, due to TypeScript's [Excess Property Checks](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks) feature, this logic did not work as expected.

For example, consider the union type `{ a: 1 } | { a: 1, b: 1 }`. Since `{ a: 1, b: 1 }` is assignable to `{ a: 1 }`
So `{ a: 1, b: 1 }` becomes redundant and can be removed.

However, if `{ a: 1, b: 1 }` is removed, the remaining type `{ a: 1 }` exhibits different behavior when declaring values.
For example, `{ a: 1 }` alone can only be assigned to variables of type `{ a: 1 }`, whereas `{ a: 1 } | { a: 1, b: 1 }` allows the declaration of both `{ a: 1 }` and `{ a: 1, b: 1 }`

Example code
```
type Foo = { a: 1 } | { a: 1, b: 1 }
const foo: Foo = { a: 1 }
const bar: Foo = { a: 1, b: 1 }

type Foo2 = { a: 1 } 
const foo2: Foo2 = { a: 1 }
const bar2: Foo2 = { a: 1, b: 1 } //error
```

So, before checking whether one type is assignable to another, I first verified that both objects have the same set of keys. Only then did I proceed with the assignability check.
For `A | B`, i checks if `A` and `B` have the same keys. If they do, it then checks assignability.

**Union Types with Intersection Types**
- Union with a Single Intersection:
  - If a union type contains an intersection, like A | (B & C & D), the intersection (B & C & D) is treated as a single object type.
- Union within an Intersection: 
  - If an intersection includes a union, like A | (B & (C | D)), TypeScript simplifies the intersection into a union of intersections.  ` B & (C | D) is equivalent to B & C | B & D,`
  - This means the assignability check becomes: Is A assignable to (B & C) | (B & D)

**Intersection**
Unlike union types, intersection types did not need excess property checks.
`A & B` in this case if  `A` is assignable to `B` B is rebundant.
- If there is no union type within the intersection, I proceed the assignability check. `A & B`
- If there is union type, I check if every type in the union is compatible with the intersection and follows the same rules.
-   `A & ( B | C)`
```
A >= B, A >= C  Then  A >= (B | C)
B >= A, C >= A  Then   (B | C) >= A
```

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-redundant-type-constituents.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-redundant-type-constituents.test.ts
@@ -291,8 +291,6 @@ ruleTester.run('no-redundant-type-constituents', rule, {
     'type T = Array<string> & [1];',
     'type T = Array<number> & Array<string>;',
     'type T = Array<number> & string[];',
-    'type T = [{ a: 1 }] & [{ a: 1; b?: 1 }];',
-    'type T = [{ a: number }] & [{ a: 1; b: 1 }];',
   ],
   invalid: [
     {
@@ -2276,6 +2274,36 @@ type C = A & B;
             redundantType: '{ a: number; }[]',
           },
           endColumn: 30,
+          messageId: 'typeOverridden',
+        },
+      ],
+    },
+    {
+      code: 'type T = [{ a: 1 }] & [{ a: 1; b?: 1 }];',
+      errors: [
+        {
+          column: 10,
+          data: {
+            container: 'intersection',
+            nonRedundantType: '[{ a: 1; b?: 1 | undefined; }]',
+            redundantType: '[{ a: 1; }]',
+          },
+          endColumn: 20,
+          messageId: 'typeOverridden',
+        },
+      ],
+    },
+    {
+      code: 'type T = [{ a: number }] & [{ a: 1; b: 1 }];',
+      errors: [
+        {
+          column: 10,
+          data: {
+            container: 'intersection',
+            nonRedundantType: '[{ a: 1; b: 1; }]',
+            redundantType: '[{ a: number; }]',
+          },
+          endColumn: 25,
           messageId: 'typeOverridden',
         },
       ],


JoshuaKGoldberg:
👋 Ping @mdm317, are you planning on un-draft ing this soon? We'd prefer not to keep drafts around for too long in case someone else would want to send one too.

mdm317:
Sorry for the delay. This issue was more challenging than I expected.
I'll try to finish it within two days. If it takes longer, I'll close the PR and reopen it once it's resolved.
Moving forward, if I anticipate a delay, I'll close it first and reopen it when it's ready.

packages/scope-manager/tests/test-utils/serializers/TSESTreeNode.ts
```diff
@@ -17,7 +17,10 @@ const EXCLUDED_KEYS = new Set([
 
 const generator = createIdGenerator();
 type Node = { type: AST_NODE_TYPES } & Record<string, unknown>;
-type Identifier = { name: string; type: AST_NODE_TYPES.Identifier } & Node;
```
mdm317, line 20: { type: AST_NODE_TYPES } is overriden by { name: string; type: AST_NODE_TYPES.Identifier }
Is it better for code readability to disable linting than to remove overridden properties?

JoshuaKGoldberg: 🔪 If we can remove it, let's do it!

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -203,15 +417,18 @@ export default createRule({
     messages: {
       errorTypeOverrides: `'{{typeName}}' is an 'error' type that acts as 'any' and overrides all other types in this {{container}} type.`,
       literalOverridden: `{{literal}} is overridden by {{primitive}} in this union type.`,
+      objectOverridden: `{{subType}} is overridden by {{superType}} in this union type.`,
       overridden: `'{{typeName}}' is overridden by other types in this {{container}} type.`,
       overrides: `'{{typeName}}' overrides all other types in this {{container}} type.`,
       primitiveOverridden: `{{primitive}} is overridden by the {{literal}} in this intersection type.`,
+      superObjectTypeOverridden: `{{superType}} is overridden by {{subType}} in this intersection type.`,
```
JoshuaKGoldberg, line 424: [Feature] This PR has a _lot_ of impressive work around type printing for both `&` and `|` types that goes into the message IDs. But, there's a reason we don't yet have plumbing for such rich type printing:

* It's a lot of work, which we frankly don't have the maintenance budget to do a good job of maintaining
* Large types look really convoluted and confusing in error reports
* For cases where a type must be printed, TypeScript already has a `typeToString` API

This rule was kind of an edge case before in that it only dealt with types that had basic literal or primitive names. Now that it's dealing with objects it'll have to expand.

The `subTypeNode.getText()` / `superTypeNode.getText()` strategy used now also isn't optimal, because:

* It'll include comments, including long JSDoc ones
* It doesn't handle composed types such as interfaces that extend other interfaces

Request: in order of preference...

* Rework these message IDs to not need to print full objects
* Use `typeToString` instead of `.getText()`

packages/eslint-plugin/tests/rules/no-redundant-type-constituents.test.ts
JoshuaKGoldberg, line 1: [Testing] There are going to be quite a few variants of types that could break the rule:

* Conditional types
* Generic types & their type parameters
* Mapped types
* Optional properties
* Interfaces
* Interfaces extending other interfaces
* `keyof` and `typeof` types
* `ReturnType`, `Omit`, etc. types
* Symbols

Etc. - I haven't though super deeply beyond that. Could you think on what other test cases should happen and fill it all out?

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -528,6 +943,33 @@ export default createRule({
...
+            );
+
+            context.report({
+              node: typeNode,
+              messageId: 'objectOverridden',
+              data: {
+                subType: mergedSubTypeName,
+                superType: superTypeName,
+              },
+            });
```
JoshuaKGoldberg, line 970: [Refactor] The two added `for (const [...` -> `context.report()` areas have very similar code around them. I was able to refactor to a shared `reportOverriddenTypes` function like:

```ts
function reportOverriddenTypes(
  overriddenObjectTypes: Map<TSESTree.TypeNode, TypeRelation[]>,
  messageId: MessageId,
  operator: '&' | '|',
) {
  for (...) {
```

WDYT?

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -203,15 +417,18 @@ export default createRule({
     messages: {
       errorTypeOverrides: `'{{typeName}}' is an 'error' type that acts as 'any' and overrides all other types in this {{container}} type.`,
       literalOverridden: `{{literal}} is overridden by {{primitive}} in this union type.`,
+      objectOverridden: `{{subType}} is overridden by {{superType}} in this union type.`,
       overridden: `'{{typeName}}' is overridden by other types in this {{container}} type.`,
       overrides: `'{{typeName}}' overrides all other types in this {{container}} type.`,
       primitiveOverridden: `{{primitive}} is overridden by the {{literal}} in this intersection type.`,
```
JoshuaKGoldberg, line 423: [Feature] Is there a need to keep differentiating between literal and primitive types? I would think the new logic should still apply to them, right?

Suggestion: have you considered unifying the existing & new logic into one area of checks? It'd have to be careful not to try to compare keys of literals or primitives, but otherwise I think it should all be roughly the same, right?

Thinking out loud: I don't think there's a strong need to keep differentiating between `intersection` and `union` in the message either...

JoshuaKGoldberg (changes requested):
OK! Sorry for the delay in reviewing. This is a really tricky one 😅.

I think the feature request is a lot trickier than the test cases here currently capture. It'll have to handle a lot of cases, including optional properties, interfaces extending each other, and so on. I suspect the core implementation will have to adjust a bunch - so left a few suggestions on how to reduce the amount of work done. Hope it's helpful, and let me know if you have questions! 🙌

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
JoshuaKGoldberg: > the result was just `F` instead of the expected `{ a: 1 }`.

If this is just for printing types, then I would think `F` is preferable, no?

mdm317: Yes i agree with you.
Just comfirming
```
type Foo = { a : number }
type Bar = { a : 1 } | Foo
```
In this case error message will be `{ a : 1 } is overriden by Foo`

JoshuaKGoldberg: Yup I believe so!

mdm317:
@JoshuaKGoldberg 

Hello I have a question about applying rules to our code.
It seems that in our codebase, we are intentionally using overridden types, like in the code below.

> https://github.com/typescript-eslint/typescript-eslint/blob/052b7f9754d81056ff81896afd3fabd132796d42/packages/ast-spec/src/declaration/ClassDeclaration/spec.ts#L15-L31

ClassDeclarationWithName  is ovveridden by ClassDeclarationWithOptionalName in this union type
  

> https://github.com/typescript-eslint/typescript-eslint/blob/052b7f9754d81056ff81896afd3fabd132796d42/packages/ast-spec/src/unions/ExportDeclaration.ts#L23-L24 
 
FunctionDeclarationWithOptionalName  is ovveridden by FunctionDeclarationWithName in this union type

In this case, should I remove them, use eslint-disable, or modify the rule itself?

JoshuaKGoldberg:
Ha, that's really interesting. Nice find. My instinct is that we should modify the `spec.ts` files because the union types really are violating the rule - cc @typescript-eslint/triage-team

packages/typescript-estree/src/parser-options.ts
```diff
@@ -244,6 +244,7 @@ export interface ParserServicesBase {
 }
 export interface ParserServicesNodeMaps {
   esTreeNodeToTSNodeMap: ParserWeakMapESTreeToTSNode;
+  // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
   tsNodeToESTreeNodeMap: ParserWeakMap<TSNode | TSToken, TSESTree.Node>;
```
mdm317, line 248: In `TSNode `'s union types, the code below is overridden. I think modifying it too complex, so I added a disable code.
> https://github.com/typescript-eslint/typescript-eslint/blob/fa265a970535e9b8d3ac8bfde9f3a24f1a1ec94a/packages/typescript-estree/src/ts-estree/ts-nodes.ts#L141

packages/eslint-plugin/src/rules/adjacent-overload-signatures.ts
```diff
@@ -49,6 +49,7 @@ export default createRule({
      * @returns the name and attribute of the member or null if it's a member not relevant to the rule.
      */
     function getMemberMethod(
+      // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
       member: Member | MemberDeclaration,
```
mdm317, line 53: Member -> ProgramStatement -> Statement -> ClassDeclarationWithName
MemberDeclaration -> DefaultExportDeclarations -> ClassDeclarationWithOptionalName
`ClassDeclarationWithName` is overriden by `ClassDeclarationWithOptionalName`
I think modifying it too complex, so I added a disable code.

mdm317 (commented):
This change was made to applying lint to code.
Let me know if you have any concerns about this adjustment

packages/eslint-plugin/tests/docs-eslint-output-snapshots/no-redundant-type-constituents.shot
```diff
@@ -22,11 +22,11 @@ type IntersectionNever = string | never;
                                   ~~~~~ 'never' is overridden by other types in this union type.
 
 type IntersectionBooleanLiteral = boolean & false;
-                                  ~~~~~~~ boolean is overridden by the false in this intersection type.
+                                  ~~~~~~~ boolean is overridden by false in this intersection type.
```
mdm317, line 25: Can we remove `the`?

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,257 @@ function addToMapGroup<Key, Value>(
...
+  return groups;
 }
 
-function isNodeInsideReturnType(node: TSESTree.TSUnionType): boolean {
-  return !!(
-    node.parent.type === AST_NODE_TYPES.TSTypeAnnotation &&
-    isFunctionOrFunctionType(node.parent.parent)
-  );
+function hasTargetOnlyOptionalProps(
+  sourceType: ts.ObjectType,
```
mdm317, line 205: Checkint target type contains no extra optional properties beyond sourceType

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,257 @@ function addToMapGroup<Key, Value>(
...
-  if (tsutils.isTrueLiteralType(type)) {
-    return 'true';
+  if (tsutils.isObjectType(sourceType) && tsutils.isObjectType(targetType)) {
+    const sourceTypeProperties = sourceType.getProperties();
+    const targetTypeProperties = targetType.getProperties();
+    if (
+      (sourceTypeProperties.length === 0) !==
+      (targetTypeProperties.length === 0)
+    ) {
+      return false;
```
mdm317, line 137: target type is non redundant
| object type | 0 property |0> property |
|:---|:---:|---:|
|  0 property | check assignable | non redundant |
| 0 > property | non redundant | check assignable |

mdm317 (commented):
I'd appreciate your feedback

packages/ast-spec/src/declaration/FunctionDeclaration/spec.ts
```diff
@@ -32,5 +32,5 @@ export interface FunctionDeclarationWithOptionalName
 }
 
 export type FunctionDeclaration =
-  | FunctionDeclarationWithName
-  | FunctionDeclarationWithOptionalName;
+  // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
+  FunctionDeclarationWithName | FunctionDeclarationWithOptionalName;
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Alternately, we can simplify this a bit by having `FunctionDeclaration` be the normal interface and `FunctionDeclarationWithName` extend it. I'll go ahead and switch it to that.

packages/typescript-estree/src/parser-options.ts
JoshuaKGoldberg: Well, if `TSNode` is overridden by `TSToken`, that means we never needed the `TSNode |` to begin with! 🚀

packages/eslint-plugin/src/rules/adjacent-overload-signatures.ts
JoshuaKGoldberg: 🤔 this feels like a bug in the rule's implementation, no? If `Member` truly is overridden by `MemberDeclaration`, then we should be able to remove the `Member |`. But trying that results in type errors.

packages/eslint-plugin/tests/docs-eslint-output-snapshots/no-redundant-type-constituents.shot
JoshuaKGoldberg: If you want 🙂 good idea! It's a nice-to-have, up to you.

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -202,83 +308,118 @@ export default createRule({
...
+      return [
+        {
+          type,
+          typeName: checker.typeToString(type),
+        },
+      ];
     }
 
     return {
       'TSIntersectionType:exit'(node: TSESTree.TSIntersectionType): void {
```
JoshuaKGoldberg: [Refactor] The rule no longer cares about when it visits the node. We can simplify the visitors to not have `:exit`:

```suggestion
      TSIntersectionType(node: TSESTree.TSIntersectionType): void {
```

(this one and the other too)

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -202,83 +308,118 @@ export default createRule({
...
+        {
+          type,
+          typeName: checker.typeToString(type),
+        },
+      ];
+    }
 
-      if (typeNode.type === AST_NODE_TYPES.TSUnionType) {
-        return typeNode.types.flatMap(getTypeNodeTypePartFlags);
+    function getInterSectionTypePart(
```
JoshuaKGoldberg: [Naming] The _S_ isn't the start of a new word:

```suggestion
    function getIntersectionTypePart(
```

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,257 @@ function addToMapGroup<Key, Value>(
...
-  if (isTypeTemplateLiteralType(type)) {
-    return 'template literal type';
+function isTargetTypeRedundantInIntersection(
+  sourceType: ts.Type,
+  targetType: ts.Type,
+  checker: ts.TypeChecker,
+): boolean {
+  if (
+    tsutils.isUnionType(sourceType) &&
+    !tsutils.isIntrinsicBooleanType(sourceType)
```
JoshuaKGoldberg: [Refactor] Removing this doesn't fail any unit tests. The subsequent `if` statement's same boolean check is needed for some unit tests, just not this one that I can tell?

```suggestion
    tsutils.isUnionType(sourceType)
```

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,257 @@ function addToMapGroup<Key, Value>(
...
+  }
+  return groups;
 }
 
-function isNodeInsideReturnType(node: TSESTree.TSUnionType): boolean {
-  return (
-    node.parent.type === AST_NODE_TYPES.TSTypeAnnotation &&
-    isFunctionOrFunctionType(node.parent.parent)
-  );
+function hasTargetOnlyOptionalProps(
```
JoshuaKGoldberg: [Naming] This function name confused me. I _think_ this is what it's actually doing? Did I get it right?

```suggestion
function targetOnlyAddsOptionalProperties(
```

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,257 @@ function addToMapGroup<Key, Value>(
...
+    );
+
+    if (!sourceProp) {
+      if (targetProp.flags & ts.SymbolFlags.Optional) {
+        return true;
+      }
+      continue;
+    }
+    const sourcePropertyType = checker.getTypeOfSymbol(sourceProp);
+    const targetPropertyType = checker.getTypeOfSymbol(targetProp);
```
JoshuaKGoldberg: [Performance] It's a bit more work, but we generally try to limit type system checks until they're definitely needed.

```suggestion
    const sourcePropertyType = checker.getTypeOfSymbol(sourceProp);
    if (!tsutils.isObjectType(sourcePropertyType)) {
      continue;
    }

    const targetPropertyType = checker.getTypeOfSymbol(targetProp);
    if (!tsutils.isObjectType(targetPropertyType)) {
      continue;
    }
```

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,257 @@ function addToMapGroup<Key, Value>(
   }
 }
 
-function describeLiteralType(type: ts.Type): string {
-  if (type.isStringLiteral()) {
-    return JSON.stringify(type.value);
-  }
+function isNodeInsideReturnType(node: TSESTree.TSUnionType): boolean {
```
JoshuaKGoldberg: [Bug] This function doesn't account for parenthesized types. This code trips it up:

```ts
declare function fn2(): (never | 'foo') | 123;
```

More nesting in types would also break it.

https://deploy-preview-10744--typescript-eslint.netlify.app/play/#ts=5.9.2&fileType=.ts&code=CYUwxgNghgTiAEAzArgOzAFwJYHtVNQEYAKASgC55UQA3EGeAH3gHJEccWAoUSWBFOmx4CAJjKVi1Og2ZsOLUk3iFRAZgDcQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1ukQBNYmvAIYtaRUrTIcmyfJXyxELVBkiJo0DtEjgwAXxB6gA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,257 @@ function addToMapGroup<Key, Value>(
...
+      if (targetProp.flags & ts.SymbolFlags.Optional) {
+        return true;
+      }
+      continue;
+    }
+    const sourcePropertyType = checker.getTypeOfSymbol(sourceProp);
+    const targetPropertyType = checker.getTypeOfSymbol(targetProp);
+    if (
+      tsutils.isObjectType(sourcePropertyType) &&
+      tsutils.isObjectType(targetPropertyType)
```
JoshuaKGoldberg: [Testing] No tests fail if either of these are removed. Type checking complains, but switching the two `*Type` parameters of `hasTargetOnlyOptionProps` to `ts.Type` fixes that too. What was this trying to check for?

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
mdm317: Yes, I think it’s fine to remove this.

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
mdm317: ~~The function name you suggested is correct.~~
 - ~~When the target is { a: 1 } and the source is { a: 1, b: 1 }, the source is assignable to the target~~
 - ~~However, when the target is { a: 1, c?: 1 } and the source is { a: 1, b: 1 }, the source is still assignable, but the target adds optional properties. In this case, the target is not redundant.~~
 
~~This function checks whether the target type only adds optional properties.~~

I removed this function

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
mdm317: > https://github.com/typescript-eslint/typescript-eslint/pull/10744/commits/01d3116d4a85afd6ff978d793c6a2742d41a3978

I fixed it here

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
mdm317: I removed this function

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,276 @@ function addToMapGroup<Key, Value>(
...
 
-  if (type.isLiteral()) {
-    // eslint-disable-next-line @typescript-eslint/no-base-to-string
-    return type.value.toString();
+function shouldCheckTypeRedundancy(
+  type: ts.Type,
+  checker: ts.TypeChecker,
+  depth = 0,
+): boolean {
+  if (depth > 10) {
```
mdm317, line 67: Skip check if depth > 10 (assume self-reference)
```
      type Node = {
        value: string;
        next?: Node;
      };
```

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,276 @@ function addToMapGroup<Key, Value>(
...
+      return false;
+    }
+    const declarations = symbol.getDeclarations();
+    const declaration = declarations?.[0];
+    if (!declaration) {
+      return false;
+    }
+    if (
+      declaration.kind !== ts.SyntaxKind.TypeLiteral &&
+      declaration.kind !== ts.SyntaxKind.InterfaceDeclaration &&
```
mdm317, line 85: Skip check if type is `function` or `class`

mdm317 (commented):
Sorry for late. 😔
This code was quite old even for me.
It took some time to review it again and fix a few issues

fixes:
 - Handle union property types correctly.
   - I used isTypeAssignableTo when the object’s property was a union type, but the result wasn’t redundant, so I fixed it.
```
type T = { a: 3 | { a: 1, b: 1 } } | { a: 3 | { a: 1 } }
               ^~~~~~~~~~~~ should not report
```
  - Added a check to skip self-referencing types (depth > 10) to prevent a maximum call stack error.
  - Changed it to only check Object types that are arrays, plain objects, or mapped types.
  - Don't split union types when dealing with reference types.
    - [comment](https://github.com/typescript-eslint/typescript-eslint/pull/10744#discussion_r2399117770)
  - Skip mapped types with no properties.
    - Not sure if it's redundant when `isTypeAssignableTo` returns true.
    
### limitation
Currently, the logic checks redundant types based on their properties, so types like functions — which have no properties — aren’t properly handled.
Therefore, the check is now limited to object, array, and mapped types only.

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
```diff
@@ -87,107 +43,395 @@ function addToMapGroup<Key, Value>(
...
+        }
+      }
+      return true;
+    }
+    return false;
   }
+  if (tsutils.isObjectType(sourceType) && tsutils.isObjectType(targetType)) {
+    const sourceProps = sourceType.getProperties();
+    const targetProps = targetType.getProperties();
+    if (targetProps.length === 0) {
```
mdm317, line 227: If the object type has no properties, such as a mapped type like Record, skip the redundant type check.

JoshuaKGoldberg (approved):
🥳 Whoohoo! Thanks so much for bearing with me on reviews and making so many refactors. This is a lot of code by necessity, but it's roughly as clean as I think it can get. Nice job!

Just to be safe I'd like to get another review from someone on @typescript-eslint/triage-team if possible (though we are swamped, and I don't want to wait beyond the end of this month).

packages/eslint-plugin/src/rules/no-redundant-type-constituents.ts
bradzacher: another way to do this is to keep track of "seen" types.
If the type is already in the set then it's recursive.

JoshuaKGoldberg:
This change is very exciting, but has a lot more edge cases and implications than we realized 😬. Reverted in #11812 - which has a list of issues. #7742 is now open again with a note.

PR #10755, opened 2025-02-01
fix: fix behavior of `config-helper.ts` when base config contains only meta keys

### LINKED ISSUE #10754: Bug: `config-helper.ts` has unexpected behavior when base config contains only meta keys

### Relevant Package

typescript-eslint

There are two bugs related to this issue: one meaningful and one mostly inconsequential.

---
### Playground Link

Reproduction with [StackBlitz](https://stackblitz.com/edit/stackblitz-starters-hld1icr5?file=eslint.config.mjs).

### ESLint Config

```javascript
import eslint from '@eslint/js';

  eslint.configs.recommended,
  tseslint.configs.recommended,
  {
    extends: [tseslint.configs.disableTypeChecked],
    ignores: ['**/*.ts', '**/*.tsx'],
  }
);
```

### Expected Result

I expect that the `ignores` key only apply to the extend config, or at the very least not create a global ignores that affects the entire ESLint config.

In other words, I expect the resulting configuration array to be equivalent to this:

```js
import eslint from '@eslint/js';

export default [
  eslint.configs.recommended,
  ...tseslint.configs.recommended,
  {
    ignores: ['**/*.ts', '**/*.tsx'],
    ...tseslint.configs.disableTypeChecked,
  }
];
```

### Actual Result

If you look into the config inspector, the resulting config is the following:

```
> ...eslint/defaults
> eslint.configs.recommended
> ...tseslint.configs.recommended
> typescript-eslint/disable-type-checked
  - Ignore: ["**/*.ts", "**.tsx"]
  - ...tseslint.configs.disable-type-checked
> anonymous #10
  - Ignore files globally: ["**/**.ts", "**/*.tsx"]
```

In other words, the equivalent config array is

```js
import eslint from '@eslint/js';

export default [
  eslint.configs.recommended,
  ...tseslint.configs.recommended,
  {
    ignores: ['**/*.ts', '**/*.tsx'],
    ...tseslint.configs.disableTypeChecked,
  },
  // Global ignores, ignores all .ts and .tsx files
  { ignores: ['**/*.ts', '**/*.tsx'] },
];
```

Extends will create a configuration with only the `ignores` key, which is parsed as a global ignores. Then, `.ts` and `.tsx` will not be linted.

This can be seen in the reproduction where `src/index.ts` is not linted as-is, while removing the `extends` config correctly allows it to be linted.

--- 
### Playground Link

Reproduction with [StackBlitz](https://stackblitz.com/edit/stackblitz-starters-b8bczhnh?file=package.json)

### ESLint Config

```javascript
import eslint from '@eslint/js';

  eslint.configs.recommended,
  tseslint.configs.recommended,
  {
    extends: [tseslint.configs.disableTypeChecked],
    files: ['file'],
    ignores: ['ignored'],
    name: 'my-config',
  }
);
```

### Expected Result

Per ESLint’s documentation ([Debug your Configuration#Print a File’s Calculated Configuration](https://eslint.org/docs/latest/use/configure/debug#print-a-files-calculated-configuration)), "You won’t see any entries for files, ignores, or name, because those are only used in calculating the final configuration and so do not appear in the result[ing configuration]."

I expect that if there are no meaningful keys for configuration and I only want to extend those configs with those keys, it should not create an unnecessary configuration file, and only updates the configs accordingly.

The ESLint configuration array should look like:

```js
import eslint from '@eslint/js';

export default [
  eslint.configs.recommended,
  ...tseslint.configs.recommended,
  {
    files: ['file'],
    ignores: ['ignored'],
    name: 'my-config',
    ...tseslint.configs.disableTypeChecked,
  }
];
```

### Actual Result

If you look into the config inspector, the resulting config is the following:

```
> ...eslint/defaults
> eslint.configs.recommended
> ...tseslint.configs.recommended
> my-config_typescript-eslint/disable-type-checked
  - Applies to files matching: ["file"]
  - Ignore: ["ignored"]
  - ...tseslint.configs.disable-type-checked
> my-config
  - Applies to files matching: ["file"]
  - Ignore: ["ignored"]
```

Or equivalently

```js
import eslint from '@eslint/js';

export default [
  eslint.configs.recommended,
  ...tseslint.configs.recommended,
  {
    name: "my-config_typescript-eslint/disable-type-checked",
    files: ["file"],
    ignore: ["ignored"],
    ...tseslint.configs.disableTypeChecked,
  },
  // This config doesn't do anything
  { 
    name: "my-config",
    files: ["file"],
    ignore: ["ignored"],
  },
];
```

The final config in the config array named `my-config` doesn't do anything and is unnecessary, and may make debugging harder in more complex ESLint configs.

---

### Versions

| package                                |  version  |
| -------------------------------------- | --------- |
| `@typescript-eslint`                   | `8.22.0`  |
| `TypeScript`                           | `5.7.3`   |
| `ESLint`                               | `9.19.0`  |

jeremy-code:
- `config-helper.ts` only adds base config to config array if it contains keys other than `files`, `ignores`, `names`.
- Fixes bug where config helper would create global ignores when extending with base config of `ignores`
- Fixes bug where config helper creates extraneous config when extending with base config of only `files`, `ignores`, `names`

## PR Checklist

- [x] Addresses an existing open issue: fixes #10754

## Overview

Checks if base config only contains the keys `files`, `ignores`, or `names`. If so, do not add it to the config array by itself, and only use it for extending. Otherwise, add it to the config array.

### DIFF
--- a/packages/typescript-eslint/src/config-helper.ts
+++ b/packages/typescript-eslint/src/config-helper.ts
@@ -138,7 +138,11 @@ export function config(
           ...(name && { name }),
         };
       }),
-      config,
+      ...(Object.keys(config).every(key =>
+        ['name', 'files', 'ignores'].includes(key),
+      )
+        ? []
+        : [config]),
     ];
   });
 }
--- a/packages/typescript-eslint/tests/config-helper.test.ts
+++ b/packages/typescript-eslint/tests/config-helper.test.ts
@@ -247,4 +247,44 @@ describe('config helper', () => {
       { rules: { rule: 'error' } },
     ]);
   });
+
+  it('does not create global ignores in extends', () => {
+    expect(
+      plugin.config({
+        extends: [{ rules: { rule1: 'error' } }, { rules: { rule2: 'error' } }],
+        ignores: ['ignored'],
+      }),
+    ).toEqual([
+      { rules: { rule1: 'error' }, ignores: ['ignored'] },
+      { rules: { rule2: 'error' }, ignores: ['ignored'] },
+      // Should not create global ignores
+      // { ignores: ['ignored'] },
+    ]);
+  });
+
+  it('does not create extraneous config in extends', () => {
+    expect(
+      plugin.config({
+        extends: [{ rules: { rule1: 'error' } }, { rules: { rule2: 'error' } }],
+        files: ['file'],
+        ignores: ['ignored'],
+        name: 'my-config',
+      }),
+    ).toEqual([
+      {
+        rules: { rule1: 'error' },
+        files: ['file'],
+        ignores: ['ignored'],
+        name: 'my-config',
+      },
+      {
+        rules: { rule2: 'error' },
+        files: ['file'],
+        ignores: ['ignored'],
+        name: 'my-config',
+      },
+      // Should not create configuration object with no keys
+      // { name: "my-config", files: ["file"], ignores: ["ignored"] },
+    ]);
+  });
 });


packages/typescript-eslint/src/config-helper.ts
```diff
@@ -138,7 +138,11 @@ export function config(
           ...(name && { name }),
         };
       }),
-      config,
+      ...(Object.keys(config).every(key =>
```
kirkwaiblinger, line 141: This is rather confusing, so maybe deserves a comment - maybe something like this?
```ts
// The resulting config objects should include the effect of the base config object 
// with the `extends` key removed so we forward it along as the last object, 
// unless it would be a noop or global ignore
```

Feel free to word this better than I have 🙂

packages/typescript-eslint/tests/config-helper.test.ts
kirkwaiblinger, line 1: I'm thinking we should add logic to handle this type of case too?

```ts
  it('does not create global ignores when extending empty configs', () => {
    expect(
      plugin.config({
        extends: [{ rules: { rule1: 'error' } }, {}],
        ignores: ['ignored'],
      }),
    ).toEqual([
      { ignores: ['ignored'], rules: { rule1: 'error' } },
      // Should not create global ignores
      // { ignores: ['ignored'] },
      // { ignores: ['ignored'] },
    ]);
  });
```

kirkwaiblinger (changes requested):
Couple questions/changes, but this is a good stuff, thanks for sending! 

BTW - you'll want to resolve the lint errors to get CI to pass too, probably `yarn lintFix` will be all you need, since I think it's just complaining about the order of fields in objects.

packages/typescript-eslint/src/config-helper.ts
jeremy-code: I do agree it is confusing, but I can't really think of a more elegant way to express it. 

It's possible it may be more readable if the logic wasn't so condensed into that one array such as:

```ts
const configArray = [...];

if (
  Object.keys(config).every(key => ['name', 'files', 'ignores'].includes(key))
) {
  return configArray;
} else {
  return configArray.concat(config);
}
```

But that may also be too verbose, so not sure.

Does the comment sound alright?

packages/typescript-eslint/tests/config-helper.test.ts
jeremy-code: Is now handled

```ts
...
expect(configWithIgnores).not.toContainEqual(
  // Should not create global ignores
  { ignores: ['ignored'] },
);
...
expect(configWithMetadata).not.toContainEqual(
  // Should not create configuration object with no non-metadata keys (noop config)
  { files: ['file'], ignores: ['ignored'], name: 'my-config' },
);
```

jeremy-code:
@kirkwaiblinger Thank you for reviewing my PR — I made some changes (added comments, tests). Do you have any thoughts on it as is?

packages/typescript-eslint/src/config-helper.ts
kirkwaiblinger: > ... if the logic wasn't so condensed into that one array

Feel free to change this! TBH I actually prefer mutating the array for a case like this.

```ts
const configArray = [...];

if (
  !Object.keys(config).every(key => ['name', 'files', 'ignores'].includes(key))
) {
  configArray.push(config);
} 

return configArray;
```

> Does the comment sound alright? 

My nitpick would be that, as written, 
```text
    // The base config should not be included in the final output if it
    // consists of only metadata keys (i.e. properties that do not affect the
    // final ESLint configuration object)
```
makes it sound like it would be a harmless noop if such a config were included in the output, and we're just doing this cleanliness, but the real reason we're here is that it's actually a _harmful_ operation in the case of global ignores. So if there's a way to convey that 🤷

packages/typescript-eslint/tests/config-helper.test.ts
kirkwaiblinger: I might be misunderstanding, but it doesn't appear that this case has been handled? To be clear, the part that I'm trying to highlight is that _in the `extends` list_, there's an empty config object, which will resolve to a `{ ignores: ['ignored'] }` if we're not careful.

kirkwaiblinger (changes requested):
Some questions - also, btw, we like to ask to not force-push since that makes it harder to figure out what has changed since a review, see [PR dos and donts](https://typescript-eslint.io/contributing/pull-requests) 🙏

kirkwaiblinger:
FYI - eslint core is making their own version of the `config()` helper in https://github.com/eslint/rewrite/pull/152 - it may be helpful to study their (in-progress) implementation and align on it as it relates to global ignores. They do all of the logic in terms of an [`isGlobalIgnores()` function](https://github.com/eslint/rewrite/pull/152/files#diff-a42aaf8268f1b91836794b8203043694f266efcd4bc078af6091e13481d3d694R90-R92) which is _very_ similar to your logic 🙂 (the difference being they don't bother with eliminating `files` noops).

Note also almost identical logic with [excluding the base config object if it's a global ignores](https://github.com/eslint/rewrite/pull/152/files#diff-a42aaf8268f1b91836794b8203043694f266efcd4bc078af6091e13481d3d694R414-R416). Let's just try and align on [the case that the extension is possibly-global-ignores](https://github.com/eslint/rewrite/pull/152/files#diff-a42aaf8268f1b91836794b8203043694f266efcd4bc078af6091e13481d3d694R317-R318).

packages/typescript-eslint/src/config-helper.ts
kirkwaiblinger: Coming from https://github.com/typescript-eslint/typescript-eslint/pull/10755#issuecomment-2673840052, probably best we do this as 

```ts
function isGlobalIgnores(config) {
   Object.keys(config).every(key => ['name', 'ignores'].includes(key)
}

const configArray = [...];

if (
  !isGlobalIgnores(config)
) {
  configArray.push(config);
} 

return configArray;
```

kirkwaiblinger:
👋 Hey @jeremy-code! Just checking in, is this still something you have time for? No worries if not - I just don't want to leave it hanging.

PR #10773, opened 2025-02-04
chore(visitor-keys): migrate from `jest` to `vitest`

aryaemami59:
## PR Checklist


JoshuaKGoldberg:
Marking as blocked on #10582. We'd like to normalize test scripts & coverage calculations before making the jump in individual packages.

packages/visitor-keys/tests/visitor-keys.test.ts
```diff
@@ -19,18 +17,9 @@ const IGNORED_KEYS = new Set([
...
-  for (const key of keys) {
-    if (IGNORED_KEYS.has(key)) {
-      if (types.has(key)) {
-        it(`${key} should not be ignored as it has an AST_NODE_TYPE defined`, () => {
-          expect(true).toBeFalsy();
-        });
-      }
-      continue;
-    }
+  const nonIgnoredKeys = [...keys].filter(key => !IGNORED_KEYS.has(key));
```
43081j, line 20: can we move this to sit next to line 6? since vitest will execute it before it runs the tests anyway (as it isn't in a hook or an `it`)

43081j:
this one looks good to me 👍

packages/visitor-keys/tests/visitor-keys.test.ts
JoshuaKGoldberg, line 1: Hmm, these `for` -> `.for` changes aren't necessary, but I do see that the ESLint plugin is reporting on them. 🤷 there aren't that many and I think it's fine.

knip.ts
```diff
@@ -11,9 +11,14 @@ export default {
     types: 'off',
     unresolved: 'off',
   },
+
+  vite: false,
+
```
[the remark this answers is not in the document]
aryaemami59: It doesn't change the result but since we have `vite` in our `devDependencies`, the vite plugin for `knip` still infers that that this is a `vite` project which is not accurate, on top of that `knip` has to do a little bit less work which makes things a teeny tiny bit faster.

JoshuaKGoldberg (dismissed):
LGTM, thanks! Just the one comment/question on the `vitest` property and I think we're good to merge. 🚀

knip.ts
JoshuaKGoldberg: Okie dokie, makes sense. I think as a followup once all projects are converted it might be good to look into removing both the devDependency and this project. Not a blocker for this PR IMO.

packages/visitor-keys/tests/visitor-keys.test.ts
aryaemami59: Yeah I did it for 2 reasons:

1. To silence the ESLint plugin like you mentioned.
2. If we were ever to run the tests concurrently, things become much faster with `.for`.

knip.ts
aryaemami59: Well, We do need `vite` as a devDependency since We're using `@nx/vite`.

packages/visitor-keys/tests/visitor-keys.test.ts
JoshuaKGoldberg: Oh interesting, does `.for` work with concurrency? I didn't see this in the docs.

aryaemami59: Yeah actually that's one of the reasons it exists in the first place, it's because `.each` didn't have access to the test context therefore did not work well with concurrent tests. 
See [**`test.for` Docs**](https://vitest.dev/api/#test-for).

PR #10778, opened 2025-02-04
chore(integration-tests): migrate from `jest` to `vitest`

aryaemami59:
## PR Checklist


packages/integration-tests/tools/integration-test-base.ts
```diff
@@ -161,11 +172,18 @@ export function eslintIntegrationTest(
...
       .replaceAll(
         new RegExp(`"filePath": ?"(/private)?${testFolder}`, 'g'),
         '"filePath": "<root>',
+      )
+      .replaceAll(/"filePath":"([^"]*)"/g, (_, testFile: string) => {
+        return `"filePath": "<root>/${path.relative(testFolder, testFile)}"`;
+      })
+      .replaceAll(
+        /C:\\\\usr\\\\linked\\\\tsconfig.json/g,
+        path.posix.join('/usr', 'linked', 'tsconfig.json'),
```
43081j, line 186: how come this didn't happen in jest?

packages/integration-tests/tools/integration-test-base.ts
```diff
@@ -97,6 +106,7 @@ function integrationTest(
             ['install', '--no-immutable'],
             {
               cwd: testFolder,
+              shell: true,
```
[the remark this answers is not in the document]
aryaemami59: Tests would fail on Windows otherwise.

packages/integration-tests/tools/integration-test-base.ts
aryaemami59: It does, it's just that we don't run the `integration-tests` on Windows during CI so We just didn't know about it. Without it, the snapshot assertions will fail on Windows.

packages/integration-tests/tools/integration-test-base.ts
```diff
@@ -50,41 +33,51 @@ function integrationTest(
 
     describe(testName, () => {
       it('should work successfully', async () => {
-        const testFolder = await tmpDir({
-          keep: KEEP_INTEGRATION_TEST_DIR,
-        });
-        if (KEEP_INTEGRATION_TEST_DIR) {
-          console.error(testFolder);
-        }
```
aryaemami59, line 58: This was removed because if We set `process.env.KEEP_INTEGRATION_TEST_DIR` to `'true'` then `console-fail-test` will cause the unit-tests to fail.

packages/integration-tests/tools/integration-test-base.ts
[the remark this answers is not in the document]
43081j: they didn't before, right? so something changed in how we run this. what was that?

packages/integration-tests/tools/integration-test-base.ts
43081j: maybe this answers my other question

the reason this and `shell` were changed is because you got these tests to run on windows, and they didn't previously?

aryaemami59: No they actually just failed when run locally, but we didn't know they would fail because We didn't run them on Windows during CI.

packages/integration-tests/tools/integration-test-base.ts
```diff
@@ -1,43 +1,26 @@
...
 
 const FIXTURES_DIR = path.join(__dirname, '..', 'fixtures');
-// an env var to persist the temp folder so that it can be inspected for debugging purposes
-const KEEP_INTEGRATION_TEST_DIR =
-  process.env.KEEP_INTEGRATION_TEST_DIR === 'true';
 
-// make sure that jest doesn't timeout the test
-jest.setTimeout(60000);
+// make sure that vitest doesn't timeout the test
+vi.setConfig({ testTimeout: 60_000 });
```
aryaemami59, line 23: We could just move this into `vitest.config.mts` file.

packages/integration-tests/typings/global.d.ts
aryaemami59, line 1: This is no longer needed.

eslint.config.mjs
```diff
@@ -32,6 +32,7 @@ const restrictNamedDeclarations = {
 const vitestFiles = [
   'packages/eslint-plugin-internal/tests/**/*.test.{ts,tsx,cts,mts}',
   'packages/typescript-eslint/tests/**/*.test.{ts,tsx,cts,mts}',
+  'packages/integration-tests/tests/**/*.test.{ts,tsx,cts,mts},',
```
aryaemami59, line 35: This could probably be simplified to something like this:

```suggestion
  'packages/integration-tests/tests/**/*.test.?(m|c)ts?(x),',
```

packages/integration-tests/vitest.config.mts
```diff
@@ -0,0 +1,30 @@
...
+      globalSetup: ['./tools/pack-packages.ts'],
+      name: packageJson.name.replace('@typescript-eslint/', ''),
+
+      poolOptions: {
+        forks: {
+          singleFork: os.platform() === 'win32',
+        },
+      },
+
+      root: import.meta.dirname,
```
43081j, line 25: note that this is only available in `>=20` of node. does this not currently fail in CI? if not, would be good to understand what is filling it in (since we run CI on 18.x which certainly doesn't have `dirname`)

aryaemami59: Vitest backports it to earlier versions of Node.js.

eslint.config.mjs
JoshuaKGoldberg: I look forward to simplifying all this stuff once the conversion is done 😄

eslint.config.mjs
```diff
@@ -31,12 +31,15 @@ const restrictNamedDeclarations = {
...
+  'packages/integration-tests/tools/integration-test-base.ts',
+  'packages/integration-tests/tools/pack-packages.ts',
   'packages/parser/tests/lib/**/*.test.{ts,tsx,cts,mts}',
   'packages/parser/tests/test-utils/**/*.{ts,tsx,cts,mts}',
-  'packages/utils/tests/**/*.test?(-d).{ts,tsx,cts,mts}',
   'packages/type-utils/tests/**/*.test.{ts,tsx,cts,mts}',
+  'packages/typescript-eslint/tests/**/*.test.{ts,tsx,cts,mts}',
+  'packages/utils/tests/**/*.test?(-d).{ts,tsx,cts,mts}',
+  'packages/visitor-keys/tests/**/*.test.{ts,tsx,cts,mts}',
 ];
```
JoshuaKGoldberg, line 43: I sorted the list 😄

PR #10781, opened 2025-02-04
feat(unified-signatures): support ignoring overload signatures with different JSDoc comments

### LINKED ISSUE #10520: Feature: [unified-signatures] Option to not report when signature has different JSDoc

### Playground Link

https://typescript-eslint.io/play/#ts=5.4.5&showAST=es&fileType=.tsx&code=FAEwpgxgNghgTmABAMwK4DsIBcCWB7dFACgA8AuRdVAWwCMw4BKCjAa3TwHd0BuYAegBUgxAAFwABwQQYWMCESD%2BicNHhI0mXAWLlEtPHihgY6ZojYdufYMiIBGRnztY4qME%2BBA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJygM0sQBNaySgHMmAQ3yxoKdGACM4MAF8QSoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
declare function f(x: number): unknown;
/** @deprecated */ declare function f(x: boolean): unknown;

f(1);
f(true);
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/unified-signatures": 1
  }
}
```

### tsconfig

_No response_

### Expected Result

It should not warn, because the two signatures cannot be unified. Note that the call `f(true)` is striked as "deprecated", and it would not be possible for it to happen if the two function declarations were merged into a single one.

### Actual Result

It reports
> These overloads can be combined into one signature taking `number | boolean`.

### Additional Info

_No response_

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10520

## Overview

This PR attempts to tackle #10520 and adds the `ignoreOverloadsWithDifferentJSDoc` option. When enabled, the following is considered valid:

```ts
/**
 * This signature does something.
 */
declare function f(x: number): void;

/**
 * This signature does something else.
 */
declare function f(x: string): void;
```

Some thoughts/notes:

- I've only considered JSDoc comments that annotate a function signature, not ones that annotate its params (or return type annotation, see below), so the following is reported, even though it has similar issues with unifying two signatures with different jsdoc comments:

  ```ts
  declare function f(x: {
    /** @deprecated */
    a: boolean;
  }): void;
  declare function f(x: { a: number }): void;
  ```
  
  This is tricky, as the following overloads, for example, can be unified (and should be reported):

  ```ts
  declare function f(x: {
    /** @deprecated */
    a: boolean;
  }): void;
  declare function f(x: { b: number }): void;
  ```
  
  I'm a bit unsure if this should be checked or not, as it wasn't described on the issue. I'd love to hear opinions on this, and would be happy to add this check for this (although this seems somewhat complex 🙈).
  
- I didn't add any checks to an annotated return type, as return type annotations are checked [by comparing the source code text](https://github.com/typescript-eslint/typescript-eslint/blob/bcdf56017a1c9d66d701d2104651ae9d9cf035dd/packages/eslint-plugin/src/rules/unified-signatures.ts#L441-L452).
  
  So the following signatures already count as unique ([playground link](https://typescript-eslint.io/play/#ts=5.7.3&fileType=.tsx&code=CYUwxgNghgTiAEAzArgOzAFwJYHtVIAoAPALngCMccIQpUBKMgbwCh54B6AKi-gAFQABzhgoGEMHhcObeFDKVqtVAG4WAXzWhIsBCnTY8hUvFTIAtuRAxG8Vu3mmLVmGs0sgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJygM0sQBNaySgHMmAQ3yxoKdFETRoAe2iRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):
  
  ```ts
  declare function f(x: boolean): {
    /** @deprecated */
    a: boolean;
  };
  declare function f(x: number): {
    a: number;
  };
  ```
  
  ---
  
  Unrelated to the change in the PR: Although this does have some false positives, as the following should be reported but doesn't:
  
  ```ts
  declare function f(x: boolean): {
    a: number;
  };
  declare function f(x: number): { a: number };
   ```

  This also includes random, non-jsdoc comments, spaces, line-breaks, etc ([playground link](https://typescript-eslint.io/play/#ts=5.7.3&showAST=es&fileType=.tsx&code=PTAEFp1ATBLAzeBTATkgdgF1AZwA4CGAxrOgOYA0oANqUuAEZoEDWOVRA9gLbcabtQSTEVDdWSUJgAWklAFdqk9J2xo8nFNkgAoHSFDrNmJNFBcUaIpmoBPHdCRFqBNKHjz012J3TuAFAAeAFygDJycSgToAJShAN6gBKHo8twMqKAAvgDcDk4ubh5emD5%2B8EEpaRkocaCJyaCp6Zm5egY40pyKZhmGSBpapmHy2NCcSDjoAOSY%2Bc6uksXevqBklWERUbEJSVUtKNl5jgtFnit%2B6yFN1ah18TqgezcHOm36YJBJ1DicoADusBkoDwKE4eBwUmkBDGEyms1A0IAbpIZJIcAQ%2BKBNI5DroPv1BiYzBYrDZ7CdCktzqVVtINuFIkhovdns0ajkwqEcJgUKQyEd5lT3DSyoiNuy7rtGpKUJyGNzefzBQTOt1qL05ANjMMGKMYHCZnNKYsRSUxSwGVtmTt6lzcEryJyZbdDm0TWdzatLddZayXQd5Yq%2BeRBUA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJygM0sQBNaySgHMmAQ3yxoKdFETRoAe2iRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)). Should I open a separate issue on this?

### DIFF
--- a/packages/eslint-plugin/src/rules/unified-signatures.ts
+++ b/packages/eslint-plugin/src/rules/unified-signatures.ts
@@ -1,10 +1,15 @@
 import type { TSESTree } from '@typescript-eslint/utils';
 
-import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+import { AST_NODE_TYPES, AST_TOKEN_TYPES } from '@typescript-eslint/utils';
 
 import type { Equal } from '../util';
 
-import { arraysAreEqual, createRule, nullThrows } from '../util';
+import {
+  arraysAreEqual,
+  createRule,
+  LINEBREAK_MATCHER,
+  nullThrows,
+} from '../util';
 
 interface Failure {
   only2: boolean;
@@ -61,6 +66,7 @@ export type MessageIds =
 export type Options = [
   {
     ignoreDifferentlyNamedParameters?: boolean;
+    ignoreOverloadsWithDifferentJSDoc?: boolean;
   },
 ];
 
@@ -91,6 +97,11 @@ export default createRule<Options, MessageIds>({
             description:
               'Whether two parameters with different names at the same index should be considered different even if their types are the same.',
           },
+          ignoreOverloadsWithDifferentJSDoc: {
+            type: 'boolean',
+            description:
+              'Whether two overloads with different JSDoc comments should be considered different even if their parameter and return types are the same.',
+          },
         },
       },
     ],
@@ -98,9 +109,13 @@ export default createRule<Options, MessageIds>({
   defaultOptions: [
     {
       ignoreDifferentlyNamedParameters: false,
+      ignoreOverloadsWithDifferentJSDoc: false,
     },
   ],
-  create(context, [{ ignoreDifferentlyNamedParameters }]) {
+  create(
+    context,
+    [{ ignoreDifferentlyNamedParameters, ignoreOverloadsWithDifferentJSDoc }],
+  ) {
     //----------------------------------------------------------------------
     // Helpers
     //----------------------------------------------------------------------
@@ -227,6 +242,15 @@ export default createRule<Options, MessageIds>({
           ) {
             return false;
           }
+        }
+      }
+
+      if (ignoreOverloadsWithDifferentJSDoc) {
+        const aComment = getJSDocCommentForNode(getExportingNode(a) ?? a);
+        const bComment = getJSDocCommentForNode(getExportingNode(b) ?? b);
+
+        if (aComment?.value !== bComment?.value) {
+          return false;
         }
       }
 
@@ -522,6 +546,18 @@ export default createRule<Options, MessageIds>({
       currentScope = scopes.pop();
     }
 
+    /**
+     * @returns the first valid JSDoc comment annotating `node`
+     */
+    function getJSDocCommentForNode(
+      node: TSESTree.Node,
+    ): TSESTree.Comment | undefined {
+      return context.sourceCode
+        .getCommentsBefore(node)
+        .reverse()
+        .find(comment => isJSDocComment(comment));
+    }
+
     function addOverload(
       signature: OverloadNode,
       key?: string,
@@ -586,7 +622,7 @@ export default createRule<Options, MessageIds>({
 });
 
 function getExportingNode(
-  node: TSESTree.TSDeclareFunction,
+  node: SignatureDefinition,
 ):
   | TSESTree.ExportDefaultDeclaration
   | TSESTree.ExportNamedDeclaration
@@ -633,4 +669,27 @@ function getStaticParameterName(param: TSESTree.Node): string | undefined {
 }
 function isIdentifier(node: TSESTree.Node): node is TSESTree.Identifier {
   return node.type === AST_NODE_TYPES.Identifier;
+}
+
+/**
+ * Checks if a comment is in JSDoc form.
+ *
+ * Based on https://github.com/eslint/eslint/blob/93c325a7a841d0fe4b5bf79efdec832e7c8f805f/lib/rules/multiline-comment-style.js#L104-L119
+ */
+function isJSDocComment(comment: TSESTree.Comment) {
+  if (comment.type !== AST_TOKEN_TYPES.Block) {
+    return false;
+  }
+
+  const lines = comment.value.split(LINEBREAK_MATCHER);
+
+  if (lines.length === 1) {
+    return lines[0].startsWith('*');
+  }
+
+  return (
+    /^\*\s*$/u.test(lines[0]) &&
+    lines.slice(1, -1).every(line => /^\s* /u.test(line)) &&
+    /^\s*$/u.test(lines[lines.length - 1])
+  );
 }
--- a/packages/eslint-plugin/docs/rules/unified-signatures.mdx
+++ b/packages/eslint-plugin/docs/rules/unified-signatures.mdx
@@ -78,6 +78,50 @@ function f(b: string): void;
 </TabItem>
 </Tabs>
 
+### `ignoreOverloadsWithDifferentJSDoc`
+
+{/* insert option description */}
+
+Examples of code for this rule with `ignoreOverloadsWithDifferentJSDoc`:
+
+<Tabs>
+<TabItem value="❌ Incorrect">
+
+```ts option='{ "ignoreOverloadsWithDifferentJSDoc": true }'
+declare function f(x: string): void;
+declare function f(x: boolean): void;
+/**
+ * @deprecate
+ */
+declare function f(x: number): void;
+/**
+ * @deprecate
+ */
+declare function f(x: null): void;
+```
+
+</TabItem>
+<TabItem value="✅ Correct">
+
+```ts option='{ "ignoreOverloadsWithDifferentJSDoc": true }'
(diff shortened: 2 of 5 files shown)


packages/eslint-plugin/src/rules/unified-signatures.ts
```diff
@@ -522,6 +546,18 @@ export default createRule<Options, MessageIds>({
...
+     * @returns the first valid JSDoc comment annotating `node`
+     */
+    function getJSDocCommentForNode(
+      node: TSESTree.Node,
+    ): TSESTree.Comment | undefined {
+      return context.sourceCode
+        .getCommentsBefore(node)
+        .reverse()
+        .find(comment => isJSDocComment(comment));
+    }
```
ronami, line 559: It seems only the first jsdoc comment affects a function signature ([playground link](https://typescript-eslint.io/play/#ts=5.7.3&fileType=.tsx&code=PQKhAIAEBMFMAcBOsDGBDALra4TAFCgSRoDOAngHYq4FwoA2ay4AZgK7UYCWA9pWwAUADwBc4AN7g04yuwC2AI1iJwAXwCU4gG69u0ANz4gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJygM0sQBNaySgHMmAQ3yxoKdFETRoAe2iRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)), and non-jsdoc comments are ignored ([playground link](https://typescript-eslint.io/play/#ts=5.7.3&fileType=.tsx&code=PTAEDMHtIKGAqeoACATApgBwE7oMYCGALuqqPMHGAEYHYwZ4A2d6EArgHZ5ECWknCAAoAHgC5QAb1AEJndgFtq6bKAC%2BASgkA3SL1QBuGEA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJygM0sQBNaySgHMmAQ3yxoKdFETRoAe2iRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)).

LukeAbby:
Not a maintainer so just a random guy's two cents.

Do you think it'd be worth considering a sibling option to NOT suggest unifying two overloads when one is deprecated and the other isn't? This is actually what the original issue's code example was. The issue I see with doing it based off of any JSDoc change is that if you dutifully document the optional parameter (but leave the rest of the JSDoc identical) then it'll be missed.

packages/eslint-plugin/src/rules/unified-signatures.ts
```diff
@@ -634,3 +670,26 @@ function getStaticParameterName(param: TSESTree.Node): string | undefined {
...
+ * Checks if a comment is in JSDoc form.
+ *
+ * Based on https://github.com/eslint/eslint/blob/93c325a7a841d0fe4b5bf79efdec832e7c8f805f/lib/rules/multiline-comment-style.js#L104-L119
+ */
+function isJSDocComment(comment: TSESTree.Comment) {
+  if (comment.type !== AST_TOKEN_TYPES.Block) {
+    return false;
+  }
+
+  const lines = comment.value.split(LINEBREAK_MATCHER);
```
JoshuaKGoldberg, line 684: [Feature] I tried replacing the rest of this function with `return true` and only two tests failed:

```ts
/**
 * This signature does something.
 **/
```

```ts
/*
 * This signature does something.
 */
```

ACK that these are technically not correctly formed for JSDocs, but TypeScript in VS Code still shows them well:

Screenshot of hovering over a 'const a = 1' variable in VS Code and seeing 'This signature does something.' intellisense


Proposal: let's just go off of whether it's a `Block`?

JoshuaKGoldberg (changes requested):
LGTM, another very thorough investigation 🔥. Just requesting changes on simplifying the block-comment/JSDoc checking.

packages/eslint-plugin/src/rules/unified-signatures.ts
ronami: Oh, nice catch!

I could make TypeScript in VSCode to show this in the first example, but I couldn't do it in the second one. Reverse engineering this a bit, it seems the comment has to start with at least two `*'s ([playground link](https://typescript-eslint.io/play#ts=5.7.2&fileType=.tsx&code=PTAEGcFNIW1AXA9hAFog7qdLIDtQBmArrgMbwCWi%2BF4oaAbpAE6QAmAUB8AFQ8egeoACopaECgHNcAQ3hFWoNokh1wiGJHhjckgHQCewDm0ikANjMXEylaoQAUADwBcoXERgAjFgEo3JADWuBi4ANxcIEoq4LgA5PAQ0HBIqBhYOPg25FQ0dIws7JH8giJialKy8orKqhAaWjr6hsamFlaQhCQ59pLObh7efgG4waERQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

"`ts
// seem to show when the function is hovered

/**
 * This signature does something.
 */
declare function f(x: number): unknown;

//doesn't seem to show when the function is hovered

/*
 * This signature does something.
 */
declare function g(x: number): unknown;
```

I agree, though, that matching TypeScript's behavior would probably be pretty complex (and possibly not worth the extra complexity). Also, for comments that aren't recognized by TypeScript (like the second example), I think it may be better not to report them since it's most likely a typo on the developer's part (or at least I assume so).

---

I've updated my PR to match your suggestion 🚀

JoshuaKGoldberg (approved):
Conan O'Brien saying "looks good" happily

PR #10823, opened 2025-02-10
docs: give release time in UTC and local time

kirkwaiblinger:
## PR Checklist

## Overview

We on the maintenance team are spread around the world, so are our users, and our release automations don't respect DST, so let's just use UTC and call it a day?

### DIFF
--- a/docs/users/Releases.mdx
+++ b/docs/users/Releases.mdx
@@ -22,7 +22,7 @@ export function LocalTimeOfRelease() {
     'Friday',
     'Saturday',
   ];
-  if (date.getDay() !== 1) {
+  if (date.getDay() !== date.getUTCDay()) {
     return `${dayNames[date.getDay()]}s at ${formatted}`;
   }
   return formatted;
--- a/docs/users/Releases.mdx
+++ b/docs/users/Releases.mdx
@@ -5,7 +5,7 @@ title: Releases
 ---
 
 export function LocalTimeOfRelease() {
-  // An arbitrary Wednesday at 17:00 UTC.
+  // An arbitrary Monday at 17:00 UTC.
   const date = new Date('1970-01-05T17:00Z');
   const formatted = date.toLocaleTimeString('en-US', {
     hour: 'numeric',
--- a/docs/users/Releases.mdx
+++ b/docs/users/Releases.mdx
@@ -13,7 +13,7 @@ title: Releases
   />
 </a>
 
-We release a latest version every Monday at 1pm US Eastern time using the latest commit to `main` at that time. This release is performed automatically by a Github action located in a private repository. This release goes to the standard `latest` tag on npm.
+We release a latest version every Monday at 17:00 UTC using the latest commit to `main` at that time. This release is performed automatically by a Github action located in a private repository. This release goes to the standard `latest` tag on npm.
 
 See [Versioning](./Versioning.mdx) for how the version number is calculated.
 
--- a/docs/users/Releases.mdx
+++ b/docs/users/Releases.mdx
@@ -9,20 +9,11 @@ export function LocalTimeOfRelease() {
   const now = new Date();
   // your time zone minus UTC time
   const offset = -now.getTimezoneOffset();
-  const formatter = new Intl.DateTimeFormat(undefined, {
+  const formatted = date.toLocaleTimeString('en-US', {
     hour: 'numeric',
     minute: '2-digit',
     timeZoneName: 'short',
   });
-  const parts = formatter.formatToParts(date);
-  const formatted = parts
-    .map(part => {
-      if (part.type === 'dayPeriod') {
-        return part.value.toLowerCase();
-      }
-      return part.value;
-    })
-    .join('');
   if (offset >= (24 - 17) * 60) {
     return 'Tuesdays at ' + formatted;
   }
--- a/docs/users/Releases.mdx
+++ b/docs/users/Releases.mdx
@@ -4,13 +4,17 @@ sidebar_label: Releases
 title: Releases
 ---
 
+import useIsBrowser from '@docusaurus/useIsBrowser';
+
 export function LocalTimeOfRelease() {
+  const isBrowser = useIsBrowser();
   // An arbitrary Monday at 17:00 UTC.
   const date = new Date('1970-01-05T17:00Z');
   const formatted = date.toLocaleTimeString('en-US', {
     hour: 'numeric',
     minute: '2-digit',
     timeZoneName: 'short',
+    timeZone: isBrowser ? undefined : 'UTC',
   });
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (dismissed):
omg yes thank you I keep forgetting about this 🤦

docs/users/Releases.mdx
```diff
@@ -13,7 +13,7 @@ title: Releases
   />
 </a>
 
-We release a latest version every Monday at 1pm US Eastern time using the latest commit to `main` at that time. This release is performed automatically by a Github action located in a private repository. This release goes to the standard `latest` tag on npm.
+We release a latest version every Monday at 17:00 UTC using the latest commit to `main` at that time. This release is performed automatically by a Github action located in a private repository. This release goes to the standard `latest` tag on npm.
```
bradzacher, line 16: Can we just use some react code here to render the time in the browser's local time to avoid making a distinction?

kirkwaiblinger: i'm the type of weirdo who would rather just see the UTC for something like this but I don't mind if we do both, like "17:00 UTC (10:00 am in your current time zone)" or something like that

Josh-Cena: Yeah, I like that idea.

bradzacher: An option the internal tools did at FB was to render local time and provide a popover which would show UTC and the original timezone on hover.

kirkwaiblinger: i went a little overboard in order to make sure that Brad knows that it release on Tuesdays for him not on Monday 😁 

Also to make sure that "am" is not capitalized because I hate that.

docs/users/Releases.mdx
```diff
@@ -4,6 +4,34 @@ sidebar_label: Releases
 title: Releases
 ---
 
+export function LocalTimeOfRelease() {
+  const date = new Date('1970-01-01T17:00Z');
+  const now = new Date();
+  // your time zone minus UTC time
+  const offset = -now.getTimezoneOffset();
+  const formatter = new Intl.DateTimeFormat(undefined, {
```
Josh-Cena, line 12: Use `"en-US"`; our docs are `en-US` and you don't want to show times in other languages.

docs/users/Releases.mdx
```diff
@@ -4,6 +4,34 @@ sidebar_label: Releases
...
+    hour: 'numeric',
+    minute: '2-digit',
+    timeZoneName: 'short',
+  });
+  const parts = formatter.formatToParts(date);
+  const formatted = parts
+    .map(part => {
+      if (part.type === 'dayPeriod') {
+        return part.value.toLowerCase();
+      }
```
Josh-Cena, line 22: Why this? I think uppercase is fine if it's produced by default.

docs/users/Releases.mdx
```diff
@@ -4,6 +4,34 @@ sidebar_label: Releases
 title: Releases
 ---
 
+export function LocalTimeOfRelease() {
+  const date = new Date('1970-01-01T17:00Z');
```
Josh-Cena, line 8: To avoid the offset checking hack, you can just start with a date you know is Monday, for example `1970-01-05`, and then enable `weekday` formatting for the `DateTimeFormat`

docs/users/Releases.mdx
kirkwaiblinger: simply because I personally find the majuscule version hideous. But if you want we can use it 🙃 That said if we pin everyone to `en-US` I think it makes even more sense to force it to be lowercase just for stylistic reasons, since we won't accidentally overwrite some different specifier format (if one exists)

docs/users/Releases.mdx
kirkwaiblinger: True - the thing that seemed kind of nice was that the time zones aren't just GMT+/-X for other places, but yeah sticking with en-US probably does make more sense

docs/users/Releases.mdx
kirkwaiblinger: Hmmm - my goal was to _not_ show the weekday if it is still Mondays though. Seemed less hacky to do this than to try to hack around with the output array from `formatToParts()`... but maybe there is a better way still?

docs/users/Releases.mdx
Josh-Cena: It just depends on whether there's data for it. For example you can get `CET`/`CEST` too.

docs/users/Releases.mdx
kirkwaiblinger: Came up with a way of doing this that seems much better of checking whether the `date.getDay() === date.getUTCDay()`

docs/users/Releases.mdx
```diff
@@ -4,6 +4,30 @@ sidebar_label: Releases
 title: Releases
 ---
 
+export function LocalTimeOfRelease() {
+  // An arbitrary Monday at 17:00 UTC.
+  const date = new Date('1970-01-05T17:00Z');
+  const formatted = date.toLocaleTimeString('en-US', {
+    hour: 'numeric',
+    minute: '2-digit',
+    timeZoneName: 'short',
```
Josh-Cena, line 13: Why not:

```suggestion
    timeZoneName: 'short',
    weekday: date.getDay() !== date.getUTCDay() ? 'long' : undefined,
```

docs/users/Releases.mdx
```diff
@@ -4,6 +4,30 @@ sidebar_label: Releases
 title: Releases
 ---
 
+export function LocalTimeOfRelease() {
```
Josh-Cena, line 7: PSA: this is a somewhat antipattern because there's a server-client hydration mismatch. You may want to do this:

```js
import { useIsBrowser } from "@docusaurus/useIsBrowser";

function LocalTimeOfRelease() {
  const isBrowser = useIsBrowser();
  const formatted = date.toLocaleTimeString('en-US', {
    // ... options
    timeZone: isBrowser ? undefined : "UTC",
  })
```

kirkwaiblinger: I trust you on this 🤷‍♂️

docs/users/Releases.mdx
kirkwaiblinger: To me it flows much more naturally to say "We release a latest version every Monday at 17:00 UTC (Tuesdays at 2:00 AM GMT+9)" than "We release a latest version every Monday at 17:00 UTC (Tuesday, 2:00 AM GMT+9)"

kirkwaiblinger: Trying to put my words into code rather than let the code's default output speak for me haha

Josh-Cena (approved):

bradzacher (approved):

ronami (approved):

PR #10849, opened 2025-02-19
feat(typescript-eslint): export util types (close #10848)

### LINKED ISSUE #10848: Enhancement: export types in typescript-eslint package

### Relevant Package

typescript-eslint

### Description

I believe there are a lot of people making there own wrapper after eslint itself as support flat config.

In this way we are likely to need the access of TSLint types from `@typescript-eslint/utils`. It is imported in `typescript-eslint` package, but not exported.

Exporting this types can benefit users, so that they no longer need to declare an extra package (AKA `@typescript-eslint/utils`) .

E.g.: 

https://github.com/Mister-Hope/eslint-config-mister-hope/blob/497fba4a0af43a53ccc78c36899090f4d670d5c4/src/ts.ts#L9

code example

### Additional Info

_No response_

Mister-Hope:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10848

## Overview

This adds export for a namespace `TSLint` which contain types used in tslint.

The main reason is that as the documentation suggests using tslint stricter types: https://typescript-eslint.io/getting-started#step-2-configuration, people may need to implicitly mark codes to these types in their own flag config, as auto-infer may contain "loose types" (e.g.: `string` instead of `'a' | 'b'`) that triggers ts errors.

For examples,see #10848

### DIFF
--- a/packages/typescript-eslint/src/index.ts
+++ b/packages/typescript-eslint/src/index.ts
@@ -16,7 +16,7 @@ import type {
 import { config } from './config-helper';
 import { getTSConfigRootDirFromStack } from './getTSConfigRootDirFromStack';
 
-export type { TSESLint } from '@typescript-eslint/utils';
+export type { FlatConfig } from '@typescript-eslint/utils/ts-eslint';
 
 export const parser: CompatibleParser =
   rawPlugin.parser as CompatibleParser satisfies FlatConfig.Parser;
--- a/docs/packages/TypeScript_ESLint.mdx
+++ b/docs/packages/TypeScript_ESLint.mdx
@@ -22,6 +22,7 @@ This package exports the following:
 | `configs`             | [Shared ESLint (flat) configs](../users/Shared_Configurations.mdx)                                |
 | `parser`              | A re-export of [`@typescript-eslint/parser`](./Parser.mdx)                                        |
 | `plugin`              | A re-export of [`@typescript-eslint/eslint-plugin`](./ESLint_Plugin.mdx)                          |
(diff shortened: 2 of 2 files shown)


kirkwaiblinger:
Noting that this issue is now accepting PRs, so we can unblock this PR. 

(thanks to @ntdiary for bringing this to our attention in https://github.com/typescript-eslint/typescript-eslint/issues/10893#issuecomment-2919686555)

packages/typescript-eslint/src/index.ts
```diff
@@ -16,6 +16,8 @@ import type {
 import { config } from './config-helper';
 import { getTSConfigRootDirFromStack } from './getTSConfigRootDirFromStack';
 
+export type { TSESLint } from '@typescript-eslint/utils';
```
JoshuaKGoldberg: From https://github.com/typescript-eslint/typescript-eslint/issues/10848#issuecomment-2767600291 & https://github.com/typescript-eslint/typescript-eslint/issues/10848#issuecomment-2789340850:

> I think we should just export `FlatConfig` and keep it limited to that.

> Accepting PRs on just exporting `FlatConfig`. If more use cases come up we can consider them individually. 🚀

I asked for clarification in https://github.com/typescript-eslint/typescript-eslint/issues/10848#issuecomment-3347070366 on whether there should be more.

Mister-Hope: I reply in that thread. In short types here are stricter that eslint config type, otherwise you don't even need to repeat writing a whole config interface here.

Mister-Hope: Oh I am fine with this part, exporting flatconfig only currently

JoshuaKGoldberg (approved):
Great, thanks!

PR #10870, opened 2025-02-22
chore: fix failing integration tests on the `main` branch

ronami:
## PR Checklist

## Overview

The `main` branch's integration tests are failing consistently. I think this is related to a recent release of `@types/eslint__js`.

The package itself seems to be deprecated and isn't necessary anymore, see:
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71994
- https://github.com/eslint/eslint/pull/19010

I'm not sure why exactly, but it seems that the version currently released for `@types/eslint__js` as the `latest` tag has no `index.d.ts` file. I think this is causing the integration tests (that check that types are valid) to fail with: `Cannot find type definition file for 'eslint__js'.` (even if the file checked is empty).

Installing the `latest` version of `@types/eslint__js` and type checking an empty file seems to error on the above.

---

Should I open an issue about this? Having an empty `index.d.ts` can help in cases it's accidentally installed (unless this is intended since it's now deprecated).

### DIFF
--- a/packages/integration-tests/fixtures/flat-config-types-@types__eslint-v8/package.json
+++ b/packages/integration-tests/fixtures/flat-config-types-@types__eslint-v8/package.json
@@ -3,7 +3,6 @@
   "devDependencies": {
     "@types/eslint__eslintrc": "latest",
     "@eslint/eslintrc": "latest",
-    "@types/eslint__js": "^8",
     "@eslint/js": "latest",
     "@types/eslint": "^8",
     "eslint": "9.9.1",


OlivierZal:
Thanks @ronami!

JoshuaKGoldberg (approved):
🩹

JoshuaKGoldberg:
> Should I open an issue about this?

Probably a good idea, yeah. Even if it's correct as-is, it's good to have SEO friendly questions about it.

PR #10874, opened 2025-02-24
chore: automatically migrate nx to the latest version with renovate

JamesHenry:
I noticed there had been a few cases in the past where only the nx package had been manually bumped in isolation.

This PR stops renovate from ignoring nx updates and instead leverages them as a trigger to run `nx migrate` automatically.

This was always the intention of the `nx-migrate` workflow, but it can only work if renovate is opening the PR for us as intended.

----

Now on an ongoing basis the flow (push-based from a release of Nx):

1. Nx team cuts new release
2. Renovate picks up on it and opens a PR to the typescript-eslint repo containing the package bumps
3. The nx-migrate workflow is triggered and reverts the bumps from renovate and reapplies them via a full `nx migrate` command which will include the results of any relevant migrations/codemods the nx release provided
4. The nx-migrate workflow commits the changes to the same PR and it is ready for review by a team member


JoshuaKGoldberg (approved):
Sweet!

PR #10886, opened 2025-02-25
chore: remove unused scripts

43081j:
Removes a couple of unused/broken scripts from
`rule-schema-to-typescript-types`.

Also adds a missing nx target such that `generate-breaking-changes` works again.

## PR Checklist

- [x] Addresses an existing open issue: relates to #10885 
- [ ] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22) (**fully aware this issue isn't marked as accepting PRs yet, but this is clearly broken and has to happen for that issue to be tackleable**)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

Issue #10885, opened 2025-02-25
Repo: execute typescript scripts/tools with node directly


43081j:
### Suggestion

In latest node, we can use `--experimental-strip-types` to run typescript sources directly with node.

We're likely not ready for this yet, but once this is less experimental and more widely available, we should drop `tsx` and switch to using node directly.

kirkwaiblinger:
Linking https://github.com/typescript-eslint/typescript-eslint/issues/10711 for visibility and somewhat duplicated discussion

43081j:
FYI #10887 fixes that issue at the root

our scripts are actually ES modules (in source), but transpiled to CJS by typescript (`type: "commonjs"`)

if we just rename them to `*.mts`, they will execute via tsx and node fine

43081j:
and here is the PR that switches to node #10888

of course, we can't have it yet but it does work in latest node. so we can leave it as a draft until some day we want it (or if you don't like clutter, close the PR and keep the branch)

kirkwaiblinger:
I don't think it's an accident or incorrect that those scripts have been CJS. Writing CJS modules in TS using `"type":"commonjs"` and ESM-style `import`s is a standard thing (especially before `verbatimModuleSyntax`'s existence). That said, I am in favor of ESM wherever possible so if converting the scripts solves our problems, I'm all for it.

43081j:
we currently have a mixture of scripts in the repo which are ESM or CJS. that should be enough reason to switch to one or the other, regardless of if they were intentionally CJS at some point in the past

additionally, node would detect them as CJS according to our `type`, and yet they're actually ES modules. so either they need the `mts` extension or they need to be written as CJS. you're allowed a mixture because tsx is transpiling them on the fly to the correct syntax (CJS)

kirkwaiblinger:
> ...additionally, node would detect them as CJS according to our `type`, and yet they're actually ES modules. 

This is the part I'm not following. In what way are they "actually" ES modules if they use *.ts in `"type": "CommonJS"` and make use of CJS globals (i.e. `__dirname`)? 

But in any case, I think it's moot, since we're in agreement here and on #10711 to convert to actual ESM.

43081j:
given you're in a `type: "commonjs"` package, that (as far as node is concerned) implies `*.js` and `*.ts` are CJS.

[...]

PR #10887, opened 2025-02-25
Nodets types prep

43081j:
Depends on #10886 

This prepares the various scripts for being consumed by `node` directly (via the strip-types feature).

Primarily this means the following:

- Our packages are `type: "commonjs"`, so some of our scripts are incorrectly `*.ts` (CJS) which `tsx` covers up by the fact it transpiles it to CJS on the fly
  - Moving any `*.ts` scripts to `*.mts` solves this
- enum are not allowed since they are not strippable types. instead, we just use a string union

## PR Checklist

- [x] Addresses an existing open issue: related to #10885 


43081j:
this should be good to go now @kirkwaiblinger

tools/release/release.mts
```diff
@@ -1,5 +1,9 @@
 import { execaSync } from 'execa';
-import { releaseChangelog, releasePublish, releaseVersion } from 'nx/release';
+import {
+  releaseChangelog,
+  releasePublish,
+  releaseVersion,
+} from 'nx/release/index.js';
```
kirkwaiblinger, line 6: What is the reasoning for this change?

kirkwaiblinger (commented):
Looks good! Thanks! One question but no action requested

tools/release/release.mts
JoshuaKGoldberg: Cross-linking: https://github.com/typescript-eslint/typescript-eslint/pull/10903#discussion_r1976544984

43081j: Same as the referenced issue I think. Typescript is unhappy when executing this. It probably worked under tsx because it was using cjs. Whatever the reason, this is the correct import

43081j:
Not too sure why ci is failing. Doesn't seem related to the changes here, but that is more concerning 😅 if main has problems

tools/release/release.mts
kirkwaiblinger: More detail in https://github.com/typescript-eslint/typescript-eslint/pull/10903#discussion_r1976698563, but this is indeed a correct change and relevant to this PR (to prevent runtime error running `yarn release` with node 23.6). Good catch!

kirkwaiblinger (approved):
Thanks for making these changes! It's always satisfying to me to get TS a step closer to out-of-the-box node.js rather than requiring layers of transpilation and/or CLI tools just to be able to execute 🙂 

Re the integration test that was failing, that was fixed in https://github.com/typescript-eslint/typescript-eslint/pull/10906 👍

PR #10889, opened 2025-02-25
feat(eslint-plugin): [explicit-module-boundary-types] add an option to ignore overload implementations

### LINKED ISSUE #4586: [explicit-module-boundary-types] New option to allow unspecified return type if there are overload signatures

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4586

## Overview

This PR addresses #4586 and add an option to ignore overload implementations:

```ts
export function test(a: string): string;
export function test(a: number): number;
export function test(a: unknown) {
  return a;
}
```

---

Some thoughts:

Would it make sense to apply the same option to `explicit-function-return-type` too?

It seems both rules share several options (e.g. `allowHigherOrderFunctions`, `allowDirectConstAssertionInArrowFunctions`, `allowTypedFunctionExpressions`) and also have [common code for handling these options](https://github.com/typescript-eslint/typescript-eslint/blob/c910ac1553e517ebc534d3271e84b51e4ac23f5d/packages/eslint-plugin/src/util/explicitReturnTypeUtils.ts).

I think the motivation for having this option in `explicit-function-return-type` is the same for `explicit-module-boundary-types`.

### DIFF
--- a/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
+++ b/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
@@ -260,16 +260,16 @@ export const foo: FooType = bar => {};
 </TabItem>
 </Tabs>
 
-### `allowImplicitReturnTypeForOverloadImplementations`
+### `allowOverloadFunctions`
 
 {/* insert option description */}
 
-Examples of code for this rule with `{ allowImplicitReturnTypeForOverloadImplementations: false }`:
+Examples of code for this rule with `{ allowOverloadFunctions: false }`:
 
 <Tabs>
 <TabItem value="❌ Incorrect">
 
-```ts option='{ "allowImplicitReturnTypeForOverloadImplementations": false }'
+```ts option='{ "allowOverloadFunctions": false }'
 export function test(a: string): string;
 export function test(a: number): number;
 export function test(a: unknown) {
@@ -280,7 +280,7 @@ export function test(a: unknown) {
 </TabItem>
 <TabItem value="✅ Correct">
 
-```ts option='{ "allowImplicitReturnTypeForOverloadImplementations": true }'
+```ts option='{ "allowOverloadFunctions": true }'
 export function test(a: string): string;
 export function test(a: number): number;
 export function test(a: unknown) {
--- a/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
+++ b/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
@@ -260,6 +260,37 @@ export const foo: FooType = bar => {};
 </TabItem>
 </Tabs>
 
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
```diff
@@ -260,6 +260,37 @@ export const foo: FooType = bar => {};
 </TabItem>
 </Tabs>
 
+### `allowOverloadFunctions`
+
+{/* insert option description */}
+
+Examples of code for this rule with `{ allowOverloadFunctions: false }`:
```
JoshuaKGoldberg, line 267: [Docs] The code blocks have the same content for _Incorrect_ and _Correct_, which as a reader I find confusing. But I can't think of anything to put in the _Correct_ one. Maybe it'd be best to just have one code block, not two tabbed ones?

JoshuaKGoldberg (changes requested):
LGTM, nice and clean! Just one thing on docs? 🧹

packages/eslint-plugin/docs/rules/explicit-module-boundary-types.mdx
ronami: Yes!

I agree, I've updated the PR to only show one code block for correct code, as the option expands what the rule allows.

JoshuaKGoldberg (approved):
Conan O'Brien dancing side-to-side

PR #10903, opened 2025-02-28
feat: support TypeScript 5.8

### LINKED ISSUE #10884: TypeScript 5.8 Support
https://devblogs.microsoft.com/typescript/announcing-typescript-5-8-rc/

_Please be patient._

------

# 🔲 lib.d.ts Updates

- Granular Checks for Branches in Return Expressions
- Support for require() of ECMAScript Modules in --module nodenext
- --module node18
- The --erasableSyntaxOnly Option
- The --libReplacement Flag
- Preserved Computed Property Names in Declaration Files
- Optimizations on Program Loads and Updates
    - Presumably this should also increase type-aware linting performance
- Restrictions on Import Assertions Under --module nodenext

dirkluijk:
Closes https://github.com/typescript-eslint/typescript-eslint/issues/10884

## PR Checklist

- [X] Addresses an existing open issue: closes #10884

## Overview

> ## lib.d.ts Updates
> We will need to regenerate our types within scope-manager.

~I am still figuring out if I need to do something for this.~

Never mind, done.


dirkluijk:
Hmm, there seems to be a linting issue in `tools/release/release.mts`. Not sure why, will look into it.

package.json
```diff
@@ -145,7 +145,7 @@
     "react-split-pane@^0.1.92": "patch:react-split-pane@npm%3A0.1.92#./.yarn/patches/react-split-pane-npm-0.1.92-93dbf51dff.patch",
     "tmp": "0.2.1",
     "tsx": "^4.7.2",
-    "typescript": "5.7.2"
+    "typescript": "~5.8.0"
```
JoshuaKGoldberg, line 148: I don't think this changes anything practically, but we've gone with explicit numbers before and I don't know of a good reason to change it:

```suggestion
    "typescript": "5.8.2"
```

Mild preference from me on going explicit just to make things... well, explicit.

tools/release/release.mts
```diff
@@ -1,5 +1,9 @@
 import { execaSync } from 'execa';
-import { releaseChangelog, releasePublish, releaseVersion } from 'nx/release';
+import {
+  releaseChangelog,
+  releasePublish,
+  releaseVersion,
+} from 'nx/release/index.js';
```
JoshuaKGoldberg, line 6: Unrelated change, is there a reason to include it here?

This showed up in other places -at least #10887, cc @kirkwaiblinger- so maybe we need to do this in `main`?

JoshuaKGoldberg (approved):
LGTM, very nicely done. Thanks for helping out! 🙌 

Just a couple of small things. Nothing that we can't do before the weekly release tomorrow if you don't have time.

bradzacher (dismissed):
Same comments as joshg

package.json
dirkluijk: I assume that's what the lock is for, right? Anyway, used a fixed version instead 😉

tools/release/release.mts
dirkluijk: Yeah, see https://github.com/typescript-eslint/typescript-eslint/pull/10903#issuecomment-2691576236

My theory is - but I haven't checked it yet - that TypeScript 5.8 behaves slightly different with deep imports in CommonJS modules that don't have specific secondary entry points. 

On TS 5.7 this is not an issue, so I think it belongs to this PR. Although we could also add this to `main` first of course.

kirkwaiblinger: We had a poke around, and the change is due to https://github.com/microsoft/TypeScript/issues/61057. TL;DR you can't omit the `/index.js` when `import`ing from ESM in node. Seemingly executing with tsx rather than node has been preventing this from being a runtime error for us so far, but the TS 5.8 error is correct, and, accordingly, so is this change. 👍  Also @JamesHenry is aware and might tweak things on the `nx` side (whether just [the docs](https://nx.dev/features/manage-releases#using-the-programmatic-api-for-nx-release) or adding a subpath `exports` field)

kirkwaiblinger: As for plan of action - both #10887 and this PR genuinely need this change for different reasons (TS type-checking error here, and runtime error when using node 23.6, which #10887 resolves), so I'd say let's just keep the change in both PRs and whichever is merged first is merged first 🤷‍♂️

bradzacher (approved):
lgtm15

Issue #10682, opened 2025-01-19
Enhancement: add strict `parent` types for nodes that have well-defined parents


ronami:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [x] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [x] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [x] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Relevant Package

types

### My proposal is suitable for this project

- [x] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Description

This PR is a very similar proposal to the one in https://github.com/typescript-eslint/typescript-eslint/issues/6225 and suggests having stricter `parent` types for some AST nodes:

- `ExportDefaultDeclaration` has parent `TSESTree.BlockStatement | TSESTree.Program | TSESTree.TSModuleBlock`
- `ExportNamedDeclaration` has parent `TSESTree.BlockStatement | TSESTree.Program | TSESTree.TSModuleBlock`
- `FunctionDeclaration` with `name`  has parent `TSESTree.BlockStatement | TSESTree.ExportDefaultDeclaration | TSESTree.ExportNamedDeclaration | TSESTree.Program`.
- `FunctionDeclaration` without `name` has parent `TSESTree.ExportDefaultDeclaration` ([reference](https://github.com/typescript-eslint/typescript-eslint/blob/2de5a10ade85d883cfe85ca79b91b5ad9f1420f4/packages/ast-spec/src/declaration/FunctionDeclaration/spec.ts#L24)).

This is a follow-up to the discussion in https://github.com/typescript-eslint/typescript-eslint/pull/10422#discussion_r1892932430.

---

Some thoughts/notes:

[...]

Issue #6255, opened 2022-12-20
Bug: [member-ordering] optionalityOrder is reversed


TFohrer:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.4&sourceType=module&code=C4TwDgpgBAsiAq5oF4oG8BQUoCcIEcBXASzwBMAFHAezAC4oBnYHYgOwHMBuLKW4YtTYBDADZVaAfgbNWnHgF8gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AW0XYCNFpaA9tAAmfZgHN0YANrhsOPtCGQANHPnysGnEVIAZSvj4BDeKgxbtOATUoCmpw4QDyIvlJiIAjrErREwrQAZn7I+JDqVhAAvpHasdoAunKx0UA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
type MyType = {
  requiredProp: string;
  optionalProp?: string;
}
```

### ESLint Config

```javascript
module.exports = {
  parser: "@typescript-eslint/parser",
  "rules": {
    "@typescript-eslint/member-ordering": [
      "error",
          {
        "typeLiterals": {
          "optionalityOrder": "required-first"
          }
          }
      ]
  }
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

I expect no error when the requiredProp is the first item. 
I expect an error when the requiredProp is not the first item. 

### Actual Result

Error is shown when requiredProp is first item: 

```
Member requiredProp should be declared after all optional members. 2:3 - 2:24
```

### Additional Info

Seems like in the implementation the result for the optionalOrRequired condition is reversed:
[...]

PR #10907, opened 2025-03-02
fix(eslint-plugin): [no-unnecessary-type-parameters] should parenthesize type in suggestion fixer if necessary

### LINKED ISSUE #10894: Bug: [no-unnecessary-type-parameters] should parenthesize type in suggestion fixer if necessary

### Playground Link

[**(BRANCH PLAYGROUND - NOT ON MAIN YET)**](https://deploy-preview-9536--typescript-eslint.netlify.app/play/#ts=5.7.2&fileType=.tsx&code=GYVwdgxgLglg9mABAKzjMAhApsOAnLAZRAHMSsBnWBAHgBVEsAPKLMAEwsSr3RMQA%2BiMCAC2AIyx4AfAAosAGwoAuRHQDaAXQCUiAN4AoRIgJQQeJIooA6UQEMADvIWIAvNMQByT4gDUjBW1rVHRZTwAaT20AbgMAXwMDUEhqJBCwAEFgVjxiMkpU5xVuKF4wfiERCSktXUNjU3NLJVtHZzcPbz8AoPSwyJj4oA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJADuIppHBgAviF1A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

### Repro Code

```TypeScript
function join<T extends string | number>(els: T[]) {
  return els.map(el => '' + el).join(',');
}
```

### ESLint Config

```javascript
module.exports = {
  "rules": {
    "@typescript-eslint/no-unnecessary-type-parameters": "warn"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

```ts
function join(els: (string | number)[]) {
  return els.map(el => '' + el).join(',');
}
```

### Actual Result

```ts
function join(els: string | number[]) {
  return els.map(el => '' + el).join(',');
}
```

### Additional Info

This is a followup to https://github.com/typescript-eslint/typescript-eslint/pull/9536#issuecomment-2685621735. note that, while this specific report case isn't merged yet, 
- I'm pretty sure it will be merged imminently
- I'm pretty sure we could figure out a repro where the parenthesization is wrong on a case the rule already reports on, but I haven't bothered thinking too hard about it

y-hsgw:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10894

## Overview

Fixed the bug described in the title.

Please note that in addition to the error examples in the #10894, I have also addressed several related cases.

🐛


JoshuaKGoldberg (approved):
Thanks! 👍

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -108,7 +108,21 @@ export default createRule({
...
                   if (reference.isTypeReference) {
                     const referenceNode = reference.identifier;
-                    yield fixer.replaceText(referenceNode, constraintText);
+                    const isCompositeType =
+                      constraint?.type === AST_NODE_TYPES.TSUnionType ||
+                      constraint?.type === AST_NODE_TYPES.TSIntersectionType;
+                    const hasMatchingAncestorType = [
+                      AST_NODE_TYPES.TSArrayType,
+                      AST_NODE_TYPES.TSIndexedAccessType,
+                    ].some(type => referenceNode.parent.parent?.type === type);
```
JoshuaKGoldberg, line 117: There's a missing code coverage point here, as reported by Codecov. Removing the `?` doesn't fail the test. 

It looks to me like another example of #6255 / #10682 where the AST types for `parent` nodes aren't precise enough to know the grandparent of an identifier is defined. If that's the case we'll want a `!` type assertion and a followup issue. Or, if not, there's a missing unit test for a case that could be hit. Could you please take a look?

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
JoshuaKGoldberg, line 1: [Testing] Just to be thorough, could you add a few tests around:

* Combination `|` union and `&` intersection types
* Types with unnecessary parenthesis around them (which might need `noFormat`)

They all worked when I tried them out, so this is just for future-proofing.

JoshuaKGoldberg (changes requested):
Good stuff! I think just a bit of extra testing & code flow trimming are needed now. 🔥

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
y-hsgw: Based on my investigation, it seems likely that when `reference.isTypeReference` is `true`, the `type` of `reference.identifier.parent` is `TSESTree.TSTypeReference`.
I agree that a follow-up might be necessary, as you suggested.
I have added the ! type assertion as well. → [91dd50d](https://github.com/typescript-eslint/typescript-eslint/pull/10907/commits/91dd50d9df0d4758deba94405cba78a6e2b4459e)

This is just a quick idea, but it seems like we might need some form of type narrowing:
```ts
  public isTypeReference(): this is Reference & {
    identifier: (TSESTree.Identifier | TSESTree.JSXIdentifier) & {
      parent: TSESTree.TSTypeReference;
    };
  } {
    return (this.#referenceType & ReferenceTypeFlag.Type) !== 0;
  }
```

packages/eslint-plugin/tests/rules/no-unnecessary-type-parameters.test.ts
y-hsgw: I added in [4cb399a](https://github.com/typescript-eslint/typescript-eslint/pull/10907/commits/4cb399a58302f85fd78a67688b17822e4640e219)!

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
JoshuaKGoldberg: Oh joy, a refactor. Yes, an issue would be good. I suspect we don't have much freedom in changing the shape of `isTypeReference`.

JoshuaKGoldberg (approved):
🚀 Looks great to me, thanks!

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
```diff
@@ -108,7 +108,22 @@ export default createRule({
...
-                    yield fixer.replaceText(referenceNode, constraintText);
+                    const isCompositeType =
+                      constraint?.type === AST_NODE_TYPES.TSUnionType ||
+                      constraint?.type === AST_NODE_TYPES.TSIntersectionType;
+                    const hasMatchingAncestorType = [
+                      AST_NODE_TYPES.TSArrayType,
+                      AST_NODE_TYPES.TSIndexedAccessType,
+                      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                    ].some(type => referenceNode.parent.parent!.type === type);
+                    if (isCompositeType && hasMatchingAncestorType) {
```
kirkwaiblinger, line 119: I think this logic isn't quite right, in that it's not general enough.... We really want (as close as we can) to answer the question - "is the destination context going to be higher precedence than the source?".  This logic only identifies the special case of intersection/union being lower precedence than TSArrayType/TSIndexedAccessType, which is true, but there is much more nuance.

For example, `TSIntersectionType` is higher precedence than `TSUnionType`, so the following test case currently doesn't work correctly:
```ts
type A = string;
type B = string;
type C = string;

declare function f<T extends A | B>(): T & C;

// fixes to 
declare function f(): A | B & C;
// (which is equivalent to)
declare function f(): A | (B & C);

// should fix to 
declare function f(): (A | B) & C;
```

Another (quite perverse) test case that isn't currently handled correctly is 

```ts
type A = string;
type B = string;
type C = string;
type D = string;

declare function f<T extends A extends B ? C : D>(): T | null;
// fixes to 
declare function f(): A extends B ? C : D | null;
// (which is equivalent to)
declare function f(): A extends B ? C : (D | null);

// should fix to 
declare function f(): (A extends B ? C : D) | null;
```

Note that it's not necessarily practical to get this 100% correct in all cases, so some extra "unnecessary" parens are totally acceptable here if it the implementation benefits. FWIW - [prettier actually puts "unnecessary" parens anyway for the mixed union/intersection case](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEATOYA2BDATnAAgDMBXKMGAS2mIAoBKJAgQQIB8CAhAgMgIGEA3AB0o6LHkKlyVGkQZNWHWtz796IsRhz5iZCtSh1GBWkq71eAwSAA0ICAAdZUAM7JQeXBADuABUk3ZBBsTB9sAE93ewAjXGwwAGs4GABlbABbOAAZSig4ZCJQ1zh7CBiAKwwYAHV4x2DHfBLcADcC2PiklNTHBLyAc2QYXBJSkBKMymHR8bgAD0c4XEos2FCAeSX4mAhcPwhXShdghFQ7EAXt1YQYUIAVZag8Sjh3JCLMEvsjqAHMOAARRIEHghWK4wqrnmqUGAOBoIKHwh9gAjiD4H5vI53iFXABafJwdDnewjbCUTCDfgQDIZbDBUKYC6-f5wZgwEaUGIkTHLXL5cFfcYACxgGUwNRFxzefTAcFSCCOVFaxwiwTArmiIFaYwAklpYKkwCtnMwxKkYBEAULviAmoc4HVsA0UE03st2hc8i0YFjsAN6bbxn1cC1glalq4TZRnBcmnlapRUDARcgABwABns+HRlHw-sDDORwrJ2BiNWTqeQACZ7CQSvdy+9Pna4BkYsSSdlsH8SAG4AAxPb0zmDRm8iAgAC+06AA), since it's just hard to read for us humans otherwise 🙃 

Have you had a look at the implementations of `getWrappingFixer` and friends? I'd think they have some of this logic already implemented more generally?

kirkwaiblinger (changes requested):

packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts
y-hsgw: Thank you. I've added the missing cases and modified it to use getWrappingFixer.

JoshuaKGoldberg (dismissed):
👍 will defer to @kirkwaiblinger, as they caught things I didn't last time. LGTM otherwise!

kirkwaiblinger:
Sorry for the delay!

kirkwaiblinger (approved):

Issue #4571, opened 2022-02-18
Rule proposal: no-unused-private-class-members


alumni:
### Description

To complement the original `no-unused-private-class-members` rule which works with typescript but doesn't detect the `private` keyword:

> Private class members that are declared and not used anywhere in the code are most likely an error due to incomplete refactoring. Such class members take up space in the code and can lead to confusion by readers.

&mdash; From: https://eslint.org/docs/rules/no-unused-private-class-members#no-unused-private-class-members

See also: https://github.com/eslint/eslint/issues/14859.

### Fail

```ts
class Foo {
    #unusedMember: number;
    private anotherUnusedMember: number;
}

class Foo {
    #usedOnlyInWrite: number;

    constructor(private anotherUsedOnlyInWrite: number) {
        // noop
    }

    method() {
        this.#usedOnlyInWrite = 42;
    }
}

class Foo {
    #usedOnlyToUpdateItself = 5;

    method() {
        this.#usedOnlyToUpdateItself++;
    }
}

class Foo {
    #unusedMethod() {}
    private anotherUnusedMethod() {}
}

class Foo {
    get #unusedAccessor() {}
    set #unusedAccessor(value) {}
    private get anotherUnusedAccessor() {}
    private set anotherUnusedAccessor(value) {}
}
```

### Pass

```ts
class Foo {
    #usedMember = 35;
    private anotherUsedMember = 7;

    method() {
        return this.#usedMember + this.anotherUsedMember;
    }
}

class Foo {
    constructor(private readonly usedMember: number) {
        // noop
    }

    doSomething() {
        return this.usedMember / 42;
    }
}


class Foo {
    #usedMethod() {
        return 35;
    }

    private anotherUsedMethod() {
        return 7;
    }

    anotherMethod() {
        return this.#usedMethod() + this.anotherUsedMethod();
    }
}

class Foo {
    get #usedAccessor() {}
    set #usedAccessor(value) {}
    private get anotherUsedAccessor() {}
    private set anotherUsedAccessor(value) {}
    
    method() {
        this.#usedAccessor = 42;
        this.anotherUsedAccessor = 42;
    }
}
```

bradzacher:
#private members are completely safe to mark as unused because they are 100%, runtime enforced to be only accessible within the class they are defined.

However `private` members are not. They are accessible outside the class via `instance['member']` (yes, that's valid and safe TS). And are ultimately just type-only sugar for a normal class member.

[...]

PR #10913, opened 2025-03-03
feat(eslint-plugin): [no-unused-private-class-members] new extension rule

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/4571

## Overview

This PR implements an extension rule for `no-unused-private-class-members` that introduces support for `private` members.

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-unused-private-class-members.mdx
+++ b/packages/eslint-plugin/docs/rules/no-unused-private-class-members.mdx
@@ -1,0 +1,82 @@
+---
+description: 'Disallow unused private class members.'
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+> 🛑 This file is source code, not the primary documentation location! 🛑
+>
+> See **https://typescript-eslint.io/rules/no-unused-private-class-members** for documentation.
+
+This rule extends the base [`eslint/no-unused-private-class-members`](https://eslint.org/docs/rules/no-unused-private-class-members) rule.
+It adds support for members declared with TypeScript's `private` keyword.
+
+## Options
+
+This rule has no options.
+
+## Examples
+
+<Tabs>
+<TabItem value="❌ Incorrect">
+
+```ts
+class A {
+  private foo = 123;
+}
+```
+
+</TabItem>
+<TabItem value="✅ Correct">
+
+```tsx
+class A {
+  private foo = 123;
+
+  constructor() {
+    console.log(this.foo);
+  }
+}
+```
+
+</TabItem>
+</Tabs>
+
+## Limitations
+
+This rule does not detect the following cases:
+
+(1) Nested classes using private members of the outer class:
+
+```ts
+class Foo {
+  private prop = 123;
+
+  method(a: Foo) {
+    return class {
+      // ✅ Detected as a usage
+      prop = this.prop;
+
+      // ❌ NOT detected as a usage
+      innerProp = a.prop;
+    };
+  }
+}
+```
+
+(2) Usages of the private member outside of the class:
+
+```ts
+class Foo {
+  private prop = 123;
+}
+
+const instance = new Foo();
+// ❌ NOT detected as a usage
+console.log(foo['prop']);
+```
+
+## When Not To Use It
+
+If you don't want to be notified about unused private class members, you can safely turn this rule off.
--- a/packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
+++ b/packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
@@ -1,0 +1,266 @@
+// The following code is adapted from the code in eslint/eslint.
+// Source: https://github.com/eslint/eslint/blob/522d8a32f326c52886c531f43cf6a1ff15af8286/lib/rules/no-unused-private-class-members.js
+// License: https://github.com/eslint/eslint/blob/522d8a32f326c52886c531f43cf6a1ff15af8286/LICENSE
+
+import type { TSESTree } from '@typescript-eslint/utils';
+
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+
+import { createRule } from '../util';
+
+type Options = [];
+export type MessageIds = 'unusedPrivateClassMember';
+
+interface PrivateMember {
+  declaredNode:
+    | TSESTree.MethodDefinitionComputedName
+    | TSESTree.MethodDefinitionNonComputedName
+    | TSESTree.PropertyDefinitionComputedName
+    | TSESTree.PropertyDefinitionNonComputedName;
+  isAccessor: boolean;
+}
+
+export default createRule<Options, MessageIds>({
+  name: 'no-unused-private-class-members',
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow unused private class members',
+      extendsBaseRule: true,
+      requiresTypeChecking: false,
+    },
+
+    messages: {
+      unusedPrivateClassMember:
+        "'{{classMemberName}}' is defined but never used.",
+    },
+
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    const trackedClassMembers: Map<string, PrivateMember>[] = [];
+
+    /**
+     * The core ESLint rule tracks methods by adding an extra property of
+     * "isUsed" to the method, which is a bug according to Joshua Goldberg.
+     * Instead, in our extended rule, we create a separate data structure to
+     * track whether a method is being used.
+     */
+    const trackedClassMembersUsed = new Set<
+      | TSESTree.MethodDefinitionComputedName
+      | TSESTree.MethodDefinitionNonComputedName
+      | TSESTree.PropertyDefinitionComputedName
+      | TSESTree.PropertyDefinitionNonComputedName
+    >();
+
+    /**
+     * Check whether the current node is in a write only assignment.
+     * @param node Node referring to a private identifier
+     * @returns Whether the node is in a write only assignment
+     * @private
+     */
+    function isWriteOnlyAssignment(
+      node: TSESTree.Identifier | TSESTree.PrivateIdentifier,
+    ): boolean {
+      const parentStatement = node.parent.parent;
+      if (parentStatement == null) {
+        return false;
+      }
+
+      const isAssignmentExpression =
+        parentStatement.type === AST_NODE_TYPES.AssignmentExpression;
+
+      if (
+        !isAssignmentExpression &&
+        parentStatement.type !== AST_NODE_TYPES.ForInStatement &&
+        parentStatement.type !== AST_NODE_TYPES.ForOfStatement &&
+        parentStatement.type !== AST_NODE_TYPES.AssignmentPattern
+      ) {
+        return false;
+      }
+
+      // It is a write-only usage, since we still allow usages on the right for
+      // reads.
+      if (parentStatement.left !== node.parent) {
+        return false;
+      }
+
+      // For any other operator (such as '+=') we still consider it a read
+      // operation.
+      if (isAssignmentExpression && parentStatement.operator !== '=') {
+        // However, if the read operation is "discarded" in an empty statement,
+        // then we consider it write only.
+        return (
+          parentStatement.parent.type === AST_NODE_TYPES.ExpressionStatement
+        );
+      }
+
+      return true;
+    }
+
+    //--------------------------------------------------------------------------
+    // Public
+    //--------------------------------------------------------------------------
+
+    function processPrivateIdentifier(
+      node: TSESTree.Identifier | TSESTree.PrivateIdentifier,
+    ): void {
+      const classBody = trackedClassMembers.find(classProperties =>
+        classProperties.has(node.name),
+      );
+
+      // Can't happen, as it is a parser error to have a missing class body, but
+      // let's code defensively here.
+      if (classBody == null) {
+        return;
+      }
+
+      // In case any other usage was already detected, we can short circuit the
+      // logic here.
+      const memberDefinition = classBody.get(node.name);
+      if (memberDefinition == null) {
+        return;
+      }
+      if (trackedClassMembersUsed.has(memberDefinition.declaredNode)) {
+        return;
+      }
+
+      // The definition of the class member itself
+      if (
+        node.parent.type === AST_NODE_TYPES.PropertyDefinition ||
+        node.parent.type === AST_NODE_TYPES.MethodDefinition
+      ) {
+        return;
+      }
+
+      // Any usage of an accessor is considered a read, as the getter/setter can
+      // have side-effects in its definition.
+      if (memberDefinition.isAccessor) {
+        trackedClassMembersUsed.add(memberDefinition.declaredNode);
+        return;
+      }
+
+      // Any assignments to this member, except for assignments that also read
+      if (isWriteOnlyAssignment(node)) {
+        return;
+      }
+
+      const wrappingExpressionType = node.parent.parent?.type;
+      const parentOfWrappingExpressionType = node.parent.parent?.parent?.type;
+
+      // A statement which only increments (`this.#x++;`)
+      if (
+        wrappingExpressionType === AST_NODE_TYPES.UpdateExpression &&
+        parentOfWrappingExpressionType === AST_NODE_TYPES.ExpressionStatement
+      ) {
+        return;
+      }
+
+      /*
+       * ({ x: this.#usedInDestructuring } = bar);
+       *
+       * But should treat the following as a read:
+       * ({ [this.#x]: a } = foo);
+       */
+      if (
+        wrappingExpressionType === AST_NODE_TYPES.Property &&
+        parentOfWrappingExpressionType === AST_NODE_TYPES.ObjectPattern &&
+        node.parent.parent?.value === node.parent
+      ) {
+        return;
+      }
+
+      // [...this.#unusedInRestPattern] = bar;
+      if (wrappingExpressionType === AST_NODE_TYPES.RestElement) {
+        return;
+      }
+
+      // [this.#unusedInAssignmentPattern] = bar;
+      if (wrappingExpressionType === AST_NODE_TYPES.ArrayPattern) {
+        return;
+      }
+
+      // We can't delete the memberDefinition, as we need to keep track of which
+      // member we are marking as used. In the case of nested classes, we only
+      // mark the first member we encounter as used. If you were to delete the
+      // member, then any subsequent usage could incorrectly mark the member of
+      // an encapsulating parent class as used, which is incorrect.
+      trackedClassMembersUsed.add(memberDefinition.declaredNode);
+    }
+
+    return {
+      // Collect all declared members/methods up front and assume they are all
+      // unused.
+      ClassBody(classBodyNode): void {
+        const privateMembers = new Map<string, PrivateMember>();
+
+        trackedClassMembers.unshift(privateMembers);
+        for (const bodyMember of classBodyNode.body) {
+          if (
+            (bodyMember.type === AST_NODE_TYPES.PropertyDefinition ||
+              bodyMember.type === AST_NODE_TYPES.MethodDefinition) &&
+            (bodyMember.key.type === AST_NODE_TYPES.PrivateIdentifier ||
+              (bodyMember.key.type === AST_NODE_TYPES.Identifier &&
+                bodyMember.accessibility === 'private'))
+          ) {
+            privateMembers.set(bodyMember.key.name, {
+              declaredNode: bodyMember,
+              isAccessor:
+                bodyMember.type === AST_NODE_TYPES.MethodDefinition &&
+                (bodyMember.kind === 'set' || bodyMember.kind === 'get'),
+            });
+          }
+        }
+      },
+
+      // Process all usages of the private identifier and remove a member from
+      // `declaredAndUnusedPrivateMembers` if we deem it used.
+      PrivateIdentifier(node): void {
+        processPrivateIdentifier(node);
+      },
+
+      // Process all usages of the identifier and remove a member from
+      // `declaredAndUnusedPrivateMembers` if we deem it used.
+      MemberExpression(node): void {
+        if (
+          !node.computed &&
+          node.object.type === AST_NODE_TYPES.ThisExpression &&
+          node.property.type === AST_NODE_TYPES.Identifier
+        ) {
+          processPrivateIdentifier(node.property);
+        }
+      },
+
+      // Post-process the class members and report any remaining members. Since
+      // private members can only be accessed in the current class context, we
+      // can safely assume that all usages are within the current class body.
+      'ClassBody:exit'(): void {
+        const unusedPrivateMembers = trackedClassMembers.shift();
+        if (unusedPrivateMembers == null) {
+          return;
+        }
+
+        for (const [
+          classMemberName,
+          { declaredNode },
+        ] of unusedPrivateMembers.entries()) {
+          if (trackedClassMembersUsed.has(declaredNode)) {
+            continue;
+          }
+          context.report({
+            loc: declaredNode.key.loc,
+            node: declaredNode,
+            messageId: 'unusedPrivateClassMember',
+            data: {
+              classMemberName:
+                declaredNode.key.type === AST_NODE_TYPES.PrivateIdentifier
+                  ? `#${classMemberName}`
+                  : classMemberName,
+            },
+          });
+        }
+      },
+    };
+  },
+});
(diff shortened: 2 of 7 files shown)


JoshuaKGoldberg (dismissed):
Conan O'Brien frankly saying 'yep'

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
```diff
@@ -0,0 +1,266 @@
...
+            });
+          }
+        }
+      },
+
+      // Process all usages of the private identifier and remove a member from
+      // `declaredAndUnusedPrivateMembers` if we deem it used.
+      PrivateIdentifier(node): void {
+        processPrivateIdentifier(node);
+      },
```
ronami, line 221: [Bug]: I've noticed some inconsistency that creates false positives with `static` `private` properties ([deploy preview playground link](https://deploy-preview-10913--typescript-eslint.netlify.app/play/#ts=5.8.2&showAST=es&fileType=.tsx&code=MYGwhgzhAEAqCmEAuBGaBvAUNaAHArgEYgCWw0yYSZ0AZgPb0AUAlBtjtAE7xL5cA7OIlQA6AMSEwXANwcAvpg4B6ZdGD0uPYEhABPaAPpJu8XJqTwAJh0rVyk6dAC80FHMWZQkGAmQAmdhwCYho7GgZmNixOUz5BYQDRKVkFJRxVU3MuSytoQnwTCAALenwQKwEAciQAfg5cLhIANyp4CiQqGhSXNw9MIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaWXWZRAE0IAdoBLANwEMAXRQgY3ieWUIFtE-AEaJoqDJHwAzaZAA04bJAACLAJ7UUHetRaEU8OrhYB6AsVLkqtRq3ZcefQSLESoY6PmiQlAXxA-IA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```ts
class Test1 {
  public static foo() {
    return Test1.#bar;
  }

  // correctly not reported
  static #bar = 1;
}

class Test2 {
  public static foo() {
    return Test2.bar;
  }

  // reported but shouldn't?
  private static bar = 1;
}
```

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
```diff
@@ -0,0 +1,266 @@
...
+          classMemberName,
+          { declaredNode },
+        ] of unusedPrivateMembers.entries()) {
+          if (trackedClassMembersUsed.has(declaredNode)) {
+            continue;
+          }
+          context.report({
+            loc: declaredNode.key.loc,
+            node: declaredNode,
+            messageId: 'unusedPrivateClassMember',
```
ronami, line 254: [Feat]:

packages/eslint-plugin/docs/rules/no-unused-private-class-members.mdx
```diff
@@ -0,0 +1,82 @@
...
+  constructor() {
+    console.log(this.foo);
+  }
+}
+```
+
+</TabItem>
+</Tabs>
+
+## Limitations
```
ronami, line 46: What do you think about also mentioning assigning `this` to a variable as a limitaiton?

```ts
class Test1 {
  private bar: number;

  foo() {
    const self = this;

    return function () {
      // ❌ NOT detected as a usage
      self.bar;
    }
  }
}
```

ronami (commented):
Excited to enable this on projects I work on 🚀🚀🚀 (I accidentally posted the review earlier while still adding comments 🙈)

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
```diff
@@ -0,0 +1,266 @@
...
+      ClassBody(classBodyNode): void {
+        const privateMembers = new Map<string, PrivateMember>();
+
+        trackedClassMembers.unshift(privateMembers);
+        for (const bodyMember of classBodyNode.body) {
+          if (
+            (bodyMember.type === AST_NODE_TYPES.PropertyDefinition ||
+              bodyMember.type === AST_NODE_TYPES.MethodDefinition) &&
+            (bodyMember.key.type === AST_NODE_TYPES.PrivateIdentifier ||
+              (bodyMember.key.type === AST_NODE_TYPES.Identifier &&
```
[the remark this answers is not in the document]
bradzacher: Generally we avoid computed properties because it's hard to validate that `private [foo]` is the same as `this[foo]`.

for example:
```ts
class Test1 {
  private [Symbol.iterator] = 1;

  method() {
    const Symbol = { iterator: 'lol' };
    return this[Symbol.iterator];
  }
}
```

```ts
function getObject(): { readonly key: string } {
  let key= '';
  return { get key() { key+= 'a'; return key; } }; 
}

const totallyStatic = getObject(); 
class Test1 {
  private [totallyStatic.key] = 1;

  method() {
    return this[totallyStatic.key];
  }
}
```

```ts
let foo = 'a';
class Test1 {
  private [foo] = 1;

  constructor() {
    foo += 'a';
  }

  method() {
    return this[foo];
  }
}
```

We _could_ support cases like `private ['a']` if we wanted to.

For v0 I'm happy to ignore computed properties and defer it to if users ask for it.

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
```diff
@@ -0,0 +1,266 @@
...
+      if (wrappingExpressionType === AST_NODE_TYPES.ArrayPattern) {
+        return;
+      }
+
+      // We can't delete the memberDefinition, as we need to keep track of which
+      // member we are marking as used. In the case of nested classes, we only
+      // mark the first member we encounter as used. If you were to delete the
+      // member, then any subsequent usage could incorrectly mark the member of
+      // an encapsulating parent class as used, which is incorrect.
+      trackedClassMembersUsed.add(memberDefinition.declaredNode);
```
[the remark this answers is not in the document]
bradzacher: This is because the base rule doesn't support that case because `this.#bar` is a syntax error inside a function expression!!

It's worth noting that TS doesn't actually catch this case either and reports `bar` as unused because the type of the `this` in the function expression is `any`!
[If you turn on `noImplicitThis` then TS will error on that `this`](https://deploy-preview-10913--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.tsx&code=MYGwhgzhAEAqCmEAuBGaBvAUNaB6X0EAFgPYCuIAJtAEbzQBO8ADiQ0vNTWUtJSYgB2AciQB%2BbNGYMAlgDcwHWmAYAuaILIBbOgwDcmSQDMSJaAF5oRsoOBIZJQdAAUASgyScTJGQZOkRDIQAHQ0KgY4AL6Y0UA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaWXWZRAE0IAdoBLANwEMAXRQgY3ieWUIFtE-AEaJoqDJHwAzaZAA04bJAACLAJ7UUHetRaEU8OrhYB6AsVLkqtRq3ZcefQSLESoY6PmiQlAXxA-IA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkRsAkkqstMngCoAWyZlemAK7oANOBqQSA1ngByg1KgDCXdNADWfMAOHiAviD1A&tokens=false)

We could support this if we wanted to. I'm leaning towards not because TS itself doesn't catch it either.

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
bradzacher: For your second example -- this is expected and is the same purposely not covered case as here:
https://github.com/typescript-eslint/typescript-eslint/pull/10913/files#diff-914330b2e3f9d71dfb628800b850412c3c5560900b13cf8a2925ff7af0905e53R50-R66

I made this rule purposely NOT use type info because the cases that type info enable are mostly going to be rarer edge cases.

This first example we could handle though with scope analysis.

bradzacher: This first example of static private member access we could handle though with scope analysis.

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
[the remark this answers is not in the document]
ronami: Yes! I've mentioned this since TypeScript reports it as unused but the rule doesn't reports it as unused ([regardless of the type error](https://deploy-preview-10913--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.tsx&code=MYGwhgzhAEAqCmEAuBGaBvAUNaB6X0EAFgPYCuIAJtAEbzQBO8ADiQ0vNTWUtJSYgB2AciQB%2BbNGYMAlgDcwHWmAbQAvNBQBuTJIBmJEgAoAlBkk4mSMg0HQ9ZQcCQySgo0iIyIALgzKGP0EyAFs6VQBfMywcWMZ4a1toT28AOhoVHTiIyRycoA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaWXWZRAE0IAdoBLANwEMAXRQgY3ieWUIFtE-AEaJoqDJHwAzaZAA04bJAACLAJ7UUHetRaEU8OrhYB6AsVLkqtRq3ZcefQSLESoY6PmiQlAXxA-IA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3QAacDToMmeCmFHopAXxAqgA&tokens=false)).

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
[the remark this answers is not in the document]
ronami: Yes, I agree 👍

bradzacher:
FYI I'm going to make @Josh-Cena very happy - I'm going to build a "class scope analyser" so we can do this right rather than just hacking it in.

This does mean that I'm going to rewrite the rule from the ground up -- but it'll be worht it

kirkwaiblinger:
> FYI I'm going to make @Josh-Cena very happy - I'm going to build a "class scope analyser" so we can do this right rather than just hacking it in.
> 
> This does mean that I'm going to rewrite the rule from the ground up -- but it'll be worht it

Do you intend for this to proceed as-is or should we pause/draft it pending this work?

bradzacher:
Let's draft it.
We need to at least handle the static property case.
Might have some time soon to pick the work back up.

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
```diff
@@ -0,0 +1,60 @@
...
+
+type Options = [];
+export type MessageIds = 'unusedPrivateClassMember';
+
+export default createRule<Options, MessageIds>({
+  name: 'no-unused-private-class-members',
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Disallow unused private class members',
```
ronami, line 14: What do you think about covering the following use-case? ([deploy preview link](https://deploy-preview-10913--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.tsx&code=MYGwhgzhAEAqCmEAuBGaBvAUNaB6X0EAFgPYCuIAJtAEbzQBO8ADiQ0vNTWUtJSYgB2AciQB%2BbNGAlByBmWBI2ACmYMAlgDcwHaADMSJAFzRBZALZ0GASgwBfTA6A&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tidmUQAmtYlQBuAQ3yJaZeOOTJaAW0RKARomioMkTdA7RI4MAF8QJoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false))

```ts
class Test1 {
  // should be reported but doesn't?
  constructor(private foo: number) {}
}
```

packages/eslint-plugin/src/util/class-scope-analyzer/classScopeAnalyzer.ts
```diff
@@ -0,0 +1,565 @@
...
+      ) {
+        const thisContext = upper.findClassScopeWithName(
+          thisType.typeName.name,
+        );
+        if (thisContext != null) {
+          super(scopeManager, upper, thisContext, false);
+          return;
+        }
+      }
+    }
```
ronami, line 597: This doesn't seem to have a test, and it throws for a function declaration/expression with no arguments ([deploy preview link](https://deploy-preview-10913--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.tsx&code=GYVwdgxgLglg9mABMOcAUBKRBvRBfAKCA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tidmUQAmtYlQBuAQ3yJaZeOOTJaAW0RKARomioMkTdA7RI4MAF8QJoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```ts
function foo() { }
```

packages/eslint-plugin/src/util/class-scope-analyzer/classScopeAnalyzer.ts
```diff
@@ -0,0 +1,565 @@
...
+
+        // TODO -- use scope analysis to do some basic type resolution to handle cases like:
+        // ```
+        // class Foo {
+        //   private prop: number;
+        //   method(thing: Foo) {
+        //     // this references the private instance member but not via `this` so we can't see it
+        //     thing.prop = 1;
+        //   }
+        // }
```
ronami: Ironically, it seems like TypeScript incorrectly reports this as unused ([playground link](https://deploy-preview-10913--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.tsx&code=MYGwhgzhAEBiD29oG8BQ1oAcBOBLAbmAC4CmW28mAXNAHYCuAtgEYnYDcq60jJRAFvAAmACgG5aAcxoJ4AShTcM4qQDoclaAF5oARk4YAvqmNA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tidmUQAmtYlQBuAQ3yJaZeOOTJaAW0RKARomioMkTdA7RI4MAF8QJoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```ts
class Foo {
  private prop: number;

  method(thing: Foo) {
    thing.prop = 1;
  }
}
```

Seems like a bug though, unless I'm totally missing something 😅 

---

Edit: Interestingly, seems like the assignment was the culprit, and the following isn't reported by TypeScript ([playground link](https://deploy-preview-10913--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.tsx&code=MYGwhgzhAEBiD29oG8BQ1oAcBOBLAbmAC4CmW28mAXNAHYCuAtgEYnYDcq60jJRAFvAAmACgG5aAcxoJ4AShTcM2PvWy1o4qQDoclThgC%2BqY0A&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tidmUQAmtYlQBuAQ3yJaZeOOTJaAW0RKARomioMkTdA7RI4MAF8QJoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```ts
class Foo {
  private prop: number;

  method(thing: Foo) {
    return thing.prop;
  }
}
```

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
ronami, line 14: I think this is a bug and shouldn't be reported ([deploy preview playground link](https://deploy-preview-10913--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.tsx&code=MYGwhgzhAEAqCmEAuBGaBvAUNaB6X0ATvAA4D2hS8AJtAEYCe0SDJiwhAliUtJNAFcAdgIg1s0ElwBuYKtABmZMgC5oIgLZ14haAB91AkCADcmCSQF0QnYPTCEAFAEoMEnEgAWnCADolZNAA-EEAvNAoZjgAvpixQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tidmUQAmtYlQBuAQ3yJaZeOOTJaAW0RKARomioMkTdA7RI4MAF8QJoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```ts
class Test1 {
  // reported but shouldn't?
  private foo: number | null;

  public bar() {
    this.foo ??= 1;
  }
}
```

bradzacher: This is a weird case that isn't currently reported by `no-unused-vars` but really should be.

[It is reported by the base `no-unused-private-class-members` rule](https://typescript-eslint.io/play/#ts=5.9.2&fileType=.tsx&code=MYGwhgzhAEAqCmEAuBGaBvAUNaB6X0ATvAA4D2hS8AJtAEYCuS0EAFmQyNQHYDkSAfmzQAxADMyZAFzRuDALZ14haAB9ZnEAG5MwkgzogAlsHphCACgCUGYTiSsjEAHTjJ0AQIC80FDpwAvpgBQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaWXWZRAE0IAdoBLANwEMAXRQgY3ieWUIFtE-AEaJoqDJDHR80SODABfEMqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

kirkwaiblinger: There was just a recent eslint bug report about these; I guess it's intentional for no-unused-vars since `x ?? = y` _reads_ `x` first (equivalent to `x != null ? x : (x = y)`) rather than _unconditionally_ assigning (`x = x != null ? x : y`). See https://github.com/eslint/eslint/issues/20029. But presumably the two rules should agree?

kirkwaiblinger: Oh, but do note that these are reported by no-useless-assignment so 🤷‍♂️ 🤷‍♂️ 

([playground](https://typescript-eslint.io/play/#ts=5.9.2&fileType=.tsx&code=KYDwDg9gTgLgBAYwDYEMDOa4BVhpgRjgG8AoOOAegrimElmABM4AjAV3jQAsI2lGAdgHIYAfjJwAxADMIEAFxwBbALYtgUOAB8lfJAG4SEsGxZIAlglYooACgCUxCeRhdzaAHQy5cUaIC8cPiG5AC%2BJOEkoPTw0mwCCDDmEAKIECqQ8Yy2IIrKSEjacHhQ5gIA5gA0cACeiigCNY6k5CC%2BAbWGkdHQsfGJyamyEADMOXl6RSVlVbX1jc0S5tJwOXD%2BgflIi%2BSt650S4eFAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaWXWZRAE0IAdoBLANwEMAXRQgY3ieWUIFtE-AEaJoqDJADuTaLkgAacNjxESZSoWbj0UGXMXKcBYuSS9CPZHQDmuQbha7ps%2BcoC%2BIT0A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false))

packages/eslint-plugin/src/util/class-scope-analyzer/classScopeAnalyzer.ts
bradzacher: also looks like my comment is wrong and I did actually add support for this case haha

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
bradzacher: we should probably file a bug for NUPCM then so it matches NUV?

packages/eslint-plugin/src/util/class-scope-analyzer/extractComputedName.ts
```diff
@@ -0,0 +1,61 @@
...
+): [Key, string] | null {
+  const name = nonComputedName.name;
+  if (nonComputedName.type === AST_NODE_TYPES.PrivateIdentifier) {
+    return [privateKey(nonComputedName), `#${name}`];
+  }
+  return [publicKey(name), name];
+}
+/**
+ * Extracts the string name for a member.
+ * @returns `null` if the name cannot be extracted due to it being a computed.
```
JoshuaKGoldberg: [Docs] Small grammar point:

```suggestion
 * @returns `null` if the name cannot be extracted due to it being computed.
```

packages/eslint-plugin/src/util/class-scope-analyzer/extractComputedName.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function extractNameForMember(node: MemberNode): [Key, string] | null {
+  if (node.computed) {
+    return extractComputedName(node.key);
```
JoshuaKGoldberg, line 81: Just noting, #11006 has some other logic around computed keys. It might be useful to unify?

JoshuaKGoldberg (changes requested):
I have very little to add that Kirk and Ronami didn't already. Very exciting!! ⚡

packages/eslint-plugin/src/rules/no-unused-private-class-members.ts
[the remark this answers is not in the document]
bradzacher: Coming back to this -- my code handles this correctly because realistically it is an actual usage of the parameter.

For example `new Test1().foo()` will return the value of `bar`.
There's an implicit binding here and so I think this is fine to keep _not_ reporting on.

packages/eslint-plugin/src/util/class-scope-analyzer/classScopeAnalyzer.ts
bradzacher: So I didn't handle the case fully, but I just added support for this case

bradzacher:
@ronami / @JoshuaKGoldberg / @kirkwaiblinger 

Added support for:

Parameter properties
```ts
class Test1 {
  constructor(
    // errors
    private parameterPropertyUnused: number,
    // no error
    private parameterPropertyUsed: number,   
  ) {}

  method() {
    return this.parameterPropertyUsed;
  }
}
```

Usage of a private class member via explicitly and simply annotated variables:
```ts
class Foo {
  // no longer reported
  private prop: number;
  // no longer reported
  private staticProp: number;

  method1(thing: Foo) {
    return thing.prop;
  }
  method2(thing: typeof Foo) {
    return thing.staticProp;
  }
}
```

Usage of a private class member via simple assignment:
```ts
class Foo {
  // no longer reported
  private prop: number;
  method() {
    const self = this;
    return self.prop;
  }
}
```

packages/eslint-plugin/src/util/class-scope-analyzer/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  | TSESTree.TSAbstractAccessorProperty
+  | TSESTree.TSAbstractMethodDefinition
+  | TSESTree.TSAbstractPropertyDefinition
+  | TSESTree.TSParameterProperty;
+export type PrivateKey = string & { __brand: 'private-key' };
+export type PublicKey = string & { __brand: 'public-key' };
+export type Key = PrivateKey | PublicKey;
+
+export function publicKey(node: string): PublicKey {
+  return node as PublicKey;
```
JoshuaKGoldberg:
```suggestion
export function publicKey(name: string): PublicKey {
  return name as PublicKey;
```

JoshuaKGoldberg (approved):
Lint complaints notwithstanding, this looks great to me! I'll defer to the other reviewers on the finer points they're calling out. But the overall scope analyzer looks great and I'm very happy with the clean rule implementation. Nicely done! 👏

ronami (approved):
Looking great 👏👏👏

PR #10950, opened 2025-03-14
fix(eslint-plugin): [promise-function-async] use a different error message for functions with promise and non-promise types

### LINKED ISSUE #10687: Bug: `promise-function-async` is reporting on `() => A | Promise<B>`

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=FAMwrgdgxgLglgewgAgEYAoCUyDexnIBOApjGISgLICGMAFgHSHUQAmCAtlsgPzIDkSYv2QAuZAAVCnOAGdiTYrIQAbAG7F0ARkwBuYAF8gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6AewFtLlFaAM1hMy%2BSpya0AhskKj0URNC7RI4MAF8QmoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
function b() {
  return Math.random() ? 'one' : Promise.resolve(1);
}
```

### ESLint Config

```javascript
_playground default_
```

### tsconfig

```jsonc
_playground default_
```

### Expected Result

No reports

### Actual Result

"Functions that return promises must be async."

Attempts to fix to `async function`, changing the return type from `A | Promise<B>` to `Promise<A | B>`

### Additional Info

While this behavior is [described in the docs/demo](https://typescript-eslint.io/rules/promise-function-async/), I believe it's wrong because the function _does not_ "return a promise," it only "MAY return a promise." 

**The change requested is unsafe** as it changes the signature/type of the function entirely.

Note that whether `() => A | Promise<B>` is a good type is not relevant to this issue. The type exists in the wild ([web extensions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage), [screenshot](https://github.com/user-attachments/assets/91c93fc8-a5cf-43e8-85b1-fedcea0b4870)) and it can be prohibited by a separate rule ([suggested here](https://github.com/typescript-eslint/typescript-eslint/issues/9296))

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10687

## Overview

This PR addresses #10687 and adds a custom error message for functions with hybrid return types (and no return type annotation).

Random thought: Should this be a suggestion instead of an auto-fix in this case?


packages/eslint-plugin/src/rules/promise-function-async.ts
```diff
@@ -37,6 +37,8 @@ export default createRule<Options, MessageIds>({
     fixable: 'code',
     messages: {
       missingAsync: 'Functions that return promises must be async.',
+      missingAsyncHybridReturn:
+        'Functions that return promises must be async. Consider adding an explicit return type annotation if the function is intended to return both a promise and a non-promise.',
```
JoshuaKGoldberg, line 41: [Proofreading] I found this a big ambiguous, like a function is returning two things at once?

```suggestion
        'Functions that return promises must be async. Consider adding an explicit return type annotation if the function is intended to return either a promise or a non-promise.',
```

JoshuaKGoldberg (approved):
LGTM! Just one suggestion on the phrasing?

packages/eslint-plugin/src/rules/promise-function-async.ts
ronami: Oh, definitely! Thanks! 🙂

packages/eslint-plugin/src/rules/promise-function-async.ts
```diff
@@ -37,6 +37,8 @@ export default createRule<Options, MessageIds>({
     fixable: 'code',
     messages: {
       missingAsync: 'Functions that return promises must be async.',
+      missingAsyncHybridReturn:
+        'Functions that return promises must be async. Consider adding an explicit return type annotation if the function is intended to return either a promise or a non-promise.',
```
kirkwaiblinger, line 41: (continuing https://github.com/typescript-eslint/typescript-eslint/pull/10950#discussion_r1999100650) Heh, I actually liked the first version better! It now sounds tautologically satisfied, like saying "every person that is named Josh or a non-Josh name is a cool person".

There's some ambiguity in this context whether we're talking about the _return type_ of the function, indicating the set from which all possible invocations of the function's returned value will be drawn, or the _type of a value returned by a specific invocation_. @ronami's version only made sense with the first interpretation (which was my interpretation on first read as well, but clearly not @JoshuaKGoldberg's), and @JoshuaKGoldberg's version only makes sense with the second interpretation.

Maybe some ideas to reduce the ambiguity?

(resolve the ambiguity by making multiple calls explicit)
```suggestion
        'Functions that return promises must be async. Consider adding an explicit return type annotation if the function is intended to sometimes return a promise and sometimes return a non-promise.',
```

(basically @ronami's original version but pluralized, to hint at the intended meaning)
```suggestion
        'Functions that return promises must be async. Consider adding an explicit return type annotation if the function is intended to return both promises and non-promises.',
```

(explicitly mention "return type")
```suggestion
        'Functions that return promises must be async. Consider adding an explicit return type annotation if the function's return type is intended to include both promises and non-promises.',
```

(break out the word "union")
```suggestion
        'Functions that return promises must be async. Consider adding an explicit return type annotation if the function is intended to return a union of promise and non-promise types.',
```

kirkwaiblinger:
> Random thought: Should this be a suggestion instead of an auto-fix in this case?

I personally like the idea of providing two suggestions: 
- add `async`
- add explicit return type

TBH it's suspect to me that this is ever an autofix rather than a suggestion, since I don't like autofixing runtime behavior changes based on type information, but that's neither here nor there.

packages/eslint-plugin/src/rules/promise-function-async.ts
ronami: Hmm I tend to agree with @kirkwaiblinger on this, but I think it's very nuanced. What's your opinion on this @JoshuaKGoldberg?

ronami:
> > Random thought: Should this be a suggestion instead of an auto-fix in this case?
> 
> I personally like the idea of providing two suggestions:
> 
> * add `async` (demoted from autofix)
> * add explicit return type
> 
> TBH it's suspect to me that this is ever an autofix rather than a suggestion, since I don't like autofixing runtime behavior changes based on type information, but that's neither here nor there.

Yes! Though I'm not sure how easy it would be to infer the function's return type.

packages/eslint-plugin/src/rules/promise-function-async.ts
JoshuaKGoldberg: Yeah I can see how both the original and current versions are ambiguous in different ways... Hmm. The _"union"_ mention seems like a must-have to avoid ambiguity then. 👍 from me on that last one.

JoshuaKGoldberg (changes requested):
👍 My vote: mentioning the type being a union, as Kirk suggested?

JoshuaKGoldberg (approved):
Great! Just pinging @kirkwaiblinger to confirm, this looks lovely to me.

Walton Goggins as Rick Hatchett in White Lotus smiling with sunglasses, moving his hands from a prayer symbol to pointing at a person

kirkwaiblinger (approved):
❤️ 🚀

Issue #10174, opened 2024-10-18
Bug: nx 20 breaks `yarn knip`


abrahamguo:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Abug) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Issue Description

Since #10128, `yarn knip` fails locally on my macOS with
```
 NX   Daemon process terminated and closed the connection

Please rerun the command, which will restart the daemon.
If you get this error again, check for any errors in the daemon process logs found in: typescript-eslint/.nx/workspace-data/d/daemon.log
```

The log file has nothing in it that appears to be a failure.

### Reproduction Repository Link

https://github.com/typescript-eslint/typescript-eslint

### Repro Steps

1) clone the repo
2) `yarn`
3) `yarn knip`

### Versions

| package                                | version |
| -------------------------------------- | ------- |
| `node`                                 | `22.9.0` |
| `yarn`     | `3.8.2` |

auvred:
Same thing! It looks like NX is being called from `jest.config.js`.

abrahamguo:
FYI not a permanent solution, but I was able to work around it using `NX_DAEMON=false`) from [knip/727](https://github.com/webpro-nl/knip/issues/727).

JoshuaKGoldberg:
Huh, I get this too. cc @JamesHenry - I've never seen it before with Knip and am confused.

JoshuaKGoldberg:
This no longer reproduces on my machine, and hasn't for a while. 😌

PR #10973, opened 2025-03-19
feat: support stringly-typed extends

### LINKED ISSUE #10934: Enhancement: support ESLint's new `defineConfig` string config references

### Relevant Package

typescript-eslint

### Description

With ESLint v9.22 they added their own `defineConfig` config helper (similar to [`config`](https://typescript-eslint.io/packages/typescript-eslint#config)), which does again support `extends`. I of course immediately tried to rewrite my ESLint config to use the new helper:

```js
import { defineConfig, globalIgnores } from "eslint/config";
import ts from "typescript-eslint";

export default defineConfig(
    {
        plugins: {
            ts: ts.plugin
        },
        extends: [
            "ts/strict",
            "ts/stylistic",
        ],
        rules: {
            "ts/no-inferrable-types": [
                "error",
                { ignoreParameters: true, ignoreProperties: true },
            ],
        },
    },
);
```

Running `npx eslint` fails with the following error:

```
Oops! Something went wrong! :(

ESLint: 9.22.0

TypeError: Plugin config "ts/strict" is an eslintrc config and cannot be used in this context.
    at deepNormalizePluginConfig (C:\Users\niklaswimmer\Dev\test\node_modules\@eslint\config-helpers\dist\cjs\index.cjs:231:9)
    at findPluginConfig (C:\Users\niklaswimmer\Dev\test\node_modules\@eslint\config-helpers\dist\cjs\index.cjs:262:9)
    at C:\Users\niklaswimmer\Dev\test\node_modules\@eslint\config-helpers\dist\cjs\index.cjs:394:25
    at Array.map (<anonymous>)
    at processExtends (C:\Users\niklaswimmer\Dev\test\node_modules\@eslint\config-helpers\dist\cjs\index.cjs:392:36)
    at C:\Users\niklaswimmer\Dev\test\node_modules\@eslint\config-helpers\dist\cjs\index.cjs:464:38
    at Array.flatMap (<anonymous>)
    at processConfigList (C:\Users\niklaswimmer\Dev\test\node_modules\@eslint\config-helpers\dist\cjs\index.cjs:464:20)
    at defineConfig (C:\Users\niklaswimmer\Dev\test\node_modules\@eslint\config-helpers\dist\cjs\index.cjs:499:9)
    at file:///C:/Users/niklaswimmer/Dev/test/eslint.config.mjs?mtime=1741610408364:37:16
```

My understandig is that when given `ts/strict` ESLint tries to access `ts.plugin.configs.strict`, which is not the same as `ts.configs.strict` and not compatible with flat configs.

### Additional Info

I marked this as enhancement and not as bug because the "fix" is to simply use `ts.configs.strict` instead of `"ts/strict"` (and use the full rule namespace instead of just `"ts"`), but I expect the current behavior to be confusing for new users (ESLint shows both options in their [docs](https://eslint.org/docs/latest/use/configure/configuration-files#using-configurations-from-plugins), but if one of them does not work, it will be confusing).

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10934 

## Overview

Previously, 
- eslintrc configs were stored at `packages/eslint-plugin/src/configs`, and were exposed via the `plugin.configs` field
- flat configs were stored at `packages/typescript-eslint/src/configs`, and were exposed via the `tseslint.configs` field.
- a typescript-eslint parser _object_ was created in the typescript-eslint package, is used inside the flat configs, and was exposed via `tseslint.parser`.
- eslintrc configs are stringly typed, so there was no need for exporting configs or parser objects from eslint-plugin (only "exception" was a little `use-at-your-own-risk` config generation function - this was just to share code, not for any runtime reason).

Now,
- eslintrc configs are stored at `packages/eslint-plugin/src/configs/eslintrc`.
- flat configs are stored at `packages/eslint-plugin/src/configs/flat`.
- The `plugin.configs` object contains all the same eslintrc configs _and_ all the flat configs, prefixed with `'flat/`. Conceptually, 
   ```ts
   const plugin = {
      configs: {
          strict: eslintrcStrict,
          'flat/strict': flatStrict,
          // likewise for the rest of the configs...
      }
      // rest of the plugin fields...
   };
   ```
   **_Only_ the eslintrc configs are acknowledged in the publicly exported types**. Therefore, this will be invisible to users in the editor. 
- the `tseslint.configs` _does not_ use the `flat/` prefixes. It _only_ exports eslint v9 flat configs, just like it always has.
- Because the flat configs need a parser object, the eslint-plugin package now creates the parser object. 
- In order for relevant objects in `eslint-plugin` to be `===` to their counterparts in `typescript-eslint` at runtime, the `flatConfigs` and `parser` are now exported from `eslint-plugin` via a new `use-at-your-own-risk` export in order to be re-exported by `typescript-eslint`.

*** 

In summary, these changes ensure:
```ts
const assert = require('node:assert');
const { defineConfig } = require('eslint/config');
const tseslint = require('typescript-eslint');
const tseslintPlugin = require('@typescript-eslint/eslint-plugin');

assert(tseslint.plugin === tseslintPlugin);
assert(tseslint.configs.strict === tseslintPlugin.configs['flat/strict']);
assert(tseslint.parser === tseslintPlugin.configs['flat/strict'][0].parser);

// doesn't throw
defineConfig({
   extends: ['tseslint/strict'],
   plugins: { tseslint: tseslintPlugin },
});

// BUT
assert(tseslint.configs !== tseslintPlugin.configs);
```


package.json
```diff
@@ -148,7 +148,8 @@
     "react-split-pane@^0.1.92": "patch:react-split-pane@npm%3A0.1.92#./.yarn/patches/react-split-pane-npm-0.1.92-93dbf51dff.patch",
     "tmp": "0.2.1",
     "tsx": "^4.7.2",
-    "typescript": "5.8.2"
+    "typescript": "5.8.2",
+    "@eslint/config-helpers": "^0.2.0"
```
kirkwaiblinger, line 152: This manual resolution is required pending release of the next version of eslint (https://github.com/eslint/eslint/pull/19533 has been merged but not yet released)

packages/eslint-plugin/src/raw-plugin.ts
```diff
@@ -0,0 +1,113 @@
+import type { TSESLint } from '@typescript-eslint/utils';
+import type { FlatConfig, Linter } from '@typescript-eslint/utils/ts-eslint';
+
+import * as parserBase from '@typescript-eslint/parser';
```
kirkwaiblinger, line 4: The inclusion of this dependency breaks some of our website mocking, and I'm not clear why. Looking for help with this 🙏 

The repro steps are easy, just run `yarn start`. The issue appears to be due to the module replacement in https://github.com/typescript-eslint/typescript-eslint/blob/2810ceb7dad9dda6c962b63b26c85f3424f01889/packages/website-eslint/build.mts#L126-L136 not succeeding anymore. Ostensibly - something is now imported from typescript-estree that wasn't previously as a result of the dependency graph change 🤷

kirkwaiblinger: Also I'm wondering if we need to change the parser from a peer dependency to a direct dependency?

JoshuaKGoldberg:
```plaintext
Error: Build failed with 1 error:
../parser/src/parser.ts:15:9: ERROR: No matching export in "../typescript-estree/src/use-at-your-own-risk.ts" for import "parseAndGenerateServices"
```

Blagh, yes, I'm stumped 😬. I don't see how this would be working now but not after the mostly-internal-to-`eslint-plugin` changes. I'm also not a high proficiency user of esbuild/etc. Maybe @armano2 or @bradzacher would be able to figure this out?

armano2: to fix this you have to exclude `parser/src/parser.ts` from build, as we do not want to import fs, stream etc.

https://github.com/typescript-eslint/typescript-eslint/blob/3241e46173918b1a981b990895fc122d7ca722c5/packages/website-eslint/build.mts#L94-L100

```
makeFilter([
  '/ts-eslint/ESLint.ts',
  '/ts-eslint/RuleTester.ts',
  '/ts-eslint/CLIEngine.ts',
  '/parser/src/parser.ts',
]),
```

armano2: sadly this is not working completely as esbuild is not correctly replacing imports,

we have to create new file, eg. parser.js and mock out `parseForESLint`
```
export function parseForESLint() {
  // do nothing
}
```
```
          build.onLoad(makeFilter(['/parser/src/parser.ts']), async args => {
            console.log('onLoad:replace', args.path);
            const contents = await requireMock('./src/mock/parser.js');
            return { contents, loader: 'js' };
          });
```

package.json
JoshuaKGoldberg: We're clear to remove this now, right?

tools/scripts/generate-configs.mts
```diff
@@ -215,12 +211,18 @@ async function main(): Promise<void> {
...
+        'configs',
+        'eslintrc',
+        `${name}.ts`,
+      ),
       classicConfigStr,
     );
 
     //
-    // 2. Flat Config - written to the core package
+    // 2. Flat Config - written to the eslint-plugin/src/configs/flat
```
JoshuaKGoldberg, line 225:
```suggestion
    // 2. Flat Config - written to eslint-plugin/src/configs/flat
```

packages/website-eslint/src/mock/parser.js
JoshuaKGoldberg, line 1:
```plaintext
Run yarn knip
Finished packing local packages.
Unused files (1)
packages/website-eslint/src/mock/parser.js
Error: Process completed with exit code 1.
```

tools/scripts/generate-configs.mts
JoshuaKGoldberg, line 1: > `PACKAGES_TYPESCRIPT_ESLINT`

Now that this file doesn't use `PACKAGES_TYPESCRIPT_ESLINT` anymore, we can delete the var in `tools/scripts/paths.mts`. I'm surprised our Knip config didn't pick it up...

JoshuaKGoldberg (approved):
LGTM, really clean swap-out. Nicely done!

Just requesting changes on a passing build & removing an unused var.

Anne Hathaway clapping at an awards show, smiling happily and tearing up

packages/website-eslint/src/mock/parser.js
kirkwaiblinger: I get 
```
Analyzing workspace ....

 NX   Daemon process terminated and closed the connection

Please rerun the command, which will restart the daemon.
```

🤔  Any idea what's going on with that?

packages/eslint-plugin/tests/worksWithStringlyTypedExtends.test.ts
```diff
@@ -0,0 +1,25 @@
+import { defineConfig } from 'eslint/config';
+
+import plugin from '../src/index';
+
+describe("The plugin object should work with eslint defineConfig's stringly typed extends array", () => {
+  it('should not error when explicitly referencing the flat version', () => {
+    defineConfig({
+      extends: ['ts/flat/strict'],
+      plugins: {
+        // @ts-expect-error -- types aren't compatible.
```
bradzacher, line 10: side note:
it's been on my todo list for a while -- we need to just bite the bullet and make them compatible.

bradzacher (approved):

package.json
kirkwaiblinger: well, on main we still have 
```
├─ eslint@npm:9.22.0
│  └─ @eslint/config-helpers@npm:0.1.0 (via npm:^0.1.0)
│
└─ eslint@npm:9.22.0 [43684]
   └─ @eslint/config-helpers@npm:0.1.0 (via npm:^0.1.0)
```

so we need either this or to upgrade eslint past [9.23](https://github.com/eslint/eslint/releases/tag/v9.23.0)

kirkwaiblinger: (I've implemented the latter, updating our eslint dependency)

packages/website-eslint/src/mock/parser.js
kirkwaiblinger: Oh, I see that log in the CI build, should be resolved in latest commit now 👍 . Not sure why I'm unable to run it locally though 🙃

package.json
kirkwaiblinger: but yeah once main has eslint >= 9.23 we can remove it.

JoshuaKGoldberg: Fun. I just clicked the checkmark in #6687 to start the Renovate PR.

packages/website-eslint/src/mock/parser.js
JoshuaKGoldberg: #10174 🤷 

```plaintext
NX_DAEMON=false
```

JoshuaKGoldberg (approved):
Two young cats humorously jumping around and one flipping into a trashcan to get to a cat toy their owner is moving around

bradzacher (approved):

PR #10978, opened 2025-03-21
chore: de-duplicate lint config

43081j:
This just simplifies the lint config slightly by doing the following:

- Separate vitest and jest rules from regular rules, so we can declare the regular rules once for both
- Change `vitestFiles` to be computed in preparation for us moving more packages to vitest (so we don't end up with a bunch of _very similar_ globs)

## PR Checklist

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -406,6 +406,7 @@ export default tseslint.config(
       '@typescript-eslint/no-unsafe-return': 'off',
     },
   },
+  // jest-specific config
   {
     files: [
       'packages/*/tests/**/*.test.{ts,tsx,cts,mts}',
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -29,9 +29,11 @@ const restrictNamedDeclarations = {
   selector: 'ExportNamedDeclaration[declaration=null][source=null]',
 };
 
-const vitestFiles = [
-  'packages/eslint-plugin-internal/tests/**/*.test.{ts,tsx,cts,mts}',
-];
+const vitestPackages = ['eslint-plugin-internal'];
+
+const vitestFiles = vitestPackages.map(
+  name => `packages/${name}/tests/**/*.test.{ts,tsx,cts,mts}`,
+);
 
 export default tseslint.config(
   // register all of the plugins up-front
@@ -392,7 +394,6 @@ export default tseslint.config(
       'packages/integration-tests/tools/integration-test-base.ts',
       'packages/integration-tests/tools/pack-packages.ts',
     ],
-    ignores: vitestFiles,
     rules: {
       '@typescript-eslint/no-empty-function': [
         'error',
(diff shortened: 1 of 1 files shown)


eslint.config.mjs
```diff
@@ -403,6 +404,18 @@ export default tseslint.config(
       '@typescript-eslint/no-unsafe-call': 'off',
       '@typescript-eslint/no-unsafe-member-access': 'off',
       '@typescript-eslint/no-unsafe-return': 'off',
+    },
+  },
+  {
```
kirkwaiblinger, line 409:
```suggestion
  // jest-specific config
  {
```

eslint.config.mjs
```diff
@@ -423,15 +436,6 @@ export default tseslint.config(
   {
```
kirkwaiblinger, line 436: change preceding comment to "vitest-specific config"

kirkwaiblinger (approved):

eslint.config.mjs
```diff
@@ -29,9 +29,11 @@ const restrictNamedDeclarations = {
...
 };
 
-const vitestFiles = [
-  'packages/eslint-plugin-internal/tests/**/*.test.{ts,tsx,cts,mts}',
-];
+const vitestPackages = ['eslint-plugin-internal'];
+
+const vitestFiles = vitestPackages.map(
+  name => `packages/${name}/tests/**/*.test.{ts,tsx,cts,mts}`,
+);
```
kirkwaiblinger, line 36: IMO let's revert this part. It creates merge conflicts on all the vitest PRs for very marginal gain. We can trivially DRY this up once the in-progress PRs are merged.

eslint.config.mjs
```diff
@@ -392,7 +394,6 @@ export default tseslint.config(
       'packages/integration-tests/tools/integration-test-base.ts',
       'packages/integration-tests/tools/pack-packages.ts',
     ],
-    ignores: vitestFiles,
```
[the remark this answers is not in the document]
kirkwaiblinger: TBH I'm kinda -1 on this. It's true that it's DRYer in the moment but the duplication will go away on its own once everything is in vitest, and if we need there to be drift between the jest/vitest parts then this makes it less flexible. So I think as-is makes good sense as a transitional approach, but it's marginal difference either way 🤷  Can go either way if others have opinions.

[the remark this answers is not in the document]
43081j: 🤷‍♂️ I'm not tied to it enough to mind either way. It's very unlikely we will alter these rules between test frameworks since they're more to do with syntax and casts

Eventually the vitest rules merge into this block, and you remove the jest block. Or your duplicate in both and remove the just block. 

We achieve the same either way, just we duplicate the rules in the latter

eslint.config.mjs
43081j: Otherwise we're repeating somewhat lengthy globs an increasing amount over time. Hoping they stay in sync

I'd rather fix this now while there's only one instead of doing it later when there's many

43081j:
Given you disagree with both parts of this pr, it seems pretty pointless now

I disagree with you on both counts for what it's worth. We shouldn't be duplicating rules, that's just a sign we got it wrong. And we shouldn't be duplicating globs because it increases maintenance burden as we add more and more of what's basically the same glob. It doesn't make sense to do this, it makes sense to have one glob that covers many packages

kirkwaiblinger:
Closing since I think this is fully superseded by the changes now on main (notably https://github.com/typescript-eslint/typescript-eslint/pull/11078)

PR #10991, opened 2025-03-23
feat(website): Show tsconfig parsing errors in tab

### LINKED ISSUE #8580: Enhancement: Display tsconfig errors in playground

### Relevant Package

website

### Description

It would be nice if playground displays tsconfig error messages in the "Errors" tab. 
Currently, the "Errors" tab shows a "All is ok!" message, but the browser console shows an error. 
- [playground](https://typescript-eslint.io/play/#ts=5.2.2&fileType=.tsx&code=FBA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3QAacDUiZ0JWKgBu6AFLyiAWVgATDugphR6KRCkBfEGaA&tokens=false)
스크린샷 2024-03-01 오전 12 27 24

This makes it difficult to accurately reproduce the issue. (https://github.com/typescript-eslint/typescript-eslint/issues/7693#issuecomment-1830977079)

### Additional Info

_No response_

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8580

## Overview

WIP. 
I'm trying to manage tsconfig errors with different markers and statuses.

Because I couldn't integrate tsconfig errors into markers since markers change depending on editor changes, and I thought it was wrong to apply tsconfig where markers are created.

### DIFF
--- a/packages/website/src/components/editor/LoadedEditor.tsx
+++ b/packages/website/src/components/editor/LoadedEditor.tsx
@@ -83,7 +83,11 @@ export const LoadedEditor: React.FC<LoadedEditorProps> = ({
     const markers = monaco.editor.getModelMarkers({
       resource: model.uri,
     });
-    onMarkersChange(parseMarkers(markers, codeActions, editor));
+
+    const test = parseMarkers(markers, codeActions, editor);
+
+    console.log(test);
+    // onMarkersChange(parseMarkers(markers, codeActions, editor));
   }, [codeActions, onMarkersChange, editor, monaco.editor]);
 
   useEffect(() => {
--- a/packages/website/src/components/linter/createLinter.ts
+++ b/packages/website/src/components/linter/createLinter.ts
@@ -19,7 +19,7 @@ import { createEventsBinder } from '../lib/createEventsBinder';
 import { parseESLintRC, parseTSConfig } from '../lib/parseConfig';
 import { defaultEslintConfig, PARSER_NAME } from './config';
 import { createParser } from './createParser';
-
+import { ErrorGroup } from '../../../../website/src/components/types';
 export interface CreateLinter {
   configs: string[];
   onLint(cb: LinterOnLint): () => void;
@@ -42,6 +42,7 @@ export function createLinter(
   system: PlaygroundSystem,
   webLinterModule: WebLinterModule,
   vfs: typeof tsvfs,
+  onMarkersChange: (value: ErrorGroup[]) => void,
 ): CreateLinter {
   const rules: CreateLinter['rules'] = new Map();
   const configs = new Map(Object.entries(webLinterModule.configs));
@@ -160,7 +161,31 @@ export function createLinter(
       console.log('[Editor] Updating', fileName, compilerOptions);
       parser.updateConfig(compilerOptions);
     } catch (e) {
-      console.error(e);
+      const errors = {
+        group: 'Typescript',
+        items: e.message
+          .trim()
+          .split('\n')
+          .map((message: string) => {
+            return {
+              fixer: undefined,
+              location: '',
+              message,
+              severity: 8,
+              suggestions: [],
+            };
+          }),
+        uri: undefined,
+      };
+
+      console.log(onMarkersChange);
+      console.log(errors);
+
+      onMarkersChange([errors]);
+
+      // const marker =
+
+      // console.error("asdf",e);
     }
   };
 
--- a/packages/website/src/components/ErrorsViewer.tsx
+++ b/packages/website/src/components/ErrorsViewer.tsx
@@ -133,6 +133,8 @@ export function ErrorViewer({
 export function ErrorsViewer({ value }: ErrorsViewerProps): React.JSX.Element {
   const [isLocked, setIsLocked] = useState(false);
 
+  console.log(value);
+
   useEffect(() => {
     setIsLocked(false);
   }, [value]);
--- a/packages/website/src/components/Playground.tsx
+++ b/packages/website/src/components/Playground.tsx
@@ -39,7 +39,7 @@ function Playground(): React.JSX.Element {
   const [tsVersions, setTSVersion] = useState<readonly string[]>([]);
   const [selectedRange, setSelectedRange] = useState<SelectedRange>();
   const [position, setPosition] = useState<number>();
-  const [activeTab, setTab] = useState<TabType>('code');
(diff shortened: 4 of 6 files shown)


armano2:
maybe we should just extends fs to contain info about errors for specific files, with that we could change/swap what we display based on file that is selected?

developer-bandi:
> 어쩌면 특정 파일의 오류 정보를 포함하도록 fs를 확장한 다음, 선택된 파일에 따라 표시하는 내용을 변경/교환할 수 있지 않을까요?
> 
> tsconfig는 모나코의 파일(모델)로 다른 파일과 유사하며 vfs 파일로 복제됩니다.

If I understand correctly, you mean changing the type of markers to `Record<TabType,ErrorGroup[]>` and displaying the corresponding error depending on the selected tab?

packages/website/src/components/linter/createLinter.ts
```diff
@@ -160,7 +167,37 @@ export function createLinter(
       console.log('[Editor] Updating', fileName, compilerOptions);
       parser.updateConfig(compilerOptions);
     } catch (e) {
-      console.error(e);
+      if (e instanceof Error) {
+        const error = {
+          group: 'Typescript',
```
JoshuaKGoldberg, line 172:
```suggestion
          group: 'TypeScript',
```

packages/website/src/components/linter/createLinter.ts
```diff
@@ -160,7 +167,37 @@ export function createLinter(
...
+            }),
+          uri: undefined,
+        };
+
+        onMarkersChange(prev => {
+          const activeTabErrors = Object.fromEntries(
+            prev.tsconfig.map(error => [error.group, error]),
+          );
+
+          activeTabErrors.Typescript = error;
```
JoshuaKGoldberg, line 193: Here and in other places that refer to `Typescript`:

```suggestion
          activeTabErrors.TypeScript = error;
```

packages/website/src/components/linter/createLinter.ts
```diff
@@ -160,7 +167,37 @@ export function createLinter(
...
+          group: 'Typescript',
+          items: e.message
+            .trim()
+            .split('\n')
+            .map((message: string) => {
+              return {
+                fixer: undefined,
+                location: '',
+                message,
+                severity: 8,
```
JoshuaKGoldberg, line 181: [Docs] It's not clear where this magic number comes from at first. Keeping in line with... https://github.com/typescript-eslint/typescript-eslint/blob/ee38b525449513ce56632aee4a7a8d40871672ee/packages/website/src/components/linter/utils.ts#L144-L146

```suggestion
                severity: 8, // MarkerSeverity.Error
```

packages/website/src/components/linter/createLinter.ts
```diff
@@ -160,7 +167,37 @@ export function createLinter(
...
       parser.updateConfig(compilerOptions);
     } catch (e) {
-      console.error(e);
+      if (e instanceof Error) {
+        const error = {
+          group: 'Typescript',
+          items: e.message
+            .trim()
+            .split('\n')
+            .map((message: string) => {
```
JoshuaKGoldberg, line 176: Do you have an example of a message that needs to be split? I've only been able to make versions with a single reported error at a time.

packages/website/src/components/editor/LoadedEditor.tsx
```diff
@@ -83,8 +83,28 @@ export const LoadedEditor: React.FC<LoadedEditorProps> = ({
...
-    onMarkersChange(parseMarkers(markers, codeActions, editor));
-  }, [codeActions, onMarkersChange, editor, monaco.editor]);
+
+    const errors = parseMarkers(markers, codeActions, editor);
+
+    onMarkersChange(prev => {
+      const tsconfigError =
+        activeTab === 'tsconfig' &&
+        Object.fromEntries(prev[activeTab].map(error => [error.group, error]))
+          .Typescript;
```
JoshuaKGoldberg, line 93: [Refactor] This confused me at first - we don't really need a whole grouped object, just the first tsconfig error.

```suggestion
        Object.values(prev[activeTab]).find(
          error => error.group === 'TypeScript',
        );
```

In fact, we don't care about this at all if `errors.length`, so this callback body can be rewritten like:

```ts
const tsconfigErrors =
  activeTab === 'tsconfig' &&
  !errors.length &&
  Object.values(prev[activeTab]).filter(
    error => error.group === 'TypeScript',
  );

return {
  ...prev,
  [activeTab]: tsconfigErrors || errors,
};
```

JoshuaKGoldberg (changes requested):
Thanks for getting this started! I think I found a bug? Also left a question and a few refactor suggestions.

packages/website/src/components/linter/createLinter.ts
developer-bandi: I'm not sure if this is realistic, but I used split because I was able to generate two errors with the settings below.

```
{
  "compilerOptions": {
    "moduleResolution": "classic",
    "resolveJsonModule": true,
    "incremental": true,
    "outFile": "dist/bundle.js"
  }
}
```

스크린샷 2025-04-12 오전 11 12 03

packages/website/src/components/editor/LoadedEditor.tsx
developer-bandi: The code you suggested is much clearer, thank you

packages/website/src/components/linter/createLinter.ts
```diff
@@ -160,7 +167,37 @@ export function createLinter(
...
+              return {
+                fixer: undefined,
+                location: '',
+                message,
+                severity: 8,
+                suggestions: [],
```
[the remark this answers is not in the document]
developer-bandi: it seems good. After applying defensive measures to the actual usage area, the type was also changed to optional.

packages/website/src/components/editor/LoadedEditor.tsx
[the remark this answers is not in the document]
developer-bandi: This was an error that occurred because the marker was not updated at the point where the error disappeared. I fixed it.

kirkwaiblinger (approved):
Seems to work!

@JoshuaKGoldberg

JoshuaKGoldberg (approved):
Lovely, thanks!

PR #10998, opened 2025-03-26
fix(eslint-plugin): [no-deprecated] report on deprecated imported variable used as property

### LINKED ISSUE #10828: Bug: [no-deprecated] Deprecated imported variable used in an object not detected/reported

### Issue Description

When using an imported deprecated variable inside an object to assign it to a field, the usage is not detected/reported by the rule, both in CLI (`npx eslint`) and the VSCode ESLint extension.

When the  variable is not imported, declared in the same file, the issue is reported correctly.
When using an intermediate variable, the issue is reported correctly.

The original case I was handling was in an Angular module `forRoot` method returning a `ModuleWithProviders` configuration object, so as to bind the now deprecated [`NGXS_PLUGINS` injection token from the NGXS library (since v18.1.6)](https://github.com/ngxs/store/blob/d290a0ef8eb500aa2089ab80c03835a27ba80c29/packages/store/plugins/src/symbols.ts#L19) to an implementation service/provider.

https://github.com/StephaneSeyvoz/typescript-eslint-no-deprecated

### Repro Steps

1) clone the repo
2) `npm install`
3) `npx eslint` or open in VSCode with the ESLint extension
```
src\main.ts
  18:18  error  `deprecatedVar` is deprecated. For some reason; Usually would be an Angular injection token        @typescript-eslint/no-deprecated  
  24:26  error  `deprecatedModuleVar` is deprecated. For some reason; Usually would be an Angular injection token  @typescript-eslint/no-deprecated  
  45:16  error  `deprecatedVar` is deprecated. For some reason; Usually would be an Angular injection token        @typescript-eslint/no-deprecated  
  52:24  error  `deprecatedModuleVar` is deprecated. For some reason; Usually would be an Angular injection token  @typescript-eslint/no-deprecated  

✖ 4 problems (4 errors, 0 warnings)
```

-> Lines 10 and 36 are not detected/reported
-> In VSCode, lines 10 and 36 are shown as deprecated by `ts(6385)`, not ESLint

### Versions

| `@typescript-eslint/eslint-plugin`     | `8.24.0` |
| `@typescript-eslint/parser`            | `8.24.0` |
| `@typescript-eslint/scope-manager`     | `8.24.0` |
| `@typescript-eslint/typescript-estree` | `8.24.0` |
| `@typescript-eslint/type-utils`        | `8.24.0` |
| `@typescript-eslint/utils`             | `8.24.0` |
| `TypeScript`                           | `5.7.3` |
| `ESLint`                               | `9.20.1` |
| `node`                                 | `20.16.0` |

### Quick self-analysis

I tried tracking the issue in the [no-deprecated rule source for tag 8.24.0](https://github.com/typescript-eslint/typescript-eslint/blob/v8.24.0/packages/eslint-plugin/src/rules/no-deprecated.ts).
It fails in `getDeprecationReason` while considering the `node.parent.type` as a property.

When the variable is local / not imported, the `getJsDocDeprecation(propertySymbol)` call at line 353 succeeds.
It returns `undefined` when the variable is imported.

In the debugger, evaluating `searchForDeprecationInAliasesChain` the same way as line 358 returns the right deprecation reason.
This seems related, as the comment above `searchForDeprecationInAliasesChain` says:

>    // Deprecated jsdoc tags can be added on some symbol alias, e.g.
>    //
>    // export { /** @deprecated */ foo }
>    //
>    // When we import foo, its symbol is an alias of the exported foo (the one
>    // with the deprecated tag), which is itself an alias of the original foo.
>    // Therefore, we carefully go through the chain of aliases and check each
>    // immediate alias for deprecated tags

So I guess it's just about applying the right behaviour with imported properties.

### Checking for duplicates

I found https://github.com/typescript-eslint/typescript-eslint/issues/10643, but:
- it seems different since no import is used there,
- the issue isn't here when the variable is local.

so I considered it was a different problem.

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10828

## Overview

This PR addresses #10828 and adjusts the rule to also check the alias chain for a possible deprecation.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-deprecated.ts
+++ b/packages/eslint-plugin/src/rules/no-deprecated.ts
@@ -351,6 +351,7 @@ export default createRule<Options, MessageIds>({
           propertySymbol?.valueDeclaration,
         );
         return (
+          searchForDeprecationInAliasesChain(propertySymbol, true) ??
           getJsDocDeprecation(property) ??
           getJsDocDeprecation(propertySymbol) ??
           getJsDocDeprecation(valueSymbol)
--- a/packages/eslint-plugin/tests/rules/no-deprecated.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-deprecated.test.ts
@@ -2075,6 +2075,25 @@ exists('/foo');
     },
     {
       code: `
+        import { deprecatedVariable } from './deprecated';
+
+        const test = {
+          someField: deprecatedVariable,
+        };
+      `,
+      errors: [
+        {
+          column: 22,
+          data: { name: 'deprecatedVariable' },
+          endColumn: 40,
+          endLine: 5,
+          line: 5,
+          messageId: 'deprecated',
+        },
+      ],
+    },
+    {
+      code: `
         import { normalFunction } from './deprecated';
 
         const foo = normalFunction;


JoshuaKGoldberg (approved):
👍 LGTM, thanks!

PR #11005, opened 2025-03-28
fix(eslint-plugin): [unified-signatures] exempt `this` from optional parameter overload check

### LINKED ISSUE #10982: Bug: [unified-signatures] thinks overloads with and without this type annotation can be merged

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.tsx&code=KYDwDg9gTgLgBDAnmYcAKUIFsCWBnYAdRxgAsAlYPCAGwDdgo8AeAFQD44BeOAbwCg4cAPTC4AOkn8Avv37CAVAsEK4AVQJQaAQwB2AEzg4sYGsCzBdMbTBwRdcCADM4AbQzZ8wcQHcSFKloGJgBdXgABGhxdAGs4UhgYMDwALlF9YAYaCBQocSwIAC8cGh1xaABzYUsAWjUAZWF9CABjPGFCYAAjYQApbTptepaoHDAYYUonRksW4GEAcWyu7RoAfQB5LoArYBaYdo9cAmE-MkpqekY8aXEVOA3dObgfVABXMAqobQyECDgAHIQX4AJhBABoXqgWno4Hgzi1SH8EKRUF03iVbLo7nAFMJ%2BKBILA4E43k9bPYXv4LkFrmx2AAKACUKXQmGORGpgSuTHpBPA0HgpPJdgcZwCl2CLA4DMEKPwrKOXgAwvY8DAoG99tB%2BCy2Z4CMRztypXzCYKSWT9qKqcbJXSZXKyAr9RzVbp1ZrtVA4AAfOBkjJOaLAfS6xXsrxGiW03kcPhyqDAGBvKAOARCISiCRSISyWRAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJygM0sQBNaySgHMmAQ3yxoKdFETRoAe2iRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
export type PromiseWithResolvers<T> = {
  // ...
}

/**
 * Userland implementation of [Promise.withResolvers]{@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers}.
 * Once we upgrade to Node 22, we can switch to the builtin.
 */
export function withResolvers<T>(): PromiseWithResolvers<T>
export function withResolvers<T>(
  this: PromiseConstructor
): PromiseWithResolvers<T>
export function withResolvers<T>(
  this: PromiseConstructor | undefined
): PromiseWithResolvers<T> {
  return {
    // ...
  }
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/unified-signatures": "error",
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

It's not possible to merge these overloads, so I expected no error

### Actual Result

These overloads can be combined into one signature with an optional parameter. 11:3 - 11:27

### Additional Info

_No response_

jedwards1211:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10982

## Overview

In `signaturesDifferByOptionalOrRestParameter`, checks the first parameter of both signatures.  If one is a `this` parameter but not the other, then `return undefined` instead of `kind: 'extra-parameter'`, so that no error is flagged.


JoshuaKGoldberg (approved):
💯 Great first PR, thanks!

packages/eslint-plugin/src/rules/unified-signatures.ts
```diff
@@ -310,6 +314,12 @@ export default createRule<Options, MessageIds>({
       const shorter = sig1.length < sig2.length ? sig1 : sig2;
       const shorterSig = sig1.length < sig2.length ? a : b;
 
+      // If one signature has explicit this type and another doesn't, they can't
+      // be unified.
+      if (isThisParam(sig1[0]) !== isThisParam(sig2[0])) {
```
kirkwaiblinger, line 319: Question -

```suggestion
      if (isThisParam(sig1[0]) || isThisParam(sig2[0])) {
```

I was kind of thinking we would also want to exempt cases where they both include `this`, since `this: Foo | void` isn't equivalent to the overload form ([playground](https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=GYVwdgxgLglg9mABHAbgUwE4Bs4EMAma%2BAKgBYwDOAanDPgBRTkUBciKt%2BAlGx3QNwAoUJFgJk6bHkIlmNOo2ZsKUDDDABzHu05CR0eElSYcBImUryGXRAG9EAX0HDwB8eEMB1GEyuLKyqrqGogAPjp02nz4do7OEAgqyIgAvLHGUmaylpwANIgeCN6%2BnHGCGaYyFtSc9FxCcAB0FdLmcrX15ZKVbTl0jRC4WFj0GGhQIBhgNQp1qQB8iADkOFhLXJ2ChWDFpH6dTdu7%2B0JHPnucA0MjYxNTMwxzKYsrcGsbQi6ihoi3k9N%2BCAAIzYT0W0SipVsgkQv3G-0QwLqQicgjQAA8AA5wDBQWIOfhAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6WJygM0sQBNaySgHMmAQ3yxoKdFADuY6E0jgwAXxBqgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false)). Is that not the case?

jedwards1211: Hmm but using different logic, we could also exempt cases where some but not all signatures have `this: void` without breaking the above example

jedwards1211: @kirkwaiblinger btw does `returnsVoid` (especially calling it with a function argument) do something tricky in your example that I'm not aware of?  It seems to me instead of

```ts
overloadedThisVoid.call(returnsVoid(() => 'lol'));
```

you could just do
```ts
overloadedThisVoid.call(1 as any as void)
```

Which triggers the same TypeScript error.

kirkwaiblinger: Hmm, yeah, I guess it would come down to special-casing around `void`, since, in addition to your example, which I agree with, we'd also want to consider:
```ts
function foo(this: Foo, x: number)
function foo(this: Bar, x: number)
```
can still be
```ts
function foo(this: Foo | Bar, x: number)
```

Wonder if special-casing around `void` is even relevant for non-`this` parameters too?

```ts
function foo(x: void);
function foo(x: number);
```

Would have to play around to figure that one out.

jedwards1211: Yes the `Foo | Bar` case was another one I had in mind.

`void` on non-`this` parameters doesn't seem like it would cause any problems with unifying signatures, but interestingly, it does allow trailing parameters to be omitted:

```ts
function foo(x: void): void;
function foo(x: number): void;
function foo() {}

foo()
foo((() => {})())
foo(1)

function bar(x: void | number) {}

bar()
bar((() => {})())
bar(1)
```

kirkwaiblinger: > @kirkwaiblinger btw does `returnsVoid` (especially calling it with a function argument) do something tricky in your example that I'm not aware of? It seems to me instead of
> 
> ```ts
> overloadedThisVoid.call(returnsVoid(() => 'lol'));
> ```
> 
> you could just do
> 
> ```ts
> overloadedThisVoid.call((() => {})())
> ```
> 
> Or more explicitly
> 
> ```ts
> overloadedThisVoid.call(1 as any as void)
> ```
> 
> Both of which trigger the same TypeScript error.

Sort of - the point of those in the issue was to demonstrate how you could sneak _any_ runtime value you like into `void` without any TS errors or type assertions. 

So, 
```ts
overloadedThisVoid.call((() => {})())
```
doesn't demonstrate that, since the runtime value actually is `undefined`, and 
```ts
overloadedThisVoid.call(1 as any as void)
```
doesn't demonstrate that, due to the type assertions.

It isn't as pertinent to this discussion anymore, yeah, since we're now just talking about what happens when the `void` type is used, not whether using `void` type is unsafe. But that's just the examples that have been on my mind 🙂

jedwards1211: It appears that TS picks the `this` type from the last overload, lol.

```
function foo(this: string): void;
function foo(this: number): void;
function foo() {}

foo.call(1)
foo.call('a') // Argument of type 'string' is not assignable to parameter of type 'number'.(2345)

function bar(this: number): void;
function bar(this: string | boolean): void;
function bar() {}

bar.call(1) // Argument of type 'number' is not assignable to parameter of type 'string'.(2345)
bar.call('a')
bar.call(true)
```

jedwards1211: ah right, the unsoundness of TS function typing rears its head again

kirkwaiblinger: > It appears that TS picks the `this` type from the last overload when it comes to `.call`, lol.

Huh. 👀 . Well - that the `.bind()`/`.call()`/`.apply()` types are going over my head at this point 😆

kirkwaiblinger: > Do you know why TS doesn't treat embedded void expressions as an error the way `no-confusing-void-expression` does? Would be curious to know the explicit rationale from the TS team.

Not off the top of my head, but I'm sure you'll get (or more likely be pointed to) a thorough answer if you ask in the TS discord.

kirkwaiblinger (approved):

packages/eslint-plugin/src/rules/unified-signatures.ts
```diff
@@ -302,16 +302,18 @@ export default createRule<Options, MessageIds>({
         : undefined;
     }
 
-    function isThisParam(
-      param: TSESTree.Parameter | undefined,
-    ): param is TSESTree.Identifier {
-      return param?.type === AST_NODE_TYPES.Identifier && param.name === 'this';
+    function isThisParam(param: TSESTree.Parameter | undefined): boolean {
```
kirkwaiblinger, line 305: FYI/PSA - I changed this not to be a type predicate since it's only valid in the `true` branch, not the `false` branch (the function returning `false` does not imply that `param` is _not_ a `TSESTree.Identifier`).

kirkwaiblinger (approved):

packages/eslint-plugin/src/rules/unified-signatures.ts
jedwards1211: Oh I didn't think about TS doing a negative refinement when the function returns `false`.  It would be useful to be able to declare an assertion that only refines in the `true` case

Issue #10740, opened 2025-01-29
Enhancement: [no-floating-promises] allowForKnownSafeCalls for test.only


koooge:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [x] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [x] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [x] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### My proposal is suitable for this project

- [x] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Link to the rule's documentation

https://typescript-eslint.io/rules/no-floating-promises/#allowforknownsafepromises

### Description

Hi there,
`@typescript-eslint/no-floating-promises` 's allowForKnownSafePromiseReturns !8404 works well for it from `node:test`. However, it doesn't help for `it.only` and `it.skip`.  They are handy during debugging so it would be great to have the ability to ignore them as well.

```js
  "allowForKnownSafePromiseReturns": [
    { "from": "package", "name": "it", "package": "node:test" }
  ]
```

### Fail

```TypeScript
import { it } from "node:test";

it.only("...", () => { /* ... */ });

it.skip("...", () => { /* ... */ });
```

### Pass

```TypeScript
import { it } from "node:test";

// eslint-disable-next-line @typescript-eslint/no-floating-promises
it.only("...", () => { /* ... */ });

// eslint-disable-next-line @typescript-eslint/no-floating-promises
it.skip("...", () => { /* ... */ });
```

JoshuaKGoldberg:
👍 this makes sense to me as a feature request. The difficult point here is that we don't yet have a way in the shared [TypeOrValueSpecifier format](https://typescript-eslint.io/packages/type-utils/type-or-value-specifier) to specify a _property_ of something imported by a package.

Copying the current interface from https://typescript-eslint.io/packages/type-utils/type-or-value-specifier#packagespecifier:

```ts
interface PackageSpecifier {
  from: 'package';
  name: string[] | string;
  package: string;
}
```

How do we think this could be augmented to allow both `it` and `it.only`? A couple of starting ideas...


```js
[...]

PR #11006, opened 2025-03-28
fix(eslint-plugin): [no-deprecated] support for literal member access (#10958)

### LINKED ISSUE #10958: Bug: [no-deprecated] doesn't work in `eslint.config.ts` for deprecated configuration name

### Issue Description

`eslint-plugin-unicorn` since version 57 deprecated the old config name for use in flat config: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v57.0.0/index.d.ts#L8-L12

TypeScript reports this, but the `no-deprecated` rule does not work in `eslint.config.ts` for the deprecated config name.

https://stackblitz.com/edit/vitejs-vite-hq4ykaue?file=eslint.config.ts

### Repro Steps

1. Open the reproduction link
2. Open the second terminal at the bottom
3. `npm run lint`

### Versions

| `@typescript-eslint/eslint-plugin`     | `8.26.1` |
| `@typescript-eslint/parser`            | `8.26.1` |
| `@typescript-eslint/scope-manager`     | `8.26.1` |
| `@typescript-eslint/typescript-estree` | `8.26.1` |
| `@typescript-eslint/type-utils`        | `8.26.1` |
| `@typescript-eslint/utils`             | `8.26.1` |
| `TypeScript`                           | `5.8.2` |
| `ESLint`                               | `9.22.0` |
| `node`                                 | `18.20.3` |

undsoft:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10958

## Overview

Adds support for literal member access like `a['b']`

### DIFF
--- a/packages/eslint-plugin/src/rules/no-deprecated.ts
+++ b/packages/eslint-plugin/src/rules/no-deprecated.ts
@@ -10,13 +10,14 @@ import {
   createRule,
   getParserServices,
   nullThrows,
-  typeOrValueSpecifiersSchema,
   typeMatchesSomeSpecifier,
+  typeOrValueSpecifiersSchema,
 } from '../util';
 
 type IdentifierLike =
   | TSESTree.Identifier
   | TSESTree.JSXIdentifier
+  | TSESTree.Literal
   | TSESTree.PrivateIdentifier
   | TSESTree.Super;
 
@@ -209,13 +210,9 @@ export default createRule<Options, MessageIds>({
       return displayParts ? ts.displayPartsToString(displayParts) : '';
     }
 
-    type CallLikeNode =
-      | TSESTree.CallExpression
-      | TSESTree.JSXOpeningElement
-      | TSESTree.NewExpression
-      | TSESTree.TaggedTemplateExpression;
+    type CalleeNode = TSESTree.Expression | TSESTree.JSXTagNameExpression;
 
-    function isNodeCalleeOfParent(node: TSESTree.Node): node is CallLikeNode {
+    function isNodeCalleeOfParent(node: TSESTree.Node): node is CalleeNode {
       switch (node.parent?.type) {
         case AST_NODE_TYPES.NewExpression:
         case AST_NODE_TYPES.CallExpression:
@@ -232,7 +229,7 @@ export default createRule<Options, MessageIds>({
       }
     }
 
-    function getCallLikeNode(node: TSESTree.Node): CallLikeNode | undefined {
+    function getCallLikeNode(node: TSESTree.Node): CalleeNode | undefined {
       let callee = node;
 
       while (
@@ -245,7 +242,7 @@ export default createRule<Options, MessageIds>({
       return isNodeCalleeOfParent(callee) ? callee : undefined;
     }
 
-    function getCallLikeDeprecation(node: CallLikeNode): string | undefined {
+    function getCallLikeDeprecation(node: CalleeNode): string | undefined {
       const tsNode = services.esTreeNodeToTSNodeMap.get(node.parent);
 
       // If the node is a direct function call, we look for its signature.
@@ -254,7 +251,10 @@ export default createRule<Options, MessageIds>({
         'Expected call like node to have signature',
       );
 
-      const symbol = services.getSymbolAtLocation(node);
+      const nodeToFind =
+        node.type === AST_NODE_TYPES.MemberExpression ? node.property : node;
+
+      const symbol = services.getSymbolAtLocation(nodeToFind);
       const aliasedSymbol =
         symbol != null && tsutils.isSymbolFlagSet(symbol, ts.SymbolFlags.Alias)
           ? checker.getAliasedSymbol(symbol)
@@ -334,14 +334,16 @@ export default createRule<Options, MessageIds>({
 
       if (
         node.parent.type === AST_NODE_TYPES.JSXAttribute &&
-        node.type !== AST_NODE_TYPES.Super
+        node.type !== AST_NODE_TYPES.Super &&
+        node.type !== AST_NODE_TYPES.Literal
       ) {
         return getJSXAttributeDeprecation(node.parent.parent, node.name);
       }
 
       if (
         node.parent.type === AST_NODE_TYPES.Property &&
-        node.type !== AST_NODE_TYPES.Super
+        node.type !== AST_NODE_TYPES.Super &&
+        node.type !== AST_NODE_TYPES.Literal
       ) {
         const property = services
           .getTypeAtLocation(node.parent.parent)
@@ -401,6 +403,7 @@ export default createRule<Options, MessageIds>({
           checkIdentifier(node);
         }
       },
+      'MemberExpression > Literal': checkIdentifier,
       PrivateIdentifier: checkIdentifier,
       Super: checkIdentifier,
     };
@@ -414,6 +417,10 @@ function getReportedNodeName(node: IdentifierLike): string {
 
   if (node.type === AST_NODE_TYPES.PrivateIdentifier) {
     return `#${node.name}`;
+  }
+
+  if (node.type === AST_NODE_TYPES.Literal) {
+    return String(node.value);
   }
 
   return node.name;
--- a/packages/eslint-plugin/tests/rules/no-deprecated.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-deprecated.test.ts
@@ -45,7 +45,31 @@ ruleTester.run('no-deprecated', rule, {
         /** @deprecated */ c: 2,
       };
 
+      a['b'];
+    `,
+    `
+      const a = {
+        b: 1,
+        /** @deprecated */ c: 2,
+      };
+
+      a['b' + 'c'];
+    `,
+    `
+      const a = {
+        b: 1,
+        /** @deprecated */ c: 2,
+      };
+
       a?.b;
+    `,
+    `
+      const a = {
+        b: 1,
+        /** @deprecated */ c: 2,
+      };
+
+      a?.['b'];
     `,
     `
       declare const a: {
@@ -56,9 +80,51 @@ ruleTester.run('no-deprecated', rule, {
       a.b;
     `,
     `
+      declare const a: {
+        b: 1;
+        /** @deprecated */ c: 2;
+      };
+
+      a['b'];
+    `,
+    `
       class A {
         b: 1;
         /** @deprecated */ c: 2;
+      }
+
+      new A().b;
+    `,
+    `
+      class A {
+        b: 1;
+        /** @deprecated */ c: 2;
+      }
+
+      new A()['b'];
+    `,
+    `
+      class A {
+        c: 1;
+      }
+      class B {
+        /** @deprecated */ c: 2;
+      }
+
+      new A()['c'];
+    `,
+    `
+      class A {
+        b: () => {};
+        /** @deprecated */ c: () => {};
+      }
+
+      new A()['b']();
+    `,
+    `
+      class A {
+        accessor b: 1;
+        /** @deprecated */ accessor c: 2;
       }
 
       new A().b;
@@ -69,7 +135,7 @@ ruleTester.run('no-deprecated', rule, {
         /** @deprecated */ accessor c: 2;
       }
 
-      new A().b;
+      new A()['b'];
     `,
     `
       declare class A {
@@ -83,11 +149,29 @@ ruleTester.run('no-deprecated', rule, {
     `
       declare class A {
         /** @deprecated */
+        static b: string;
+        static c: string;
+      }
+
+      A['c'];
+    `,
+    `
+      declare class A {
+        /** @deprecated */
         static accessor b: string;
         static accessor c: string;
       }
 
       A.c;
+    `,
+    `
+      declare class A {
+        /** @deprecated */
+        static accessor b: string;
+        static accessor c: string;
+      }
+
+      A['c'];
     `,
     `
       namespace A {
@@ -99,6 +183,15 @@ ruleTester.run('no-deprecated', rule, {
       A.c;
     `,
     `
+      namespace A {
+        /** @deprecated */
+        export const b = '';
+        export const c = '';
+      }
+
+      A['c'];
+    `,
+    `
       enum A {
         /** @deprecated */
         b = 'b',
@@ -106,6 +199,15 @@ ruleTester.run('no-deprecated', rule, {
       }
 
       A.c;
+    `,
+    `
+      enum A {
+        /** @deprecated */
+        b = 'b',
+        c = 'c',
+      }
+
+      A['c'];
     `,
     `
       function a(value: 'b' | undefined): void;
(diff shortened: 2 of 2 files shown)


JoshuaKGoldberg (changes requested):
🙌 Great start, thanks for kicking it off! I think there's just one case around `[key]` that I think is still missing?

undsoft:
@JoshuaKGoldberg 
Okay, I've added supported for simple identifiers and template strings.

```ts
const myObject = {
  /** @deprecated */
  recommended: null,
};
const key = 'recommended';
const k2 = 'recom';
const k3 = 'mended';

myObject[key]; // Works
myObject[`${k2}${k3}`]; // Works
```

However, more complex identifiers like:

```ts
const myObject = {
  /** @deprecated */
  recommended: null,
};
const key = {
  nested: 'recommended'
} as const;

myObject[key.nested]; // Doesn't work
```
are not supported, because `getPropertyName` that I use doesn't seem to understand this construction.

I feel like fixing this may be outside of the scope for this PR? Or otherwise I could use some guidance as to how to do this.

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -254,6 +288,10 @@ export default createRule<Options, MessageIds>({
         'Expected call like node to have signature',
       );
 
+      if (node.type === AST_NODE_TYPES.MemberExpression && node.computed) {
+        return getComputedPropertyDeprecation(node);
+      }
```
JoshuaKGoldberg, line 293: [Refactor] `getComputedPropertyDeprecation` is already being called in `getDeprecationReason`, just after the `getCallLikeDeprecation`. If we swap the order to try the computed property check first then we'll be able to remove this call here:

```suggestion
```

```ts

    function getDeprecationReason(node: IdentifierLike): string | undefined {
      if (isInComputedProperty(node)) {
        return getComputedPropertyDeprecation(
          node.parent as TSESTree.MemberExpression,
        );
      }

      const callLikeNode = getCallLikeNode(node);
      if (callLikeNode) {
        return getCallLikeDeprecation(callLikeNode);
      }
```

packages/eslint-plugin/tests/rules/no-deprecated.test.ts
```diff
@@ -691,6 +898,43 @@ exists('/foo');
...
+        const a = {
+          /** @deprecated */ b: 1,
+        };
+        const key = 'b';
+        a[key];
+      `,
+      errors: [
+        {
+          column: 11,
+          data: { name: 'key' },
```
JoshuaKGoldberg, line 912: [Bug] `key` isn't deprecated, `b` is. TypeScript correctly reports `'b' is deprecated.` so the rule should too. 

https://deploy-preview-11006--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.ts&code=MYewdgzgLgBAhjAvDA3gKBjA9AKhzAAQBMBTABwCcTg4oSiYcsMYAjALhgEYAaNAXwDcaNKEiwA1iQCeSGAHJW84WjgBtKdIC6goA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNFjpEZAIb5E3dFETRoHaJHBgAviEVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

(here and elsewhere)

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -232,7 +266,7 @@ export default createRule<Options, MessageIds>({
       }
     }
 
-    function getCallLikeNode(node: TSESTree.Node): CallLikeNode | undefined {
+    function getCallLikeNode(node: TSESTree.Node): CalleeNode | undefined {
```
JoshuaKGoldberg, line 269: [Naming] If the type's name is switched from `CalleeNode` to `CallLikeNode`, the corresponding `get*` function's name should too:

```suggestion
    function getCalleNode(node: TSESTree.Node): CalleeNode | undefined {
```

JoshuaKGoldberg (changes requested):
Really lovely progress! 💪

JoshuaKGoldberg:
> fixing this may be outside of the scope for this PR?

Agreed, yeah - I think it'd be good as a followup issue.

undsoft:
Okay, I believe I've addressed the comments.

JoshuaKGoldberg:
I'll handle the merge conflicts so we can ideally merge this in time for the release today.

JoshuaKGoldberg:
😬 these are non-trivial merge conflicts. I ran out of time today to try to get this to work.

@undsoft I don't suppose you have bandwidth to fix these up?

undsoft:
@JoshuaKGoldberg Hi, at the moment unfortunately I don't have spare time to fix this. I may have time in a couple of months.
It looks like there was a PR for similar functionality that partially implemented was this PR is trying to achieve. It will probably be necessary to re-do this PR.

packages/eslint-plugin/tests/rules/no-deprecated.test.ts
```diff
@@ -461,30 +620,6 @@ exists('/foo');
...
-        {
-          allow: [
-            {
-              from: 'file',
-              name: 'AllowedType',
-            },
-          ],
-        },
-      ],
-    },
```
JoshuaKGoldberg, line 487: 🤔.

JoshuaKGoldberg, line 487: 🤔. This is an existing test, from #10867, that no longer passes with the new unified logic. But I now think that it's not correct. The `AllowedType` interface is itself what's marked as allowed, not its `prop`. I think that that kind of `allow` property lookup is separately tracked by #10740.

packages/eslint-plugin/tests/rules/no-deprecated.test.ts
```diff
@@ -1128,82 +1348,306 @@ exists('/foo');
...
+        declare class A {
+          /** @deprecated */
+          b(): string;
+        }
+
+        declare const a: A;
+        const key = 'b';
+
+        a[key]();
+      `,
```
ronami, line 1490: Interestingly, all of the test cases here pass on the main branch except for this one. On the main branch this reports two errors (one false error) instead of one ([main branch playground link](https://typescript-eslint.io/play/#ts=5.9.2&fileType=.tsx&code=CYUwxgNghgTiAEkoGdnwILwN4Ch7wHoAqI%2BAAVAAc4woAXEYeIgveAIwAoBKALnmR0YASwB2AcwDcOAL44coJHEQB7UYPhR%2B6aWDUaA1iACe8ALzwA5O0vScBAvAAG7J-GFoqNeowB09xycjYzcPeC9wH2B-KABtYIBdHmkgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNFjpEZAIb5E3dFETRoHaJHBgAviEVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false), [deploy preview playground link](https://deploy-preview-11006--typescript-eslint.netlify.app/play/#ts=5.9.2&fileType=.tsx&code=CYUwxgNghgTiAEkoGdnwILwN4Ch7wHoAqI%2BAAVAAc4woAXEYeIgveAIwAoBKALnmR0YASwB2AcwDcOAL44coJHEQB7UYPhR%2B6aWDUaA1iACe8ALzwA5O0vScBAvAAG7J-GFoqNeowB0OKABtI2MAXR5pIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNFjpEZAIb5E3dFETRoHaJHBgAviEVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)).

It seems support for most edge cases was added in #10867 along with fixing the original issue reported in #10958.

What do you think about limiting the scope of this PR to specifically address this bug, or add additional tests that don't work well on the `main` branch?

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -189,6 +197,19 @@ export default createRule<Options, MessageIds>({
...
+    function isInComputedProperty(
+      node: IdentifierLike,
+    ): node is IdentifierInComputedProperty {
+      return (
+        node.parent.type === AST_NODE_TYPES.MemberExpression &&
+        node.parent.computed &&
+        node === node.parent.property &&
+        (node.type === AST_NODE_TYPES.Literal ||
+          node.type === AST_NODE_TYPES.TemplateLiteral ||
+          node.type === AST_NODE_TYPES.Identifier)
```
ronami: Bug: This can also be an expression other than literal, template literal or an identifier:

```ts
declare function x(): 'b';

const a = {
  /** @deprecated */
  b: 'string',
};

// should be reported but doesn't
const c = a[x()];
```

This seems like regression from the `main` branch which does correctly report this ([deploy preview playground link](https://deploy-preview-11006--typescript-eslint.netlify.app/play/#ts=5.9.2&fileType=.tsx&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXwA8AKASgC54ByAI0oG4AoBsPAZw3ingF54BvBvHgB6AFSj4AAVAAHOGCgYQweKOGD41CpXYwsqAOaUANAwC%2BjBsOHxWACxzIIK6gjgycMJS%2BQdgOEFZUSgxmNg4wHk4AbWISAF1GIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNFjpEZAIb5E3dFETRoHaJHBgAviEVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false), [`main` branch playground link](https://typescript-eslint.io/play/#ts=5.9.2&fileType=.tsx&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXwA8AKASgC54ByAI0oG4AoBsPAZw3ingF54BvBvHgB6AFSj4AAVAAHOGCgYQweKOGD41CpXYwsqAOaUANAwC%2BjBsOHwWMeRggBPeHBk4YS4MzYcwPTgBtYhIAXUYgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNFjpEZAIb5E3dFETRoHaJHBgAviEVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)).

---

Similarly ([deploy preview playground link](https://deploy-preview-11006--typescript-eslint.netlify.app/play/#ts=5.9.2&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3gDwFzwG94BPfAcgCMz4BfAbgCgHl0sp4BeQh%2BeAegBUA%2BAAFQABzhgoGEMHgC%2BPeBXKYYASxQBzMgBoG9Jnz7w0ACyQBXCPIoI44pDFl2rWYEhBoUZDM1RMRE54KABtbAA6YgBdRiA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNFjpEZAIb5E3dFETRoHaJHBgAviEVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false), [`main` branch playground link](https://typescript-eslint.io/play/#ts=5.9.2&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3gDwFzwG94BPfAcgCMz4BfAbgCgHl0sp4BeQh%2BeAegBUA%2BAAFQABzhgoGEMHgC%2BPeBXKYYASxQBzMgBoG9Jnz6IkMKRgjF4ccednBmqTIk7woAbWwA6YgF1GIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oBNFjpEZAIb5E3dFETRoHaJHBgAviEVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```ts
declare const x: { y: 'b' };

const a = {
  /** @deprecated */
  b: 'string',
};

// should be reported but doesn't
const c = a[x.y];
```

packages/eslint-plugin/tests/rules/no-deprecated.test.ts
```diff
@@ -804,6 +939,22 @@ exists('/foo');
         },
       ],
     },
+    {
+      code: `
+        /** @deprecated */ const a = { b: 1 };
+        console.log(a['b']);
```
ronami: Tests: Can this and other tests be simplified to:


```suggestion
        /** @deprecated */ const a = { b: 1 };
        a['b'];
```

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -210,13 +231,47 @@ export default createRule<Options, MessageIds>({
...
+
+      const type = services.getTypeAtLocation(node.property);
+
+      if (type.isStringLiteral()) {
+        return type.value;
+      }
+
+      if (type.isLiteral()) {
+        return String(type.value as number);
+      }
```
ronami: Nit: Making the following change doesn't seem to break any tests, is it safe or did I miss an edge case?


```suggestion
      if (type.isNumberLiteral()) {
        return String(type.value);
      }
```

ronami: Test: Seems like this can be commented out without failing tests, can we add a test or simplify?

packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -327,8 +382,41 @@ export default createRule<Options, MessageIds>({
...
+        return 'super';
+      }
+
+      if (node.type === AST_NODE_TYPES.PrivateIdentifier) {
+        return `#${node.name}`;
+      }
+
+      if (node.type === AST_NODE_TYPES.Literal) {
+        return String(node.value);
+      }
```
ronami: Test: Seems like this can be commented out without failing tests, can we add a test or simplify?

packages/eslint-plugin/src/rules/no-deprecated.ts
JoshuaKGoldberg: This took a bit of finagling 😎 

```ts
declare const Keys: {
  a: 1,
};

const a = {
  /** @deprecated reason for deprecation */
  [1]: 'string',
};

const key = Keys.a;
const c = a[key];
```

Added a test.

packages/eslint-plugin/tests/rules/no-deprecated.test.ts
JoshuaKGoldberg: I like it! Trying now..

JoshuaKGoldberg:
Ok! You're right, this become _much_ cleaner after a reset onto main. Nice.

ronami (approved):
Looks amazing, thanks! 🚀

Issue #10388, opened 2024-11-25
Docs: Blog post or other explainer on improving test coverage in RuleTester tests


JoshuaKGoldberg:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

I think one of, if not the most, common requests for changes I put in PR reviews is uncovered test lines in lint rules. Folks oftentimes don't test cases the AST is forcing them to write edge case code for. Or the reverse: they use type assertions to get around cases the AST types are trying to force them to handle.

I very often end up posting some comment like:

> Looks like there are some gaps in unit test coverage. For each gap in coverage, at least one of the following is probably true:
>
> * The code is reachable, so a unit test should exercise it
> * The code can be refactored to not include that case
> * This is a difficult-to-represent edge case in types and a type assertion is warranted

I'd love to have an explainer resource that explains what each of those mean, with an example. That'd save me a lot of typing - and would help promote understanding of how to work with the AST properly.

Proposal: let's make this into a blog post?

### Affected URL(s)

Maybe: `https://typescript-eslint.io/blog/tips-for-increasing-lint-rule-test-coverage`?

### Additional Info

I have a general blog post on this type of request drafted here: https://github.com/JoshuaKGoldberg/dot-com/pull/324. https://github.com/JoshuaKGoldberg/dot-com/blob/d7d2f348ecf38fb2e3012fb06c85db16f1267cdf/src/content/blog/so-youve-got-a-gap-in-code-coverage-for-a-lint-rule/index.mdx?plain=1 is a partially-done version of it specific to lint rules. I'd started writing that, then realized I wanted to be more general to JS/TS in my own blog.

PR #11032, opened 2025-04-05
feat(eslint-plugin): [no-explicit-any] suggest to replace keyof any with PropertyKey

### LINKED ISSUE #11009: Enhancement: [no-explicit-any] suggest to replace keyof any with PropertyKey

https://typescript-eslint.io/rules/no-explicit-any

### Description

Suggest to replace keyof any (and keyof never?) with PropertyKey as it is more explicit.
This can even be an auto fixable thing.

### Fail

```TypeScript
const integer = <TKey extends keyof any, TTarget extends { [K in TKey]: number }>(
  target: TTarget,
  key: TKey
) => { /* ... */ };
```

### Pass

```TypeScript
const integer = <TKey extends PropertyKey, TTarget extends { [K in TKey]: number }>(
  target: TTarget,
  key: TKey
) => { /* ... */ };
```

### Additional Info

Extracted from https://github.com/typescript-eslint/typescript-eslint/issues/3206

dbarabashh:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11009

## Overview

Enhanced the `no-explicit-any` rule to suggest replacing `keyof any` with PropertyKey. added detection, helpful suggestion messages, auto-fix support, and test coverage.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-explicit-any.ts
+++ b/packages/eslint-plugin/src/rules/no-explicit-any.ts
@@ -197,10 +197,10 @@ export default createRule<Options, MessageIds>({
      */
     function createPropertyKeyFixer(node: TSESTree.Node) {
       return (fixer: TSESLint.RuleFixer) => {
-        if (node.parent && node.parent.type === AST_NODE_TYPES.TSTypeOperator) {
-          return fixer.replaceText(node.parent, 'PropertyKey');
-        }
-        return fixer.replaceText(node, 'unknown');
+        return fixer.replaceText(
+          node.parent as TSESTree.TSTypeOperator,
+          'PropertyKey',
+        );
       };
     }
 
--- a/packages/eslint-plugin/src/rules/no-explicit-any.ts
+++ b/packages/eslint-plugin/src/rules/no-explicit-any.ts
@@ -195,24 +195,9 @@ export default createRule<Options, MessageIds>({
      * @returns a function that will fix the node.
      * @private
      */
-    function createPropertyKeyFixer(node: TSESTree.Node) {
+    function createPropertyKeyFixer(node: TSESTree.TSAnyKeyword) {
       return (fixer: TSESLint.RuleFixer) => {
-        return fixer.replaceText(
-          node.parent as TSESTree.TSTypeOperator,
-          'PropertyKey',
-        );
-      };
-    }
-
-    /**
-     * Creates a fixer that replaces any with unknown
-     * @param node the node to be fixed.
-     * @returns a function that will fix the node.
-     * @private
-     */
-    function createUnknownFixer(node: TSESTree.Node) {
-      return (fixer: TSESLint.RuleFixer): TSESLint.RuleFix => {
-        return fixer.replaceText(node, 'unknown');
+        return fixer.replaceText(node.parent, 'PropertyKey');
       };
     }
 
@@ -233,7 +218,7 @@ export default createRule<Options, MessageIds>({
         const unknownSuggestion: TSESLint.SuggestionReportDescriptor<MessageIds> =
           {
             messageId: 'suggestUnknown',
-            fix: createUnknownFixer(node),
+            fix: fixer => fixer.replaceText(node, 'unknown'),
           };
 
         const neverSuggestion: TSESLint.SuggestionReportDescriptor<MessageIds> =
@@ -257,7 +242,7 @@ export default createRule<Options, MessageIds>({
         if (fixToUnknown) {
           fixOrSuggest.fix = isKeyofAny
             ? createPropertyKeyFixer(node)
-            : createUnknownFixer(node);
+            : fixer => fixer.replaceText(node, 'unknown');
         }
 
         context.report({
--- a/packages/eslint-plugin/src/rules/no-explicit-any.ts
+++ b/packages/eslint-plugin/src/rules/no-explicit-any.ts
@@ -182,9 +182,9 @@ export default createRule<Options, MessageIds>({
      * @returns true if the node is within a keyof any expression, false otherwise
      * @private
      */
-    function isNodeWithinKeyofAny(node: TSESTree.Node): boolean {
+    function isNodeWithinKeyofAny(node: TSESTree.TSAnyKeyword): boolean {
       return (
-        node.parent?.type === AST_NODE_TYPES.TSTypeOperator &&
+        node.parent.type === AST_NODE_TYPES.TSTypeOperator &&
         node.parent.operator === 'keyof'
       );
     }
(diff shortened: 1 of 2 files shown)


dbarabashh:
Hey, 

I ran into a small issue. codecov is flagging line of code that is not covered by tests:
```ts
function createPropertyKeyFixer(node: TSESTree.Node) {
  return (fixer: TSESLint.RuleFixer) => {
    if (node.parent && node.parent.type === AST_NODE_TYPES.TSTypeOperator) {
      return fixer.replaceText(node.parent, 'PropertyKey');
    }
    return fixer.replaceText(node, 'unknown'); // ← This line isn't covered
  };
}
```

It is fallback that should never actually run I only call this function when we've already checked that the node has a parent that іs a keyof operator.

what is the best way to handle this?
- Just add ignore comment to make Codecov happy?
- Delete fallback?
- Try to somehow test this edge case (though that seems tricky)

Any thoughts on what is usually preferred for these kinds of situations? Thanks!

JoshuaKGoldberg:
Great question! We've been hoping to write a blog post on the subject: #10388.

Roughly, it's either of the following:

* The code truly can never be hit, so it should be removed
* The code can actually be hit, so test coverage should be added

If the type checker is pressuring you to write code that you don't think is necessary (e.g. the `node.parent` checks) then you can either:

* Use an `as` assertion
  * This tends to come up with `node.parent` checks: #6225, #10682, etc.
  * If you think there's a `node.parent` that should have a more specific type, please do file an issue, this is a pretty regular headache 🙃  
* Refactor the code to get around trade-offs in control flow analysis (e.g. storing a variable with a type-narrowed value)
  * If you ever want to spend a "fun" afternoon deep diving into this: https://github.com/microsoft/TypeScript/issues/9998

dbarabashh:
Thanks for the detailed reply @JoshuaKGoldberg 

I’ll go through all the topics you mentioned in your comment.

dbarabashh:
Thanks again for the comment you left it had a lot of great insights, and I especially enjoyed your article.

So, I pushed a commit with my solution. 

Of course, its not the best one out there, but following your article, I intentionally picked the worst and most unsafe approach 🙃. That said, the situation isn't as simple as it seems, so I will try to explain it in words maybe someone will find it helpful in the future.

[Using type assertion just to bypass type checking is generally a bad idea](https://github.com/JoshuaKGoldberg/dot-com/blob/d7d2f348ecf38fb2e3012fb06c85db16f1267cdf/src/content/blog/so-youve-got-a-gap-in-code-coverage-for-a-lint-rule/index.mdx?plain=1). Without solid test coverage, this approach can go wrong fast, and other node types might end up being handled incorrectly.

in this particular case, I came to the following conclusions:
- Since we have a check inside [isNodeWithinKeyofAny](https://github.com/typescript-eslint/typescript-eslint/pull/11032/files#diff-d4d6a7d9bc22336db45ece7f4f6e775621e71e5f1ec94bf4a45f4d8a5d43b5d1R185) that guarantees `node.parent` exists and is of type `TSTypeOperator` with `operator === 'keyof'`,
- and [createPropertyKeyFixer](https://github.com/typescript-eslint/typescript-eslint/pull/11032/files#diff-d4d6a7d9bc22336db45ece7f4f6e775621e71e5f1ec94bf4a45f4d8a5d43b5d1R198) is only called after that check,
- the previous version of createPropertyKeyFixer had an extra check for something that was already verified.

according to @JoshuaKGoldberg  article, in our case, type assertion actually *is* the best strategy, because:
- the check is indeed redundant (yikes),
- and we know for sure that the type is correct (so we are not really bypassing the type system).

packages/eslint-plugin/src/rules/no-explicit-any.ts
```diff
@@ -170,36 +176,88 @@ export default createRule<Options, MessageIds>({
...
+      );
+    }
+
+    /**
+     * Creates a fixer that replaces a keyof any with PropertyKey
+     * @param node the node to be fixed.
+     * @returns a function that will fix the node.
+     * @private
+     */
+    function createPropertyKeyFixer(node: TSESTree.Node) {
```
JoshuaKGoldberg, line 198: I should have mentioned this more explicitly, sorry: by _"refactor the code"_ I meant refactoring either the runtime logic _or_ the types. In this case `createPropertyKeyFixer` is only ever called with a node that we know to have a `.parent`. So we can do this:

```suggestion
    function createPropertyKeyFixer(node: MakeRequired<TSESTree.Node, 'parent'>) {
```

Or, even better, we know `createPropertyKeyFixer` is only ever called with a `TSESTree.TSAnyKeyword`. So we can do this instead:

```suggestion
    function createPropertyKeyFixer(node: TSESTree.AnyKeyword) {
```

...and either way, the `as TSESTree.TSTypeOperator` is unnecessary.

Change request: switch to `: TSESTree.AnyKeyword` and remove the `as`.

packages/eslint-plugin/src/rules/no-explicit-any.ts
```diff
@@ -170,36 +176,88 @@ export default createRule<Options, MessageIds>({
...
+            fix: createUnknownFixer(node),
+          };
+
+        const neverSuggestion: TSESLint.SuggestionReportDescriptor<MessageIds> =
+          {
+            messageId: 'suggestNever',
+            fix(fixer: TSESLint.RuleFixer): TSESLint.RuleFix {
+              return fixer.replaceText(node, 'never');
+            },
+          };
```
JoshuaKGoldberg, line 245: [Refactor] Could you move these back inline, please? These objects don't need to be created until & unless they're needed. Plus this is a lot of explicit typing that can be inferred.

packages/eslint-plugin/tests/rules/no-explicit-any.test.ts
```diff
@@ -1199,8 +1199,68 @@ const test = <T extends Partial<never>>() => {};
...
+                output:
+                  'const integer = <TKey extends PropertyKey, TTarget extends { [K in TKey]: number }>(target: TTarget, key: TKey) => { /* ... */ };',
+              },
+            ],
+          },
+        ],
+      },
     ] as RuleInvalidTestCase[]
   ).flatMap(testCase => {
+    if (testCase.code.includes('keyof any')) {
```
JoshuaKGoldberg, line 1237: [Testing] 😬 I know this is existing code with its own pattern, but - we're trying to move away from auto-generated tests. They're much harder to statically analyze and tend to batch together a lot of redundant tests with multiple edge cases per test. Could you please keep to manually written tests for the new code?

JoshuaKGoldberg (changes requested):
Thanks for keeping working on this - I think you're in the right track with the removing a runtime check. 👍

packages/eslint-plugin/tests/rules/no-explicit-any.test.ts
JoshuaKGoldberg: Explicit suggestions for each is the way to go. If you search for `suggest`/`suggestions` in existing files you should be able to find examples.

packages/eslint-plugin/src/rules/no-explicit-any.ts
JoshuaKGoldberg: This request is still open, did you mean to push some local code? 🙂

JoshuaKGoldberg (approved):
💯 thanks!

PR #11075, opened 2025-04-15
feat(eslint-plugin): [only-throw-error] add option `allowRethrowing`

### LINKED ISSUE #10376: Enhancement: [only-throw-error] allow re-throwing a caught error

https://typescript-eslint.io/rules/only-throw-error

### Description

`only-throw-error` (and `prefer-promise-reject-errors`) disallow throwing non-error values. `only-throw-error` (and with #10375 `prefer-promise-reject-errors` too) can be configured to allow throwing `unknown` and `any` types in addition to error values.

It's necessary to enable `allowThrowingUnknown` so re-throwing a cached error is valid (or `any` in some `tsconfig.json` configurations):

```ts
function test() {
  try {
    // ...
  } catch (error) {
    if (x) {
      throw new Error()
    }

    // this would otherwise fail
    throw error;
  }
}
```

While configuring `allowThrowingUnknown` makes the above pass, it allows potentially unsafe throwing of an `unknown` type:

```ts
function foo(): unknown {
  // not necessarily returns an `Error` object
}

function test() {
  // this should fail
  throw foo();
}
```

This issue proposes an additional option, a stricter form of `allowThrowingUnknown,` which will allow re-throwing a caught error.

_This is relevant both to [`only-throw-error`](https://typescript-eslint.io/rules/only-throw-error) and [`prefer-promise-reject-errors`](https://typescript-eslint.io/rules/prefer-promise-reject-errors)._

### Fail

```TypeScript
function foo(): unknown {
  // not necessarily returns an `Error` object
}

function test() {
  // this should fail
  throw foo();
}
```

### Pass

```TypeScript
function test() {
  try {
    // ...
  } catch (error) {
    if (x) {
      throw new Error()
    }

    throw error;
  }
}
```

### Additional Info

_No response_

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10376


packages/eslint-plugin/package.json
```diff
@@ -56,7 +56,7 @@
     "generate:breaking-changes": "tsx tools/generate-breaking-changes.mts",
     "generate:configs": "npx nx generate-configs repo",
     "lint": "npx nx lint",
-    "test": "vitest --run --config=$INIT_CWD/vitest.config.mts",
```
kirkwaiblinger, line 59: revert

packages/eslint-plugin/docs/rules/only-throw-error.mdx
```diff
@@ -123,6 +123,13 @@ interface Options {
     | string
   )[];
 
+  /**
+   * Whether to allow rethrowing caught values that are not `Error` objects.
+   *
+   * @default false
+   */
+  allowRethrowing?: boolean;
```
JoshuaKGoldberg, line 131: [Docs] The options mentions in this page don't explain why any of the options exist. I suppose that's existing/separate docs work for the other ones. But do you think it'd be reasonable to at least mention `allowRethrowing` in this PR?

If you'd rather them all tackled together as a followup, I wouldn't block.

packages/eslint-plugin/src/rules/only-throw-error.ts
```diff
@@ -65,13 +78,74 @@ export default createRule<Options, MessageIds>({
...
+        def.node.type === AST_NODE_TYPES.ArrowFunctionExpression &&
+        def.node.params.length >= 1 &&
+        def.node.params[0] === def.name &&
+        def.node.parent.type === AST_NODE_TYPES.CallExpression
+      ) {
+        const callExpression = def.node.parent;
+
+        const parsedPromiseHandlingCall =
+          parseCatchCall(callExpression, context) ??
+          parseThenCall(callExpression, context);
```
JoshuaKGoldberg, line 127: [Refactor] Both this rule and `no-floating-promises` run `parseCatchCall(node, context) ?? parseThenCall(node, context)`. Nowhere else uses either. Is there a reason to expose both functions?

```suggestion
          parsePromiseHandlingCall(callExpression, context);
```

packages/eslint-plugin/src/rules/only-throw-error.ts
```diff
@@ -65,13 +78,74 @@ export default createRule<Options, MessageIds>({
...
+        if (parsedPromiseHandlingCall != null) {
+          const { object, onRejected } = parsedPromiseHandlingCall;
+          if (onRejected === def.node) {
+            const tsObjectNode = services.esTreeNodeToTSNodeMap.get(object);
+
+            // make sure we're actually dealing with a promise
+            if (
+              isThenableType(
+                services.program.getTypeChecker(),
+                tsObjectNode satisfies TSESTreeToTSNode<TSESTree.Expression> as ts.Expression,
```
JoshuaKGoldberg, line 137: [Refactor] I think I see why this is here: it's to make sure that if the types change so that `object` -> `tsObjectNode` aren't both expressions, we'll be notified? If so: I think just having `object` as type `TSESTree.Expression` is good enough. We've never had to use `TSESTreeToTSNode` this way before, and it feels odd to me.

JoshuaKGoldberg (approved):
Really nifty stuff! I like the extraction+reuse for Promise detection. Left a few comments, nothing close to blocking IMO. 🔥

packages/eslint-plugin/src/rules/only-throw-error.ts
kirkwaiblinger: IMHO - yes. It's "coincidental" in a sense that these functions are both used essentially as `(parseThenCall() ?? parseCatchCall())?.onRejected` in the two rules they're used in. But you could imagine even if we just wanted to differentiate the error message based on whether it's a `.then()` or a `.catch()` we'd need these exposed separately, let alone if we need separate logic between the methods in a rule at some point. Slightly speculative I suppose but not really a stretch I don't think.

packages/eslint-plugin/src/rules/only-throw-error.ts
kirkwaiblinger: Yeah, I think we're probably saying the same thing but my thought process was 
1. `isThenableType()` requires a `ts.Expression`, but unfortunately `services.esTreeNodeToTSNodeMap.get(object)` has type `TSESTreeToTSNode<TSESTree.Expression>` which is _not_ `ts.Expression`. So a type assertion is necessary
2. Not understanding in great detail why `TSESTreeToTSNode<TSESTree.Expression> != ts.Expression`, I thought to do the type assertion in the most conservative way, i.e. validate type first (via `satisfies`) then assert, rather than unconditionally assert. 

I do think it at least makes sense to move around the assertion to the declaration of `tsObjectNode`, so I'll move that... I'm not 100% sure whether you're suggesting further changes or not though so lmk if you approve of my change.

packages/eslint-plugin/src/rules/only-throw-error.ts
JoshuaKGoldberg: Makes sense to me, thanks. Resolving as wontfix 🔪

packages/eslint-plugin/src/rules/only-throw-error.ts
JoshuaKGoldberg: Yeah the mapping thing is confusing and I don't understand at all why they're `!=` 🙃. Agreed with the tradeoffs. 🚢

.vscode/extensions.json
```diff
@@ -6,7 +6,8 @@
       "esbenp.prettier-vscode",
       "streetsidesoftware.code-spell-checker",
       "tlent.jest-snapshot-language-support",
-      "mrmlnc.vscode-json5"
+      "mrmlnc.vscode-json5",
+      "unifiedjs.vscode-mdx"
```
JoshuaKGoldberg, line 10: Unrelated?

PR #11076, opened 2025-04-15
chore: update vscode launch tasks to use vitest

### LINKED ISSUE #11074: Repo: Make F5 Launch config work with vitest
### Suggestion

We suggest launch tasks in order to run the tests for a file: https://typescript-eslint.io/contributing/local-development#tests

But this assumes we're using jest, and is broken with the switch to vitest. Let's make sure we can run/debug tests in the editor with vitest (most important is eslint-plugin).

cc @43081j @aryaemami59 

related: #11068 

### Additional Info

_No response_

43081j:
Updates the launch task to use vitest for running tests.

Note that I didn't (yet?) enable single thread mode in vitest as it isn't quite the equivalent to jest's `runInBand`. From what I understand, vitest will always run tests in another thread even in single thread mode (as in there will be one process running vitest, and one running tests).

I also noticed there was a duplicate path in the existing skip lists, so have updated that.

cc @kirkwaiblinger  @aryaemami59 

## PR Checklist

- [x] Addresses an existing open issue: fixes #11074 


kirkwaiblinger:
Tried this out locally - for the most part, it works great! One thing that's kind of clunky though is that debugging a test case can cause it to fail due to test timeouts. Is there a way to fix that?

43081j:
maybe we just need to disable the timeout in the task?

as in `--testTimeout=0`

kirkwaiblinger:
Nice, yeah the `--testTimeout=0` solved that error 👍 There's still some issue with `Error: [vitest-worker]: Timeout calling "onTaskUpdate"` but it doesn't really block anything 🤷 

Also I tried all the options at https://vitest.dev/guide/debugging.html#vs-code and it didn't seem to solve it. Oh well!

Issue #10334, opened 2024-11-15
Bug: [no-unnecessary-type-assertion] false positive on evolving type


OpportunityLiu:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.6.3&fileType=.tsx&code=LAKAJgpgxgNghgJwgAgGYFcB2UAuBLAe0zQIIAoBKALmU3QFsAjCBZAH2S0lT0wjFCRYiFBmz4iyRoko06TFoOjwkaLLkLFpALzKY5DZgmrIAbgTxgA3KFBiNk%2BnF4BGSsgDeoZMhgQcyAAeNiA%2BgcgAvFIyFCFhkSTksd7RuoEAhMkgAL62IPYSxE68AEzuXqG%2B-kFxQQnSCJS14VGopE0pOmSBWbkgQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDArugDTg10NM8AOXapUAYQAW6aAGsylDul4BfECqA&tokens=false

### Repro Code

```TypeScript
declare function foo(): number | undefined
declare function bar(): number
declare function baz(n: number): void;

function main1() {
  let x;
  x = bar();
  x = foo();
  baz(x!);
}

function main2() {
  let x;
  x = bar();
  x = foo();
  baz(x);
}
```

### ESLint Config

```javascript
module.exports = {
  parser: "@typescript-eslint/parser",
  rules: {
    "@typescript-eslint/no-unnecessary-type-assertion": "error"
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true
  }
}
```

### Expected Result

[...]

PR #11082, opened 2025-04-16
fix(eslint-plugin): [no-unnecessary-type-assertion] false positive on non-null assertion after an implicitly-any variable gets initialised inside conditional block

### LINKED ISSUE #11054: Bug: @typescript-eslint/no-unnecessary-type-assertion False positive on non-null assertion after an implicitly-any variable gets initialised inside a conditional or loop

### Playground Link

https://typescript-eslint.io/play/#ts=5.8.2&fileType=.ts&code=GYVwdgxgLglg9mABAUwB4EMC2AHANsgCgEpEBvAKEUXykVQG5LEZhECBZdKACwDoAndGAAmcTMUQA%2BRAFoAjCQpUqqRAF5EAJkZUAvk37IoIfklQBCRAGpEcxvvJoseQkUZA&eslintrc=N4KABGBECmAeAu0B2ATAzpAXGA2uCUADgDYCuA5gJZKYAC8AnodGgMYBOlh8AtC8dXgB6dtFYB7ALaTkKaCh6NmPVgAsxAa3mR8AXQA0%2BSO1LEWWMMAC%2BIK0A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false

### Repro Code

```TypeScript
function example() {
  let x;
  if (Math.random() > -1) {
    x = 2;
  }
  return x! + 1;
}

example();
```

### ESLint Config

```javascript

  tseslint.configs.recommendedTypeChecked,
  {
    languageOptions: {
      parserOptions: {
        projectService: true,
        tsconfigRootDir: import.meta.dirname,
      },
    },
  },
  {
    files: ['eslint.config.mjs'],
    extends: [tseslint.configs.disableTypeChecked],
  }
);
```

### tsconfig

```jsonc
{
    "include": ["**/*.ts"],
    "compilerOptions": {
        "strict": true,
        
    }
}
```

### Expected Result

No errors, since the non-null assertion (`!`) *is* necessary here; without it, tsc will refuse to compile the code, complaining that `'x' is possibly 'undefined'!

### Actual Result

```
6:10  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
```

### Additional Info

Things that seem necessary to produce the error above:

* `x` is not explicitly typed when declared (so is implicitly `any`)
* the initial assignment of `x` takes place inside a conditional or a loop, and TypeScript therefore can't tell, without the non-null assertion, that the initial assignment is guaranteed to take place
* `strict: true` is enabled

The example code above is of course contrived, but a more realistic example where this would in fact be totally reasonable code would be one where, instead of an `if (...) {...}` that's guaranteed to run, the assignment takes place within a *loop* that's guaranteed to run (and `x` gets used to store some sort of information about the final iteration of the loop that gets used after the loop finishes).

mdm317:
> [!CAUTION]
> Currently blocked by https://github.com/typescript-eslint/typescript-eslint/issues/10334 / https://github.com/microsoft/TypeScript/issues/60514. See https://github.com/typescript-eslint/typescript-eslint/pull/11082#discussion_r2047412747

***

## PR Checklist

- [x] Addresses an existing open issue: fixes #11054 

## Overview
I have updated it so that [isPossiblyUsedBeforeAssigned](https://github.com/typescript-eslint/typescript-eslint/blob/23c5aa725ede26fbb88ac77cb950181ac2fca4d7/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts#L68-L149) returns `true` when the variable is not explicitly typed at the time of declaration

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
@@ -118,9 +118,11 @@ export default createRule<Options, MessageIds>({
         if (
           // is it `const x!: number`
           declaration.initializer == null &&
-          declaration.exclamationToken == null &&
-          declaration.type != null
+          declaration.exclamationToken == null
         ) {
+          if (declaration.type == null) {
+            return true;
+          }
           // check if the defined variable type has changed since assignment
           const declarationType = checker.getTypeFromTypeNode(declaration.type);
           const type = getConstrainedTypeAtLocation(services, node);
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
@@ -429,6 +429,13 @@ declare function foo<T extends unknown>(bar: T): T;
 const baz: unknown = {};
 foo(baz!);
     `,
+    `
+let x;
+if (Math.random() > -1) {
+  x = 2;
+}
+x!;
+    `,
   ],
 
   invalid: [


packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -118,9 +118,11 @@ export default createRule<Options, MessageIds>({
         if (
           // is it `const x!: number`
           declaration.initializer == null &&
-          declaration.exclamationToken == null &&
-          declaration.type != null
+          declaration.exclamationToken == null
         ) {
+          if (declaration.type == null) {
+            return true;
```
kirkwaiblinger, line 124: We're definitely in the right ballpark, but I notice that this introduces a regression with the following case:
```ts
function foo() {
  let x
  if (Math.random() > 0.5) {
    x = 3;
  } else {
    x = 4;
  }
  // should be flagged as unnecessary
  x!;
}
```
I wonder if it's possible to get that case correct still? (it may or may not be feasible to prevent some false positives or negatives)

kirkwaiblinger: actually - I think you can ignore this regression... This is most likely a symptom of https://github.com/typescript-eslint/typescript-eslint/issues/10334 / https://github.com/microsoft/TypeScript/issues/60514

mdm317: Thanks for review!  🙏

The regression also occurs in this simplified case:
```
function foo() {
  let x;
  x =4;
  // should be flagged as unnecessary
  x!;
}
```

> This is most likely a symptom of https://github.com/typescript-eslint/typescript-eslint/issues/10334 / https://github.com/microsoft/TypeScript/issues/60514

As you mentioned, it's not possible to fix that case for now.

Should I close this PR until that issue is resolved?

kirkwaiblinger: Hmmm. I'm not sure which choice is the lesser of two evils here... fixing the bug at the cost of that rather undesirable-looking regression, or marking the bug as blocked. I'm gently leaning towards making this work as blocked unless we can find away to proceed while mitigating this regression 🤔 But then again if it was that important of a case, you'd think we might have a test case that caught it? 

Looking for thoughts from @typescript-eslint/triage-team too.

JoshuaKGoldberg: Yeah I don't really want to wade into these murky waters 🙃. +1 on marking as blocked.

PR #11115, opened 2025-04-29
fix(typescript-estree): correct `TSImportType` property name when `assert`

### LINKED ISSUE #11114: Bug: `TSImportType`'s `option` object always has key `with`

### Relevant Package

typescript-estree

### Playground Link

https://typescript-eslint.io/play/#ts=5.8.2&showAST=es&fileType=.tsx&code=C4TwDgpgBAKlC8UCWBbMB7ATsAFAIhTwBooBvKAdyWAAsAuMqPTCAZ3QBsBXYJdAOwC0KdABMIeBnlQZseKAF9FASgDcAKFCQoAVQTI0WXAWKMAhq1YRsDcszacefISPGSmMo-KUK16oA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
type T = import("m", { with: { "resolution-mode": "import" } });
type U = import("m", { assert: { "resolution-mode": "import" } });
```

### ESLint Config

```javascript

```

### tsconfig

```jsonc

```

### Expected Result

Both statements include a `TSImportType` with an `options` property containing an `ObjectExpression`. Both `ObjectExpression`s have a single `Property`.

The key of `Property` in 1st statement is `with`.

The key of `Property` in 2nd statement is also `with`! I'd have expected it to be `assert`.

### Actual Result

(see above)

### Additional Info

I *think* this is a bug. But apologies if this is a stupid query. I don't know TypeScript at all well.

I notice that TypeScript's AST doesn't seem to differentiate between the two either, so maybe it's not possible for TS-ESLint to?

### Versions

| `@typescript-eslint/parser`            | `8.31.1` |

overlookmotel:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11114

## Overview

I *guess* `token.getText()` is the best API to use to obtain the property name, but I'm unfamiliar with both `typescript-eslint` and TypeScript compiler, so I may well be on the wrong track. Ditto I *think* I've put the test in the right place.

If any changes required (or if I've gone about this in completely the wrong way), please do let me know.


overlookmotel:
Could someone please advise on how to fix the `codecov/patch` CI fail? I'm a bit baffled as the changes made to tests does exercise multiple possible values of `withOrAssertName`. Is it just due to using longer var names?

packages/ast-spec/src/type/TSImportType/fixtures/type-import-type-with-import-attributes/fixture.ts
```diff
@@ -1,3 +1,4 @@
 // TODO: This fixture might be too large, and if so should be split up.
 
 type A = import("A", { with: { type: "json" } });
+type B = import("B", { assert: { type: "json" } });
```
JoshuaKGoldberg, line 4: > the test in the right place.

You're close 🙂. These ast-spec fixtures are the right place in general, but we generally try to keep the fixtures small and single-purpose. I think the right step here would be to:

* Copy this `type-import-type-with-import-attributes` folder to a new one
* Have one of these fixtures test `assert` and the other test `with`
* Remove the `// TODO: ` comments from each - those were automatically added and I think they're not accurate in this case

JoshuaKGoldberg:
> Could someone please advise on how to fix the `codecov/patch` CI fail? I'm a bit baffled as the changes made to tests does exercise multiple possible values of `withOrAssertName`. Is it just due to using longer var names?

#6701 🫠. This PR is fine, it's the test coverage reporting that's wrong. You can ignore codecov. That's what I'd do (along with an exasperated sigh). 

If you want to be awesome and go the extra mile, you could try to fix #6701. 😄

packages/typescript-estree/src/convert.ts
```diff
@@ -3118,22 +3118,27 @@ export class Converter {
...
           )!;
-          const withToken = findNextToken(openBraceToken, node, this.ast)!;
-          const withTokenRange = getRange(withToken, this.ast);
+          const withOrAssertToken = findNextToken(
+            openBraceToken,
+            node,
+            this.ast,
+          )!;
+          const withOrAssertTokenRange = getRange(withOrAssertToken, this.ast);
+          const withOrAssertName = withOrAssertToken.getText();
```
JoshuaKGoldberg, line 3127: > `token.getText()`

That's probably fine, but it would be a little more precise to check the `.kind`:

```suggestion
          const withOrAssertName =
            withOrAssertToken.kind === ts.SyntaxKind.AssertKeyword
              ? 'assert'
              : 'with';
```

IMO this is slightly preferable. But if another team member wants to come in and tell me `.getText()` is good enough and more succinct I won't complain.

JoshuaKGoldberg (changes requested):
Great first PR! 👏 

I'm just requesting changes on making a dedicated fixture for this. The text parsing stuff I'm not confident on.

Jack Torrence in The Shining at the bar, happily waggling his eyebrows and holding a drink

packages/ast-spec/src/type/TSImportType/fixtures/type-import-type-with-import-attributes/fixture.ts
overlookmotel: Addressed in 725831a58638ec62d0316db26d2a80e7e883639c.

packages/typescript-estree/src/convert.ts
overlookmotel: Addressed in 7af8e537c88e4166ca5b7bb6b8bc161fe1ed52db.

Since no-one else has (yet) expressed an opinion, I've gone with yours.

overlookmotel:
> If you want to be awesome and go the extra mile, you could try to fix #6701. 😄

I aspire to being awesome, but I'm sorry to say I wouldn't have a clue how to do that!

bradzacher (approved):
LGTM - great work!

JoshuaKGoldberg (dismissed):
🙌

PR #11116, opened 2025-04-29
fix(eslint-plugin): decode unicode escapes in identifier tokens

### LINKED ISSUE #11036: Bug: The token value of an escaped identifier differs with espree

### Relevant Package

typescript-estree

### Playground Link
https://typescript-eslint.io/play#ts=5.6.2&showAST=es&fileType=.tsx&code=DoVwDGBsCMQ&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=true

### Repro Code

```TypeScript
\u0061
```

### ESLint Config

```javascript

```

### tsconfig

```jsonc

```

### Expected Result

`.ast.tokens[0].value` should be `a` (Espree)

### Actual Result

`.ast.tokens[0].value` is `\\u0061`

### Additional Info

Whether a token value should be the escaped text / the raw text is certainly up to debate. If the goal is to align with espree, then I think we should use the escaped text.

### Versions

| `@typescript-eslint/typescript-estree` | `8.29.1` |

dbarabashh:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11036 

## Overview
Screenshot 2025-04-29 at 17 20 16
Screenshot 2025-04-29 at 17 21 03

### DIFF
--- a/packages/typescript-estree/tests/lib/node-utils.test.ts
+++ b/packages/typescript-estree/tests/lib/node-utils.test.ts
@@ -1,4 +1,7 @@
-import { unescapeStringLiteralText } from '../../src/node-utils';
+import {
+  unescapeStringLiteralText,
+  unescapeUnicodeIdentifier,
+} from '../../src/node-utils';
 
 describe(unescapeStringLiteralText, () => {
   it('should not modify content', () => {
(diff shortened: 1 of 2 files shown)


dbarabashh:
I’m not sure the current CI/CD error is actually caused by my changes but I cant say for certain. I think I need some help.
@JoshuaKGoldberg @bradzacher

packages/typescript-estree/tests/lib/node-utils.test.ts
```diff
@@ -42,3 +45,30 @@ describe(unescapeStringLiteralText, () => {
...
+  it('should leave non-escape text unchanged', () => {
+    expect(unescapeUnicodeIdentifier('foo')).toBe('foo');
+    expect(unescapeUnicodeIdentifier('a\\u0062c')).toBe('abc');
+  });
+
+  it('should not decode invalid escapes', () => {
+    expect(unescapeUnicodeIdentifier('\\u00ZZ')).toBe('\\u00ZZ');
+    expect(unescapeUnicodeIdentifier('\\u{ZZ}')).toBe('\\u{ZZ}');
+  });
+});
```
JoshuaKGoldberg, line 74: [Testing] Nit: I'm a fan of having one thing under test per `it()` case. I think it would make sense to split each of these two-paired ones into two `it()`s. But I wouldn't block on it, this file already doesn't align to that (and even more so in existing code!).

JoshuaKGoldberg (dismissed):
LGTM, thanks! 🚀 

I would definitely prefer someone like @bradzacher who's more familiar with text parsing review. But from my less-than-expert perspective this all looks great.

bradzacher (commented):
my only concern here is that we're going to be doing a regex scan over every single `Identifier` in every single file we parse just so that we can potentially catch one specific edge-case that (a) likely isn't present in 99.9999% of code and (b) likely doesn't impact anyone's lint rules...

acorn gets away with producing a token with the non-encoded character because it is a true parser -- it derives the AST from the tokens it produces.

Truth be told -- I'm honestly not completely sold on us landing this to align with acorn/espree. WDYT @typescript-eslint/triage-team?

dbarabashh:
@bradzacher what if add a check before running the regex? Something like:

```ts
if (!text.includes('\\u')) {
  return text;
}
```

this way we keep the compatibility but only run the expensive regex when there's actually a unicode escape. Should solve the performance concern since 99.9% of identifiers will take the fast path. any thoughts?

bradzacher:
See the update comment I just posted on the related issue: https://github.com/typescript-eslint/typescript-eslint/issues/11036#issuecomment-3016264500

Thank you so much for raising a PR for this -- but we have decided that the impacts of this change are too great for a rare (to never ever used) edge-case.

PR #11199, opened 2025-05-09
docs(eslint-plugin): [no-explicit-any] add alternatives

### LINKED ISSUE #11063: Docs: Add explainer around `any` alternatives with no-explicit-any

### Suggested Changes

We sometimes get requests for `no-explicit-any` like:

* #642
* #7751
* https://x.com/JoshuaKGoldberg/status/1910814958515871943

...where folks have situations they believe `any` to be a correct and safe choice to use. Almost all -or maybe all?- of those situations end up having safety holes and/or being times when `never` or `unknown` would be more appropriate.

But, https://typescript-eslint.io/rules/no-explicit-any doesn't have a mention of `never`, and only light documentation on strategies for the points in those requests.

Suggestion: how about we bolster the `no-explicit-any` docs with some FAQs?

### Affected URL(s)

https://typescript-eslint.io/rules/no-explicit-any

### Additional Info

#10362 -> https://typescript-eslint.io/blog/avoiding-anys#banning-explicit-anys added  some docs, but isn't super comprehensive. It's more of a high-level overview with select situations and strategies.

💖

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11063


packages/eslint-plugin/docs/rules/no-explicit-any.mdx
```diff
@@ -145,6 +145,36 @@ interface Garply {
 }
 ```
 
+## Alternatives to `any`
+
+Which type(s) to use instead of `any` depends on the context.
+
```
JoshuaKGoldberg, line 151: I can't find references of other non-straightforward scenarios to add... are there others we should mention?

JoshuaKGoldberg, line 151: TODO: Add more straightforward cases from earlier in the file...

- If the type is known, describing it in an `interface` or `type`
  - Maybe mention `[i: string]`/etc. index signatures?
- If the type is not known, using the safer `unknown` type

kirkwaiblinger (approved):

PR #11200, opened 2025-05-10
docs: add note about `max-semi-size` to performance section

maschwenk:
## PR Checklist

## Overview

This setting has improved the performance of typed linting of our biggest app (~2.5 million LoC) by a factor of 2x.

### DIFF
--- a/docs/troubleshooting/typed-linting/Performance.mdx
+++ b/docs/troubleshooting/typed-linting/Performance.mdx
@@ -38,6 +38,14 @@ In particular for typed linting:
   - [Performance Tracing](https://github.com/microsoft/TypeScript/wiki/Performance#performance-tracing) can spotlight specific slow types within your project.
 - [Using Project References](https://github.com/microsoft/TypeScript/wiki/Performance#using-project-references) -which requires enabling the [new "project service" (`parserOptions.projectService`) in v8](/blog/announcing-typescript-eslint-v8-beta#project-service)- can be helpful to speed up type checking on larger projects.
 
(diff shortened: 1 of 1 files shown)


maschwenk:
My reproduction case here really requires you have a disgustingly large app to repro, so unsure if there are ideas

docs/troubleshooting/typed-linting/Performance.mdx
```diff
@@ -38,6 +38,14 @@ In particular for typed linting:
   - [Performance Tracing](https://github.com/microsoft/TypeScript/wiki/Performance#performance-tracing) can spotlight specific slow types within your project.
 - [Using Project References](https://github.com/microsoft/TypeScript/wiki/Performance#using-project-references) -which requires enabling the [new "project service" (`parserOptions.projectService`) in v8](/blog/announcing-typescript-eslint-v8-beta#project-service)- can be helpful to speed up type checking on larger projects.
 
+If none of the above work, you can try adjusting the `--max-semi-space-size` of Node. Increasing the max size of a semi-space can improve performance at the cost of more memory consumption. You can read more about [the setting here](https://nodejs.org/api/cli.html#--max-semi-space-sizesize-in-mib).
```
JoshuaKGoldberg, line 41: [Accessibility] https://www.mtu.edu/accessibility/training/web/link-text - link text should make sense out of context.

```suggestion
If none of the above work, you can try adjusting the `--max-semi-space-size` of Node. Increasing the max size of a semi-space can improve performance at the cost of more memory consumption. You can [read more about setting space size in Node.js here](https://nodejs.org/api/cli.html#--max-semi-space-sizesize-in-mib).
```

JoshuaKGoldberg (approved):
Seems reasonable to me, but I'd like to defer to @bradzacher as the tsc-at-scale 👑

Issue #11206, opened 2025-05-12
Repo: Speed up semantic-breaking-change-pr.yml


JoshuaKGoldberg:
### Suggestion

Splitting out of #11204: our [`semantic-breaking-change-pr.yml`](https://github.com/typescript-eslint/typescript-eslint/blob/f8db925cdbaf2d85575de852ac9347d56e718604/.github/workflows/semantic-breaking-change-pr.yml) workflow is good and useful for making sure PRs that introduce breaking changes have the appropriate title and description.

But, per https://github.com/typescript-eslint/typescript-eslint/actions/metrics/usage, that action currently is our third biggest slice of total CI minutes. I'm seeing 613 minutes across 359 runs over the last month.

Much of that time comes from the action having to go through our full installation command:

https://github.com/typescript-eslint/typescript-eslint/blob/f8db925cdbaf2d85575de852ac9347d56e718604/.github/workflows/semantic-breaking-change-pr.yml#L18

Let's speed that up! I can think of a couple options to try:

* Quick & easy: Instead of our shared installation action, we could install just the `@github/*` deps the JS file actually uses
* More comprehensive: @bradzacher suggested internally that we could switch it to a standalone repo's action. Doing so would allow us to pre-compile the action as JS and make it reusable

Proposal: let's go with quick & easy for now?

### Additional Info

Also related is #11205, for speeding up CI install times. But I don't think this workflow should be running a full repository install at all.

💖

kirkwaiblinger:
This has decreased from [~1m](https://github.com/typescript-eslint/typescript-eslint/actions/runs/15324359336) to [~30s](https://github.com/typescript-eslint/typescript-eslint/actions/runs/15335524196) since the merging of #11219. Yay! 

Leaving this open though since there's still room for further improvement as described in the issue.

JoshuaKGoldberg:
#11286 brings it down to just an embedded JS script. I'm satisfied with this for now. Eventually OctoGuide (#11203) might be configurable to do more, in which case we could reuse that action.

Issue #11205, opened 2025-05-12
Repo: Per-job CI Install step takes a minute


JoshuaKGoldberg:
### Suggestion

Splitting out of #11204: right now, each CI job typically spends over a minute in the _Install_ step:

https://github.com/typescript-eslint/typescript-eslint/blob/f8db925cdbaf2d85575de852ac9347d56e718604/.github/workflows/ci.yml#L58

Examples from a docs-only PR, #11200:

* https://github.com/typescript-eslint/typescript-eslint/actions/runs/14972488668/job/42056678571?pr=11200 took 57s
* https://github.com/typescript-eslint/typescript-eslint/actions/runs/14972488668/job/42056678626?pr=11200 took 1m 15s

That seems very long to me. In theory this should all be cached nicely, so all CI needs to do is download a lockfile and run a bit of disk linking/processing. Why does this take a full minute? Can we speed it up?

### Additional Info

💖

maschwenk:
<img width="691" alt="Image" src="https://github.com/user-attachments/assets/0f6b79a2-ca15-4bb7-99e3-75b7efa77521" />

fwiw looks like almost all of that is in the link step

kirkwaiblinger:
Guessing that's because of the postinstall build?

kirkwaiblinger:
After the merging of #11219, the install step is only taking about 15s. See CI logs on main: https://github.com/typescript-eslint/typescript-eslint/actions/runs/15326909449

PR #11135, opened 2025-05-05
chore: clean up ESLint config post `vitest` migration

aryaemami59:
## PR Checklist

- [X] Addresses an existing open issue: fixes #11132
- [X] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [X] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

PR #11219, opened 2025-05-14
chore: remove postinstall steps from CI to speed up prepare-install action

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue
   - #11205 
   - #11206

## Overview

Take all of the following with a grain of salt since this is not my area of expertise...

The postinstall script can take a while (locally, at least) since it builds the whole project from scratch. I'm not sure to what extent this is mitigated by nx caching in CI, but, locally, the difference between `SKIP_POSTINSTALL=true yarn` and `yarn` is 9 seconds  vs 68 seconds... So I'd think that explains the 1 minute prepare-install time mentioned in #11205 

For CI steps that need the tasks done by `yarn build`, nx's dependency graph _should_ still execute the required build steps, meaning that we'll mostly just be shifting running time out of the prepare-install step, and into the main CI job. So, likely not a huge overall speedup for those cases. However, for actions that _don't_ have any upstream build dependencies, like semantic-breaking-changes-pr, I would expect this to be a significant speedup. 

I'd be surprised if this didn't break anything if merged to main, tbh. Previously, many of the nx build dependencies weren't accurately declared, and, therefore, several commands depended on the "build everything" task built into `yarn install`. Note that @aryaemami59 has made quite a few fixes in this area, though! So perhaps things will work.

Merging this would also introduce a downside that CI failures would be harder for community contributors to debug since they would generally not know to do `SKIP_POSTINSTALL=true yarn && yarn my-failing-command` to replicate the behavior in CI.


JoshuaKGoldberg (dismissed):
💯 💯 _yes!_

IMO we should merge this ASAP and see what breaks. If I'm understanding the nx philosophy right then all of our scripts should know to run their dependencies first. In a perfect world I think that means we'd be able to eventually _delete_ the postinstall script and have it all automagically inferred. 🔪

bradzacher (approved):

JamesHenry:
There were some confusing nx config changes just merged in https://github.com/typescript-eslint/typescript-eslint/pull/11135 which have caused conflicts with this one

JamesHenry:
I'll share my thoughts in a loom to try and bring everyone on the same page

kirkwaiblinger:
> In a perfect world I think that means we'd be able to eventually _delete_ the postinstall script and have it all automagically inferred. 🔪 

For now, the postinstall script also does a bit of devx setup, in that it installs husky git hooks. So we probably still want it for that? For this PR, I went with `SKIP_POSTINSTALL=true` instead of `SKIP_POSTINSTALL_BUILD=true` (which I periodically use locally) because I figured we don't need husky git hooks in CI.

But yeah possibly the part after the `SKIP_POSTINSTALL_BUILD` condition maybe could go in the future?

JamesHenry:
The PR and loom are here: https://github.com/typescript-eslint/typescript-eslint/pull/11226

.github/workflows/ci.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         uses: ./.github/actions/prepare-install
         with:
           node-version: ${{ env.PRIMARY_NODE_VERSION }}
-      - run: yarn generate-configs
```
kirkwaiblinger, line 81: directly running `yarn generate-configs` seems to execute the yarn script directly, ignoring the `dependsOn`. So, I think `yarn nx run generate-configs` executes the nx task instead, which includes the necessary `dependsOn`.

bradzacher (approved):

PR #11243, opened 2025-05-23
feat(eslint-plugin): add a default-off option to autofix remove unused imports

### LINKED ISSUE #11223: Enhancement: [no-unused-vars] add a default-off option to autofix remove unused imports

https://typescript-eslint.io/rules/no-unused-vars/

### Description

Currently all `no-unused-vars` errors are not fixable. This is for good reasons, eg:
- side-effects are possible nearly everywhere in JS and it's impossible for us to know if an expression has a side-effect
- unused variables are a common part of the developer workflow and many people use autofix-on-save in their IDE -- so a fixer would cause "soon to be used" variables from being auto-deleted accidentally (an obviously frustrating DevX!)

With that being said there is one subset of unused variables which are _generally_ safe to autofix -- imported names. These are generally safe to fix because:
- the vast, vast, vast majority of modules do not have top-level side-effects
   - most bundlers and minifiers run on the assumption that modules DO NOT have side-effects!
- imports are mostly managed by the IDE (eg an IDE will auto-add import on usage of imported thing) and so removing a "soon to be used" import generally is low-impact from DevX.

There's evidence that the community wants this -- for example [`eslint-plugin-unused-imports`](https://www.npmjs.com/package/eslint-plugin-unused-imports) exists and is pretty widely used (they have ~8.4% of our weekly downloads). This plugin introduces two rules -- both are just monkey-patched extension rules on top of our `no-unused-vars` implementation where one rule only reports on variables and the other only reports on imports and provides a fixer.

My proposal is the following:
1) add a new option, say `enableAutofixRemoval: { imports: boolean }` with default `enableAutofixRemoval: { imports: false }` (i.e. disable the autofixer by default).
2) build an autofixer for imported names that does the following:
    1) if the unused name is a namespace import, remove the entire import declaration.
    2) if the unused name is a default import:
        1) if there are no used named imports in the declaration, remove the entire import declaration.
        2) if there are used named imports in the declaration, just remove the unused default import.
    3) if the unused name is a named import:
        1) if there are no other used names in the declaration, remove the entire import declaration.
        2) if there are used names in the declaration, just remove the unused named import.
3) if `enableAutofixRemoval.imports === true`
    1) then set the fixer from (2) as an **_auto-fixer_**
    2) else set the fixer from (2) as a **_suggestion fixer_**

### Fail

```TypeScript
todo delete section as it doesn't make sense
```

### Pass

```TypeScript
todo delete section as it doesn't make sense
```

### Namespace:
```ts
import * as Unused from 'foo';
import * as Used from 'foo';

export { Used };
```
autofixes to:
```ts
import * as Used from 'foo';

export { Used };
```

### Default:

```ts
import Unused1 from 'foo';
import Unused2, { Used } from 'foo';

export { Used };
```
autofixes to:
```ts
import { Used } from 'foo';

export { Used };
```

### Named:

```ts
import { Unused1 } from 'foo';
import Used1, { Unused2 } from 'foo';
import { Unused3, Used2 } from 'foo';
import Used3, { Unused4, Used4 } from 'foo';

export { Used1, Used2, Used3, Used4 };
```
autofixes to:
```ts
import Used1 from 'foo';
import { Used2 } from 'foo';
import Used3, { Used4 } from 'foo';

export { Used1, Used2, Used3, Used4 };
```

nayounsang:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11223

## Overview

**DRAFT**
- When add option `enableAutofixRemoval: { imports: true } `, auto fix and remove unused import statement
- Add test case in issue
- More test cases are needed. First of all, since it is my first contribution, it is very difficult, so I will think about it after uploading the PR. I can check the comments quickly.
- There may be something wrong with the comma handling. First, pass the test case.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unused-vars.ts
+++ b/packages/eslint-plugin/src/rules/no-unused-vars.ts
@@ -38,6 +38,9 @@ export type Options = [
       reportUsedIgnorePattern?: boolean;
       vars?: 'all' | 'local';
       varsIgnorePattern?: string;
+      enableAutofixRemoval?: {
+        imports: boolean;
+      };
     },
 ];
 
@@ -52,6 +55,9 @@ interface TranslatedOptions {
   reportUsedIgnorePattern: boolean;
   vars: 'all' | 'local';
   varsIgnorePattern?: RegExp;
+  enableAutofixRemoval?: {
+    imports: boolean;
+  };
 }
 
 type VariableType =
@@ -74,6 +80,7 @@ export default createRule<Options, MessageIds>({
       extendsBaseRule: true,
       recommended: 'recommended',
     },
+    fixable: 'code',
     messages: {
       unusedVar: "'{{varName}}' is {{action}} but never used{{additional}}.",
       usedIgnoredVar:
@@ -116,6 +123,16 @@ export default createRule<Options, MessageIds>({
                 type: 'string',
                 description:
                   'Regular expressions of destructured array variable names to not check for usage.',
+              },
+              enableAutofixRemoval: {
+                type: 'object',
+                properties: {
+                  imports: {
+                    type: 'boolean',
+                    description:
+                      'Whether to enable autofix for removing unused imports.',
+                  },
+                },
               },
               ignoreClassWithStaticInitBlock: {
                 type: 'boolean',
@@ -207,6 +224,10 @@ export default createRule<Options, MessageIds>({
             firstOption.destructuredArrayIgnorePattern,
             'u',
           );
+        }
+
+        if (firstOption.enableAutofixRemoval) {
+          options.enableAutofixRemoval = firstOption.enableAutofixRemoval;
         }
       }
 
@@ -687,6 +708,87 @@ export default createRule<Options, MessageIds>({
               data: unusedVar.references.some(ref => ref.isWrite())
                 ? getAssignedMessageData(unusedVar)
                 : getDefinedMessageData(unusedVar),
+              fix:
+                options.enableAutofixRemoval?.imports &&
+                unusedVar.defs.some(
+                  d => d.type === DefinitionType.ImportBinding,
+                )
+                  ? fixer => {
+                      const def = unusedVar.defs.find(
+                        d => d.type === DefinitionType.ImportBinding,
+                      );
+                      if (!def) {
+                        return null;
+                      }
+
+                      const source = context.sourceCode;
+                      const node = def.node;
+                      const decl = node.parent as TSESTree.ImportDeclaration;
+
+                      // Remove import declaration line if no specifiers are left
+                      if (decl.specifiers.length === 1) {
+                        const next = source.getTokenAfter(decl) ?? {
+                          range: [decl.range[1], decl.range[1]],
+                        };
+                        return fixer.removeRange([
+                          decl.range[0],
+                          next.range[0],
+                        ]);
+                      }
+
+                      // case: remove { unused }
+                      const restNamed = decl.specifiers.filter(
+                        s =>
+                          s === node &&
+                          s.type === AST_NODE_TYPES.ImportSpecifier,
+                      );
+                      if (restNamed.length === 1) {
+                        const nextBraceToken = source.getTokenAfter(node);
+                        const prevBraceToken = source.getTokenBefore(node);
+                        if (
+                          nextBraceToken?.value === '}' &&
+                          prevBraceToken?.value === '{'
+                        ) {
+                          // remove comma
+                          const prevComma =
+                            source.getTokenBefore(prevBraceToken);
+
+                          return fixer.removeRange([
+                            prevComma?.value === ','
+                              ? prevComma.range[0]
+                              : prevBraceToken.range[0],
+                            nextBraceToken.range[1],
+                          ]);
+                        }
+                      }
+
+                      // case: Remove comma after node
+                      const nextCommaToken = source.getTokenAfter(node);
+                      if (nextCommaToken?.value === ',') {
+                        const nextToken = source.getTokenAfter(nextCommaToken, {
+                          includeComments: true,
+                        });
+
+                        return fixer.removeRange([
+                          node.range[0],
+                          nextToken
+                            ? nextToken.range[0]
+                            : nextCommaToken.range[1],
+                        ]);
+                      }
+
+                      // case: Remove comma before node
+                      const prevCommaToken = source.getTokenBefore(node);
+                      if (prevCommaToken?.value === ',') {
+                        return fixer.removeRange([
+                          prevCommaToken.range[0],
+                          node.range[1],
+                        ]);
+                      }
+                      // Remove the current specifier and all tokens until the next specifier
+                      return fixer.remove(node);
+                    }
+                  : undefined,
             });
 
             // If there are no regular declaration, report the first `/*globals*/` comment directive.
--- a/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
@@ -1734,6 +1734,122 @@ export {};
       ],
       filename: 'foo.d.ts',
     },
+    {
+      code: `
+import * as Unused from 'foo';
+import * as Used from 'bar';
+export { Used };
+      `,
+      errors: [
+        {
+          column: 13,
+          data: {
+            action: 'defined',
+            additional: '',
+            varName: 'Unused',
+          },
+          line: 2,
+          messageId: 'unusedVar',
+        },
+      ],
+      options: [{ enableAutofixRemoval: { imports: true } }],
+      output: `
+import * as Used from 'bar';
+export { Used };
+      `,
+    },
+    {
+      code: `
+import Unused1 from 'foo';
+import Unused2, { Used } from 'bar';
+export { Used };
+      `,
+      errors: [
+        {
+          column: 8,
+          data: {
+            action: 'defined',
+            additional: '',
+            varName: 'Unused1',
+          },
+          line: 2,
+          messageId: 'unusedVar',
+        },
+        {
+          column: 8,
+          data: {
+            action: 'defined',
+            additional: '',
+            varName: 'Unused2',
+          },
+          line: 3,
+          messageId: 'unusedVar',
+        },
+      ],
+      options: [{ enableAutofixRemoval: { imports: true } }],
+      output: `
+import { Used } from 'bar';
+export { Used };
+      `,
+    },
+    {
+      code: `
+import { Unused1 } from 'foo';
+import Used1, { Unused2 } from 'bar';
+import { Used2, Unused3 } from 'baz';
+import Used3, { Unused4, Used4 } from 'foobar';
+export { Used1, Used2, Used3, Used4 };
+      `,
+      errors: [
+        {
+          column: 10,
+          data: {
+            action: 'defined',
+            additional: '',
+            varName: 'Unused1',
+          },
+          line: 2,
+          messageId: 'unusedVar',
+        },
+        {
+          column: 17,
+          data: {
+            action: 'defined',
+            additional: '',
+            varName: 'Unused2',
+          },
+          line: 3,
+          messageId: 'unusedVar',
+        },
+        {
+          column: 17,
+          data: {
+            action: 'defined',
+            additional: '',
+            varName: 'Unused3',
+          },
+          line: 4,
+          messageId: 'unusedVar',
+        },
+        {
+          column: 17,
+          data: {
+            action: 'defined',
+            additional: '',
+            varName: 'Unused4',
+          },
+          line: 5,
+          messageId: 'unusedVar',
+        },
+      ],
+      options: [{ enableAutofixRemoval: { imports: true } }],
+      output: `
+import Used1 from 'bar';
+import { Used2 } from 'baz';
+import Used3, { Used4 } from 'foobar';
+export { Used1, Used2, Used3, Used4 };
+      `,
+    },
   ],
 
   valid: [
@@ -2939,39 +3055,6 @@ export namespace Foo {
 declare module 'foo' {
   export import Bar = Something.Bar;
   const foo: 1234;
-}
-      `,
-      filename: 'foo.d.ts',
-    },
-    {
-      code: `
-export import Bar = Something.Bar;
-const foo: 1234;
-      `,
-      filename: 'foo.d.ts',
-    },
-    {
-      code: `
-declare module 'foo' {
-  export import Bar = Something.Bar;
-  const foo: 1234;
-  export const bar: string;
-  export namespace NS {
-    const baz: 1234;
-  }
-}
-      `,
-      filename: 'foo.d.ts',
-    },
-    {
-      code: `
-export namespace Foo {
-  export import Bar = Something.Bar;
-  const foo: 1234;
-  export const bar: string;
-  export namespace NS {
-    const baz: 1234;
-  }
 }
       `,
       filename: 'foo.d.ts',


packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
```diff
@@ -2946,39 +3062,6 @@ declare module 'foo' {
     {
       code: `
 export import Bar = Something.Bar;
-const foo: 1234;
```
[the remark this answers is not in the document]
nayounsang: revert it

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -687,6 +708,80 @@ export default createRule<Options, MessageIds>({
...
+                    const def = unusedVar.defs.find(
+                      d => d.type === DefinitionType.ImportBinding,
+                    );
+                    if (!def) {
+                      return null;
+                    }
+
+                    const source = context.sourceCode;
+                    const node = def.node;
+                    const decl = node.parent as TSESTree.ImportDeclaration;
```
kirkwaiblinger: I think that this will be unsound for `import = require`-style imports, for example
```ts
import x = require('foo')
```

Let's be sure to include some test coverage with that style of `import`.

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -687,6 +708,80 @@ export default createRule<Options, MessageIds>({
               data: unusedVar.references.some(ref => ref.isWrite())
                 ? getAssignedMessageData(unusedVar)
                 : getDefinedMessageData(unusedVar),
+              fix: options.enableAutofixRemoval?.imports
```
kirkwaiblinger: This seems useful as a suggestion, even if the autofix option is disabled. Consider using `getFixOrSuggest` to conditionally provide this fix as a suggestion?

https://github.com/typescript-eslint/typescript-eslint/blob/5b2ebcdea9011a6de7e657118874ae85c08d5a0e/packages/eslint-plugin/src/util/getFixOrSuggest.ts#L3-L21

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -687,6 +708,80 @@ export default createRule<Options, MessageIds>({
...
+                    const beforeNodeToken = source.getTokenBefore(node);
+                    const prevBeforeNodeToken = beforeNodeToken
+                      ? source.getTokenBefore(beforeNodeToken)
+                      : null;
+
+                    // Remove import declaration line if no specifiers are left, import unused from 'a';
+                    if (decl.specifiers.length === 1) {
+                      return fixer.removeRange([
+                        decl.range[0],
+                        decl.range[1] + 1, // +1 to include "\n"
```
kirkwaiblinger: > to include "\n"

Why do we want to do this exactly? This causes a bug with this technically valid code (please include as a test case - hint, you'll need to use `noFormat`):
```ts
import x from 'foo';import y from 'bar'
```

My instinct is to say we're better off just leaving a blank line and letting the user decide how to format it afterwards (whether by hand or with prettier or similar).

kirkwaiblinger (changes requested):
A great start! 

I'm leaving a bit of a drive-by review since I see that there are several major TODOs in the code that need to get resolved. Feel free to reach out if you're looking for help with those! In the meantime, I'm going to leave a Changes Requested review so that this isn't in our ready-to-review queue.

Thanks!

packages/eslint-plugin/src/rules/no-unused-vars.ts
nayounsang: Oh, nice catch. This is `TSImportEqualsDeclaration` type.
I test with this code:
```ts
    {
      code: `
import x = require('foo');
import y = require('bar');
export { y };
      `,
      errors: [{ messageId: 'unusedVar' }],
      options: [{ enableAutofixRemoval: { imports: true } }],
      output: `
import y = require('bar');
export { y };
      `,
    }
```
In the existing logic, an error occurs and a more accurate logic is used rather than type casting.
related commit: https://github.com/typescript-eslint/typescript-eslint/pull/11243/commits/11ac4fa7833a17b0883875952b82be34ad39126d

packages/eslint-plugin/src/rules/no-unused-vars.ts
nayounsang: Good idea. However, it would be better to provide suggestions to existing test cases and then resolve [this comment](https://github.com/typescript-eslint/typescript-eslint/pull/11243#discussion_r2131085820) before processing.

packages/eslint-plugin/src/rules/no-unused-vars.ts
nayounsang: > we're better off just leaving a blank line and letting the user decide how to format it afterwards (whether by hand or with prettier or similar).

Wow, that's right. That should have been it. Plus, it lowers the difficulty of solving the problem and the code becomes simpler.
related commit: https://github.com/typescript-eslint/typescript-eslint/pull/11243/commits/9f76441a8a109bbc4d63317f95eeb321769ef300 & https://github.com/typescript-eslint/typescript-eslint/pull/11243/commits/633d2c8c00d7882f7e98331bec09d24fd0502288

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
```diff
@@ -1734,6 +1734,206 @@ export {};
...
+      errors: [{ messageId: 'unusedVar' }],
+      options: [{ enableAutofixRemoval: { imports: true } }],
+      output: `
+
+import y = require('bar');
+export { y };
+      `,
+    },
+    // TODO: Logic to remove multiple unused vars in one-line
+    //     {
```
nayounsang: The reason for this was that the **fixer conflict** was right.
```
import Unused1, { Unused2, Unused3 } from "foo";
import { Unused4, Unused5, Unused6 } from "bar";
```
The cases that occur are as follows:
```ts
import Unused1, { Unused2 } from "foo";
import { Unused4, Unused5, Unused6 } from "bar";
```
1. If the number of `ImportDefaultSpecifiers` + the number of `ImportSpecifiers` in a single line of `ImportDeclaration` is 2 or more and all of them are **unused**.
[playground](https://typescript-eslint.io/play#ts=5.6.3&showAST=es&fileType=.tsx&code=JYWwDg9gTgLgBAVQHYFcDOBTAJgRgDRwDeiqmWATHAL5wBmUEIcARLRBMwNwBQoksREumwAWAsmFYArONLYAbNToMmzAEYBDKFyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

```ts
import Used1, { Unused4, Unused5, Unused6 } from "bar";
export { Used1 };
```
2. If there are more than 2 `ImportSpecifiers` in a single line of `ImportDeclaration` and they are all **unused**
[playground](https://typescript-eslint.io/play#ts=5.6.3&showAST=es&fileType=.tsx&code=JYWwDg9gTgLgBAVQM4FMAmBGANHA3ogOwFdU0AWHBY0gVkuvQDY4BfOAMyghDgCIAjAIZReAbgBQKAB6RYeRKQytRQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

WHY?
This is because, after deleting one `ImportSpecifier` or `ImportDefaultSpecifer` respectively, it tries to delete a line if **only one remains** and it is not used.

nayounsang: Current
1. Fixer see `Unused1`, and remove `range(Unused1,)`
2. Fixer see `Unused2`, and remove `range(Unused2,)`
3. Fixer see `Unused3`, since there is only one Specifier left, fixer remove `range(Declaration)`, **CONFLICT**

Plan (Please evaluate it.)
1. Calculate all ranges that need to be deleted for each ImportDeclaration
2. Merge conflicting ranges into a larger range through a sweeping algorithm
3. When deleting a (Default) Specifier, delete the range calculated in step 2.

For ex)
```ts
import Unused1, { Unused2, Unused3 } from "foo";
```
If you want to quick fix for `Unused2`, expect
```ts
```

packages/eslint-plugin/src/rules/no-unused-vars.ts
nayounsang: Add new message: `unusedVarSuggestion`. However, I'm concerned about the proposed message: `Remove unused variable.`. Any thoughts?

related commit: https://github.com/typescript-eslint/typescript-eslint/pull/11243/commits/2c8b7406a98205f95f5413cb0522f23c824a7abf & https://github.com/typescript-eslint/typescript-eslint/pull/11243/commits/9a15d4955b77dcd35a50e1e8f93b6517336cb58a

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
nayounsang: image
If I need to erase a larger range and a collision occurs, I may want to split the larger range.

nayounsang: This seems like it would require a conversation with the maintainer and they will implement it and Resolve it

nayounsang:
@kirkwaiblinger Hi. I committed my code about your review and TODOs! Please make sure I'm on the right track.
The problem to be solved now is: Problem with unformatted code. (Based on 0-index) Odd specifiers are not deleted.
Um... I run console.log in the local and see results of the ranges need to be deleted, and in this case, the range to delete did not conflict! I don't know why.
```ts
// test case
import {A,B,C,D,E,F,G,Used1,H,I,J,K} from 'foo';
export {Used1};

// expect
import {Used1} from 'foo';
export {Used1};

// received
import {B,D,F,Used1,I,K} from 'foo'
export {Used1};
```
How can I solve this problem? My idea is to merge the contiguous segments and remove multiple specifiers at once.
For ex
```ts
import {A,B,C,D,E,F,G,Used1,H,I,J,K} from 'foo';  // ->
import {Used1,H,I,J,K} from 'foo'; // ->
import {Used1} from 'foo';
```

nayounsang:
@bradzacher 
yes, it is not supported. When handling this, a conflict occurs in the range that the fixer removes, and is therefore ignored.
```ts
// received 
import {a,b} from 'foo';
// expected

// received
import a, {b,c} from 'foo';
console.log(a);
// expected
import a from 'foo';
console.log(a)
```
I'm curious about the direction.
I think this case could be implemented with logic that removes multiple unused specifiers at once, rather than deleting each specifier individually. Is this the right approach?
```js
// To express it in pseudocode...
const allNamedSpecifersCount = decl.specifiers.filter(
   s => s.type === AST_NODE_TYPES.ImportSpecifier,
);

if (unusedNamedSpecifiersCount ===  allNamedSpecifiersCount) {

  return fixer.removeRange([
    startBraceToken.range[0],
    endBraceToken.range[1],
   ]);

}
```

- That case wasn't found by [eslint](https://github.com/eslint/eslint/blob/main/tests/lib/rules/no-unused-vars.js) either. It delete specifiers one by one. 
- I'm not sure if it would be a good idea to have fixer affect other variables that it's not currently handling.

Is there another better way? Anyway I'll try implementing it my way and then req for review.
If you have any comments before then, please let me know! It would be very helpful.

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -642,6 +672,44 @@ export default createRule<Options, MessageIds>({
...
+        const unusedImportSpecifiersMap = new Map<
+          string,
+          (
+            | TSESTree.ImportDefaultSpecifier
+            | TSESTree.ImportNamespaceSpecifier
+            | TSESTree.ImportSpecifier
+          )[]
+        >();
+
+        function getSpecifiersMapKey(range: TSESTree.Range): string {
```
nayounsang: I might consider using the decl itself as the key, rather than the location of the import decl.

1. Will it break the reference?
2. Will it use a lot of memory?

We should consider these two things.

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -642,6 +672,44 @@ export default createRule<Options, MessageIds>({
       'Program:exit'(programNode): void {
         const unusedVars = collectUnusedVariables();
 
+        // collect unused named import specifiers for import decl
+        const unusedImportSpecifiersMap = new Map<
+          string,
```
nayounsang: In the same context, I should use `WeakMap` to apply that [comment](https://github.com/typescript-eslint/typescript-eslint/pull/11243/files#r2297402435).

packages/eslint-plugin/src/rules/no-unused-vars.ts
nayounsang: For reference variables, there is no need to worry about memory since only a reference is used when using it as a key.

nayounsang: Anyway, the current code does not break the reference. pass in local, pray for ci

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -681,12 +746,101 @@ export default createRule<Options, MessageIds>({
...
+              if (unusedSpecifiers && unusedSpecifiers.length > 0) {
+                const restNamed = decl.specifiers.filter(
+                  s => s.type === AST_NODE_TYPES.ImportSpecifier,
+                );
+                const lastBraceToken = source.getTokenAfter(
+                  unusedSpecifiers[unusedSpecifiers.length - 1],
+                );
+                const firstBraceToken = source.getTokenBefore(
+                  unusedSpecifiers[0],
+                );
```
nayounsang, line 793: This logic assume that unusedVariables collected sequentially from the code.
Deductively, it works fine.
But,  I'm not really sure if that's true in all cases.
```ts
      const analysisResults = collectVariables(context);
      const variables = [
        ...Array.from(analysisResults.unusedVariables, variable => ({
          used: false,
          variable,
        })),
        ...Array.from(analysisResults.usedVariables, variable => ({
          used: true,
          variable,
        })),
      ];
```

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -74,8 +85,13 @@ export default createRule<Options, MessageIds>({
       extendsBaseRule: true,
       recommended: 'recommended',
     },
+    fixable: 'code',
+    // If generate suggest dynamically, disable the eslint rule.
```
JoshuaKGoldberg, line 89:
```suggestion
    // Suggestions are generated dynamically by getFixOrSuggest
```

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -74,8 +85,13 @@ export default createRule<Options, MessageIds>({
...
+    // eslint-disable-next-line eslint-plugin/require-meta-has-suggestions
+    hasSuggestions: true,
     messages: {
       unusedVar: "'{{varName}}' is {{action}} but never used{{additional}}.",
+      unusedVarSuggestion: 'Remove unused variable.',
```
JoshuaKGoldberg, line 94: [Bug] This is specific to imports, so we should be more specific here I think.

```suggestion
      removeUnusedImport: 'Remove unused import.',
```

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -642,6 +672,41 @@ export default createRule<Options, MessageIds>({
       'Program:exit'(programNode): void {
         const unusedVars = collectUnusedVariables();
 
+        /**
+         * Collect unused import specifiers for the import declaration.
+         * Access the map by the import declaration.
+         */
+        const unusedImportSpecifiersMap = new WeakMap<
```
JoshuaKGoldberg, line 679: [Style] A few interlinked suggestions:

* This comment doesn't add anything IMO: the variable is already named "unused import specifiers", and its type says that import declarations are the keys
* The `Map` suffix in the name isn't necessary either - it's a map type already
* It's kind of confusing to read through so much new logic, where a few dozen lines are just to populate a map. Let's move the logic into a function outside of the rule.

```suggestion
        const unusedImportSpecifiers =
          collectUnusedImportSpecifiers(unusedVars);
```

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
```diff
@@ -1715,6 +1977,242 @@ export {};
...
+    },
+    {
+      code: `
+import { Unused1, Unused2 } from 'foo';
+      `,
+      errors: [{ messageId: 'unusedVar' }, { messageId: 'unusedVar' }],
+      options: [{ enableAutofixRemoval: { imports: true } }],
+      output: `
+
+      `,
```
JoshuaKGoldberg, line 2188: Moving https://github.com/typescript-eslint/typescript-eslint/pull/11243#pullrequestreview-3126171006 & https://github.com/typescript-eslint/typescript-eslint/pull/11243#issuecomment-3194426223 into a thread:

> yes, it is not supported. When handling this, a conflict occurs in the range that the fixer removes, and is therefore ignored.

Just to confirm what it looks like the implementation went with: conflicts are ok! ESLint runs up to 10 passes of auto-fixing. If there are 2-3 fixes in a range and only 1-2 get applied in a time, that's fine. It'll get all of them eventually. Fixes are by design meant to always improve codebase quality. They never backtrack.

JoshuaKGoldberg (approved):
Looking good! Just a few small touchup suggestions from me, nothing serious.

If @bradzacher and/or @kirkwaiblinger have time to review that'd be great - so I'll hold off on merging for a bit. But if they don't (they're busy this month & we've got a lot of PRs going on) then I feel comfortable merging this soon. 

Thanks! 😊

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
nayounsang: Okay. Then, I think I can just delete the complex (conflict-causing) test case and the logic surrounding [this comment](https://github.com/typescript-eslint/typescript-eslint/pull/11243#discussion_r2486841150) (this will greatly simplify the overall logic). 
I'll wait for other people's opinions.

nayounsang:
TODO: This pr adds options separate from [eslint's options](https://eslint.org/docs/latest/rules/no-unused-vars#options), so I think it needs to add something in [the documentation](https://typescript-eslint.io/rules/no-unused-vars/#options).

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -681,12 +746,101 @@ export default createRule<Options, MessageIds>({
...
+                return fixer.removeRange([
+                  beforeNodeToken.range[0],
+                  node.range[1],
+                ]);
+              }
+
+              return null;
+            };
+
             context.report({
```
kirkwaiblinger, line 829: (readability refactor)

This single `report()` call handles too many things right now. It handles

1. Unused _variables_, which can never be removed by a fix or suggestion
2. Unused _imported symbols_, which can _always_ be removed by a fix or a suggestion (whether fix or suggestion is controlled by the `enableAutofixRemoval` option. These should have a fix message that makes sense semantically (like @JoshuaKGoldberg's comment https://github.com/typescript-eslint/typescript-eslint/pull/11243/files#r2486826380))

I propose: can these branches be split apart please? The long `const fix = ...` function should be part of the import branch only, since it's completely inapplicable to the variables branch

roughly,
```ts
if (!isUnusedImport) {
   context.report(
     // ...
     // no fixer message or function
   )
} else {
   const fix = ...; // complicated fixer.
   context.report({
     // imports can never be reassigned in valid TS code so probably don't need the "written to but not read" message here at all
     // ...
     ...getFixOrSuggest(...)
   });
}

File: packages/eslint-plugin/src/rules/no-unused-vars.ts [hunk 14]
```diff
@@ -642,6 +672,41 @@ export default createRule<Options, MessageIds>({
...
+            | TSESTree.ImportDefaultSpecifier
+            | TSESTree.ImportNamespaceSpecifier
+            | TSESTree.ImportSpecifier
+          )[]
+        >();
+
+        for (const unusedVar of unusedVars) {
+          // Find the import statement
+          const def = unusedVar.defs.find(
+            d => d.type === DefinitionType.ImportBinding,
```
kirkwaiblinger, line 691: This might have some odd edge cases around declaration merging. For example, might be worth looking into test cases along the lines of 

```ts
// should this have a fixer? Maybe not?
import { A } from 'a';

interface A {
  anotherProperty: string
}
```

Needs some investigation into the edge cases but let's be sure to add a few test cases for strange cases if it's possible for there to be cases where other declarations/declaration types may also be present?

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -642,6 +672,41 @@ export default createRule<Options, MessageIds>({
...
+          );
+          if (!def) {
+            continue;
+          }
+
+          const node = def.node;
+          const decl = node.parent;
+          if (
+            decl.type !== AST_NODE_TYPES.ImportDeclaration ||
+            node.type === AST_NODE_TYPES.TSImportEqualsDeclaration
```
kirkwaiblinger, line 701: If I replace this with `if (false)`, no tests fail. Is this a condition that can be reached/needs to be handled?

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -681,12 +746,101 @@ export default createRule<Options, MessageIds>({
               },
             };
 
+            const fixer: TSESLint.ReportFixFunction = fixer => {
+              // Find the import statement
+              const def = unusedVar.defs.find(
```
kirkwaiblinger, line 751: Same comment as https://github.com/typescript-eslint/typescript-eslint/pull/11243/files#r2590828734 

Also is there a way to eliminate this duplication?

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -681,12 +746,101 @@ export default createRule<Options, MessageIds>({
...
+                d => d.type === DefinitionType.ImportBinding,
+              );
+              if (!def) {
+                return null;
+              }
+
+              const source = context.sourceCode;
```
kirkwaiblinger, line 758: Nit - if you're going to alias this, call it `sourceCode`. But, preferably just don't alias it?

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -681,12 +746,101 @@ export default createRule<Options, MessageIds>({
...
+              const node = def.node;
+              const decl = node.parent;
+              if (decl.type !== AST_NODE_TYPES.ImportDeclaration) {
+                // decl.type is Program, import foo = require('bar');
+                return fixer.remove(node);
+              }
+
+              const afterNodeToken = source.getTokenAfter(node);
```
kirkwaiblinger, line 766: nit - why are these defined so far away from where they're used? Why not just move them down to where they're checked?

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
```diff
@@ -832,24 +1080,38 @@ export = Foo;
       ],
     },
     {
-      code: `
+      code: noFormat`
```
kirkwaiblinger, line 1083: Why the `noFormat`? This test doesn't appear to test anything relating to formatting.

```suggestion
      code: `
```

kirkwaiblinger (changes requested):

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
kirkwaiblinger: Echoing what Josh has said, having a fix per variable is perfectly ok. This gets sorted out by eslint, by taking multiple passes.

I checked out commit [eb3fd2b](https://github.com/typescript-eslint/typescript-eslint/pull/11243/commits/eb3fd2b2443f45b4dbb3d766e01ea1b0d453f67e) and it works just fine for a multi-variable fix.

```ts
import { unused1, unused2 } from 'foo';
```

However, to write a test case for it, you'll want to use the array of fixes, see https://typescript-eslint.io/packages/rule-tester/#usage (search for "multi-pass"). 

In short, your test case will look like

```ts
    {
      code: `
import { unused1, unused2 } from 'foo';
      `,
      errors: [{ messageId: 'unusedVar' }, { messageId: 'unusedVar' }],
      options: [{ enableAutofixRemoval: { imports: true } }],
            output: [
              `
      import {  unused2 } from 'foo';
            `,
              `

            `,
            ],
    },
```

FWIW - this multipass strategy is exactly how the community solution eslint-plugin-unused-imports works. Consider generally studying their implementation at https://github.com/sweepline/eslint-plugin-unused-imports/blob/master/src/rules/predicates.ts#L74-L130

kirkwaiblinger: > Okay. Then, I think I can just delete the complex (conflict-causing) test case and the logic surrounding [this comment](https://github.com/typescript-eslint/typescript-eslint/pull/11243#discussion_r2486841150) (this will greatly simplify the overall logic). I'll wait for other people's opinions.

Note -  please do NOT delete the test cases around multiple variables. This is a critical aspect of this feature! But preferably let it be handled implicitly rather than explicitly in the implementation, and ensure that it is tested! 

Also, please write test cases where multiple _suggestions_ are generated (i.e.
```ts
import { unused1, unused2 } from 'foo';
```
but with `enableAutofixRemoval: { imports: false } `

PR #11254, opened 2025-05-30
docs: fix up config links

kirkwaiblinger:
## PR Checklist

## Overview

Affects https://typescript-eslint.io/users/configs

Addresses 2 issues.

1. https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslintrc/eslint-recommended.ts is a bad link to use for `eslint-recommended`. It's technically true but doesn't show its contents. So, I've changed it to https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts

2. The wording "See  `configs/<config-name>.ts` for the exact contents of this config" doesn't really work because those configs are now at `configs/eslintrc/<config-name>.ts` (with the exception of bullet point 1, which is at `configs/eslint-recommended-raw.ts`). Since putting `eslintrc` and `*-raw` in the link text is IMO distracting, I've opted to change the wording to "See `the source code for this config` for the exact contents"... Not dead set on this solution, but I'd like to avoid putting the technically-false `configs/<config-name>.ts` in the text.

### DIFF
--- a/docs/users/Shared_Configurations.mdx
+++ b/docs/users/Shared_Configurations.mdx
@@ -203,7 +203,7 @@ module.exports = {
 </Tabs>
 
 Some rules also enabled in `recommended` default to more strict settings in this configuration.
-See [the source code for this config](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslintrc/strict.ts) for the exact contents.
+See [the source code for the `strict` config](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslintrc/strict.ts) for the exact contents.
 
 :::tip
 We recommend a TypeScript project extend from `plugin:@typescript-eslint/strict` only if a nontrivial percentage of its developers are highly proficient in TypeScript.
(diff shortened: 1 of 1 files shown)


bradzacher (dismissed):

docs/users/Shared_Configurations.mdx
```diff
@@ -144,7 +144,7 @@ module.exports = {
 </TabItem>
 </Tabs>
 
-See [`configs/recommended.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslintrc/recommended.ts) for the exact contents of this config.
+See [the source code for this config](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslintrc/recommended.ts) for the exact contents.
```
JoshuaKGoldberg, line 147: https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html - the problem with having the same text on all links is they can't be disambiguated from just that text. Maybe...

```suggestion
See [the source code for the `recommended` config](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslintrc/recommended.ts) for the exact contents.
```

would be a good fix?

kirkwaiblinger: Yes, agreed that that's better. Thanks for the suggestion!

JoshuaKGoldberg (approved):
🙌

PR #11260, opened 2025-05-31
chore: upgrade Docusaurus to 3.8

benmccann:
## PR Checklist

- [ ] Addresses an existing open issue: fixes #000
- [ ] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [ ] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Upgrades from Docusaurus 3.6.1 (what's in the lockfile) to Docusaurus 3.8.1

I have never really used React or Docusaurus, so someone else might have to take this over as it's turned out to be more complicated than I envisioned

The copied over components significantly changed: https://docusaurus.io/blog/releases/3.8#code-block-refactor
I ran `npx docusaurus swizzle @docusaurus/theme-classic CodeBlock/Buttons --eject` and then made a new `TryInPlaygroundButton` pulled out from the old code as this new version refactors things to break up the components

PR #11262, opened 2025-05-31
chore: upgrade to Docusaurus 3.7

benmccann:
## PR Checklist

## Overview

Upgrades to Docusaurus 3.7. It would be nice to upgrade to Docusaurus 3.8, but I can't figure out how to do that with this repo's setup. See here for my attempt at that: https://github.com/benmccann/typescript-eslint/tree/docusaurus-3.8

### DIFF
--- a/packages/website/package.json
+++ b/packages/website/package.json
@@ -28,12 +28,12 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.24.4",
-    "@docusaurus/core": "^3.2.1",
-    "@docusaurus/plugin-client-redirects": "^3.2.1",
-    "@docusaurus/plugin-pwa": "^3.2.1",
-    "@docusaurus/preset-classic": "^3.2.1",
-    "@docusaurus/remark-plugin-npm2yarn": "^3.2.1",
-    "@docusaurus/theme-common": "^3.2.1",
+    "@docusaurus/core": "~3.7.0",
+    "@docusaurus/plugin-client-redirects": "~3.7.0",
+    "@docusaurus/plugin-pwa": "~3.7.0",
+    "@docusaurus/preset-classic": "~3.7.0",
+    "@docusaurus/remark-plugin-npm2yarn": "~3.7.0",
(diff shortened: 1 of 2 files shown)


packages/website/package.json
```diff
@@ -55,7 +55,7 @@
     "typescript": "*"
   },
   "devDependencies": {
-    "@docusaurus/module-type-aliases": "^3.2.1",
+    "@docusaurus/module-type-aliases": "~3.7.0",
```
JoshuaKGoldberg, line 58: Almost everywhere else uses `^`. I think the two existing `~`s might have been unintentional. Is there a reason to go with `~` here?

If it doesn't make a strong difference, I'd think we'd rather stick with `^` for consistency.

JoshuaKGoldberg (approved):
🚀 Functionally it all looks good to me. Just checking in on the `^`/`~`?

packages/website/package.json
benmccann: The reason was because larger changes are needed to adopt 3.8. Ideally we'd just upgrade to 3.8, but I was having trouble with that given that I'm not a React developer. I don't know if those changes would be hard for someone familiar with Docusaurus and React, but they might be a bit beyond me.

JoshuaKGoldberg: Ah gotcha, so this is to make sure we don't accidentally bump to a `3.8` of something then? In that case if it makes it easier to not accidentally bump, that seems reasonable to me.

JoshuaKGoldberg (approved):
👍 assuming we'd switch back to `^` once 3.8 happens.

benmccann:
Yes, I was thinking we'd switch back to `^` for 3.8. You can see my attempt at it here: https://github.com/typescript-eslint/typescript-eslint/pull/11260. I'm not sure if I should leave it open or not as I'm not sure whether I can complete it, but it should be directionally correct and provide details about what needs to happen if there's someone who either wants to take it over or make their own attempt.

Issue #11257, opened 2025-05-30
Bug: [no-floating-promises] doesn't detect nested arrays to `Promise.all()`


libre-man:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.
- [x] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [x] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=IYZwngdgxgBAZgV2gFwJYHsI2ACgJQwDeAUDNgO7CrIwAKATugLaogCmAdMADbc4DapGPwbNWnemxDpuANzb4AugBohivMQC%2BxIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUFs%2B5AAgC4CeADigMbQCWphAtCvNQHaED0LA9vQGbxcAhoVYBzeqWhcAttWQp0URNCnRI4MAF8QmoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
async function a() {
  await Promise.all([
    [Promise.resolve()],
  ])
}
```

### ESLint Config

```javascript
{
  "rules": {
            "@typescript-eslint/no-floating-promises": "error"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I would expect this code to be flagged, as `Promise.all([[promise]])` doesn't await `promise`.

### Actual Result

There was no issue detected.

### Additional Info

This is quite a subtle bug and it would be great if `typescript-eslint` could catch it.

kirkwaiblinger:
Hi @libre-man!

Note that https://github.com/typescript-eslint/typescript-eslint/issues/1804 is related and also solves this specific case, by different logic.

[...]

PR #11267, opened 2025-06-01
feat(eslint-plugin): [await-thenable] report invalid (non-promise) values passed to promise aggregator methods

### LINKED ISSUE #1804: [await-thenable] Unexpected `await` of a non-Promise does not work if wrapped with Promise.all

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1804

## Overview

This PR tackles #1804 and adjusts the rule to report on invalid (non-promise) input passed to promise aggregator methods (`Promise.all`, `Promise.race`, `Promise.allSettled`, and `Promise.any`):

```ts
declare const x: number[];

// Unexpected iterator of non-Promise (non-"Thenable") values passed to promise aggregator.
Promise.all(x);
```

### DIFF
--- a/packages/eslint-plugin/src/rules/await-thenable.ts
+++ b/packages/eslint-plugin/src/rules/await-thenable.ts
@@ -265,13 +265,12 @@ function getValueTypesOfArrayLike(
   }
 
   if (checker.isArrayLikeType(type)) {
-    const indexType = type.getNumberIndexType();
-
-    if (indexType != null) {
-      return [indexType];
-    }
-
-    return null;
+    return [
+      nullThrows(
+        type.getNumberIndexType(),
+        'number index type should exist on an array-like',
+      ),
+    ];
   }
 
   // `Iterable<...>`
--- a/packages/eslint-plugin/tests/rules/await-thenable.test.ts
+++ b/packages/eslint-plugin/tests/rules/await-thenable.test.ts
@@ -342,6 +342,17 @@ class C<R extends unknown> {
 
     {
       code: `
+// @ts-expect-error
+Promise.all();
+      `,
+    },
+    {
+      code: `
+Promise.all([,]);
+      `,
+    },
+    {
+      code: `
 declare const x: unknown;
 
 // @ts-expect-error


libre-man:
It would be great if a test case for for `Promise.all(x)` where `x` is `Array<Array<Promise<T>>` could be added, as this the bug that we recently had in #11257.

JoshuaKGoldberg (dismissed):
Liu Kang from Mortal Kombat clenching his fist as fire erupts from it. Caption: "FLAWLESS VICTORY"

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -39,6 +43,8 @@ export default createRule<[], MessageId>({
       convertToOrdinaryFor: 'Convert to an ordinary `for...of` loop.',
       forAwaitOfNonAsyncIterable:
         'Unexpected `for await...of` of a value that is not async iterable.',
+      invalidPromiseAggregatorInput:
+        'Unexpected iterator of non-Promise (non-"Thenable") values passed to promise aggregator.',
```
kirkwaiblinger: Technically, I think the wording should be like so?

```suggestion
        'Unexpected iterable of non-Promise (non-"Thenable") values passed to promise aggregator.',
```

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -176,3 +209,55 @@ export default createRule<[], MessageId>({
...
+  // non array/tuple/iterable types already show up as a type error
+  if (!isIterable(type, checker)) {
+    return false;
+  }
+
+  for (const part of tsutils.unionConstituents(type)) {
+    if (tsutils.isTypeReference(part)) {
+      const typeArguments = checker.getTypeArguments(part);
+
+      // only check the first type argument of `Iterator<...>` or `Array<...>`
```
kirkwaiblinger: The heuristics here seem to have minor edge case bugs.

```ts
interface MyArray<Unused, T> extends Array<T> {};

declare const arrayOfNull: MyArray<Promise<void>, null>;
declare const arrayOfPromises: MyArray<null, Promise<void>>;

Promise.all(arrayOfNull); // no report; should report
Promise.all(arrayOfPromises); // does report; shouldn't report
```

Note that if you switch `interface` to `type`, this works correctly 🧐

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -1,10 +1,12 @@
 import type { TSESLint, TSESTree } from '@typescript-eslint/utils';
```
kirkwaiblinger: Noting that there was talk on the issue of [adding an option](https://github.com/typescript-eslint/typescript-eslint/issues/1804#issuecomment-604560677) for `.some()` vs `.every()` semantics for whether to report things like `Promise.all([2, Promise.resolve(3)])`.

I think the `.some()` semantics used here are good and if someone wants the `.every()` semantics option after it's released we can entertain that as a followup, but no need for this version.

packages/eslint-plugin/src/rules/await-thenable.ts
ronami: Yes! Thanks 👍

packages/eslint-plugin/src/rules/await-thenable.ts
ronami: Interesting! I didn't consider this edge case!

I took some time trying to figure this out, here are my thoughts:

- `checker.isArrayType()` doesn't flag this type as array, and I've only managed to compare this kind of value is with `checker.isArrayLikeType()` (which seems to check if the value is assignable to `Array<any>`).

  Getting the type of the element(s) of the array seems to be a bit trickier, and may be possible through TypeScript's internal [`getIterationTypesOfIterable`](https://github.com/microsoft/TypeScript/blob/2778e84ed88894156e2c5a17d74a25c73cc96ffe/src/compiler/checker.ts#L45725) or similar.
  
  Unless additional APIs are exposed, the best I've managed to come up with is using `checker.isArrayLikeType()` and getting the value of the array via `arrayType.getNumberIndexType()`. This seems to work OK, though a similar case that extends `Iterable` still has this issue ([playground link](https://deploy-preview-11267--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.tsx&code=JYOwLgpgTgZghgYwgAgLIE8CSkpwEYA2EAPAKogCuAzhACYA0yAKgHzIQAekItVy20fEWKtkAb2QBfAFC0ICAnCgoEAexBUwyDgC40WHEJIAFKKoC2wGsQBuq4LRaNKBAiwDc06QHpvyKgAWqhQEtMjKAA6qUFp4FFq0qhBUIADkYNKmFlYQAHRwrgAUHACUnkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AQwHd3K78ALRE3YAjJOiiJo0APbRI4MAF8QSoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):
  
    ```ts
    interface MyIterable<Unused, T> extends Iterable<T> { }
    declare const x: MyIterable<Promise<void>, null>;
    
    // should report but doesn't
    Promise.all(x);
    ```
- This issue seems to affect additional rules, I was able to create these reproducible ones (should I open issues for them? I'm not sure how often this case gets used in the wild):
  - `no-base-to-string` ([link to playground](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=PTAEDsHsCcFsEMA2iCepoFMAOMAuBnUAYxkyN1QCgATDIxeTYycfXUADwC5QBBaaPBQAeSACMAVnVwA%2BANyUABgBIA3hwC%2BihZRCh8AC0gBXRNXTY8oMcfbVIGfOADkuSgEtwuDNABm8IgxQAFkUfkERAFVwY3wMagAaUAAVGVAMDm9wakJwoWFU0FVQDQVaekYgklZ2RgiAeV96yWl8HlC8kRjkJPEpcnlKJTU6oUbm-oItBSA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oCMBDZRLXxdk%2BKkwDm6KImjQO0SODABfECqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):
  
    ```ts
    // normally reports correctly
    declare const x: Array<object>;
    `${x}`;
    
    // should report but doesn't
    interface MyArray<Unused, T> extends Array<T> { };
    declare const arrayOfObjects: MyArray<null, object>;
    
    `${arrayOfObjects}`;
    ```
  - `no-floating-promises` ([link to playground](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=PTAEDsHsCcFsEMA2iCepoFMAOMAuBnUAYxkyN1QCgATDIxeTYycfXUADwC5QBBaaPBQAeAArRIsAJb4MwgG6Qp1AHwqA3JQ6bKIUPgAWkAK6Jq6bHlAAjY%2B2qQM%2BcAHJclKeFwZoAM3hEGKAAsij8giIAquDGstQANKAAKiqgGBze4NSE4ULCKaAA3qAAvpq09IxBJKzsjBEA8r7ikjJOPKG5IjHIiS3SsgpKqhqUlPVCTf1t%2BJpAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDN4OBDfMwDmtYtA4BbSshTooiaBOiRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):
  
    ```ts
    // normally reports correctly
    declare const x: Array<Promise<void>>;
    x;
    
    // should report but doesn't
    interface MyArray<Unused, T> extends Array<T> { };
    declare const arrayOfPromises: MyArray<null, Promise<void>>;
    
    arrayOfPromises;
    ```
  - `prefer-reduce-type-parameter` ([link to playground](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=PTAEDsHsCcFsEMA2iCepoFMAOMAuBnUAYxkyN1QCgATDIxeTYycfXUADwC5QBBaaPBQAeNtACW4AOYA%2BSpQ4A6TNQCuRDAApKoUJvhEiq2ABoI8WBgCUoALwy9Abx27Qi9waOmXugNrgLDABdHlxoVQwTFwBfKyjdR2jQeEIAJToYalEwySkzACNISEQMeHAZKKsAbnkQUHwAC0hVRGp0bDxQfNV2akgMfHAAclxKSVwMaAAzAwxQAFkUfkERAFVwVXwMajMAFQcMDgnwakJloWF90EdQaJqaOgYmElZ2RhWAeSmAZRzpfB4i3OIg2yDMYlyMnu7yEX1%2BEn%2Bym26i0Ln0hmMZgClhs9icPjcHgx3lcoH8gRCoDCEXitziLkSyTSGWgWQh0gKRRKZQqlGqlCAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wls4AmsMolpFStYgENoUgLaJ8vdFF7QA9tEjgwAXxC6gA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):
  
    ```ts
    // normally reports correctly
    declare const x: Array<string>
    
    x.reduce(
      (accum, name) => ({
        ...accum,
        [name]: true,
      }),
      {} as Record<string, boolean>,
    );
    
    // should report but doesn't
    interface MyArray<Unused, T> extends Array<T> { };
    
    declare const arrayOfStrings: MyArray<null, string>;
    
    arrayOfStrings.reduce(
      (accum, name) => ({
        ...accum,
        [name]: true,
      }),
      {} as Record<string, boolean>,
    );
    ```
- I updated the PR with the changes described above, would love to hear your thoughts.

Edit: I think the `main` branch has failing tests which cause this PR to be red too.

kirkwaiblinger: Hey! I'm going to be away from a computer until beginning of August. I can look at it then, or feel free to move this along in the meantime 🙂

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -176,3 +209,73 @@ export default createRule<[], MessageId>({
...
+          return true;
+        }
+      }
+    }
+  }
+
+  return false;
+}
+
+function getValueTypesOfArrayLike(type: ts.Type, checker: ts.TypeChecker) {
```
kirkwaiblinger: IMO this would benefit from a return type

packages/eslint-plugin/tests/rules/await-thenable.test.ts
```diff
@@ -339,6 +339,218 @@ class C<R extends unknown> {
 }
       `,
     },
+
+    {
+      code: `
+declare const x: unknown;
+Promise.all(x);
```
kirkwaiblinger: This is a TS type error, so, not sure if we need the test. If we keep it let's add a comment that this is a type error, though.

packages/eslint-plugin/tests/rules/await-thenable.test.ts
```diff
@@ -339,6 +339,218 @@ class C<R extends unknown> {
...
+      code: `
+function f<T extends Promise<unknown>>(x: Array<T>) {
+  Promise.all(x);
+}
+      `,
+    },
+    {
+      code: `
+declare const x: Array<unknown>;
+Promise.all(x);
```
kirkwaiblinger, line 404: this maybe should be flagged? Debatable I suppose since you don't know `x` _isn't_ a promise. But usually, `unknown` isn't a promise, I'd say.

For example, we don't flag this with no-floating-promises:

```ts
declare function returnsUnknown(): unknown;

returnsUnknown();
```

since `unknown` doesn't "look like" a promise. 

Up to your judgement!

packages/eslint-plugin/tests/rules/await-thenable.test.ts
```diff
@@ -339,6 +339,218 @@ class C<R extends unknown> {
...
+    {
+      code: `
+declare const x: Array<any>;
+Promise.all(x);
+      `,
+    },
+    {
+      code: `
+declare const x: number | Array<Promise<number>>;
+Promise.all(x);
```
kirkwaiblinger, line 418: type error

packages/eslint-plugin/tests/rules/await-thenable.test.ts
```diff
@@ -339,6 +339,218 @@ class C<R extends unknown> {
...
+    },
+    {
+      code: `
+declare const x: [unknown, any];
+Promise.all(x);
+      `,
+    },
+    {
+      code: `
+declare const x: number | [Promise<number>];
```
kirkwaiblinger, line 462: type error

packages/eslint-plugin/tests/rules/await-thenable.test.ts
```diff
@@ -339,6 +339,218 @@ class C<R extends unknown> {
...
+    {
+      code: `
+declare const x: Iterable<any>;
+Promise.all(x);
+      `,
+    },
+    {
+      code: `
+declare const x: number | Iterable<Promise<number>>;
+Promise.all(x);
```
kirkwaiblinger, line 520: type error

packages/eslint-plugin/tests/rules/await-thenable.test.ts
```diff
@@ -339,6 +339,218 @@ class C<R extends unknown> {
...
+    {
+      code: `
+declare const x: Array<Promise<string>> | [Promise<string>, Promise<unknown>];
+Promise.all(x);
+      `,
+    },
+
+    {
+      code: `
+Promise.all();
```
kirkwaiblinger: type error

packages/eslint-plugin/tests/rules/await-thenable.test.ts
```diff
@@ -339,6 +339,218 @@ class C<R extends unknown> {
...
+      `,
+    },
+    {
+      code: `
+Promise.all(1);
```
kirkwaiblinger, line 554: type error

packages/eslint-plugin/tests/rules/await-thenable.test.ts
```diff
@@ -339,6 +339,218 @@ class C<R extends unknown> {
...
+      `,
+    },
+    {
+      code: `
+declare const x: Promise<number>;
+Promise.all(x);
```
kirkwaiblinger, line 562: type error

packages/eslint-plugin/tests/rules/await-thenable.test.ts
kirkwaiblinger, line 1: (testing)

I think let's add some test cases around generators as well? Like

```ts
function* generatorOfPromises() {
  yield Promise.resolve(1)
  yield Promise.resolve(2)
  yield Promise.resolve(3)
}

Promise.all(generatorOfPromises()); // should not report


function* generatorOfNumbers() {
  yield 1;
  yield 2;
  yield 3;
}

Promise.all(generatorOfNumbers()); // should report
```

and/or using the `Generator<...>` built-in type.

Also, `ReadonlyArray` and friends?


Also, array literals (most likely the primary way `Promise.all()` will be used:

```ts
Promise.all([Promise.resolve(1), 2, Promise.resolve(3)]);
```

Ideally this would be special-cased when the `[]` syntax is found and would report specifically on the 2 rather than on the whole array?

packages/eslint-plugin/src/rules/await-thenable.ts
kirkwaiblinger: I always appreciate how deeply you investigate things like this, and how clear the examples you come up with are 🙏 

> Interesting! I didn't consider this edge case!
> 
> I took some time trying to figure this out, here are my thoughts:
> 
> * `checker.isArrayType()` doesn't flag this type as array, and I've only managed to compare this kind of value is with `checker.isArrayLikeType()` (which seems to check if the value is assignable to `Array<any>`).
>   Getting the type of the element(s) of the array seems to be a bit trickier, and may be possible through TypeScript's internal [`getIterationTypesOfIterable`](https://github.com/microsoft/TypeScript/blob/2778e84ed88894156e2c5a17d74a25c73cc96ffe/src/compiler/checker.ts#L45725) or similar.
>   Unless additional APIs are exposed, the best I've managed to come up with is using `checker.isArrayLikeType()` and getting the value of the array via `arrayType.getNumberIndexType()`. This seems to work OK, though a similar case that extends `Iterable` still has this issue ([playground link](https://deploy-preview-11267--typescript-eslint.netlify.app/play/#ts=5.8.2&fileType=.tsx&code=JYOwLgpgTgZghgYwgAgLIE8CSkpwEYA2EAPAKogCuAzhACYA0yAKgHzIQAekItVy20fEWKtkAb2QBfAFC0ICAnCgoEAexBUwyDgC40WHEJIAFKKoC2wGsQBuq4LRaNKBAiwDc06QHpvyKgAWqhQEtMjKAA6qUFp4FFq0qhBUIADkYNKmFlYQAHRwrgAUHACUnkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AQwHd3K78ALRE3YAjJOiiJo0APbRI4MAF8QSoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):
>   ```ts
>   interface MyIterable<Unused, T> extends Iterable<T> { }
>   declare const x: MyIterable<Promise<void>, null>;
>   
>   // should report but doesn't
>   Promise.all(x);
>   ```
> * This issue seems to affect additional rules, I was able to create these reproducible ones (should I open issues for them? I'm not sure how often this case gets used in the wild):
>   
>   * `no-base-to-string` ([link to playground](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=PTAEDsHsCcFsEMA2iCepoFMAOMAuBnUAYxkyN1QCgATDIxeTYycfXUADwC5QBBaaPBQAeSACMAVnVwA%2BANyUABgBIA3hwC%2BihZRCh8AC0gBXRNXTY8oMcfbVIGfOADkuSgEtwuDNABm8IgxQAFkUfkERAFVwY3wMagAaUAAVGVAMDm9wakJwoWFU0FVQDQVaekYgklZ2RgiAeV96yWl8HlC8kRjkJPEpcnlKJTU6oUbm-oItBSA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oCMBDZRLXxdk%2BKkwDm6KImjQO0SODABfECqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):
>     ```ts
>     // normally reports correctly
>     declare const x: Array<object>;
>     `${x}`;
>     
>     // should report but doesn't
>     interface MyArray<Unused, T> extends Array<T> { };
>     declare const arrayOfObjects: MyArray<null, object>;
>     
>     `${arrayOfObjects}`;
>     ```

I actually think this behavior is arguably correct here. It's only with a true built-in array that the `.toString()` behavior is known to subsequently call its elements' `.toString()`s, whereas other "array"s/arraylikes may have different stringification. So in this case, we're reasoning based on implementation details of built-in arrays.

>   * `no-floating-promises` ([link to playground](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=PTAEDsHsCcFsEMA2iCepoFMAOMAuBnUAYxkyN1QCgATDIxeTYycfXUADwC5QBBaaPBQAeAArRIsAJb4MwgG6Qp1AHwqA3JQ6bKIUPgAWkAK6Jq6bHlAAjY%2B2qQM%2BcAHJclKeFwZoAM3hEGKAAsij8giIAquDGstQANKAAKiqgGBze4NSE4ULCKaAA3qAAvpq09IxBJKzsjBEA8r7ikjJOPKG5IjHIiS3SsgpKqhqUlPVCTf1t%2BJpAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1oDN4OBDfMwDmtYtA4BbSshTooiaBOiRwYAL4h1QA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):
>     ```ts
>     // normally reports correctly
>     declare const x: Array<Promise<void>>;
>     x;
>     
>     // should report but doesn't
>     interface MyArray<Unused, T> extends Array<T> { };
>     declare const arrayOfPromises: MyArray<null, Promise<void>>;
>     
>     arrayOfPromises;
>     ```
>   * `prefer-reduce-type-parameter` ([link to playground](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=PTAEDsHsCcFsEMA2iCepoFMAOMAuBnUAYxkyN1QCgATDIxeTYycfXUADwC5QBBaaPBQAeNtACW4AOYA%2BSpQ4A6TNQCuRDAApKoUJvhEiq2ABoI8WBgCUoALwy9Abx27Qi9waOmXugNrgLDABdHlxoVQwTFwBfKyjdR2jQeEIAJToYalEwySkzACNISEQMeHAZKKsAbnkQUHwAC0hVRGp0bDxQfNV2akgMfHAAclxKSVwMaAAzAwxQAFkUfkERAFVwVXwMajMAFQcMDgnwakJloWF90EdQaJqaOgYmElZ2RhWAeSmAZRzpfB4i3OIg2yDMYlyMnu7yEX1%2BEn%2Bym26i0Ln0hmMZgClhs9icPjcHgx3lcoH8gRCoDCEXitziLkSyTSGWgWQh0gKRRKZQqlGqlCAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wls4AmsMolpFStYgENoUgLaJ8vdFF7QA9tEjgwAXxC6gA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):
>     ```ts
>     // normally reports correctly
>     declare const x: Array<string>
>     
>     x.reduce(
>       (accum, name) => ({
>         ...accum,
>         [name]: true,
>       }),
>       {} as Record<string, boolean>,
>     );
>     
>     // should report but doesn't
>     interface MyArray<Unused, T> extends Array<T> { };
>     
>     declare const arrayOfStrings: MyArray<null, string>;
>     
>     arrayOfStrings.reduce(
>       (accum, name) => ({
>         ...accum,
>         [name]: true,
>       }),
>       {} as Record<string, boolean>,
>     );
>     ```

With these two... It's debatable IMO, and I could be convinced either way what the "correct" behavior is. We're linting for how these arraylikes quack, rather than the exact implementation details. Though (maybe?) an argument maybe could be made that the signature of `.reduce()` is an implementation detail rather than a general feature of arraylikes. 

In any case I don't think there's a clear cut wrong behavior in any of these that warrants proactively filing issues (versus reacting to user feedback if anyone encounters these edge cases in the wild).

> * I updated the PR with the changes described above, would love to hear your thoughts.

I think this current state looks great! I'd say this is a very good "best effort" approach at replicating the `getIterationTypesOfIterable` TS API (which really would be nice to have!), and we'll find out from user feedback if the remaining edge cases really are necessary to solve. Thanks for digging so deep into this.

> Edit: I think the `main` branch has failing tests which cause this PR to be red too.

Yep, should be fixed now! (after merging from main)

kirkwaiblinger (changes requested):
Couple of small things, mostly around testing, but this otherwise looks great! Oh and one larger question around special-casing array literals.

packages/eslint-plugin/tests/rules/await-thenable.test.ts
ronami: Yes! I think it may still be valuable having such a test to make sure the rule doesn't report _in addition_ to Typescript, making it extra noisy.

Do you think it's worth having these kind of tests? If we keep them, should I add a `// @ts-expect-error` above them (referencing https://github.com/typescript-eslint/typescript-eslint/issues/8298).

I think this is relevant for the other type errors (as they should all be under the `valid` cases to test the rule is not overly noisy).

packages/eslint-plugin/tests/rules/await-thenable.test.ts
ronami: Hmm... interesting!

My thinking is that either way can work, but that this should work consistently with the rest of the rule. Since the following is allowed by the rule (`await`ing an `unknown`), I think it should be OK to pass an `unknown` to a promise aggregator method ([playground link](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3mgXPArigNYpIDuKA3AFDVRoCeKY8AZoWBgJaptJIAKAJTwA3tXjwoZKFyxoaAX2pA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6AQwHd3K78ALRE3YAjJOiiJo0APbRI4MAF8QSoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)):

```typescript
declare const s: unknown;

async function foo() {
  await s;
}
```

kirkwaiblinger: Oh, yeah, that's a convincing argument 😀

packages/eslint-plugin/tests/rules/await-thenable.test.ts
kirkwaiblinger: I think that's a fine resolution to that, yeah 👍

packages/eslint-plugin/tests/rules/await-thenable.test.ts
ronami: Yes! I didn't consider generators and friends, thanks for catching this!

Regarding array literals - Initially I was bit hesitant to special case reporting these as this can turn a single report (`Promise.all(x)`) to multiple reports (e.g. `Promise.all([1, Promise.resolve(2), 3])`). Though I don't think this would have a realistic auto-fix or suggestion, and I agree that reporting exactly where the issue is a lot better (especially since this pattern seem to be a very common one).

I adjusted the PR to special case array literals and added the relevant tests, thanks for commenting on this 🚀

ronami:
Thanks @kirkwaiblinger! I'm really sorry for how long it took me to get back to this.

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -84,6 +91,53 @@ export default createRule<[], MessageId>({
...
 
+      CallExpression(node: TSESTree.CallExpression): void {
+        if (!isPromiseAggregatorMethod(context, services, node)) {
+          return;
+        }
+
+        const argument = node.arguments.at(0);
+
+        if (argument == null) {
+          return;
```
kirkwaiblinger, line 102: (cov nit)

This can be hit with 
```ts
// @ts-expect-error
Promise.all();
```

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -84,6 +91,53 @@ export default createRule<[], MessageId>({
...
+        }
+
+        if (argument.type === TSESTree.AST_NODE_TYPES.ArrayExpression) {
+          for (const element of argument.elements) {
+            if (element == null) {
+              continue;
```
kirkwaiblinger, line 108: (cov nit)

This can be hit with 
```ts
Promise.all([,]);
```

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -176,3 +230,76 @@ export default createRule<[], MessageId>({
...
+  }
+
+  if (checker.isArrayLikeType(type)) {
+    const indexType = type.getNumberIndexType();
+
+    if (indexType != null) {
+      return [indexType];
+    }
+
+    return null;
```
kirkwaiblinger: (cov nit)

I think this is probably truly unreachable - consider using `nullThrows()`? or leaving as is 🤷

packages/eslint-plugin/src/rules/await-thenable.ts
```diff
@@ -176,3 +230,76 @@ export default createRule<[], MessageId>({
...
+  }
+
+  // `Iterable<...>`
+  if (tsutils.isTypeReference(type)) {
+    return checker.getTypeArguments(type).slice(0, 1);
+  }
+
+  return null;
```
kirkwaiblinger, line 281: (cov nit) this might also be unreachable? 🤷 But if so `return null;` seems correct as-is here rather than a `nullThrows()` 👍

kirkwaiblinger (approved):
I had a look and a few of the missing cov lines are reachable for tests, but this is good with or without those tweaks 👍

packages/eslint-plugin/src/rules/await-thenable.ts
ronami: Oh, I completely missed codcov, thanks! I've updated the PR with your suggestions. I'm not sure about this one though, it's hard to tell if it's truly unreachable or if there's a case I've missed.

I think this is OK either way (though I think this would require throwing [like is done here](https://github.com/typescript-eslint/typescript-eslint/blob/ef9173c56fa488737bff9ab463338ea65cb83373/packages/eslint-plugin/src/rules/strict-boolean-expressions.ts#L934) instead of `nullThrows()`).

Zamiell:
Hey ronami! I've been wanting this bug to be fixed for many years now, so thanks for all your hard work on this!

PR #11272, opened 2025-06-05
fix(eslint-plugin): [prefer-optional-chain] ignore `check` option for most RHS of a chain

### LINKED ISSUE #11270: Bug: [prefer-optional-chain] should report case that can be converted to optional function call `?.()`

### Playground Link

https://typescript-eslint.io/play/#ts=5.6.3&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3gMyUgXPAN4BQ88AtiBgBZLCECuKo2AliiMPAD7wAU-AJTwAvAD54ANyRtgQkgF8SJXEgB0VWvXgAyXTjybqdYMIDcQA&eslintrc=N4AhCgwIgJwVwDYFMDOUBcJiTCKABAFwE8AHVAYxgEtTCBaVBagO0IHpSYkAzJGegHs61QSwCGCehQAW41hjwB3cTBZQcAX3CagA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDAruuGAL4g9A&tokens=false

### Repro Code

```TypeScript
declare const foo: {
  method: undefined | (() => void)
}

foo.method && foo.method();
```

### ESLint Config

```javascript
module.exports = {  
  "rules": {
    "@typescript-eslint/prefer-optional-chain": "warn"
  }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true
  }
}
```

### Expected Result

I expect that the rule should report and suggest `foo.method?.()`.

### Actual Result

No report

### Additional Info

This is specifically implied in [the docs examples](https://typescript-eslint.io/rules/prefer-optional-chain/#examples) with the line `foo && foo.a && foo.a.b && foo.a.b.method && foo.a.b.method();`

Relates to #9605

nayounsang:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11270 

## Overview

- Apply check option to LHSs except most RHS to fix the bug where void functions were not reported


packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
```diff
@@ -111,6 +111,11 @@ function isValidFalseBooleanCheckType(
   if (options.checkBigInt === true) {
     allowedFlags |= ts.TypeFlags.BigIntLike;
   }
+
+  if (options.checkVoid === true) {
+    allowedFlags |= ts.TypeFlags.Void;
+  }
+
   return types.every(t => isTypeFlagSet(t, allowedFlags));
 }
```
kirkwaiblinger: I'm wondering if this section should be flipped to only specify the flags that aren't checked instead of the ones that are?

For now, no tests fail if I switch the logic to the following, where we don't bother with explicit options for or handling of `void`:
```ts
  let flagsToExcludeFromCheck = 0;
  if (options.checkAny !== true) {
    flagsToExcludeFromCheck |= ts.TypeFlags.Any;
  }
  if (options.checkUnknown !== true) {
    flagsToExcludeFromCheck |= ts.TypeFlags.Unknown;
  }
  if (options.checkString !== true) {
    flagsToExcludeFromCheck |= ts.TypeFlags.StringLike;
  }
  if (options.checkNumber !== true) {
    flagsToExcludeFromCheck |= ts.TypeFlags.NumberLike;
  }
  if (options.checkBoolean !== true) {
    flagsToExcludeFromCheck |= ts.TypeFlags.BooleanLike;
  }
  if (options.checkBigInt !== true) {
    flagsToExcludeFromCheck |= ts.TypeFlags.BigIntLike;
  }

  return !types.some(t => isTypeFlagSet(t, flagsToExcludeFromCheck));
```

Can we come up with some test cases that would
- demonstrate which approach is preferable
- solidify that approach rather than passing with either choice of logic

I haven't thought deeply about which logic makes more sense, so it would be lovely if you have time to give this some thought and justify with examples which way is preferable. Thanks!

For reference, I ask this because: 
- [prefer-nullish-coalescing](https://typescript-eslint.io/rules/prefer-nullish-coalescing/#ignoreprimitives) does have the "only ignore specific types" approach, rather than "only check some hardcoded types plus allowed primitives" as in the existing implementation of this rule. 
- the answer will determine whether creation of a new option is warranted

kirkwaiblinger (commented):
This generally looks great! Just one big question on the implementation approach we want to take. Looking forward to hearing your thoughts!

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
nayounsang: I will think about it more as I work, but I will leave a quick comment first. The more immediate the communication, the better. Even if it's not accurate!
1. I would recommend adding `void` options. My philosophy is that the more open source users can customize with options, the more use cases users can cover. (Adding options would likely require some additional documentation changes beyond the current commit.)
2. Since the defaults are given as true, it seems more natural to change to the suggested code. Even better, it prevents optional chaining from being skipped on unexpected types.
3. I guess I need to consider side effects when `flagsToExcludeFromCheck = 0`. 
```ts
let allowedFlags = NULLISH_FLAGS | ts.TypeFlags.Object;
```
All tests will pass if `flagsToExcludeFromCheck = 0` for now and I don't think there's a problem at all.(Due to 2), but something feels off because `allowedFlags` has default value...

Perhaps the test case would be to determine what error/suggestion occurs depending on whether the option is turned off.

nayounsang: Yes, I looked into it further and there is no problem logically even if the default value is 0.
And, I found a tc that causes an lint error even though I set all check options to false.
```ts
      {
        code: `
declare const foo: { x: () => { y: string } } | null;
foo && foo.x();
        `,
        options: [
          {
            checkAny: false,
            checkBigInt: false,
            checkBoolean: false,
            checkNumber: false,
            checkString: false,
            checkUnknown: false,
            checkVoid: false,
          },
        ],
      },

// --------------------------------
      {
        code: `
declare const foo: { x: { y: string } } | null;
foo && foo.x;
        `,
        options: [
          {
            checkAny: false,
            checkBigInt: false,
            checkBoolean: false,
            checkNumber: false,
            checkString: false,
            checkUnknown: false,
            checkVoid: false,
          },
        ],
      },
```

```bash
AssertionError: Should have no errors but had 1: [
  {
    ruleId: '@rule-tester/prefer-optional-chain',
    severity: 2,
    message: "Prefer using an optional chain expression instead, as it's more concise and easier to read.",
    line: 3,
    column: 1,
    nodeType: null,
    messageId: 'preferOptionalChain',
    endLine: 3,
    endColumn: 15,
    suggestions: [ [Object] ]
  }
]
```
In summary, this is a case for nested Objs. So the commit only adds one of the two tc
related commit: https://github.com/typescript-eslint/typescript-eslint/pull/11272/commits/4b92ad4340a6128d29f50f9b92fc68baf645a794

There are probably a few more, but I'll give them credit for the discovery.
Is there anything else I should know & work? @kirkwaiblinger

kirkwaiblinger:
This PR has the effect that the following code is flagged (with `checkVoid` enabled)
```ts
declare const maybeVoid: void | { x: () => { some: 'object' } };
maybeVoid && maybeVoid.x();
```

I'm not sure that that's what we want? I'd say instead the goal is that a `void`-returning function call at the end of a chain whose return value isn't used should be reported regardless of the `check*` options.

In other words, as far as implementation - we're not looking to include possibly-`void` values in checked chains _except_ as the type of a function call at the end of the chain. It's not obvious to me that we need to create an option at all for this...

WDYT?

kirkwaiblinger:
I've realized — I think this is a much more general bug. The `check*` options should _not_ apply to the _right hand side_ member of a chain. Only the _left_ member. That is to say, 

```ts
declare const x: { a: string };

// checkString: true
const y = x && x.a; // correctly reports

// checkString: false
const y = x && x.a; // (BUG) doesn't report
```

So, if we generally fix the logic so that the type checking only applies to the LHS of a potential chain expression `x?.a`, that should also automatically fix the reported bug regarding `x?.a()` as a special case.

nayounsang:
Then, I guess we can check it except for right most. I will finish the work on void and reply.

nayounsang:
After thinking about it further with comments and usage examples, I think the checkVoid option is not necessary. checkXXX is useful when there is an accessible method. When considering real-world usage, I couldn't think of a case where checkXXX would be used for void.

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1680,6 +1680,65 @@ describe('hand-crafted cases', () => {
...
+        `,
+              },
+            ],
+          },
+        ],
+        options: [{ checkBoolean: false }],
+      },
+      // Exclude for everything else, an error occurs
+      {
+        code: noFormat`declare const foo: { x: { y: string } } | null; foo && foo.x;`,
```
kirkwaiblinger: This doesn't _test_ anything related to formatting, so there's no reason to use `noFormat` here. Let's remove it.

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1680,6 +1680,65 @@ describe('hand-crafted cases', () => {
...
+          declare const x: { y: boolean };
+          x?.y;
+        `,
+              },
+            ],
+          },
+        ],
+        options: [{ checkBoolean: false }],
+      },
+      // Exclude for everything else, an error occurs
```
kirkwaiblinger: ?

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1680,6 +1680,65 @@ describe('hand-crafted cases', () => {
...
+        code: `
+declare const foo: void;
+foo && foo();
+        `,
+        errors: [{ messageId: 'preferOptionalChain' }],
+        output: `
+declare const foo: void;
+foo?.();
+        `,
+      },
```
kirkwaiblinger: This should not report (see https://github.com/typescript-eslint/typescript-eslint/pull/11272#issuecomment-3043139036). Please revert.

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
```diff
@@ -92,26 +90,27 @@ function isValidFalseBooleanCheckType(
...
   }
-  if (options.checkBigInt === true) {
-    allowedFlags |= ts.TypeFlags.BigIntLike;
+  if (options.checkBigInt !== true) {
+    flagsToExcludeFromCheck |= ts.TypeFlags.BigIntLike;
   }
-  return types.every(t => isTypeFlagSet(t, allowedFlags));
+
+  return types.every(t => !isTypeFlagSet(t, flagsToExcludeFromCheck));
 }
```
kirkwaiblinger: Please revert these changes, since they turn out not to be related to the necessary changes.

nayounsang:
Oh, sorry. I misunderstood your review. Now I know what to do
I'll revert all of changes because, I changed so much and produce unnecessary codes.
Then, applying the logic to check except most RHS, I will focus on whether the issue case passes.
And if there is any problem, I will ask you. Thank you.

kirkwaiblinger:
Hey @nayounsang — is this ready for another review pass? If so, we [ask that you request review in the GUI](https://typescript-eslint.io/contributing/pull-requests#addressing-feedback-and-beyond) so we know to take another look. If not, take your time and don't mind me! Just looked like it might be ready for another pass. Thanks!

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1917,6 +1957,10 @@ describe('hand-crafted cases', () => {
         !x || x.a;
       `,
       "typeof globalThis !== 'undefined' && globalThis.Array();",
+      `
+        declare const x: void;
```
kirkwaiblinger: This is a TS error as written, which we prefer to avoid in test cases when possible.


```suggestion
        declare const x: void | (() => void);
        x && x();
```

kirkwaiblinger (approved):
Thanks!

kirkwaiblinger (approved):

PR #11292, opened 2025-06-10
fix(website): update hash value if problem occurs due to ts version error

### LINKED ISSUE #11281: Bug: playground just generally doesn't work

### Relevant Package

website

### Playground Link

_No response_

### Repro Code

```TypeScript
-
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

I'm able to use the playground

### Actual Result

Most of the features don't actually seem to work - I can interact with elements like checkboxes and tabs, type in code which gets red squiggles and highlighting, but most of the interaction doesn't actually do anything e.g. switching from "code" to "tsconfig" to "eslintrc" doesn't change the editor content (though it does cause a "Visual Editor" button to appear which does nothing when clicked); likewise for the "errors", "estree", etc.

If I switch to the "ESTree" tab, I do get an "ESQuery filter" input which immediately gives me this warning when I type anything:

```
Invalid Selector
Cannot read properties of undefined (reading 'parse')
```

Image

I have tried this on both latest Chrome and Edge, with extensions installed like ublock origins (which isn't reporting anything being blocked, and has been disabled for the site with no luck), with no extensions, and a few other combos, all with the same result.

There doesn't seem to be any errors in the console

Image

Here's another set of screenshots after I've clicked through some stuff and typed again into the editor:

Image

Image

### Additional Info

I'd like to have attached a screen recording, but don't have any software for that setup currently - I can look into that next week if needed.

Also let me know if there's any sections of template which you'd like filled out in some way - I skipped most of them as they didn't seem relevant and I couldn't find a more appropriate

### Versions

-

developer-bandi:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11281

## Overview

If an error occurs due to the ts version, hash's ts is modified to load the latest version.

### DIFF
--- a/packages/website/src/components/editor/useSandboxServices.ts
+++ b/packages/website/src/components/editor/useSandboxServices.ts
@@ -140,7 +140,6 @@ export const useSandboxServices = (
         });
       })
       .catch((err: unknown) => {
-        // props.setState({ ts: process.env.TS_VERSION });
         if (err instanceof Error) {
           if (
             err.message ===
--- a/packages/website/src/components/editor/useSandboxServices.ts
+++ b/packages/website/src/components/editor/useSandboxServices.ts
@@ -7,7 +7,7 @@ import semverSatisfies from 'semver/functions/satisfies';
 import type { createTypeScriptSandbox } from '../../vendor/sandbox';
 import type { CreateLinter } from '../linter/createLinter';
 import type { PlaygroundSystem } from '../linter/types';
-import type { RuleDetails } from '../types';
+import type { ConfigModel, RuleDetails } from '../types';
 import type { CommonEditorProps } from './types';
 
 import rootPackageJson from '../../../../../package.json';
@@ -23,6 +23,7 @@ export interface SandboxServicesProps {
     ruleDetails: RuleDetails[],
     tsVersions: readonly string[],
   ) => void;
+  readonly setState: (value: Partial<ConfigModel>) => void;
   readonly ts: string;
 }
 
@@ -139,6 +140,7 @@ export const useSandboxServices = (
         });
       })
       .catch((err: unknown) => {
+        props.setState({ ts: process.env.TS_VERSION });
         if (err instanceof Error) {
           setServices(err);
         } else {
--- a/packages/website/src/components/editor/useSandboxServices.ts
+++ b/packages/website/src/components/editor/useSandboxServices.ts
@@ -140,8 +140,14 @@ export const useSandboxServices = (
         });
       })
       .catch((err: unknown) => {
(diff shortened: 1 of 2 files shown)


packages/website/src/components/editor/useSandboxServices.ts
```diff
@@ -140,6 +141,12 @@ export const useSandboxServices = (
       })
       .catch((err: unknown) => {
         if (err instanceof Error) {
+          if (
+            err.message ===
+            'Could not get all the dependencies of sandbox set up!'
```
JoshuaKGoldberg: [Refactor] A couple issues with the implementation as-is:
* I had to do a find-all to see where this magic string came from. -1 from me on using magic: https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings
* I also think that continuing to use `catch` / `reject` as a form of control flow is not ideal here. Especially now that we know there's a real case to be handled with an "error" state. https://www.joshuakgoldberg.com/blog/try-catch-as-little-as-possible

But, stepping back, I think this is too wide of a check. _Any_ error rejection from `loadSandbox` will run `props.setState({ ts: process.env.TS_VERSION });` Even if it wasn't a TS version error in the first place!

Requesting changes:

* The reset of `ts` state should only happen if:
    a. The existing `ts` version is an unsupported one
    b. The failure is specifically for the `vs/editor/editor.main`
* Switch the return type of `loadSandbox` to be something like a discriminated union, so the failure state is captured in the type system

developer-bandi: I modified it in the above way, which seems to solve the problem you mentioned by checking if the user is trying to use a valid TypeScript version before loading the sandbox, and executing setState if not.

packages/website/src/components/editor/useSandboxServices.ts
```diff
@@ -44,107 +62,122 @@ export const useSandboxServices = (
...
-
-        setServices({
-          sandboxInstance,
-          system,
-          webLinter,
-        });
+      })
+      .then(() => {
+        sandboxSingleton(props.ts)
+          .then(async ({ lintUtils, main, sandboxFactory }) => {
```
JoshuaKGoldberg, line 73: [Refactor] The code already has an `async` inside a `.then(() => {`. Could you please refactor the new code so that it keeps to the `async`/`await` pattern where possible? That'll reduce the noise from this diff. Which is non-trivial and making it hard to review.

JoshuaKGoldberg:
Closing out as awaiting response for a while. Thanks for starting on this, though!

Issue #8315, opened 2024-01-30
Bug: [consistent-type-imports] False negatives when type name also refers to a local value


haskellcamargo:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.3.3&fileType=.tsx&code=PTAEAEBcGcFoFMAeAHeBjSCBOWD2XRZQAVAC3lAFtcATAVwBsKbd5oA7AcklCQEtokADSgARnR6RSA0DPa4eWeEwBuAQ3aQAdACg%2BlZPh4BvUAAUGatPFK4GNeAQC%2BoAGZ5KoTskvXb9x04Abh0dJEMsSQBPVFAAYVx2Vz4Ac1AAXlBjHVBQHysbOwcsAC5zX0KArBCnUJBQAEFQaCjKUTtQAHc%2BKVApCmg1Sgp2IYo1aD7yPpj4HVc6dgw%2BRPKC-2KACgBKLJzQJUg6LHZQdkYGGrqwAFVoPnY0-tAGXDQ1BmbW9s%2BJ0DVQOoGHQ5uEjG5FstVgBZNQPHZ7XKHY6nAA8FnWRUcoGAAD4rkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6MgeyeUuX0Ra0ipWpQC2xDtHyoMkRNGhTI4MAF8QaoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
// @ts-expect-error - The module doesn't exist, but this is not relevant.
import { Placeholder } from 'placeholder';

export type Config = {
  // Using it only as a type
  placeholder: Placeholder;
}

// A symbol with the same name as the type
function Placeholder() {
  return null;
}

[...]

PR #11322, opened 2025-06-20
fix(eslint-plugin): [no-unused-vars] no `is assigned a value but only used as a type` error when it has a same name in export statement

### LINKED ISSUE #10658: Bug: [no-unused-vars] no `is assigned a value but only used as a type` error when it has a same name type alias declaration exported

### Playground Link

https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=MYewdgzgLgBAgjAvDADAbgFAYKYA8AOIATrFAJ77bxIzmUgBm8aQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaWXWZRAE0IDcBDaVDSfAMxcgBpxtIABAFwCeABxQBjaAEth-QiniTc-APQFipclToN0URNGj5okbgF8QpoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
const A = 0;
export type A = typeof A;
```

### ESLint Config

```javascript
module.exports = {
  {
   "rules": {
    "no-unused-vars": "off",
    }
  }
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

Should report an error same as https://typescript-eslint.io/play/#ts=5.7.2&fileType=.ts&code=MYewdgzgLgBAgjAvDADAbgFAagTwA4Cm8SMuhIAZvGkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaWXWZRAE0IDcBDaVDSfAMxcgBpxtIABAFwCeABxQBjaAEth-QiniTc-APQFipclToN0URNGj5okbgF8QpoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

```ts
const A = 0;
type A = typeof A;
```

### Actual Result

No error reported

### Additional Info

Related: https://github.com/oxc-project/oxc/issues/8488#issuecomment-2589932781

nayounsang:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10658 

### Strict checks on exports with same type and variable name
- If the name is the same, the variable is `isTypeVariable && isValueVariable`.
- `ClassName`, `TSEnumName`, `TSModuleName`, `ImportBinding` and `arguments` also have same characteristics
- Therefore, only the cases that correspond to the issue are properly verified.
### Perform additional validating for abnormal cases
- `def.node.type !== AST_NODE_TYPES.TSTypeAliasDeclaration`
### Add test cases
- tc mentioned in issue and similar but valid


packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -415,6 +418,29 @@ function isSelfReference(
...
+ * In edge cases, the existing used logic does not work.
+ * When the type and variable name of the variable are the same
+ * @ref https://github.com/typescript-eslint/typescript-eslint/issues/10658
+ * @param variable the variable to check
+ * @returns true if it is an edge case
+ */
+function isSafeUnusedExportCondition(variable: ScopeVariable): boolean {
+  if (variable instanceof ESLintScopeVariable) {
+    return true;
+  }
```
nayounsang: ## need advice
I don't know why this condition is necessary. Can't the arg type of `isExported`, `isMergableExported` be used as `Variable`?
Actually, it is used as `Variable` only.

packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -465,7 +495,12 @@ function isExported(variable: ScopeVariable): boolean {
     }
 
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    return node.parent!.type.startsWith('Export');
+    const isExportedFlag = node.parent!.type.startsWith('Export');
```
nayounsang: ## need advice
Is there a reason why didn't use optional chaining before?

packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -427,6 +453,7 @@ const MERGABLE_TYPES = new Set([
  * @param variable the variable to check
  */
 function isMergableExported(variable: ScopeVariable): boolean {
```
nayounsang: fixable typo: `mergeable`

packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -15,6 +17,7 @@ import {
   ESLintUtils,
   TSESLint,
 } from '@typescript-eslint/utils';
+import { Node } from 'typescript';
```
nayounsang: oh, remove this....

packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -441,7 +467,9 @@ function isMergableExported(variable: ScopeVariable): boolean {
         def.node.parent?.type === AST_NODE_TYPES.ExportNamedDeclaration) ||
       def.node.parent?.type === AST_NODE_TYPES.ExportDefaultDeclaration
     ) {
-      return true;
+      return (
+        safeFlag || def.node.type !== AST_NODE_TYPES.TSTypeAliasDeclaration
```
JoshuaKGoldberg: [Performance] There's no need to run through `isSafeUnusedExportCondition(...)` until it's used:

```suggestion
        def.node.type !== AST_NODE_TYPES.TSTypeAliasDeclaration || isSafeUnusedExportCondition(variable)
```

packages/eslint-plugin/src/util/collectUnusedVariables.ts
JoshuaKGoldberg: tl;dr: the `!` is the right thing to use here. `node.parent` definitely always exists, so it'd be unnecessary to add extra runtime logic with a `?.`. The types are wrong! 

Other code using `?.` isn't ideal, but is existing cruft that we'll clean up eventually.

---

Digging in for fun...

To see what node type(s) don't know that the `parent` property definitely exists, you can hover or use a [twoslash query](https://marketplace.visualstudio.com/items?itemName=Orta.vscode-twoslash-queries) on:

```ts
node;
// ^? let node: TSESTree.CatchClause | TSESTree.ArrayPattern | TSESTree.ObjectPattern | TSESTree.Identifier | TSESTree.ClassDeclarationWithName | ... 192 more ... | TSESTree.YieldExpression

if (!node.parent) {
  node;
  // ^? let node: TSESTree.Program
}
```

`TSESTree.Program` is the AST root representing a file. Why is `node` allowed to be `TSESTree.Program`?

Its definition comes from:

https://github.com/typescript-eslint/typescript-eslint/blob/2f474b3f902606e151b01f31e05ae16899a6946c/packages/eslint-plugin/src/util/collectUnusedVariables.ts#L487

* `definition.node` is the `DefinitionBase` type parameter `Node extends ESTree.Node`
* `definition` itself comes from `variable.defs`, which is the `VariableBase` property `defs: Definition[]`

I think the root issue, then, is that `Definition` says its node can be _any_ `TSESTree.Node`, including `TSESTree.Program` - the only node type that doesn't have a `parent`. Which feels weird and wrong to me. What kind of scope-manager definition would come from a `TSESTree.Program`?

Filed: #11334. Once that's fixed then lint rules will force this code to remove unnecessary `!.`s and `?.`s.

packages/eslint-plugin/src/util/collectUnusedVariables.ts
JoshuaKGoldberg: > it is used as `Variable` only

👍 any reason not to change its type from `ScopeVariable` to `Variable`? I tried the three `: ScopeVariable` parameters to `: Variable` locally and everything looked happy.

nayounsang:
Its a different issue:https://github.com/typescript-eslint/typescript-eslint/issues/8315, but we need to figure out if this is a workable solution.

nayounsang:
> Its a different issue:#8315, but we need to figure out if this is a workable solution.

Ah, it is different. should change validate logic. 
I think I need to work on another PR. If this PR is headed in the right direction.

JoshuaKGoldberg:
> I think I need to work on another PR. If this PR is headed in the right direction.

@nayounsang apologies, I'm not following - are you suggesting we should or shouldn't review this PR?

nayounsang:
@JoshuaKGoldberg 
Oh, please review this PR. I'm ready.

The comments are my own monologue. I tend to take notes of everything and I just saw another issue that seemed related to this PR. After looking into it, it seems that even if this PR is resolved, there will be additional work needed to resolve the other issues.

packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -415,6 +417,29 @@ function isSelfReference(
...
+ * @returns true if it is an edge case
+ */
+function isSafeUnusedExportCondition(variable: ScopeVariable): boolean {
+  if (variable instanceof ESLintScopeVariable) {
+    return true;
+  }
+  if (variable.isTypeVariable && variable.isValueVariable) {
+    return !variable.defs
+      .map(d => d.type)
+      .every(t => exportExceptDefTypes.includes(t));
```
JoshuaKGoldberg: [Testing] If I comment out this `.every`, then all tests still pass. In fact, if the inside of the `if` just does `return false;` then tests still pass. So either this is unnecessary code or there's a gap in test coverage.

packages/eslint-plugin/src/util/collectUnusedVariables.ts
nayounsang: Yes, I learned more and your opinion is reasonable.
The number of `defs` is small, so the overhead of calling them unconditionally is greater than the complexity due to nested loops under certain conditions.
I committed.

packages/eslint-plugin/src/util/collectUnusedVariables.ts
nayounsang: I committed.

packages/eslint-plugin/src/util/collectUnusedVariables.ts
nayounsang: Good, thanks to you I was able to learn a lot.
I also wondered why there could be no node.parent, but of course it was an unavoidable part of the type declaration due to the root (Program).

packages/eslint-plugin/src/util/collectUnusedVariables.ts
nayounsang: The reason for this is because of https://github.com/typescript-eslint/typescript-eslint/pull/11322#discussion_r2162022682.
If type of `node.parent` start with `Export` and node is a `TSTypeAliasDeclaration`, then these syntaxes exist:
```
export type Foo = /* What */

declare namespace(module) Foo {
  export type A = / * What */
}
```
Now, let's see what's possible with `/* What */`.
This issue is about variable that `isTypeVariable && isValueVariable`
So, tests are passed even if commented code. 
But it's inconvenient. I think there's something I haven't anticipated yet. I need to dig deeper & I can refactor the code to make it clean.
I will test [these cases in play ground](https://typescript-eslint.io/play/#ts=5.7.2&fileType=.tsx&code=MYGwhgzhAECyCeBhcVoG9oF8BQBTAHgA4D2ATgC7TnyG7SLQC8cSKEA3NtgPTfQC0gocJE8%2BsMPABGuADRUAFrnjQwpOsGIA3XOoAm0KSuABXUuoB2lEMQDmAS2AA6LhbABbXBEJhgdAILo2NDQBCQU0JoWEJT4TNAAjJw4YWSU1LTQgcwZuMQAZlmc2PkmFsDk9sQW0ABCABQAlOhY2G6e3r50tUEhqRFRMdCEpMSE8UnYKURpVDTd8bkFdcW8AiIb-K4eXj5%2BWYFowaEzA9VDccyT0%2BHp8wfZc7TL-v7FpeWV1XUNzRg47V2XR%2BvROt0i50oIzGE2SeFOd0ytWRi3my2R7CAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaWXWZRAE0IDcBDaVDSRaafaSAGnG0gAEALgE8ADigDG0AJYiBhFPCm4BAegLFS5KnQbooLNhx4BfEMaA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

nayounsang: + More detail memo...
1. TSTypeAliasDeclaration & export

declare has been verified before, so we only test one (which also succeeds locally).
```
export type Foo = /* What */
```
2. isTypeVariable && isValueVariable

Merged declaration, `ClassName`, `TSEnumName`,`TSEnumMemberDefinition`, `TSModuleName`, `ImportBinding` and `arguments` satisfy this condition.
Most of these are impossible syntax with `export type A = /* ... */ `.

nayounsang: Just refactor dupe validate logic.
However, the question of whether this is enough has not yet been resolved. Please wait a few more days :)

nayounsang: After thinking about it, I found that the current logic is sufficient. I couldn't think of any test cases or examples that go against this logic.

packages/eslint-plugin/src/util/collectUnusedVariables.ts
```diff
@@ -415,6 +416,22 @@ function isSelfReference(
...
+function isMergedTypeDeclaration(
+  variable: Variable,
+  node: TSESTree.Node,
+): boolean {
+  return (
+    node.type === AST_NODE_TYPES.TSTypeAliasDeclaration &&
+    variable.isTypeVariable &&
+    variable.isValueVariable
+  );
+}
```
bradzacher, line 434: A better implementation to catch the interface case would leverage `variable.defs`:

```ts
function isMergedTypeDeclaration(
  variable: Variable,
): boolean {
  return (
    variable.defs.length > 1 &&
    variable.isValueVariable &&
    variable.isTypeVariable &&
    variable.defs.some(d => d.type === DefinitionType.Type)
  );
}
```

bradzacher (changes requested):
this is almost there -- but it misses this case:
```ts
const A = 0;
export interface A {}
```

packages/eslint-plugin/src/util/collectUnusedVariables.ts
nayounsang: It can't cover this case:
```ts
// valid, but report
export const Foo = 0;
export type Foo = typeof Foo;

export const Foo = 0;
export interface Foo {}
```
What I meant was to check for the `export type Foo = ...` statement. But in this case, it returns true if there are other type declarations besides the `export` statement. This is because `node` exists within the `export` statement when the `isMergedTypeDeclaration` is called.

So I've taken a different approach, can you please tell me if I'm going in the right direction?

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
```diff
@@ -3018,5 +3073,30 @@ declare class Bar {}
...
+      code: `
+const A = 0;
+type A = typeof A;
+export { A };
+      `,
+    },
+    {
+      code: `
+class A {}
+export type B = A;
```
nayounsang, line 3142: Now that I think about it, isn't this wrong? I can't remember why I added it to valid..... and currently it is treated as valid.

I'll try to fix it.

nayounsang: No, This case is diff from `only used as a type`. `class` is `isValueVariable` & `isTypeVariable`, and it is assigned anyway.
I think it would be better to make it a different issue if necessary...

JoshuaKGoldberg (approved):
LGTM, thanks! I'll defer to @bradzacher as well on review. 👍

nayounsang:
Hi. @bradzacher, There are some valid test cases I don't understand among the ones wrote before.
This is same with `case 3 -- just value exported -- expected type is reported` on your comment.
```
    `
interface Foo {
  bar: string;
}
export const Foo = 'bar';
    `,
    `
export const Foo = 'bar';
interface Foo {
  bar: string;
}
    `,
```
This is added from https://github.com/typescript-eslint/typescript-eslint/pull/6873
This is a PR that check all the declarations not just the first declaration.
At the time, the review seemed to have determined that the test case was not problematic. So I understood this to be an intended behavior.(In my personal opinion, a report is needed.) What do you think?
+cc @JoshuaKGoldberg

bradzacher:
@nayounsang the test cases may have been added in the past, but that doesn't mean they're right! They were right in so far as they allowed the rule to understand when a multi-declaration variable was exported in its second declaration, but the result is wrong.

```ts
interface Foo {
//        ^^^ Type Foo is unused
  bar: string;
}
export const Foo = 'bar';
```
```ts
export const Foo = 'bar';
interface Foo {
//        ^^^ Type Foo is unused
  bar: string;
}
```

This is what we should expect to report for these two cases -- we should be able to understand that the variable's second declaration is unused.

nayounsang:
@kirkwaiblinger I'm going to delegate this issue on to another contributor. 
It's a difficult issue, but I don't have the time to invest that much effort. I sincerely apologize for keeping maintainers waiting.

JoshuaKGoldberg:
No worries, thanks for getting started on this! It's totally reasonable to pass on issues you don't have time for. FWIW we've also been pretty busy in the maintenance team.

If anybody seeing this wants to continue this work / generally tackle #10658: please do! #10658 at time of writing is still accepting PRs. Just note that if you reuse code from this PR, please add a co-authored-by attribution to give nayounsang credit for the work they've done that's being used.

Cheers! 💙

PR #11323, opened 2025-06-20
feat(eslint-plugin-internal): [no-dynamic-tests] new internal Lint rule to ban dynamic syntax in generating tests

### LINKED ISSUE #11052: Repo: Lint rule to ban dynamic syntax in generating tests
### Suggestion

Following #9691 and its ~dozen linked test cleanup issues, we'd like to statically enforce that folks don't add more auto-generated / dynamic tests. I bet this could be done with a lint rule to ban:

* Putting function calls or other dynamic syntax inside `RuleTester` tests
* Placing `RuleTester` tests anywhere other than the root of a file (except maybe `if`s, for environment/version checks)

### Additional Info

I could see this eventually being upstreamed into `eslint-plugin-eslint-plugin`...

💖

nayounsang:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11052 

## Overview

no-dynamic-tests

Ban these:
```ts
{
  invalid: {...spread},
  valid: identifier,
  invalid: bi + nary,
  valid: callExp(),
  invalid: `${templateLiteralWithDynamic()}`,
  valid: obj.member,
  invalid: condition? foo : bar,
}
```
For test case object, It validate these keys recursively: `code`, `error`.
No checks are performed on other keys.

Applied the rules to the test file in `eslint-plugin` & `eslint-plugin-internal`. [Bulk suppression feature](https://eslint.org/blog/2025/04/introducing-bulk-suppressions/) was applied.
Create the script to update suppression.


packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
```diff
@@ -0,0 +1,94 @@
...
+        case AST_NODE_TYPES.Literal:
+        default:
+          return false;
+      }
+    }
+
+    return {
+      // Check RuleTester.run calls
+      CallExpression(node) {
+        if (isRuleTesterCall(node)) {
```
JoshuaKGoldberg: Nit: I think this comment is already pretty clear from the `isRuleTesterCall` call.

```suggestion
      CallExpression(node) {
        if (isRuleTesterCall(node)) {
```

packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
```diff
@@ -0,0 +1,94 @@
...
+          const testObject = node.arguments[2];
+          if (testObject.type === AST_NODE_TYPES.ObjectExpression) {
+            for (const prop of testObject.properties) {
+              if (
+                prop.type === AST_NODE_TYPES.Property &&
+                prop.key.type === AST_NODE_TYPES.Identifier &&
+                (prop.key.name === 'valid' || prop.key.name === 'invalid') && // Check each element in the array
```
JoshuaKGoldberg: Nit: similarly, this comment is only relevant a few lines down, above a `prop.value.elements.forEach` that reads pretty much as the same thing IMO.

```suggestion
                (prop.key.name === 'valid' || prop.key.name === 'invalid') &&
```

packages/eslint-plugin-internal/tests/rules/no-dynamic-tests.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      errors: [
+        {
+          column: 11,
+          line: 3,
+          messageId: 'noDynamicTests',
+        },
+      ],
+    },
+  ],
+  valid: [
```
JoshuaKGoldberg, line 239: [Testing] TaggedTemplateExpressions like `noFormat` should have test coverage to make sure that's allowed:

```ts
    {
      code: `
ruleTester.run('test', rule, {
  valid: [noFormat\`const x = 1;\`],
  invalid: [],
});
      `,
    },
```

JoshuaKGoldberg (changes requested):
LGTM, thanks! Only the one request for an added test is blocking IMO. Everything else is nitpicks that can be ignored if you don't like them. 🙂

packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
nayounsang: I applied :)

packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
nayounsang: I applied :)

packages/eslint-plugin-internal/tests/rules/no-dynamic-tests.test.ts
nayounsang: This is an important case. I added a test case and used fall-through to handle it as non-dynamic.

JoshuaKGoldberg (approved):
🔥 thanks!

packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
```diff
@@ -0,0 +1,94 @@
...
+              if (
+                prop.type === AST_NODE_TYPES.Property &&
+                prop.key.type === AST_NODE_TYPES.Identifier &&
+                (prop.key.name === 'valid' || prop.key.name === 'invalid') &&
+                prop.value.type === AST_NODE_TYPES.ArrayExpression
+              ) {
+                prop.value.elements.forEach(element => {
+                  if (element && isDynamicExpression(element)) {
+                    context.report({
+                      node: element,
```
kirkwaiblinger: Rather than reporting on the whole object expression, let's report on the specific offending node(s). The closer the report is to the issue, the the easier it is for a user to understand what's wrong and how to fix it.

For example, in `no-relative-paths-to-internal-packages`, we should report like this

```ts
    {
      code: "import packageJson from '../../package.json' with { type: 'json' };",
      filename: path.resolve(PACKAGES_DIR, 'ast-spec/vitest.config.mts'),
//              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    },
```
rather than the entire object

packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
```diff
@@ -0,0 +1,94 @@
...
+          );
+        case AST_NODE_TYPES.ObjectExpression:
+          return node.properties.some(prop => {
+            if (prop.type === AST_NODE_TYPES.SpreadElement) {
+              return true;
+            }
+            return isDynamicExpression(prop.value);
+          });
+        case AST_NODE_TYPES.Literal:
+        case AST_NODE_TYPES.TaggedTemplateExpression:
```
kirkwaiblinger, line 85: This should _only_ be `false` if the tag is the `noFormat` function. Otherwise, it should report. For example, all the usage of `wrap` in plugin-test-formatting.test.ts should report.

packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
```diff
@@ -0,0 +1,94 @@
...
+        default:
+          return false;
+      }
+    }
+
+    return {
+      CallExpression(node) {
+        if (isRuleTesterCall(node)) {
+          const testObject = node.arguments[2];
```
kirkwaiblinger, line 111: Let's handle the case that there isn't a `node.arguments[2]`... Obviously, valid code will always have three arguments, but this can crash the eslint server during development

nayounsang: > but this can crash the eslint server during development

I learned a lot from these details. I committed it.

packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
nayounsang: I committed!

packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
nayounsang: Nice review. I committed. I made it so that recursive functions report on elements rather than returning whether they are dynamic.

nayounsang:
# Change
As In the interest of the current issue and in my opinion, I enable dynamic object value.
```ts
// ex
{
  output: getOutput()
}
```

1. From the perspective of repo maintenance, this case is better. I think that writing values ​​by hand increases the inconvenience and does not significantly harm readability.
2. I don't think this change causes the following problems:
```
They tended to exponentially grow in size as more cases get grown
This is especially bad for slow typed tests
They make it hard to see what cases are or aren't being tested
They're not very statically analyzable
```
However, the **code** and **error** properties are excluded as an exception. These two properties often make analysis difficult when dynamically allocated.

+1
I forgot this case:
```
invalid: foo
```
So, I fix and test it.

packages/eslint-plugin-internal/src/rules/no-dynamic-tests.ts
```diff
@@ -0,0 +1,133 @@
...
+          node.properties.forEach(prop => {
+            if (prop.type === AST_NODE_TYPES.SpreadElement) {
+              context.report({
+                node: prop,
+                messageId: 'noDynamicTests',
+              });
+            } else {
+              // InvalidTestCase extends ValidTestCase
+              type TestCaseKey = keyof InvalidTestCase<string, []>;
+              const keyToValidate: TestCaseKey[] = ['code', 'errors'];
```
nayounsang, line 69: If do not handwrite these two keys, you will have problems with readability and analysis difficulty.
https://github.com/typescript-eslint/typescript-eslint/pull/11323#issuecomment-3052486935

eslint-suppressions.json
nayounsang, line 1: If this PR is merged, I'll mention bulk suprresion on issues about hand written test case.

nayounsang:
Probably if CI is run again, there will be no prob: Docusaurus can't find @typescript-eslint/eslint-plugin. This PR has nothing to do with this failure.

This is my first time using `bulk suppressions`, but it's amazing. **Let's try it out.**
However, this requires guidance and a process for disposal when issues about `hand written test cases` are resolved.
If only that could be maintained, management costs would be greatly reduced.

---------
This is perf comparison for `bulk suppressions` feature. There is no difference between the two. Because size of `eslint-suppressions.json` is small.
```bash
# with suppressions 
=== result ===
total time: 55.0sec
max CPU: 109.1%
avg CPU: 1.37%
max Memory: 1.0%
avg Memory: .30%
max RSS: 168256KB
avg RSS: 49736.64KB
=== result ===
total time: 52.0sec
max CPU: 90.4%
avg CPU: 1.44%
max Memory: 1.0%
avg Memory: .28%
max RSS: 167744KB
avg RSS: 46318.16KB

# without suppressions
=== result ===
total time: 57.0sec
max CPU: 114.0%
avg CPU: 1.50%
max Memory: 1.0%
avg Memory: .26%
max RSS: 167952KB
avg RSS: 43129.38KB
=== result ===
total time: 52.0sec
max CPU: 85.1%
avg CPU: 1.29%
max Memory: 1.0%
avg Memory: .31%
max RSS: 171712KB
avg RSS: 52057.46KB
```

Monitor with this sh:
```bash
#!/bin/bash

start=$(date +%s.%N)

# RUN
yarn nx lint eslint-plugin --skip-nx-cache > /dev/null 2>&1 &
pid=$!

# Variables
max_cpu=0; max_mem=0; max_rss=0
total_cpu=0; total_mem=0; total_rss=0
sample_count=0

# Monitor
while true; do

    if ! kill -0 $pid 2>/dev/null; then
        break
    fi

    ps_info=$(ps -p $pid -o %cpu,%mem,rss 2>/dev/null | tail -1)
    if [ ! -z "$ps_info" ] && [ "$ps_info" != "%CPU" ]; then
        cpu=$(echo "$ps_info" | awk '{print $1}')
        mem=$(echo "$ps_info" | awk '{print $2}')
        rss=$(echo "$ps_info" | awk '{print $3}')

        # isNumber?
        if [[ "$cpu" =~ ^[0-9]+\.?[0-9]*$ ]] && [[ "$mem" =~ ^[0-9]+\.?[0-9]*$ ]] && [[ "$rss" =~ ^[0-9]+$ ]]; then
            [ $(echo "$cpu > $max_cpu" | bc -l) -eq 1 ] && max_cpu=$cpu
            [ $(echo "$mem > $max_mem" | bc -l) -eq 1 ] && max_mem=$mem
            [ $(echo "$rss > $max_rss" | bc -l) -eq 1 ] && max_rss=$rss
            total_cpu=$(echo "$total_cpu + $cpu" | bc -l)
            total_mem=$(echo "$total_mem + $mem" | bc -l)
            total_rss=$(echo "$total_rss + $rss" | bc -l)
            sample_count=$((sample_count + 1))
        fi
    fi
    sleep 0.5
done

wait $pid
end=$(date +%s.%N)
time=$(echo "$end - $start" | bc -l)

if [ $sample_count -gt 0 ]; then
    avg_cpu=$(echo "scale=2; $total_cpu / $sample_count" | bc -l)
    avg_mem=$(echo "scale=2; $total_mem / $sample_count" | bc -l)
    avg_rss=$(echo "scale=2; $total_rss / $sample_count" | bc -l)
else
    avg_cpu=0
    avg_mem=0
    avg_rss=0
fi

echo ""
echo ""
echo "=== result ==="
echo "total time: ${time}sec"
echo "max CPU: ${max_cpu}%"
echo "avg CPU: ${avg_cpu}%"
echo "max Memory: ${max_mem}%"
echo "avg Memory: ${avg_mem}%"
echo "max RSS: ${max_rss}KB"
echo "avg RSS: ${avg_rss}KB"

```

package.json
```diff
@@ -37,12 +37,14 @@
...
     "lint-markdown": "markdownlint \"**/*.md\" --config=.markdownlint.json --ignore-path=.markdownlintignore",
     "lint-stylelint": "nx lint website stylelint",
+    "lint-prune-suppressions": "nx run-many -t lint --projects=eslint-plugin-internal,eslint-plugin --prune-suppressions",
     "lint": "nx run-many -t lint",
     "postinstall": "tsx tools/scripts/postinstall.mts",
     "pre-commit": "lint-staged",
     "release": "tsx tools/release/release.mts",
     "start": "nx run website:start",
     "test": "nx run-many -t test --exclude integration-tests website website-eslint",
+    "test-tmp": "vitest packages/eslint-plugin-internal/tests/rules/no-dynamic-tests.test.ts",
```
kirkwaiblinger: revert

nayounsang: I committed. Thank you for reviewing this weird script. 😂

package.json
```diff
@@ -77,7 +78,7 @@
     "console-fail-test": "^0.5.0",
     "cross-fetch": "^4.0.0",
     "cspell": "^9.0.0",
-    "eslint": "^9.26.0",
+    "eslint": "^9.30.1",
```
nayounsang: Bulk suppressions work even without updating packages. It supports after `9.24.0`.
https://eslint.org/blog/2025/04/eslint-v9.24.0-released/#bulk-suppressions

But, the bulk-suppressions feature has been updated after `9.26.0`.
https://eslint.org/blog/2025/05/eslint-v9.27.0-released/#sorted-eslint-suppressions.json
https://eslint.org/blog/2025/05/eslint-v9.28.0-released/#--pass-on-unpruned-suppressions-cli-option
https://eslint.org/blog/2025/06/eslint-v9.29.0-released/#features

eslint.config.mjs
```diff
@@ -416,6 +416,7 @@ export default tseslint.config(
     ],
     name: 'eslint-plugin-and-eslint-plugin-internal/test-files/rules',
```
nayounsang, line 429: Anyway, I think there is no reason not to apply it even though the issue doesn't mention `eslint-plugin-internal`. Testing the lint rules is the same.

JoshuaKGoldberg (dismissed):
LGTM. +1 to @kirkwaiblinger's suggestions on improving the reporting of the rule. I'll want to hear from Kirk before merging too.

package.json
kirkwaiblinger, line 1: Let's remove the package.json and yarn.lock changes; we're on eslint 9.35 now

kirkwaiblinger (changes requested):
(see comment)

package.json
nayounsang: I committed. I left the script of bulk suppression.

JoshuaKGoldberg (approved):
Looks great to me, thanks for jumping on this! 🙌 

I'll note that the rule isn't _completely_ covered by unit tests. But since it's internal & specific to this repo (for now?) I think that's fine.

PR #11327, opened 2025-06-21
fix(typescript-estree): infers singleRun as true for project service

### LINKED ISSUE #9854: Bug(typescript-estree): never infers singleRun as true for project service

### Relevant Package

typescript-estree

### Playground Link

_No response_

### Repro Code

```TypeScript
n/a
```

### ESLint Config

```javascript
import tseslint from "typescript-eslint";

export default tseslint.config(tseslint.configs.base, {
	files: ["**/*.ts"],
	languageOptions: {
		parserOptions: {
			projectService: true,
			tsconfigRootDir: import.meta.dirname,
		},
	},
	rules: {
		"@typescript-eslint/no-floating-promises": "error",
	},
});
```

### tsconfig

```jsonc
n/a
```

### Expected Result

Internally, `inferSingleRun` should return `true` when we're in single-run mode. Executing this config with something like `npx eslint .` should be determined to be in single-run mode.

### Actual Result

`singleRun` is always inferred to be `false` right now, because of:

https://github.com/typescript-eslint/typescript-eslint/blob/382e87b73644c138b3eade360a88ff2fff6af9c8/packages/typescript-estree/src/parseSettings/inferSingleRun.ts#L25-L33

### Additional Info

This doesn't generally have any impact on runtime performance right now. The project service doesn't generally use any single-run optimizations in single-run mode. But:

* It's confusing internally to see `singleRun` set to `false`
* There'll likely eventually be some optimizations, such as the caching TS system host mentioned in https://github.com/microsoft/TypeScript/issues/59338#issuecomment-2239701363 

---

We should also investigate whether the `options?.extraFileExtensions?.length` check can be bypassed with `options.projectService` per #9504:

https://github.com/typescript-eslint/typescript-eslint/blob/382e87b73644c138b3eade360a88ff2fff6af9c8/packages/typescript-estree/src/parseSettings/inferSingleRun.ts#L18-L23

### Versions

| `@typescript-eslint/typescript-estree` | `8.2.0` |

nayounsang:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9854 

## Overview

As `project` is null when `projectService` is true, we have to fix validate logic

### DIFF
--- a/packages/typescript-estree/src/parseSettings/inferSingleRun.ts
+++ b/packages/typescript-estree/src/parseSettings/inferSingleRun.ts
@@ -18,10 +18,7 @@ export function inferSingleRun(options: TSESTreeOptions | undefined): boolean {
   // https://github.com/typescript-eslint/typescript-eslint/issues/9504
   // There's no support (yet?) for extraFileExtensions in single-run hosts.
   // Only watch program hosts and project service can support that.
-  if (
-    options?.extraFileExtensions?.length &&
-    (!options.projectService || options.project)
-  ) {
+  if (options?.extraFileExtensions?.length && options.project) {
     return false;
   }
 


nayounsang:
Why tests was failed in [CI](https://cloud.nx.app/cipes/685692852f52e26f78bab110)?
In my local, there was no problem.

nayounsang:
Can't reproduce test errors in local development.
In test, `parseSettings.singleRun` from `packages/typescript-estree/src/parser.ts/parseAndGenerateServices` is `false`. 
So, it doesn't seem to be related to this commit.
I verified that the test results and (of course) the singleRun results did not change before and after the commit.
This work only affects `singleRun`.

packages/typescript-estree/src/parseSettings/inferSingleRun.ts
JoshuaKGoldberg, line 1: 😕 I also don't reproduce the failures locally. Even when running completely outside of Nx:

```shell
yarn
cd packages/typescript-estree
TYPESCRIPT_ESLINT_PROJECT_SERVICE=true yarn vitest tests/lib/parse.test.ts
```

This might be a question for @JamesHenry. Maybe some other cache shenanigans are in play?

JamesHenry: I finalized the general nx config follow ups PR and merged it into here, hopefully that helps, otherwise I'll look more specifically at this branch

JamesHenry:
@JoshuaKGoldberg @nayounsang Ok this one drove me a little insane working my way to a reproduction, but it actually has nothing to do with caching, it is all to do with whether `CI=true` is set or not (which is why it happens only on CI and not locally by default).

In other words, you should be able to repro with:

```sh
CI=true TYPESCRIPT_ESLINT_PROJECT_SERVICE=true nx test typescript-estree --skip-nx-cache
```

nayounsang:
Thanks for helping us find the cause of this weird bug. Now let's fix our logic to fix the CI test failure.

nayounsang:
Hi @JoshuaKGoldberg . Thank you for always helping me. I would like to ask you a question.
I want to test that singleRun works without any problems even when `extraFileExtensions` is set & `projectService` is truthy. But, I don't know how to write test scenarios and where to test them.

(I can see the processing `extraFileExtensions` logic in `packages/typescript-estree/src/useProgramFromProjectService.ts`
, but... I don't know if it's related to this issue.)

nayounsang:
When `singleRun` & `projectService` is true and given `extraFileExtensions` in `packages/typescript-estree/tests/lib/semanticInfo-singleRun.test.ts`, check the program is generated.
```ts
const resultProgram = parseAndGenerateServices(code, options).services
        .program;
expect(resultProgram).toStrictEqual(mockProgram);
```

Because, the code is to solve the bug that occurs when programInstance is not found in the singleRun environment.

packages/typescript-estree/tests/lib/semanticInfo-singleRun.test.ts
```diff
@@ -260,4 +260,17 @@ describe('semanticInfo - singleRun', () => {
       );
     },
   );
+
+  it('should create program when extraFileExtensions is provided in projectService and singleRun mode', () => {
+    vi.stubEnv('TSESTREE_SINGLE_RUN', 'true');
+
+    const resultProgram = parseAndGenerateServices(code, {
```
nayounsang: As asked, I'm not sure if testing like this is desirable.
`parseAndGenerateService -> getProgramAndAST (in singleRun, get AST and program) -> useProgramFromProjectService -> extraFileExtensions`

packages/typescript-estree/src/parseSettings/inferSingleRun.ts
```diff
@@ -18,13 +18,13 @@ export function inferSingleRun(options: TSESTreeOptions | undefined): boolean {
   // https://github.com/typescript-eslint/typescript-eslint/issues/9504
   // There's no support (yet?) for extraFileExtensions in single-run hosts.
   // Only watch program hosts and project service can support that.
-  if (options?.extraFileExtensions?.length) {
+  if (options?.extraFileExtensions?.length && !options.projectService) {
```
JoshuaKGoldberg: [Bug] I don't think this is right? If `options.projectService` is false but `options.project` is true`, single-run mode might still be active.

packages/typescript-estree/tests/lib/semanticInfo-singleRun.test.ts
JoshuaKGoldberg: Yeah I'd remove this test change.

JoshuaKGoldberg (changes requested):
The first `&&` added seems off to me, but the other looks good. Am I reading it right?

packages/typescript-estree/src/parseSettings/inferSingleRun.ts
nayounsang: I committed.
I looked into the related issues and studied them further, and it was something I should have carefully considered.

packages/typescript-estree/tests/lib/semanticInfo-singleRun.test.ts
nayounsang: I committed

packages/typescript-estree/tests/lib/inferSingleRun.test.ts
```diff
@@ -143,6 +143,17 @@ describe(inferSingleRun, () => {
     expect(actual).toBe(false);
   });
 
+  it('returns true when singleRun can be inferred options.extraFileExtensions contains entires with projectService', () => {
```
JoshuaKGoldberg:
```suggestion
  it('returns true when options.extraFileExtensions contains entries and projectService is true', () => {
```

packages/typescript-estree/src/parseSettings/inferSingleRun.ts
```diff
@@ -18,13 +18,16 @@ export function inferSingleRun(options: TSESTreeOptions | undefined): boolean {
   // https://github.com/typescript-eslint/typescript-eslint/issues/9504
   // There's no support (yet?) for extraFileExtensions in single-run hosts.
   // Only watch program hosts and project service can support that.
-  if (options?.extraFileExtensions?.length) {
+  if (
+    options?.extraFileExtensions?.length &&
+    (!options.projectService || options.project)
```
JoshuaKGoldberg: [Bug] Why would `options.projectService` be negated (`!`), but not `options.project`?

JoshuaKGoldberg (changes requested):

packages/typescript-estree/src/parseSettings/inferSingleRun.ts
nayounsang: According to the comments, in that case `projectService` must be truthy to infer singlerun when `extraFileExtensions` is given. I'm not sure about support for project options.
+It seems like it would be better to remove the conditions for `options.project`. What do you think?

JoshuaKGoldberg: #11333 will make this easier. Once that's merged, we can assume that only 0-1 of `project` or `projectService` will be enabled -- not both.

I _think_ then it will be roughly:

```ts
  if (options?.extraFileExtensions?.length && options.project) {
    return false;
  }
```

nayounsang: Good. I've committed.

packages/typescript-estree/tests/lib/inferSingleRun.test.ts
nayounsang: I committed

bradzacher (approved):

Issue #11319, opened 2025-06-18
Bug: Parser should error when both `projectService` and `project` are set


MariaSolOs:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.
- [x] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Abug) and found none that matched my issue.
- [x] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Relevant Package

parser

### Repro Code

```TypeScript
N/A. This issue is about a confusing configuration behavior.
```

### ESLint Config

```javascript
// base-config.js
module.exports = {
  plugins: [ "@typescript-eslint" ],
  parser: "@typescript-eslint/parser",
  parserOptions: { projectService: true }
}

// .eslintrc.js
module.exports = {
  extends: "./base-config.js",
  parserOptions: { project: "./tsconfig.eslint.json" }
}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

The docs aren't clear of what the parser will use when both `projectService` and `project` are set. This can be particularly confusing when extending a configuration that sets `parserOptions` which are later combined into a single merged object. I would expect for the parser to throw an explicit error in such misconfiguration scenarios, or have a well-documented explanation of what happens when both settings are provided.

### Actual Result

`eslint --print-config` shows that the merged configuration has `{ project: "./tsconfig.eslint.json", projectService: true }`, and it's not clear how the TypeScript program will be discovered.

### Versions

| package                                | version |
| -------------------------------------- | ------- |
| `@typescript-eslint/eslint-plugin`     | `^8.31.0` |
| `@typescript-eslint/parser`            | `^8.31.0` |
| `TypeScript`                           | `5.8.3` |
| `ESLint`                               | `^8.57.0` |
| `node`                                 | `22.14.0` |

JoshuaKGoldberg:
This is a good point. For most misconfiguration items I would normally point to #6403 and say we should address that holistically. But this _"both `project` and `projectService`"_ case has come up a few times. IMO it'd be better to get a quicker, tailored fix now and not wait for #6403's breaking change some major version in the future.

cc @typescript-eslint/triage-team - WDYT?

JoshuaKGoldberg:
[...]

PR #11333, opened 2025-06-23
fix(parser): error when both `projectService` and `project` are set

MariaSolOs:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/11319

## Overview

As mentioned in the issue, the docs aren't clear of what the parser will use when both `projectService` and `project` are set. When such case is detected, I think it's better to be explicit about the misconfiguration and error as early as possible.


packages/typescript-estree/src/parseSettings/createParseSettings.ts
```diff
@@ -145,6 +145,12 @@ export function createParseSettings(
     tsconfigRootDir,
   };
 
+  if (parseSettings.projectService && tsestreeOptions.project) {
+    throw new Error(
+      'Cannot use both projectService and project parser options.',
```
JoshuaKGoldberg: [Docs] I think users who are enabling both are probably not super familiar with them. We'll need to guide them a bit more. Maybe something like...

```suggestion
      'Enabling 'project' does nothing when 'projectService' is already enabled. You can remove 'project'.',
```

Note that we can't assume `project: true` rather than string / string[]. Example from the wild, https://sourcegraph.com/github.com/software-mansion/smelter/-/blob/ts/eslint.config.js?L28-35

packages/typescript-estree/src/parseSettings/createParseSettings.ts
JoshuaKGoldberg, line 1: [Testing] Looks like our own test infrastructure is getting in the way. Both are set because that was easier to do in CI:

https://github.com/typescript-eslint/typescript-eslint/blob/702cea862db34b980e580f46314141c24024ee47/.github/workflows/ci.yml#L256-L260

https://github.com/typescript-eslint/typescript-eslint/blob/4ad8a0713376e5b88c56997312861a2c8cb049ab/packages/typescript-estree/src/parseSettings/createParseSettings.ts#L116-L125 

Maybe we should temporarily give users a way to disable the new error? `TYPESCRIPT_ESLINT_IGNORE_PROJECT_AND_PROJECT_SERVICE_ERROR` or some similarly unwieldy name? I'll post in #11319.

MariaSolOs:
@JoshuaKGoldberg Thanks for the review! I'll wait for further thoughts from other maintainers before I continue working on this :)

packages/typescript-estree/src/parseSettings/createParseSettings.ts
JoshuaKGoldberg: Yes, this :)

packages/typescript-estree/src/parseSettings/createParseSettings.ts
MariaSolOs: > Note that we can't assume `project: true` rather than string / string[]. Example from the wild, https://sourcegraph.com/github.com/software-mansion/smelter/-/blob/ts/eslint.config.js?L28-35

We still want to throw if `project` isn't a boolean though (?).

packages/typescript-estree/src/parseSettings/createParseSettings.ts
MariaSolOs: Done!

.github/workflows/ci.yml
```diff
@@ -19,6 +19,7 @@ env:
   NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
   # This increases the verbosity of the logs for everything, including Nx Cloud, but will hopefully surface more info about recent lint failures
   NX_VERBOSE_LOGGING: false
+  TYPESCRIPT_ESLINT_IGNORE_PROJECT_AND_PROJECT_SERVICE_ERROR: true
```
MariaSolOs: I tried setting this variable [here](https://github.com/typescript-eslint/typescript-eslint/blob/63d37d4a6ac58bc98605f1faef0fc15e29026b08/.github/workflows/ci.yml#L270) but it was undefined when running the tests on CI 🤷🏻‍♀️

packages/typescript-estree/src/parseSettings/createParseSettings.ts
JoshuaKGoldberg: I don't follow, what do you mean? `project: boolean | string | string[]` is always valid on its own.

MariaSolOs: I just don’t understand what you wanted me to do when you pointed this out:

> Note that we can't assume `project: true` rather than string / string[]

The current if condition handles all cases, unless I’m missing something?

MariaSolOs: Or are you saying that when project is a string or a string array, then it isn’t replaceable by setting `projectService: true` only?

JoshuaKGoldberg: Oh gotcha sorry, I'm being unclear 🙈. I think you can disregard my comments. I just wanted to point out that the type of `project` isn't just `boolean`.

In general, devs should _never_ set both `project` and `projectService`. Regardless of types of either, you should only ever set at most one. If both are set to any truthy value, that should be a config error.

Does that answer what you're looing for?

MariaSolOs: No worries, all clear now!

.github/workflows/ci.yml
MariaSolOs: The problem of setting it here is that then I'm not able to write a test that hits the new error codepath, and so that's why the coverage check is screaming at me.

JoshuaKGoldberg: Yeah this gets tricky. Instead of going through CI envs, I'd stub this out with `vi.stubEnv` the way `TYPESCRIPT_ESLINT_PROJECT_SERVICE` is:

https://github.com/typescript-eslint/typescript-eslint/blob/270fa594102856ca9cc06ca897819c3c196c1314/packages/typescript-estree/tests/lib/createParseSettings.test.ts#L18

https://github.com/typescript-eslint/typescript-eslint/blob/270fa594102856ca9cc06ca897819c3c196c1314/packages/typescript-estree/tests/lib/createParseSettings.test.ts#L28

packages/typescript-estree/src/parseSettings/createParseSettings.ts
```diff
@@ -180,6 +180,16 @@ export function createParseSettings(
     tsconfigRootDir,
   };
 
+  if (
```
JoshuaKGoldberg, line 185: [Docs] Just to make it clear in the code that this ad-hoc validation isn't the long-term strategy...

```suggestion
  // TODO: Eventually, parse settings will be validated more thoroughly.
  // https://github.com/typescript-eslint/typescript-eslint/issues/6403
  if (
```

JoshuaKGoldberg (changes requested):
The source code change looks great. Just requesting adding tests: you should be able to use the existing `vi.stubEnv`-powered `TYPESCRIPT_ESLINT_PROJECT_SERVICE` ones around `projectService` in `createParseSettings.test.ts` as reference. LMK if that actually doesn't work though!

MariaSolOs:
@JoshuaKGoldberg I'm having trouble fixing these tests as I can't reproduce the errors locally. I've been running `yarn nx test eslint-plugin-internal` from the root of the repo, is that not the right command?

JoshuaKGoldberg:
Ah, yes, these come from explicitly setting `TYPESCRIPT_ESLINT_PROJECT_SERVICE`:

https://github.com/typescript-eslint/typescript-eslint/blob/eb4ecd8c156133265eb13dd4228521df74ba9b7b/.github/workflows/ci.yml#L244-L245

https://github.com/typescript-eslint/typescript-eslint/blob/eb4ecd8c156133265eb13dd4228521df74ba9b7b/.github/workflows/ci.yml#L267-L271

I normally run Vitest directly when only a handful of files are failing:

```shell
TYPESCRIPT_ESLINT_PROJECT_SERVICE=true yarn vitest packages/eslint-plugin-internal/tests/rules/plugin-test-formatting.test.ts
```

MariaSolOs:
@JoshuaKGoldberg got it, thank you for the hint!

packages/eslint-plugin-internal/tests/rules/no-poorly-typed-ts-props.test.ts
```diff
@@ -8,6 +8,7 @@ const ruleTester = new RuleTester({
   languageOptions: {
     parserOptions: {
       project: './tsconfig.json',
+      projectService: false,
```
MariaSolOs, line 11: Not sure if this is the right way to fix the tests in files that just use the `RuleTester`.

packages/typescript-estree/tests/lib/semanticInfo.test.ts
```diff
@@ -43,6 +43,11 @@ describe('semanticInfo', async () => {
     vi.stubEnv('TSESTREE_SINGLE_RUN', '');
   });
 
+  vi.stubEnv(
```
MariaSolOs, line 50: I don't know why this isn't enough to disable the test failures 🫠

packages/eslint-plugin-internal/tests/rules/no-poorly-typed-ts-props.test.ts
JoshuaKGoldberg: Hmm, good question. Ideally we wouldn't have to do anything in the plugin tests themselves. I'm poking around...

packages/typescript-estree/tests/lib/semanticInfo.test.ts
JoshuaKGoldberg: For 7 of the 8 failures, because it's outside the `beforeEach` - the env is getting reset each test. But then also the `const testCases = ...` _also_ needs the env variable stubbed. So this has to happen both times. 36ad58e8d

JoshuaKGoldberg:
Ok, I think I have a handle on things and am most of the way through normalizing test files to handle the change. I have to step away, but in the meantime filed #11676. Expect more soon. Fun stuff! 🫠

MariaSolOs:
> Ok, I think I have a handle on things and am most of the way through normalizing test files to handle the change. I have to step away, but in the meantime filed #11676. Expect more soon. Fun stuff! 🫠

Thanks a lot for the help here! <3

packages/eslint-plugin-internal/tests/rules/no-poorly-typed-ts-props.test.ts
JoshuaKGoldberg: OK, yes, this is the right way.

Recapping for posterity: it's not apparent from the naming _(sigh)_ but the `TYPESCRIPT_ESLINT_PROJECT_SERVICE` env variable's purpose isn't purely to _enable_ `projectService`. It's to _switch_ `project` to `projectService` if `project` is enabled. But it doesn't apply when `project: false` is present.

packages/eslint-plugin/tests/RuleTester.ts
```diff
@@ -2,9 +2,40 @@ import type {
...
+  }
+  // Otherwise, use the project service for types if requested in the env
+  else if (process.env.TYPESCRIPT_ESLINT_PROJECT_SERVICE) {
+    parserOptions.projectService = true;
+  }
+  // Finally, default to project: true as the standard (legacy) behavior
+  // See: https://github.com/typescript-eslint/typescript-eslint/issues/11676
+  else {
+    parserOptions.project = true;
+  }
```
JoshuaKGoldberg, line 28: Linking #11676 here so we don't forget: if & when we remove `eslint-plugin` from the `TYPESCRIPT_ESLINT_PROJECT_SERVICE` matrix,  we could switch the default to `projectService: true`...

JoshuaKGoldberg (commented):
This was a lot more work than I was expecting it to be 🙃. Sorry for "surprise!" gifting you this big of a task @MariaSolOs!

Summarizing what's going on:

* Previously, if `process.env.TYPESCRIPT_ESLINT_PROJECT_SERVICE` was enabled, rule tests with `project` set were being switched to `projectService`
  * This is why some of them set `projectService: false` -- to prevent that switching
  * See [logic in `createParseSettings.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/c2c6ca168e05e9ebd1e27d4ae00e94064a83ecca/packages/typescript-estree/src/parseSettings/createParseSettings.ts#L151-L155)
* With this PR's intended change, that use case errors intentionally
* I switched typed rule tests from manually calling `new RuleTester` to using a helper `createRuleTesterWithTypes` that:
  * If it receives an enabled `project`, sets `projectService: false` like before
  * Enables `projectService` if `process.env.TYPESCRIPT_ESLINT_PROJECT_SERVICE`
  * Otherwise enables `project`

MariaSolOs:
> Sorry for "surprise!" gifting you this big of a task @MariaSolOs!

No worries at all! At least I'm relieved that something non-trivial was going on and it wasn't just me having skill issues 😂 

>   * See [logic in `createParseSettings.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/c2c6ca168e05e9ebd1e27d4ae00e94064a83ecca/packages/typescript-estree/src/parseSettings/createParseSettings.ts#L151-L155)

Oh my, that `tsestreeOptions.project && tsestreeOptions.projectService !== false && process.env.TYPESCRIPT_ESLINT_PROJECT_SERVICE === 'true'` condition does feel a bit hacky. Besides the tests in this repo, do you think there are projects out in the wild with this setup?

> * I switched typed rule tests from manually calling `new RuleTester` to using a helper `createRuleTesterWithTypes` that:
>   
>   * If it receives an enabled `project`, sets `projectService: false` like before
>   * Enables `projectService` if `process.env.TYPESCRIPT_ESLINT_PROJECT_SERVICE`
>   * Otherwise enables `project`

You're awesome, thank you for all the work here! 🦸

JoshuaKGoldberg:
> Besides the tests in this repo, do you think there are projects out in the wild with this setup?

I hope not... https://sourcegraph.com/search?q=context:global+TYPESCRIPT_ESLINT_PROJECT_SERVICE+-file:.*node_modules.*+-repo:%5Egithub%5C.com/typescript-eslint/typescript-eslint%24+&patternType=keyword&sm=0 shows none. I suppose we'll never know for sure though. 🤷

JoshuaKGoldberg:
The ESLint lint complaint is external.

MariaSolOs:
@JoshuaKGoldberg Thank you for all the guidance and helping me get this merged! <3

JoshuaKGoldberg:
It was a pleasure working with you @MariaSolOs, thanks for sticking with us on it! I hope you keep contributing 😁

jkowalleck:
wow, this thing causes an error - which breaks build.
so this is clearly a breaking change.

Please roll it back and issue a warning, instead of an error.

JoshuaKGoldberg:
> clearly a breaking change

We don't see this as a breaking change. We see this as a bug fix. As in, from our perspective, your config was incorrect and invalid to begin with. This change fixed the parser to correctly error when it should have before.

If you see differently, please follow [our contributing guide](https://typescript-eslint.io/contributing) and file an issue with a reproduction. Thanks!

GKersten:
I understand the fix is not seen as a breaking change, but actually it is breaking my build currently when updating my packages. Using `@ngrx/eslint-plugin`. It took my way too much time to find out what was causing this error, since the config was hidden in that plugin and I had no direct indication where to look for the failing config. 

Wanted to leave this comment here for anyone running into the same issue.
"Fixed" it for now by disabling ngrx-eslint rules. See https://github.com/ngrx/platform/issues/4993

If possible, would be great if the error message can be improved to better indicate what config is causing the error.
Parsing error: Enabling project does nothing when projectService is enabled. You can remove the project setting.

JoshuaKGoldberg:
Ah, we don't always check comments on merged PRs, I'd missed the latest batch of comments here. #11869 also noted that this change is more "breaking"-feeling than the typical bugfix. It's under `evaluating community engagement` now.

> If possible, would be great if the error message can be improved to better indicate what config is causing the error.

I like this! I don't know how to do it, though. ESLint configs are pretty restrictive in what information they give to parsers. By the time it reaches `@typescript-eslint/parser` -> `@typescript-eslint/typescript-estree`, it's "just" a JavaScript object. If anybody knows of a good way to get this to work, please do file an issue (or post here, or in our Discord) - that'd be great.

PR #11355, opened 2025-06-28
fix(typescript-estree): If the template literal is tagged and the text has an invalid escape, `cooked` will be `null`

### LINKED ISSUE #11353: Bug: Wrong cooked value of `TemplateElement` in `TaggedTemplateExpression `

### Relevant Package

typescript-estree

### Playground Link

https://typescript-eslint.io/play/#ts=5.8.2&showAST=es&fileType=.tsx&code=MoFwTglgdg5gdGAhgdwAYB0CuANX3VA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
String.raw`\uXXXX`
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

`TemplateElement.value.cooked` should be `null`.

### Actual Result

`TemplateElement.value.cooked` is "\uXXXX".

### Additional Info

_No response_

### Versions

| `@typescript-eslint/typescript-estree` | `8.35.0` |

nayounsang:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11353 

## Overview

- Add a flag `isInTaggedTemplate` indicating whether template literal is tagged
- Function to validate escape
- Add test case

### DIFF
--- a/packages/typescript-estree/src/convert.ts
+++ b/packages/typescript-estree/src/convert.ts
@@ -90,10 +90,10 @@ function isEntityNameExpression(
 }
 
 export class Converter {
+  #isInTaggedTemplate = false;
   private allowPattern = false;
   private readonly ast: ts.SourceFile;
   private readonly esTreeNodeToTSNodeMap = new WeakMap();
-  private isInTaggedTemplate = false;
   private readonly options: ConverterOptions;
   private readonly tsNodeToESTreeNodeMap = new WeakMap();
 
@@ -400,6 +400,38 @@ export class Converter {
         }
       }
     }
+  }
+
+  #isValidEscape(arg: string): boolean {
+    const unicode = /\\u([0-9a-fA-F]{4})/g;
+    const unicodeBracket = /\\u\{([0-9a-fA-F]+)\}/g; // supports ES6+
+    const hex = /\\x([0-9a-fA-F]{2})/g;
+    const validShort = /\\[nrtbfv0\\'"]/g;
+
+    const allEscapes =
+      /\\(u\{[^}]*\}|u[0-9a-fA-F]{0,4}|x[0-9a-fA-F]{0,2}|[^ux])/g;
+
+    let match: RegExpExecArray | null;
+    while ((match = allEscapes.exec(arg)) != null) {
+      const escape = match[0];
+
+      if (
+        unicode.test(escape) ||
+        (unicodeBracket.test(escape) &&
+          (() => {
+            const cp = parseInt(escape.match(unicodeBracket)![1], 16);
+            return cp <= 0x10ffff;
+          })()) ||
+        hex.test(escape) ||
+        validShort.test(escape)
+      ) {
+        continue;
+      }
+
+      return false;
+    }
+
+    return true;
   }
 
   #throwError(node: number | ts.Node, message: string): asserts node is never {
@@ -1890,7 +1922,10 @@ export class Converter {
               type: AST_NODE_TYPES.TemplateElement,
               tail: true,
               value: {
-                cooked: node.text,
+                cooked:
+                  this.#isValidEscape(node.text) && this.#isInTaggedTemplate
+                    ? node.text
+                    : null,
                 raw: this.ast.text.slice(
                   node.getStart(this.ast) + 1,
                   node.end - 1,
@@ -1919,7 +1954,7 @@ export class Converter {
       }
 
       case SyntaxKind.TaggedTemplateExpression: {
-        this.isInTaggedTemplate = true;
+        this.#isInTaggedTemplate = true;
         const result = this.createNode<TSESTree.TaggedTemplateExpression>(
           node,
           {
@@ -1934,7 +1969,7 @@ export class Converter {
               ),
           },
         );
-        this.isInTaggedTemplate = false;
+        this.#isInTaggedTemplate = false;
         return result;
       }
       case SyntaxKind.TemplateHead:
@@ -1945,7 +1980,10 @@ export class Converter {
           type: AST_NODE_TYPES.TemplateElement,
           tail,
           value: {
-            cooked: node.text,
+            cooked:
+              this.#isValidEscape(node.text) && this.#isInTaggedTemplate
+                ? node.text
+                : null,
             raw: this.ast.text.slice(
               node.getStart(this.ast) + 1,
               node.end - (tail ? 1 : 2),
--- a/packages/typescript-estree/src/convert.ts
+++ b/packages/typescript-estree/src/convert.ts
@@ -93,6 +93,7 @@ export class Converter {
   private allowPattern = false;
   private readonly ast: ts.SourceFile;
   private readonly esTreeNodeToTSNodeMap = new WeakMap();
+  private isInTaggedTemplate = false;
   private readonly options: ConverterOptions;
   private readonly tsNodeToESTreeNodeMap = new WeakMap();
 
@@ -1917,19 +1918,25 @@ export class Converter {
         return result;
       }
 
-      case SyntaxKind.TaggedTemplateExpression:
-        return this.createNode<TSESTree.TaggedTemplateExpression>(node, {
-          type: AST_NODE_TYPES.TaggedTemplateExpression,
-          quasi: this.convertChild(node.template),
-          tag: this.convertChild(node.tag),
-          typeArguments:
-            node.typeArguments &&
-            this.convertTypeArgumentsToTypeParameterInstantiation(
-              node.typeArguments,
-              node,
-            ),
-        });
-
+      case SyntaxKind.TaggedTemplateExpression: {
+        this.isInTaggedTemplate = true;
+        const result = this.createNode<TSESTree.TaggedTemplateExpression>(
+          node,
+          {
+            type: AST_NODE_TYPES.TaggedTemplateExpression,
+            quasi: this.convertChild(node.template),
+            tag: this.convertChild(node.tag),
+            typeArguments:
+              node.typeArguments &&
+              this.convertTypeArgumentsToTypeParameterInstantiation(
+                node.typeArguments,
+                node,
+              ),
+          },
+        );
+        this.isInTaggedTemplate = false;
+        return result;
+      }
       case SyntaxKind.TemplateHead:
       case SyntaxKind.TemplateMiddle:
       case SyntaxKind.TemplateTail: {
--- a/packages/typescript-estree/tests/lib/parse.test.ts
+++ b/packages/typescript-estree/tests/lib/parse.test.ts
@@ -926,4 +926,58 @@ describe(parser.parseAndGenerateServices, () => {
       });
     },
   );
+
+  describe('template literal cooked values', () => {
+    const getTemplateElement = (
+      code: string,
+    ): parser.TSESTree.TemplateElement | null => {
+      const result = parser.parse(code, {
+        comment: true,
+        loc: true,
+        range: true,
+        tokens: true,
+      });
+
+      const taggedTemplate = result.body.find(
+        b => b.type === parser.AST_NODE_TYPES.ExpressionStatement,
+      );
+      const expression = taggedTemplate?.expression;
+      if (expression?.type !== parser.AST_NODE_TYPES.TaggedTemplateExpression) {
+        return null;
+      }
+      return expression.quasi.quasis[0];
+    };
+
+    it('should set cooked to null for invalid escape sequences in tagged template literals', () => {
+      const code = 'String.raw`\\uXXXX`';
+      const templateElement = getTemplateElement(code);
+
+      expect(templateElement?.value.cooked).toBeNull();
+      expect(templateElement?.value.raw).toBe('\\uXXXX');
+    });
+
+    it('should set cooked to null for other invalid escape sequences', () => {
+      const code = 'String.raw`\\unicode and \\u{55}`';
+      const templateElement = getTemplateElement(code);
+
+      expect(templateElement?.value.cooked).toBeNull();
+      expect(templateElement?.value.raw).toBe('\\unicode and \\u{55}');
+    });
+
+    it('should set cooked to parsed value for valid escape sequences', () => {
+      const code = 'String.raw`\\n\\t\\u0041`';
+      const templateElement = getTemplateElement(code);
+
+      expect(templateElement?.value.cooked).toBe('\n\tA');
+      expect(templateElement?.value.raw).toBe('\\n\\t\\u0041');
+    });
+
+    it('should handle mixed valid and invalid escape sequences', () => {
+      const code = 'String.raw`\\n\\uXXXX\\t`';
+      const templateElement = getTemplateElement(code);
+
+      expect(templateElement?.value.cooked).toBeNull();
+      expect(templateElement?.value.raw).toBe('\\n\\uXXXX\\t');
+    });
+  });
 });


nayounsang:
It's strange. In my local, I cleared the cache and the changed types were built in the types package. Isn't the CI environment a new virtual machine that runs anyway?

-------------------------------

`typescript-estree` has a dependency on `types`. `TemplateElement` comes from `types`
`TemplateElement` in `types` is auto-generated by `ast-spec`.
In nx cloud, `types` is hitting cache in CI.
So, changes are not updated and an error occurs.
Do I need to do something separate?

nayounsang:
hi @JamesHenry
Thank you for your efforts on this difficult problem.
- Can I continue working on this PR?
- Is it okay if I merge with the main branch and resolve conflicts freely?
- If CI-related work is not completed, can I help?

JamesHenry:
@nayounsang Yes, sorry I forgot about this one, I obviously kept iterating on your other one to discover that one wasn't a caching issue. This one might be related to your other one or it might be totally separate.

Please revert my changes when you fix up the conflicts and let's get back to a clean slate. If it's still failing unexpectedly after that I can take another look

JamesHenry:
@nayounsang I see you're doing a lot of pushes that result in failed tasks. Are you running these things locally before pushing? Ideally you would iterate locally to get the high priority tasks passing and then push up to CI to verify in a neutral environment

nayounsang:
@JamesHenry Sorry, I will push after resolving all failures that occurred in CI next time.

When I work on the command for the project locally, my laptop has a performance issue that makes it not work for a long time. So I tried to check the work result in CI. Still, I was in a hurry.

packages/parser/tsconfig.build.json
```diff
@@ -8,9 +8,6 @@
     {
       "path": "../typescript-estree/tsconfig.build.json"
     },
-    {
-      "path": "../types/tsconfig.build.json"
-    },
```
nayounsang, line 13: Because of this message by nx,
```
NX   The workspace is out of sync

[@nx/js:typescript-sync]: Some TypeScript configuration files are missing project references to the projects they depend on or contain outdated project references.

? Would you like to sync the identified changes to get your workspace up to date? … 
Yes, sync the changes and run the tasks
No, run the tasks without syncing the changes
```
There are a lot of changes like this due to sync, but even after reading the documentation, I'm not sure if this is correct.
I reverted all dependency related changes.

packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts
```diff
@@ -282,7 +282,7 @@ export default createRule<Options, MessageIds>({
         return;
       }
 
-      const text = literal.quasis[0].value.cooked;
+      const text = literal.quasis[0].value.cooked ?? '';
```
nayounsang: TODO: I don't think this is the best way to handle it, but I need to dig into it a bit more.

packages/eslint-plugin/src/rules/no-unsafe-assignment.ts
```diff
@@ -172,6 +172,10 @@ export default createRule({
       senderType: ts.Type,
       senderNode: ts.Node,
     ): boolean {
+      function isNull(value: unknown): value is null {
```
nayounsang: I had to write a separate type guard function because of lint. Is there a better way?

nayounsang:
What is this err on CI?
https://cloud.nx.app/runs/Ol9VgADs38/task/eslint-plugin%3Atypecheck
```
> tsc --build --emitDeclarationOnly

vitest.config.mts(4,34): error TS6305: Output file '/home/runner/work/typescript-eslint/typescript-eslint/dist/vitest.config.base.d.mts' has not been built from source file '/home/runner/work/typescript-eslint/typescript-eslint/vitest.config.base.mts'.
> tsc --build --emitDeclarationOnly

vitest.config.mts(4,34): error TS6305: Output file '/home/runner/work/typescript-eslint/typescript-eslint/dist/vitest.config.base.d.mts' has not been built from source file '/home/runner/work/typescript-eslint/typescript-eslint/vitest.config.base.mts'.
Warning: command "tsc --build --emitDeclarationOnly" exited with non-zero status code
```

This is my first time seeing this and I don't know how to fix it in local as content of err msg and success in the local.

packages/eslint-plugin-internal/src/rules/plugin-test-formatting.ts
```diff
@@ -448,9 +452,13 @@ export default createRule<Options, MessageIds>({
     }
 
     function checkForUnnecesaryNoFormat(
-      text: string,
+      text: string | null,
       expr: TSESTree.TaggedTemplateExpression,
     ): void {
+      if (text == null) {
```
nayounsang, line 466: This change affects three rules.
Would it be better to simply ignore it? How should I handle it?
Once this is decided, I will also add tests.

packages/typescript-estree/src/convert.ts
```diff
@@ -401,6 +402,78 @@ export class Converter {
...
+      '$',
+    ];
+
+    for (let index = 0; index < text.length; index++) {
+      const char = text[index];
+      if (char !== '\\') {
+        continue;
+      }
+
+      const nextChar = text[index + 1];
```
nayounsang: Don't care about if nextChar doesn't exist because of [typescript error & parsing error](https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=MYewdgzgLgBAhjAvDAylATgSzAcwHTpwDuABgDolA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false).

packages/parser/tsconfig.build.json
JoshuaKGoldberg: cc @JamesHenry - I believe `main` is now in a healthy state, and any tsconfig changes can be reverted?

JamesHenry: Yeah should be

packages/typescript-estree/src/convert.ts
```diff
@@ -90,6 +90,7 @@ function isEntityNameExpression(
 }
 
 export class Converter {
+  #isInTaggedTemplate = false;
```
JoshuaKGoldberg: [Bug] Tagged templates can have arbitrary nodes as children. This is totally valid:

```ts
`
${new (class {
    value = `
  
  ${(() => `C:\D`)()}}
  `
  })().value
  }
`;
```

https://typescript-eslint.io/play#ts=5.8.2&showAST=ts&fileType=.tsx&code=AYKAJA3gdgpg7gAgBQGMA2BDAzlhEQKEIBuGaArjAgLwKiEEKRJICUNAfHQMIBcAOgBFgrNgF8xjegjGjWAOlIUYjScADcQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

I don't think any kind of state on `Converter` is going to scale. Instead, I think checking `node.parent` might be the only way to do it?

packages/typescript-estree/src/convert.ts
```diff
@@ -401,6 +402,78 @@ export class Converter {
     }
   }
 
+  #isValidEscape(text: string): boolean {
```
JoshuaKGoldberg, line 95: 😬 this and `#isInTaggedTemplate` are a lot of logic to add. If it does stay in, I think this should at least be extracted to its own separate function to keep `Converter` from getting even more complex. Finding a package on npm that already does this would be even better.

But, can we get away with checking if `template.text === template.rawText`?

JoshuaKGoldberg (changes requested):
A good start!

packages/typescript-estree/src/convert.ts
nayounsang: `node.parent` is more safe & simple. I committed.

packages/typescript-estree/src/convert.ts
nayounsang: I think `template.text === template.rawText` is difficult to solve this issue.
As `node.text`(node: ts.NoSubstitutionTemplateLiteral | ts.TemplateHead | ts.TemplateMiddle | ts.TemplateTail) is handle its escape char and , this condition is difficult when valid and invalid escape chrs are mixed.

So, find package is better. I found [unraw](https://www.npmjs.com/package/unraw) package that parse escape chr and throws err when it is invalid. (I wanted to compare different packages, but I couldn't find any other suitable ones.)

I committed.

nayounsang:
CI is failed again.. I guess more work is needed

packages/typescript-estree/src/convert.ts
fisker: Maybe we can eval `node.rawText` it directly?

```
try {
  new Function(`return \`${node.rawText}\`;`)();
  return true;
} catch {
  return false;
}
```

I see they are stored in `.templateSpans`

nayounsang: Ah, that's better. Thank you. For me, it's a much lighter and simpler way than adding dependencies, so I committed it. However, I know that eval has performance and security issues. Will this be okay? @JoshuaKGoldberg 

--------
p.s I tried to use `ts.createSourceFile`, but it fails in test cases.
```ts
ts.createSourceFile(
   'temp.ts',
   `const str = \`${text}\`;`,
    ts.ScriptTarget.Latest,
);
```

```bash
convert > tagged template literal cooked > should set cooked to null for invalid escape sequences in tagged template literals
AssertionError: expected '\n\uXXXXᄑ\t' to be null

- Expected: 
null

+ Received: 
"
\\uXXXXᄑ       "

convert > tagged template literal cooked > should set cooked to null for mixed valid and invalid escape sequences
AssertionError: expected '\uXXXX\xQW' to be null

- Expected: 
null

+ Received: 
"\\uXXXX\\xQW"

```

packages/types/package.json
```diff
@@ -67,6 +67,9 @@
       "build": {
         "dependsOn": [
           "copy-ast-spec"
+        ],
+        "inputs": [
+          "{projectRoot}/src/generated/**/*"
```
nayounsang, line 75: Since there is no `ast-spec` dependency in `types`, need to set `inputs` to update the cache hash.
+I think also enhance `copy-ast-spec` cache.
```json
      "copy-ast-spec": {
        "cache":true,
        "command": "tsx tools/copy-ast-spec.mts",
        "dependsOn": [
          "ast-spec:build"
        ],
        "options": {
          "cwd": "{projectRoot}"
        },
        "outputs": [
          "{projectRoot}/src/generated"
        ]
       "inputs":[
         "file of types..."
       ]
      },
```

By the way, should I create a separate issue for this task?

packages/typescript-estree/tests/lib/convert.test.ts
```diff
@@ -449,4 +449,77 @@ describe('convert', () => {
...
+      }
+      const init = variableDeclarator.init;
+      if (init?.type !== AST_NODE_TYPES.TemplateLiteral) {
+        throw new Error('TemplateLiteral not found');
+      }
+      const templateElement = init.quasis[0];
+
+      expect(templateElement.value.cooked).toBe(`\\uXXXX`);
+    });
+  });
```
fisker, line 546: We should add some tests for template with expressions.

nayounsang: Ah, thanks to you I found and fixed the bug.

packages/typescript-estree/src/convert.ts
JoshuaKGoldberg: I think I should defer to @bradzacher on this. I'm quite weary of taking on `eval` or `new Function`. Even if we assume there are no security concerns, both are quite slow. I think we'd need to see benchmarks showing they don't significantly change parsing times for very large files that include many large template literals and edge cases.

fisker: If the `new Function` is unacceptable, except `unraw` package, `@babel/helper-string-parser` can be considered to use too.

nayounsang: I'm having a hard time measuring benchmarks and frankly, it's hard to come up with meaningful results. 
If it's important, I'd keep spending time on this, but I think it's better to spend time analyzing the behavior of `unraw` and `@babel/helper-string-parser`.
Because I also have concerns about `eval`.

JoshuaKGoldberg: > it's hard to come up with meaningful results

FWIW the lack of a measurable difference _is_ a meaningful result! If you can show that doing this many times doesn't slow down anything noticeably, that's great.

But, yes, I think `eval()` is probably a non-starter.

nayounsang: I used unraw. Here's why:
1. As `@babel/helper-string-parser` in internal package, the document does not exist. I think it's important to make sure that the code is easy to understand later.
2. `@babel/helper-string-parser` is difficult to use in a TypeScript environment because there's no `d.ts` file. It's currently in v8 beta version. 
3. There are cases where the test produces false negatives when use `@babel/helper-string-parser`.

packages/eslint-plugin-internal/tests/rules/plugin-test-formatting.test.ts
```diff
@@ -886,5 +886,14 @@ const test = [
   errors: [],
 }));
     `,
+    // cooked is null for invalid escape sequences in tagged template literals. ignore it.
+    `
+   ruleTester.run({
```
bradzacher:
```suggestion
ruleTester.run({
```

packages/typescript-estree/src/convert.ts
bradzacher: Could we approximate the check with something like:
```ts
if (/\\[xu]/.test(template.text) && template.text === template.rawText) {
  // invalid
} else {
  // valid
}
```

This is a pretty rare case -- it's an edge-case in the spec.
And TBH I don't think we should bother being "perfectly correct" at the cost of perf / extra 3rd-party dependencies.

For reference -- the parser has worked this way **_since its creation 10 years ago_** and nobody has ever complained.

So I think "fast and close" is more than enough for us given we aren't doing the parsing ourselves.

fisker: IMO, correctness is top priority for a "parser", if performance is important, why don't we always set it to `null`? User can't relay on it anyway. Or consider use a getter?

fisker: Maybe we can ask ESTree to remove `.cooked` completely? Not much value in this property.

fisker: Raised an issue in ESTree https://github.com/estree/estree/issues/333

bradzacher: > IMO, correctness is top priority for a "parser

I'd agree -- but we're not a parser! TypeScript is the parser -- we're just an intermediary on top of it.
IMO a little bit of leeway is okay because we're not a parser that's able to validate tokens inline as it lexes the code string -- having to back-track to validate with relatively complex regexes isn't really worth it unless there's a good motivating case.

Because it's an edge case that clearly few people use/care about I think that a bit of leeway to trade-off perf with "mostly correct" behaviour is fine, esp when the alternatives are:
- the current and wrong behaviour
- always using `null` which is also incorrect and could lead people to believe all escapes are invalid always
- adding an expensive regex to catch a rare-to-likely-never case

nayounsang: I committed as suggested by @bradzacher.
- estree/estree#333 is closed.
- It has no external dependencies and does not eval.
- Above all, it's a decent level of work. The logic is simple.

packages/eslint-plugin/src/rules/no-unsafe-assignment.ts
```diff
@@ -202,7 +202,11 @@ export default createRule({
           receiverProperty.key.type === AST_NODE_TYPES.TemplateLiteral &&
           receiverProperty.key.quasis.length === 1
         ) {
-          key = receiverProperty.key.quasis[0].value.cooked;
+          const cooked = receiverProperty.key.quasis[0].value.cooked;
+          if (cooked == null) {
```
kirkwaiblinger: This is an impossible condition (and the test provided doesn't reach it).

This branch refers to keys whose node type is a `TemplateLiteral`... Which necessarily means they are untagged. If it were tagged it would have the `TaggedTemplateExpression`. 

Suggestion:
Use `nullThrows()` instead

packages/eslint-plugin/src/rules/no-duplicate-enum-values.ts
```diff
@@ -62,7 +62,9 @@ export default createRule({
           } else if (isNumberLiteral(member.initializer)) {
             value = member.initializer.value;
           } else if (isStaticTemplateLiteral(member.initializer)) {
-            value = member.initializer.quasis[0].value.cooked;
+            // cooked can only be null inside a TaggedTemplateExpression, which is not possible.
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            value = member.initializer.quasis[0].value.cooked!;
```
kirkwaiblinger: Suggestion: Use `nullThrows()` here instead

packages/typescript-estree/src/convert.ts
```diff
@@ -1938,32 +1955,43 @@ export class Converter {
...
-          quasi: this.convertChild(node.template),
-          tag: this.convertChild(node.tag),
-          typeArguments:
-            node.typeArguments &&
-            this.convertTypeArgumentsToTypeParameterInstantiation(
-              node.typeArguments,
-              node,
-            ),
-        });
+        const result = this.createNode<TSESTree.TaggedTemplateExpression>(
```
kirkwaiblinger: Revert?

JoshuaKGoldberg:
👋 @nayounsang just checking in, is this still something you have time & energy for?

nayounsang:
> 👋 @nayounsang just checking in, is this still something you have time & energy for?

I'm so sorry, I forgot.
I've been busy lately. It's still difficult to work quickly, but I can work on it over the weekend.
If I need to respond quickly, please let other contributors do the work.

packages/eslint-plugin/src/rules/no-duplicate-enum-values.ts
```diff
@@ -62,7 +62,10 @@ export default createRule({
           } else if (isNumberLiteral(member.initializer)) {
             value = member.initializer.value;
           } else if (isStaticTemplateLiteral(member.initializer)) {
-            value = member.initializer.quasis[0].value.cooked;
+            value = nullThrows(
+              member.initializer.quasis[0].value.cooked,
+              'impossible condition',
```
kirkwaiblinger:
```suggestion
              'cooked can only be null inside a TaggedTemplateExpression, which is not possible here',
```

packages/eslint-plugin/src/rules/no-unsafe-assignment.ts
```diff
@@ -202,7 +202,11 @@ export default createRule({
           receiverProperty.key.type === AST_NODE_TYPES.TemplateLiteral &&
           receiverProperty.key.quasis.length === 1
         ) {
-          key = receiverProperty.key.quasis[0].value.cooked;
+          const cooked = nullThrows(
+            receiverProperty.key.quasis[0].value.cooked,
+            'impossible condition',
```
kirkwaiblinger:
```suggestion
            'cooked can only be null inside a TaggedTemplateExpression, which is not possible here',
```

kirkwaiblinger (dismissed):

kirkwaiblinger:
@nayounsang I think this looks good but unfortunately it has some merge conflicts now. Could you resolve those please 🙏

nayounsang:
Additionally, after looking into the failed ci, I removed one tc from `plugin-test-formatting`, as I felt it was pointless to test for behavior when cooked is null: 
5e32527

this is because when cooked is null, it just ignore.

kirkwaiblinger (approved):

kirkwaiblinger:
Gonna merge this in, as it's been essentially ready to merge for a while, and no one else has had any issues with it in a few review cycles 🙂 . 

Thanks!

PR #11359, opened 2025-06-30
fix(eslint-plugin)[no-deprecated]: should report deprecated exports and reexports

mainframev:
## PR Checklist

## Overview

- Removed export statements from `isInsideExportOrImport` guard function and renamed it to `isInsideImport` to skip only imports.
- Updated the test file with export / reexports cases


packages/eslint-plugin/src/rules/no-deprecated.ts
```diff
@@ -436,7 +435,18 @@ export default createRule<Options, MessageIds>({
     }
 
     return {
-      Identifier: checkIdentifier,
+      ExportSpecifier(node): void {
+        const symbol = services.getSymbolAtLocation(node.exported);
+
+        if (searchForDeprecationInAliasesChain(symbol, false) == null) {
+          checkIdentifier(node.exported as TSESTree.Identifier);
```
nayounsang: safer to use conditional statements than type casting.
```ts
        if (
          searchForDeprecationInAliasesChain(symbol, false) == null &&
          node.exported.type === AST_NODE_TYPES.Identifier
        ) {
          checkIdentifier(node.exported);
        }
```
or
```ts
function checkIdentifier(node: IdentifierLike | TSESTree.StringLiteral): void
```

It seems like a rare case, this is example of literal. 
```ts
export { foo as 'bar' } from 'baz';
export { 'foo' } from 'bar';
```
(Maybe, there will be no problem in operation. Could you test these cases?)

mainframev: @nayounsang hi, thanks! updated

mainframev: @nayounsang hi, thanks! updated 👍🏻

packages/eslint-plugin/tests/rules/no-deprecated.test.ts
```diff
@@ -222,6 +222,15 @@ ruleTester.run('no-deprecated', rule, {
...
       } from 'typescript';
     `,
+    `
+      export { deprecatedFunction as 'bur' } from './deprecated';
+    `,
+    `
+      export { 'deprecatedFunction' } from './deprecated';
+    `,
+    `
+      export { deprecatedFunction as 'bar' } from './deprecated';
```
nayounsang: Nice, I think this is far from a real-world use case and can be safely ignored. But, this tc is dupe: `export { deprecatedFunction as 'bur' } from './deprecated';`

mainframev: @nayounsang removed, thanks! 

Also, I’ve tweaked it again a bit. While tests and lint were passing locally, on our side CI lint step was getting stuck in an infinite loop. After updating the `searchForDeprecationInAliasesChain` to handle alias chains, the hang no longer occurs

JoshuaKGoldberg (dismissed):
LGTM, thanks! 🚀

JoshuaKGoldberg:
`yarn lint` passes locally.

PR #11367, opened 2025-07-01
chore: make copy-ast-spec non cacheable

JamesHenry:
The `types:copy-ast-spec` task could hit the remote cache when the `ast-spec` build outputs were changing. This change ensures that is now impossible by making it so that `copy-ast-spec` will always run. It's a very cheap operation anyway.

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -70,7 +70,7 @@
         ]
       },
       "copy-ast-spec": {
-        "cache": true,
+        "cache": false,
         "command": "tsx tools/copy-ast-spec.mts",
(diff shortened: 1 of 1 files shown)


packages/ast-spec/package.json
```diff
@@ -77,7 +77,7 @@
           "cwd": "{projectRoot}"
         },
         "outputs": [
-          "{projectRoot}/dist/**/*.ts"
```
JamesHenry, line 80: This change might have been enough by itself TBH, but the explicit `inputs` ensures it

PR #11456, opened 2025-08-05
feat(eslint-plugin): [no-unused-vars] support explicit resource management with `ignoreUsingDeclarations` option

### LINKED ISSUE #11424: Enhancement: [no-unused-vars] support explicit resource management

https://typescript-eslint.io/rules/no-unused-vars/

### Description

From the ESLint v9.32.0 release notes:

> [no-unused-vars](https://eslint.org/docs/latest/rules/no-unused-vars): Treats variables declared with using/await using as used, since their Symbol.dispose is implicitly called at the end of their scope. Also adds an option to ignore unused using declarations.

### Fail

```TypeScript
var replace = 'me';
```

### Pass

```TypeScript
using foo = disposable();
```

### Additional Info

_No response_

nayounsang:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11424

## Overview

TODO: docs? maybe it can skip. [Because `option` depends on `eslint`](https://typescript-eslint.io/rules/no-unused-vars/#options)

- As eslint supports explict resource management(using, await using) with `ignoreUsingDeclarations` option, 
Add boolean type option, default is `false`. [ref](Ignore even if not referenced in the code block.)
- When `ignoreUsingDeclarations` is enable, ignore even if not referenced in the code block.
```ts
// pass
{
  using foo = getResource();
}
```
- Implement this with `options.ignoreUsingDeclarations` is enable && `def.parent.kind` is `using` or `await using` conditions.

ref
- [no-unused-vars ignoreUsingDeclarations](https://eslint.org/docs/latest/rules/no-unused-vars#ignoreusingdeclarations)
- [explicit-resource-management](https://github.com/tc39/proposal-explicit-resource-management?tab=readme-ov-file#ecmascript-explicit-resource-management)
- [implement on eslint](https://github.com/eslint/eslint/blob/main/lib/rules/no-unused-vars.js#L373C12-L373C42)

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
@@ -3063,7 +3063,7 @@ declare class Bar {}
     {
       code: `
 using resource = getResource();
-resource();
+resource;
       `,
       languageOptions: {
         parserOptions: {


packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -546,6 +557,12 @@ export default createRule<Options, MessageIds>({
             });
           }
           continue;
+        } else if (
```
kirkwaiblinger: Surprised this isn't caught by a lint rule, but we shouldn't have `else if` after the unconditional `continue` (along the lines of [no-else-return](https://eslint.org/docs/latest/rules/no-else-return)).

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
kirkwaiblinger, line 1: For consistency, could you make sure that the tests equivalent to https://github.com/eslint/eslint/pull/19828/files#diff-a011296b092c66311bc2ef19de30721cb1c5e9164be696f6f581ad4ba6231998 are included?

kirkwaiblinger (changes requested):
LGTM, just requesting a bit of alignment with https://github.com/eslint/eslint/pull/19828 

Thanks!

packages/eslint-plugin/src/rules/no-unused-vars.ts
nayounsang: Based on lint option([allowElseIf:false](https://eslint.org/docs/latest/rules/no-else-return#allowelseif-false)), conditional statement is separated into single if.

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars.test.ts
nayounsang: I specified the ECMA version and aligned the variable names. I also added a test case from the reference.

JoshuaKGoldberg (approved):
Swell! 🙌

PR #11467, opened 2025-08-08
docs: mark rules that are frozen on rules index page

### LINKED ISSUE #10669: Docs: mark rules that are frozen on rules index page

### Suggested Changes

Mark rules that are frozen with ❄️ on the rules index page. Same as https://github.com/eslint/eslint/issues/18694

### Affected URL(s)

https://typescript-eslint.io/rules/

### Additional Info

_No response_

nayounsang:
## PR Checklist

- [x] Addresses an existing open issue: fixes #10669

## Overview

1. mark [these rule](https://github.com/typescript-eslint/typescript-eslint/issues/10669#issuecomment-2601025196) as frozen
2. add frozen emoji in rule table on rule index page
3. add frozen description on each rule page
4. add frozen explanation on rule index page (Obviously, the frozen property is for contributors, so I'm not sure if the rule index page is the right place. ESLint also has an explanation in its contribution documentation. However, I couldn't find a suitable place in the contribution pages either.)
5. add `frozen` property in rule.docs meta.

```ts
// packages/utils/src/ts-eslint/Rule.ts

export interface DeprecatedInfo {
  /**
   * The estimated version when the rule is removed (probably the next major
   * version). null means the rule is "frozen" (will be available but will not
   * be changed).
   */
  availableUntil?: string | null;
}
```
We can determine if something is frozen by `deprecated.availableUntil === null`, but I found this method strange, so I add new property: `docs.frozen`.
- This interface is necessary for compatibility with eslint, but **deprecated** and **frozen** are different.
-  `deprecated` can be boolean or object. Additional conditions are needed to distinguish between the two. 
- Type guards (or additional conditions) are required to access object property.
- `availableUntil` can be undefined or null. Adding unnecessary comments to bypass the eqeqeq lint rule
- There are many codes written for these reason, and it is judged that they actually have a negative impact on maintenance.

TODO
- Check if other rules in eslint are frozen.


nayounsang:
Additionally, I found more frozen rule: [`default-param-last`](https://eslint.org/docs/latest/rules/default-param-last).
I don't think there will be any more in the extension rule.

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -25,6 +25,11 @@ export interface RuleMetaDataDocs {
    * The URL of the rule's docs.
    */
   url?: string;
+
+  /**
+   * Mark this rule as frozen.
+   */
+  frozen?: boolean;
```
nayounsang, line 32: If we decide where to put the description of frozen, can link it here too.

packages/eslint-plugin/docs/rules/README.md
```diff
@@ -55,3 +55,16 @@ module.exports = {
 ```
 
 [Search for `🧱 extension rule`s](?=extension#rules) in this page to see all extension rules.
+
+## Frozen Rules
+
+When rules are feature complete, they are marked as frozen (indicated with ❄️ in the documentation). This applies to standalone rules that are complete, as well as [extension rules](#extension-rules) whose underlying core ESLint rules are frozen. After that point, we expect users to use [disable comments](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) when they find an edge case that isn’t covered.
```
nayounsang, line 61: This is slightly different from the [ESLint's](https://eslint.org/docs/latest/contribute/core-rules#frozen-rules).
I don't quite understand the 80% policy, so I added a line that says the dependent ESLint rule is frozen.
However, my English isn't very good, so I'm not sure if the meaning is natural.

packages/website/src/components/RulesTable/styles.module.css
```diff
@@ -145,3 +145,9 @@
   text-align: center;
   cursor: default;
 }
+
+.ruleNameWrapper {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
```
nayounsang, line 152: ESLint uses this style.
I like it too. Appropriate gap

packages/eslint-plugin/docs/rules/README.md
JoshuaKGoldberg: Agreed, 80% isn't quite the same as our policy. This is perfect 🙂 I wouldn't have picked up on any problems with your English.

packages/website/src/components/RulesTable/styles.module.css
```diff
@@ -145,3 +145,9 @@
   text-align: center;
   cursor: default;
 }
+
+.ruleNameWrapper {
+  display: flex;
```
JoshuaKGoldberg: [Bug] Putting this new div into block flow is causing extra padding

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
Before screenshot of a few rules table rows showing no vertical space between rule name and description
</td>
<td>
After screenshot of a few rules table rows showing about 1rem of vertical space between rule name and description
</td>
</tbody>
</table>

The flex is good, but it should be inline:

```suggestion
  display: inline-flex;
```

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -25,6 +25,11 @@ export interface RuleMetaDataDocs {
    * The URL of the rule's docs.
    */
   url?: string;
+
+  /**
+   * Mark this rule as frozen.
```
JoshuaKGoldberg: [Docs] Nit: just to make this comment a bit more informative.

```suggestion
   * Mark this rule as feature-frozen.
```

packages/website/src/components/RulesTable/styles.module.css
nayounsang: image
Great. I committed.

packages/utils/src/ts-eslint/Rule.ts
nayounsang: I committed.

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -25,6 +25,11 @@ export interface RuleMetaDataDocs {
    * The URL of the rule's docs.
    */
   url?: string;
+
+  /**
+   * Mark this rule as feature = frozen.
```
JoshuaKGoldberg: ` = ` is not the same as `-` grammatically.

```suggestion
   * Mark this rule as feature-frozen.
```

JoshuaKGoldberg (dismissed):
Lovely, thanks! 🙌

PR #11469, opened 2025-08-08
fix(scope-manager): exclude Program from DefinitionBase node types

### LINKED ISSUE #11334: Bug: DefinitionBase shouldn't allow its Node type to be TSESTree.Program

### Relevant Package

scope-manager

### Playground Link

_No response_

### Repro Code

See https://github.com/typescript-eslint/typescript-eslint/pull/11322/files#r2158282689

### Expected Result

The types of nodes on scope-manager definitions are, to my knowledge, always nodes _inside_ a source file. They're never the `TSESTree.Program` themselves. Thus, they should always have a defined `parent` property set to some parent node.

### Actual Result

`DefinitionBase`'s `node` property's type is from a type parameter constrained to `TSESTree.Node`:

https://github.com/typescript-eslint/typescript-eslint/blob/a157fd4c62741d2dec2e4f47f738f499b616035d/packages/scope-manager/src/definition/DefinitionBase.ts#L11

https://github.com/typescript-eslint/typescript-eslint/blob/a157fd4c62741d2dec2e4f47f738f499b616035d/packages/scope-manager/src/definition/DefinitionBase.ts#L32

`TSESTree.Node` includes `TSESTree.Program` as a type constituent. Which is problematic because that's the only type of node that doesn't have a required `parent` property.

We see in https://github.com/typescript-eslint/typescript-eslint/pull/11322/files#r2158282689 that some areas of code have to use a `!` after `node.parent`.

### Additional Info

My initial instinct for solving this is to:

1. Make a `TSESTree.NodeWithParent` type union with all the constituents of `TSESTree.Node` _except_ `TSESTree.Program`
2. Use that `TSESTree.NodeWithParent` in `DefinitionBase` and other scope-manager types that can only ever be a node inside a file (i.e. with a parent)

### Versions

| package                                | version  |
| -------------------------------------- | -------- |
| `@typescript-eslint/scope-manager`     | `8.34.1` |
| `@typescript-eslint/types`             | `8.34.1` |

💖

dbarabashh:
Fix #11334

## PR Checklist

- [x] Addresses an existing open issue: fixes #11334

## Overview

This PR improves type safety in the `scope-manager` package by preventing `TSESTree.Program` nodes from being used as definition nodes.


dbarabashh:
I am not sure that the linter failed now exactly because of my changes, because locally this command was successful

JoshuaKGoldberg:
Try a `yarn build` first, then a `yarn lint`? It looks like the types are better informed because of this change (nice!), so lint rules are pointing out previously missed unnecessary `?.`s and the like.

dbarabashh:
@JoshuaKGoldberg thank you, done

kirkwaiblinger:
@dbarabashh Would you mind also auditing for uses of `nullThrows` that can be removed? For example, 

https://github.com/typescript-eslint/typescript-eslint/blob/63ab00225eea8c9e4c8fadc22d5c3081daf13e18/packages/eslint-plugin/src/rules/no-confusing-void-expression.ts#L368

packages/scope-manager/src/definition/DefinitionBase.ts
```diff
@@ -5,10 +5,15 @@ import type { DefinitionType } from './DefinitionType';
 import { createIdGenerator } from '../ID';
 
 const generator = createIdGenerator();
+/**
+ * Helper type to exclude Program from valid definition nodes
+ * Program is the root node and doesn't have a parent property
+ */
+export type NodeWithParent = Exclude<TSESTree.Node, TSESTree.Program>;
```
JoshuaKGoldberg: 🤔 This type doesn't really belong to the scope-manager. Any consumer of the TSESTree nodes is likely to want it, such as the `collectUnusedVariables` change already in this PR. I this this should really be `TSESTree.NodeWithParent`.

JoshuaKGoldberg (changes requested):
+1 to what Kirk said. I also think the new type is good enough that it should be moved to `@typescript-eslint/types` altogether. Thanks for starting this! 🙌

packages/scope-manager/src/definition/DefinitionBase.ts
dbarabashh: I got it, will do

dbarabashh: I hope the commit I made meets the expectations. If something is needed, let me know.

JoshuaKGoldberg (approved):
🤤 look at all those removed `?~s and `nullThrows()`s! Absolutely lovely. Thanks!

kirkwaiblinger (approved):
🎉🎉

PR #11471, opened 2025-08-10
docs(utils): grammar

Zamiell:
## PR Checklist

- [ ] Addresses an existing open issue: n/a
- [ ] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22): n/a

## Overview

Semi colons were being used incorrectly.

### DIFF
--- a/packages/utils/src/ts-eslint/eslint/FlatESLint.ts
+++ b/packages/utils/src/ts-eslint/eslint/FlatESLint.ts
@@ -49,13 +49,12 @@ export namespace FlatESLint {
      */
     ignorePatterns?: string[] | null;
     /**
-     * The path to a configuration file.
+     * The path to a configuration file. Overrides all configurations used with this instance.
+     * The options.overrideConfig option is applied after this option is applied.
      *
-     * - Overrides all configurations used with this instance.
-     * - The options.overrideConfig option is applied after this option is applied.
-     * - Searches for default config file when falsy.
-     * - Does not do any config file lookup when `true`.
-     * - Considered to be a config filename when a string.
+     * - When falsy, searches for default config file.
+     * - When `true`, does not do any config file lookup.
+     * - When a string, considered to be a config file name.
      *
      * @default false
      */


packages/utils/src/ts-eslint/eslint/FlatESLint.ts
```diff
@@ -51,7 +51,7 @@ export namespace FlatESLint {
     /**
      * The path to a configuration file, overrides all configurations used with this instance.
      * The options.overrideConfig option is applied after this option is applied.
-     * Searches for default config file when falsy; doesn't do any config file lookup when `true`; considered to be a config filename when a string.
+     * Searches for default config file when falsy. Does not do any config file lookup when `true`. Considered to be a config filename when a string.
```
Josh-Cena: Previous text is correct and makes slightly more sense to me, although I would personally make it a bulleted list

Zamiell: @Josh-Cena I have updated it to this:
```ts
    /**
     * The path to a configuration file.
     *
     * - Overrides all configurations used with this instance.
     * - The options.overrideConfig option is applied after this option is applied.
     * - Searches for default config file when falsy.
     * - Does not do any config file lookup when `true`.
     * - Considered to be a config filename when a string.
     *
     * @default false
     */
    overrideConfigFile?: boolean | string;
 ```
 Is that what you wanted?

File: packages/utils/src/ts-eslint/eslint/FlatESLint.ts [hunk 2]
```diff
@@ -49,9 +49,14 @@ export namespace FlatESLint {
...
-     * The path to a configuration file, overrides all configurations used with this instance.
-     * The options.overrideConfig option is applied after this option is applied.
-     * Searches for default config file when falsy; doesn't do any config file lookup when `true`; considered to be a config filename when a string.
+     * The path to a configuration file.
+     *
+     * - Overrides all configurations used with this instance.
+     * - The options.overrideConfig option is applied after this option is applied.
+     * - Searches for default config file when falsy.
+     * - Does not do any config file lookup when `true`.
+     * - Considered to be a config filename when a string.
```
Josh-Cena:
```suggestion
     * The path to a configuration file. Overrides all configurations used with this instance.
     * The options.overrideConfig option is applied after this option is applied.
     *
     * - When falsy, searches for default config file.
     * - When `true`, does not do any config file lookup.
     * - When a string, considered to be a config file name.
```

Issue #7654, opened 2023-09-15
Bug: prefer-optional-chain suggests optional chaining during strict null equality check


justinbhopper:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.
- [X] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [X] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.2.2&fileType=.tsx&code=LAKAlgdgLgpgTgMwIYGMYAIAiSpPQb1HXQHoT0BJAWwAcB7OXaALnQAZ0wBndPANyQAbMABN0AwQFcYRcUOmsIkqgCN46AD7olgwQG5QAX1CgUdCFyjoROJK2y5N2ybvQBedAHMYUB0gAUAJQGIKBk6ADCABYwKADWkJ7oKnRQUehpOOgABja42Zw8UHCSaQCe6ACCAHKYGVFZubYAdBLSBdzaqc6u-mooSJJcGBydEqKBoGAI6P55eABkC9YtbRgAhG4eOoKBBLJmFnSCMM2CdJ7%2BAETVympwrFfoANQruK3yMMFG6DCCw-sQMRDlxjqdzpcrhQIOMxPNmldviBjKEQAhJBAUFAwOYvD4-EF7LYnDtAcQ4D5JHAIAQ5FIYKwAMzoQwhFGgIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uii9o-aJHBgAviHVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```typescript
const data: Data | null = getData();

// Checking both that `data` is truthy AND that `data.value` is not null (because 0 is valid)
if (data && data.value !== null) {
  console.log("Number: " + data.value);
} else {
  console.log("Invalid data.");
}
```

[...]

PR #11533, opened 2025-08-27
fix(eslint-plugin): [prefer-optional-chain] include mixed "nullish comparison style" chains in checks

### LINKED ISSUE #7170: [prefer-optional-chain] include "mixed truthiness" chains in checks
https://github.com/typescript-eslint/typescript-eslint/pull/6397#issuecomment-1560490598

In v6 we have rewritten the `prefer-optional-chain` rule from the ground up.
The new logic assesses the "nullish comparison style" of logical operand in relation to the logical operator to determine if it should be consider part of a chain. EG `||` will only assess "equal nullish" checks, and `&&` will only assess "not equal to nullish" checks.

This does lead to one hole in the rule:

```ts
declare const foo: { bar: number } | null;
foo === null || foo.bar;
foo !== null && !foo.bar;
```

In the first case `foo.bar` is seen as a "truthy boolean" check (eg "not equal nullish") - which isn't a valid operand in an "OR" chain.
Similarly in the second case `!foo.bar` is seen as a "falsy boolean" check (eg "equal nullish") - which isn't a valid operand in an "AND" chain.

This means that the rule will ignore it and thus will not include it in the chain to report and fix.

This logic is correct and sound - however it doesn't properly encapsulate developer intent. Specifically when that "invalid" chain operand is the very last member in a chain - in certain cases we will actually want to include it as part of the chain.

A slightly longer example:
```ts
foo && foo.bar && foo.bar.baz === 1
// rule will currently fix to
foo?.bar && foo.bar.baz === 1
// but ideally it would fix it to
foo?.bar?.baz === 1
```

I think there's a subset of checks that it should include:
- "boolean" checks (eg `foo` and `!foo`)
- `!==`/`===`/`!=`/`==` comparing against anything except `typeof`
   - not `typeof` because `foo && typeof foo.bar` will return `undefined | typeof foo.bar` whereas `typeof foo?.bar` will return `typeof foo.bar | 'undefined'` - which is VERY different
- should not include `>`/`>=`/`<`/`<=` because TS will hard error if one of the operands is nullish.

mdm317:
## PR Checklist

- [x] Addresses an existing open issue: fixes #7170

## Overview

This is still a work in progress, but I have one unclear point. I have a question and was hoping you could help.
I want to make sure the categories are clearly defined: `suggestion`, `error`, and `valid`.
I’m especially unsure about suggestion — could you explain that part?

### Valid
Cases where converting to optional chaining may change the behavior -- especially with truthy and falsy values. In such cases we do not report at all.
Example from [issue 7654](https://github.com/typescript-eslint/typescript-eslint/issues/7654)
```
const data: Data | null = getData();

// Checking both that `data` is truthy AND that `data.value` is not null (because 0 is valid)
if (data && data.value !== null) {
  console.log("Number: " + data.value);
} else {
  console.log("Invalid data.");
}
```

### Error
Conversion to optional chaining either produces the same result as before, or introduces an option value.
```
foo != null && foo.bar
foo?.bar
```
### Suggestion 
Optional chaining introduces an additional undefined type, which changes the result type but preserves the same truthy/falsy behavior.
```
declare const foo: { bar: number } | null | undefined;
foo != undefined && foo.bar;         => false | number
foo?.bar                                           => undefined | number
```
Currently, even when the type changes, especially truthy and false some cases report as a suggestion.
examples 
```
declare const foo : { bar : number | null } | null

foo == null || foo.bar === null    => true when foo is null
foo?.bar === null                         => false when foo is null
```
[repro](https://typescript-eslint.io/play/#ts=5.2.2&fileType=.tsx&code=LAKAJgpgxgNghgJwgAigewHYGcAuyBmaaAXMgN7IBGipGArgLaUQLIA%2By9MMyAvu5zrcA3KFABLfMgAUhNMgC8Cwd3Yc5AOmqslyrjACU5UMgJFhyAPSWz89BhxxxGFTFC8JU2UQD8WxIpKrkZkJrYW1racaHj2js6u7kA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uii9o-aJHBgAviHVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)
I think we should fix these cases not to be reported as suggestions.

Am I understanding this correctly and approaching it the right way?
If not, please clarify how it should continue to be reported as a suggestion.

### DIFF
--- a/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
+++ b/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
@@ -326,6 +326,10 @@ function getReportDescriptor(
     lastOperand.comparisonType === NullishComparisonType.StrictEqualUndefined ||
     lastOperand.comparisonType ===
       NullishComparisonType.NotStrictEqualUndefined ||
+    lastOperand.comparisonType === ComparisonType.Equal ||
+    lastOperand.comparisonType === ComparisonType.NotEqual ||
+    lastOperand.comparisonType === ComparisonType.NotStrictEqual ||
+    lastOperand.comparisonType === ComparisonType.StrictEqual ||
     (operator === '||' &&
       lastOperand.comparisonType === NullishComparisonType.NotBoolean)
   ) {
@@ -652,6 +656,31 @@ export function analyzeChain(
       //                    ^^^^^^^ invalid OR chain logical, but still part of
       //                            the chain for combination purposes
 
+      if (lastOperand) {
+        const comparisonResult = compareNodes(
+          lastOperand.comparedName,
+          operand.comparedName,
+        );
+        switch (operand.comparisonType) {
+          case NullishComparisonType.NotStrictEqualUndefined: {
+            if (comparisonResult === NodeComparisonResult.Subset) {
+              subChain.push(operand);
+            }
+          }
+          case NullishComparisonType.NotStrictEqualNull: {
+            if (
+              comparisonResult === NodeComparisonResult.Subset &&
+              isTruthyOperand(
+                lastOperand.comparedName,
+                lastOperand.comparisonType,
+                parserServices,
+              )
+            ) {
+              subChain.push(operand);
+            }
+          }
+        }
+      }
       maybeReportThenReset();
       continue;
     }
--- a/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
+++ b/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
@@ -10,10 +10,10 @@ import type {
 } from '@typescript-eslint/utils/ts-eslint';
 
 import { AST_NODE_TYPES } from '@typescript-eslint/utils';
-import { unionConstituents } from 'ts-api-utils';
+import { isFalsyType, unionConstituents } from 'ts-api-utils';
 import * as ts from 'typescript';
 
-import type { ValidOperand } from './gatherLogicalOperands';
+import type { LastChainOperand, ValidOperand } from './gatherLogicalOperands';
 import type {
   PreferOptionalChainMessageIds,
   PreferOptionalChainOptions,
@@ -31,7 +31,11 @@ import {
 } from '../../util';
 import { checkNullishAndReport } from './checkNullishAndReport';
 import { compareNodes, NodeComparisonResult } from './compareNodes';
-import { NullishComparisonType } from './gatherLogicalOperands';
+import {
+  ComparisonType,
+  NullishComparisonType,
+  OperandValidity,
+} from './gatherLogicalOperands';
 
 function includesType(
   parserServices: ParserServicesWithTypeInformation,
@@ -46,6 +50,66 @@ function includesType(
     }
   }
   return false;
+}
+
+function isTruthyOperand(
+  comparedName: TSESTree.Node,
+  nullishComparisonType: NullishComparisonType | ComparisonType,
+  parserServices: ParserServicesWithTypeInformation,
+): boolean {
+  let ANY_UNKNOWN_FLAGS = ts.TypeFlags.Any | ts.TypeFlags.Unknown;
+  const comparedNameType = parserServices.getTypeAtLocation(comparedName);
+
+  if (isTypeFlagSet(comparedNameType, ANY_UNKNOWN_FLAGS)) {
+    return false;
+  }
+  switch (nullishComparisonType) {
+    case NullishComparisonType.Boolean:
+      const types = unionConstituents(comparedNameType);
+      return types.every(type => !isFalsyType(type));
+    case NullishComparisonType.NotStrictEqualUndefined:
+      return !isTypeFlagSet(comparedNameType, ts.TypeFlags.Undefined);
+    case NullishComparisonType.NotStrictEqualNull:
+      return !isTypeFlagSet(comparedNameType, ts.TypeFlags.Null);
+    case NullishComparisonType.NotEqualNullOrUndefined:
+      return !isTypeFlagSet(
+        comparedNameType,
+        ts.TypeFlags.Null | ts.TypeFlags.Undefined,
+      );
+    default:
+      return false;
+  }
+}
+
+function isValidLastChainOperand(
+  ComparisonValueType: TSESTree.Node,
+  operator: TSESTree.BinaryExpression['operator'],
+  parserServices: ParserServicesWithTypeInformation,
+) {
+  const type = parserServices.getTypeAtLocation(ComparisonValueType);
+  let ANY_UNKNOWN_FLAGS = ts.TypeFlags.Any | ts.TypeFlags.Unknown;
+
+  switch (operator) {
+    case '==': {
+      const types = unionConstituents(type);
+      const isNullish = types.some(t =>
+        isTypeFlagSet(
+          t,
+          ANY_UNKNOWN_FLAGS | ts.TypeFlags.Null | ts.TypeFlags.Undefined,
+        ),
+      );
+      return !isNullish;
+    }
+    case '===': {
+      const types = unionConstituents(type);
+      const isUndefined = types.some(t =>
+        isTypeFlagSet(t, ANY_UNKNOWN_FLAGS | ts.TypeFlags.Undefined),
+      );
+      return !isUndefined;
+    }
+    default:
+      return false;
+  }
 }
 
 // I hate that these functions are identical aside from the enum values used
@@ -66,14 +130,6 @@ const analyzeAndChainOperand: OperandAnalyzer = (
 ) => {
   switch (operand.comparisonType) {
     case NullishComparisonType.Boolean: {
-      const nextOperand = chain.at(index + 1);
-      if (
-        nextOperand?.comparisonType ===
-          NullishComparisonType.NotStrictEqualNull &&
-        operand.comparedName.type === AST_NODE_TYPES.Identifier
-      ) {
-        return null;
-      }
       return [operand];
     }
 
@@ -521,10 +577,11 @@ export function analyzeChain(
   node: TSESTree.Node,
   operator: TSESTree.LogicalExpression['operator'],
   chain: ValidOperand[],
+  lastChainOperand?: LastChainOperand,
 ): void {
   // need at least 2 operands in a chain for it to be a chain
   if (
-    chain.length <= 1 ||
+    chain.length + (lastChainOperand ? 1 : 0) <= 1 ||
     /* istanbul ignore next -- previous checks make this unreachable, but keep it for exhaustiveness check */
     operator === '??'
   ) {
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/no-base-to-string.ts
```diff
@@ -270,6 +270,7 @@ export default createRule<Options, MessageIds>({
 
       const declarations = toString.getDeclarations();
 
+      // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
       if (declarations == null || declarations.length !== 1) {
```
mdm317, line 321: > const declarations: ts.Declaration[] | undefined

When the new logic was applied, this code
`declarations == null || declarations.length !== 1`
was changed to
`declarations?.length !== 1 `

But I think the original code is more readable and intuitive.
For cases like `foo == null || <equaliy check>`, should we avoid optional chaining?
What do you think?
[example new playground](https://deploy-preview-11533--typescript-eslint.netlify.app/play/#ts=5.9.2&fileType=.tsx&code=FAEwpgxgNghgTmABBA9gOwM4BdHmjLAS3QwC5Fs5C0BzAbQF1EAfRAVzXADNqwRhghLogAUeWERKIAvNMRo2UKC1biCxTADooYWlgAWiAISzEARgCUiAN7BEuSBI0ZgAXwFDRayZgD823RoDY1NLGzsHfB8XVyA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uii9o-aJHBgAviHVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
```diff
@@ -48,6 +48,109 @@ function includesType(
   return false;
 }
 
+function isAlwaysTruthyOperand(
```
mdm317, line 51: I'm not convinced this function is necessary.
This function checks if the previous operand is always true:
- `foo != null && foo.bar == somevalue` → `foo?.bar == somevalue` (same result when foo is not nullish)
- `foo && foo.bar == somevalue` → `foo?.bar == somevalue` (same result when foo is truthy)

But if we know `foo` is not nullish, we can simply write `foo.bar == somevalue` without optional chaining at all.

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
```diff
@@ -65,18 +168,7 @@ const analyzeAndChainOperand: OperandAnalyzer = (
   chain,
 ) => {
   switch (operand.comparisonType) {
-    case NullishComparisonType.Boolean: {
```
mdm317, line 68: This logic moved to line 700.

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
```diff
@@ -595,6 +697,35 @@ export function analyzeChain(
...
+        switch (operand.comparisonType) {
+          case NullishComparisonType.StrictEqualUndefined:
+          case NullishComparisonType.NotStrictEqualUndefined: {
+            if (comparisonResult === NodeComparisonResult.Subset) {
+              lastChain = operand;
+            }
+            break;
+          }
+          case NullishComparisonType.StrictEqualNull:
+          case NullishComparisonType.NotStrictEqualNull: {
```
mdm317, line 714: When the last parameter is StrictEqualNull like `foo && foo.bar !== null`, we can't convert to optional chaining because:
- `foo && foo.bar !== null` returns `false` when foo is nullish
- `foo?.bar !== null` returns `true` when foo is nullish

Also StrictEqualNull case foo == null | foo.bar !== null

- `foo == null || foo.bar !== null` returns `true` when foo is nullish
- `foo?.bar !== null` returns `false` when foo is nullish

This issue is documented [here](https://github.com/typescript-eslint/typescript-eslint/issues/7654).

But as I mentioned above, I'm still curious whether this logic is necessary at all.

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
```diff
@@ -624,7 +755,33 @@ export function analyzeChain(
...
+        ? isValidAndLastChainOperand
+        : isValidOrLastChainOperand;
+    if (
+      comparisonResult === NodeComparisonResult.Subset &&
+      (isAlwaysTruthyOperand(
+        lastOperand.comparedName,
+        lastOperand.comparisonType,
+        parserServices,
+      ) ||
+        isValidLastChainOperand(
```
mdm317, line 776: This function checks if the last operand is valid as outlined in the table.

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
```diff
@@ -374,4 +429,32 @@ export function gatherLogicalOperands(
...
   }
+
+  function getBinaryComparisonChain(node: TSESTree.BinaryExpression) {
+    const { left, right } = node;
+    let isYoda = false;
+    const isLeftMemberExpression =
+      left.type === AST_NODE_TYPES.MemberExpression;
+    const isRightMemberExpression =
+      right.type === AST_NODE_TYPES.MemberExpression;
+    if (isLeftMemberExpression && !isRightMemberExpression) {
```
mdm317, line 440: Only checks MemberExpression because equality checks cannot be the first chain element.

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1991,16 +3489,58 @@ describe('base cases', () => {
...
+              `,
+              errors: [{ messageId: 'preferOptionalChain' }],
+              output: `
+                declare const foo: {
+                  bar: () =>
+                    | { baz: { buzz: (() => number) | null | undefined } | null | undefined }
+                    | null
+                    | undefined;
+                };
+                foo.bar?.() !== null &&
```
mdm317, line 3526: If `foo.bar` is guaranteed to be non-nullish, we can change it to optional chaining.

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -2114,20 +3738,62 @@ describe('base cases', () => {
...
+            {
+              code: `
+                declare const foo: {
+                  bar: () =>
+                    | { baz: { buzz: (() => number) | null | undefined } | null | undefined }
+                    | null
+                    | undefined;
+                };
+                foo.bar === undefined ||
+                  foo.bar() === undefined ||
```
mdm317, line 3766: Since `operand` is a nullish check, 
`operand || foo.bar() === undefined` can be converted to optional chaining.

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
JoshuaKGoldberg: Thanks for these tips, they made reviewing this (already, by necessity, complex) logic easier. 🙂

packages/eslint-plugin/src/rules/no-base-to-string.ts
JoshuaKGoldberg: If the choices are these:

* Original: `if (declations == null || declations.length !== 1) {`
* Updated: `if (declations?.length !== 1) {`

Personal anecodtal thoughts: I find the updated form more readable. I think when `?.` was new I would probably have preferred the original for being more familiar. But at this point I'm pretty comfortable with `?.` and `??`, and it feels natural to me.

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
JoshuaKGoldberg, line 1: [Praise] Such thorough testing! Lovely. 👏 🔥

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/gatherLogicalOperands.ts
```diff
@@ -182,61 +197,101 @@ export function gatherLogicalOperands(
...
-                    : NullishComparisonType.StrictEqualUndefined,
-                  isYoda,
-                  node: operand,
-                  type: OperandValidity.Valid,
-                });
-                continue;
+        // x == something :(
+        // x === something :(
+        // x != something :(
+        // x !== something :(
```
JoshuaKGoldberg, line 257: [Praise] The `:(`s made me chuckle. 😄

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
JoshuaKGoldberg: Removing it fails some unit tests (that have legitimate cases, I believe). E.g.:

```ts
declare const foo: { bar: string } | null;
foo !== null && foo.bar !== null;
```

bradzacher (approved):

B4nan:
This PR is too aggressive and suggests optional chaining in cases where it would change behavior.

Few cases from https://github.com/mikro-orm/mikro-orm/pull/6924 (when applying the suggested fixes, the better half of the test suite breaks).

```diff
-    if (meta.properties[propertyName] && meta.properties[propertyName].kind !== reference) {
+    if (meta.properties[propertyName]?.kind !== reference) {
```

```diff
-    if (!meta.embeddable && (!meta.primaryKeys || meta.primaryKeys.length === 0)) {
+    if (!meta.embeddable && (meta.primaryKeys?.length === 0)) {
```

```diff
-      if (!prop.ownColumns || prop.ownColumns.length === 0) {
+      if (prop.ownColumns?.length === 0) {
```

```diff
-    if (!last || last[0] !== entityName || last[1] !== prop) {
+    if (last?.[0] !== entityName || last[1] !== prop) {
```

Basically, all the changes that the PR suggests are wrong.

kirkwaiblinger:
@B4nan We ask that you please file a new issue rather than commenting on a closed PR, see https://typescript-eslint.io/contributing/issues#commenting.

I'm not able to reproduce what you're seeing in the playground based on my best guess at creating a reproduction... This code does not result in a lint report:

([playground](https://typescript-eslint.io/play#ts=5.9.3&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3lAXPA3vAEYD8eKArgLZEgwDaAuvAL4DcAUBwJYBm8ACgCEUAHRF4AH0nZxoiCBQBzDAAt4AXi3wADAEoCHePA4sgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uigB3UdCaRwYAL4gNQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false))

```ts
declare const a: { b?: number[] };

if (!a.b || a.b.length === 0) {
  
}
```

Can you please help us help you by creating an isolated reproduction we can try, and filing a new issue with it? Thanks!

aboks:
We also observe overly 'aggresive' behavior from this rule, as described in the comment by @B4nan. I've opened https://github.com/typescript-eslint/typescript-eslint/issues/11700 with a reproduction case.

PR #11591, opened 2025-09-10
fix(typescript-estree): add `TSImportType.source`

### LINKED ISSUE #11583: Enhancement: align `TSImportType` with `ImportExpression`

### Relevant Package

ast-spec

### Description

For similar code

```ts
let value = import("foo")
type type = import("foo")
```

The first was `ImportExpression` with `source`(`Literal`) property, while the second was `TSImportType` with `argument`(`TSLiteralType`).

Do you think they can share the same structure by changing `TSImportType.argument` to `TSImportType.source` and removing the `TSLiteralType` wrapper?

### Additional Info

[Playground](https://typescript-eslint.io/play/#ts=5.3.3&showAST=es&fileType=.tsx&code=DYUwLgBAbghsCuIIF4IEsC2AHA9gJzAAoAiAMxx2IEoBYAKDAE8skmWV1t8iyLqg&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

fisker:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11583


packages/visitor-keys/package.json
```diff
@@ -45,7 +45,7 @@
     "typecheck": "yarn run -BT nx typecheck"
   },
   "dependencies": {
-    "@typescript-eslint/types": "8.43.0",
+    "@typescript-eslint/types": "workspace:*",
```
fisker: Changed this because it can't build with stable version.

fisker: Still not working..

fisker:
Cherry-picked fix for https://github.com/typescript-eslint/typescript-eslint/issues/11593#issuecomment-3551414168, hope the test will pass.

fisker:
Strange, this was released, but not mentioned in the changelog. https://github.com/typescript-eslint/typescript-eslint/commit/6fb1551634b2ff11718e579098f69e041a2ff92c

Did I use the wrong `fix(typescript-estree)`?

```
> require('@typescript-eslint/typescript-estree').parse('type a = import("a")').body[0].typeAnnotation
{
  type: 'TSImportType',
  options: null,
  qualifier: null,
  source: { type: 'Literal', raw: '"a"', value: 'a' },
  typeArguments: null
}
> require('@typescript-eslint/typescript-estree').parse('type a = import("a")').body[0].typeAnnotation.argument
{
  type: 'TSLiteralType',
  literal: { type: 'Literal', raw: '"a"', value: 'a' },
  range: [ 16, 19 ],
  loc: { end: { column: 19, line: 1 }, start: { column: 16, line: 1 } }
}
```

JoshuaKGoldberg:
Hmm, the PR title should be correct. Maybe it's an issue with changelog generation?

fisker:
Also, we are supposed to make this a non-breaking change, but it breaks eslint-plugin-unicorn, https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2829#issuecomment-3574407664

I think that's because we removed `argument` from visitor keys. Unicorn is fine with this, but I don't know if other plugin will be broken.

How do we usually do with deprecated property? Keep or remove?

JamesHenry:
First guess without looking into it deeply is that the commit body wasn't cleaned up before merging, it looks a little whacky with those revert commit references:

image

JamesHenry:
Yeah that's it. We detect reverted hashes present in a commit in order to determine if a change was merged and reverted within the same release and can therefore strip it so that it isn't noise on the changelog.

Because there are reverted references in the commit body, the overall change gets classed as a revert, which is hidden from the changelog

JamesHenry:
I don't think there is any way for the release tooling to differentiate these scenarios, sadly I think it's down to maintainers to ensure the commit bodies are clean before merging. Tbh I always nuke the entire body when I squash merge...

Revert commits during the lifecycle of a PR are a bit more of an uncommon thing in my experience, so I think that's why this specific case hasn't come up before.

But yeah, overall those revert references have zero value for posterity because they were reverts of commits within a deleted PR branch, so I think the ultimate solution here is to make sure they are cleaned up on merge (SO wish GitHub had added squash merge validation hooks by now, it's been so long)

PR #11592, opened 2025-09-10
chore(typescript-estree): add `Converter#convertChildren()`

fisker:
## PR Checklist

### DIFF
--- a/packages/typescript-estree/src/convert.ts
+++ b/packages/typescript-estree/src/convert.ts
@@ -1388,7 +1388,7 @@ export class Converter {
           return this.createNode<TSESTree.ArrayPattern>(node, {
             type: AST_NODE_TYPES.ArrayPattern,
             decorators: [],
-            elements: this.convertChildren(node.elements),
+            elements: node.elements.map(el => this.convertPattern(el)),
             optional: false,
             typeAnnotation: undefined,
           });
--- a/packages/typescript-estree/src/convert.ts
+++ b/packages/typescript-estree/src/convert.ts
@@ -650,7 +650,7 @@ export class Converter {
    * @returns the converted ESTree node list
    */
   private convertChildren(
-    children?: ts.NodeArray<ts.Node> | ts.Node[],
+    children?: ts.Node[] | ts.NodeArray<ts.Node>,
     parent?: ts.Node,
   ): any {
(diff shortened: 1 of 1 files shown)


packages/typescript-estree/src/convert.ts
```diff
@@ -643,6 +643,19 @@ export class Converter {
...
+  /**
+   * Converts TypeScript node array into an ESTree node list.
+   * @param children the child `ts.NodeArray` or `ts.Node[]`
+   * @param parent parentNode
+   * @returns the converted ESTree node list
+   */
+  private convertChildren(
+    children?: ts.Node[] | ts.NodeArray<ts.Node>,
+    parent?: ts.Node,
+  ): any {
```
kirkwaiblinger: i don't think this should have a return type of `any`? If it's tedious to specify we can leave it inferred?

kirkwaiblinger: Looks like the real type is `undefined | any[]`, which is not super specific, but still meaningfully better than `any`.

packages/typescript-estree/src/convert.ts
```diff
@@ -643,6 +643,19 @@ export class Converter {
...
   }
 
+  /**
+   * Converts TypeScript node array into an ESTree node list.
+   * @param children the child `ts.NodeArray` or `ts.Node[]`
+   * @param parent parentNode
+   * @returns the converted ESTree node list
+   */
+  private convertChildren(
+    children?: ts.Node[] | ts.NodeArray<ts.Node>,
```
kirkwaiblinger: This is never invoked with the `children` parameter `undefined`... So let's either require it

```suggestion
    children: ts.Node[] | ts.NodeArray<ts.Node>,
```

or add logic to handle the undefined array the way the call sites do? Like `return (children ?? []).map(...)`

packages/typescript-estree/src/convert.ts
kirkwaiblinger: (depending on https://github.com/typescript-eslint/typescript-eslint/pull/11592/files#r2352644007, it may end up being `any[]` instead of `undefined | any[]`)

kirkwaiblinger (changes requested):
Just minor touchups requested

PR #11616, opened 2025-09-18
feat(eslint-plugin): expose rule name via RuleModule interface

### LINKED ISSUE #11608: [feature request] Expose the rule name as a property
```ts
const createRule = ESLintUtils.RuleCreator(
  (name) => `https://github.com/gajus/eslint-plugin-panda/blob/main/docs/rules/${name}.md`,
)

const myRule = createRule({
  name: 'no-config-function-in-source',
  // ...
});
```

At the moment, there is no way to get the name of the `myRule`.

The current situation forces weird patterns like:

```ts
export const RULE_NAME = 'no-config-function-in-source'

export default createRule({
  name: RULE_NAME,
  // ...
});
```

Ideally, we could get it via `myRule.name`.

y-hsgw:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11608

## Overview

This PR introduces an `name` property to the `RuleModule` interface.

🚀

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts
@@ -40,6 +40,7 @@ ruleTester.defineRule('use-every-a', {
     schema: [],
     type: 'problem',
   },
+  name: 'rule',
 });
 
 /**
--- a/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts
@@ -40,7 +40,7 @@ ruleTester.defineRule('use-every-a', {
     schema: [],
     type: 'problem',
   },
-  name: 'rule',
+  name: 'use-every-a',
 });
 
 /**
--- a/packages/utils/src/eslint-utils/RuleCreator.ts
+++ b/packages/utils/src/eslint-utils/RuleCreator.ts
@@ -30,6 +30,7 @@ export interface RuleCreateAndOptions<
   defaultOptions: Readonly<Options>;
 }
 
+/** @deprecated use `'RuleWithMetaAndName'` */
 export interface RuleWithMeta<
   Options extends readonly unknown[],
   MessageIds extends string,
@@ -76,6 +77,7 @@ export function RuleCreator<PluginDocs = unknown>(
           url: urlCreator(name),
         },
       },
+      name,
       ...rule,
     });
   };
@@ -89,7 +91,8 @@ function createRule<
   create,
   defaultOptions,
   meta,
-}: Readonly<RuleWithMeta<Options, MessageIds, PluginDocs>>): RuleModule<
+  name,
+}: Readonly<RuleWithMetaAndName<Options, MessageIds, PluginDocs>>): RuleModule<
   MessageIds,
   Options,
   PluginDocs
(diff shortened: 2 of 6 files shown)


packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts
```diff
@@ -40,6 +40,7 @@ ruleTester.defineRule('use-every-a', {
     schema: [],
     type: 'problem',
   },
+  name: 'rule',
```
JoshuaKGoldberg: [Testing] For consistency with the rest of the block:

```suggestion
  name: 'use-every-a',
```

🤔 I wonder if we should write a lint rule for this...

packages/utils/src/ts-eslint/Rule.ts
```diff
@@ -735,6 +735,10 @@ export interface RuleModule<
    * Metadata about the rule
    */
   meta: RuleMetaData<MessageIds, Docs, Options>;
+  /**
```
JoshuaKGoldberg, line 739: [Style] Nit: the other properties have a blank line too.

```suggestion

  /**
```

packages/utils/src/eslint-utils/RuleCreator.ts
```diff
@@ -30,6 +30,7 @@ export interface RuleCreateAndOptions<
   defaultOptions: Readonly<Options>;
 }
 
+/** @deprecated use `'RuleWithMetaAndName'` */
 export interface RuleWithMeta<
```
JoshuaKGoldberg, line 33: [Refactor] I'm on the fence here: what if we instead added the `name` property to `RuleWithMeta`? `RuleWithMetaAndName` is a bit more precise, but it's a few more words. And `RuleWithMeta` already exists & may be used externally...?

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts
y-hsgw: That makes sense.  
By the way, instead of adding a separate lint rule, do you think we could change the `defineRule` logic itself so that it automatically keeps the `name` consistent, like in the following example?


https://github.com/typescript-eslint/typescript-eslint/blob/c1980522cd11f2de1a49ff6a30b4be7765a843ff/packages/rule-tester/src/RuleTester.ts#L447-L448

```ts
  defineRule(rule: AnyRuleModule): void {
    this.#rules[rule.name] = {

packages/utils/src/eslint-utils/RuleCreator.ts
y-hsgw: Thanks for the feedback 😃  
I've updated the RuleWithMeta type to include the name property. c77d386

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts
JoshuaKGoldberg: Oh, that's an interesting idea. I like it. But I think it'd be a behavioral change that differs from the ESLint core `RuleTester`. We're restricted from doing those too much. It'll have to be something we talk about separately 😞.

JoshuaKGoldberg (approved):
_(wrong button)_

y-hsgw:
Thanks for the review! I’ve updated the description ✅

JoshuaKGoldberg:
Swell! We might have docs work to do later on to more prominently mention this. But for now I think "soft launching" with just the types change is best. 🚀

PR #11651, opened 2025-09-27
chore: convert eslint-plugin-internal to ESM

kirkwaiblinger:
## PR Checklist

### DIFF
--- a/packages/eslint-plugin-internal/src/index.ts
+++ b/packages/eslint-plugin-internal/src/index.ts
@@ -1,17 +1,12 @@
 import type { Linter } from '@typescript-eslint/utils/ts-eslint';
 
-import rules from './rules';
+import rules from './rules/index.js';
+import packageJson from './util/packagejson.js';
 
-// note - cannot migrate this to an import statement because it will make TSC copy the package.json to the dist folder
-const { name, version } = require('../package.json') as {
-  name: string;
-  version: string;
-};
-
-export = {
+export default {
   meta: {
-    name,
-    version,
+    name: packageJson.name,
+    version: packageJson.version,
   },
   rules,
 } satisfies Linter.Plugin;
--- a/packages/eslint-plugin-internal/src/rules/index.ts
+++ b/packages/eslint-plugin-internal/src/rules/index.ts
@@ -1,13 +1,13 @@
 import type { Linter } from '@typescript-eslint/utils/ts-eslint';
 
-import debugNamespace from './debug-namespace';
-import eqeqNullish from './eqeq-nullish';
-import noPoorlyTypedTsProps from './no-poorly-typed-ts-props';
-import noRelativePathsToInternalPackages from './no-relative-paths-to-internal-packages';
-import noTypescriptDefaultImport from './no-typescript-default-import';
-import noTypescriptEstreeImport from './no-typescript-estree-import';
-import pluginTestFormatting from './plugin-test-formatting';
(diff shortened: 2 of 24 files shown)


JoshuaKGoldberg (approved):
Swell. 👌

packages/eslint-plugin-internal/src/rules/index.ts
```diff
@@ -1,13 +1,13 @@
...
-import pluginTestFormatting from './plugin-test-formatting';
-import preferASTTypesEnum from './prefer-ast-types-enum';
+import debugNamespace from './debug-namespace.js';
+import eqeqNullish from './eqeq-nullish.js';
+import noPoorlyTypedTsProps from './no-poorly-typed-ts-props.js';
+import noRelativePathsToInternalPackages from './no-relative-paths-to-internal-packages.js';
+import noTypescriptDefaultImport from './no-typescript-default-import.js';
+import noTypescriptEstreeImport from './no-typescript-estree-import.js';
+import pluginTestFormatting from './plugin-test-formatting.js';
+import preferASTTypesEnum from './prefer-ast-types-enum.js';
```
fisker, line 10: Better make this work

```suggestion
export {default as 'prefer-ast-types-enum'} from './prefer-ast-types-enum.js';
```

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/eb97068dc85210dd8cc5c0abd7aab2e0597582e6/rules/index.js#L1

kirkwaiblinger: I guess? 

If I understand, the suggested change would be to export a bunch of named exports, and then import them as `import * as rules`, and then directly stick the `rules` null-prototype object into the exported plugin?

Seems more straightforward in my mind to keep it the way it is 🤷‍♂️

fisker: > If I understand, the suggested change would be to export a bunch of named exports, and then import them as import * as rules

Yes.

> Seems more straightforward in my mind to keep it the way it is 🤷‍♂️

The benefit of re-export way is you don't need a temporary variable name.

packages/eslint-plugin-internal/src/index.ts
```diff
@@ -1,17 +1,12 @@
 import type { Linter } from '@typescript-eslint/utils/ts-eslint';
 
-import rules from './rules';
+import rules from './rules/index.js';
+import packageJson from './util/packagejson.js';
```
fisker, line 4: Can't we use `with {type: "json"}`?

kirkwaiblinger: Could, but runs into the same issue as the code it replaces https://github.com/typescript-eslint/typescript-eslint/pull/11651/files#diff-7277e31556b0d6fb9ca81f5033df07640ddc6f3dde373d58757a1974702766faL5

packages/eslint-plugin-internal/src/rules/index.ts
kirkwaiblinger: I mean, true, but I'm not swayed that it's worth changing 😅 Seems slightly tedious but about even whichever way you slice it 🤷 

If someone else has an opinion I can change it though

JoshuaKGoldberg: I'll go ahead and merge as-is to keep the diff smaller.

PR #11659, opened 2025-09-29
feat(eslint-plugin): [no-unsafe-member-access] add allowOptionalChaining option

### LINKED ISSUE #2728: [no-unsafe-member-access] Add exception for optional chaining
- [X] I have [read the FAQ](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/FAQ.md) and my problem is not listed.

**Repro**

```JSON
{
  "rules": {
    "@typescript-eslint/no-unsafe-member-access": ["error"]
  }
}
```

```TS
const handleEvent = (event: { detail: any }): void => {
	const uuid: unknown = event.detail?.uuid;
	if (typeof uuid === "string") {
		console.log(uuid);
	}
};
```

```JSON
{
	"compilerOptions": {
		"allowSyntheticDefaultImports": true,
		"forceConsistentCasingInFileNames": true,
		"lib": [
			"es5",
			"es2015.promise",
			"es2015.iterable",
			"dom",
			"dom.iterable"
		],
		"incremental": true,
		"module": "esnext",
		"moduleResolution": "node",
		"jsx": "react",
		"noEmitOnError": true,
		"noFallthroughCasesInSwitch": true,
		"noImplicitReturns": true,
		"noUnusedLocals": true,
		"noUnusedParameters": true,
		"resolveJsonModule": true,
		"esModuleInterop": true,
		"sourceMap": false,
		"strict": true,
		"target": "es5"
	}
}
```

**Expected Result**

I'd like to not have an error when accessing `event.detail?.uuid`, either by default or by special rule config, because I'm not seeing what is unsafe about this member access. I've already satisfied the `no-unsafe-assignment` rule by declaring `uuid` to be of type `unknown`, and with optional chaining, it should not be possible to get a runtime error here, no matter what the value of `event.detail` actually is.

**Actual Result**

The `.uuid` access yielded:

```
error  Unsafe member access .uuid on an any value         @typescript-eslint/no-unsafe-member-access
```

**Versions**

| `@typescript-eslint/eslint-plugin` | `4.5.0` |
| `@typescript-eslint/parser`        | `4.5.0` |
| `ESLint`                           | `7.11.0` |
| `node`                             | `14.12.0` |

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2728

## Overview

This was a little tricker than I'd hoped. The AST positions `ChainExpression`s as the parent of all `MemberExpression`s in the chain, rather than inline in the stack. This PR modifies the node checker to have a third `State`, `State.Chained`, to save for nodes.

This results in an edge case where if you have an optional property in the middle of a stack of `any` `MemberExpression`s (`value.outer?.middle.inner`), the rule reports twice. I don't think this is a very common use case that we need to optimize for.

💖


packages/eslint-plugin/docs/rules/no-unsafe-member-access.mdx
```diff
@@ -65,6 +65,43 @@ arr[idx++];
...
+outer?.inner;
+outer?.middle?.inner;
+```
+
+</TabItem>
+</Tabs>
+
+:::caution
+We only recommend using `allowOptionalChaining` to help transition an existing project towards fully enabling `no-unsafe-member-access`.
+Optional chaining makes is more safe than normal property accesses in that you won't get a runtime error if the parent value is `null` or `undefined`.
```
ronami: Nit:

```suggestion
Optional chaining makes it safer than normal property accesses in that you won't get a runtime error if the parent value is `null` or `undefined`.
```

packages/eslint-plugin/src/rules/no-unsafe-member-access.ts
```diff
@@ -41,10 +53,25 @@ export default createRule({
...
     },
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        additionalProperties: false,
+        properties: {
+          allowOptionalChaining: {
+            type: 'boolean',
+            description: 'Whether to allow `?.` optional chains on any values.',
```
ronami: Nit: Matching the style of other descriptions, for example: https://github.com/typescript-eslint/typescript-eslint/blob/db09a7d5e551964b52b14cdb7bcd08ab094ef2b0/packages/eslint-plugin/src/rules/no-unsafe-member-access.ts#L30

```suggestion
            description: 'Whether to allow `?.` optional chains on `any` values.',
```

ronami (dismissed):
🚀

packages/eslint-plugin/src/rules/no-unsafe-member-access.ts
```diff
@@ -114,6 +153,13 @@ export default createRule({
       'MemberExpression[computed = true] > *.property'(
         node: TSESTree.Expression,
       ): void {
+        if (
+          allowOptionalChaining &&
+          (node.parent as TSESTree.MemberExpression).optional
```
ronami, line 159: Tests: This line can be commented out without a test failing

JoshuaKGoldberg: Nice! 7b9caae9c

Issue #11662, opened 2025-09-29
Bug: (regression) `name` should be optional in RuleCreator.withoutDocs


kirkwaiblinger:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.
- [x] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Abug) and found none that matched my issue.
- [x] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Relevant Package

utils

### Repro Code

```TypeScript
import { RuleCreator } from "@typescript-eslint/utils/eslint-utils";

RuleCreator.withoutDocs({
  meta: {
    messages: {},
    schema: [],
    type: 'problem',
  },
  defaultOptions: [],
  create() { return {}; }
});
```

### ESLint Config

```javascript
n/a
```

### tsconfig

```jsonc
n/a
```

### Expected Result

I expect there to be no type error

### Actual Result

type error because `name` is now required

```
Argument of type '{ meta: { messages: {}; schema: never[]; type: "problem"; }; defaultOptions: []; create(context: Readonly<RuleContext<never, []>>): {}; }' is not assignable to parameter of type 'Readonly<RuleWithMeta<[], never, unknown>>'.
  Property 'name' is missing in type '{ meta: { messages: {}; schema: never[]; type: "problem"; }; defaultOptions: []; create(context: Readonly<RuleContext<never, []>>): {}; }' but required in type 'Readonly<RuleWithMeta<[], never, unknown>>'.ts(2345)
RuleCreator.d.ts(12, 5): 'name' is declared here.
```

### Additional Info

`name` used to be optional but was made required in https://github.com/typescript-eslint/typescript-eslint/pull/11616 (breaking)

cc @y-hsgw 

### Versions

| package                                | version |
| -------------------------------------- | ------- |
| `@typescript-eslint/utils`             | `8.45.0` |

kirkwaiblinger:
Curiously, it seems like `name` is actually not an eslint core property. https://eslint.org/docs/latest/extend/custom-rules

It looks like the only meaning of `name` is to insert the rule name into the documentation with 
```ts
import { RuleCreator } from '@typescript-eslint/utils/eslint-utils';

const myPluginRuleCreator = RuleCreator(
[...]

PR #11663, opened 2025-09-29
Revert "feat(eslint-plugin): expose rule name via RuleModule interface"

JoshuaKGoldberg:
Reverts typescript-eslint/typescript-eslint#11616

I'd like to get the change back in, but it's non-trivial to work around #11662.

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-eslint.test.ts
@@ -40,7 +40,6 @@ ruleTester.defineRule('use-every-a', {
     schema: [],
     type: 'problem',
   },
-  name: 'use-every-a',
 });
 
 /**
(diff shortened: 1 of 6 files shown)


kirkwaiblinger (approved):
I don't remember the details, but I remember previously nx was super picky about the PR title for a revert PR, so might be worth double checking that before merging

jfirebaugh:
This is a breaking change, no? `RuleModule` is a public type, so code written against v8.45.0 such as:

```
import { TSESLint, TSESTree } from '@typescript-eslint/utils'
const rule: TSESLint.RuleModule<'foo', []> = {
  name: 'my-rule',
  ...
}
```

Now fails to typecheck with v8.46.0+.

bradzacher:
It was added in one version and immediately removed in the next version because it introduced bugs. From our POV that is NOT a breaking change - it should not have landed at all.

kirkwaiblinger:
@jfirebaugh FWIW we were intending to do an immediate patch release to revert it, but ran into CI/automation issues, so it ended up just going out as part of the next regularly-scheduled minor, rather than its own patch.

jfirebaugh:
I see... I guess we upgraded at exactly the wrong time, such that we had to add `name` properties to keep things building with 8.45.0 and then remove them again with 8.46.0.

PR #11682, opened 2025-10-05
feat(eslint-plugin): [unbound-method] report property signatures with function types

### LINKED ISSUE #3503: [unbound-method] Should prevent conforming to an bound-style interface method with an unbound method

ronami:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3503

## Overview

Waiting for resolution on https://github.com/typescript-eslint/typescript-eslint/issues/3503#issuecomment-3369321736 before opening for review.

### DIFF
--- a/packages/eslint-plugin/tests/rules/unbound-method.test.ts
+++ b/packages/eslint-plugin/tests/rules/unbound-method.test.ts
@@ -463,7 +463,11 @@ class Foo {
 type foo = new ({ unbound }: Foo) => void;
     `,
     'const { unbound } = { unbound: () => {} };',
-    'function foo({ unbound }: { unbound: () => void } = { unbound: () => {} }) {}',
+    `
+function foo(
+  { unbound }: { unbound: (this: void) => void } = { unbound: () => {} },
+) {}
+    `,
     // https://github.com/typescript-eslint/typescript-eslint/issues/1866
     `
 class BaseClass {
@@ -479,6 +483,30 @@ class OtherClass extends BaseClass {
 }
 const oc = new OtherClass();
 oc.superLogThis();
+    `,
+    `
+interface Foo {
+  bound: (this: void) => number;
+}
+
+declare const foo: Foo;
+foo.bound;
+    `,
+    `
+interface Foo {
+  bound: (this: void) => number;
+}
+
+declare const foo: Foo;
+const { bound } = foo;
+    `,
+    `
+type Foo = {
+  bound: (this: void) => number;
+};
+
+declare const foo: Foo;
+const { bound } = foo;
     `,
   ],
   invalid: [
@@ -965,7 +993,7 @@ function foo({ unbound }: { unbound: () => string } | Foo) {}
       errors: [
         {
           line: 5,
-          messageId: 'unbound',
+          messageId: 'unboundWithoutThisAnnotation',
         },
       ],
     },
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/tests/rules/unbound-method.test.ts
```diff
@@ -463,7 +463,11 @@ class Foo {
 type foo = new ({ unbound }: Foo) => void;
     `,
     'const { unbound } = { unbound: () => {} };',
-    'function foo({ unbound }: { unbound: () => void } = { unbound: () => {} }) {}',
+    `
+function foo(
+  { unbound }: { unbound: (this: void) => void } = { unbound: () => {} },
```
ronami, line 468: To my understanding, this previously valid test is in fact invalid without a `this: void` parameter type.

packages/eslint-plugin/tests/rules/unbound-method.test.ts
```diff
@@ -965,7 +993,7 @@ function foo({ unbound }: { unbound: () => string } | Foo) {}
       errors: [
         {
           line: 5,
-          messageId: 'unbound',
+          messageId: 'unboundWithoutThisAnnotation',
```
ronami, line 996: This is interesting, it seems the rule iterates on union types and reports the first violation: https://github.com/typescript-eslint/typescript-eslint/blob/7f5fed7922385479613f4d79a28615014174a556/packages/eslint-plugin/src/rules/unbound-method.ts#L263-L275

In this case, the union starts with `{ unbound: () => string }`, with `Foo` being second. Currently, `{ unbound: () => string }` doesn't report, which causes the report to come from `Foo`.

This PR made it so `{ unbound: () => string }` does trigger a report, which changes the error message accordingly.

ronami: Somewhat related: https://github.com/typescript-eslint/typescript-eslint/issues/11683.

JoshuaKGoldberg:
👋 @ronami are you waiting on anything for this PR?

JoshuaKGoldberg:
Closing out as an old draft. But we can always re-open!

PR #11693, opened 2025-10-11
feat(typescript-estree): gate all errors behind allowInvalidAST

fisker:
## PR Checklist

## Overview

Currently, these errors are all syntax errors, and having two methods is confusing. Let's put all these errors behind `allowInvalidAST` option.

### DIFF
--- a/packages/typescript-estree/src/convert.ts
+++ b/packages/typescript-estree/src/convert.ts
@@ -159,7 +159,13 @@ export class Converter {
     checkModifiers(node);
   }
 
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (approved):
If I understand this, it's actually a pretty big change for the conversion logic, right? If so - that makes sense to me, and I think it's actually quite nice to put all the syntax errors behind the _"should the converter throw on invalid syntax?"_ option. But someone else from @typescript-eslint/triage-team should probably gut check this too.

PR #11702, opened 2025-10-16
fix(eslint-plugin): [prefer-optional-chain] skip optional chaining when it could change the result

### LINKED ISSUE #11700: Bug: @typescript-eslint/prefer-optional-chain flags situations where an optional chain would change behavior

### Playground Link

https://typescript-eslint.io/play/#ts=5.9.3&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3lAXPA3vAEYD8emMAligObwC%2BA3AFDOUBm8AFAIRTwAfAdgB0ReAF4p8AETskSGQEoCzePGb0gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uigB3UdCaRwYAL4gNQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
declare const a: { b?: string };

if (!a || a.b === "foo") {
  
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/prefer-optional-chain": ["error", {}],
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {}
}
```

### Expected Result

I expect that no error from `@typescript-eslint/prefer-optional-chain` is triggered on the third line, as an optional chain would change behavior. 

### Actual Result

An error from `@typescript-eslint/prefer-optional-chain` is triggered on the third line. When autofixed, this produces the code
```ts
declare const a: { b?: string };

if (a?.b === "foo") {
  
}
```
When `a.b` is undefined, the original comparison produced the value `true`, executing the code within the `if`-block. The 'autofixed' code produces `false` in that case, skipping the code within the `if`-block.

### Additional Info

This seems due to the changes in https://github.com/typescript-eslint/typescript-eslint/pull/11533. The comment in https://github.com/typescript-eslint/typescript-eslint/pull/11533#issuecomment-3400575571 already hints at this incorrect behavior.

mdm317:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11700

When applying optional chaining, if the transformation has the potential to alter the evaluation result,
the previous operand should not be included, even when it appears useless.

Conversely, if the transformation is guaranteed to preserve the result,
the previous operand should be included in the chain, even when it’s useless.

## Not convert to optional-chain
```
declare const a: { b?: string };

a && a.b != "4"
a == null || a.b === "foo"
...more cases
```
## Conver to optional-chain
```
declare const a: { b?: string };

a && a.b != null
a && a.b != undefined
a && a.b  !== undefined
...more cases
```
[see Algorithm](https://github.com/typescript-eslint/typescript-eslint/pull/11533#issue-3359701002)

### DIFF
--- a/packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
+++ b/packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
@@ -1778,20 +1778,20 @@ describe('chain ending with comparison', () => {
       'foo == null && foo.bar != null;',
       'foo == null && foo.bar != undefined;',
       `
-        declare const x: false | { a: string };
-        x && x.a == x;
+        declare const foo: false | { a: string };
+        foo && foo.a == undeclaredVar;
       `,
       `
-        declare const x: '' | { a: string };
-        x && x.a == x;
+        declare const foo: '' | { a: string };
+        foo && foo.a == undeclaredVar;
       `,
       `
-        declare const x: 0 | { a: string };
-        x && x.a == x;
+        declare const foo: 0 | { a: string };
+        foo && foo.a == undeclaredVar;
       `,
       `
-        declare const x: 0n | { a: string };
-        x && x.a;
+        declare const foo: 0n | { a: string };
+        foo && foo.a;
       `,
       '!foo || foo.bar != undeclaredVar;',
       '!foo || foo.bar != null;',


packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1938,6 +1387,174 @@ describe('chain ending with comparison', () => {
...
+        declare const foo: { bar: number };
+        foo != null && foo.bar !== true;
+      `,
+      `
+        declare const foo: { bar: number };
+        foo != null && foo.bar !== null;
+      `,
+      `
+        declare const foo: { bar: number };
+        foo != null && foo.bar !== x;
```
kirkwaiblinger: Looks like `x` is undeclared here, so this is an error type. Might be good to make it explicit?

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
```diff
@@ -1938,6 +1387,174 @@ describe('chain ending with comparison', () => {
       'foo != null && foo.bar != false;',
       'foo != null && foo.bar != true;',
       'foo != null && foo.bar != x;',
+      `
+        declare const foo: { bar: number };
+        foo && foo.bar == x;
```
kirkwaiblinger: Let's declare `x` with a type here? Unless we're intentionally testing undeclared variables?

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
kirkwaiblinger, line 1: Looks like there's a minor bug somewhere with:
```ts
declare const foo: { bar: number | null } | undefined;
if (foo === undefined || foo.bar === null) { // suggests to foo?.bar === null, which is not equivalent.
} else {
  foo.bar.toExponential(); // after accepting fix, this crashes if `foo` is `undefined`.
}
```

I guess this should either not report or it could maybe try to use `foo?.bar == null`? But this seems kind of bad since it is impossible to work with for an `=== nullishValue`-prefering user

kirkwaiblinger (changes requested):
Looks like it's on the right track, but minor touchups and a question about one bug that I saw.

Thanks for looking at this so quickly! ❤️

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
mdm317: I didn’t declare it because I wanted to test that it doesn’t convert to an optional chain when we don’t know `x`’s type.
But thinking about what you said, testing with an error type feels odd and doesn’t clearly show the test’s purpose.

How about declaring x with the type `any` instead?

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
mdm317: > https://github.com/typescript-eslint/typescript-eslint/blob/3f5fbf698e75ddd87874885ffbf937913761cdb0/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts#L277-L284

I believe this bug didn’t originate from this PR — it seems to have existed beforehand.
Since I wasn’t the original author, it might take some time to fix it properly.

If possible, we could apply a quick hotfix for the current issue first and then open a new issue to track it, but it’s just a suggestion — feel free to decide as you think best.

In any case, I’ll start digging into it now.

packages/eslint-plugin/tests/rules/prefer-optional-chain/prefer-optional-chain.test.ts
kirkwaiblinger: I think both are potentially valid tests! But I would just make the intention explicit like:
```ts
declare const foo: { bar: number };
declare const anee: any;
foo && foo.bar == anee;
```

```ts
// intentionally tests that an undeclared variable doesn't trigger an autofix
declare const foo: { bar: number };
foo && foo.bar == undeclaredVar;
```

packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
kirkwaiblinger: That's fine if we defer it! 

Just out of curiosity, do you mean to say it didn't originate in _this_ PR, ie #11702, or in #11533? (either way we can defer it just good to know)

mdm317: Apologies for the late reply — I didn’t get a notification.
It seems this issue has existed since this [#6397](https://github.com/typescript-eslint/typescript-eslint/pull/6397)

I think the code below is causing the issue above.
> https://github.com/typescript-eslint/typescript-eslint/blob/dd969477e8cb10a4402eccc051a1d1206853e3b0/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts#L152-L165

I’ll create an issue for this.

PR #11704, opened 2025-10-17
fix(eslint-plugin): replace unclear "`error` typed" with more helpful description

### LINKED ISSUE #9591: Bug: [no-unsafe-call] Unclear message "Unsafe call of an `error` type typed value" (especially in JSDoc)

### Playground Link

https://typescript-eslint.io/play/#ts=5.5.2&fileType=.js&code=MYewdgzgLgBApgGxgXhgegFQZgASgTwAc4YBvAUQTgFs4woBfGDNGAClIYEoAoHxAHQALAE5wAZgICWYYAgCuAEzgQ2AcgCGAI2BreQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tieQEMAZolpk%2B8eOiiJo0DtEjgwAXxBKgA&tsconfig=&tokens=false

### Repro Code

```TypeScript
const el = /** @type {Element} */ ({})

el.href.includes('abc')
```

### ESLint Config

```javascript
module.exports = {
  rules: {
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
  }
}
```

### Expected Result

I expected the error message:

```
Unsafe call of a method on an `any` typed value
```

Differences:

- mentions the type `any` (which is the only observable type in JSDoc without `checkJs: true`)
- doesn't mention the type `error`
- less important: mentions "method"

### Actual Result

I observed the error message:

```
Unsafe call of an `error` type typed value
```

Confusing parts:

- no mention of `any`
- mentions the type `error`, which is unobservable in JSDoc without `checkJs: true`
  - also, if I hear "an `error` type typed value", then that could sound like something like `Error`, so even if this "error" language is kept in some form, maybe it should be disambiguated from `Error`
- less important: does not mention "method"

### Additional Info

The type for this seems to be coming from the `ts-api-utils` function `isIntrinsicErrorType`

- https://github.com/JoshuaKGoldberg/ts-api-utils/blob/2d6af536e6ecf9fae9a2db445dbe1db115955afa/src/types/typeGuards/intrinsic.ts#L92-L111

cc @JoshuaKGoldberg

wjhsf:
## PR Checklist

- [x] Addresses an existing open issue: fixes #9591
- [ ] ~Steps in [Contributing](https://typescript-eslint.io/contributing) were taken~

## Overview

As described in #9591, the error message "Unsafe call of a(n) `error` type typed value". Namely, what is "an `error` type typed value"? It's unclear to anyone who does not know that TypeScript has a special flavor of `any` for types that could not be resolved. (Probably just about everyone who doesn't work on TypeScript tooling.)

This PR aims to provide clarity by changing "`error` typed" to "unresolved due to error", resulting in error messages like "Unsafe call of a(n) unresolved due to error typed value". Not the most natural phrasing, and it would probably be better to use a distinct error message, but my goal was to make a minimal change and not have to go stomping through the code to figure things out. 😅


packages/eslint-plugin/src/rules/no-unsafe-call.ts
```diff
@@ -73,7 +73,7 @@ export default createRule<[], MessageIds>({
           node: reportingNode,
           messageId,
           data: {
-            type: isErrorType ? '`error` type' : '`any`',
+            type: isErrorType ? 'unresolved due to error' : '`any`',
```
JoshuaKGoldberg: > Not the most natural phrasing

🤔 yeah, I'm not sure that this specific change is better for users. Looking at [the deploy preview playground](https://deploy-preview-11704--typescript-eslint.netlify.app/play/#ts=5.9.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3gNyhAriAFzwBySGA0ikgO4oDcAUE7gSABQCUDQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tieQEMAZolpk%2B8eOiiJo0DtEjgwAXxBKgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false), the report message is:

```plaintext
Unsafe call of a(n) unresolved due to error typed value.
```

It took me a few tries to parse out the English grammar in that sentence, and I had to rely on already knowing the rule & this PR. 

I'm drawing a blank here (and am a little swamped), but let's bikeshed. What's a more clear phrase we could go with?

wjhsf: Using an entirely new message would be best. Something along the lines of "Unsafe call of a type that could not be resolved." That's what I originally started doing in this PR, but then I had to touch more files than I wanted to, so I changed to the kludgy phrasing.

JoshuaKGoldberg: FWIW I like that phrasing more. I can't think of a way to more concisely tell the user _"this type could not be resolved"_ than just saying so.

wjhsf: I've updated the rules to have sentences that humans actually use.

packages/eslint-plugin/src/rules/no-unsafe-call.ts
```diff
@@ -27,13 +31,22 @@ export default createRule<[], MessageIds>({
...
+        'Unsafe call of a `this` type that could not be resolved.',
+        'You can try to fix this by turning on the `noImplicitThis` compiler option, or adding a `this` parameter to the function.',
+      ].join('\n'),
+      errorNew: 'Unsafe construction of a type that could not be resolved.',
+      errorTemplateTag:
+        'Unsafe use of a template tag whose type could not be resolved.',
+      unsafeCall: 'Unsafe call of {{article}} {{type}} typed value.',
       unsafeCallThis: [
-        'Unsafe call of a(n) {{type}} typed value. `this` is typed as {{type}}.',
+        'Unsafe call of {{article}} {{type}} typed value. `this` is typed as {{type}}.',
```
wjhsf: `{{article}}` isn't strictly necessary. Actually, since it's always `{{article}} {{type}}`, the values for `type` could just be changed to "an \`any\`" or "a \`Function\`".

packages/eslint-plugin/src/rules/no-unsafe-call.ts
```diff
@@ -49,7 +62,8 @@ export default createRule<[], MessageIds>({
     function checkCall(
       node: TSESTree.Node,
       reportingNode: TSESTree.Node,
-      messageId: MessageIds,
+      unsafeMessageId: Extract<MessageIds, `unsafe${string}`>,
+      errorMessageId: Extract<MessageIds, `error${string}`>,
```
wjhsf, line 62: These types are most likely unnecessarily pedantic.

packages/eslint-plugin/src/rules/no-unsafe-call.ts
```diff
@@ -27,13 +31,22 @@ export default createRule<[], MessageIds>({
       requiresTypeChecking: true,
     },
     messages: {
-      unsafeCall: 'Unsafe call of a(n) {{type}} typed value.',
+      errorCall: 'Unsafe call of a type that could not be resolved.',
+      errorCallThis: [
+        'Unsafe call of a `this` type that could not be resolved.',
+        'You can try to fix this by turning on the `noImplicitThis` compiler option, or adding a `this` parameter to the function.',
```
JoshuaKGoldberg: [Bug] The problem here is that sometimes there _is_ a `this` type already, but it couldn't be resolved. I would recommend just removing this second line.

```ts
function wrongThis(this: NotInScope) {}
```

packages/eslint-plugin/src/rules/no-unsafe-call.ts
JoshuaKGoldberg: Yeah, since code is already providing `{{type}}`, +1 to just including the article in that data.

JoshuaKGoldberg (changes requested):
I like the direction this is taking, +1! I think the only blocker for me is the removing the second line of the `this` messages - WDYT?

JoshuaKGoldberg (dismissed):
🤩 This looks great to me, thank you for all the iterations! Just pinging @typescript-eslint/triage-team to see if anybody else wants to provide input.

I'll also note there are some CI failures. If you have time to clean those up, that'd be nice. But we can always step in before merging - I know this PR has had a lot of back-and-forth.

packages/eslint-plugin/src/rules/no-unsafe-call.ts
```diff
@@ -27,13 +31,20 @@ export default createRule<[], MessageIds>({
       requiresTypeChecking: true,
     },
     messages: {
-      unsafeCall: 'Unsafe call of a(n) {{type}} typed value.',
+      errorCall: 'Unsafe call of a type that could not be resolved.',
+      errorCallThis: [
+        'Unsafe call of a `this` type that could not be resolved.',
+      ].join('\n'),
```
bradzacher:
```suggestion
      errorCallThis: 'Unsafe call of a `this` type that could not be resolved.',
```

nit unnecessary array

bradzacher (approved):

PR #11715, opened 2025-10-23
fix(scope-manager): reconcile type-value distinction for implicit globals across libs

### LINKED ISSUE #11714: Scope Manager fails to pick up references to implicit globals

### Relevant Package

scope-manager

### Playground Link

_No response_

### Repro Code

```TypeScript
undefined; globalThis; NaN; Object; Boolean; String; Math; Date; Array; Map; Set; var foo; foo;
```

### ESLint Config

```javascript
Not relevant
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "lib": ["esnext"]
  }
}
```

### Expected Result

All identifiers but `foo` are analyzed to be determined references to implicit global variables.

### Actual Result

Only ~~`String`~~ `Boolean` and `Object` are determined to be references to implicit global variables.

### Additional Info

- I encountered this while working on https://github.com/oxc-project/oxc/pull/14890.
### What's going wrong?
- The generated code for ts libs may include multiple mentions of the same variable, in some cases indicating the variable is a [`TYPE`](https://github.com/typescript-eslint/typescript-eslint/blob/55ca033ee88cc95cfac4ad6dea2257fbeb1d4657/packages/scope-manager/src/lib/base-config.ts#L16), and `TYPE_VALUE` in others.
- When `Referencer.populateGlobalsFromLib` calls`GlobalScope.defineImplicitVariable` which calls `ScopeBase.defineVariable` and the variable is already defined, it doesn't entertain the possibility that the variable may have different type/value attributes and simply moves on.
- The result: `@typescript-eslint/scope-manager` incorrectly indicates to the user project that a variable is a type, and the scope analysis does not pick up value identifiers as references to the variable.

### Versions

| `@typescript-eslint/scope-manager`     | `8.46.1'` |
| `TypeScript`                           | `5.9.3` |
| `node`                                 | `24.10.0` |

lilnasy:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11714

## Overview

To fix the issue, implicit variables are mutated as the global scope is being populated.

I admittedly did not spend much time thinking of a proper solution. Though it was tricky to pinpoint, the fix is simple. I imagine the "correct" location for the fix is a matter of maintainer preference.

Direct commits from maintainers welcome.

### DIFF
--- a/packages/scope-manager/src/scope/ScopeBase.ts
+++ b/packages/scope-manager/src/scope/ScopeBase.ts
@@ -326,6 +326,11 @@ export abstract class ScopeBase<
     }
     // variable exists on the scope
 
+    // implicit lib variables should always be resolved statically
+    if (variable instanceof ImplicitLibVariable) {
+      return true;
+    }
+
     // in module mode, we can statically resolve everything, regardless of its decl type
     if (scopeManager.isModule()) {
       return true;
@@ -391,10 +396,11 @@ export abstract class ScopeBase<
           : nameOrVariable;
       set.set(name, variable);
       variables.push(variable);
-    } else if (variable instanceof ImplicitLibVariable && nameOrVariable instanceof ImplicitLibVariable) {
-      // @ts-expect-error
+    } else if (
+      variable instanceof ImplicitLibVariable &&
+      nameOrVariable instanceof ImplicitLibVariable
+    ) {
       variable.isTypeVariable ||= nameOrVariable.isTypeVariable;
-      // @ts-expect-error
       variable.isValueVariable ||= nameOrVariable.isValueVariable;
     }
 
--- a/packages/scope-manager/src/variable/ImplicitLibVariable.ts
+++ b/packages/scope-manager/src/variable/ImplicitLibVariable.ts
@@ -25,12 +25,12 @@ export class ImplicitLibVariable
   /**
    * `true` if the variable is valid in a type context, false otherwise
    */
-  public readonly isTypeVariable: boolean;
+  public isTypeVariable: boolean;
 
   /**
    * `true` if the variable is valid in a value context, false otherwise
    */
-  public readonly isValueVariable: boolean;
+  public isValueVariable: boolean;
 
   public constructor(
     scope: Scope,


lilnasy:
@bradzacher I'm thinking it would be even better to go one level higher.

The first year-wise occurrence of the globals I glimpsed at seems to always be `TYPE_VALUE`. If that holds true for all, we could simply ensure the iteration is in chronological order at runtime and it would work. The test could be made more comprehensive to prevent regressions.

lilnasy:
I have a branch where I [flatten](https://github.com/typescript-eslint/typescript-eslint/blob/843f144797c0a94272cdb002c00c5639cf0797c6/packages/scope-manager/src/referencer/Referencer.ts#L64) the lib references at codegen time instead of at runtime, fixing the bug right there. It increases the package size a little but reduces `ScopeManager` instantiation time.

I can move forward with whichever approach maintainers prefer.

bradzacher:
I believe the behaviour that we will need to add as part of the eslint v10 upgrades will provide you the behaviour you want.

This behaviour will resolve all references to globals as part of the analysis rather than leaving them all open.

JoshuaKGoldberg:
@bradzacher so just confirming, is there any change we should make here? Is #11714 a no-op/wontfix given the upcoming ESLint v10 changes?

Linking for visibility: #11762 tracks updating scope-manager for compatibility with ESLint v10.

lilnasy:
> I believe the behaviour that we will need to add as part of the eslint v10 upgrades will provide you the behaviour you want.

I agree.

> Is https://github.com/typescript-eslint/typescript-eslint/issues/11714 a no-op/wontfix given the upcoming ESLint v10 changes?

@JoshuaKGoldberg  I think that depends on what you meant in the comment here: https://github.com/typescript-eslint/typescript-eslint/issues/11762#issuecomment-3543053307. Could you clarify? Can `@typescript-eslint/scope-manager` remain compatible with both ESLint v8 and v10?

JoshuaKGoldberg:
We haven't looked deeply into it yet, but I really hope so. Dropping support for v8 and v9 would be a breaking change that we're not ready to release a new major version for anytime soon.

bradzacher:
Well technically by being compatible with v10 we're also compatible with v8/v9 - cos eslint will just have nothing to do.

lilnasy:
Thanks, I thought that might be the case. Fixing this as part of ESLint 10 compat seems appropriate. Oxlint will be able to remove the patch we made to `@typescript-eslint/scope-manager` then (if it happens before our rust reimplementation).

PR #11719, opened 2025-10-25
feat(eslint-plugin): expose rule name via RuleModule interface

### LINKED ISSUE #11608: [feature request] Expose the rule name as a property

y-hsgw:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11608

## Overview

The previous PR [#11616](https://github.com/typescript-eslint/typescript-eslint/pull/11616) was reverted after a discussion in [#11662](https://github.com/typescript-eslint/typescript-eslint/issues/11662) pointing out that the `name` argument in `withoutDocs` should be optional.
This PR reintroduces the change from #11616 while keeping name optional.

🍂


y-hsgw:
I’ve added tests. [91e97bc](https://github.com/typescript-eslint/typescript-eslint/pull/11719/commits/91e97bc63a186b3cdd1b6560dbf2e6ee40eded8e)

packages/utils/src/eslint-utils/RuleCreator.ts
```diff
@@ -101,6 +104,7 @@ function createRule<
     },
     defaultOptions,
     meta,
+    name: name ?? '',
```
JoshuaKGoldberg: 🤔 I don't think this is right. `''` is kind of an odd default for the field. I think we'd want the type system know that `name` is `undefined` if it wasn't provided.

y-hsgw: Got it—I've fixed in 213983b.

packages/rule-tester/src/RuleTester.ts
```diff
@@ -444,9 +444,10 @@ export class RuleTester extends TestFramework {
     return normalizedTests;
   }
 
-  defineRule(name: string, rule: AnyRuleModule): void {
+  defineRule(name: string, rule: Omit<AnyRuleModule, 'name'>): void {
```
kirkwaiblinger: I don't think this really matters since it gets overwritten anyways 
```suggestion
  defineRule(name: string, rule: AnyRuleModule): void {
```

packages/rule-tester/tests/RuleTester.test.ts
kirkwaiblinger, line 1: Think these can be removed since the `name` isn't used or required

packages/utils/tests/eslint-utils/RuleCreator.test.ts
```diff
@@ -42,5 +42,55 @@ describe(ESLintUtils.RuleCreator, () => {
       schema: [],
       type: 'problem',
     });
+    expect(rule.name).toBe('test');
```
kirkwaiblinger, line 45: (nice-to-have, not a blocker)

I think it would be neat if we could also make sure that `rule.name satisfies string` rather than being possibly-undefined, since it's required now for rules created by `createRule()`.

kirkwaiblinger (dismissed):

packages/utils/tests/eslint-utils/RuleCreator.test.ts
y-hsgw: Hey, I tried making `rule.name` a required string type in 083f6b5. What do you think of that?

packages/utils/src/eslint-utils/RuleCreator.ts
```diff
@@ -67,17 +77,23 @@ export function RuleCreator<PluginDocs = unknown>(
...
         },
       },
+      name,
       ...rule,
     });
+
+    return {
+      ...ruleWithDocs,
+      name,
+    };
```
kirkwaiblinger: Seems like overkill to make another runtime change here just to satisfy the checker.

Maybe we just use an assertion here?

```suggestion
    return ruleWithDocs as typeof ruleWithDocs & { name: string };
```

kirkwaiblinger (dismissed):

packages/utils/src/eslint-utils/RuleCreator.ts
y-hsgw: Fixed it in 28af1c9.

kirkwaiblinger (approved):

kirkwaiblinger (approved):

PR #11735, opened 2025-11-03
feat(eslint-plugin): [no-unnecessary-condition] detect unnecessary `Array.isArray` calls

### LINKED ISSUE #11716: Enhancement: [no-unnecessary-condition] detect unnecessary calls to Array.isArray

https://typescript-eslint.io/rules/no-unnecessary-condition

### Description

I constantly see LLMs generating code that checks variables by calling Array.isArray on them despite them being strictly typed as array. Adding logic into `no-unnecessary-condition` to detect these redundant calls would reduce code by avoiding unnecessary conditions and helps in reviewing AI generated code.

### Fail

```TypeScript
const items: number[] = [1, 2, 3];
if (Array.isArray(items)) {
  console.log("items is an array");
}
```

### Pass

```TypeScript
const items: number[] | null = [1, 2, 3];
if (Array.isArray(items)) {
  console.log("items is an array");
}
```

### Additional Info

Maybe also include logic to detect unnecessary `items instanceof Array`?

azat-io:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11716

## Overview

Adds detection of redundant Array.isArray() calls when checkTypePredicates is enabled.

Reports error when the argument is already typed as an array (including readonly arrays).

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -288,6 +288,71 @@ export default createRule<Options, MessageId>({
         .some(part => checker.isTupleType(part));
     }
 
+    function typeIsExclusivelyArrayOrTuple(type: ts.Type): boolean {
+      return tsutils
+        .unionConstituents(type)
+        .every(part => checker.isArrayType(part) || checker.isTupleType(part));
+    }
+
+    function getDeclaredTypesOfExpression(
+      node: TSESTree.Expression,
+    ): ts.Type[] {
+      const tsNode = services.esTreeNodeToTSNodeMap.get(node);
+      const symbol = checker.getSymbolAtLocation(tsNode);
+
+      if (!symbol) {
+        return [];
+      }
+
+      const declarations = symbol.getDeclarations();
+
+      if (!declarations || declarations.length === 0) {
+        return [];
+      }
+
+      function getTypeNodeFromDeclaration(
+        declaration: ts.Declaration,
+      ): ts.TypeNode | undefined {
+        if (
+          ts.isParameter(declaration) ||
+          ts.isPropertySignature(declaration) ||
+          ts.isPropertyDeclaration(declaration) ||
+          ts.isVariableDeclaration(declaration)
+        ) {
+          return declaration.type ?? undefined;
+        }
+
+        if (ts.isBindingElement(declaration)) {
+          const bindingPattern = declaration.parent;
+          const parentDeclaration = bindingPattern.parent;
+
+          if (
+            ts.isVariableDeclaration(parentDeclaration) ||
+            ts.isParameter(parentDeclaration) ||
+            ts.isPropertyDeclaration(parentDeclaration) ||
+            ts.isPropertySignature(parentDeclaration)
+          ) {
+            return parentDeclaration.type ?? undefined;
+          }
+        }
+
+        return undefined;
+      }
+
+      const types: ts.Type[] = [];
+
+      for (const declaration of declarations) {
+        const typeNode = getTypeNodeFromDeclaration(declaration);
+        if (!typeNode) {
+          continue;
+        }
+
+        types.push(checker.getTypeFromTypeNode(typeNode));
+      }
+
+      return types;
+    }
+
     function isArrayIndexExpression(node: TSESTree.Expression): boolean {
       return (
         // Is an index signature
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
```diff
@@ -1138,6 +1138,55 @@ isString('falafel');
       `,
       options: [{ checkTypePredicates: true }],
     },
+
+    {
+      code: `
+const items: number[] | null = [1, 2, 3];
```
kirkwaiblinger: You'll want to do it like this, I think, here and elsewhere:
```suggestion
declare const items: number[] | null;
```
TS ignores the `| null` because it can see that it's not null ([playgroun](https://typescript-eslint.io/play/#ts=5.9.3&fileType=.ts&code=MYewdgzgLgBAllApgWwgLhmArsgRogJwG0BdGAH0ywBtqYBeGIgRgBoYAmdgZhIG4AUAhQRBAejEwAegH4hAMxgAKAIIECAQwCeAOjgQ1mrUuGoAlGZgBvATBgS7jp87uzbMUJBDVEO6iABzJQByUwh4cI0wGA11bWCzQQBfIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6AewFtLlFaAM1hMy%2BSpya0AhskKj0URNC7RI4MAF8QmoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false))

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
kirkwaiblinger, line 1: If I replace this entire diff with 

```diff
-          if (typeOfArgument === typeGuardAssertedArgument.type) {
+          if (
+            checker.isTypeAssignableTo(
+              typeOfArgument,
+              typeGuardAssertedArgument.type,
+            )
+          ) {
```

(which is what was intended in https://github.com/typescript-eslint/typescript-eslint/issues/11716#issuecomment-3439853329), the only test case that doesn't pass (apart from spurious reasons) is 

```ts
function processArray(arr: readonly string[]) {
  if (Array.isArray(arr)) {
    return arr.length;
  }
}
```

So, I'm still thinking that changing to the assignability API makes sense, at least as a first step. In other words, let's take on the assignability API change before worrying about the edge case of `readonly` arrays with `Array.isArray()` (which, FWIW, I think is basically a corollary of https://github.com/microsoft/TypeScript/issues/17002). What do you think?

Also cc @ronami

azat-io: Done! Changed to use `checker.isTypeAssignableTo()` for Array.isArray

packages/eslint-plugin/tests/rules/no-unnecessary-condition.test.ts
azat-io: Done ✅

packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
kirkwaiblinger: @azat-io almost! I really mean the entire diff, though, not just the Array.isArray case. The thought is to change all unnecessary type predicate detection to be based on assignability. 

Note that this change would require accompanying report message changes and additional testing

azat-io: Done! Changed to use `isTypeAssignableTo()` for all type predicates. Removed `isArrayIsArrayCall()` function and updated the tests accordingly.

kirkwaiblinger (changes requested):
I don't have time to do a fully detailed review right now, but jotting down the notes that I'll want to come back to. Feel free to look into these in the meantime:

1. We'll probably need to update the text of the typeGuardAlreadyIsType message. 
1. The `getConstrainedTypeAtLocation` may actually be wrong/unnecessary now. Needs tests with generics to prove one way or another.
1. Like with no-unsafe-type-assertion, we should double-check and add testing around object literals and excess property checking.
1. Needs further tests around subtype cases.

azat-io:
Dialed the change back to focus solely on `Array.isArray`: the rule still reports when the argument type exactly matches the predicate type, and now additionally reports redundant Array.isArray checks when the argument is assignable to the guard’s array type (excluding `any`/`unknown`). Added valid/invalid coverage for unions, generics, and tuples so the new path stays constrained.

kirkwaiblinger:
Hi, so I see in the latest changeset you changed the PR back to special-case logic for `Array.isArray()`. Per https://github.com/typescript-eslint/typescript-eslint/issues/11716#issuecomment-3439853329, I'd like to first see if we can't get `Array.isArray()` for free as part of a more general logic change around unnecessary type predicates where the asserted argument is a subtype of the asserted type... For better clarity, I've split that out explicitly now into https://github.com/typescript-eslint/typescript-eslint/issues/11798. 

Therefore, unfortunately, I'd like to consider this work blocked for now until we can resolve #11798 and then we can revisit what, if anything, remains in #11716. I appreciate your understanding 🙏

kirkwaiblinger:
Superseded by https://github.com/typescript-eslint/typescript-eslint/pull/12147

PR #11748, opened 2025-11-09
chore: extract AST check from convert.ts to ast-checks.ts

### LINKED ISSUE #11695: Repo: Extract AST check from `Converter`
### Suggestion

The [`Converter`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-estree/src/convert.ts) is getting huge, I think it's possible to split the AST check from it.

### Additional Info

_No response_

Lonercode:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11695

## Overview

This PR extracts the AST check (checkForStatementDeclaration) from `convert.ts` to a file `ast-checks.ts` as requested in the issue due to the increasing length of code in the file. Therefore, a new fie has been added (`ast-checks.ts`), one modified (`convert.ts`) and tests have been run.

### DIFF
--- a/packages/typescript-estree/src/check-syntax-errors.ts
+++ b/packages/typescript-estree/src/check-syntax-errors.ts
@@ -1,17 +1,21 @@
 import * as ts from 'typescript';
 
+import type { TSNode } from './ts-estree';
+
 import { checkModifiers } from './check-modifiers';
 import { isValidAssignmentTarget, createError } from './node-utils';
 
-export function checkSyntaxError(node: ts.Node): void {
-  checkModifiers(node);
+const SyntaxKind = ts.SyntaxKind;
+
+export function checkSyntaxError(tsNode: ts.Node): void {
+  checkModifiers(tsNode);
+
+  const node = tsNode as TSNode;
 
   switch (node.kind) {
-    case ts.SyntaxKind.ForInStatement:
-    case ts.SyntaxKind.ForOfStatement: {
-      checkForStatementDeclaration(
-        node as ts.ForInStatement | ts.ForOfStatement,
-      );
+    case SyntaxKind.ForInStatement:
+    case SyntaxKind.ForOfStatement: {
+      checkForStatementDeclaration(node);
       break;
     }
     default: {
@@ -24,7 +28,7 @@ function checkForStatementDeclaration(
   node: ts.ForInStatement | ts.ForOfStatement,
 ): void {
   const { initializer, kind } = node;
-  const loop = kind === ts.SyntaxKind.ForInStatement ? 'for...in' : 'for...of';
+  const loop = kind === SyntaxKind.ForInStatement ? 'for...in' : 'for...of';
   if (ts.isVariableDeclarationList(initializer)) {
     if (initializer.declarations.length !== 1) {
       throw createError(
@@ -45,7 +49,7 @@ function checkForStatementDeclaration(
       );
     }
     if (
-      kind === ts.SyntaxKind.ForInStatement &&
+      kind === SyntaxKind.ForInStatement &&
       initializer.flags & ts.NodeFlags.Using
     ) {
       throw createError(
@@ -55,8 +59,8 @@ function checkForStatementDeclaration(
     }
   } else if (
     !isValidAssignmentTarget(initializer) &&
-    initializer.kind !== ts.SyntaxKind.ObjectLiteralExpression &&
-    initializer.kind !== ts.SyntaxKind.ArrayLiteralExpression
+    initializer.kind !== SyntaxKind.ObjectLiteralExpression &&
+    initializer.kind !== SyntaxKind.ArrayLiteralExpression
   ) {
     throw createError(
       initializer,


fisker:
Can we run the AST check from https://github.com/Lonercode/typescript-eslint/blob/65803e4127b9557ec6cf4802df12e588370d9930/packages/typescript-estree/src/convert.ts#L509, and have a big switch-case in the new file to check every ts node?

Lonercode:
> Can we run the AST check from https://github.com/Lonercode/typescript-eslint/blob/65803e4127b9557ec6cf4802df12e588370d9930/packages/typescript-estree/src/convert.ts#L509, and have a big switch-case in the new file to check every ts node?

AST check `checkModifier` function is in https://github.com/Lonercode/typescript-eslint/blob/65803e4127b9557ec6cf4802df12e588370d9930/packages/typescript-estree/src/check-modifiers.ts#L139. Do you mean it should be in the new file with a check for every ts node? Or I modify `checkForStatementDeclaration` in the new file?

fisker:
No need change check-modifiers.ts.

In the new file export a function to check ts node, it calls `checkModifier` and do other checks currently lives in convert.ts

packages/typescript-estree/src/convert.ts
```diff
@@ -915,7 +873,7 @@ export class Converter {
         });
 
       case SyntaxKind.ForInStatement:
-        this.#checkForStatementDeclaration(node.initializer, node.kind);
+        this.#checkTSNode(node, node.initializer, node.kind);
```
fisker: This should be able to be removed?

Lonercode:
@fisker Please let me know if I need to remove every call to the ast check function from convert.ts.

fisker:
`checkTSNode` should only call in `converter`, like the old `checkModofiers` did.

packages/typescript-estree/src/convert.ts
```diff
@@ -105,58 +105,16 @@ export class Converter {
...
-    }
-  }
-
-  #checkModifiers(node: ts.Node): void {
     if (this.options.allowInvalidAST) {
       return;
     }
 
-    checkModifiers(node);
+    checkTSNode(node, this.#throwError, initializer, kind);
```
fisker: `initializer` and `kind` should access from `node` directly in the check function.

packages/typescript-estree/src/ast-checks.ts
```diff
@@ -0,0 +1,80 @@
+import * as ts from 'typescript';
+
+import type { TSESTree } from './ts-estree';
+
+import { checkModifiers } from './check-modifiers';
+import { isValidAssignmentTarget } from './node-utils';
+
+export function checkTSNode(
```
fisker: I prefer export as `checkSyntaxError`, and rename the file accordingly.

packages/typescript-estree/src/ast-checks.ts
```diff
@@ -0,0 +1,80 @@
...
+  node: ts.Node,
+  throwError: (
```
fisker: This won't be necessary to pass, after we merge #11775

packages/typescript-estree/src/ast-checks.ts
```diff
@@ -0,0 +1,80 @@
...
+      );
+      break;
+    }
+    default: {
+      break;
+    }
+  }
+}
+
+export function checkForStatementDeclaration(
```
fisker: This should be an internal functionality, no need export.

packages/typescript-estree/tests/lib/ast-checks.test.ts
```diff
@@ -0,0 +1,196 @@
+import * as ts from 'typescript';
```
fisker: This test is unnecessary, they have already been well tested.

Lonercode: Was trying to align with codecov requirements. I can remove the file if it could still be merged. Should we also skip testing the new `checkSyntaxError` function?

packages/typescript-estree/src/ast-checks.ts
Lonercode: This means we should remove the `throwError` parameter then?

fisker: #11775 is merged, we should be able to throw directly.

packages/typescript-estree/src/check-syntax-errors.ts
```diff
@@ -0,0 +1,67 @@
...
+    default: {
+      break;
+    }
+  }
+}
+
+function checkForStatementDeclaration(
+  initializer: ts.ForInitializer,
+  kind: ts.SyntaxKind,
+): void {
```
fisker:
```suggestion
      checkForStatementDeclaration(node);
      break;
    }
    default: {
      break;
    }
  }
}

function checkForStatementDeclaration(node: ts.ForInStatement | ts.ForOfStatement): void {
  const {initializer, kind} = node;
```

Lonercode: The only issue with this is that an error is thrown unless I specify the node type in `checkForStatementDeclaration` so I may still need to do `checkForStatmentDeclaration(node as ts.ForInStatment | ts.ForOfStatment)` when calling it in `checkSyntaxErrors`

fisker: Ts can't narrow type?

Looking at the difference, 

```
checkSyntaxError(node: ts.Node)
```

while `convertNode` uses `TSNode`

https://github.com/Lonercode/typescript-eslint/blob/ee1ada9db5229902f70faa19dc27950dc457cc7d/packages/typescript-estree/src/convert.ts#L666

maybe change to it will fix the issue?

packages/typescript-estree/src/check-syntax-errors.ts
```diff
@@ -0,0 +1,66 @@
+import * as ts from 'typescript';
+
+import { checkModifiers } from './check-modifiers';
+import { isValidAssignmentTarget, createError } from './node-utils';
+
```
fisker, line 7: We should also define a 

```
const SyntaxKind = ts.SyntaxKind;
```

here, https://github.com/Lonercode/typescript-eslint/blob/ee1ada9db5229902f70faa19dc27950dc457cc7d/packages/typescript-estree/src/convert.ts#L42

so other checks can be copied easier to here.

fisker:
@JoshuaKGoldberg Are you fine with just one check in this PR? Or should all checks done in this one? If we want to copy variable declaration syntax check, better wait for #11777 and #11778 to merge first.

packages/typescript-estree/src/check-syntax-errors.ts
fisker: As I tested, it works

image

Lonercode: That's true but if I do this, I get an error in the converter method in `convert.ts` as node is `ts.Node` not `TSNode`:

ts-eslint-err

fisker: How about

```
function checkSyntaxError(tsNode: ts.node): void {
	checkModifiers(tsNode)

	const node = tsNode as TSNode;

	switch (...) {}
}
```

convert.ts also do similar.

Lonercode: That works :)

packages/typescript-estree/src/check-syntax-errors.ts
```diff
@@ -0,0 +1,70 @@
...
+  switch (node.kind) {
+    case SyntaxKind.ForInStatement:
+    case SyntaxKind.ForOfStatement: {
+      checkForStatementDeclaration(node);
+      break;
+    }
+    default: {
+      break;
+    }
```
fisker, line 23: Add this because ESLint complaint? Will this work?

```suggestion
    // No default
```

fisker:
Looks good to me, thanks! 🙏 

Can't approve since I'm not a team member.

packages/typescript-estree/src/check-syntax-errors.ts
Lonercode: I think it's valid, exiting with the default, for now at least.

Lonercode:
Thanks @fisker

packages/typescript-estree/src/check-syntax-errors.ts
JoshuaKGoldberg: Ehh should we? We don't generally do namespace property shorthands like that in the typescript-eslint monorepo. IMO it just adds a bit of clutter.

fisker: We'll add lots of `case`s in switch, better align with convert.ts.

fisker: At least keep it before it's done?

packages/typescript-estree/src/check-syntax-errors.ts
```diff
@@ -0,0 +1,70 @@
...
+import { isValidAssignmentTarget, createError } from './node-utils';
+
+const SyntaxKind = ts.SyntaxKind;
+
+export function checkSyntaxError(tsNode: ts.Node): void {
+  checkModifiers(tsNode);
+
+  const node = tsNode as TSNode;
+
+  switch (node.kind) {
```
JoshuaKGoldberg, line 15: [Refactor] I don't think this is a positive change for the logic? Converter functions generally directly call the `checkFor*` logic they need. For example, for-in and for-of statements directly call `checkForStatementDeclaration`. Now this new `checkSyntaxError` function has to again switch on `node.kind`. That's extra code structure and work.

@fisker please correct me if I'm wrong here, but I think it'd still be in the spirit of the issue to have `checkForStatementDeclaration` be a standalone function separated out from `convert.ts`, and still called directly?

As in, the diffs to `convert.ts` would look more like:

```diff
+ import { checkForStatementDeclaration } from "./checks/...";
...
      case SyntaxKind.ForInStatement:
-       this.#checkForStatementDeclaration(node.initializer, node.kind);
+       checkForStatementDeclaration(node.initializer, node.kind);
        return this.createNode<TSESTree.ForInStatement>(node, {
...
-  #checkForStatementDeclaration(...) { ... }
```

JoshuaKGoldberg (changes requested):
Thanks for getting started on this! I think we should discuss the direction a bit.

packages/typescript-estree/src/check-syntax-errors.ts
fisker: Take this https://github.com/typescript-eslint/typescript-eslint/blob/16cf0f7ff1f67b5be7bd4cc1defeee7745af9e9a/packages/typescript-estree/src/convert.ts#L802-L811 as example, normally the check just few lines (there are even simpler checks, eg: [1](https://github.com/typescript-eslint/typescript-eslint/blob/16cf0f7ff1f67b5be7bd4cc1defeee7745af9e9a/packages/typescript-estree/src/convert.ts#L834-L836)). Not every node need such long code like `SyntaxKind.ForInStatement`.

If we call from covert.ts directly after the `case ...`, we'll have lot of lines to add. and forcing many functions to import/export.

I think a single call like `checkModifiers()` used did and duplicate the big `switch(){}` in check-syntax-error.ts is much easier to maintain.

Most kind of node check will inline, instead of separate function.

Lonercode: @JoshuaKGoldberg @fisker Would it prove more efficient to separate out `checkForStatmentDeclaration`? I thought the point of the issue was to extract those checks?

fisker:
@JoshuaKGoldberg

Which part do you insist on changing for landing this PR? I have some other syntax check changes to make, and I'm waiting for this to merge first.

Personally, I like the current implementation, the struct will be easy to extract other checks. How about we take another look after all the checks are extracted? I can make the changes as you request after we finish the migration.

JoshuaKGoldberg:
Oh, I'm not picky here. I'm happy to just see what you want to get refactored happen first. I'm not sure I'm following though. Are you proposing landing this as-is, or doing some refactors first?

fisker:
Land as it is, and continue to extract all syntax checks in follow up PR.

JoshuaKGoldberg (approved):
Cool! Although this isn't my personal instinct for how I would write things, I think it's a good direction to try out. And I do think there's a nice simplicity in being able to "just" call `checkSyntaxError` and having `checkModifiers` + whatever node-specific checks are needed be run. So perhaps my instincts are wrong 😄.

Either way, thanks for bearing with the reviews @Lonercode, I appreciate you pushing forward on this! And also thanks to @fisker for taking on reviewing.

fisker:
I'll send follow-up PR.

Issue #11593, opened 2025-09-10
Repo: Can't build packages when new property added to Node


fisker:
### Suggestion

Add a new property in Node type definition, add the new property to visitorKeys, the build script fails.

Found this during #11591

kirkwaiblinger:
Can you provide some more info? I took commit 5684f6a8afba9b616fce5130df013637c9dffb3d (currently latest) from #11591, then applied this diff:

```diff
diff --git a/packages/visitor-keys/src/visitor-keys.ts b/packages/visitor-keys/src/visitor-keys.ts
index e1daf8dbb..09ceae691 100644
--- a/packages/visitor-keys/src/visitor-keys.ts
+++ b/packages/visitor-keys/src/visitor-keys.ts
@@ -213,8 +213,7 @@ const additionalKeys: AdditionalKeys = {
   TSExternalModuleReference: ['expression'],
   TSFunctionType: SharedVisitorKeys.FunctionType,
   TSImportEqualsDeclaration: ['id', 'moduleReference'],
-  // TODO: Update to use `source`
-  TSImportType: ['argument', 'options', 'qualifier', 'typeArguments'],
+  TSImportType: ['source', 'options', 'qualifier', 'typeArguments'],
   TSIndexedAccessType: ['objectType', 'indexType'],
   TSIndexSignature: ['parameters', 'typeAnnotation'],
   TSInferType: ['typeParameter'],
(END)
```

ran `yarn build` and had no problems locally.

fisker:
I don't know, can't build on CI  with this change. This is also the same as the first commit. https://github.com/typescript-eslint/typescript-eslint/actions/runs/17628354411/job/50090539833

kirkwaiblinger:
Probably an nx thing then @JamesHenry

JamesHenry:
Sorry for the delay here folks, I pushed a fix on this other PR:

https://github.com/typescript-eslint/typescript-eslint/pull/11760#issuecomment-3551412595

PR #11760, opened 2025-11-14
fix(typescript-estree): disallow binding patterns in parameter properties

### LINKED ISSUE #11708: Enhancement: throw an error for invalid `TSParameterProperty`s

### Relevant Package

ast-spec

### Description

```ts
class Test1 {
  constructor(
    private [foo] = [1],
    private { bar } = { bar: 1 },
    private ...baz
  ) { }
}
```

Currently these three parameter properties are semantic errors in TS and are similarly allowed by our parser.
However they are all non-sensical. The first two [don't even transpile to generate a property](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEAqCmEAuBGaBvAUNawD2AdsgE4CuwSexAFNjtAA7ECWAbmEvNANoBmeeALrQAvDxSCANHRxM2HLumgAjMMWgBfURhVqAXNDQbp9Ri3adoAOhuqAXnQCUOjZldA)!

We should make these hard errors to block the syntax so that we can refine our [`TSParameterProperty.parameter`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/src/parameter/TSParameterProperty/spec.ts) type to exclude the invalid states.

An additional case which is already a hard TS parser error that we should pre-narrow the types for -- the `.left` of an assignment parameter property _must_ always be an `Identifier`.

dbarabashh:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11708 

## Overview
This PR implements validation to block invalid parameter property patterns that typescript treats as semantic errors:
- Array binding patterns: `private [foo] = [1]`
- Object binding patterns: `private { bar } = { bar: 1 }`
- Rest parameters: `private ...baz`


packages/eslint-plugin/src/util/class-scope-analyzer/extractComputedName.ts
```diff
@@ -59,22 +60,10 @@ function extractNonComputedName(
...
-      case AST_NODE_TYPES.AssignmentPattern:
-        if (node.parameter.left.type !== AST_NODE_TYPES.Identifier) {
-          return null;
-        }
-        return extractNonComputedName(node.parameter.left);
-
-      case AST_NODE_TYPES.Identifier:
-        return extractNonComputedName(node.parameter);
-    }
+    // After #11708, parameter can only be Identifier or AssignmentPattern
```
JoshuaKGoldberg: [Style] Long-term, reading devs don't need to know about specific issues like this. These comments have a habit of getting out-of-date over time.

```suggestion
```

packages/eslint-plugin/src/util/getParameterPropertyIdentifier.ts
```diff
@@ -0,0 +1,26 @@
+import type { TSESTree } from '@typescript-eslint/utils';
+
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+
+/**
+ *
+ * Since #11708, parameter properties can only be simple identifiers or assignment patterns,
+ * and when it's an assignment pattern, the left side must be an identifier.
+ *
```
JoshuaKGoldberg:
```suggestion
 * Parameter properties can only be simple identifiers or assignment patterns,
 * and when it's an assignment pattern, the left side must be an identifier.
```

packages/typescript-estree/tests/lib/convert.test.ts
```diff
@@ -282,6 +282,78 @@ describe('convert', () => {
     });
   });
 
+  describe('invalid parameter properties', () => {
```
JoshuaKGoldberg: [Testing] We don't generally put _"does this specific invalid syntax throw an error?"_ tests in `convert.test.ts`. They're generally put in the dedicated `fixtures/` tests.   `convert.test.ts`'s `describe()` blocks are only used for high-level behavior tests for `Converter`'s behaviors.

You can find references of how other fixtures deal with errors by searching for their error text. For example, `string literal cannot be used` shows packages/ast-spec/src/declaration/ExportNamedDeclaration/fixtures/_error_/braced-string-literal-aliased-to-identifier-without-source/snapshots/1-TSESTree-Error.shot and co.

JoshuaKGoldberg:
🤔 I don't get these failures locally. Re-running...

packages/eslint-plugin/src/util/getParameterPropertyIdentifier.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export function getParameterPropertyIdentifier(
```
JoshuaKGoldberg: That being said, I don't think this function is necessary? There should be no need to, at runtime, assert on its type. The AST types should properly indicate what the parameter can be. If you're seeing that the AST node types aren't narrow enough, the solution would be to fix the AST node types.

dbarabashh:
@JoshuaKGoldberg can you help with github actions? I can't reproduce this locally. Locally everything passes and everything is ok

packages/eslint-plugin/src/util/getParameterPropertyIdentifier.ts
dbarabashh: Thanks a lot for the comment I'll keep that in mind for the future as well

JoshuaKGoldberg:
Blurgh, yes, I also don't reproduce them locally. We can ignore them in this PR for now - let's just pretend the build is passing. In parallel I'll try to clear the Nx cache and work with @JamesHenry on #11593.

packages/eslint-plugin/src/util/class-scope-analyzer/extractComputedName.ts
```diff
@@ -59,22 +59,11 @@ function extractNonComputedName(
...
-        return extractNonComputedName(node.parameter.left);
-
-      case AST_NODE_TYPES.Identifier:
-        return extractNonComputedName(node.parameter);
-    }
+    const identifier =
+      node.parameter.type === AST_NODE_TYPES.Identifier
+        ? node.parameter
+        : (node.parameter.left as TSESTree.Identifier);
+    return extractNonComputedName(identifier);
```
JoshuaKGoldberg, line 66: [Praise] 👌 I like how much code is getting removed here!

JamesHenry:
I did two items of repo housekeeping to resolve the CI build issue:

- Removed the old `SKIP_AST_SPEC_REBUILD` env var: https://github.com/typescript-eslint/typescript-eslint/pull/11760/commits/c2ee584e8ff2781d06a508284745e1df0adad4cc
  - Based on how it was being checked it was doing the opposite of what the name implied, and it was also imperatively triggering tasks nested inside the task graph which is always going to be a recipe for subtle issues
- Fixed the project graph to accurately reflect the fact that the ast-spec is implicitly depended upon by essentially everything else: https://github.com/typescript-eslint/typescript-eslint/pull/11760/commits/348e7d79de1eedc9e34c13035574d20127d12dc2

**Before**

`ast-spec` was orphaned without any traced relationships, which clearly is not correct

image

**After**

`ast-spec` is appropriately one of the core nodes at the base of the graph

image

dbarabashh:
@JoshuaKGoldberg you [left a comment](https://github.com/typescript-eslint/typescript-eslint/pull/11760#discussion_r2534977260) here and I removed those tests but now there’s a problem with the coverage rate

packages/eslint-plugin/src/rules/no-unnecessary-parameter-property-assignment.ts
```diff
@@ -109,8 +109,7 @@ export default createRule({
         ((node.parameter.type === AST_NODE_TYPES.Identifier && // constructor (public foo) {}
           node.parameter.name === name) ||
           (node.parameter.type === AST_NODE_TYPES.AssignmentPattern && // constructor (public foo = 1) {}
-            node.parameter.left.type === AST_NODE_TYPES.Identifier &&
-            node.parameter.left.name === name))
+            (node.parameter.left as TSESTree.Identifier).name === name))
```
bradzacher, line 112: is this assertion still necessary?

bradzacher (approved):

packages/eslint-plugin/src/rules/no-unnecessary-parameter-property-assignment.ts
JoshuaKGoldberg: It's not, I'll go ahead and remove the assertion before merging.

PR #11764, opened 2025-11-16
fix(eslint-plugin): [restrict-template-expressions] check base types in allow list

### LINKED ISSUE #11759: Bug: [restrict-template-expressions] should check if superclass is ignored

### Playground Link

https://typescript-eslint.io/play/#ts=5.2.2&fileType=.tsx&code=MYGwhgzhAEBCkFNoG9oF8BQpIwCIICcBLANwQBNoEAPAFwQDtyZ4IlVNgB7BiW6AGZcu0ALzQGCAO5xEACgCUAbiw8%2B0AEZgCYidOj5iZcopUADACTIhXNFa0E0ZpUA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6aFfKsu-RAW2LwAhn3oAPYh2TJKAeyaoMAbXDYciaNDnRIAGjXrgUYfHhyA7ujDLIAIWHJEkALoBfQ2BdqPboA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
class Base { }
class Derived extends Base { }
const foo = new Base();
const bar = new Derived();
`${foo}${bar}`;
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/restrict-template-expressions": [
      "error",
      { "allow": ["Base"]}
    ]
  }
}
```

### tsconfig

```jsonc
{
  "rules": {
    "@typescript-eslint/restrict-template-expressions": [
      "error",
      { "allow": ["Base"]}
    ]
  }
}
```

### Expected Result

No warnings - if `Base` is allowed and every instance of `Derived` is also a `Base`, then `Derived` should be allowed. (This would also be consistent with the `no-base-to-string` rule - see #11520.)

### Actual Result

> Invalid type "Derived" of template literal expression. 5:10 - 5:13

### Additional Info

_No response_

Higangssh:
Fixes #11759

  ## PR Checklist

  - [x] Addresses an existing open issue: fixes #11759
  - [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%
  3Aopen+is%3Aissue+label%3A%22accepting+prs%22)

  ## Overview

  This PR fixes an issue where the `restrict-template-expressions` rule incorrectly rejects derived class instances
  in template literals when their base class is in the allow list.

  ## Problem

  Currently, when a base class is included in the `allow` option, derived classes extending it are still flagged as
  errors, violating the Liskov Substitution Principle where subtypes should be substitutable for their supertypes.

  Example:
  ```typescript
  // Configuration: { allow: [{ from: 'file', name: 'Base' }] }
  class Base { }
  class Derived extends Base { }
  const bar = new Derived();
  `${bar}`;  // Error: Invalid type "Derived"
```
  
  ## Solution

  Implemented recursive base type checking following the pattern established in the no-base-to-string rule:

  1. Added hasBaseTypes() helper function to identify types with base types (Interface or Class)
  2. Added isAllowedTypeOrBase() function that recursively checks if a type or its base types match the allow list
  3. Included cycle detection using a Set to prevent infinite recursion
  4. Updated recursivelyCheckType() to utilize the new base type checking mechanism

   ## Changes

  - packages/eslint-plugin/src/rules/restrict-template-expressions.ts: Added base type checking logic
  - packages/eslint-plugin/tests/rules/restrict-template-expressions.test.ts: Added test case for inheritance
  scenario

### CODE REVIEW

File: packages/eslint-plugin/tests/rules/restrict-template-expressions.test.ts [hunk 1]
```diff
@@ -344,6 +344,52 @@ ruleTester.run('restrict-template-expressions', rule, {
     'const msg = `arg = ${undefined}`;',
     'const msg = `arg = ${123}`;',
     "const msg = `arg = ${'abc'}`;",
+    // https://github.com/typescript-eslint/typescript-eslint/issues/11759
+    // allow should check base types
```
JoshuaKGoldberg:
```suggestion
```

[Style] Nit: we don't generally put these comments in, unless it's really not clear what the tests are doing. Comments aren't enforced by anything to be correct -- and they have a habit of getting out-of-date over time.

The tests you're adding are nice and clear (yay! nicely done!) so I don't think this or the other comments are needed.

packages/eslint-plugin/src/rules/restrict-template-expressions.ts
```diff
@@ -130,6 +131,33 @@ export default createRule<Options, MessageId>({
       ({ option }) => options[option],
     );
 
+    function hasBaseTypes(type: Type): type is InterfaceType {
```
JoshuaKGoldberg: [Cleanup] This is the same `hasBaseTypes` as in `no-base-to-string.ts`. Let's move it to a file under `../util` to deduplicate.

packages/eslint-plugin/src/rules/restrict-template-expressions.ts
```diff
@@ -130,6 +131,33 @@ export default createRule<Options, MessageId>({
...
+      return (
+        isObjectType(type) &&
+        isObjectFlagSet(type, ObjectFlags.Interface | ObjectFlags.Class)
+      );
+    }
+
+    function isAllowedTypeOrBase(type: Type, seen = new Set<Type>()): boolean {
```
JoshuaKGoldberg: [Cleanup] This is _almost_ the same as the `isIgnoredTypeOrBase` in `no-base-to-string.ts`. The only difference is that this one properly uses `typeMatchesSomeSpecifier`, whereas that one uses the legacy `ignoredTypeNames`. But both of these are tricky logic that's relevant to any _"check if this class extends a specified base"_.

Change request: similar to deduplicating `hasBaseTypes`, let's deduplicate this too. It can take in a `matcher: (type: ts.Type) => boolean` to customize the _"is this type allowed?"_ logic.

```ts
export function matchesTypeOrBaseType(
  services: ParserServicesWithTypeInformation,
  matcher: (type: ts.Type) => boolean,
  type: Type,
seen = new Set<Type>(),
): boolean {
```

```ts
// no-base-to-string.ts
matchesTypeOrBaseType(services, type => ignoredTypeNames.includes(getTypeName(checker, typeName)), type)
```

```ts
// restrict-template-expressions.ts
matchesTypeOrBaseType(services, type => typeMatchesSomeSpecifier(type, allow, program), type)
```

Does that make sense?

Higangssh:
Thanks for the review! I've addressed all the feedback:
  - Removed unnecessary test comments
  - Extracted hasBaseTypes to shared util
  - Created matchesTypeOrBaseType with matcher pattern

packages/eslint-plugin/tests/rules/restrict-template-expressions.test.ts
JoshuaKGoldberg: @Higangssh I think you missed this comment?

Higangssh: @JoshuaKGoldberg  Thank you comment again , All comments removed now.
 I left line 380 (`// allow list with type alias without base types`) since it explains a specific edge case, but
  let me know if you'd prefer me to remove it too

packages/eslint-plugin/tests/rules/restrict-template-expressions.test.ts
JoshuaKGoldberg, line 1: [Testing] Oh, sory, I just realized - this is missing test coverage for extending multiple classes.

* `A extends B, C` where `B` is listed
* `A extends B, C` where `C` is listed
* `A extends B, C` + `B extends D, E` where `E` is listed

etc.

Higangssh:
@JoshuaKGoldberg Added tests for multiple interface inheritance. Thanks for the thorough review - really
  appreciate you taking the time to guide me through this! Let me know if there's anything else that needs
  adjusting

PR #11769, opened 2025-11-19
fix(eslint-plugin): [consistent-type-exports] check value flag before resolving alias

### LINKED ISSUE #11761: Bug: [consistent-type-exports] Importing a shadowing type breaks consistent-type-imports/exports

### Issue Description

The following piece of code has two issues:

```ts
import { SomeThing } from './types.js';

const SomeThing = { foo: 'bar' };

export { SomeThing };
```

Firstly, it does not flag that `import { SomeThing } from './types.js';` should be `import { type SomeThing } from './types.js';` (problem with `consistent-type-imports`).

Secondly, it does error with "All exports in the declaration are only used as types.", even though the export is very clearly _not_ a type.

This issue manifested itself in v8.43.0. I suspect it is a regression to do with https://github.com/typescript-eslint/typescript-eslint/pull/11457

https://github.com/simon-abbott/ts-eslint-issue-mre-import-shadowing

### Repro Steps

1) clone the repo
2) `npm install`
3) `npm run lint`

Running `npm run lint` _should_ error on the `import` in `src/index.ts`, and _should not_ error on the export. Instead it does the reverse.

### Versions

| `@typescript-eslint/eslint-plugin`     | `8.46.4` |
| `@typescript-eslint/parser`            | `8.46.4` |
| `TypeScript`                           | `5.9.2` |
| `ESLint`                               | `8.57.1` |
| `node`                                 | `22.19.0` |

ntdiary:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11761

## Overview

Make sure the current symbol doesn't have `SymbolFlags.Value` before resolving alias.

I submit the recursive version here, which is just 1 line shorter than the `while` iteration.
If we prefer using iteration, can be done like this:

```ts
while (symbol && !checker.isUnknownSymbol(symbol)) {
  if (
    symbol.getDeclarations()?.some(ts.isTypeOnlyImportOrExportDeclaration)
  ) {
    return true;
  }
  if (symbol.flags & ts.SymbolFlags.Value) {
    return false;
  }
  if (!(symbol.flags & ts.SymbolFlags.Alias)) {
    return true;
  }
  symbol = checker.getImmediateAliasedSymbol(symbol)
}
return undefined;
```
💖


packages/eslint-plugin/src/rules/consistent-type-exports.ts
```diff
@@ -90,18 +90,20 @@ export default createRule<Options, MessageIds>({
...
+      if (
+        symbol.getDeclarations()?.some(ts.isTypeOnlyImportOrExportDeclaration)
+      ) {
+        return true;
+      }
+      if (symbol.flags & ts.SymbolFlags.Value) {
+        return false;
+      }
+      return symbol.flags & ts.SymbolFlags.Alias
+        ? isSymbolTypeBased(checker.getImmediateAliasedSymbol(symbol))
```
ntdiary, line 105: replace `getAliasedSymbol` with `getImmediateAliasedSymbol`.
Because the latter can handle this case as well:
```ts
// a.ts
export const NAME = 1;
export type NAME = 1;
// b.ts
import { type NAME as Foo } from './a';
export { Foo };
```

JoshuaKGoldberg:
I ended up not having time to merge this in for the Monday release this week. Will leave open.

PR #11773, opened 2025-11-21
chore(eslint-plugin): correct TODO comment on the wrong line

Zamiell:
I noticed that the TODO comment in this file was actually in the wrong spot, implying that "no-new-native-nonconstructor" was the deprecated rule instead of "no-new-symbol".

I moved the comment, and then expanded it to make it more descriptive.

### DIFF
--- a/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts
+++ b/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts
@@ -28,12 +28,12 @@ const config = (
     'no-func-assign': 'off', // ts(2630)
     'no-import-assign': 'off', // ts(2632) & ts(2540)
     'no-new-native-nonconstructor': 'off', // ts(7009)
-     // "no-new-symbol" was deprecated in ESLint 9.0.0 and will be removed in
-     // ESLint v11.0.0. See:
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/src/configs/eslint-recommended-raw.ts
```diff
@@ -27,8 +27,13 @@ const config = (
     'no-dupe-keys': 'off', // ts(1117)
     'no-func-assign': 'off', // ts(2630)
     'no-import-assign': 'off', // ts(2632) & ts(2540)
-    // TODO - remove this once we no longer support ESLint v8
     'no-new-native-nonconstructor': 'off', // ts(7009)
+     // "no-new-symbol" was deprecated in ESLint 9.0.0 and will be removed in
+     // ESLint v11.0.0. See:
+     // https://eslint.org/docs/latest/rules/no-new-symbol
+     // We need to keep the rule disabled until TSESlint drops support for
```
JoshuaKGoldberg:
```suggestion
     // We need to keep the rule disabled until TSESLint drops support for
```

JoshuaKGoldberg (dismissed):
Thanks!

PR #11789, opened 2025-11-24
feat(eslint-plugin): [no-unnecessary-type-assertion] report more cases based on assignability

### LINKED ISSUE #8107: Rule proposal: no `as any` when type is already expected type

### Description

This proposal is for a rule that detects and can auto-fix casts to any, when the type is already the type that's expected in that situation.

It's semi-common in test code, to use `as any` to make a stub or similar pass type checks when passing as a function parameter, however this is hard to get away from once already done because it's hard to identify when it's made redundant.

My proposal is basically for the following situation to warn that the `as any` is a redundant cast, as the type already satisfies the required conditions.

```
function doThing(a: number) {
}

doThing(5 as any); // <-- Rule fails on this line, as `5` already satisfies the type of number 
```

A possible extension to this rule would be identifying the pattern of `5 as unknown as number` or `5 as any as number`, another common pattern used within tests to forcibly make stubs conform to a given type. However, that might also make sense as an extension to `no-unnecessary-type-assertion` as it also fits into "when a type assertion does not change the type of an expression" unlike the main case of this proposed rule.

### Fail Cases

```TypeScript
interface A {
  required: string;
  alsoRequired: number;
}

function doThing(a: A) {}

// Fails as this already fulfils the type requirement, the `as any` is not preventing a type error here
doThing({ required: 'yes', alsoRequired: 'alsoYes' } as any);
```

### Pass Cases

```TypeScript
interface A {
  required: string;
  alsoRequired: number;
}

function doThing(a: A) {}

// The rule accepts this code, as it does not fulfil the requirements and the `any` is masking a type error
doThing({ required: 'yes' } as any);
```

### Additional Info

I can see an argument where making this a part of `no-unnecessary-type-assertion` would make sense, and am happy to switch this proposal to that if requested. Given the underlying "cause" is different though it felt like it made more sense to propose this as a separate rule.

ulrichstark:
## PR Checklist

- [x] Addresses an existing open issue: fixes #8107

## Overview

It now also reports cases like:

```ts
const v: number = 5;
const x = v as any as number;
```

And fixes it to:

```ts
const v: number = 5;
const x = v;
```


ulrichstark:
This was a pretty tricky change to get right. Coverage is at 100%, it reported multiple cases while dogfooding, I turned the false positives into test cases and fixed the true positives. Let the review iterations begin!

ulrichstark:
At least two missing cases that I need to fix:

```ts
declare function fn(param: number);
fn(42 as unknown as number);
```

```ts
declare function fn(param: number);
fn(42 as any as number);
```

packages/typescript-estree/src/node-utils.ts
```diff
@@ -386,12 +386,12 @@ export function findFirstMatchingAncestor(
...
   predicate: (node: ts.Node) => boolean,
 ): ts.Node | undefined {
-  let current: ts.Node | undefined = node;
+  let current = node as ts.Node | undefined;
   while (current) {
     if (predicate(current)) {
       return current;
     }
-    current = current.parent as ts.Node | undefined;
+    current = current.parent;
```
JoshuaKGoldberg, line 394: [Bug?] I believe this was more correct before?

* For the `let current` declaration, `node` _isn't_ `ts.Node | undefined`. We know it's `ts.Node`
* For this assignment, `current.parent` _is_ `ts.Node | undefined`. The types are just wrong.

So I think the rule is being overly aggressive about reporting `current = current.parent as ts.Node | undefined`.

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -219,13 +223,148 @@ export default createRule<Options, MessageIds>({
...
+      if (
+        isTypeFlagSet(uncast, ts.TypeFlags.NonPrimitive) &&
+        !isTypeFlagSet(cast, ts.TypeFlags.NonPrimitive)
+      ) {
+        return false;
+      }
+
+      if (hasIndexSignature(uncast) && !hasIndexSignature(cast)) {
+        return false;
+      }
```
JoshuaKGoldberg: [Testing] Removing this doesn't fail any unit tests. Missing tests or unnecessary code?

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -219,13 +223,148 @@ export default createRule<Options, MessageIds>({
...
+
+      if (containsAny(uncast) || containsAny(cast)) {
+        return false;
+      }
```
JoshuaKGoldberg: [Testing] Removing this doesn't fail any unit tests. Missing tests or unnecessary code?

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -219,13 +223,148 @@ export default createRule<Options, MessageIds>({
...
+      if (isConceptuallyLiteral(expression)) {
+        return false;
+      }
+
+      if (
+        isTypeFlagSet(uncast, ts.TypeFlags.NonPrimitive) &&
+        !isTypeFlagSet(cast, ts.TypeFlags.NonPrimitive)
+      ) {
+        return false;
+      }
```
JoshuaKGoldberg, line 241: [Testing] Removing this doesn't fail any unit tests. Missing tests or unnecessary code?

JoshuaKGoldberg (changes requested):
👏 This is so cool! It's always surprising to see a bunch of new fixes in the codebase. Nice.

Requesting changes on two points: a few areas that don't fail tests when removed, and also I think an incorrect report in existing repo code?

packages/typescript-estree/src/node-utils.ts
ulrichstark: I don't think the rule is overly aggressive. Instead of fixing the rule, I would prefer to rewrite findFirstMatchingAncestor like so:

```ts
let current = node;
while (true) {
  if (predicate(current)) {
    return current;
  }
  if (current.kind === SyntaxKind.SourceFile) {
    break;
  }
  current = current.parent;
}
return undefined;
```

JoshuaKGoldberg: I suppose that could work, but the rule is wrong to report on the code as-is, no?

ulrichstark: I would say no, because [this](https://deploy-preview-11789--typescript-eslint.netlify.app/play#ts=5.9.2&fileType=.tsx&code=JYOwLgpgTgZghgYwgAgPICMBWEFmQbwChlkAHKAe1OjAE8AuZEAVwFt1oBuQgX0MIQUQAZzwUsjDNlzIAvAWJlK1KHUYBWADS9uhADYQ8YOFADmhyVhx4APsmYgAJhBigIj7sbOG5ycZmQ4YTQrGTsHZ1cQd10gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false) is also a true positive by the rule and essentially the same code as `findFirstMatchingAncestor`. There's no need to cast `obj` to a wider type, because `target` already accepts `obj`'s type, therefore making the type assertion unnecessary.

ulrichstark:
This PR is going back into draft. After today's testing in our repo, I found some cases that weren't yet reported. But sadly this also lead to more false positives when I dogfooded my changes on the TypeScript repo. This rule with my changes currently finds 145 unnecessary type assertion in the TypeScript repo. When applying the automatic fixes, there will be 33 TypeScript errors. So It might take a bit of time to fix those remaining false positives.

ulrichstark:
I think this PR is now finally ready for review! I fixed much more false positives after testing this rule across multiple projects. This rule is able to find unnecessary type assertion in almost every project as you see in the following PRs:

https://github.com/microsoft/TypeScript/pull/62969
https://github.com/TanStack/router/pull/6365
https://github.com/TanStack/query/pull/10028

ulrichstark:
This needs more work sadly... Going back into draft....

ulrichstark:
Now it's finally ready. Despite having an extensive test collection, I still expect there will be many false positives when this change is released into the wild. It required a lot of back and forth to get to this implementation and I tried my best to add tests along the way to not regress.

image

packages/typescript-estree/src/node-utils.ts
kirkwaiblinger: This really has me thinking hard. The trouble is that the widening assertion affects narrowing in subsequent code. For example, this code is allowed by TS, and will cause a runtime error.

```ts
current = current.parent;
current.kind === SyntaxKind.Whatever;
```

This code is rejected by TS because `current.kind` is possibly-`undefined`
```ts
current = current.parent as typeof current.parent | undefined;
current.kind === SyntaxKind.Whatever;
```

So, 

1. This specific code IMO should not be changed to remove the `| undefined` widening assertion. The widening assertion is there to make the code safer. 

Should it be reported by the rule (which is to say, should [this](https://deploy-preview-11789--typescript-eslint.netlify.app/play/#ts=5.9.2&fileType=.tsx&code=JYOwLgpgTgZghgYwgAgPICMBWEFmQbwChlkAHKAe1OjAE8AuZEAVwFt1oBuQgX0MIQUQAZzwUsjDNlzIAvAWJlK1KHUYBWADS9uhADYQ8YOFADmhyVhx4APsmYgAJhBigIj7sbOG5ycZmQ4YTQrGTsHZ1cQd10gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false) be reported? That's a separate question. I'm _leaning_ towards no... 

However, this clearly is a false positive; the type assertion is _narrowing_, and it's unnecessary.

```ts
declare const maybeUndefined: string | undefined;

declare let s: string | undefined;

s = maybeUndefined as string
```

So, I wonder if it's useful to contemplate whether a type assertion is narrowing, widening, or neither, in deciding whether to flag assertions as unnecessary? OR - should the rule flag all of them, and we need to communicate that some such reports will be false positives in which the rule should be disabled; do not assume that removing the "unnecessary" assertion is the correct remedy to a report.

kirkwaiblinger: Also, this type of analysis is particularly relevant where the type of the expression will determine the type of a symbol/expression. Something like `f(x as T)` where `T` is unnecessary (and `f` isn't generic) is _probably_ reasonable to flag no matter whether `as T` is widening, narrowing, or neither 🤔

kirkwaiblinger:
This is an impressive PR and clearly a lot of effort and attention to detail has been put in. Looking at our own repo, it also appears to catch _quite_ a few legitimate unnecessary assertions. Kudos!

However, I am a little perplexed because it seems to go well beyond the scope of #8107... As I understand that issue, it's more or less purely about `as any`-like assertions. 

I'm not so sure there's been discussion/consensus on things beyond that? Furthermore, "upcasts" (widening assertions) are explicitly mentioned as being something not in scope to report in https://github.com/typescript-eslint/typescript-eslint/issues/8107#issuecomment-1868473933, which comes into play with https://github.com/typescript-eslint/typescript-eslint/pull/11789#discussion_r2649223429. There may well be room for us to consider and think deeply about such cases, but I'm thinking this should warrant its own issue for visibility, rather than being implicitly part of the PR to resolve #8107... 

Please let me know if I'm missing something and there has been previous discussion on all this! And if so please link it in the PR description 🙏

ulrichstark:
Thank you @kirkwaiblinger!

Yes you are right that my rule detects more cases than discussed in the original issue. I was a bit too motivated and my logic was too broad in the beginning, so it detected these additional cases in internal and public projects (this, TypeScript, Tanstack router and query repo). These additional cases seemed reasonable to report and the false positives easy to fix, so I did continue going that route.

I can trim down (or maybe redo) my changes to the rule to detect less potentially false positive cases, but honestly I'm not sure anymore which of my "invalid" test cases should be in scope or not. Looking at the original issue, most of my "invalid" test cases fit the discussion in that issue (either chained assertions with the original and target type being assignable, or function calls where some value is cast to the contextual type of that parameter it's assigned to).

JoshuaKGoldberg:
@kirkwaiblinger my read of the discussion was to go past just `as any`, and to fully use the assignability APIs. WDYT?

i.e. https://github.com/typescript-eslint/typescript-eslint/issues/8107#issuecomment-1868429853 -> https://github.com/typescript-eslint/typescript-eslint/issues/8107#issuecomment-1868481469

kirkwaiblinger:
Oh, I see.... I think I misunderstood (or missed entirely) the comment about being assignable to the contextual type, my apologies.

JoshuaKGoldberg (approved):
This LGTM, but given how dangerous it is we should definitely also get review from @bradzacher and @kirkwaiblinger if they have time.

For context, we've had to revert at least IIRC 2 uses of the assignability API after releasing to production.

bradzacher (approved):

jimmy-zhening-luo:
Tons of breaking FP auto-fixes, removes type assertions as "unnecessary", which then breaks TypeScript transpilation.

bradzacher:
@jimmy-zhening-luo as per our contributing guide -- do not comment on closed PRs.
A one line comment like this is completely useless spam and doesn't help anyone understand or fix problems.

Please raise an issue with an reproduction.

PR #11826, opened 2025-12-07
fix(eslint-plugin): [no-unnecessary-type-assertion] correct handling of undefined vs. void

### LINKED ISSUE #11529: Bug: [no-unnecessary-type-assertion] false positive when casting IIFE with `void` return type to `undefined`

### Playground Link

https://typescript-eslint.io/play/#ts=5.9.2&fileType=.ts&code=PTAEhlyUDMEMBsDOBTUAHA9gglgFywNyQCgBjdAOwR1AFUA5AEQFEAxASTqYdAF5QAKfgEpeAPlABvUAF8hw0DASgAruQAmSKFnJI1AbiJEQoQKDkoOulBIATtfTXSFKtF4CRPcVOkGylavWZ2TgYAJlcoeUUVdU1tXQMgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWkOTJE0fJQ5N0UOdA7RI4MAF8QOoA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
// ❌ false positive
const UNDEFINED = (() => { })() as undefined;

// ✅ No error
const f = () => { };
const UNDEFINED2 = f() as undefined;
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-unnecessary-type-assertion": "error"
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

No error

### Actual Result

Error: this assertion is unnecessary since it does not change the type of the expression.

### Additional Info

This might be a minor issue, but it could indicate a problem with the rule. Casting an IIFE with a `void` return type to `undefined` triggers a false positive.

Tamashoo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11529 

## Overview

Added special handling for type checking of IIFEs in the no-unnecessary-type-assertion rule.
Introduced an isIIFE type guard and a getUncastType helper function to correctly resolve IIFE return types, converting implicit undefined to void for functions without explicit return type annotations.
This improves the accuracy of unnecessary assertion detection for IIFE expressions.

### DIFF
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
@@ -1542,7 +1542,7 @@ interface Overloaded {
   (value: string): undefined;
 }
 
-((value => {}) as Overloaded)("");
+((value => {}) as Overloaded)('');
       `,
     },
   ],


packages/eslint-plugin/tests/rules/no-unnecessary-type-assertion.test.ts
```diff
@@ -450,6 +450,22 @@ declare const a: T.Value1;
 const b = a as const;
       `,
     },
+    {
+      code: `
+const UNDEFINED = (() => {})() as undefined;
+      `,
+    },
```
JoshuaKGoldberg, line 457: [Testing] There's no need to store these values in variables - other than to test that storing in a variable isn't broken. Let's split this into two, simpler tests:

```ts
const value = (() => {})() as undefined;
```

```ts
(() => {})() as undefined;
```

...and trim out the unnecessary variables from the rest:

```ts
const f = () => {};
f() as undefined;
```

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -226,6 +226,46 @@ export default createRule<Options, MessageIds>({
...
+      node: TSESTree.TSAsExpression | TSESTree.TSTypeAssertion,
+    ): ts.Type {
+      // Special handling for IIFE: extract the function's return type
+      if (isIIFE(node.expression)) {
+        const callee = node.expression.callee;
+        const functionType = services.getTypeAtLocation(callee);
+        const signatures = functionType.getCallSignatures();
+
+        if (signatures.length > 0) {
+          const returnType = checker.getReturnTypeOfSignature(signatures[0]);
```
JoshuaKGoldberg, line 251: [Testing] This code hardcodes the return type to what's in `signatures[0]`, not the specific resolved signature in case of an overload:

```ts
interface Overloaded {
  (): undefined; // or void?
  (value: string): void; // or undefined?
}

(((value) => { }) as Overloaded)("") as undefined;
```

I couldn't find a way to break this rule with overloads (phew!) - but these cases should be added as tests to prevent regressions. Overload signatures are a mess to deal with and tend to break over time.

packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts
```diff
@@ -226,6 +226,46 @@ export default createRule<Options, MessageIds>({
...
+
+          // If the function has no explicit return type annotation and returns undefined,
+          // treat it as void (TypeScript infers () => {} as () => undefined, but it should be void)
+          if (
+            callee.returnType == null &&
+            isTypeFlagSet(returnType, ts.TypeFlags.Undefined)
```
JoshuaKGoldberg, line 257: [Testing] Another common soruce of bugs is knowing when to check for a type flag being set vs. something specifically and only being a type. Let's add in at least one test for `undefined | void` just to be safe:

```ts
interface Unioned {
  (): undefined | void;
}

((() => { }) as Unioned)() as undefined;
```

JoshuaKGoldberg (changes requested):
This is looking good, thanks for the PR! I poked at the rule in a bunch of different ways and couldn't figure out amy way to break the new implementation. Nice work 🙂.

Just requesting changes on more thorough testing. This area of TypeScript (inferred types, overloads/signatures) tends to cause bugs in lint rules a lot.

Tamashoo:
@JoshuaKGoldberg 
Thank you for the review!

I’ve added the minimal test cases you suggested and fixed the existing tests.
I’m planning to add an additional test that involves generics before re-requesting review.

If you have any further thoughts on edge cases worth covering, please let me know.
Thank you!

Tamashoo:
I added test cases for overloaded and union return types in 2b0b517.
I also added additional cases covering both generics and overloads in 18316e1, with some minor fixes.

JoshuaKGoldberg (approved):
🙌 wonderful, thanks!

Issue #11416, opened 2025-07-21
Docs: Establish and document expectations around use of AI in contributions


kirkwaiblinger:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [x] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [x] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

Let's be clear about how we'll respond to AI contributions. Roughly, at a minimum, we'd like to establish a policy that makes it clear that we will summarily reject AI contributions that are both overtly AI and of poor quality. 

I think we should also be clear that we will not accept contributions where it becomes the case that we are effectively babysitting someone's Claude instance via the code review process.

Perhaps we should set specific expectations around AI usage falling outside of these obviously bad categories, or perhaps we should simply leave it vague, i.e. "we will use our best judgement" in dealing with AI contributions, or "we expect all contributions to be of human quality, and we expect you to have a full understanding of any contribution you make". At the end of the day, a good AI-assisted contribution is one that we can't really tell was AI-assisted

### Affected URL(s)

https://typescript-eslint.io/contributing/pull-requests

### Additional Info

Examples of recent overt and bad uses of AI:
- https://github.com/typescript-eslint/typescript-eslint/pull/11415
- https://discord.com/channels/1026804805894672454/1394396639833227356/1394969255879118999
- https://github.com/typescript-eslint/typescript-eslint/pull/11435
- https://github.com/typescript-eslint/typescript-eslint/pull/11426
- https://github.com/typescript-eslint/typescript-eslint/pull/11419
- https://github.com/typescript-eslint/typescript-eslint/issues/11830

PR #11832, opened 2025-12-09
feat(scope-manager): add ESLint 10 globals opt-in and guard

michaelgiraldo:
ESLint 10 changed how scope managers are supposed to work.

TypeScript-ESLint didn’t implement the new contract, which is why ESLint 10 crashes with the missing `addGlobals` method.

This PR basically patches that gap:

1. **Implements `scopeManager.addGlobals()`**
   This is the method ESLint 10 expects. Without it, ESLint blows up before running any rule.

2. **Adds a flag: `resolveGlobalVarsInScript`** (default: false)
   ESLint 10 resolves globals differently in script mode.
   **They didn’t want to break everyone on ESLint 8/9, so this flag gates the new behavior.**

3. **Makes injected globals “value-only”**
   Prevents accidental type-binding, which would hide legit TS errors.

4. **Adds cleanup logic for unresolved references**
   ESLint 10 requires tidying up script-mode references; TS-ESLint wasn’t doing that.

That’s it. It’s compatibility plumbing.

---

ESLint 10 changed the rules of the game, and TypeScript-ESLint wasn’t playing by them.

This PR just makes the scope manager stop crashing and behave the way ESLint 10 expects *without* screwing over people still on older ESLint versions.

**If you're upgrading to ESLint 10, this PR is the missing piece that unbreaks the world.**

---

When you move to ESLint 10, turn this on:

```json
{
  "parserOptions": {
    "resolveGlobalVarsInScript": true
  }
}
```

Otherwise script-mode globals won't resolve the way ESLint 10 expects.

---

🍊💾

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1587,21 +1587,21 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@cspell/cspell-bundled-dicts@npm:9.3.1":
-  version: 9.3.1
-  resolution: "@cspell/cspell-bundled-dicts@npm:9.3.1"
+"@cspell/cspell-bundled-dicts@npm:9.4.0":
+  version: 9.4.0
+  resolution: "@cspell/cspell-bundled-dicts@npm:9.4.0"
   dependencies:
     "@cspell/dict-ada": ^4.1.1
     "@cspell/dict-al": ^1.1.1
-    "@cspell/dict-aws": ^4.0.15
+    "@cspell/dict-aws": ^4.0.16
     "@cspell/dict-bash": ^4.2.2
     "@cspell/dict-companies": ^3.2.7
-    "@cspell/dict-cpp": ^6.0.14
+    "@cspell/dict-cpp": ^6.0.15
     "@cspell/dict-cryptocurrencies": ^5.0.5
     "@cspell/dict-csharp": ^4.0.7
     "@cspell/dict-css": ^4.0.18
     "@cspell/dict-dart": ^2.3.1
-    "@cspell/dict-data-science": ^2.0.11
+    "@cspell/dict-data-science": ^2.0.12
     "@cspell/dict-django": ^4.1.5
     "@cspell/dict-docker": ^1.1.16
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg:
👋 @michaelgiraldo thanks for the PR!

The description you've written is _very_ verbose. We have a very limited maintainer bandwidth and deeply reading through the whole thing would take up an unnecessarily large amount of time in it. We're working on docs in https://github.com/typescript-eslint/typescript-eslint/issues/11416 -> https://github.com/typescript-eslint/typescript-eslint/pull/11836. The tl;dr is: regardless of whether you use AI, excessively long descriptions actually harm understandability. Pithiness is good.

@bradzacher and I each took a cursory look at the code and there are a lot more changes than either of us expected for this feature. I'm not sure yet whether this is necessary, but given that the PR description seems to be AI-generated and not well human-reviewed, I don't think we want to spend our limited bandwidth reviewing a PR that might be AI-generated and not well human-reviewed.

I'm converting this to a draft for now. If you're still interested in this feature, please:

* Reduce the PR description to something in line with what a human would normally write
* Double-check the implementation to make sure it's something you can stand by

Feel free to re-request review & un-draft (you should have permissions to, ping me if not) once you've done those things. 

Cheers!

michaelgiraldo:
@JoshuaKGoldberg is the write up is better? any other guidance, coding style, more tests? I want  "stand by" my code with testing coverage.

JoshuaKGoldberg:
Oh sorry I missed the tag! Honestly, this description is better, but it still looks AI-authored and unnecessarily verbose. And it still doesn't follow the template. See recently merged PRs and compare to this.

Also, the strategy of needing an explicit `parserOptions` entry is not something I think we want. Things should "just work" for users regardless of ESLint version. Asking to add some new thing to their ESLint config is a non-starter.

JoshuaKGoldberg:
Closing in favor of #11914, which is the strategy we're looking to go with. Thank you for looking into this!

PR #11834, opened 2025-12-10
feat(type-utils): detect nested any in isUnsafeAssignment to flag unsafe object arguments

### LINKED ISSUE #11801: Enhancement: [no-unsafe-*] check `any`s nested inside object arguments

https://typescript-eslint.io/rules/no-unsafe-argument

### Description

Currently `no-unsafe-argument` doesn't catch object arguments with nested `any` types. For example:

```ts
// Imagine this is a value received from a library.
declare const fromLib: { foo: any };

// Imagine this is a function defined in user-land.
declare const fn: (x: { foo: number }) => string;

// This is unsafe due to the inner `any` type.
fn(fromLib);
```

A real world example of where this comes up is React Router's `useLocation` hook, which returns a `Location` object where the `state` property has type `any`.

Would it make sense to extend this lint rule to check for `any`s nested inside object types?

### Fail

```TypeScript
// Imagine this is a value received from a library.
declare const fromLib: { foo: any };

// Imagine this is a function defined in user-land.
declare const fn: (x: { foo: number }) => string;

// This is unsafe due to the inner `any` type.
fn(fromLib);
```

### Pass

```TypeScript
declare const fromLib: { foo: number };

declare const fn: (x: { foo: number }) => string;

fn(fromLib);
```

### Additional Info

_No response_

OleksandraKordonets:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11801 

## Overview

Extend `isUnsafeAssignment` to walk object types structurally with cycle guards and early exits, so nested `any` in object arguments, unions, tuples, arrays, and index signatures are detected even when shapes are inline or via aliases, not only when the generic bases match. This aligns with the goal of the no-unsafe rules to catch any flowing through call sites and assignments, and it expands on the helper’s current simple top level behavior. All no-unsafe rules that call the helper benefit, including `no-unsafe-argument`, `no-unsafe-assignment`, and, by extension, related rules that rely on the same type-level check, the generic fast path remains for identical bases while inline shapes now flag, and `no-unsafe-assignment` suppresses a duplicate top level diagnostic when a nested property already errors. 

Tests cover shallow and nested objects in `packages/type-utils/tests/isUnsafeAssignment.test.ts` plus the issue’s repro in `packages/eslint-plugin/tests/rules/no-unsafe-argument.test.ts`, a small website config tweak avoids spreading any to satisfy the stricter checks, and performance is guarded with early exits, default lib skipping, and cycle tracking.

### DIFF
--- a/packages/type-utils/src/isUnsafeAssignment.ts
+++ b/packages/type-utils/src/isUnsafeAssignment.ts
@@ -138,6 +138,46 @@ function isUnsafeAssignmentWorker(
     visited.set(type, new Set([receiver]));
   }
 
+  if (checker.isTupleType(type) && checker.isTupleType(receiver)) {
+    const senderElements = checker.getTypeArguments(type);
+    const receiverElements = checker.getTypeArguments(receiver);
+    const length = Math.min(senderElements.length, receiverElements.length);
+
+    for (let i = 0; i < length; i += 1) {
+      const unsafe = isUnsafeAssignmentWorker(
+        senderElements[i],
+        receiverElements[i],
+        checker,
+        senderNode,
+        visited,
+      );
+      if (unsafe) {
+        return { deep: true, receiver, sender: type };
+      }
+    }
+
+    return false;
+  }
+
+  if (checker.isArrayType(type) && checker.isArrayType(receiver)) {
+    const senderElementType = checker.getTypeArguments(type)[0];
+    const receiverElementType = checker.getTypeArguments(receiver)[0];
+
+    const unsafe = isUnsafeAssignmentWorker(
+      senderElementType,
+      receiverElementType,
+      checker,
+      senderNode,
+      visited,
+    );
+
+    if (unsafe) {
+      return { deep: true, receiver, sender: type };
+    }
+
+    return false;
+  }
+
   if (tsutils.isTypeReference(type) && tsutils.isTypeReference(receiver)) {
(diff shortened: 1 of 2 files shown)


OleksandraKordonets:
@JoshuaKGoldberg 
Hi, I wanted to ask for some guidance on the CI lint failures here.
Earlier I fixed the lint errors for this branch in [commit 1d01bd4](https://github.com/typescript-eslint/typescript-eslint/pull/11834/commits/1d01bd47ec967f31e5b6a5f24f60fef6f7d01ea3), but after I added tests, the lint job started failing again. I can fix these, but it would mean touching extra files outside the core of this PR. Before I do that, I wanted to ask what you prefer. Should I keep this PR focused on my solution and leave the other lint issues for a follow up, or is it ok to include those extra lint fixes here so CI passes? I want to follow the project style and avoid unnecessary changes, so any guidance would be very helpful. Thank you

kirkwaiblinger:
Hi @OleksandraKordonets, without diving deeper into the failures you're seeing and such, if there are new lint failures anywhere in the codebase occurring as a result of the changes to the rule logic in this PR, then, generally, yes, we encourage/require them to be fixed! This is a great dogfooding opportunity to ensure that reports make sense. In some cases, disable comments or suppressions are appropriate, but generally we prefer the fixes be made where reasonable.

Does that answer your question or are you referring to something different?

PR #11860, opened 2025-12-18
feat(utils): extend node and token type checks to accept string literals

9romise:
## PR Checklist

## Overview

Previously, functions like `isNodeOfType()` required using enum values (e.g., `AST_NODE_TYPES.ArrayPattern`). This PR updates them to also accept plain string literals like `'ArrayPattern'`.

Now both styles are valid:
```ts
isNodeOfType('ArrayPattern') // New, shorter style, and editor suggests all node type strings
isNodeOfType(AST_NODE_TYPES.ArrayPattern) // Existing style still works
```

### DIFF
--- a/packages/utils/src/ast-utils/helpers.ts
+++ b/packages/utils/src/ast-utils/helpers.ts
@@ -1,26 +1,29 @@
 import type { AST_NODE_TYPES, AST_TOKEN_TYPES, TSESTree } from '../ts-estree';
 
+type ASTNodeTypes = `${AST_NODE_TYPES}` | AST_NODE_TYPES;
+type ASTTokenTypes = `${AST_TOKEN_TYPES}` | AST_TOKEN_TYPES;
+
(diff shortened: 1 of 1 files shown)


bradzacher (changes requested):
We purposely use the enum references and not raw strings for a multitude of reasons.

TypeScript enums should not be treated as transparent bags of values - instead they should be treated as opaque, nominal data.

Making code slightly shorter at the cost of these goals is not something we're interested in.

Additionally as per our contributor guide (which you did not tick as having read...) we ask that you only submit PRs for issues that are marked as accepting PRs.

For all these reasons I am going to close this PR.

9romise:
However, in practice, while `AST_NODE_TYPES` and `AST_TOKEN_TYPES` are defined as enums, their values are not arbitrary, each one is specific and meaningful. Furthermore, simply shortening the code is not the full intent of this PR.

But in any case, the final decision is up to you and team.

PR #11885, opened 2025-12-27
feat(eslint-plugin): add namespace to plugin meta

### LINKED ISSUE #11113: Enhancement: Support `plugin.meta.namespace`

### Relevant Package

eslint-plugin

### Description

Currently, this is an invalid ESLint configuration:

```js
import { defineConfig } from "eslint/config";
import { plugin as ts } from "typescript-eslint";

export default defineConfig(
    {
        plugins: {
            ts,
        },
        extends: ["ts/stylistic"],
        rules: {
            "ts/array-type": "off",
        },
    },
);
```

This does not work as I would intuitively expect: it does not disable the `@typescript-eslint/array-type` rule added by the `ts/stylistic` config. Instead, it disables the `ts/array-type` rule but leaves the `@typescript-eslint/array-type` rule enabled.

ESLint recently added the `plugin.meta.namespace` property. A plugin can use this property to specify the namespace it is using for its rule definitions. As far as I understand, the `defineConfig` function will then replace this namespace with the user provided namespace (`ts` in the above example) in all rules of the extended configurations (I think one needs to reference the config via a string, not sure though).

The end result is that the above config should work as one would intuitively expect, i.e. the `array-type` lint is turned off completely.

### Additional Info

I originally opened this as eslint/eslint#19655 as I thought the whole namespace thing is not yet implemented, but I was told that it is and it is now up to the plugins to support it. Hence this enhancement request.

If this enhancement is accepted, I should be able to make the change and send a PR.

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11113

## Overview

Also applies this to `@typescript-eslint/eslint-plugin-internal`. And adds the property to the `SharedConfig.PluginMeta` type from `@typescript-eslint/utils`.

💖

### DIFF
--- a/packages/utils/src/ts-eslint/Config.ts
+++ b/packages/utils/src/ts-eslint/Config.ts
@@ -44,7 +44,10 @@ export namespace SharedConfig {
      */
     name: string;
     /**
-     * The meta.namespace property should match what comes after eslint-plugin- in your package name.
+     * If you followed the classic style to name your package
+     * (e.g. `eslint-plugin-X`, `@X/eslint-plugin`, or `@X/eslint-plugin-Y`),
+     * then this should match the classic namespace inferred for your package
+     * (e.g. `X`, `@X`, or `@X/Y`, respectively).
      */
     namespace?: string;
     /**


packages/utils/src/ts-eslint/Config.ts
```diff
@@ -43,6 +43,10 @@ export namespace SharedConfig {
      * The meta.name property should match the npm package name for your plugin.
      */
     name: string;
+    /**
+     * The meta.namespace property should match what comes after eslint-plugin- in your package name.
```
kirkwaiblinger: What about `@typescript-eslint/eslint-plugin`? Isn't that convention relevant too?

JoshuaKGoldberg: Hmm, yeah. I wasn't sure how to phrase this... 🤔

kirkwaiblinger: There are some docs around this at https://eslint.org/docs/latest/extend/plugins#meta-data-in-plugins... Maybe we just link to that? Idk? 

Seems a little absurd to say "This should match the namespace inferred by eslint's old config lookup system, that now no longer technically exists, but is de facto fossilized as a recommended naming conventions for plugins"

bradzacher: Perhaps something like...
(I hate the "legacy" and "new"/"flat"/"current" so I'm gonna call it "classic")

> If you followed the classic style to name your package (eg `eslint-plugin-X`, `@X/eslint-plugin`, or `@X/eslint-plugin-Y`) then this should match the classic namespace inferred for your package (eg `X`, `@X`, or `@X/Y`, respectively).

JoshuaKGoldberg:
Merging sooner than our normal waiting period to get this in to the release today.

PR #11778, opened 2025-11-23
chore: check variable declaration syntax error against tsNode

fisker:
## PR Checklist

- [ ] Addresses an existing open issue: fixes #000
- [ ] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Same as #11777

> The syntax check should always check against TS node instead of ESTree node before convert, prepare for https://github.com/typescript-eslint/typescript-eslint/issues/11695

No logic changes, just refactor.

Note: this contains changes from #11777, but fixed both `VariableDeclaration`, `VariableDeclarationList`, and `VariableStatement`.

If smaller changes are preferred, I'll rebase after #11777 gets merged.

PR #11910, opened 2025-12-30
fix(typescript-estree): fix syntax check for `using` declaration

fisker:
## PR Checklist

## Overview

In #11778, I accidentally fixed a bug.

```
for (using foo = bar;;);
```

Should not be a syntax error, [TS playground](https://www.typescriptlang.org/play/?ts=5.2.2#code/GYewTgBAFArgzgSwHYHMKhBAvBARgQzAG4iBKIoA), but previously was forbidden.

I found this until Prettier upgrade `typescript-eslint` https://github.com/prettier/prettier/pull/18551

The mistake is [here](https://github.com/typescript-eslint/typescript-eslint/commit/f7ddd95c09bf143756adbbcb335cf68a2b30e2b1#diff-341d9ffb40c62139c7da69b49d5081cbce7eabce79b58fe569de987a91e63c14R1004) I changed it to check no initializer, [original is check has initializer](https://github.com/typescript-eslint/typescript-eslint/commit/f7ddd95c09bf143756adbbcb335cf68a2b30e2b1#diff-341d9ffb40c62139c7da69b49d5081cbce7eabce79b58fe569de987a91e63c14L1023)

I found there is no tests covers these lines, and some code is actually dead.


packages/typescript-estree/src/check-syntax-errors.ts
```diff
@@ -48,15 +48,6 @@ function checkForStatementDeclaration(
         `The variable declaration of a '${loop}' statement cannot have a type annotation.`,
       );
     }
-    if (
-      kind === SyntaxKind.ForInStatement &&
-      initializer.flags & ts.NodeFlags.Using
-    ) {
-      throw createError(
-        initializer,
-        "The left-hand side of a 'for...in' statement cannot be a 'using' declaration.",
```
fisker, line 400: This was moved because the error message should be 

```diff
-  "The left-hand side of a 'for...in' statement cannot be a 'using' declaration.",
+  "The left-hand side of a 'for...in' statement cannot be a 'await using' declaration.",
```

When using `await using`

JoshuaKGoldberg (approved):
> accidentally fixed a bug

Amusing. 👍

bradzacher (changes requested):
Looks good -- but I believe that the check logic will need to move from `packages/typescript-estree/src/convert.ts` to `packages/typescript-estree/src/check-syntax-errors.ts` based on the other PRs we just merged

Issue #10940, opened 2025-03-11
Enhancement: Use TS 7 (tsgo / typescript-go) for type information


JoshuaKGoldberg:
### Before You File a Proposal Please Confirm You Have Done The Following...

- [x] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22enhancement%3A+plugin+rule+option%22) and found none that match my proposal.
- [x] I have searched the [current rule list](https://typescript-eslint.io/rules/#rules) and found no rules that match my proposal.
- [x] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Relevant Package

typescript-estree

### My proposal is suitable for this project

- [x] I believe my proposal would be useful to the broader TypeScript community (meaning it is not a niche proposal).

### Description

TypeScript is getting ported to Go! It's going to be 10x faster! Blazing fast! Speed! 🔥 ⚡

https://devblogs.microsoft.com/typescript/typescript-native-port/

If we could use this to power typed linting, that'd be great for users. Type checking performance is _the_ bottleneck for the vast majority of correctly configured lint rules that use typescript-eslint.

We believe there are currently _at least_ three major difficulties with typescript-eslint using tsgo in 2025:

1. ESLint does not yet support asynchronous parsers: https://github.com/eslint/eslint/issues/15475
    * ~~It's very likely that tsgo will be consumed via native binding APIs (WASM), and so will almost certainly be asynchronous~~
    * ~~typescript-eslint could in theory use something like [`synckit`](https://www.npmjs.com/package/synckit) but that presents data serialization difficulties~~
    * Edit: per https://github.com/typescript-eslint/typescript-eslint/issues/10940#issuecomment-2715502934 and https://github.com/typescript-eslint/typescript-eslint/issues/10940#issuecomment-2715587307, this might not be a blocker for long (thanks @jakebailey and @nicolo-ribaudo for the notes!)
2. tsgo is not stable and is many months away from being stable. It won't likely be the primary stable version of TypeScript within the next ~1-2 major versions of typescript-eslint.
3. ESLint rules in and out of typescript-eslint still rely on the JS-land AST, so even if we had a stable async parser, there's design work to be done communicating AST nodes _and_ type information back from Go/WASM land to JS-land

So, although we'd very much like to be 10x faster (!!), it will take a lot of design exploration and work.

[...]

PR #11918, opened 2026-01-03
feat(eslint-plugin): [no-unnecessary-type-arguments] report inferred required types using assignability API

### LINKED ISSUE #4978: Enhancement: [no-unnecessary-type-arguments] Detect inferred type

https://typescript-eslint.io/rules/no-unnecessary-type-arguments

### Description

Assuming this is feasible, I would like redundant type arguments due to inferred type to be flagged as well. Even if the type argument is required.

The examples below uses a `MatDialog` from `@angular/material/dialog`. Both result in `dialogRef: MatDialogRef<RemoveItemDialogComponent, any>`

```js
export interface ComponentType<T> {
    new (...args: any[]): T;
}
export declare class MatDialogRef<T, R = void> {
   // ...
}
// ...
  open<T, D = any, R = any>(component: ComponentType<T>, config?: MatDialogConfig<D>): MatDialogRef<T, R>;
```

### Fail

```typescript
// Complex case: Inferred type will be the same
const dialogRef = this.matDialog.open<RemoveItemDialogComponent>(RemoveItemDialogComponent);

// Simple case: Is already a string
const a_string = '' as string;
const fn = <T>(arg: T) => arg;
const result = fn<string>(a_string);
```

### Pass

```typescript
// Complex case
const dialogRef = this.matDialog.open(RemoveItemDialogComponent);

// Simpler cases
const fn = <T>(arg: T) => arg;
const a_string = '' as string;

// result is string
const result = fn(a_string);

// Changes '' to string
const result = fn<string>('');
// or (opinionated)
const result: string = fn('');
```

### Additional Info

_No response_

ej-shafran:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4978

## Overview

Updates the `@typescript-eslint/no-unnecessary-type-arguments` rule to disallow explicitly passing type arguments when the type can be trivially inferred without them.

### Invalid code

```ts
function f<T>(x: T) {}
f<number>(10);
declare const x: number;
f<number>(x);
declare function y(): number;
f<number>(y());
enum E { A, B }
f<E>(E.A);
```

### Valid code

```ts
function f<T>(x: T) {}
f<number | null>(10);
f<10>(10);
f(10);
declare const x: number;
f(x);
declare function y(): number;
f(y());
enum E { A, B }
f(E.A);
```

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
@@ -42,10 +42,31 @@ export default createRule<[], MessageIds>({
     const services = getParserServices(context);
     const checker = services.program.getTypeChecker();
 
+    function isEmptyObjectType(type: ts.Type) {
+      if (!(type.flags & ts.TypeFlags.Object)) {
+        return false;
+      }
+
+      if (type.getProperties().length) {
+        return false;
+      }
+
+      if (type.getStringIndexType() || type.getNumberIndexType()) {
+        return false;
+      }
+
+      return true;
+    }
+
     function areTypesEquivalent(a: ts.Type, b: ts.Type) {
-      // If either type is `any` or unresolved,
+      // If either type is `any` (including when they're unresolved) or `{}`,
       // they should be considered equivalent if they're explicitly the same reference
-      if (a.flags & ts.TypeFlags.Any || b.flags & ts.TypeFlags.Any) {
+      if (
+        a.flags & ts.TypeFlags.Any ||
+        b.flags & ts.TypeFlags.Any ||
+        isEmptyObjectType(a) ||
+        isEmptyObjectType(b)
+      ) {
         return a === b;
       }
 
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-arguments.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-arguments.test.ts
@@ -28,6 +28,45 @@ function f<T>(x: T) {}
 f<10>(10);
     `,
     `
+function f<T>(x: T) {}
+declare const x: any;
+f<string>(x);
+    `,
+    `
+function f<T>(x: T) {}
+f<Record<string, boolean>>({});
+    `,
+    `
+function f<T>(x: T) {}
+declare const x: {};
+f<Record<string, boolean>>(x);
+    `,
+    `
+function f<T>(x: T) {}
+declare const x: Record<string, never>;
+f<Record<string, boolean>>(x);
+    `,
+    `
+function f<T>(x: T) {}
+declare const x: any;
+f<{}>(x);
+    `,
+    `
+function f<T>(x: T) {}
+declare const x: {};
+f<any>(x);
+    `,
+    `
+function f<T>(x: T) {}
+interface F {}
+declare const x: {};
+f<F>(x);
+    `,
+    `
+function f<T>(x: T) {}
+f<number[]>([]);
+    `,
+    `
 function f<T = number>(x: T) {}
 f(10);
     `,
@@ -313,6 +352,60 @@ f<number>(x);
       output: `
 function f<T>(x: T) {}
 declare const x: number;
+f(x);
+      `,
+    },
+    {
+      code: `
+function f<T>(x: T) {}
+declare const x: any;
+f<any>(x);
+      `,
+      errors: [{ messageId: 'canBeInferred' }],
+      output: `
+function f<T>(x: T) {}
+declare const x: any;
+f(x);
+      `,
+    },
+    {
+      code: `
+function f<T>(x: T) {}
+declare const x: {};
+f<{}>(x);
+      `,
+      errors: [{ messageId: 'canBeInferred' }],
+      output: `
+function f<T>(x: T) {}
+declare const x: {};
+f(x);
+      `,
+    },
+    {
+      code: `
+function f<T>(x: T) {}
+declare const x: Record<string, never>;
+f<Record<string, never>>(x);
+      `,
+      errors: [{ messageId: 'canBeInferred' }],
+      output: `
+function f<T>(x: T) {}
+declare const x: Record<string, never>;
+f(x);
+      `,
+    },
+    {
+      code: `
+function f<T>(x: T) {}
+interface F {}
+declare const x: F;
+f<F>(x);
+      `,
+      errors: [{ messageId: 'canBeInferred' }],
+      output: `
+function f<T>(x: T) {}
+interface F {}
+declare const x: F;
 f(x);
       `,
     },


ej-shafran:
Okay, I think the basic idea is properly implemented now. All that's left is my question in the "notes" of the description.

As for the first test (which checks an existing identifier's type), I can make it pass with the following diff:

```diff
diff --git a/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts b/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
index 6acccb81b..cbf1deb8a 100644
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
@@ -143,7 +143,16 @@ export default createRule<[], MessageIds>({
             return;
           }
 
-          if (isInferrable(typeArgument, argument)) {
+          const argumentSymbol = services.getSymbolAtLocation(argument);
+
+          if (
+            (argumentSymbol &&
+              areTypesEquivalent(
+                typeArgumentType,
+                checker.getTypeOfSymbol(argumentSymbol),
+              )) ||
+            isInferrable(typeArgument, argument)
+          ) {
             context.report({
               node: typeArgument,
               messageId: 'canBeInferred',
```

But that doesn't handle the second test (an enum's member being inferred as the enum), or other possibilities where it makes sense not to have an explicit declaration of the type argument (for example, calling `f` with the value returned from a function means `T` can be inferred as the returned type of that function).

I wonder if this is too far of a rabbit hole, though - or even, if this is kinda intended behavior for both of these rules; if a user writes:

```ts
function f<T>(x: T) {}
declare function g(): number;
f<number>(g());
```

Should that be flagged by this rule? What about:

```ts
declare function g(): number;
const x: number = g();
```

Should that be flagged by `no-inferrable-types`?

It seems to me like both "yes" and "no" are valid answers to those questions, and I'm not sure which one makes more sense. If "yes", I can try to implement a more involved solution that uses type checking to infer the type of an expression (or maybe something like that already exists?) - in kind of opposite fashion to `getContextualType`. If "no", then I think this PR is basically good to go as-is, though maybe I can add the diff above...

Hopefully I'm not ranting too badly here... 🫠

ej-shafran:
I've updated the description so it's not as overbearing with information about the specific implementation details, since that's changed already and I suspect it may change again.

Right now I've moved from an implementation that extracts `isInferrable` from `no-inferrable-types` to one that uses `checker.getBaseTypeOfLiteralType`, which works pretty well and covers all the examples I've listed in the description and the test cases.

However, there's two cases it doesn't currently cover that I think maybe it should:

```ts
function f<T>(x: T) {}
f<10>(10 as const);
```

Should this error? Particularly - should we error when a user uses both an `as` expression and an explicit type argument (doing that feels like going against the "spirit" of the rule - don't specify the same information twice when once will suffice)? Or maybe just an `as const` expression should negate the usage of `getBaseTypeOfLiteralType`...?

```ts
function g<T>(x: T, y: T) {}
f<number | null>(10, null);
```

Should this error? i.e. should we collect all the different arguments being passed for the type parameter and union them, and compare _that_ with the explicitly passed type argument?

packages/eslint-plugin/tests/rules/no-unnecessary-type-arguments.test.ts
JoshuaKGoldberg, line 1: [Testing] The following features tend to break things with type parameters:

* `extends` constraints
* `=` defaults
* Currying, like `const curried = <Outer>(outer: Outer) => <Inner>(inner: Inner) => ({ outer, inner });`
* Function overloads

Could you add test coverage for these wild and wacky cases, please?

packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
```diff
@@ -61,50 +63,118 @@ export default createRule<[], MessageIds>({
       };
     }
 
+    // TODO: would like checker.areTypesEquivalent. https://github.com/Microsoft/TypeScript/issues/13502
```
JoshuaKGoldberg: At this point:

* `isTypeAssignableTo` _is_ available for usage, so checking `isTypeAssignableTo(a, b)` and `isTypeAssignableTo(b, a)` is something we can do
* TS 5/6 has a roughly frozen API, so anything more is out-of-scope

Either something is supported with the current TS APIs, or it's not something we can plan to support (#10940 - tsgo APIs are beyond our ability to plan).

Meaning: this wouldn't be a TODO. Either we can do it or we can't.

ACK that this is a move of existing code & comments. I see this PR as addressing what the comment was looking for.

JoshuaKGoldberg (changes requested):
Nice! I like where this is going. Thanks for getting started on it! ✨ 

This is a tricky area of logic, so I think we'll need some more comprehensive tests for cases that tend to break things.

packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
ej-shafran: Hmm.

For some reason, when I replace `areTypesEquivalent(a, b)` with `checker.isTypeAssignableTo(a, b) && checker.isTypeAssignableTo(b, a)`...

<details>

<summary>View diff</summary>

```diff
diff --git a/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts b/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
index 8fbd07460..9219fd1e6 100644
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
@@ -145,7 +145,10 @@ export default createRule<[], MessageIds>({
             services.getTypeAtLocation(argument),
           );
 
-          if (areTypesEquivalent(typeArgumentType, argumentType)) {
+          if (
+            checker.isTypeAssignableTo(typeArgumentType, argumentType) &&
+            checker.isTypeAssignableTo(argumentType, typeArgumentType)
+          ) {
             context.report({
               node: typeArgument,
               messageId: 'canBeInferred',
@@ -168,7 +171,12 @@ export default createRule<[], MessageIds>({
       }
 
       const defaultType = checker.getTypeAtLocation(typeParameter.default);
-      if (!areTypesEquivalent(defaultType, typeArgumentType)) {
+      if (
+        !(
+          checker.isTypeAssignableTo(defaultType, typeArgumentType) &&
+          checker.isTypeAssignableTo(typeArgumentType, defaultType)
+        )
+      ) {
         return;
       }
```

</details>

...I get these test failures:

```
 ❯  eslint-plugin  tests/rules/no-unnecessary-type-arguments.test.ts (95 tests | 2 failed) 1312ms

(...)

 FAIL   eslint-plugin  tests/rules/no-unnecessary-type-arguments.test.ts > no-unnecessary-type-arguments > valid > 
import { F } from './missing';
function bar<T = F>() {}
bar<F<number>>();
    
AssertionError: Should have no errors but had 1: [
  {
    ruleId: '@rule-tester/no-unnecessary-type-arguments',
    severity: 2,
    message: 'This is the default value for this type parameter, so it can be omitted.',
    line: 4,
    column: 5,
    nodeType: 'TSTypeReference',
    messageId: 'isDefaultParameterValue',
    endLine: 4,
    endColumn: 14,
    fix: { range: [Array], text: '' }
  }
]

1 !== 0


- Expected
+ Received

- 0
+ 1

 ❯ RuleTester.#testValidTemplate ../rule-tester/dist/RuleTester.js:639:31
 ❯ ../rule-tester/dist/RuleTester.js:410:56

(...)

 FAIL   eslint-plugin  tests/rules/no-unnecessary-type-arguments.test.ts > no-unnecessary-type-arguments > valid > 
type A<T = Element> = T;
type B = A<HTMLInputElement>;
    
AssertionError: Should have no errors but had 1: [
  {
    ruleId: '@rule-tester/no-unnecessary-type-arguments',
    severity: 2,
    message: 'This is the default value for this type parameter, so it can be omitted.',
    line: 3,
    column: 12,
    nodeType: 'TSTypeReference',
    messageId: 'isDefaultParameterValue',
    endLine: 3,
    endColumn: 28,
    fix: { range: [Array], text: '' }
  }
]

1 !== 0


- Expected
+ Received

- 0
+ 1

 ❯ RuleTester.#testValidTemplate ../rule-tester/dist/RuleTester.js:639:31
 ❯ ../rule-tester/dist/RuleTester.js:410:56

(...)
```

Which seem quite strange! Especially the second one. If I add some logging there and run only that test...

<details>

<summary>View diff</summary>

```diff
diff --git a/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts b/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
index 9219fd1e6..59989319f 100644
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
@@ -171,6 +171,9 @@ export default createRule<[], MessageIds>({
       }
 
       const defaultType = checker.getTypeAtLocation(typeParameter.default);
+      console.log(
+        `${checker.typeToString(defaultType)} ${checker.isTypeAssignableTo(defaultType, typeArgumentType) ? 'is' : 'is not'} assignable to ${checker.typeToString(typeArgumentType)}`,
+      );
       if (
         !(
           checker.isTypeAssignableTo(defaultType, typeArgumentType) &&
diff --git a/packages/eslint-plugin/tests/rules/no-unnecessary-type-arguments.test.ts b/packages/eslint-plugin/tests/rules/no-unnecessary-type-arguments.test.ts
index e3f6b7efd..a4119e99e 100644
--- a/packages/eslint-plugin/tests/rules/no-unnecessary-type-arguments.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unnecessary-type-arguments.test.ts
@@ -186,10 +186,13 @@ import { F } from './missing';
 function bar<T = F>() {}
 bar<F<number>>();
     `,
-    `
+    {
+      code: `
 type A<T = Element> = T;
 type B = A<HTMLInputElement>;
     `,
+      only: true,
+    },
     `
 type A<T = Map<string, string>> = T;
 type B = A<Map<string, number>>;
```

</details>

...I see this bizarre conclusion:

```
(...) 

 FAIL   eslint-plugin  tests/rules/no-unnecessary-type-arguments.test.ts > no-unnecessary-type-arguments > valid > 
type A<T = Element> = T;
type B = A<HTMLInputElement>;
    
Error: Oh no! Your test called the following console method:
  * log (1 call)
    > Call 0: "Element is assignable to HTMLInputElement"
 ❯ createComplaint ../../node_modules/console-fail-test/src/complaining/createComplaint.ts:15:16
 ❯ ../../node_modules/console-fail-test/src/cft.ts:53:20
 ❯ ../../node_modules/console-fail-test/src/environments/vitest.ts:21:3

(...)
```

I have no idea why this is happening, and it seems counterintuitive to me. If I write the following code in some file, say `scratch.ts`:

```ts
declare function f(a: HTMLInputElement): void;
declare const x: Element;
f(x);
```

I get an error `Argument of type 'Element' is not assignable to parameter of type 'HTMLInputElement'.`.

So it seems to me like something weird is going on. Would appreciate some help so I can keep pushing this PR forward! And of course, thank you so much for your review and your time.

JoshuaKGoldberg: Ah, this old friend! Looking in the debugger, types `a` and `b` as passed to `areTypesEquivalent` do indeed have `.asymbol.escapedName: 'Element'` and `.symbol.escapedName: 'HTMLElement'` as described. But:

* If you look at their `.symbol.members`, they're both `Map() {size: 0}`
* Their `declarations` are both `[NodeObject]` (size 1) where the node's source file is at path `'.../typescript-eslint/node_modules/@types/react/global.d.ts'`

Looking at `node_modules/@types/react/global.d.ts`:

```ts
/*
React projects that don't include the DOM library need these interfaces to compile.
React Native applications use React, but there is no DOM available. The JavaScript runtime
is ES6/ES2015 only. These definitions allow such projects to compile with only `--lib ES6`.

Warning: all of these interfaces are empty. If you want type definitions for various properties
(such as HTMLInputElement.prototype.value), you need to add `--lib DOM` (via command line or tsconfig.json).
*/

...

interface Element {}
...
interface HTMLInputElement extends HTMLElement {}
```

Adding the DOM types to the test case get it to pass:

```ts
languageOptions: {
  parserOptions: {
    project: './tsconfig.lib-dom.json',
    projectService: false,
  },
},
```


I.e. it's a problem with how the backing test infra sets up types, not the rule.

Does that make sense?

ej-shafran: Ah!!! Wow, glad you caught that. I'll modify the test case and make the change to `areTypesEquivalent` 😸

JoshuaKGoldberg: Great! This comes up surprisingly often, it's on my _"check on instinct"_ list. 😄

JoshuaKGoldberg:
👋 Just checking in @ej-shafran did you mean to re-request my review?

ej-shafran:
@JoshuaKGoldberg whoops, I think I did!

JoshuaKGoldberg:
> Should this error? i.e. should we collect all the different arguments being passed for the type parameter and union them, and compare _that_ with the explicitly passed type argument?

Oh, and - I think the PR is fine as-is. This is pretty tricky logic and expanding it a lot strikes me as kind of delicate. IMO it'd be best to go with what's working now, and let folks file a follow-up if they really want. Good questions!

alumni:
@JoshuaKGoldberg @bradzacher could we undo this PR to fix #12158?

Seems it's causing quite a lot of false positives. Since the rule is fixable, it removes useful type information from the code.

bradzacher:
Please keep all discussion limited to the issue instead of commenting on closed PRs.

PR #11905, opened 2025-12-29
chore: extract syntax check (part 2)

fisker:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11695
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

Second part of ##11695
Follow up #11748

PR #11923, opened 2026-01-05
chore: extract syntax check (part 3)

### LINKED ISSUE #11695: Repo: Extract AST check from `Converter`

fisker:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11695

## Overview

Third part of #11695
Follow up #11748 #11905

### DIFF
--- a/packages/typescript-estree/src/check-syntax-errors.ts
+++ b/packages/typescript-estree/src/check-syntax-errors.ts
@@ -394,8 +394,7 @@ export function checkSyntaxError(tsNode: ts.Node): void {
         node.arguments.length !== 2
       ) {
         throw createError(
-          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Bug?
-          node.arguments[2] ?? node,
+          node.arguments.length > 1 ? node.arguments[2] : node,
           'Dynamic import requires exactly one or two arguments.',
         );
       }


packages/typescript-estree/src/check-syntax-errors.ts
```diff
@@ -357,6 +358,49 @@ export function checkSyntaxError(tsNode: ts.Node): void {
...
+
+    case SyntaxKind.CallExpression:
+      if (
+        node.expression.kind === SyntaxKind.ImportKeyword &&
+        node.arguments.length !== 1 &&
+        node.arguments.length !== 2
+      ) {
+        throw createError(
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Bug?
+          node.arguments[2] ?? node,
```
fisker: Is this a bug in the `no-unnecessary-condition` rule? Why it assume `node.arguments[2]` is not nullish?

bradzacher: cos array indexes in TS are not typed as nullish.
the rule operates solely on types.

we could turn on [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess) to fix this if we wanted but it does come with a bit of burden.

PR #11927, opened 2026-01-05
chore: refactor `ModuleDeclaration` syntax check against TS node

fisker:
## PR Checklist

## Overview

Hope we have tests cover this part, so I won't make mistakes. 😄


packages/typescript-estree/src/check-syntax-errors.ts
```diff
@@ -400,6 +400,40 @@ export function checkSyntaxError(tsNode: ts.Node): void {
...
+      if (ts.isStringLiteral(node.name)) {
+        return;
+      }
+
+      if (node.body == null) {
+        throw createError(node, 'Expected a module body');
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Fixme: confirm if it's possible
+      if (node.name.kind !== ts.SyntaxKind.Identifier) {
```
bradzacher, line 586: The possible case is `namespace 'foo' {}`
Because `module 'foo' {}` is valid.

But above (on L421) you guarded against this case with a `return`.

it also looks like you already have this exact check on L411?

fisker: > it also looks like you already have this exact check on L411?

L411 is for `global`?

---

It seems that it can be another type.

```
module 1 {}
```

https://www.typescriptlang.org/play/?ssl=1&ssc=12&pln=1&pc=1#code/LYewJgrgNgpgBARjgbwL5A


I didn't change because it worked this way before.


Anyway, TypeScript already dropped support for `module identifier {}` https://github.com/microsoft/TypeScript/pull/62876. Let's keep it this way?

bradzacher: Oh sorry I didn't notice that 411 was nested in a block.

bradzacher (dismissed):

bradzacher:
Merge master and resolve the conflicts and we can merge this!

PR #11689, opened 2025-10-09
fix(typescript-estree): forbid invalid modifiers in object methods

fisker:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11687
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

PR #11931, opened 2026-01-06
fix(typescript-estree): forbid invalid modifier in object expression

### LINKED ISSUE #11688: Enhancement: Forbid invalid modifiers in object properties

### Relevant Package

typescript-estree

### Description

Similar to https://github.com/typescript-eslint/typescript-eslint/issues/11655, but for properties.

```ts
({abstract property: 1});
({async property: 1});
({declare property: 1});
({export property: 1});
({override property: 1});
({private property: 1});
({protected property: 1});
({public property: 1});
```

### Additional Info

_No response_

fisker:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11688

### DIFF
--- a/packages/typescript-estree/src/check-modifiers.ts
+++ b/packages/typescript-estree/src/check-modifiers.ts
@@ -131,7 +131,7 @@ function nodeHasIllegalDecorators(
 function getModifiers(node: ts.Node): ts.Modifier[] {
   return (
     // @ts-expect-error intentional to access `node.modifiers` instead of `ts.getModifiers(node)` to access all modifiers
-    ((node.modifiers as ts.Modifier[]) ?? []).filter(
+    ((node.modifiers as ts.Modifier[] | undefined) ?? []).filter(
       (modifier): modifier is ts.Modifier => !ts.isDecorator(modifier),
     )
   );
@@ -338,7 +338,7 @@ export function checkModifiers(node: ts.Node): void {
       modifier.kind === SyntaxKind.ProtectedKeyword ||
       modifier.kind === SyntaxKind.PrivateKeyword
     ) {
-      for (const anotherModifier of getModifiers(node) ?? []) {
+      for (const anotherModifier of getModifiers(node)) {
         if (
           anotherModifier !== modifier &&
           (anotherModifier.kind === SyntaxKind.PublicKeyword ||


packages/typescript-estree/src/check-modifiers.ts
```diff
@@ -128,6 +128,15 @@ function nodeHasIllegalDecorators(
   );
 }
 
+function getModifiers(node: ts.Node): ts.Modifier[] {
+  return (
+    // @ts-expect-error intentional to access `node.modifiers` instead of `ts.getModifiers(node)` to access all modifiers
+    ((node.modifiers as ts.Modifier[]) ?? []).filter(
```
fisker: @bradzacher `ts.getModifiers()` can't access `declare` modifier in `({declare a=1})`, is it fine to access `node.modifiers` directly?

https://github.com/typescript-eslint/typescript-eslint/pull/11689#discussion_r2416713772

packages/typescript-estree/src/check-modifiers.ts
```diff
@@ -128,6 +128,15 @@ function nodeHasIllegalDecorators(
...
 }
 
+function getModifiers(node: ts.Node): ts.Modifier[] {
+  return (
+    // @ts-expect-error intentional to access `node.modifiers` instead of `ts.getModifiers(node)` to access all modifiers
+    ((node.modifiers as ts.Modifier[] | undefined) ?? []).filter(
+      (modifier): modifier is ts.Modifier => !ts.isDecorator(modifier),
+    )
+  );
+}
```
bradzacher, line 138: question -- why did you do this instead of using the existing `getModifiers` function?

fisker: https://github.com/typescript-eslint/typescript-eslint/pull/11931#discussion_r2664342563

packages/typescript-estree/src/check-modifiers.ts
bradzacher: https://typescript-eslint.io/play/#ts=5.9.3&fileType=.tsx&code=BQbwBAJgpgxgNgQwE5TAsBeMBGMBfASgCgg&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

looks like we currently have a parser error on that code.
so we probably don't need to do this?

fisker: Mistake `({declare a:1})` (colon)

fisker:
```
p = ts.createSourceFile('foo.ts', '({async a:1})').statements[0].expression.expression.properties[0]
ts.getModifiers(p)
undefined
```

bradzacher: the problem is that this property was purposely removed from the TS API types in TS4.9
accessing it is, ultimately, sketchy -- which is why the `getModifiers` util was created.

Should we just ignore these entirely if TS is choosing to not give us an official way to access them?

fisker: What is the worst result that can be? TypeScript removes `.modifiers`?

How about this, switch back to `ts.getModifiers()`, but make an exception for object properties, if `.modifiers` is removed one day, the syntax check for object properties just silently ignore them. What do you think?

packages/typescript-estree/src/check-modifiers.ts
JoshuaKGoldberg: @bradzacher WDYT?

fisker:
@JoshuaKGoldberg https://github.com/typescript-eslint/typescript-eslint/pull/11931#discussion_r2666662391 not addressed.

fisker:
https://github.com/typescript-eslint/typescript-eslint/issues/12027

JoshuaKGoldberg:
Ah gotcha, I misinterpreted. thanks!

PR #11937, opened 2026-01-06
chore(typescript-estree): remove `createError`

fisker:
## PR Checklist

## Overview

Construct error with `new` directly


fisker:
Since `TSError` is a public API, I added a subclass instead of changing the signature, we can remove the legacy interface in next major release.

[Search result](https://github.com/search?q=TSError+%40typescript-eslint%2Ftypescript-estree&type=code&p=2) shows no one actually constructs it, should we change it directly?

StyleShit:
Hey @fisker! Just checking in, is this still something you have the time/energy to work on?

> [Search result](https://github.com/search?q=TSError+%40typescript-eslint%2Ftypescript-estree&type=code&p=2) shows no one actually constructs it

I'm not familiar enough with this package, but I wouldn't rely on github search. It doesn't include private repos that might be using this API

fisker:
>  but I wouldn't rely on github search.


That's a question, it's not my decision to make.

StyleShit:
CC'ing @typescript-eslint/triage-team 

But I think let's discuss this in an issue

Issue #11840, opened 2025-12-11
Bug: [prefer-optional-chain] unsafe fix when the chain ends with a null comparison (a.b === null)


mdm317:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.
- [x] I have [searched for related issues](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3A%22package%3A+eslint-plugin%22) and found none that matched my issue.
- [x] I have [read the FAQ](https://typescript-eslint.io/linting/troubleshooting) and my problem is not listed.

### Playground Link

https://typescript-eslint.io/play/#ts=5.9.3&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3gMyUgXPAN7wBGshKArgLakgzwA%2B81EE8Avs-FSqNgCWKEMADcAKEHZ4AClxJ4AXhW9%2BIISODMWCgHTlGKpayrsAlMXgB6a-DRUA5o5CY08DIoUB%2BA7GWqbBAANPAA7gAWgmAR8ILuKEhYIACOVIIAblAQICgYehLcIBBoCEQS8Dh4fjB6ngCiAB4ADqi5GILZsuZiNnZQ2BgM8FBgYCDNHSiOOIKNoRhR7mAwUGgRrnEyAAYK23Hu23wCwqLbBZwSEkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uii9o-aJHBgAviBDqgA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
declare const foo: { bar: number | null } | undefined;
if (foo === undefined || foo.bar === null) { // suggests to foo?.bar === null, which is not equivalent.
} else {
  foo.bar.toExponential(); // after accepting fix, this crashes if `foo` is `undefined`.
}
```

### ESLint Config

```javascript
{
  "rules": {
    "@typescript-eslint/prefer-optional-chain": "error"
  }

}
```

### tsconfig

```jsonc
{
  "compilerOptions": {
[...]

PR #11942, opened 2026-01-08
fix(eslint-plugin): [prefer-optional-chain] make unsafe fixes suggestions

### LINKED ISSUE #11917: Bug: [prefer-optional-chain] invalid fix with array access by index

### Playground Link

https://typescript-eslint.io/play/#ts=5.9.3&fileType=.tsx&code=JYOwLgpgTgZghgYwgAgGIHt3IN4FgBQyyARnFAFzIgCuAtsdANwEC%2BBBM1ICYw6IyWnFAAKGJkoYsAHyrUANvIA0ycegDOkzAG0AugEpKxTPIhwBeQsmAxkYzMgBkj1ZgB0pKMgC8v1xu0ABl0Afg8yfRwCIiIoCDBqKAEwKGoIaOQ2KziEpNU4eXV0-BYgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uii9o-aJHBgAviHVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eFYDArugDTg10NM8AOXapUAYQAW6aAGsylDul4BfECqA&tokens=false

### Repro Code

```TypeScript
interface Foo {
  bar: number;
}

function main(foo: Foo | null, foos: Foo[]): boolean {
  if (foo && foo.bar === foos[0]?.bar) {
    return true
  }
  return false
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/prefer-optional-chain": "error"
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strict": true,
    "strictNullChecks": true
  }
}
```

### Expected Result

No error or autofix

### Actual Result

Autofix provides an invalid result:

```ts
function main(foo: Foo | null, foos: Foo[]): boolean {
  if (foo?.bar === foos[0]?.bar) {
    return true
  }
  return false
}
```

`main(null, [])` was `false` before the fix but now it's `true`

### Additional Info

Happens after upgrading to `8.51.0`, probably introduced with https://github.com/typescript-eslint/typescript-eslint/pull/11835

JEmilyKoo:
…r #11917
## PR Checklist

- [x] Addresses an existing open issue: fixes #11917

## Overview

This draft PR addresses issue #11917. The `prefer-optional-chain` rule currently auto-fixes array index access (e.g., `foo && foo[0]`), which can change runtime behavior (unsafe fix).

I have added a failing test case to reproduce the issue.
The test expects a `suggestion` instead of an `auto-fix` for array access patterns, but currently fails because the rule still attempts to auto-fix it.

I will modify the rule implementation to downgrade these specific unsafe fixes to suggestions.

💖

### DIFF
--- a/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
+++ b/packages/eslint-plugin/src/rules/prefer-optional-chain-utils/analyzeChain.ts
@@ -129,6 +129,37 @@ function isValidOrLastChainOperand(
 // I can't think of a good way to reuse the code here in a way that will preserve
 // the type safety and simplicity.
 
+function isUnsafeEqualityCheck(
+  parserServices: ParserServicesWithTypeInformation,
+  lastChainOperand: LastChainOperandForReport,
+  operator: '&&' | '||',
+): boolean {
+  if (operator !== '&&') {
+    return false;
+  }
+
+  if (
+    lastChainOperand.comparisonType !== ComparisonType.Equal &&
+    lastChainOperand.comparisonType !== ComparisonType.StrictEqual
+  ) {
+    return false;
+  }
+  const rhsNode = lastChainOperand.comparisonValue;
+  if (rhsNode.type === AST_NODE_TYPES.ChainExpression) {
+    return true;
+  }
+
+  const type = parserServices.getTypeAtLocation(rhsNode);
+  const constituents = unionConstituents(type);
+
+  return constituents.some(t =>
+    isTypeFlagSet(
+      t,
+      ts.TypeFlags.Undefined | ts.TypeFlags.Any | ts.TypeFlags.Unknown,
+    ),
+  );
+}
+
 type OperandAnalyzer = (
   parserServices: ParserServicesWithTypeInformation,
   operand: ValidOperand,
@@ -532,13 +563,21 @@ function getReportDescriptor(
       start: sourceCode.getLocFromIndex(reportRange[0]),
     },
     messageId: 'preferOptionalChain',
-    ...getFixOrSuggest({
-      fixOrSuggest: useSuggestionFixer ? 'suggest' : 'fix',
-      suggestion: {
-        fix,
-        messageId: 'optionalChainSuggest',
-      },
-    }),
+    ...(operator === '&&' &&
+    lastChain &&
+    isUnsafeEqualityCheck(
+      parserServices,
+      lastChain as LastChainOperandForReport,
+      operator,
+    )
+      ? {}
+      : getFixOrSuggest({
+          fixOrSuggest: useSuggestionFixer ? 'suggest' : 'fix',
+          suggestion: {
+            fix,
+            messageId: 'optionalChainSuggest',
+          },
+        })),
   };
 
   interface FlattenedChain {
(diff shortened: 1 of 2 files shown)


kirkwaiblinger (changes requested):
Hi @JEmilyKoo! Unfortunately, I don't think this seems to be the correct resolution. Assuming @bradzacher and I are aligned (I think we're saying the same thing in https://github.com/typescript-eslint/typescript-eslint/issues/11917#issuecomment-3706620498 and https://github.com/typescript-eslint/typescript-eslint/issues/11917#issuecomment-3761305139), the intended resolution is that the reports added in https://github.com/typescript-eslint/typescript-eslint/pull/11533 (i.e. `a && a.b === c` => `a?.b === c`) should be changed to suggestions, rather than fixes.

Hypnosphi:
@kirkwaiblinger that wouldn't be a correct suggestion though, because `a && a.b === c` is actually `a && (a.b === c)`

kirkwaiblinger:
> @kirkwaiblinger that wouldn't be a correct suggestion though, because `a && a.b === c` is actually `a && (a.b === c)`

We're familiar with the precedence here... The intention behind this is/was that the following two functions have the identical behavior

```ts
function f(a?: { b: string }, c: string) {
  if (a && a.b === c) {
    console.log('hello!');
  }
}

function g(a?: { b: string }, c: string) {
  if (a?.b === c) {
    console.log('hello!');
  }
}
```

This is subject to some conditions on the types involved:
1. `c` must not be possibly-`undefined` (the biggest wrinkle with #11917, since `c` is "undetectably" possibly-`undefined`)
2. Several other nuanced combinations of possibly-`undefined`/possibly-`null` and `==`/`===` that the rule doesn't currently always get right, e.g. https://github.com/typescript-eslint/typescript-eslint/issues/11840

As noted elsewhere I _personally_ don't like this suggestion[^1] but this is a thing that has been previously discussed and decided, eg https://github.com/typescript-eslint/typescript-eslint/issues/7170

[^1]: indeed, _because_ of its sensitivity to the TS types being correct at runtime in a situation where they often aren't

Hypnosphi:
https://github.com/typescript-eslint/typescript-eslint/issues/7170 talks about comparison with a literal which is fine because it never can turn out to be nullish in runtime

kirkwaiblinger:
@Hypnosphi I don't think that's my interpretation of #7170, but that is an interesting idea. Let's move this conversation to #11917 please for visibility.

kirkwaiblinger:
Hi @JEmilyKoo!

Do you have time to address the feedback in https://github.com/typescript-eslint/typescript-eslint/pull/11942#pullrequestreview-3726682205? I don't want to leave this PR hanging.

JEmilyKoo:
> Hi @JEmilyKoo!
> 
> Do you have time to address the feedback in [#11942 (review)](https://github.com/typescript-eslint/typescript-eslint/pull/11942#pullrequestreview-3726682205)? I don't want to leave this PR hanging.

Happy to, @kirkwaiblinger! I'll work on those adjustments and update the PR shortly.

kirkwaiblinger:
Closing as stale, since there hasn't been progress here in a while, but feel free to ping us if you end up picking this up again, and we can reopen. Thanks for your work on this!

PR #11954, opened 2026-01-12
fix(eslint-plugin): [prefer-readonly-parameter-types] preserve type alias infomation

### LINKED ISSUE #11725: Bug: [@typescript-eslint/prefer-readonly-parameter-types] Can't ignore inlined union type

### Playground Link

https://typescript-eslint.io/play/#ts=5.9.2&fileType=.tsx&code=JYOwLgpgTgZghgYwgAgGrAgdwMpgJ4A2KA3sjBgQCYCMAXMiAK4C2ARtMgL4BQ%2BADigBicAgGc8yALzJGIShHIgIlZAB8GjAgTVkRolOoDkhgNzduoSLEQoAShASMoo4ADcIAQShQ4eADwAKgB83MjIEAAekHKiyF4%2B-gE69nCUAPYgBHjxvoFByQ5OLu45iUH5pDy8eALI9gDmwKJWyriEEHlSGmwcAGTIpAD6g1AQjc3QrfhEAEI%2BcvRJnGbVtW1EAApQaXydkqE6AQfqDU0tlOsdwcd1hc5unt65SSdjZ5MX01f56sJieCEwr89HgVggMs0yDAugAKA5wKD1OhoDA4L4AGnhiIATPRLlsdn50FhLiEAJT0VxpYAqSQVA7gkCiNJEAB0BDS9RhCKR6OQPOxZO4PCAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0Wls4CGAEwD2TeIVrEB0AQFtE%2BXrSKlUGANrhsOyL2gjokADTadWHZcgD48EQHd0YLZcsXXryFwNynXykgmZh66TPKIThqQAMpESAAKBsQmUABqlIj2sYSBwSEAunmWAL5FEIWupToVYKXFQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
interface ViewStyle { field1: number }
type Falsy = undefined | null | false | '';

interface RecursiveArray<T>
  extends Array<T | ReadonlyArray<T> | RecursiveArray<T>> { }

type RegisteredStyle<T> = number & { __registeredStyleBrand: T };

type StyleProp<T> =
  | T
  | RegisteredStyle<T>
  | RecursiveArray<T | RegisteredStyle<T> | Falsy>
  | Falsy;

const ff = (
  arg1: ViewStyle,
  arg2: StyleProp<ViewStyle>
): void => {
  console.log(arg1, arg2)
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/prefer-readonly-parameter-types": [
        "error",
      {
          "allow": [
          {
              "from": "file",
              "name": ["StyleProp", "ViewStyle"
            ]
          }
        ]
      }
    ]
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

I expected ESLint won't issue `Parameter should be a read only type` for `StyleProp<ViewStyle>`.

### Actual Result

ESLiint issued `Parameter should be a read only type` for `StyleProp<ViewStyle>` until I changed my `allow` option to be
```
allow: [
            {
              from: 'file',
              name: ['RecursiveArray', 'RegisteredStyle', 'ViewStyle']
            }
          ]
```

### Additional Info

`StyleProp` and `ViewStyle` are types / interfaces from `react-native`.

Tamashoo:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11725

## Overview

Refactored prefer-readonly-parameter-types to preserve type alias information by extracting getParameterType() helper that retrieves types from annotation nodes. This prevents type aliases from being inlined into union types, ensuring aliasSymbol is available for allow-list matching. Added test coverage for type alias handling and type inference paths.

👍


packages/eslint-plugin/src/rules/prefer-readonly-parameter-types.ts
```diff
@@ -84,6 +85,47 @@ export default createRule<Options, MessageIds>({
...
+    ): ts.Type {
+      const checker = services.program.getTypeChecker();
+      if (actualParam.typeAnnotation?.typeAnnotation) {
+        // Get type from annotation node to preserve aliasSymbol
+        const tsTypeNode = services.esTreeNodeToTSNodeMap.get(
+          actualParam.typeAnnotation.typeAnnotation,
+        ) as ts.TypeNode;
+        return checker.getTypeFromTypeNode(tsTypeNode);
+      }
+      return services.getTypeAtLocation(actualParam);
```
JoshuaKGoldberg: [Refactor] Nit: we can remove the `const checker` creation and some indentation by returning early.

```suggestion
      if (!actualParam.typeAnnotation?.typeAnnotation) {
        return services.getTypeAtLocation(actualParam);
      }

      // Get type from annotation node to preserve aliasSymbol
      const checker = services.program.getTypeChecker();
      const tsTypeNode = services.esTreeNodeToTSNodeMap.get(
        actualParam.typeAnnotation.typeAnnotation,
      ) as ts.TypeNode;
      return checker.getTypeFromTypeNode(tsTypeNode);
```

Not a blocker at all, feel free to ignore this. 🙂

packages/eslint-plugin/src/rules/prefer-readonly-parameter-types.ts
```diff
@@ -84,6 +85,47 @@ export default createRule<Options, MessageIds>({
...
+    }
+
+    function checkIsReadonly(type: ts.Type): boolean {
```
JoshuaKGoldberg: [Bug] The `allow` option for `prefer-readonly-parameter-types` intentionally uses the shared plumbing of [`TypeOrValueSpecifier`](https://typescript-eslint.io/packages/type-utils/type-or-value-specifier). That format is meant to work the same across all rules. I don't think it's ideal to put specific logic around name matching into any one rule.

Could you please move this logic into the `TypeOrValueSpecifier` areas of `type-utils`? (or if I'm wrong that this can be moved, say why?)

packages/eslint-plugin/tests/rules/prefer-readonly-parameter-types.test.ts
JoshuaKGoldberg, line 1: [Testing] Could you add cases for the snippet mentioned in the OP? I think it exercises some interesting quirks of aliases and branded types we'll want to make sure stay supported.

packages/eslint-plugin/src/rules/prefer-readonly-parameter-types.ts
Tamashoo: @JoshuaKGoldberg 
Thank you for reviewing!

Before I start implementing, just to confirm:
is it okay to move this logic into a new file under type-utils?

Thanks.

packages/eslint-plugin/src/rules/prefer-readonly-parameter-types.ts
Tamashoo: I realized this could just be declared inside the if statement, so I moved the declaration there.
[0f7d953](https://github.com/typescript-eslint/typescript-eslint/pull/11954/commits/0f7d9531149d279d7b4f854292dab5b8b21c5532)

packages/eslint-plugin/tests/rules/prefer-readonly-parameter-types.test.ts
Tamashoo: I added a test case in the OP in this commit. [3dde86d](https://github.com/typescript-eslint/typescript-eslint/pull/11954/commits/3dde86d945ea075c20d21e3b59ce2a46d84cfb87)

packages/eslint-plugin/src/rules/prefer-readonly-parameter-types.ts
Tamashoo: I thought it would be better to make the change in the isTypeReadonly function.
So I'm currently implementing it that way and working on the tests.

JoshuaKGoldberg: Either way works 🙂

Tamashoo: Finally, I moved this logic to isTypeReadonly in this commit [d9cf084](https://github.com/typescript-eslint/typescript-eslint/pull/11954/commits/d9cf0845082169a79452a9d54af51c7c55b058cf).

Tamashoo:
I've addressed all the comments. Could you please take another look?
Also, it seems the integration test failure is not caused by this PR.

Thanks!

packages/type-utils/src/isTypeReadonly.ts
```diff
@@ -231,6 +231,23 @@ function isTypeReadonlyRecurser(
   const checker = program.getTypeChecker();
   seenTypes.add(type);
 
+  // If the type has an alias symbol, check it against the allow list first
+  if (type.aliasSymbol && options.allow) {
+    const aliasSymbol = type.aliasSymbol;
```
JoshuaKGoldberg, line 236: [Style] Nit: eventually I'm going to write a lint rule against this. AIs are popping up this unnecessary local variable assignments all over the place. Even if you wrote this by hand (no worries either way) it is pretty much unnecessary IMO.

```suggestion
```

JoshuaKGoldberg: _Ah_ it's because of the `if` statement's type narrowing. Fun. Never mind 🙂

JoshuaKGoldberg:
As for the PR title complaint... huh. Interesting one. I don't know what to do here. Because 99.99%+ of consumers just care about `eslint-plugin`, I'll limit it to that.

Issue #11362, opened 2025-06-30
⚡️ Performance: Potential overhead populating implicit globals that don't appear in files


JoshuaKGoldberg:
### Suggestion

Coming over from #9575 -> #10561: @dmichon-msft created a very tantalizing proof-of-concept that might help improve linting performance. Copying the description here:

> ## Overview
> This PR optimizes `populateGlobalsFromLib` by deduplicating work to calculate the set of variables available in the configured `libs` array, and by only injecting implicit variable definitions for variables whose names appear in the file being analyzed.
> 
> As a consequence of only injecting implicit variables that are actually used in the file, we can avoid the overhead of lint rules spending time processing irrelevant injected globals, for example in `typescript-eslint/no-redeclare` and `typescript-eslint/naming-convention`.

Nobody's had the dedicated time to take that PR to completion. But it'd be really good if we could get it in! That should involve cleaning up the test coverage from the PR, testing it to make sure it's still better for performance, and sending a new PR.

### Additional Info

If you send a PR that builds off the code in #10561, please add a [co-author attribution](https://github.blog/news-insights/product-news/commit-together-with-co-authors/#how-it-works) to credit the original author.

💖

PR #11969, opened 2026-01-17
chore(scope-manager): refactor populateGlobalsFromLib with tests and benchmarks

minyeong981:
## PR Checklist

- [x] Addresses an existing open issue: #11362

## Overview

This PR builds on the performance discussion raised in #11362 and the earlier proof-of-concept explored in #10561 (originally discussed in #9575).

In that earlier work, an optimization of `populateGlobalsFromLib` was proposed but never fully integrated, tested, or benchmarked.

This PR revisits that idea with the goal of making it concrete and reviewable by:

- Extracting lib resolution logic into a dedicated private method (`resolveLibDefinitions`)
- Refactoring `populateGlobalsFromLib` to use `resolveLibDefinitions` for explicit and testable dependency flattening
- Adding snapshot-based tests to lock in the observable behavior of global variable population
- Adding Vitest benchmarks to measure performance across different `lib` configurations (es5, es2018, esnext.full, repeated analysis)

There are **no intentional functional or behavioral changes** in how globals are resolved or exposed.
All existing semantics are preserved and verified via snapshot tests, allowing structural changes to be evaluated safely.

Benchmarks show a small regression in raw throughput in some scenarios.
This appears to be confined to a non-hot-path operation, while the refactor improves code readability and separation of concerns.
The resulting structure provides a clearer foundation for follow-up optimizations if desired.

I would appreciate feedback on:
- Whether the refactored structure aligns with the project's maintainability goals
- Whether the observed performance trade-off is acceptable, or if further optimization should be pursued before merging

### Benchmark Results

**Before:**

Benchmark results before refactor

**After:**

Benchmark results after refactor

<br/>

> This work builds on the original proof-of-concept in #10561 by @dmichon-msft.
> Relevant commits include a co-author attribution to credit the original author.

Co-authored-by: @dmichon-msft

💖

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -21,8 +21,4 @@
   "typescript.preferences.quoteStyle": "single",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "nxConsole.generateAiAgentRules": true,
-  "chat.tools.terminal.autoApprove": {
-    "git ls-tree": true,
-    "awk": true
-  },
 }


bradzacher:
this will need to be rebased on main but otherwise we can land this!
thanks for this!

packages/scope-manager/tests/__snapshots__/populate-globals-from-lib.test.ts.snap
JoshuaKGoldberg, line 1: This would be the largest `*.test.*` file in the codebase by almost 2x (248,958 bytes vs 133,600 bytes for `no-floating-promises.test.ts`). Huge! I'm going to try reducing it by removing empty and falsy values from the snapshot.

JoshuaKGoldberg: Got it down to ~66k bytes in 9a28c91 by filtering out falsy values, compacting the `is*Variable` properties into one, and compacting `{ variable: string }` objects into just `string`.

JoshuaKGoldberg (approved):
ACK that this is a pure internal refactor and doesn't address the original issue. Looks good as-is - thanks for the added test coverage!

packages/scope-manager/tests/__snapshots__/populate-globals-from-lib.test.ts.snap
minyeong981: Thanks for the improvement. I found it helpful to see how reducing redundant structure can make tests smaller without losing meaning. I’ll keep this in mind going forward. 🙏

PR #11974, opened 2026-01-19
feat(eslint-plugin-internal): add prefer-tsutils-methods rule

### LINKED ISSUE #11625: Repo: Write lint rule to prefer tsutils methods over bitwise checks
### Suggestion

Coming over from #11617: it's not well-known that [`ts-api-utils`](https://github.com/JoshuaKGoldberg/ts-api-utils) provides a handful of shortcut/wrapper methods around TypeScript bitwise enums/flags. For example, this method:

```ts
function hasBaseTypes(type: ts.Type): type is ts.InterfaceType {
  return (
    (type.flags & ts.TypeFlags.Object) !== 0 &&
    (((type as ts.ObjectType).objectFlags & ts.ObjectFlags.Interface) !==
      0 ||
      ((type as ts.ObjectType).objectFlags & ts.ObjectFlags.Class) !== 0)
  );
}
```

...can be rewritten as:

```ts
function hasBaseTypes(type: ts.Type): type is ts.InterfaceType {
  return (
    tsutils.isObjectType(type) &&
    tsutils.isObjectFlagSet(
      type,
      ts.ObjectFlags.Interface | ts.ObjectFlags.Class,
    )
  );
}
```

Proposal: let's write a general-purpose lint rule that enforces the latter when possible?

### Additional Info

Eventually this might make sense to move to https://github.com/eslint-community/eslint-plugin-eslint-plugin, perhaps.

💖

Kimsoo0119:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11625
- [x] That issue was marked as accepting prs
- [x] Steps in Contributing were taken

## Overview

Add `prefer-tsutils-methods` rule to `eslint-plugin-internal` that detects direct bitwise flag checks and suggests using ts-api-utils methods instead.

Detects patterns like:
- `type.flags & ts.TypeFlags.X` → `tsutils.isTypeFlagSet(type, ts.TypeFlags.X)`
- `symbol.flags & ts.SymbolFlags.X` → `tsutils.isSymbolFlagSet(symbol, ts.SymbolFlags.X)`
- `type.objectFlags & ts.ObjectFlags.X` → `tsutils.isObjectFlagSet(type, ts.ObjectFlags.X)`

Includes auto-fix support.

## Example Output 
Lint result showing detected violations

👍

### DIFF
--- a/packages/eslint-plugin-internal/src/rules/prefer-tsutils-methods.ts
+++ b/packages/eslint-plugin-internal/src/rules/prefer-tsutils-methods.ts
@@ -6,12 +6,13 @@ import { createRule } from '../util/index.js';
 
 type FlagType = 'ObjectFlags' | 'SymbolFlags' | 'TypeFlags';
 
-interface FlagConfig {
-  flagsProperty: string;
-  method: string;
+function isFlagType(name: string): name is FlagType {
+  return (
+    name === 'TypeFlags' || name === 'SymbolFlags' || name === 'ObjectFlags'
+  );
 }
 
-const FLAG_CONFIGS: Record<FlagType, FlagConfig> = {
+const FLAG_CONFIGS = {
   ObjectFlags: {
     flagsProperty: 'objectFlags',
     method: 'isObjectFlagSet',
@@ -43,9 +44,7 @@ export default createRule({
   },
   defaultOptions: [],
   create(context) {
-    function isTsFlagAccess(
-      node: TSESTree.Node,
-    ): { flagType: FlagType; fullText: string } | null {
+    function isTsFlagAccess(node: TSESTree.Node) {
       if (
         node.type === AST_NODE_TYPES.MemberExpression &&
         node.object.type === AST_NODE_TYPES.MemberExpression &&
@@ -54,15 +53,8 @@ export default createRule({
         node.object.property.type === AST_NODE_TYPES.Identifier
       ) {
         const flagType = node.object.property.name;
-        if (
-          flagType === 'TypeFlags' ||
-          flagType === 'SymbolFlags' ||
-          flagType === 'ObjectFlags'
-        ) {
-          return {
-            flagType,
-            fullText: context.sourceCode.getText(node),
-          };
+        if (isFlagType(flagType)) {
+          return { flagType, fullText: context.sourceCode.getText(node) };
         }
       }
       return null;
@@ -96,7 +88,7 @@ export default createRule({
     function isFlagPropertyAccess(
       node: TSESTree.Node,
       expectedFlagType: FlagType,
-    ): { objectText: string } | null {
+    ) {
       if (
         node.type === AST_NODE_TYPES.MemberExpression &&
         node.property.type === AST_NODE_TYPES.Identifier
@@ -112,10 +104,7 @@ export default createRule({
       return null;
     }
 
-    function analyzeBitwiseAnd(node: TSESTree.BinaryExpression): {
-      flagAccess: { flagType: FlagType; flagText: string };
-      objectText: string;
-    } | null {
+    function analyzeBitwiseAnd(node: TSESTree.BinaryExpression) {
       const { left, right } = node;
 
       const rightFlag = isTsFlagAccess(right) ?? isFlagOrExpression(right);
@@ -123,10 +112,8 @@ export default createRule({
         const leftPropAccess = isFlagPropertyAccess(left, rightFlag.flagType);
         if (leftPropAccess) {
           return {
-            flagAccess: {
-              flagText: rightFlag.fullText,
-              flagType: rightFlag.flagType,
-            },
+            flagText: rightFlag.fullText,
+            flagType: rightFlag.flagType,
             objectText: leftPropAccess.objectText,
           };
         }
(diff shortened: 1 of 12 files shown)


packages/eslint-plugin-internal/src/rules/prefer-tsutils-methods.ts
```diff
@@ -0,0 +1,194 @@
...
+        const { flagAccess, objectText } = analysis;
+        const { flagText, flagType } = flagAccess;
+        const method = FLAG_CONFIGS[flagType].method;
+
+        let nodeToReplace: TSESTree.Node = node;
+        let isNegated = false;
+
+        const parent = node.parent;
+        if (
+          parent.type === AST_NODE_TYPES.BinaryExpression &&
```
JoshuaKGoldberg: [Style] Nit: this _"store a single property in a variable"_ pattern isn't really necessary; I don't think it's worth the extra line.

```suggestion
        if (
          node.parent.type === AST_NODE_TYPES.BinaryExpression &&
```

(here and elsewhere)

Aside: I've been seeing this a lot from AIs/agents and have been confused why they keep defaulting to it. I didn't think it was so common...?

packages/eslint-plugin-internal/src/rules/prefer-tsutils-methods.ts
```diff
@@ -0,0 +1,194 @@
...
+          };
+        }
+      }
+      return null;
+    }
+
+    function analyzeBitwiseAnd(node: TSESTree.BinaryExpression): {
+      flagAccess: { flagType: FlagType; flagText: string };
+      objectText: string;
+    } | null {
```
JoshuaKGoldberg: [Style] Nit: we haven't enforced explicit return types for non-exported functions in a while. Non-trivial ones like this union with a deep object take up a lot of syntax space and IMO make it a little harder to read. Could you please trim out unnecessary return type annotations from this PR? 

```suggestion
    function analyzeBitwiseAnd(node: TSESTree.BinaryExpression) {
```

(here and elsewhere)

A hunch: was this an AI-authored thing? (not a problem at all if so, just curious 🙂)

packages/eslint-plugin-internal/src/rules/prefer-tsutils-methods.ts
```diff
@@ -0,0 +1,194 @@
...
+import { createRule } from '../util/index.js';
+
+type FlagType = 'ObjectFlags' | 'SymbolFlags' | 'TypeFlags';
+
+interface FlagConfig {
+  flagsProperty: string;
+  method: string;
+}
+
+const FLAG_CONFIGS: Record<FlagType, FlagConfig> = {
```
JoshuaKGoldberg: [Style] Nit: no need for this type or type annotation, TypeScript will let us know if anything is mismatched.

```suggestion
const FLAG_CONFIGS = {
```

packages/eslint-plugin-internal/src/rules/prefer-tsutils-methods.ts
```diff
@@ -0,0 +1,194 @@
...
+          return;
+        }
+
+        const analysis = analyzeBitwiseAnd(node);
+        if (!analysis) {
+          return;
+        }
+
+        const { flagAccess, objectText } = analysis;
+        const { flagText, flagType } = flagAccess;
```
JoshuaKGoldberg: [Refactor] I don't see any strong reason to nest the `flagAccess` information. Let's simplify the data structure a little:

```suggestion
        const { flagText, flagType, objectText } = analysis;
```

packages/eslint-plugin-internal/src/rules/index.ts
JoshuaKGoldberg, line 1: [Feature] Now that this is added to the plugin, it'll need to be enabled in the root `eslint.config.mjs`. 

https://github.com/typescript-eslint/typescript-eslint/blob/4d01fa41db40e90376071c5ba2258bbdc0f9371c/eslint.config.mjs#L200-L204

JoshuaKGoldberg (changes requested):
A good start, nice! The only real blocker is that we'll need this running on the repo itself. 

I also was able to get the file down from 195 lines to a little under 180 with some reductions in code verbosity. 

Thanks!

packages/eslint-plugin-internal/src/rules/prefer-tsutils-methods.ts
Kimsoo0119: Just my habit to avoid long dot chains 😅
Fixed to follow the convention!

packages/eslint-plugin-internal/src/rules/prefer-tsutils-methods.ts
Kimsoo0119: AI helped overall, but explicit types are my habit too 😄 Fixed!

packages/eslint-plugin-internal/src/rules/prefer-tsutils-methods.ts
JoshuaKGoldberg: Well in that case I apologize for implying it was from an API 😄 it's a very reasonable habit!

Issue #11204, opened 2025-05-12
Repo: Improve CI performance to avoid throttling


JoshuaKGoldberg:
### Suggestion

According to https://github.com/typescript-eslint/typescript-eslint/actions/metrics/usage, our biggest total minutes in the last month are:

1. `ci.yml` (linux): 17,478 minutes across 425 runs
2. `ci.yml` (windows): 7,610 minutes across 207 runs
3. `semantic-breaking-change-pr.yml`: 613 minutes across 359 runs
4. `semantic-pr-titles.yml`: 282 minutes across 282 runs
5. `a11y-alt-bot.yml`: 201 minutes across 201 runs

17478 / 425 = >40 minutes of CI time per commit. That's quite a lot! We've started to see more GHA throttling recently. Jobs/runs have been getting stuck waiting to start -> then for GHA runners to free up.

Filing this issue as a general one: what can we do to reduce the throttling in CI?

### Additional Info

These more specific issues are split out of or generally related to this one:

* #3223: using ESLint for Markdown linting would reduce the number of jobs per commit by one
* #11203 tracks combining `semantic-pr-titles.yml` and `a11y-alt-bot.yml`. That should improve a teeny bit for reducing our number of runs per commit / event.
* #11205
* #11206

💖

ntdiary:
I've been exploring switching tests to vfs recently (i.e. microsoft/typescript-go#1891), it might also help reduce ci time a bit? 🤔
I mimicked a project parser and a project service parser, for `consistent-type-exports.test.ts`, `vitest` time drops from about 3.5s to around 2.2s:

<img height="584" alt="vitest time image" src="https://github.com/user-attachments/assets/79146d9e-bae0-4268-8059-7721fde09eea" />

JoshuaKGoldberg:
Some kind of VFS work would be _great_, yes please!

JoshuaKGoldberg:
💡 I just had an idea. What if we skip running _rule_ unit tests on Node.js 18 and on Windows? The majority of our CI unit test time is spent in them, but rules generally don't care about the differences between operating systems or Node.js versions.

bradzacher:
> but rules generally don't care about the differences between operating systems or Node.js versions

OS's -- yes they don't care
Node versions -- no they do care

Cos we wouldn't want a rule to use eg a newer JS feature

JoshuaKGoldberg:
Another point to look into: 

[...]

PR #11987, opened 2026-01-21
chore: run vitest only for changed files

StyleShit:
Trying to validate the (great) idea from https://github.com/typescript-eslint/typescript-eslint/issues/11204#issuecomment-3719039296


StyleShit:
ok, seems like it's working

see [this run](https://github.com/typescript-eslint/typescript-eslint/actions/runs/21209992052/job/61015925714?pr=11987) that contained a change in single rule

.github/workflows/ci.yml
```diff
@@ -212,7 +212,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-depth: 2
+          fetch-depth: 0 # So vitest `--changed` will be able to compare against main
```
JoshuaKGoldberg: Blurgh, for long-lived repos like this one that's a lot of commits. But I think we should be able to use `git fetch` on the base SHA to build up precisely enough history for `vitest --changed`.

How about something like this?

```yml
- uses: actions/checkout@v4
  with:
    fetch-depth: 1

- run: |
    git fetch origin ${{ github.event.pull_request.base.sha }} --depth=1
    npx vitest --changed
```

JoshuaKGoldberg (changes requested):
Hell yeah, +1 to this direction from me. I think we can get rid of the full history -which sometimes takes a while IME- by doing a more targeted `git fetch`. WDYT?

Thanks for starting this! 🙏

kirkwaiblinger:
I'm worried about this potentially causing many more test failures to make it to main. How well is this risk mitigated? I'm worried about 
1. dependency changes. the docs say that `package.json` changes cause the whole suite to be rerun but can we be sure to validate how that will work with the monorepo and workspace `package.json`s ?

1. indirectly tested utilities... If I have rule-a and rule-b, both of which rely on utils/some-util.ts, and I change rule-b and some-util.ts, does that mean that rule-a's tests won't be rerun? Or would they? 

1. (lesser concern) changes in upstream workspaces. If one of the other workspaces has changes, this won't incur a package.json change. Will that mean that the upstream workspace of esilnt-plugin wouldn't cause the eslint-plugin tests to be rerun?

Definitely not opposed to experimenting with this but just would like to understand the details

---

Also btw could you pls add the docs link to the description: https://vitest.dev/guide/cli.html#changed

.github/workflows/ci.yml
StyleShit: Ah, great idea!
done

StyleShit:
@kirkwaiblinger, these are some great points

> dependency changes ... how that will work with the monorepo and workspace `package.json`s ?

Well, that's a problematic one...
We can [force vitest not to run](https://vitest.dev/config/forcereruntriggers.html) on specific `package.json` changes (e.g. packages that are not the "current" one), but it involves some ugly configs and copy-pasting the same logic to all `packages/*/vitest.config.mts` (which is not necessarily a bad thing).

It's gonna look something like this:

```TS
// packages/*/vitest.config.mts

forceRerunTriggers: [
  // Root package.json
  path.resolve(import.meta.dirname, '../../package.json'),
  // Current package's package.json
  path.resolve(import.meta.dirname, 'package.json'),

  // ... the same for `vitest.config.mts`
],
```

This would work (tested locally), but it might cause issues when there is a dependency between some packages, no?
Say, for example, when `rule-tester` package updates its deps, it might affect the `eslint-plugin` package, so it might probably be safer to just run everything I guess?

If we really want to, we can write a function that will create an array of all deps based on the `package.json` of the current package, and will build the `forceRerunTriggers` according to it, but it sounds like a great place for bugs and a maintenance nightmare


> indirectly tested utilities.

In theory, this will work
The problem is, we have a barrel file in the `src/util` folder, so any util change triggers everything...
It might be a good idea to drop it

Anyway, tested locally with a newly created util file (without touching the barrel file), imported it directly from 2 rules, committed, changed only the util file, ran `vitest --changed HEAD`, and these 2 rules were the only ones to re-run

> changes in upstream workspaces

Seems like they don't trigger it
I changed the index file of `packages/types`, ran `vitest --changed` inside `packages/eslint-plugin`, and nothing ran

> pls add the docs link

Good idea
Done

StyleShit:
Seems like something in the tests is flaky and I can't see why...

It's failing only in some cases, not sure whether it's related to my code

inaccessible job log

EDIT: seems like they're all failing on windows and node 20

EDIT 2: everything passes on my fork though

https://github.com/StyleShit/typescript-eslint/actions/runs/21387238778/job/61566135800

image


EDIT 3: Never mind, now it passes... probably some GitHub Actions issue?

kirkwaiblinger:
> @kirkwaiblinger, these are some great points
> 
> > dependency changes ... how that will work with the monorepo and workspace `package.json`s ?
> 
> Well, that's a problematic one... We can [force vitest not to run](https://vitest.dev/config/forcereruntriggers.html) on specific `package.json` changes (e.g. packages that are not the "current" one), but it involves some ugly configs and copy-pasting the same logic to all `packages/*/vitest.config.mts` (which is not necessarily a bad thing).
> 
> It's gonna look something like this:
> 
> ```ts
> // packages/*/vitest.config.mts
> 
> forceRerunTriggers: [
>   // Root package.json
>   path.resolve(import.meta.dirname, '../../package.json'),
>   // Current package's package.json
>   path.resolve(import.meta.dirname, 'package.json'),
> 
>   // ... the same for `vitest.config.mts`
> ],
> ```
> 
> This would work (tested locally), but it might cause issues when there is a dependency between some packages, no? Say, for example, when `rule-tester` package updates its deps, it might affect the `eslint-plugin` package, so it might probably be safer to just run everything I guess?
> 
> If we really want to, we can write a function that will create an array of all deps based on the `package.json` of the current package, and will build the `forceRerunTriggers` according to it, but it sounds like a great place for bugs and a maintenance nightmare

I'm a bit confused because at first you talk about forcing vitest _not_ to run tests but then later seem to talk about forcing vitest _to_ run tests... 

In any case, I think a dependency update should most likely run all tests in the repo (especially for things like renovate PRs). Maybe using `yarn.lock` as a rerun trigger would work, since that describes all the deps in the whole repo?

> > indirectly tested utilities.
> 
> In theory, this will work The problem is, we have a barrel file in the `src/util` folder, so any util change triggers everything... It might be a good idea to drop it
> 
> Anyway, tested locally with a newly created util file (without touching the barrel file), imported it directly from 2 rules, committed, changed only the util file, ran `vitest --changed HEAD`, and these 2 rules were the only ones to re-run

👍 That's cool that vitest traverses the module graph to detect the changes. IMO removing the barrel file is not a blocker, so it can easily be considered as a followup

> > changes in upstream workspaces
> 
> Seems like they don't trigger it I changed the index file of `packages/types`, ran `vitest --changed` inside `packages/eslint-plugin`, and nothing ran

Well for what it's worth, that package is just types so maybe that specific change didn't trigger the rerun but a substantive one would? 🤔 But yeah this will be a concern IMO if this doesn't work.

---

Also, not sure if this is mentioned somewhere, but IMO the `main` branch should definitely run everything.

StyleShit:
OK, sorry, bad explanations. lemme try again

> I'm a bit confused because at first you talk about forcing vitest not to run tests but then later seem to talk about forcing vitest to run tests...

By default, `forceRerunTriggers` contains all `package.json`s & `vitest.config`s, causing any change of any of these patterns in the repo to re-run the whole test suite.

I assumed it's due to dependency changes that vitest can't detect by itself, so it's kind of a safe measure (maybe I'm wrong, or maybe that's only partially right)

Anyway, I experimented with this locally and we can override the default `forceRerunTriggers` array and make vitest run only when the current package's `package.json`/`vitest.config` has changed. I'm not sure this is a good idea though... 

> barrel file is not a blocker, so it can easily be considered as a followup

sure, sound good to me

> that package is just types so maybe that specific change didn't trigger the rerun but a substantive one would?

Yeah, my bad explanation again... It was an example of a change in a package that's unrelated to the currently tested package (i.e. `eslint-plugin`) that doesn't cause a test re-run (which is good)

But you made me recheck that, and for some reason it actually really doesn't catch up changes in dependencies 😕
I'm assuming it's because we're technically importing the `dist` of each module from `node_modules`, which is probably excluded by vitest (dependency changes are handled by the `**/package.json` pattern in `forceRerunTriggers`, which we didn't change when changing a package source code in the repo)

Not sure how we can solve this without changing the way things are working right now (AKA reading directly from `src` rather than `dist`)

___

EDIT:

ok, seems like I was right. This is the function that handles the `--changed` calculation:
https://github.com/vitest-dev/vitest/blob/7ce3417b1be71689a784cc7a717bd0bce18e39a3/packages/vitest/src/node/specifications.ts#L122

It essentially gets the changed files from git (they call it `related`), calculates all the dependencies of each test file, and checks whether one of these files has changed (e.g. exists in the `related` array).
The problem is that the dependencies are calculated from the `dist` folder, which is not present in the changed files in git.

___ 

EDIT 2:

We can force vitest to run on every workspace dependency change in some ugly way:

```ts
import { defineConfig, configDefaults } from 'vitest/config';

defineConfig({
	// ...

    test: {
		// ...
		forceRerunTriggers: getForceRerunTriggers(),
    }
});

function getForceRerunTriggers() {
  const dependencies = Object.entries({
    ...packageJson.dependencies,
    ...packageJson.devDependencies,
  })
	.filter(([, value]) => value.startsWith('workspace:'))
	.map(([key]) => {
		key = key.replace('@typescript-eslint/', '');
	
		return `**/packages/${key}/**/*`;
	});

  return [
    ...configDefaults.forceRerunTriggers,
    ...dependencies,
  ];
}
```

not sure if it's worth it at this point...
wdyt?

.github/workflows/ci.yml
JamesHenry: The commit references themselves aren't that expensive, it's the data that gets associated with them that isn't needed most of the time, we recommend the combination of:

```
fetch-depth: 0
filter: 'tree:0'
```

JamesHenry: Then no imperative fetch commands are needed and the checkout step remains efficient

kirkwaiblinger:
I think I personally lean against this... I think the complexity to get it to be granular without enough to benefit us, but conservative enough not to fail to run necessary test appears to be a bit too much as this investigation currently stands, and it will be hard to debug when something goes wrong. This has already been a headache when sometimes nx caching has not worked correctly, and adding another source of potential obscure behavior will make such headaches more painful when they happen. The CI delays just haven't been _that_ painful IMO. 

Up to others, though, if someone wants to advocate for the benefit outweighing the drawbacks.

Regardless of the outcome here, huge thanks to @StyleShit for the deep investigations!!

StyleShit:
I'm closing this to clean up the open PRs
Will reopen if we decide to do it

lishaduck:
> I think I personally lean against this... I think the complexity to get it to be granular without enough to benefit us, but conservative enough not to fail to run necessary test appears to be a bit too much as this investigation currently stands, and it will be hard to debug when something goes wrong.

For the interested, IIRC Josh was pushing this here because we'd gotten it to work well in Flint; we ended up reverting it this week because it wasn't granular enough and had let failing PRs break main. 😞
So yes, probably a good idea to keep a hold on it.

PR #11922, opened 2026-01-05
feat(eslint-plugin): [no-unused-vars] add a fixer to remove unused imports

bradzacher:
## PR Checklist

- [x] Addresses an existing open issue: fixes https://github.com/typescript-eslint/typescript-eslint/issues/11223, closes #11243
- [x] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [x] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview

This PR implements a new option in `no-unused-vars` -- `enableAutofixRemoval` which will allow users to turn on automatic removal of unused variables from their code. If the relevant sub-option is set to `true` then the rule will provide an autofixer to remove the unused code, otherwise the rule will provide a suggestion fixer to do the same thing.

This first cut:
- adds the framework for all the other types of fixers we might add,
- adds `enableAutofixRemoval.imports`:
  - If all specifiers in the import are unused then the entire import is removed.
  - Otherwise just the unused specifiers will be removed.

PR #11990, opened 2026-01-21
fix(eslint-plugin): [no-unused-vars] remove trailing newline when removing entire import

### LINKED ISSUE #11996: Bug: [no-unused-vars] Unused imports fixer creates blank lines when removing entire import statement

### Playground Link

https://typescript-eslint.io/play/#ts=5.8.2&fileType=.tsx&code=JYWwDg9gTgLgBAQTgMyhEcBEA6A9AQ0wG4AoUSWOAIRTQx1wCNiSSBjCAOwGcIAbAKbY%2BEAOYAKKgEoSQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tidmUQAmtAG4BDaKgwBtSImjQO0SABpM4bDkRNRAIyQBBWPg4AzSgA8ASogC2HMfHSYolG8UX5JYfHERgAvuoQ-gC66oH%2BQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
import A from "./a";
import B from "./b";

console.log(B)
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-unused-vars": ["error", {
       "enableAutofixRemoval": { "imports": true }
    }],
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

After applying the fix to remove the unused import of `A`, I expected the corresponding trailing new line to also be removed, such that the file contents become:
```ts
import B from "./b";

console.log(B)
```

### Actual Result

There's now an unnecessary new line at the top of the file:
```ts

import B from "./b";

console.log(B)
```

### Additional Info

As mentioned in https://github.com/typescript-eslint/typescript-eslint/pull/11990#issuecomment-3781914761, before this fixer was added I was using `eslint-plugin-unused-imports`, which [does handle this edge case with blank lines](https://github.com/sweepline/eslint-plugin-unused-imports/issues/7). I do prefer the built-in fixer, but this feature is something that I miss from the other plugin.

MariaSolOs:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11996

## Overview

The fixer introduced in https://github.com/typescript-eslint/typescript-eslint/pull/11922 leaves blank lines when removing the entire import statement. This PR modifies the fixer to remove trailing new lines when the entire statement is unused.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unused-vars.ts
+++ b/packages/eslint-plugin/src/rules/no-unused-vars.ts
@@ -523,6 +523,32 @@ export default createRule<Options, MessageIds>({
       return options;
     })();
 
+    function removeNodeWithTrailingNewline(
+      fixer: TSESLint.RuleFixer,
+      node: TSESTree.Node,
+    ): TSESLint.RuleFix {
+      const sourceCode = context.sourceCode;
+      const { line: startLine } = node.loc.start;
+      const { line: endLine } = node.loc.end;
+      // Expand range: start of first line to start of next line (or EOF)
+      const lineRangeStart = sourceCode.getIndexFromLoc({
+        column: 0,
+        line: startLine,
+      });
+      const lineRangeEnd =
+        endLine < sourceCode.lines.length
+          ? sourceCode.getIndexFromLoc({ column: 0, line: endLine + 1 })
+          : sourceCode.text.length;
+      // If node is the only non-whitespace on its line(s), remove full line(s)
+      if (
+        sourceCode.getText(node) ===
+        sourceCode.text.slice(lineRangeStart, lineRangeEnd).trim()
+      ) {
+        return fixer.removeRange([lineRangeStart, lineRangeEnd]);
+      }
+      return fixer.remove(node);
+    }
+
     function getImportFixer(def: ImportBindingDefinition): {
       fix: TSESLint.ReportFixFunction;
       messageId: MessageIds;
@@ -533,7 +559,7 @@ export default createRule<Options, MessageIds>({
           // just remove entire import declaration
           return {
             messageId: 'removeUnusedImportDeclaration',
-            fix: fixer => fixer.remove(def.node),
+            fix: fixer => removeNodeWithTrailingNewline(fixer, def.node),
           };
 
         case AST_NODE_TYPES.ImportDefaultSpecifier: {
@@ -546,7 +572,7 @@ export default createRule<Options, MessageIds>({
             // declaration
             return {
               messageId: 'removeUnusedImportDeclaration',
-              fix: fixer => fixer.remove(importDecl),
+              fix: fixer => removeNodeWithTrailingNewline(fixer, importDecl),
             };
           }
 
@@ -603,7 +629,7 @@ export default createRule<Options, MessageIds>({
             // declaration
             return {
               messageId: 'removeUnusedImportDeclaration',
-              fix: fixer => fixer.remove(importDecl),
+              fix: fixer => removeNodeWithTrailingNewline(fixer, importDecl),
             };
           }
 
@@ -715,7 +741,7 @@ export default createRule<Options, MessageIds>({
           const importDecl = def.node.parent;
           return {
             messageId: 'removeUnusedImportDeclaration',
-            fix: fixer => fixer.remove(importDecl),
+            fix: fixer => removeNodeWithTrailingNewline(fixer, importDecl),
           };
         }
       }
--- a/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-enableAutofixRemoval.test.ts
+++ b/packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-enableAutofixRemoval.test.ts
@@ -39,7 +39,6 @@ export {};
           },
         ],
         output: `
-
 export {};
         `,
       },
@@ -66,7 +65,6 @@ export {};
           },
         ],
         output: `
-
 export {};
         `,
       },
@@ -93,7 +91,6 @@ export {};
           },
         ],
         output: `
-
 export {};
         `,
       },
@@ -128,7 +125,6 @@ export {};
           },
         ],
         output: `
-
 export {};
         `,
       },
@@ -271,7 +267,6 @@ export {};
           },
         ],
         output: `
-
 export {};
         `,
       },
@@ -559,6 +554,48 @@ export { Used };
         output: `
 import { Used, } from 'module';
 export { Used };
+        `,
+      },
+      {
+        code: `
+/* this is an important comment */ import assert from 'assert';
+        `,
+        errors: [
+          {
+            line: 2,
+            messageId: 'unusedVar',
+          },
+        ],
+        options: [
+          {
+            enableAutofixRemoval: {
+              imports: true,
+            },
+          },
+        ],
+        output: `
+/* this is an important comment */ 
+        `,
+      },
+      {
+        code: `
+import assert from 'assert'; /* this is an important comment */
+        `,
+        errors: [
+          {
+            line: 2,
+            messageId: 'unusedVar',
+          },
+        ],
+        options: [
+          {
+            enableAutofixRemoval: {
+              imports: true,
+            },
+          },
+        ],
+        output: `
+ /* this is an important comment */
         `,
       },
     ],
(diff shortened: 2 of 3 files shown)


JoshuaKGoldberg:
> leaves blank lines when removing the entire import statement

As much as I dislike leaving blank lines around, I don't think this is in the scope of what we'd want to add a lot of code to handle? Detection/practices around blank lines are not in the purview of the rule. The general project stance is that if folks care about that, they should use a separate tool: preferably (to us) a dedicated formatter, or alternately a dedicated lint rule.

If this PR was just a couple lines it'd be a lot more palatable. But at >30 lines, although I appreciate the work that went into it, I don't think we'd want to take on that maintenance burden.

MariaSolOs:
@JoshuaKGoldberg I understand. I do want to mention that before this fixer was added I was using `eslint-plugin-unused-imports`, which [does handle this edge case with blank lines](https://github.com/sweepline/eslint-plugin-unused-imports/issues/7). I do prefer the built-in fixer, but this feature is something that I miss from the other plugin.

> they should use a separate tool: preferably (to us) a dedicated formatter, or alternately a dedicated lint rule.

I agree that whitespace should preferably be always handled with formatters, but in this specific case a tool like Prettier won't remove the new blank line (it just enforces that there are no multiple lines between imports). Moreover lint rules like `import/order` can be configured to prohibit blank lines between import groups, which conflict with the modifications of this fixer.

> Detection/practices around blank lines are not in the purview of the rule.

I again agree with you here, but intuitively I think that the least "disruptive" modification when removing an import statement is to remove the new line.

kirkwaiblinger:
@MariaSolOs this would require an issue first in order to be considered. It's a nontrivial PR for an ordinary feature that doesn't fall under the "minor documentation changes" exception; see https://typescript-eslint.io/contributing/pull-requests

Marking as draft for now, pending issue acceptance.

MariaSolOs:
@kirkwaiblinger fair enough, my bad for not following the right order here. Issue created: https://github.com/typescript-eslint/typescript-eslint/issues/11996

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -523,6 +523,41 @@ export default createRule<Options, MessageIds>({
       return options;
     })();
 
+    function removeNodeWithTrailingNewline(
```
kirkwaiblinger, line 526: I tried to simplify the implementation a bit using `.trim()` rather than thinking about white space and line ending manually...

It's not quite as simple as I'd initially hoped but I wonder if this general idea can be used to further golf the code and make it simpler

```ts
    function removeNodeWithTrailingNewline(
      fixer: TSESLint.RuleFixer,
      node: TSESTree.Node,
    ): TSESLint.RuleFix {
      const fullText = context.sourceCode.getText();
      const startLine = context.sourceCode.getLocFromIndex(node.range[0]).line;
      const endLine = context.sourceCode.getLocFromIndex(node.range[1]).line;

      // get the text expanded to the start of the starting line and the start of the line
      // following the ending line (if there is a following line; else the end of the file)
      const expandedRange = [
        context.sourceCode.getIndexFromLoc({ column: 0, line: startLine }),
        endLine !== context.sourceCode.lines.length
          ? context.sourceCode.getIndexFromLoc({ column: 0, line: endLine + 1 })
          : fullText.length,
      ] as const;
      const expandedRangeText = fullText.slice(
        expandedRange[0],
        expandedRange[1],
      );

      if (context.sourceCode.getText(node) === expandedRangeText.trim()) {
        return fixer.removeRange(expandedRange);
      }

      return fixer.remove(node);
    }
```

MariaSolOs: That's fair! Let me try making this a bit more concise.

packages/eslint-plugin/src/rules/no-unused-vars.ts
```diff
@@ -523,6 +523,30 @@ export default createRule<Options, MessageIds>({
...
```
kirkwaiblinger, line 526: This version does weird things when a leading comment is present. Consider this test case:

```ts
    {
      code: `
/* this is an important comment */ import assert from 'assert';
      `,
      errors: [
        {
          line: 2,
          messageId: 'unusedVar',
        },
      ],
      options: [
        {
          enableAutofixRemoval: {
            imports: true,
          },
        },
      ],
      output: `
/* this is an important comment */       `,
    },
```

kirkwaiblinger: to be clear, my assertion is that the trailing newline should not have been removed from this test case, so a correct test case would be 
```ts
    {
      code: `
/* this is an important comment */ import assert from 'assert';
      `,
      errors: [
        {
          line: 2,
          messageId: 'unusedVar',
        },
      ],
      options: [
        {
          enableAutofixRemoval: {
            imports: true,
          },
        },
      ],
      output: `
/* this is an important comment */${' '}
      `,
    },
```

(or, maybe, not autofix at all, and put the removal in the sugggestion 🤷)

MariaSolOs: I see, this is a good edge case. I've updated the implementation in https://github.com/typescript-eslint/typescript-eslint/compare/f6d67b9950b69b4579c5ba8111713509628d44d5..fbaa7a0b5efb914a5290a17fcd16685a93618395, which also includes the usage of `.trim()` as suggested.

MariaSolOs:
Hmm not sure why the CI checks were cancelled...

packages/eslint-plugin/tests/rules/no-unused-vars/no-unused-vars-enableAutofixRemoval.test.ts
```diff
@@ -561,6 +556,27 @@ import { Used, } from 'module';
...
+          {
+            enableAutofixRemoval: {
+              imports: true,
+            },
+          },
+        ],
+        output: `
+/* this is an important comment */
+        `,
+      },
```
bradzacher, line 600: Could you add this test case:

```js
      {
        code: `
import assert from 'assert'; /* this is an important comment */
        `,
        errors: [
          {
            line: 2,
            messageId: 'unusedVar',
          },
        ],
        options: [
          {
            enableAutofixRemoval: {
              imports: true,
            },
          },
        ],
        output: `
 /* this is an important comment */
        `,
      },
```

just for completeness

MariaSolOs: Done!

MariaSolOs:
> This approach should work.
> 
> An alternate approach might be to use tokens:
> 
> 1. get the token (incl comments) before the import decl
>    
>    1. if it's on the same line then the decl is not on its own line
>    2. else it's on its own line and and set delete start to `prevToken.range[1]`
> 2. get the token (incl comments) after the import decl
>    
>    1. if it's on the same line then the decl is not on its own line
>    2. else it's on its own line and and set delete start to `nextToken.range[0]`
> 
> That would let us guarantee that we can delete extra whitespace (which is a nice benefit).
> 
> There are two edge cases to handle with this though:
> 
> 1. if the import decl is the first statement in the file
> 2. if the import decl is the last statement before non-decls
> 
> IDK if this is better than your current approach.. probably not due to the edge cases?
> 
> 1 extra test and I think we're probably good to go?

Yeah because of edge cases I think this is as pretty as it gets.

MariaSolOs:
> I'm happy with this as-is. The token approach sounds fine to me too so feel free to go with that but it looks like it'll just be a bit longer since you have to do the same wrangling of the edge case when there's no next line 🤷‍♂️
> 
> Thanks for sending this PR!

Thanks to you for all the feedback! Is always a pleasure to contribute to this project :)

JoshuaKGoldberg (approved):
Ok I have to admit, it is nice to see all the removed lines in the test file 😄 

Great work, thanks @MariaSolOs! Both on advocating for the change and implementing (+ re-implementing) it.

PR #11994, opened 2026-01-22
chore: skip jobs for non-affected packages

StyleShit:
Continuing on #11204

Having NX cache is nice and everything, but we're still spinning up lots of jobs of non-affected packages just to get their results from cache.
This PR could potentially save roughly up to 44min of redundant CI time (`44 skipable jobs` x `~1min`)

This PR causes all packages to be affected because I'm changing a workflow, but you can see a filtered run in my fork:

https://github.com/StyleShit/typescript-eslint/pull/1
https://github.com/StyleShit/typescript-eslint/actions/runs/21256454117

image

I'm not sure how to handle GitHub's "required jobs" feature in skipped cases though...

image


.github/workflows/ci.yml
```diff
@@ -52,6 +52,74 @@ jobs:
...
+              'rule-schema-to-typescript-types',
+              'rule-tester',
+              'scope-manager',
+              'tsconfig-utils',
+              'type-utils',
+              'typescript-eslint',
+              'typescript-estree',
+              'utils',
+              'visitor-keys',
+            ];
```
bradzacher: I wonder -- can we use `yarn` to grab this and avoid needing to keep this in sync manually in future? 

EG use `yarn workspaces list --json`

Would be nice to avoid touching this when we next add a package.

StyleShit: Technically we don't even need that, nx affected will give us anything that's affected

I kept this list because I assumed it's intentionally hard coded since it doesn't contain all packages, and each list runs the tests differently

JoshuaKGoldberg: It's not super intentionally hardcoded, we just don't have unit tests in every package. I think the best thing here would be to glob `vitest.config.*` files and grab the parent directory.

JoshuaKGoldberg (dismissed):
Sorry for the slow review :c but this looks very good to me. I think if @bradzacher and/or @JamesHenry have time to review as more Nx-capable people than me that'd be good.

My only suggestion is the non-blocking one around detecting which packages care about unit tests. We can always treat that as a followup.

⚡

.github/workflows/ci.yml
StyleShit: I think a better approach (for future-proofing, and less implementation details awareness) would be `nx show projects --affected --withTarget=test`, but not sure it's better in terms of performance

JoshuaKGoldberg: Whatever solution finds the tests automatically and doesn't take a noticeable amount of time is 👍 from me.

.github/workflows/ci.yml
```diff
@@ -54,6 +54,66 @@ jobs:
         with:
           node-version: ${{ env.PRIMARY_NODE_VERSION }}
 
+  affected:
```
JamesHenry: This is very specifically affected projects WITH A TEST TARGET, so let's have a more specific job ID and name please so that that fact isn't lost in future and this list gets reused in subtly incorrect ways

Otherwise LGTM, this whole setup would be so much simpler with Nx Cloud 😅 but I don't have the bandwidth for overhauling it right now

JamesHenry (changes requested):

.github/workflows/ci.yml
StyleShit: ah oops, wanted to do it and forgot
done

JamesHenry (approved):

PR #11995, opened 2026-01-22
fix(scope-manager): prevent misidentification of `"use strict"` directives

### LINKED ISSUE #11993: Bug: Scope manager: Identifying `"use strict"` directives is too liberal

### Relevant Package

scope-manager

### Playground Link

https://typescript-eslint.io/play/#ts=5.8.2&sourceType=script&showAST=scope&fileType=.js&code=GYVwdgxgLglg9mABMOcAUBKRBvRaDkIAzgKaJFQBOM0%2BGA3IlABYxEB0JAbgIYA2jAL4AoIA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
An ESlint test case for `no-eval` rule:


// script
function foo() { ('use strict'); this.eval; }
```

### ESLint Config

```javascript

```

### tsconfig

```jsonc

```

### Expected Result

Scope for function `foo` should have `isStrict: false`. `('use strict')` is not a directive.

### Actual Result

Scope for function `foo` has `isStrict: true`.

The above example produces an error with `no-eval` rule with ESLint's built-in scope manager (correctly) but does not with TS-ESLint's scope manager (incorrect). [ESLint playground](https://eslint.org/play/#eyJ0ZXh0IjoiZnVuY3Rpb24gZm9vKCkgeyAoJ3VzZSBzdHJpY3QnKTsgdGhpcy5ldmFsOyB9XG4iLCJvcHRpb25zIjp7InJ1bGVzIjp7Im5vLWV2YWwiOlsiZXJyb3IiXX0sImxhbmd1YWdlT3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319LCJlY21hVmVyc2lvbiI6ImxhdGVzdCIsInNvdXJjZVR5cGUiOiJzY3JpcHQifX19)

This same problem also causes 3 tests to fail for `no-invalid-this` rule.

### Additional Info

I believe the cause is that checking for `"use strict"` directives is overly liberal:

https://github.com/typescript-eslint/typescript-eslint/blob/aaa7ca2d6afd8def8cf9d01d0d9deedd38ae1495/packages/scope-manager/src/scope/ScopeBase.ts#L88-L108

It accepts `Literal` nodes without `directive` property defined.

IMO it should only check the `directive` field:

```js
for (const stmt of body.body) {
  if (stmt.type !== AST_NODE_TYPES.ExpressionStatement) {
    break;
  }

  if (stmt.directive == null) {
    break;
  }

  if (stmt.directive === 'use strict') {
    return true;
  }
}
```

This code dates back to the original introduction of `scope-manager` package (#1939). Perhaps the more liberal check was there for some historical reason e.g.

1. Copied from `eslint-scope`, but `eslint-scope` has been fixed since?
2. A workaround for compatibility with another parser which didn't correctly produce `directive` property on `ExpressionStatement`?

### Versions

| `@typescript-eslint/scope-manager`     | `8.53.1` |

overlookmotel:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11993

## Overview

Fixes #11993.

Previously, `ExpressionStatement`s without a `directive` property were incorrectly treated as directives. For example this function incorrectly had `isStrict: true`:

```js
// script
function f() {
  ("use strict");
}
```

Fix this by only treating `ExpressionStatement`s as directives if they have a `directive` property.

I may have gone overboard with the tests, but I couldn't find any existing tests for `"use strict"` directives at top level or in functions (only arrow functions), so I thought may as well add tests for these at the same time.

### DIFF
--- a/packages/scope-manager/src/scope/ScopeBase.ts
+++ b/packages/scope-manager/src/scope/ScopeBase.ts
@@ -86,22 +86,14 @@ function isStrictScope(
 
   // Search 'use strict' directive.
   for (const stmt of body.body) {
-    if (stmt.type !== AST_NODE_TYPES.ExpressionStatement) {
+    if (
+      stmt.type !== AST_NODE_TYPES.ExpressionStatement ||
+      stmt.directive == null
+    ) {
       break;
     }
 
     if (stmt.directive === 'use strict') {
-      return true;
-    }
-
-    const expr = stmt.expression;
-    if (expr.type !== AST_NODE_TYPES.Literal) {
-      break;
-    }
-    if (expr.raw === '"use strict"' || expr.raw === "'use strict'") {
-      return true;
-    }
-    if (expr.value === 'use strict') {
       return true;
     }
   }
--- a/packages/scope-manager/tests/eslint-scope/use-strict-directive.test.ts
+++ b/packages/scope-manager/tests/eslint-scope/use-strict-directive.test.ts
@@ -1,0 +1,191 @@
+import { AST_NODE_TYPES } from '@typescript-eslint/types';
+
+import { ScopeType } from '../../src/index.js';
+import { parseAndAnalyze } from '../test-utils/index.js';
+
+describe('"use strict" directives', () => {
+  describe('top level', () => {
+    it('"use strict" at top level makes the global scope strict', () => {
+      const { scopeManager } = parseAndAnalyze('"use strict";', 'script');
+
+      expect(scopeManager.scopes).toHaveLength(1);
+
+      const globalScope = scopeManager.scopes[0];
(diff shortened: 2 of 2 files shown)


overlookmotel:
Not sure why CI typecheck task is failing. `yarn typecheck` also fails locally, but with a different error! Both the local error and the CI error seem unrelated to this PR.

bradzacher (approved):
thanks!

PR #12001, opened 2026-01-28
chore(website): use `@docusaurus/faster` for faster docusaurus builds

StyleShit:
Continuing on #11204

I noticed that the longest task we have is `website:build`. I started with some hacking around and yarn resolutions to override webpack with rspack to make things faster, but then I realized that there is a [`@docusaurus/faster`](https://github.com/facebook/docusaurus/issues/10556) package 🥳 

Ran a few benchmarks locally (M4 Pro, 24GB), and they seem to be aligned with what we should have expected (3x faster):

```
hyperfine --prepare "yarn clean" "yarn docusaurus build" "FASTER=true yarn docusaurus build"
Benchmark 1: yarn docusaurus build
  Time (mean ± σ):     41.454 s ±  0.593 s    [User: 76.653 s, System: 9.456 s]
  Range (min … max):   40.775 s … 42.707 s    10 runs
 
Benchmark 2: FASTER=true yarn docusaurus build
  Time (mean ± σ):     13.396 s ±  0.541 s    [User: 21.162 s, System: 2.974 s]
  Range (min … max):   12.867 s … 14.344 s    10 runs
 
Summary
  FASTER=true yarn docusaurus build ran
    3.09 ± 0.13 times faster than yarn docusaurus build
```

I benchmarked only cold builds since [we don't share the builds](https://github.com/typescript-eslint/typescript-eslint/blob/d423e57622e1c8e789a1b84109d1ab2c15497507/.github/actions/prepare-build/action.yml) between runs anyway

Also, I added caching for the `website:build` task and its dependency `website-eslint:build`, because we're currently cold-building both of them on every PR for no apparent reason.

I'm wondering whether we should remove the exclusion of these 2 builds from the [`prepare-build`](https://github.com/typescript-eslint/typescript-eslint/blob/main/.github/actions/prepare-build/action.yml#L32) action, since we're building them anyway as dependencies of other tasks `website:typecheck` for example, wdyt?

### DIFF
--- a/packages/project-service/package.json
+++ b/packages/project-service/package.json
@@ -49,8 +49,8 @@
     "typescript": ">=4.8.4 <6.0.0"
   },
   "dependencies": {
-    "@typescript-eslint/tsconfig-utils": "^8.54.0",
-    "@typescript-eslint/types": "^8.54.0",
+    "@typescript-eslint/tsconfig-utils": "workspace:^",
+    "@typescript-eslint/types": "workspace:^",
     "debug": "^4.4.3"
   },
   "devDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -16,7 +16,7 @@ importers:
     devDependencies:
       '@eslint-community/eslint-plugin-eslint-comments':
         specifier: ^4.5.0
-        version: 4.5.0(eslint@9.39.2(jiti@2.6.1))
+        version: 4.6.0(eslint@9.39.2(jiti@2.6.1))
       '@eslint/eslintrc':
         specifier: ^3.3.3
         version: 3.3.3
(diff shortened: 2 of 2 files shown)


StyleShit:
@bradzacher wdyt about building these packages in prepare-build?
I'm assuming this action is to warmup the caches?
If so, as i said, we need these 2 packages built anyway

bradzacher:
> wdyt about building these packages in prepare-build?
> I'm assuming this action is to warmup the caches?
> If so, as i said, we need these 2 packages built anyway

We only need them for the typecheck action, right?
So by deferring their build to the typecheck action then all the actions that depend on the build warmup won't wait for them and thus will start earlier.

JoshuaKGoldberg:
Haha, thanks, I'd wanted to try to merge things in but got distracted with other stuff...

PR #12002, opened 2026-01-28
chore: remove redundant tsconfig reference

StyleShit:
Continuing on #11204 again

When working on #12001, I noticed that we have a seemingly redundant reference to `packages/typescript-eslint`

It causes the `repo:typecheck` to also check some of the packages themselves, rather than just the root files, resulting in slightly (but probably negligible) slower run and duplicated errors.

To see the issue, checkout `main`, add a type error in one of the referenced packages (i.e. `eslint-package`), and run `yarn nx run-many --target typecheck`.
You should see the same error twice

### DIFF
--- a/tsconfig.repo-config-files.json
+++ b/tsconfig.repo-config-files.json
@@ -16,9 +16,6 @@
   ],
   "references": [
     {
-      "path": "./packages/typescript-eslint/tsconfig.build.json"
-    },
-    {
       "path": "./tsconfig.spec.json"
     }
   ]


tsconfig.repo-config-files.json
```diff
@@ -15,9 +15,6 @@
     ".github/**/*.js"
   ],
   "references": [
-    {
-      "path": "./packages/typescript-eslint/tsconfig.build.json"
```
kirkwaiblinger, line 19: I think the reason for this was that in order to typecheck `eslint.config.mjs`, you need to `import tseslint from 'typescript-eslint'`... Does removing this not cause problems?

StyleShit: hmm

well, it fails only if you don't build it, but we do build it in CI so it passes

if we still want to make it work locally without build, maybe we can use [live types with custom conditions](https://colinhacks.com/essays/live-types-typescript-monorepo#:~:text=5%2E%20Custom%20conditions%20in%20%22exports) instead?

kirkwaiblinger: My tentative understanding of the nx TS setup is that the tsconfigs should fully represent the dependency graph via project references... But @JamesHenry is the expert on such matters so we'll see if he has time to chime in?

Or others in @typescript-eslint/triage-team ?

JoshuaKGoldberg: Yes I believe the current way it is is correct & necessary. Removing the reference might coincidentally work from files already being built, but that's not a guarantee I think is safe to rely on.

StyleShit: OK, sure

So let's close this one, or you wanna wait for @JamesHenry?

JoshuaKGoldberg: Let's close it out. Thank you though!

PR #12038, opened 2026-02-05
fix(type-utils): use exact package name matching instead of substring regex

### LINKED ISSUE #11946: Bug: [only-throw-error] only-throw-error.allow[].package unexpectedly works for prefixes

### Playground Link

https://eslint-online-playground.netlify.app/#eNp1Uk1v2zAM/SuELm6BxMYG7DDvEqAbsNuAobeqB0FmEqeyJEhK08Lwfy/1YTetUR8k4/Hx8VHkyLyTDb6IwSqsg2ct6wdrXIAR/oZg/zhn3H/01miPMMHemQGqndCHsxKukWYYjG6OxKx+cc11ODpzAY2XdfbNLduw4KXR+/5Qn7zRVGycCESveh3qJUJ408Au+K08onyKuoupDve9xrtEXfxkAXIT0Wik0IPPkUILrxa9dL0N24xnz/iSyKQszirdS4UbrmFRKQZ97TD2jbrD7p4k76JJ7DaRO8YDQMUHEgf8Z0NPvbczDmCF8+jWOEWcOaEMLQR3xhmekur77c4Kr9Oq3aqpxmj1uk2D2GIcQNXCw3uZKmNFL35XHgCEUuZCCWN6sxYqK+QTdVJtQIsBCVgNlkKFRNHr1ahgeryqM/cAMKMZiectTYI2oegs2xHDnMUWOWvpbzAdPQBnKY+zDm2cgpY9+kgonXD2aUFz8ref9ff6B2dzzazw/PtLkfKepwRzpkRAH0rxSMjxL4KruXzkJRNcT2x6A9ZJKNQ=

### Repro Code

```TypeScript
import { HttpErrorResponse } from '@angular/common/http';

throw new HttpErrorResponse()
```

### ESLint Config

```javascript
// @ts-check

import { defineConfig } from 'eslint/config';

export default defineConfig(
  tseslint.configs.recommendedTypeChecked,
  {
    languageOptions: {
      parserOptions: {
        project: true
      },
    },
    rules: {
      '@typescript-eslint/only-throw-error': [
        'error',
        {
          allow: [{ from: 'package', name: 'HttpErrorResponse', package: '@angular/com' }],
        },
      ],
    },
  },
);
```

### tsconfig

```jsonc
{}
```

### Expected Result

```
Expected an error object to be thrown.
[(@typescript-eslint/only-throw-error)](https://typescript-eslint.io/rules/only-throw-error)
[3:7-30]
```

because `HttpErrorResponse` was only allowed from package `@angular/com`, not from package `@angular/common/http`

### Actual Result

No error: the import seems to be whitelisted.

### Additional Info

I created this issue on @JoshuaKGoldberg’s [suggestion](https://github.com/typescript-eslint/typescript-eslint/issues/11817#issuecomment-3673125244) from https://github.com/typescript-eslint/typescript-eslint/issues/11817.

andreahlert:
## PR Checklist

- [x] Addresses an existing open issue: Fixes #11946
- [x] That issue was mass-labeled as `accepting prs`

## Overview

The previous implementation of `typeDeclaredInDeclarationFile` used regex substring matching which caused package specifiers like `@angular/com` to incorrectly match types from `@angular/common/http`.

### The Bug

The original code:
```typescript
const matcher = new RegExp(`${packageName}|${typesPackageName}`);
// ...
matcher.test(packageIdName)
```

For a package specifier `@angular/com`, this would create regex `/@angular\/com|angular__com/` which would match `@angular/common` because it's a substring match.

### The Fix

This change:
1. Extracts the package name from module specifiers that may include subpaths (e.g., `@types/babel__code-frame/index.d.ts` → `@types/babel__code-frame`)
2. Uses exact string comparison instead of regex
3. Properly handles `@types` packages for scoped packages (e.g., `@babel/code-frame` → `@types/babel__code-frame`)

The fix ensures that `{ from: 'package', name: 'Foo', package: '@angular/com' }` will NOT match types from `@angular/common/http`, while still correctly matching types from packages that use the @types convention.


JoshuaKGoldberg (changes requested):
The logic looks reasonable, but we generally need tests for core/important areas like this. Could you please add some?

Apologies for the delay in reviewing -> asking for something straightforward. 😅

andreahlert:
@JoshuaKGoldberg added tests for the exact matching behavior. three cases covering the substring bug: partial non-scoped names (`typescrip`, `semve`) and partial scoped names (`@babel/code`), all asserting they don't match.

also removed an unreachable branch in `extractPackageName` that was causing the codecov patch check to fail. scoped packages always have at least `@scope/name`, so the length guard was dead code.

packages/type-utils/src/typeOrValueSpecifiers/typeDeclaredInPackageDeclarationFile.ts
```diff
@@ -28,20 +28,42 @@ function typeDeclaredInDeclareModule(
...
+      return false;
+    }
+    // Extract just the package name from the module specifier
+    const extractedPackageName = extractPackageName(packageIdName);
     return (
-      packageIdName != null &&
-      matcher.test(packageIdName) &&
+      (extractedPackageName === packageName ||
+        extractedPackageName === typesPackageName ||
+        extractedPackageName === atTypesPackageName) &&
```
JoshuaKGoldberg, line 66: [Style] Nit: the variable is only used once, and it's pretty clear what it does.

```suggestion
        extractedPackageName === `@types/${typesPackageName}`) &&
```

packages/type-utils/tests/TypeOrValueSpecifier.test.ts
```diff
@@ -576,6 +576,23 @@ describe('TypeOrValueSpecifier', () => {
         'import type {Node as TsNode} from "typescript"; type Test = TsNode;',
         { from: 'package', name: 'TsNode', package: 'typescript' },
       ],
+      // Substring matching bug: partial package names should not match
```
JoshuaKGoldberg, line 579: [Docs] Nit: this is pretty clear from the code. We don't generally add comments like this unless they're really necessary (and then, there's actually already a `name` property).

```suggestion
```

packages/type-utils/src/typeOrValueSpecifiers/typeDeclaredInPackageDeclarationFile.ts
```diff
@@ -28,20 +28,42 @@ function typeDeclaredInDeclareModule(
...
+  // Also handle @types packages for scoped packages
+  const atTypesPackageName = `@types/${typesPackageName}`;
 
-  const matcher = new RegExp(`${packageName}|${typesPackageName}`);
   return declarationFiles.some(declaration => {
     const packageIdName = program.sourceFileToPackageName.get(declaration.path);
+    if (packageIdName == null) {
+      return false;
+    }
+    // Extract just the package name from the module specifier
```
JoshuaKGoldberg, line 61: [Docs] Nit: this is just a restatement of the same line of code:

```suggestion
```

packages/type-utils/tests/TypeOrValueSpecifier.test.ts
JoshuaKGoldberg, line 1: [Tests] A good start, but we've got a few more to request adding 🙂. https://github.com/typescript-eslint/typescript-eslint/issues/11946#issuecomment-3774624175 mentions some good cases, both matching and non-matching. Could you add those in too, please?

JoshuaKGoldberg (changes requested):
Looking good! We like to be thorough - could you add the suggested tests from the issue? No other blockers from me, assuming the code still works with them added. 🚀

PR #12042, opened 2026-02-08
chore: reduce build step cache size

### LINKED ISSUE #12035: Repo: "Post Build" step is taking forever
### Suggestion

https://github.com/typescript-eslint/typescript-eslint/actions/runs/21673047655/job/62485738215?pr=12033#step:6:1

Image

not sure whether it's related to the pnpm migration or something else, but this is really bad

### Additional Info

_No response_

StyleShit:
Closes #12035

The cache step included files also from node_modules, resulting in a huge amount of time wasted collecting and compressing unnecessary files

This PR also reduces the build cache size from 82MB down to 780KB

2 consecutive runs in this PR resulted in basically zero time wasted:

post build time after fix

https://github.com/typescript-eslint/typescript-eslint/actions/runs/21795297762/job/62881742409?pr=12042

post build time after fix

https://github.com/typescript-eslint/typescript-eslint/actions/runs/21794563757/job/62879810596?pr=12042

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -187,7 +187,7 @@ importers:
         version: 7.28.5
       '@microsoft/api-extractor':
         specifier: ^7.55.2
-        version: 7.55.2(@types/node@22.19.8)
+        version: 7.56.0(@types/node@22.19.8)
       '@types/babel__code-frame':
         specifier: ^7.0.6
         version: 7.27.0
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (approved):
⚡ WOW. ⚡⚡⚡ WOW WOW WOW. What a find!! 👏 

I was concerned at first that this would wholly remove `node_modules/` caching altogether. Just recapping the details for future referenced:

1. `actions/setup-node` with `cache: 'pnpm'` intentionally doesn't cache `node_modules/`; it instead caches the pnpm store path (`pnpm store path`)
2. When the store path is populated, pnpm installs end up mostly creating symlinks in `node_modules/` to that

...so this should be a work reduction, not a work move-to-different-step-in-the-process.

Just to be safe, I deleted the recent caches under https://github.com/typescript-eslint/typescript-eslint/actions/caches -especially any that mentioned 12042- and updated from main. Assuming that build passes I think we'll be good to merge!

PR #12046, opened 2026-02-08
feat(type-utils): detect nested any types

minyeong981:
## PR Checklist

- [x] Addresses an existing open issue: #11801

## Overview
This change extends the existing any-type detection logic to cover nested any occurrences inside complex type structures.

Previously, the analysis only detected any at the top level, which caused cases such as objects, unions, intersections, and function signatures containing nested any to be missed.
This update introduces a recursive traversal of TypeScript types to ensure that any is detected even when deeply nested.

Additionally, types originating from the default TypeScript library (lib*.d.ts) are explicitly excluded to avoid false positives from built-in definitions.

Test cases have been updated accordingly to validate the new detection behavior across these scenarios.
💖


packages/type-utils/src/isUnsafeAssignment.ts
```diff
@@ -1,11 +1,32 @@
...
 import * as tsutils from 'ts-api-utils';
+import * as ts from 'typescript';
 
 import { isTypeAnyType, isTypeUnknownType } from './predicates';
 
+/**
+ * Checks if a type is from the default library (lib.*.d.ts files).
+ * We skip these to avoid false positives from built-in types like Object, Array, etc.
+ */
+function isFromDefaultLibrary(type: ts.Type): boolean {
```
JoshuaKGoldberg: [Bug] This will get tricked by types in files like `./src/gotcha.lib.d.ts`. We also already have logic for this in `type-utils` -> `isSymbolFromDefaultLibrary`. See that & other places that call `program.isSourceFileDefaultLibrary`.

packages/type-utils/src/isUnsafeAssignment.ts
```diff
@@ -115,5 +136,135 @@ function isUnsafeAssignmentWorker(
     return false;
   }
 
+  // Check object types - compare properties, index signatures, and call signatures
+  if (tsutils.isObjectType(type) && tsutils.isObjectType(receiver)) {
+    // Only check properties for non-default-library types to avoid
+    // false positives from built-in types like Object, Array, etc.
+    if (!isFromDefaultLibrary(type)) {
```
JoshuaKGoldberg: > Additionally, types originating from the default TypeScript library (lib*.d.ts) are explicitly excluded to avoid false positives from built-in definitions.

[Question] Why? If the built-in type definitions are unsafe, we would want to catch them, no?

[Testing] I tried out this PR locally with the `isFromDefaultLibrary` check removed. It passed all test cases. So this can probably be directly removed from the PR, right? Either way I think there are some missing test cases around "Object, Array, etc." that should be added.

packages/eslint-plugin/tests/rules/no-unsafe-assignment.test.ts
JoshuaKGoldberg, line 1: [Feature] The rule logic works nicely, but now presents a message ambiguity: it's no longer always clear where the `any` comes from! Consider this test case:

```ts
type Foo = { bar: { a: number } };
type Inner = { a: any };
declare const bar: Inner;
const foo: Foo = { bar };
```

This implementation reports:

```plaintext
Unsafe assignment of type `{ bar: Inner; }` to a variable of type `Foo`.
```

This'll get even more complex for places that are more nested. We'll have to think a bit about how to surface this to users.

Discussion / feature request: what do you think about about specifying the actual path(s) to the unsafe `any` in the types? e.g.

```plaintext
Unsafe assignment of type `{ bar: Inner; }` to a variable of type `Foo`.
  `bar.a` is the unsafe `any` type.
```

JoshuaKGoldberg (changes requested):
Sorry for taking >1 month to get to this! For full transparency, we've been pretty swamped lately. But this is a good feature area and I'm excited to keep it going. Please do keep at it, and let us know if you have any questions! 🔥

packages/type-utils/src/isUnsafeAssignment.ts
minyeong981: I checked the second comment additionally. I modified it to replace it with `isSymbolFromDefaultLibrary`, but as you pointed out, this check itself seems unnecessary. I will remove this guard and add a built-in test case as well.

packages/type-utils/src/isUnsafeAssignment.ts
minyeong981: That's a good point. I excluded built-in like Object and EventTarget because they had so many properties that they thought it could cause unexpected false positives in the recursive comparison. However, if all tests actually pass, this check is not required, and if the built-in is safe, it is correct to detect it. I will remove it. I will also supplement the test cases related to Object and Array.

packages/eslint-plugin/tests/rules/no-unsafe-assignment.test.ts
minyeong981: I thought about a couple of possible approaches:

1. Include the property path in the message (e.g. `bar.a`), as you suggested.

2. Report directly on the problematic property node (e.g. `bar` in object literals):

```ts
const foo: Foo = { bar };
//                 ^^^
// Unsafe assignment: `Inner` contains `any` at property `.a`
```

We could also combine both approaches, but that would make the reporting behavior inconsistent between object literals and simple assignments.

While (2) is more intuitive, (1) seems more consistent with existing rules.

Which direction would you prefer?

JoshuaKGoldberg: 1 is necessary because we might not have the node available to report on.

```ts
import { Apple, banana } from 'elsewhere';
const apple: Apple = banana;
```

minyeong981: Got it, that makes sense.

I’ll update this PR to include the property path based on (1). Let me know if you’d prefer this as a follow-up change instead.

minyeong981:
@JoshuaKGoldberg

Hi! I started working on this and pushed an initial implementation.

I’m seeing CI fail with the following NX error after `types:build` completes successfully:

`Cannot read properties of undefined (reading 'cacheError')`

Because of that, I’m not sure whether the issue is in my changes or if this is related to CI / NX itself. Could you take a quick look and let me know whether I’m on the right track?

JoshuaKGoldberg:
Yeah that's definitely an Nx issue. You don't have to worry about it.

packages/eslint-plugin/src/util/class-scope-analyzer/classScopeAnalyzer.ts
```diff
@@ -687,8 +687,8 @@ class ClassScope extends ThisScope implements ClassScopeResult {
    * The class's members, keyed by their name
    */
   public readonly members: ClassScopeResult['members'] = {
-    instance: new Map(),
-    static: new Map(),
+    instance: new Map<Key, Member>(),
+    static: new Map<Key, Member>(),
```
StyleShit, line 691: [Question]

Why do we need this change? The `Map` type is already being inferred by `ClassScopeResult['members']`, no?

packages/type-utils/src/isUnsafeAssignment.ts
```diff
@@ -108,7 +108,146 @@ function isUnsafeAssignmentWorker(
...
+        // here is non-trivial. For now we propagate any nested path as-is.
+        return { path: unsafe.path, receiver, sender: type };
+      }
+    }
+
+    return false;
+  }
+
+  // Check object types - compare properties, index signatures, and call signatures
+  if (tsutils.isObjectType(type) && tsutils.isObjectType(receiver)) {
```
StyleShit, line 122: [Missing Tests]

Removing this doesn't fail any test:

```suggestion
  if (tsutils.isObjectType(type)) {
```

StyleShit, line 122: [Simplification, Readability, Non-Blocking]

This condition body is pretty big. It can probably be separated into ~3 smaller function calls (for props, indexes, and function signatures) rather than having all the logic in the same place. IMO it'll be easier to read & follow

packages/type-utils/src/isUnsafeAssignment.ts
```diff
@@ -108,7 +108,146 @@ function isUnsafeAssignmentWorker(
...
+          return { path: [], receiver, sender: type };
+        }
+      }
+    }
+
+    // Check call signatures (function return types)
+    const senderSignatures = type.getCallSignatures();
+    const receiverSignatures = receiver.getCallSignatures();
+
+    if (senderSignatures.length > 0 && receiverSignatures.length > 0) {
```
StyleShit, line 177: [Missing Tests]

Removing this condition doesn't fail any test

```suggestion
```

packages/type-utils/src/isUnsafeAssignment.ts
```diff
@@ -108,7 +108,146 @@ function isUnsafeAssignmentWorker(
...
+      if (typeIndexInfo && receiverIndexInfo) {
+        const unsafe = isUnsafeAssignmentWorker(
+          typeIndexInfo.type,
+          receiverIndexInfo.type,
+          checker,
+          senderNode,
+          visited,
+        );
+        if (unsafe) {
+          // TODO: Index signature paths could be represented as `[string]` or
```
StyleShit, line 166: [Docs]

I don't think this is a TODO?
If this is something we can't really know/fix, let's just keep the explanation without prefixing with "TODO:"

Same for similar TODOs in this file

packages/type-utils/src/isUnsafeAssignment.ts
```diff
@@ -108,7 +108,146 @@ function isUnsafeAssignmentWorker(
...
+        receiverSignatures.length,
+      );
+
+      for (let i = 0; i < minSigs; i += 1) {
+        const senderReturnType = checker.getReturnTypeOfSignature(
+          senderSignatures[i],
+        );
+        const receiverReturnType = checker.getReturnTypeOfSignature(
+          receiverSignatures[i],
+        );
```
StyleShit, line 189: [Bug]

Every signature of `sender` should be checked against every signature of `receiver`:

```TS
declare function senderFn1(): any;
declare function senderFn1(): string;

declare function senderFn2(): string;
declare function senderFn2(): any;

const test1: { fn: () => string } = { fn: senderFn1 }; // Error, great!
const test2: { fn: () => string } = { fn: senderFn2  }; // Whoops
```

[Playgorund](https://deploy-preview-12046--typescript-eslint.netlify.app/play/#ts=6.0.2&fileType=.ts&code=CYUwxgNghgTiAEAzArgOzAFwJYHtXwGcRVQYAxVARgAoBKALnilQE8BuAKFElgRXWx5CxUhRoNCGGFlQBzTl3DQ4SNJlz4iJEOVQAmOowJSZ8xTxX91QraP2GmrBWDzH4GEMcqMA3klSMdPAAvAB8ktJy8AC%2BIfB%2BiAHC2rqUMWzwAPSZ8ACiMDA4MAA08LJwUBgAhBwuqG4exnq%2B-oG0IeHGkbIxcQlJtjoUevDpWTkA6gAWODgADgRAA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tieQEMAZolp9kySgHMmAW0Qt0URNGgdokcGAC%2BITUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

packages/type-utils/src/isUnsafeAssignment.ts
```diff
@@ -108,7 +108,146 @@ function isUnsafeAssignmentWorker(
...
+          // TODO: A call signature's return type path could be represented as
+          // `(return type)`, but there is no standard notation for this.
+          return { path: [], receiver, sender: type };
+        }
+      }
+    }
+  }
+
+  // Check union types - compare constituent types pairwise
+  if (type.isUnion() && receiver.isUnion()) {
```
StyleShit, line 208: [Missing Tests]

Removing either of `.isUnion()` here doesn't fail any test:

```suggestion
  if (type.isUnion()) {
```

packages/type-utils/src/isUnsafeAssignment.ts
```diff
@@ -108,7 +108,146 @@ function isUnsafeAssignmentWorker(
...
+    const senderMembers = type.types;
+    const receiverMembers = receiver.types;
+
+    if (senderMembers.length === receiverMembers.length) {
+      for (let i = 0; i < senderMembers.length; i += 1) {
+        const unsafe = isUnsafeAssignmentWorker(
+          senderMembers[i],
+          receiverMembers[i],
```
StyleShit, line 216: [Question]

Why does the member count need to be the same?

[Bug]

Same as with the signatures, each of the members of `receiver` should be compared with every member of `sender`:

```TS
declare const sender: { a: any } | { b: string };

const test: { b: string } | { a: number } = sender;
```

[Playground](https://deploy-preview-12046--typescript-eslint.netlify.app/play/#ts=6.0.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3mkLQwC54BveKIqFAT3gF94AfE%2BAIyMxgEsUBzOgbgBQg5OiwYQmIqTbYMXXnUbNy8FAFcAtixAwlAXmy58-IA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tieQEMAZolp9kySgHMmAW0Qt0URNGgdokcGAC%2BITUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

EDIT: The above also applies to intersections ([Playground](https://deploy-preview-12046--typescript-eslint.netlify.app/play/#ts=6.0.2&fileType=.ts&code=CYUwxgNghgTiAEYD2A7AzgF3mkLQwC54BveKIqFAT3gF94AyE%2BAIyMxgEsUBzOgbgBQg5OiwYQmIqTbYMXXnUbNy8FAFcAtixAwlAXmy58-IA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tieQEMAZolp9kySgHMmAW0Qt0URNGgdokcGAC%2BITUA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false))

StyleShit (changes requested):
Thanks for working on this!

This is pretty clean. Added comments mainly on bugs/testing/coverage
I didn't review the tests themselves since they'll probably change after this review

Also, I'm not seeing anything related to `no-unsafe-argument`, or actually any `no-unsafe-*` other than `no-unsafe-assignment`. Is that intentional? We'll probably need a follow-up then

bradzacher:
This PR has not had a response from the author in 5 months -- marked as stale as per our PR maintenance policies.

minyeong981:
Thanks for the reminder! There are quite a few changes to address from the review, so I haven't been able to get back to this as quickly as I'd hoped.

I'm still planning to continue working on this PR and address the feedback. Thanks for your patience!

StyleShit:
Hey!

I'm closing this PR for now since it has been open for a long time without activity.
Feel free to reopen when you have time to work on this. Thanks! 💖

PR #12047, opened 2026-02-09
chore: use pnpm catalog

### LINKED ISSUE #11266: Repo: "*" should not be used which maybe broken with `yarn-berry-deduplicate`
### Suggestion

https://github.com/search?q=repo%3Atypescript-eslint%2Ftypescript-eslint+%2F%3A+%5C%22%5C*%5C%22%2F+path%3A**%2Fpackage.json&type=code

related #11216

https://github.com/typescript-eslint/typescript-eslint/blob/e4933170f75b3a8a0c9bf3985fb4d2ddb6e4b4c6/packages/website/package.json#L71

`"make-dir": "*"` currently installs `5.0.0`, but there is also a `make-dir@npm:^4.0.0` installed by `istanbul-lib-report` -> `istanbul-reports` -> `@vitest/coverage-v8`, so with `yarn-berry-deduplicate`, `"make-dir": "*"` will be narrowed into `make-dir@npm:^4.0.0` which is unexpected for `website` package actually.

### Additional Info

_No response_

StyleShit:
Closes #11266 

Should we also read things like this from the catalog?
https://github.com/typescript-eslint/typescript-eslint/blob/f645b7f70e11acd953285057b515b20843f69924/packages/type-utils/package.json#L55

it [gets replaced](https://pnpm.io/catalogs#publishing) by pnpm upon publish, but it means we're coupled to pnpm for publishing, not sure this is the best idea.

in addition, should we add a CI check (i.e. [synpack](https://www.npmjs.com/package/syncpack)) that ensures nobody uses `*` anymore? or that nobody uses a different version than what's specified in the catalog?

### DIFF
--- a/packages/ast-spec/src/expression/ObjectExpression/fixtures/_error_/property-modifier-abstract/snapshots/1-TSESTree-Error.shot
+++ b/packages/ast-spec/src/expression/ObjectExpression/fixtures/_error_/property-modifier-abstract/snapshots/1-TSESTree-Error.shot
@@ -3,5 +3,5 @@
 exports[`AST Fixtures > expression > ObjectExpression > _error_ > property-modifier-abstract > TSESTree - Error`]
 TSError
 > 1 | ({abstract property: 1})
-    |   ^^^^^^^^ 'abstract' modifier can only appear on a class, method, or property declaration.
+    |   ^^^^^^^^ 'abstract' modifier cannot be used here.
   2 |


packages/integration-tests/tools/pack-packages.ts
```diff
@@ -222,3 +226,18 @@ export const teardown = async (): Promise<void> => {
...
+async function getPnpmCatalog() {
+  const pnpmWorkspace = await fs.readFile(
+    path.join(ROOT_DIR, 'pnpm-workspace.yaml'),
+    { encoding: 'utf-8' },
+  );
+
+  const parsed: PnpmWorkspace = yaml.parse(pnpmWorkspace);
+
+  return parsed.catalog;
+}
```
StyleShit, line 251: might be slightly overkill to parse the pnpm-workspace.yaml (which also introduces a new dev dependency), but I didn't find a better/simpler way

The problem is that `pnpm install` fails because the generated `package.json` contains `"package-name": "catalog:"` (pulled from the root `package.json`), but there is no catalog in the installation directory

open for suggestions

StyleShit:
`integration-tests:test` is failing for me also locally on `main` with the same error, not sure it's related to this PR, unless i'm missing something here?

also, `tsVersionRangeSync.test.ts` is failing. not sure how I should handle that (parse the yaml and compare to the catalog as I did in the integrations tests? read  this value from the catalog in build time? something else?)

pnpm-workspace.yaml
```diff
@@ -1,8 +1,18 @@
...
   - packages/*
 
-overrides:
-  "@types/node": ^22.0.0
-  "@types/react": ^18.2.14
-  tsx: ^4.7.2
-  typescript: 5.9.2
+catalog:
+  "@vitest/coverage-v8": "^3.2.4"
+  "@eslint/js": "^9.39.2"
```
bradzacher:
```suggestion
  "@eslint/js": "^10.0.0"
```

pnpm-workspace.yaml
```diff
@@ -1,8 +1,18 @@
...
+  "@types/natural-compare": "^1.4.3"
+  "@types/node": "^22.19.3"
+  "@types/react": ^18.3.21
+  eslint: "^9.39.2"
```
bradzacher:
```suggestion
  eslint: "^10.0.0"
```

packages/ast-spec/package.json
```diff
@@ -52,14 +52,14 @@
     "@types/babel__code-frame": "^7.0.6",
     "@types/babel__core": "^7.20.5",
     "@typescript-eslint/typescript-estree": "workspace:*",
-    "@vitest/coverage-v8": "^3.2.4",
+    "@vitest/coverage-v8": "catalog:",
     "@vitest/pretty-format": "^3.2.4",
     "@vitest/utils": "^3.2.4",
```
bradzacher: do we wanna just be consistent and do all vitest packages?

packages/eslint-plugin-internal/package.json
```diff
@@ -31,10 +31,10 @@
     "prettier": "3.8.0"
```
bradzacher: we've got prettier listed in multiple package.jsons

packages/parser/package.json
```diff
@@ -58,12 +58,12 @@
     "debug": "^4.4.3"
```
bradzacher, line 58: question: can we do non-dev deps?
I don't remember how our publishing setup works as to whether it would work...

packages/ast-spec/package.json
bradzacher: We have `@vitest/pretty-format` listed in multiple spots at least

packages/type-utils/package.json
```diff
@@ -57,12 +57,12 @@
   "devDependencies": {
     "@types/babel__code-frame": "^7.0.6",
     "@typescript-eslint/parser": "workspace:*",
-    "@vitest/coverage-v8": "^3.2.4",
+    "@vitest/coverage-v8": "catalog:",
     "ajv": "^6.12.6",
```
bradzacher: we have this in a few package.jsons

packages/website/package.json
```diff
@@ -57,15 +57,15 @@
...
     "typedoc-plugin-markdown": "^4.9.0",
-    "typescript": "*",
+    "typescript": "catalog:",
     "zod": "^3.25.76"
   },
   "devDependencies": {
     "@docusaurus/mdx-loader": "^3.9.2",
     "@docusaurus/types": "^3.9.2",
     "@types/history": "^4.7.11",
     "@types/mdast": "^4.0.4",
```
bradzacher: we have this in multiple package.jsons

packages/website/package.json
bradzacher: and these

bradzacher (changes requested):
much nicer!

pnpm-workspace.yaml
StyleShit: 🥳 

Done

pnpm-workspace.yaml
StyleShit: Done

packages/ast-spec/package.json
StyleShit: Good idea, done

packages/parser/package.json
StyleShit: ahh, great question...
@JoshuaKGoldberg will probably know

I'm assuming we can do it since we do have `workspace:` prod deps already:

https://github.com/typescript-eslint/typescript-eslint/blob/68a074fb61a10c3627f09678859fb4d42a3b88f5/packages/eslint-plugin/package.json#L60-L65

packages/eslint-plugin-internal/package.json
StyleShit: Moved all shared dev dependencies to the catalog (easily, thanks to AI)

skipped prod deps for now, waiting for response in https://github.com/typescript-eslint/typescript-eslint/pull/12047#discussion_r2807362905

StyleShit:
> Sweet! 🙏

uhh... CI is still failing and there are some open questions 😅 

https://github.com/typescript-eslint/typescript-eslint/pull/12047#discussion_r2807362905
https://github.com/typescript-eslint/typescript-eslint/pull/12047#issuecomment-3869909878
https://github.com/typescript-eslint/typescript-eslint/pull/12047#discussion_r2781075783
https://github.com/typescript-eslint/typescript-eslint/pull/12047#issue-3914574488

JoshuaKGoldberg:
Ah.

packages/integration-tests/tools/pack-packages.ts
bradzacher: There's packages like `@pnpm/catalogs.resolver` and `@pnpm/catalogs.config` -- can we use those packages?

https://github.com/pnpm/pnpm/tree/main/catalogs

bradzacher:
> Should we also read things like this from the catalog?
> ```
> "typescript": ">=4.8.4 <6.0.0" 
> ```
> it [gets replaced](https://pnpm.io/catalogs#publishing) by pnpm upon publish, but it means we're coupled to pnpm for publishing, not sure this is the best idea.

Let's leave it in for now. I too don't love the idea of tying ourselfs to `pnpm publish` for now.

packages/parser/package.json
bradzacher: lets avoid the non-dev deps for this first pass so we don't tie ourselves to `pnpm publish` out of the gate. https://github.com/typescript-eslint/typescript-eslint/pull/12047#issuecomment-3949505704

We can always revisit this later.

packages/integration-tests/tools/pack-packages.ts
StyleShit: Cool, didn't know it exists

Anyway, seems like you need to pass an already parsed YAML into it, so I'm not sure it's helping us here 😞

bradzacher: yeah I guess not.
well if we've got a static setup and we know exactly what's coming then we can easily make this work.

How's about you make 1 final change cos we're in the realm of untyped -- just add `asserts` to assert that the dependencies we access exist in the yaml file.
That way there'll be a clear error message if we remove one in future.

packages/eslint-plugin-internal/package.json
bradzacher: given this is an internal-only package -- I'm fine if you want to make this particular package use `catalog:` for the prod deps.

But yeah as per [this comment](https://github.com/typescript-eslint/typescript-eslint/pull/12047#discussion_r2844810154) we'll leave all the published packages as is.

packages/integration-tests/tools/pack-packages.ts
StyleShit: Great idea, done!

StyleShit:
>  I too don't love the idea of tying ourselfs to `pnpm publish` for now.

ok, sounds good to me
I still have [this test](https://github.com/typescript-eslint/typescript-eslint/actions/runs/22033040369/job/63661369851?pr=12047) failing, not sure how to handle that

See https://github.com/typescript-eslint/typescript-eslint/pull/12047#issuecomment-3869909878

Also, wdyt about https://github.com/typescript-eslint/typescript-eslint/pull/12047#issue-3914574488?
> in addition, should we add a CI check (i.e. [synpack](https://www.npmjs.com/package/syncpack)) that ensures nobody uses * anymore? or that nobody uses a different version than what's specified in the catalog?

I'm assuming we could do it as a follow-up if we want to go this route

packages/eslint-plugin-internal/package.json
StyleShit: Done

bradzacher:
https://github.com/typescript-eslint/typescript-eslint/blob/a09921e2de2e8790e6a803016b825815ca9409d8/packages/typescript-estree/tests/lib/tsVersionRangeSync.test.ts

This test will need to be updated to read from the catalog instead

bradzacher (dismissed):
amazing -- thanks so much for this!

bradzacher:
Once we have CI passing we can merge this in. All the changes are looking good.

PR #12066, opened 2026-02-15
fix(typescript-estree): skip createIsolatedProgram fallback for projectService

### LINKED ISSUE #12065: Bug: createIsolatedProgram fallback incorrectly triggered with projectService when external parser calls parseAndGenerateServices multiple times

### Issue Description

When using `projectService: true` with `vue-eslint-parser` v9 in a CI environment (where `singleRun` is inferred as `true`), type-aware lint rules incorrectly report that `strictNullChecks` is not enabled, even though `strict: true` is set in `tsconfig.json`.

**Root cause:** `vue-eslint-parser` calls `@typescript-eslint/parser`'s `parseAndGenerateServices()` multiple times for the same file path — once for the `<script>` block and additional times for template expressions (e.g., `{{ x }}`, `v-if`, `v-slot`). This has been the case since well before v9 and is a fundamental part of how `vue-eslint-parser` handles template expression linting. In `singleRun` mode, the second call increments `parseAndGenerateServicesCalls` above 1, which triggers the `createIsolatedProgram` fallback ([`parser.ts` L234-L239](https://github.com/typescript-eslint/typescript-eslint/blob/68a074fb6b4a91f04c4bbbd42239c22aec498e78/packages/typescript-estree/src/parser.ts#L234-L239)).

The current fallback logic assumes that "same file parsed more than once in singleRun = ESLint `--fix` cycle". However, it may also be worth considering the case where external parsers like `vue-eslint-parser` legitimately call the parser multiple times for the same file during a single lint pass (for template expressions, etc.).

`createIsolatedProgram` does not read `tsconfig.json`, so all compiler options (including `strictNullChecks`) fall back to TypeScript defaults (`false`), breaking type-aware rules.

**Regression:** This behavior has been latent for a long time, but was never triggered because `inferSingleRun` unconditionally returned `false` when `extraFileExtensions` was set — effectively preventing `singleRun=true` for all `.vue` users. PR #11327 (v8.48.0, November 2025) changed this guard to apply only to the `project` option, allowing `projectService` users to enter `singleRun=true` mode and exposing the pre-existing incompatibility.

https://github.com/ryo-chin/typescript-eslint-issue-repro

### Repro Steps

1. Clone the repo
2. `npm install`
3. `CI=true npx eslint --no-cache src/App.vue`

**Expected:** Rule reports actual code issues (e.g., "Unexpected nullish value in conditional").

**Actual:** Rule reports `This rule requires the 'strictNullChecks' compiler option to be turned on to function correctly`.

**Verification:** Running with `TSESTREE_SINGLE_RUN=false npx eslint --no-cache src/App.vue` produces the correct result.

### Suggested Fix

Skip the `createIsolatedProgram` fallback when `projectService` is in use:

```diff
  const { ast, program } =
    parseSettings.singleRun &&
    tsestreeOptions.filePath &&
-   parseAndGenerateServicesCalls[tsestreeOptions.filePath] > 1
+   parseAndGenerateServicesCalls[tsestreeOptions.filePath] > 1 &&
+   !parseSettings.projectService
      ? createIsolatedProgram(parseSettings)
      : getProgramAndAST(parseSettings, hasFullTypeInformation);
```

This fallback exists for the legacy `project` option, which uses AOT-compiled programs that can become stale during `--fix` cycles. `projectService` does not use AOT programs — it passes the latest source code to `openClientFile()` on every call ([`useProgramFromProjectService.ts` L150-L156](https://github.com/typescript-eslint/typescript-eslint/blob/68a074fb6b4a91f04c4bbbd42239c22aec498e78/packages/typescript-estree/src/useProgramFromProjectService.ts#L150-L156)), so no fallback is necessary. Verified that `projectService` + `singleRun=true` + `--fix` works correctly without the fallback.

### Scope and Impact

This reproduction uses `vue-eslint-parser` v9, but the multiple-call behavior for template expressions has been present in `vue-eslint-parser` since well before v9. It was never an issue because the `extraFileExtensions` guard in `inferSingleRun` kept `.vue` users in `singleRun=false` mode, preventing the fallback from being triggered.

`vue-eslint-parser` v10 (released March 2025) optimized its internals to call the parser only once per file, so projects using v10 are not affected. Since `typescript-eslint` v8.48.0 (which introduced this regression) was released in November 2025, most actively maintained Vue projects had likely already migrated to `eslint-plugin-vue` v10 / `vue-eslint-parser` v10 by that time. This may explain why no issues have been filed so far — the affected combination (`vue-eslint-parser` v9 + `projectService` + CI) is relatively uncommon at this point.

That said, the underlying issue is not specific to `vue-eslint-parser`. The `createIsolatedProgram` fallback can be triggered by any external parser or plugin that calls `parseAndGenerateServices()` multiple times for the same file path in a single lint pass. A similar issue was previously reported with `eslint-plugin-import` (#5880 / #5136), where out-of-band parsing incremented the call counter and triggered the fallback. That case was fixed in PR #5893 by preventing singleRun mode when not in type-aware linting. However, today's case differs: the multiple calls happen *during* type-aware linting (template expression parsing), so that guard does not apply. Fixing this in `typescript-estree` for the `projectService` case would make the fallback logic more robust against this class of interaction.

### Workarounds

- Set `disallowAutomaticSingleRunInference: true` in `parserOptions`
- Set environment variable `TSESTREE_SINGLE_RUN=false` in CI
- Upgrade `vue-eslint-parser` to v10 (which calls the parser only once per file)

### Related Issues

- #11327 — PR that exposed this bug by allowing `projectService` users into `singleRun=true` mode
- #5880 / #5136 — Same class of bug: `eslint-plugin-import`'s out-of-band parsing incremented the call counter, triggering `createIsolatedProgram`. Fixed in PR #5893 by preventing singleRun mode when not in type-aware linting. Today's case differs: the multiple calls happen *during* type-aware linting, so that guard does not help.
- #9344 — Same class of bug: `createIsolatedProgram` triggered during `--fix` cycles
- #5856 — Refactor issue noting `createIsolatedProgram` fallback "should never be hit"

### Versions

| package                                | version   |
| -------------------------------------- | --------- |
| `@typescript-eslint/eslint-plugin`     | `8.55.0`  |
| `@typescript-eslint/parser`            | `8.55.0`  |
| `@typescript-eslint/typescript-estree` | `8.55.0`  |
| `TypeScript`                           | `5.9.3`   |
| `ESLint`                               | `10.0.0`  |
| `node`                                 | `24.7.0`  |

ryo-chin:
## PR Checklist

- [x] Addresses an existing open issue: fixes #12065
- [x] That issue was marked as [`accepting prs`](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)

## Overview

When using `projectService` with external parsers like `vue-eslint-parser` (v9) in `singleRun` mode (e.g., CI with `CI=true`), the `createIsolatedProgram` fallback is incorrectly triggered, causing `tsconfig.json` compiler options (such as `strictNullChecks`) to be lost.

### Root Cause

`vue-eslint-parser` calls `parseAndGenerateServices()` multiple times for the same file path during a single lint pass — once for the `<script>` block and additional times for template expressions. In `singleRun` mode, the second call increments `parseAndGenerateServicesCalls` above 1, triggering the `createIsolatedProgram` fallback. This fallback does not read `tsconfig.json`, so all compiler options fall back to TypeScript defaults.

### Fix

Skip the `createIsolatedProgram` fallback when `projectService` is in use. This is safe because:

- The fallback exists for the legacy `project` option, which uses AOT-compiled programs that can become stale during `--fix` cycles.
- `projectService` does not use AOT programs — it passes the latest source code to `openClientFile()` on every call, so no fallback is necessary.
- Verified that `projectService` + `singleRun=true` + `--fix` works correctly without the fallback.

### Reproduction

https://github.com/ryo-chin/typescript-eslint-issue-repro

```bash
npm install
CI=true npx eslint --no-cache src/App.vue
# Actual: "This rule requires the `strictNullChecks` compiler option to be turned on"
# Expected: "Unexpected nullish value in conditional"
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/typescript-estree/src/parser.ts
+++ b/packages/typescript-estree/src/parser.ts
@@ -225,6 +225,12 @@ export function parseAndGenerateServices<
    *
    * In this scenario we cannot rely upon the singleRun AOT compiled programs because the SourceFiles will not contain the source
    * with the latest fixes applied. Therefore we fallback to creating the quickest possible isolated program from the updated source.
+   *
+   * Note: This fallback is only needed for the legacy `project` option which uses AOT compiled programs.
+   * When `projectService` is used, the TypeScript language service always provides up-to-date programs,
+   * so no fallback is necessary. Additionally, external parsers like vue-eslint-parser may call
+   * parseAndGenerateServices() multiple times for the same file in a single lint pass, which would
+   * incorrectly trigger this fallback.
    */
   if (parseSettings.singleRun && tsestreeOptions.filePath) {
     parseAndGenerateServicesCalls[tsestreeOptions.filePath] =
@@ -234,7 +240,8 @@ export function parseAndGenerateServices<
   const { ast, program } =
     parseSettings.singleRun &&
     tsestreeOptions.filePath &&
-    parseAndGenerateServicesCalls[tsestreeOptions.filePath] > 1
+    parseAndGenerateServicesCalls[tsestreeOptions.filePath] > 1 &&
+    !parseSettings.projectService
       ? createIsolatedProgram(parseSettings)
       : getProgramAndAST(parseSettings, hasFullTypeInformation);
 


kirkwaiblinger:
Hi @ryo-chin, any chance that you can add some testing to demonstrate this change works? This would help us out in reviewing.

JoshuaKGoldberg:
Oh, shoot - I missed @kirkwaiblinger's comment. Sorry about that. I shouldn't have merged. There are some unit tests for this function in `tests/lib/semanticInfo.test.ts`, I'll add those now.

JoshuaKGoldberg:
Sent: #12152

PR #12067, opened 2026-02-15
chore(eslint-plugin): add default excludes to vitest

StyleShit:
I added this exclude in a previous PR, but suddenly realized that I forgot to add the default excludes, meaning that vitest will try to look for tests inside `node_modules` and such


jayaddison:
@JoshuaKGoldberg NB: I think this might be missing from the changelog for the v8.56.1 release?

JoshuaKGoldberg:
It's intentionally filtered out because it only touches internals of typescript-eslint, not anything end-users would see.

jayaddison:
@JoshuaKGoldberg Ok; thank you!

PR #12082, opened 2026-02-24
chore: shard eslint-plugin tests

StyleShit:
Continuing on #11204 

This PR adds [sharding](https://vitest.dev/guide/improving-performance.html#sharding) for the eslint-plugin package

TBH, I'm not sure where the sweet spot is.
The total run time [with sharding](https://github.com/typescript-eslint/typescript-eslint/actions/runs/22342824313?pr=12082) of 4 is `5m35s` compared to `7m42s` [without sharding](https://github.com/typescript-eslint/typescript-eslint/actions/runs/22340211291) (~27% less total duration), but it requires more machines in total, so... idk whether it's helping or not (in terms of freeing up machines)

I'm referring to this part:
> Jobs/runs have been getting stuck waiting to start -> then for GHA runners to free up.
https://github.com/typescript-eslint/typescript-eslint/issues/11204#issue-3057141208

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,21 +190,25 @@ jobs:
         node-version: [18, 20]
         package:
           [
-            'ast-spec',
-            'eslint-plugin',
-            'eslint-plugin-internal',
-            'parser',
-            'project-service',
-            'rule-schema-to-typescript-types',
-            'rule-tester',
-            'scope-manager',
-            'tsconfig-utils',
-            'type-utils',
-            'typescript-eslint',
-            'typescript-estree',
-            'utils',
-            'visitor-keys',
+            { name: 'ast-spec', params: '' },
+            { name: 'eslint-plugin', params: '--shard=1/4' },
+            { name: 'eslint-plugin', params: '--shard=2/4' },
+            { name: 'eslint-plugin', params: '--shard=3/4' },
+            { name: 'eslint-plugin', params: '--shard=4/4' },
+            { name: 'eslint-plugin-internal', params: '' },
+            { name: 'parser', params: '' },
+            { name: 'project-service', params: '' },
+            { name: 'rule-schema-to-typescript-types', params: '' },
+            { name: 'rule-tester', params: '' },
+            { name: 'scope-manager', params: '' },
+            { name: 'tsconfig-utils', params: '' },
+            { name: 'type-utils', params: '' },
+            { name: 'typescript-eslint', params: '' },
+            { name: 'typescript-estree', params: '' },
+            { name: 'utils', params: '' },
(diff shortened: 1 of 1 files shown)


rubiesonthesky:
Does the sharding use previous test time data to divide the tests? If it doesn’t have timing data available, sharding can lead one shard taking a lot more time than the others. It may not be a problem here, but if you want to make the CI pipeline faster, then you would want every shard to finish same time. (I tried to check this and it seems that this is not something that GitHub can offer)

Finding the sweet spots for these settings is always tiresome. Using 4 shards with it having that little effect sounds to me that it’s doing same job in multiple shards or that they are not balanced.

I’m not sure if GitHub reports the total minutes used somehow, but it could be useful to check if the total compute minutes used went down or stayed almost same. 

I think the test suite is big, but if possible I would aim for a shard to take max 3 min time. So maybe 8 shards could be tested? 

I wonder also does it matter if the sharding is done manually like here or with GitHub actions matrix parameters.

StyleShit:
> Does the sharding use previous test time data to divide the tests?

AFAIK, no, it's something basic like:

```ts
allTests
	.sort()
	.slice(currentShard - 1, currentShard * amountPerShard)
```

> ... every shard to finish same time

Seems like they're pretty close:

No coverage (total machine time):
1. `1m26s`
2. `1m34s`
3. `1m49s`
4. `1m25s`

With coverage (total machine time):
1. `1m19s`
2. `1m32s`
3. `1m48s`
4. `1m19s`

> ...doing same job in multiple shards

Well, that's why it's hard to find the same spot
Every machine has to checkout, install deps, build, etc.

> if GitHub reports the total minutes used

This is exactly what GitHub reports, and these are the times that I mentioned in the [PR's comment](https://github.com/typescript-eslint/typescript-eslint/pull/12082#issue-3982277758):

time with sharding

time without sharding

> ... a shard to take max 3 min time

The whole suite without sharding is taking about 3min IIRC. Am I missing something?


> sharding is done manually like here or with GitHub actions matrix parameters

Not sure what this means, could you please elaborate?

rubiesonthesky:
What I meant with total minutes was combination of duration of the CI run multiplied with the machines. In non-free projects the billed minutes is usually the important bit. 
That said I misunderstood what the duration reduction meant in OP. I thought it referred only the sharded test steps but I was wrong.

The matrix parameters that I referred is this what Playwright for example uses here https://playwright.dev/docs/test-sharding#github-actions-example - so same what is now used for different node versions. This probably doesn’t matter here.

If the shards seem to take similar time with each other, you can disregard my comment about that too. 

I have spend too much time with tuning things like this but I have not done that with GitHub Actions, so my comments probably weren’t helpful at all.

PR #12086, opened 2026-02-25
feat(eslint-plugin): [strict-void-return] add fix suggestions

phaux:
Adds fixes which were originally part of #9707 so they can be discussed separately.

Most notable example:

```js
button.addEventListener("click", async (ev) => {
  await doSomethingAsync(ev)
})
// ❌ strict-void-return: Async function used in a context where a void function is expected.
```

`strict-void-return` shows a suggestion to wrap the function body in an async IIFE and remove async keyword from outer function:

```js
button.addEventListener("click", (ev) => {
  (async () => {
    await doSomethingAsync(ev)
  })()
  // ❌ no-floating-promises: Promises must be awaited, end with a call to .catch...
})
```

Then `no-floating-promises` kicks in and tells you to handle the error:

```js
button.addEventListener("click", (ev) => {
  (async () => {
    await doSomethingAsync(ev)
  })().catch((err) => {
    console.error(err)
  })
})
```

This guides the user step by step towards a working solution which satisfies all the rules. This will help more users discover this approach instead of being forced to disable the rule because they don't know what to do.


packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -420,5 +446,49 @@ export default util.createRule<Options, MessageId>({
...
+        util.NullThrowsReasons.MissingToken('async keyword', funcNode.type),
+      );
+      const afterAsyncToken = util.nullThrows(
+        sourceCode.getTokenAfter(asyncToken),
+        util.NullThrowsReasons.MissingToken(
+          'token after async keyword',
+          funcNode.type,
+        ),
+      );
+      return fixer.removeRange([asyncToken.range[0], afterAsyncToken.range[0]]);
```
kirkwaiblinger: We usually just remove the token, rather than up until the next token, in order to avoid deleting comments. For example, this test case deletes `important comment` ([playground](https://deploy-preview-12086--typescript-eslint.netlify.app/play/#ts=5.9.2&fileType=.tsx&code=CYUwxgNghgTiAEYD2A7AzgF3gIwK4Y1QC54BRCEAWxBQwG4AoUSWBAM1xTAwEtV5gSAMpJqGABY8UAcwCCaAJ5cAFADp1saWhJQUCgJQkACjFE80IADwA3JD2AA%2BRgwZ4CqVVGDBS1mhgAZcwwaEBhlACJIHjAAawiAGngoRS54AHoAKngeSgAHJBgMXSxkSmpaeEz0%2BGUQa314AF4HeABvBnhkgHcoHixBETFJGXklMDqGhgBffQYgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZfKsugNwHtKAE1rRE%2BWNCbooAdwCGkyODABfECqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3QAacDUipkAIwpgA2pAAmCSOKjoSRdAA88kALpSAviDNA&tokens=false))

```ts
declare const button: Element;
declare function doSomethingAsync(...args: any): Promise<void>;


button.addEventListener("click", async /* important comment */ (ev) => {
  await doSomethingAsync(ev)
})
```

Would you mind tweaking that to avoid removing the comment?

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -357,6 +364,25 @@ export default util.createRule<Options, MessageId>({
...
       if (funcNode.async) {
         // The provided function is an async function.
+
+        if (
+          funcNode.body.type === AST_NODE_TYPES.BlockStatement &&
+          funcNode.body.body.length > 0
+        ) {
+          // This function has a block body.
+          // Suggest wrapping its body in an async IIFE.
+          return context.report({
```
kirkwaiblinger, line 402: The report is the same, only the `suggest` is different... Usually, we'd put this as part of one `report()` call and just pass in the `suggest` data. Would you mind restructuring that?

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -420,5 +446,49 @@ export default util.createRule<Options, MessageId>({
 
       // No invalid returns found. The function is allowed.
     }
+
+    /**
+     * Wraps function body in an inner async IIFE which allows the outer function to be non-async and return void.
+     * The passed function node must be an async function with a block body.
```
kirkwaiblinger: Is there a reason for this? Surely we can also rewrite functions that have an expression body?

kirkwaiblinger (changes requested):

JoshuaKGoldberg:
👋 Just checking in @phaux, do you still have time for this? Or would you like us to take over?

phaux:
@JoshuaKGoldberg sorry, I was busy with real life. I will work on it this weekend.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: added a test case and fixed

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: it originally had fixer for it but I removed it to make it simple. I added it back and extracted similar logic in no-confusing-void-expr to utils.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: I don't see an obvious way to do it. I think its fine like this.

phaux:
Sorry again for delay. It's because I haven't been working on any project which uses eslint for some time but now that changed.

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
```diff
@@ -1954,6 +2167,16 @@ ruleTester.run('strict-void-return', rule, {
           column: 25,
           line: 4,
           messageId: 'asyncFunc',
+          suggestions: [
+            {
+              messageId: 'suggestWrapInAsyncIIFE',
+              output: `
+        type Cb = () => void;
+        declare function Foo(props: { cb: Cb; s: string }): unknown;
+        return <Foo cb={ function () { (async () => {})(); }} s="!@#jp2gmd" />;
```
StyleShit: Should cases of an empty function body only remove the async keyword instead of creating an empty, useless async IIFE?

Maybe we can add this as another suggestion in these cases?

[Playground](https://deploy-preview-12086--typescript-eslint.netlify.app/play/#ts=5.9.2&fileType=.tsx&code=O4SwdgJg9sB0CGEIFEBuBTMAXAMiAzlpugE4AUARAMYA2IVA1hQDQAE8%2BAnmFawGYBXHlhBQwrMgEpWAb1YBfSQChQkGAiRpMuAkTClKtekzYduvKawC8APlkLlS%2BACNCJeFSytaHfKwBiUFCySqzsrljunt7OUgBcrKhQIBAA3ErySj74fgBC8CSs6AAeehB%2BgcEyoexcPDGWMpnyQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6ZfKsugNwHtKAE1rRE%2BWNCbooAdwCGkyODABfECqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3QAacDUipkAIwpgA2pAAmCSOKjoSRdAA88kALpSAviDNA&tokens=false)

StyleShit (commented):
Sorry for the late review
LGTM, thanks!

Added one question/suggestion

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
phaux: Original PR had this and I wanted to reintroduce every feature one by one in separate PRs but I might as well add them all at once by copying the whole rule code from that old commit :D

phaux:
Added all the suggestions. Some of them might be changed to auto fixes if you want.

phaux:
Now that 

```js
button.addEventListener("click", async (ev) => {
  await doSomethingAsync(ev)
})
```

can be auto fixed to

```js
button.addEventListener("click", (ev) =>
  void (async () => {
    await doSomethingAsync(ev)
  })()
)
```

it doesn't trigger no-floating-promises anymore, because it ignores void operator by default, which is unfortunate.

I can revert it but a better idea would be to change the default config of no-floating-promises to not ignore void in future release, at least in the strict config.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -23,13 +30,18 @@ export default util.createRule<Options, MessageId>({
...
     messages: {
       asyncFunc:
         'Async function used in a context where a void function is expected.',
       nonVoidFunc:
         'Value-returning function used in a context where a void function is expected.',
       nonVoidReturn:
         'Value returned in a context where a void return is expected.',
+      suggestAddVoidOp: 'Add a void operator to discard the return value.',
+      suggestMakeSyncFunc: 'Make the function synchronous.',
+      suggestRemoveFuncBody: 'Remove the function body.',
```
StyleShit: The user will see something like "Async function used in a context where a void function is expected. Suggestion: Async function used in a context where a void function is expected."

IMO it's not informative enough. Maybe we should explain why removing the body is "ok" in this case

It also makes the function synchronous, which is not mentioned here

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -387,22 +399,114 @@ export default util.createRule<Options, MessageId>({
...
       if (funcNode.async) {
         // The provided function is an async function.
-        // Async functions aren't allowed.
+
+        if (
+          funcNode.body.type === AST_NODE_TYPES.BlockStatement &&
+          funcNode.body.body.length === 0
+        ) {
+          // This async function is empty, e.g. `async () => {}`.
+          // Replace it with a regular empty function.
```
StyleShit: [Nitpick]

```suggestion
          // Replace it with a sync empty function.
```

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -387,22 +399,114 @@ export default util.createRule<Options, MessageId>({
...
+              {
+                messageId: 'suggestAddVoidOp',
+                fix: fixer =>
+                  addVoidToArrowFix(
+                    fixer,
+                    funcNode as TSESTree.ArrowFunctionExpression,
+                  ),
+              },
+            ],
+          });
```
StyleShit: [~Nitpick~, See edit below]

I think it's more readable to have the truthy case as the early return (but maybe it's just me?)

e.g.:

```TS
if (ASTUtils.hasSideEffect(funcNode.body, sourceCode)) {
	return // add void
}

return // remove body
```

(also in other places where applicable)

EDIT: This is the exact same report except the suggestion. can we do something like this instead?

```TS
return context.report({
	// ... all shared props

	suggest: [ASTUtils.hasSideEffect(funcNode.body, sourceCode) ? {
		// add void
	} : {
		// remove body
	}]
})
```

or maybe even always suggest both and let the user decide?

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -387,22 +399,114 @@ export default util.createRule<Options, MessageId>({
...
         });
       }
 
+      // This function is a regular or arrow sync function.
+
       if (funcNode.body.type !== AST_NODE_TYPES.BlockStatement) {
-        // The provided function is an arrow function shorthand without braces.
+        // This function is an arrow function shorthand without braces.
+
+        if (!ASTUtils.hasSideEffect(funcNode.body, sourceCode)) {
```
StyleShit: Same as above regarding condition/grouping

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
```diff
@@ -1129,6 +1238,44 @@ ruleTester.run('strict-void-return', rule, {
...
+          messageId: 'asyncFunc',
+          suggestions: [
+            {
+              messageId: 'suggestWrapInAsyncIIFE',
+              output: `
+        declare function foo(cb: () => void): void;
+        foo( function () { (async () => {
+          return -Math.random();
+        })(); });
+      `,
```
StyleShit, line 1289: Wrapping this with IIFE results in a weird code that nobody would actually write.

I think in such cases we should suggest both removing async and wrapping with IIFE, because we don't know what would be the "correct" fix for the user

StyleShit (changes requested):
After reading most of the test cases, the suggestions feels slightly unexpected / confusing to me as a user

I think we should simplify things and always suggest all possible fixes. But maybe I'm wrong here. I would wait for other team members to hear their thoughts on this

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
phaux: That would require scanning the whole function to check if it doesn't contain awaits. Right now removing async is only suggested when the function is empty or performed together with removing whole body (no side effects)

StyleShit: > ... require scanning the whole function

Exactly, which is why I think we shouldn't do that, and suggest both "wrap with IIFE" & "remove async", so the user will decide

phaux: But removing async when there are awaits will result in invalid code

StyleShit: Because it's a suggestion, and because `no-floating-promises` also adds `await`s without adding `async`, it's fine IMO

But pinging @typescript-eslint/triage-team for opinions

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: I removed this suggestion altogether

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: reduced duplicated code

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
phaux: I agree that we shouldn't be smart and guess which suggestion will be the best. I changed it back to suggest IIFE unconditionally. Maybe the user actually wants to insert an empty IIFE into an empty async function and continue writing code inside the IIFE. I think that's useful.

I also removed the simple suggestions because they aren't even worth having in code and barely save any keystrokes. Selecting one word and pressing backspace is faster than opening the quick fix menu and choosing a suggestion.

So basically it's like it was before except there's also "add void operator" suggestion.

StyleShit (changes requested):

StyleShit (dismissed):

phaux:
One extra thing worth considering is to either:

- change "add void operator to arrow" to be an auto fix.
- change the identical auto fix in `no-confusing-void-expression` to be a suggestion.

Right now they do the same thing (even use the same fixer function), but one is a suggestion and the other is an auto fix.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -450,5 +505,68 @@ export default util.createRule<Options, MessageId>({
...
+      // Wrap body in async IIFE
+      const bodyRange = util.getRangeWithParens(funcNode.body, sourceCode);
+      if (funcNode.type === AST_NODE_TYPES.ArrowFunctionExpression) {
+        yield fixer.insertTextBeforeRange(bodyRange, 'void (async () => ');
+        yield fixer.insertTextAfterRange(bodyRange, ')()');
+      } else {
+        yield fixer.insertTextBeforeRange(bodyRange, '{ (async () => ');
+        yield fixer.insertTextAfterRange(bodyRange, ')(); }');
+      }
+    }
```
kirkwaiblinger, line 581: We may be able to avoid making the `util.getRangeWithParens` helper here... As far as I can tell, this seems to work just as well:

```suggestion
    /**
     * Makes the function sync and wraps the body in an inner async IIFE.
     */
    function* wrapInAsyncIIFEFix(
      fixer: TSESLint.RuleFixer,
      funcNode: TSESTree.ArrowFunctionExpression | TSESTree.FunctionExpression,
    ): Generator<TSESLint.RuleFix> {
      yield* makeSyncFuncFix(fixer, funcNode);

      // Wrap body in async IIFE
      if (funcNode.type === AST_NODE_TYPES.ArrowFunctionExpression) {
        const arrowToken = nullThrows(
          sourceCode.getTokenBefore(funcNode.body, {
            filter: token => token.value === '=>',
          }),
          util.NullThrowsReasons.MissingToken(
            'arrow token',
            'before arrow function body',
          ),
        );

        yield fixer.insertTextAfter(arrowToken, ' void (async () =>');
        yield fixer.insertTextAfter(funcNode, ')()');
      } else {
        yield fixer.insertTextBefore(funcNode.body, '{ (async () => ');
        yield fixer.insertTextAfter(funcNode.body, ')(); }');
      }
    }
```

Maybe there are differences with how comments show up? 

WDYT? FWIW I'm entirely ok with the `getRangeWithParens()` helper too if it gives better results.

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
kirkwaiblinger: I think it makes enough sense to convert to a non-returning sync function and wrap the body with async IIFE. This is more or less how I would write it 🤷

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -450,5 +505,68 @@ export default util.createRule<Options, MessageId>({
...
+        node: funcNode.body,
+        sourceCode,
+        wrap: code => `void ${code}`,
+      })(fixer);
+    }
+
+    /**
+     * Makes the function sync and wraps the body in an inner async IIFE.
+     */
+    function* wrapInAsyncIIFEFix(
```
kirkwaiblinger, line 556: I had to think long and hard about `function` vs arrow function semantics and `this` context here 🙃 I'm pretty sure your solution is the correct one: preserve the _outer_ function style (arrow function or `function`) and _always_ use an arrow function on the inner function (ensuring it has the same `this` value as the surrounding function, and therefore the same `this` value as it would have seen before the suggestion).

Would you kindly add a comment to that hints at this to future users? Something like

```ts
// This uses the original `function`/arrow style for the _outer_ function
// and an arrow function for the _inner_ function, so that `this` in the
// body will always refer to the same value as before.
```

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -450,5 +505,68 @@ export default util.createRule<Options, MessageId>({
...
+      fixer: TSESLint.RuleFixer,
+      funcNode: TSESTree.ArrowFunctionExpression | TSESTree.FunctionExpression,
+    ): Generator<TSESLint.RuleFix> {
+      yield* makeSyncFuncFix(fixer, funcNode);
+      // Wrap body in async IIFE
+      const bodyRange = util.getRangeWithParens(funcNode.body, sourceCode);
+      if (funcNode.type === AST_NODE_TYPES.ArrowFunctionExpression) {
+        yield fixer.insertTextBeforeRange(bodyRange, 'void (async () => ');
```
kirkwaiblinger: I guess this is fine but I'd much prefer just using an arrow function expression with body, i.e. `(/* params */) => { (async () => { /* body */ })(); }`, rather than a `void` expression. The `void` keyword is to be avoided whenever possible IMO.

packages/eslint-plugin/src/rules/strict-void-return.ts
```diff
@@ -387,22 +395,69 @@ export default util.createRule<Options, MessageId>({
...
+
       if (funcNode.body.type !== AST_NODE_TYPES.BlockStatement) {
-        // The provided function is an arrow function shorthand without braces.
+        // This function is an arrow function shorthand without braces.
         return context.report({
           node: funcNode.body,
-          messageId: `nonVoidReturn`,
+          messageId: 'nonVoidReturn',
+          suggest: ASTUtils.hasSideEffect(funcNode.body, sourceCode)
+            ? // This function has side effects, e.g. `() => doSomething()`.
```
kirkwaiblinger, line 444: Same here, much prefer just adding braces... `() => { doSomething(); }`

kirkwaiblinger:
> Now that
> 
> ```js
> button.addEventListener("click", async (ev) => {
>   await doSomethingAsync(ev)
> })
> ```
> 
> can be auto fixed to
> 
> ```js
> button.addEventListener("click", (ev) =>
>   void (async () => {
>     await doSomethingAsync(ev)
>   })()
> )
> ```
> 
> it doesn't trigger no-floating-promises anymore, because it ignores void operator by default, which is unfortunate.
> 
> I can revert it but a better idea would be to change the default config of no-floating-promises to not ignore void in future release, at least in the strict config.

Why not fix to

```ts
button.addEventListener("click", (ev) => {
  (async () => {
    await doSomethingAsync(ev);
  })();
});
```

instead?

kirkwaiblinger (commented):
Several comments/questions, but I don't think any of them blocking...

Most are around why we'd want to add `void` operator (yucky) where `() => { /* never returns */ }` would work instead. I don't see this discussed, and it seems to contradict the description.

packages/eslint-plugin/tests/rules/strict-void-return.test.ts
kirkwaiblinger: If the function doesn't contain `await`s at all, that's just a [`require-await`](https://typescript-eslint.io/rules/require-await/) concern, I'd think

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: good idea

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: done

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: This is what I originally implemented but I used it for a while and decided to change it. It all boils down to which formatting you prefer after you apply all the suggested code actions and format the code. Let's assume we put everything inline initially. For example:

```jsx
function App() {
  return (
    <button
      onClick={async () => {
        await doSomethingAsync()
      }}
    >
      Click
    </button>
  )
}
```

Then we do the following steps:

1. Wrap callback in async IIFE.
2. Handle the floating promise by adding a `.catch` after the IIFE
3. Extract function to constant in enclosing scope (code action).

And we end up with:

```jsx
function App() {
  async function handleClick() {
    await doSomethingAsync()
  }
  return <button onClick={() => void handleClick().catch(handleError)}>Click</button>
}
```

If we didn't use `void` before the IIFE we would end up with this:

```jsx
function App() {
  async function handleClick() {
    await doSomethingAsync()
  }
  return (
    <button
      onClick={() => {
        handleClick().catch(handleError)
      }}
    >
      Click
    </button>
  )
}
```

Which is 7 extra lines in extreme case. For regular callback code (for example addEventListener) it's usually 2 per callback. 

Another example:

```js
observable.subscribe({
  next: (value) => void handleValue(value).catch(handleError),
  error: (err) => handleError(err),
  complete: () => void handleComplete().catch(handleError),
})

// vs

observable.subscribe({
  next: (value) => {
    handleValue(value).catch(handleError)
  },
  error: (err) => {
    handleError(err)
  },
  complete: () => {
    handleComplete().catch(handleError)
  },
})
```

If I want the more compact formatting I have to do one more extra step (replace braces with void manually). Of course this is just a matter of preference. If somebody prefers the spread out style and void is the default then *they* must do the extra step.

You could argue that that the spread out style is better because coverage tools show   coverage per line so its clearer whether the callback was hit. I sometimes use that style just to see coverage better but sometimes I reformat it back after.

So basically I'm conflicted on this one. I might change it back just because other rules have so many problems with it.

packages/eslint-plugin/src/rules/strict-void-return.ts
phaux: See my other comment. It boils down to whether you prefer more compact or spread out style:

```jsx
<button
  onClick={() => {
    doSomething()
  }}
>
  Click
</button>

// vs

<button onClick={() => void doSomething()}>Click</button>
```

I think the compact one looks better more often especially in JSX.

If this stays as a suggestion and not an autofix then we can suggest both.

packages/eslint-plugin/src/rules/strict-void-return.ts
kirkwaiblinger: I hear you're saying here, but I'm still in the camp of stylistically preferring the `() => { /* stuff */ }` approach. I totally acknowledge it's really just a personal preference thing at the end of the day.

In any case, it's been a while (I've been low on bandwidth to review), you've put a lot of good work into this, and I think it's better to ship this, and we can tweak the autofixes down the road if needed, rather than continuing to block this work.

Thank you for thoughtful discussion!

packages/eslint-plugin/src/rules/strict-void-return.ts
kirkwaiblinger: I do like the idea of suggestin both 🤔 

Perhaps a future enhancement!

kirkwaiblinger (dismissed):

StyleShit (dismissed):

StyleShit:
LGTM, but there are some conflicts
Will merge when they're resolved

StyleShit (dismissed):

StyleShit (approved):

PR #12089, opened 2026-02-28
fix(eslint-plugin): [no-base-to-string] fix false positive for toString with overloads

### LINKED ISSUE #11945: Bug: [@typescript-eslint/no-base-to-string] False positive on * import containing toString function with overloads

### Issue Description

If you `*` import everything from a module, and that module contains a `toString` method that has function overloads, then the `@typescript-eslint/no-base-to-string` rule incorrectly reports that the `toString` call is going to use the default implementation.

The function overloads appear to be an important part of the repro. Without them the rule doesn't report anything.

https://github.com/bschlenk/no-base-to-string-repro

### Repro Steps

1) clone the repo
2) `pnpm install`
3) `pnpm lint`

### Versions

| `typescript-eslint`     | `8.52.0` |
| `TypeScript`                           | `5.9.3` |
| `ESLint`                               | `9.39.2` |
| `node`                                 | `22.17.1` |

bschlenk:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11945

## Overview

Instead of returning immediately if there are multiple declarations (overloads), we check each overload to see if it's the base `toString`, and if _any_ of them are not, we return without reporting the line.

Added tests for both class `toString` member with overloads, and an `import *` of a `toString` function with overloads.

### DIFF
--- a/packages/ast-spec/src/expression/ObjectExpression/fixtures/_error_/property-modifier-abstract/snapshots/1-TSESTree-Error.shot
+++ b/packages/ast-spec/src/expression/ObjectExpression/fixtures/_error_/property-modifier-abstract/snapshots/1-TSESTree-Error.shot
@@ -3,5 +3,5 @@
 exports[`AST Fixtures > expression > ObjectExpression > _error_ > property-modifier-abstract > TSESTree - Error`]
 TSError
 > 1 | ({abstract property: 1})
-    |   ^^^^^^^^ 'abstract' modifier can only appear on a class, method, or property declaration.
+    |   ^^^^^^^^ 'abstract' modifier cannot be used here.
   2 |


packages/eslint-plugin/tests/rules/no-base-to-string.test.ts
```diff
@@ -2429,5 +2455,21 @@ a.toString();
         },
       ],
     },
+    {
+      code: `
+        type Mapped = { [K in 'toString']: () => string };
```
kirkwaiblinger, line 2462: What exactly is this test case trying to show? Maybe a comment would help here?

packages/eslint-plugin/tests/rules/no-base-to-string.test.ts
```diff
@@ -398,6 +398,32 @@ import { String } from 'foo';
...
+    `
+class Foo {
+  toString(prefix?: string): string {
+    return (prefix ?? '') + 'Hello, world!';
+  }
+}
+'' + new Foo();
+    `,
+    `
+declare module 'guid' {
```
kirkwaiblinger, line 419: (praise) Nice job getting a single-file test case working like this 🙂

packages/eslint-plugin/src/rules/no-base-to-string.ts
```diff
@@ -354,19 +354,20 @@ export default createRule<Options, MessageIds>({
...
+        // see https://github.com/typescript-eslint/typescript-eslint/issues/8585
+        // see https://github.com/typescript-eslint/typescript-eslint/issues/11945
         if (
-          !ts.isInterfaceDeclaration(declarations[0].parent) ||
-          declarations[0].parent.name.text !== 'Object'
+          declarations.some(
+            declaration =>
+              !ts.isInterfaceDeclaration(declaration.parent) ||
+              declaration.parent.name.text !== 'Object',
+          )
```
kirkwaiblinger, line 372: Up to you but I think this might read better (and correspond to the comment better) if we apply deMorgan

```suggestion
          declarations.some(
            declaration =>
              !(ts.isInterfaceDeclaration(declaration.parent) &&
              declaration.parent.name.text === 'Object'),
          )
```

WDYT?

bschlenk: sounds good!

packages/eslint-plugin/tests/rules/no-base-to-string.test.ts
bschlenk: Yeah this one is a bit weird, it was the only way I could figure out how to hit 100% coverage. Because of the `if (!declarations?.length) {` on line 357, we need a case where we have a `toString` method to cause us to enter the loop, but where `getDeclarations()` returns undefined.

I'll add a comment

bschlenk:
I flipped the condition and added a comment to that weird test. Do I need to merge the branch with main? I'm used to just rebasing + force pushing but I saw your contributing guide said not to force push.

JoshuaKGoldberg (dismissed):
We've got you 👍 thanks for this! Very clever testing 🙂

PR #12098, opened 2026-03-04
docs: convert config examples to use `files` and `extends`

### LINKED ISSUE #12092: Docs: Always use `files` in config examples

### Suggested Changes

As of https://github.com/eslint/eslint/issues/20075 -> https://github.com/eslint/eslint/pull/20158, the ESLint docs suggest to always include `files` in config objects. That way you only enable rules on a specific set of files known to work on them. This is important because ESLint is (and actually has been for a while) a multi-language linter. Rules made for one language will often break on files of other languages.

Suggestion: let's follow the ESLint docs guidance and always include `files` in our suggested configs?

### Affected URL(s)

https://typescript-eslint.io/getting-started/legacy-eslint-setup
https://typescript-eslint.io/getting-started/typed-linting
etc.

### Additional Info

See issues from users with borked configs:

* HTML: #11300
* JSON: #12091

Typed rules & parser services in particular are what surface as breakages.

💖

kirkwaiblinger:
## PR Checklist

- [x] Addresses an existing open issue: fixes #12092 

## Overview

Wraps our config examples in 
```diff
export default defineConfig(
+ {
+   files: ['**/*.{js,cjs,mjs,jsx,ts,cts,mts,tsx}'],
+   extends: [
    stuff,
+   ]
+ }
);
```
in most places, wherever it seemed to make sense to do so.

Be sure to ignore whitespace when viewing diff!

I've also gone and converted our own setup to this pattern, so that we dogfood appropriately.


docs/getting-started/Quickstart.mdx
```diff
@@ -38,17 +38,18 @@ import eslint from '@eslint/js';
 import { defineConfig } from 'eslint/config';
 import tseslint from 'typescript-eslint';
 
-export default defineConfig(
-  eslint.configs.recommended,
-  tseslint.configs.recommended,
-);
+export default defineConfig({
+  files: ['**/*.{js,cjs,mjs,jsx,ts,cts,mts,tsx}'],
```
JoshuaKGoldberg: 😬 I don't like throwing such as large glob in front of people. Especially since the vast majority of users are only ever going to want `.ts` or `.{ts,tsx}`. Maybe we could do something like...


```suggestion
  files: ['**/*.ts'], // (add globs as needed for any other file extensions)
```

...or maybe a Flint-style `tseslint.globs.ts`?

docs/getting-started/Quickstart.mdx
```diff
@@ -38,17 +38,18 @@ import eslint from '@eslint/js';
...
+  extends: [eslint.configs.recommended, tseslint.configs.recommended],
+});
 ```
 
 This code will enable our [recommended configuration](../users/Shared_Configurations.mdx) for linting.
 
 #### Details
 
 - `defineConfig(...)` is an optional helper function built in to current versions of ESLint. See [the ESLint configuration docs](https://eslint.org/docs/latest/use/configure/configuration-files) for more detail.
+- `files`/`extends`: this setup is an optional feature of `defineConfig(...)` that restricts the `extends` config to JS and TS files. We recommend this in order to avoid issues if you enable ESLint to run on other file types, but it's entirely optional.
```
JoshuaKGoldberg: This almost makes it sound like the `extends` feature itself is what restricts to JS and TS, not the particular usage of it.

```suggestion
- `files`/`extends`: this setup is an optional feature of `defineConfig(...)` that restricts the `extends` config to this config object's files. We recommend this in order to avoid issues if you enable ESLint to run on other file types, but it's entirely optional.
```

JoshuaKGoldberg (changes requested):
I like the direction overall, but I think the glob is going to be confusing to people - or at least seem complex. WDYT about simplifying it in some way?

docs/getting-started/Quickstart.mdx
kirkwaiblinger: I think that providing the TS glob may be the best choice. I dislike `*.ts` as it makes the default case un-general, and this would potentially need to be explained in every location in which it appears, which would be distracting. I also dislike the long, exhaustive glob being written everywhere, for obvious reasons. With a glob exported by tseslint, we get to provide documentation along the lines of "what this glob is and why you can ignore it if you prefer to write your own".

kirkwaiblinger:
Marking this work blocked on https://github.com/typescript-eslint/typescript-eslint/issues/12102 (see https://github.com/typescript-eslint/typescript-eslint/pull/12098#discussion_r2886868069)

vladshcherbin:
Would be nice to see this merged. This is important part of configuration and waiting for exported globs support doesn't feel like a real blocker for users reading current docs.

kirkwaiblinger:
Marked as a draft while https://github.com/typescript-eslint/typescript-eslint/issues/12102 &rarr; https://github.com/typescript-eslint/typescript-eslint/pull/12105 proceeds

kirkwaiblinger:
We decided to remove #12102 as a blocker and move forward with this in the meantime. Just need to converge on a resolution to https://github.com/typescript-eslint/typescript-eslint/pull/12098#discussion_r2886696815 without exported globs in order to proceed

docs/getting-started/Quickstart.mdx
```diff
@@ -38,17 +38,21 @@ import js from '@eslint/js';
 import { defineConfig } from 'eslint/config';
 import tseslint from 'typescript-eslint';
 
-export default defineConfig(
-  js.configs.recommended,
-  tseslint.configs.recommended,
-);
+export default defineConfig({
+  files: ['**/*.{js,ts}'],
```
JoshuaKGoldberg, line 42: I thought we'd resolved on including at least `mjs`, so that the `eslint.config.mjs` we tell folks to use will also be linted?

JoshuaKGoldberg (changes requested):
LGTM _except_ the whole `.mjs` inclusion shenanigan?

docs/getting-started/Quickstart.mdx
kirkwaiblinger: Oh, my takeaway from https://discord.com/channels/1026804805894672454/1503817469511602307/1503879343263186994 and the subsequent message is that we want to continue using `eslint.config.mjs` as the default config file type to demo, and that therefore we would accept the wart that `eslint.config.mjs` would be figuratively, but not literally, included in the example lint config. 😆

kirkwaiblinger: @bradzacher @Josh-Cena

kirkwaiblinger: I also tried to mitigate this with the note

> To enable linting for all standard JS and TS file extensions, consider using `files: ['**/*.{js,cjs,mjs,jsx,ts,cts,mts,tsx}']`.

JoshuaKGoldberg: Gotcha. I think that's a good compromise for now, yeah.

JoshuaKGoldberg (approved):
✅✅✅

PR #12109, opened 2026-03-07
feat(eslint-plugin): [no-misused-promises] check promise non-promise unions in conditionals

### LINKED ISSUE #5767: Feat: [no-misused-promises] non-thenable thenable union function does not require await in conditional

### Playground Link

https://typescript-eslint.io/play/#ts=4.8.4&sourceType=module&code=PQKgBApgzgNglgOwC5gAJIJ4AdoGMBOcWSAtNPMsAgPYkC2cUArlBACYlb7UOtQBcYAOQR83fELAhgAKBmYcYAGZMEuJHGoIAKtghgAvGADeMsMtXrNCAOpwkACwDKGNQEEEbN1Fe5BACn8ASkMAPjAAI2pqGAgAQwQQgB8wQJCDcIAFbl4IAB4omPiEUKCZAF85XC0oFBU1QXqrLV1FI1NzJo0tO0cXd09vXwD0rJzGCAA6fGgYgDcIfyU4mFYggBoKuWBgMABRAA8cdXZIMWp8MAdRfShEXH0u6xkdsBmkJnwEKEjo2ISwClsjwJmACn9iqEZHAlEtLJMnj17M5fB4vD41MEQh0wDsKkA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUELjAXxDyA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

### Repro Code

```typescript
type functionType = {
  functionWithSyncAndAsync: (() => boolean) | (() => Promise<boolean>)
}

const func: functionType = {
  functionWithSyncAndAsync: () => Promise.resolve(false),
}

// Expected error here since function
// returns boolean | Promise <boolean>
if(func.functionWithSyncAndAsync()) {
  //
}
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-misused-promises": ["error"],
  },
};
```

### tsconfig

```jsonc
{}
```

### Expected Result

I expected `if(func.functionWithSyncAndAsync())` to fail since it can return a Promise<boolean> which should be a misused promise.

### Actual Result

There was no error with the provided code.

### Additional Info

_No response_

### Versions

| `@typescript-eslint/eslint-plugin`     | `5.39.0` |
| `@typescript-eslint/parser`            | `5.39.0` |
| `node`                                 | `14.19.0` |

Nick-NCSU:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5767 

## Overview

Adds support to check conditionals based on these options:
`all` - Flags all conditionals for which at least one constituent is thenable.
`narrow` - Flags conditionals for which at least one constituent is thenable and no constituent is potentially falsy.
`none` - Flags conditionals for which all constituents of a union are thenable (current behavior).

I have attempted to cover the concerns in the comments of the original issue and the duplicate issues of it. It's possible there could still be a fourth group who would want only nullish values to be allowed to be narrowed and not all falsy values. So
```ts
declare const a: Promise<boolean> | boolean;
if(a) {}
```
would be seen as unintentional, but
```ts
declare const a: Promise<boolean> | null;
if(a) {}
```
would be seen as intentional. Definitely let me know if I should make some changes to the logic here.

Also worth mentioning is that the spread operator and the array predicates feature added in #9955 behave like the `all` option in this PR. Neither of the following are currently valid:
```ts
declare const promise: Promise<string> | string;

[1, 2, 3].filter(() => promise);

declare const fn: () => Promise<{}> | {};

console.log({ foo: 42, ...fn() });
```


packages/eslint-plugin/docs/rules/no-misused-promises.mdx
```diff
@@ -39,6 +39,21 @@ If you don't want to check conditionals, you can configure the rule with `"check
 
 Doing so prevents the rule from looking at code like `if (somePromise)`.
 
+The following sub-options are supported:
```
JoshuaKGoldberg: [Docs] None of these three sub-headings show configuring the rule or examples of invalid/valid code. That's necessary for readers to actually understand the option. Could you please add that in?

JoshuaKGoldberg: [Docs] Also, it's not super clear from just this section that `checksConditionals` is `boolean | (some object)`. And the type is complex enough that I'd bet most readers aren't deeply reading it.

```suggestion
`checksConditionals` can also be provided as an object to granularly specify how unions are handled in conditionals.
The following sub-options are supported:
```

JoshuaKGoldberg (changes requested):
👏 The core logic & tests look great, thanks for this! Very cool to see a PR ~3.5 years after the original issue from the author. ❤️ 

Just requesting changes on filling out docs. But I know docs can be tricky to work with, let us know if you'd like us to step in. Cheers~!

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
Nick-NCSU: I've added in an example of configuring the rule via the object and modified the headings to match the format to configure the rule.

As for the examples, it looks like for this rule all of the existing examples are located in the `Examples` section. I added in the new examples into that section, but should I move them to be in the `Options` section instead? I went ahead and added some sub-headings down there to make them more visible either way.

kirkwaiblinger:
> Also worth mentioning is that the spread operator and the array predicates feature added in #9955 behave like the `all` option in this PR. Neither of the following are currently valid:
> 
> ```ts
> declare const promise: Promise<string> | string;
> 
> [1, 2, 3].filter(() => promise);
> 
> declare const fn: () => Promise<{}> | {};
> 
> console.log({ foo: 42, ...fn() });
> ```

This is a good callout. 

On the spread operator, I agree with the behavior. For the array predicates, I think I disagree: we normally treat those as "conditionals" as well (for example they are considered "conditional" locations by strict-boolean-expressions and no-unnecessary-condition) and have them follow the same options as other conditional expressions.

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
```diff
@@ -39,6 +39,35 @@ If you don't want to check conditionals, you can configure the rule with `"check
 
 Doing so prevents the rule from looking at code like `if (somePromise)`.
 
+`checksConditionals` can also be provided as an object to granularly specify how unions are handled in conditionals. Default: `'none'`.
```
kirkwaiblinger, line 42: `'none'` is not the default value of `checksConditionals`... If I understand correctly, `true` is? Can you find a way to reword this to be more precise?

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
```diff
@@ -39,6 +39,35 @@ If you don't want to check conditionals, you can configure the rule with `"check
...
+        "flagUnions": "none"
+      }
+    }
+  ]
+}
+```
+
+#### `{ "flagUnions": "all" }`
+
+Flags any time a union containing a misused Promise is encountered in a conditional.
```
kirkwaiblinger, line 59:
```suggestion
Reports any time a union containing a misused Promise is encountered in a conditional.
```

Let's use the word "report" for this. We use "flag" informally when discussing, but in the docs, we almost universally use the word "report".

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
kirkwaiblinger, line 27: In this PR the options structure changes from 

```ts
type Options = {
    /** Whether to warn when a Promise is provided to conditional statements. */
    checksConditionals?: boolean;

    // ...
}
```

to 

```ts
type Options = {
    /** Whether to warn when a Promise is provided to conditional statements. */
    checksConditionals?: /**
       * Whether to warn when a Promise is provided to conditional statements.
       * Whether to disable checking all conditional statements.
       */
      | boolean
      /** Which forms of conditionals may be checked. */
      | {
          /** Whether to warn when a promise non-promise union is provided. */
          flagUnions?:
            | 'narrow'
            | 'none'
            /** Whether to warn when a promise non-promise union is provided. */
            | 'all';
        };

    // ...
}
```

The default used to be `true`. Let's be clear what the new default is (I think it's `{ flagUnions: 'none' }`?), and, especially, that `true` is an alias for that (so that people who previously might have specified `true` or relied on the default know what the new behavior is)

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
kirkwaiblinger, line 59:
```suggestion
Flags any time a union containing a Promise is encountered in a conditional.
```

What would the difference between a misused and non-misused promise in the union be? It's the presence in the conditional that makes it misused, no?

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
```diff
@@ -39,6 +39,35 @@ If you don't want to check conditionals, you can configure the rule with `"check
...
+
+#### `{ "flagUnions": "narrow" }`
+
+Flags when a Promise non-Promise union is encountered in a conditional only when
+no constituent of the union can be falsy. This allows patterns where the conditional
```
kirkwaiblinger, line 64: > only when no constituent of the union can be falsy

This has some weird consequences...

```ts
declare const promiseUnion: Promise<string> | string;

if (promiseUnion) {
}
```

The above code has a union constituent `string` which _can be falsy_, and therefore it's not flagged by the rule as a misuse of a promise. To me, this is definitely a case that you'd want flagged.

Depending on how you look at it, this also doesn't quite pass the test in https://github.com/typescript-eslint/typescript-eslint/issues/5767#issuecomment-1272653142, where `Promise<T> | T` should be flagged in the intermediate option, due to things like this:

```ts
type MaybeObject = { a: string } | undefined
declare function maybeObjectFunction(): Promise<MaybeObject> | MaybeObject

// flags under { "checksConditionals": { "flagUnions": "all" } }...
// but maybe shouldn't?

if (maybeObjectFunction()) {
}
```

I'm personally a bit skeptical whether these are the correct semantics, and would be interested for this to be discussed more... But, either way, even if we go with this version, I think the docs here should provide much more in the way of examples so that it's clear what the consequences of this are options are. Could you update the docs with some examples please?

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
```diff
@@ -39,6 +39,35 @@ If you don't want to check conditionals, you can configure the rule with `"check
...
+}
+```
+
+#### `{ "flagUnions": "all" }`
+
+Flags any time a union containing a misused Promise is encountered in a conditional.
+
+#### `{ "flagUnions": "narrow" }`
+
+Flags when a Promise non-Promise union is encountered in a conditional only when
```
kirkwaiblinger, line 63: > Promise non-Promise union

This is informal, jargon-y wording with questionable grammar, which is completely fine for technical discussions on github, but which isn't great for end-user docs... Can you find a way to clean it up for a general audience?

packages/eslint-plugin/docs/rules/no-misused-promises.mdx
```diff
@@ -126,7 +155,9 @@ If you don't want to check object spreads, you can add this configuration:
 
 {/* insert option description */}
 
-Examples of code for this rule with `checksConditionals: true`:
+#### `{ "flagUnions": "none" }`
+
+Examples of code for this rule with `checksConditionals: true` or `checksConditionals: { "flagUnions": "none" }`:
```
kirkwaiblinger, line 160: Building off of https://github.com/typescript-eslint/typescript-eslint/pull/12109/changes#r3259945527, this is the first hint that `true` and `{ "flagUnions": "none" }` are aliases here. Rather than it being implicitly shown, please state this explicitly.

packages/eslint-plugin/src/rules/no-misused-promises.ts
```diff
@@ -183,18 +229,31 @@ export default createRule<Options, MessageId>({
...
-        checkConditional(node.argument, true);
-      },
-      WhileStatement: checkTestConditional,
-    };
+    checksConditionals = parseChecksConditionals(checksConditionals);
+    const conditionalChecks: TSESLint.RuleListener = checksConditionals
+      ? {
+          'CallExpression > MemberExpression': checkArrayPredicates,
+          ConditionalExpression: node =>
+            checkTestConditional(node, checksConditionals.flagUnions),
```
kirkwaiblinger, line 237: Instead of passing the option `checksConditionals.flagUnions` down as a parameter, consider instead just accessing the captured value within the functions?

kirkwaiblinger (changes requested):
This looks pretty good (and is entirely correct as far as I can tell in terms of the added rule logic) but could use some improved documentation. Good work!

StyleShit:
👋 Hey @Nick-NCSU, just checking in - do you still have time to work on this?

PR #12116, opened 2026-03-09
chore: [ci] only run `eslint-plugin` tests on the lowest node version

bradzacher:
## Overview

- Fixes: https://github.com/typescript-eslint/typescript-eslint/issues/11204

This is an attempt at reducing the number of jobs we execute per PR.

Currently we spawn 3 jobs per package to run unit tests:
- windows node latest
- linux node latest
- linux node oldest

And we also spawn 4x sharded jobs for the plugin tests.

This means we spawn 13 packages x 3 + 4 x 3 = 51 jobs just for tests.
This is a bit obscene!

If we make the assumption that the plugin tests by themselves provide us enough coverage of all the packages then we can isolate the multi-node testing to just the plugin. So this means that instead we'll spawn 2 jobs for most packages and 4x3 for the plugin -- i.e. 13x2 + 4x3 = 38 jobs for tests (13 fewer jobs).

Hopefully this should free up some action bandwidth during high-load periods and reduce queueing.


JoshuaKGoldberg (dismissed):
IIRC there was past talk of wanting to make sure individual rules that deal with OS/Node.js-specific things work well in the matrix. But I can't find it on quick demand. And we're still getting slammed in CI. I vote 👍 !

StyleShit:
> IIRC there was past talk of wanting to make sure individual rules that deal with OS/Node.js-specific things work well in the matrix. But I can't find it on quick demand. And we're still getting slammed in CI. I vote 👍 !

Ah, sorry I'm late to the party
Here it is https://github.com/typescript-eslint/typescript-eslint/issues/11204#issuecomment-3717297569

Issue #5246, opened 2022-06-25
Docs: Line addition diffs don't have enough color contrast in text


JoshuaKGoldberg:
### Before You File a Documentation Request Please Confirm You Have Done The Following...

- [X] I have looked for existing [open or closed documentation requests](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+label%3Adocumentation) that match my proposal.
- [X] I have [read the FAQ](https://typescript-eslint.io/docs/linting/troubleshooting) and my problem is not listed.

### Suggested Changes

The red text on green line background in the color diffs has a [color contrast ratio of 3.38:1 (`#FF0000` on `#B8FAE0`)](https://webaim.org/resources/contrastchecker/?fcolor=FF0000&bcolor=B8FAE0), which is not strong enough for all users to be able to perceive the text. We should up the contrast to make sure the lines are readable by everyone - at least 4.5:1. See https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html for accessibility docs.

![Screenshot of the added lines diff with not enough color contrast for redon green text](https://user-images.githubusercontent.com/3335181/175779276-f01017e9-7796-4c91-97c7-d98acf1637e7.png)

(I personally found it a little difficult at first to read such a bold green background.)

Suggestion: how about darkening the code text to `#DD0000`, halving the opacity of the green background to `00ff9520`, and making the `+` glyphs bold?

![Screenshot of the added lines diff with enough color contrast for redon green text](https://user-images.githubusercontent.com/3335181/175779551-4d306bdd-4939-4565-abb2-79c08c11e447.png)

### Affected URL(s)

https://typescript-eslint.io/docs/linting/type-linting

JoshuaKGoldberg:
Absolutely! For visibility: if something is marked as `accepting prs` then anybody anywhere can tackle it. We'd love your contributions regardless of whether you've contributed before! ❤️

PR #12117, opened 2026-03-09
docs(website): improve added-line contrast in code diffs (#5246)

karimFin:
This improves color contrast for red string/char tokens within green "added" lines on the website code diffs.\n\n- Adjusts only token color inside .code-block-added-line to a darker red (#dd0000) for readability.\n- Keeps existing background and + glyph styling; no logic or rule changes.\n- Website-only CSS.\n\nCloses #5246.

### DIFF
--- a/packages/website/src/css/custom.css
+++ b/packages/website/src/css/custom.css
@@ -202,7 +202,7 @@ h6 {
 
 .code-block-added-line .token.string,
 .code-block-added-line .token.char {
-  color: #dd0000;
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Current contrast already passes WCAG AA. Using `#DD0000` (which is in itself a really glaring color) doesn't make anything better. You are also referencing a closed issue. If you think there's a real problem, please open a separate issue.

PR #12124, opened 2026-03-10
feat: support TypeScript 6

### LINKED ISSUE #12123: TypeScript 6 Support
## TypeScript 6.0 Support

[TypeScript 6.0 RC Announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0-rc/)

### 🔲 `lib.d.ts` Updates

### 🔲 Less Context-Sensitivity on `this`-less Functions

TypeScript 6.0 reduces context-sensitivity for functions that don't reference `this`. This may change type inference results in some edge cases.

### 🔲 Deprecated: `target: es5`

`target: "es5"` is deprecated and produces an error without `ignoreDeprecations: "6.0"`. We'll need to update tsconfig fixtures accordingly.

### 🔲 Deprecated: `no-default-lib` Directives

`/// <reference no-default-lib="true"/>` and `hasNoDefaultLib` are deprecated. Any code relying on `sourceFile.hasNoDefaultLib` to identify might need to be updated

### 🔲 Deprecated: `--moduleResolution node` (AKA `node10`)

`node10` module resolution is deprecated. Documentation should mention this. Might also break tests.

### 🔲 Deprecated: `asserts` Keyword on Imports

Import assertions (`assert {}`) are deprecated in favor of import attributes (`with {}`). Might break tests.

### 🔲 New Default Compiler Options

TS 6.0 changes many defaults: `strict: true`, `module: "esnext"`, `moduleResolution: "bundler"`, `target: "es2025"`, `types: []`, `rootDir: "."`, `esModuleInterop: true`. We'll need to update tsconfig fixtures accordingly.

### 🔲 JSX Closing Tag Tokenization Change

TypeScript 6.0 splits the `</` JSX closing tag token into `<` and `/` as separate tokens. AST snapshots in `ast-spec` need to be updated.
(This is not mentioned explicitly in the blog post.)

### Other changes with no impact to us

- Subpath Imports Starting with `#/`
- Combining `--moduleResolution bundler` with `--module commonjs`
- The `--stableTypeOrdering` Flag
- Deprecated: `--esModuleInterop false` and `--allowSyntheticDefaultImports false`
- Deprecated: `--baseUrl`
- Deprecated: `--downlevelIteration`
- Deprecated: `amd`, `umd`, and `systemjs` values of `module`
- Deprecated: `--moduleResolution classic`
- Deprecated: `--alwaysStrict false`
- Deprecated: `outFile`
- Deprecated: legacy `module` syntax for namespaces
- Specifying command-line files when `tsconfig.json` exists is now an error

___

_AI Disclamer: I was lazy, so this list has been generated with AI as a starting point, and edited manually by me_

### Additional Info

_No response_

StyleShit:
Closes #12123

~This is a starting point, some tests will probably fail~.

Used AI for most of the repetitive tasks.

It took me hours to understand why `isSourceFileDefaultLibrary` tests were failing...
~Turns out [it's a bug in TS 6](https://github.com/microsoft/typescript/issues/63228)~ not anymore 🎊

### DIFF
--- a/packages/parser/src/parser.ts
+++ b/packages/parser/src/parser.ts
@@ -12,7 +12,10 @@ import type { VisitorKeys } from '@typescript-eslint/visitor-keys';
 import type * as ts from 'typescript';
 
 import { analyze } from '@typescript-eslint/scope-manager';
-import { parseAndGenerateServices } from '@typescript-eslint/typescript-estree';
+import {
+  parseAndGenerateServices,
+  typescriptVersionIsAtLeast,
+} from '@typescript-eslint/typescript-estree';
 import { visitorKeys } from '@typescript-eslint/visitor-keys';
 import debug from 'debug';
 import { ScriptTarget } from 'typescript';
@@ -50,8 +53,11 @@ function getLib(compilerOptions: ts.CompilerOptions): Lib[] {
       .filter((lib): lib is Lib => !!lib);
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-deprecated -- Deprecated in TS 6 but we support TS < 6
-  const target = compilerOptions.target ?? ScriptTarget.ES5;
+  const defaultTarget = typescriptVersionIsAtLeast['6.0']
+    ? ScriptTarget.LatestStandard
+    : ScriptTarget.ES5; // eslint-disable-line @typescript-eslint/no-deprecated -- Deprecated in TS 6 but we support TS < 6
+
+  const target = compilerOptions.target ?? defaultTarget;
   // https://github.com/microsoft/TypeScript/blob/35ff23d4b0cc715691323ebe54f523c16fe6e3a5/src/compiler/utilitiesPublic.ts#L312-L346
   switch (target) {
     case ScriptTarget.ES2015:
--- a/packages/parser/tests/lib/parser.test.ts
+++ b/packages/parser/tests/lib/parser.test.ts
@@ -142,7 +142,6 @@ describe('parser', () => {
     ['es6', ScriptTarget.ES2015],
     // eslint-disable-next-line @typescript-eslint/no-deprecated -- Deprecated in TS 6 but we support TS < 6
     ['lib', ScriptTarget.ES5],
-    ['lib', undefined],
   ] as const)(
     'calls analyze() with `lib: [%s]` when the compiler options target is %s',
     ([lib, target], { expect }) => {
@@ -165,6 +164,54 @@ describe('parser', () => {
           },
         },
       } as typescriptESTree.ParseAndGenerateServicesResult<typescriptESTree.TSESTreeOptions>);
+
+      parseForESLint(code, config);
+
+      expect(spy).toHaveBeenCalledExactlyOnceWith(
+        expect.anything(),
+        expect.objectContaining({
+          lib: [lib],
+        }),
+      );
+    },
+  );
+
+  it.for([
+    ['lib', '5.9'],
+    [`es${new Date().getFullYear() - 1}.full`, '6.0'],
+  ] as const)(
+    'calls analyze() with `lib: [%s]` as the default when TS version is %s',
+    ([lib, tsVersion], { expect }) => {
+      const code = 'const valid = true;';
+      const spy = vi.spyOn(scopeManager, 'analyze');
+      const config: ParserOptions = {
+        filePath: 'isolated-file.src.ts',
+        project: 'tsconfig.json',
+        tsconfigRootDir: FIXTURES_DIR,
+      };
+
+      vi.spyOn(
+        typescriptESTree,
+        'parseAndGenerateServices',
+      ).mockReturnValueOnce({
+        ast: {},
+        services: {
+          program: {
+            // Explicitly not setting a target so that the default is used
+            getCompilerOptions: () => ({ target: undefined }),
+          },
+        },
+      } as typescriptESTree.ParseAndGenerateServicesResult<typescriptESTree.TSESTreeOptions>);
+
+      vi.spyOn(
+        typescriptESTree,
+        'typescriptVersionIsAtLeast',
+        'get',
+      ).mockReturnValue({
+        '5.9': false,
+        '6.0': false,
+        [tsVersion]: true,
+      } as Record<string, boolean>);
 
       parseForESLint(code, config);
 


packages/ast-spec/src/jsx/JSXAttribute/fixtures/element-non-self-closing/snapshots/2-TSESTree-Tokens.shot
```diff
@@ -81,11 +81,21 @@
...
+      end: { column: 18, line: 1 },
+    },
+  },
+  Punctuator {
+    type: "Punctuator",
+    value: "/",
+
+    range: [18, 19],
+    loc: {
+      start: { column: 18, line: 1 },
```
StyleShit, line 98: the jsx tokenization has changed in ts 6

packages/eslint-plugin-internal/tests/fixtures/tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
   "compilerOptions": {
     "jsx": "preserve",
-    "target": "es5",
+    "target": "es2015",
```
StyleShit, line 4: this is now the new minimum, es5 is deprecated
maybe we could remove some of these completely. I just didn't bother and ran a simple find & replace

packages/eslint-plugin/package.json
```diff
@@ -94,7 +94,7 @@
   "peerDependencies": {
     "@typescript-eslint/parser": "workspace:^",
     "eslint": "^8.57.0 || ^9.0.0 || ^10.0.0",
-    "typescript": ">=4.8.4 <6.0.0"
+    "typescript": ">=4.8.4 <7.0.0"
```
StyleShit: should be `<7.0.0` or `<6.1.0`?

packages/eslint-plugin/src/rules/no-base-to-string.ts
```diff
@@ -321,9 +322,10 @@ export default createRule<Options, MessageIds>({
         node.name.expression.expression.text === 'Symbol' &&
         ts.isIdentifier(node.name.expression.name) &&
         node.name.expression.name.text === 'toPrimitive' &&
-        checker
-          .getSymbolAtLocation(node.name.expression.expression)
-          ?.valueDeclaration?.getSourceFile().hasNoDefaultLib
```
StyleShit, line 326: this is deprecated in TS 6, so I replaced it with an explicit check against the default library

packages/eslint-plugin/tests/fixtures/tsconfig-with-dom.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "module": "commonjs",
```
StyleShit, line 4: same as the es5 thing - this is a simple find&replace because commonjs was the implicit default in TS < 6

we might not need this in all cases

packages/integration-tests/fixtures/eslint-v8/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
-    "strict": true
+    "strict": true,
```
StyleShit, line 3: technically we could remove `"strict": true` in all places since it's the new default, but many things will fail due to an [incompatibility of ts-api-utils with TS 6 ](https://github.com/JoshuaKGoldberg/ts-api-utils/issues/870) (mostly assumptions about default config values that have been changed)

I kept it with strict here just so ci will pass, but it's a blocker for this PR

packages/parser/tests/lib/parser.test.ts
```diff
@@ -122,13 +122,16 @@ describe('parser', () => {
...
+      lib: ['es6'],
       sourceType: 'script',
     });
   });
 
   it.for([
     ['esnext.full', ScriptTarget.ESNext],
+    ['es2025.full', ScriptTarget.ES2025],
+    ['es2024.full', ScriptTarget.ES2024],
+    ['es2023.full', ScriptTarget.ES2023],
```
StyleShit, line 134: 23-24 were missing so i added them as well
i assume it wasn't intentional?

packages/parser/tests/lib/parser.test.ts
```diff
@@ -137,8 +140,9 @@ describe('parser', () => {
     ['es2017.full', ScriptTarget.ES2017],
     ['es2016.full', ScriptTarget.ES2016],
     ['es6', ScriptTarget.ES2015],
+    // eslint-disable-next-line @typescript-eslint/no-deprecated -- Deprecated in TS 6 but we support TS < 6
     ['lib', ScriptTarget.ES5],
```
StyleShit, line 144: should we remove it completely?

packages/type-utils/src/builtinSymbolLikes.ts
```diff
@@ -183,7 +183,11 @@ export function isBuiltinSymbolLikeRecurser(
     )
   ) {
     const checker = program.getTypeChecker();
-    for (const baseType of checker.getBaseTypes(type as ts.InterfaceType)) {
+    const declaredType = checker.getDeclaredTypeOfSymbol(symbol);
+
+    for (const baseType of checker.getBaseTypes(
+      declaredType as ts.InterfaceType,
+    )) {
```
StyleShit, line 190: in TS 6, `getBaseTypes()` is slightly stricter than TS 5 when it comes to constructor types

This, for example, will be broken in `unbound-method` when using TS 6:
```TS
class Foo extends Array {}

const x = Foo.isArray;
```

Before the fix, `getBaseTypes()` returned an empty array instead of `[Array]`

I noticed we don't have any tests for this function, so I added some tests (with help of AI, but manually checked and refined)

We can separate this into its own PR if needed

patches/typescript-6.patch
```diff
@@ -0,0 +1,22 @@
+# Temporary patch. See https://github.com/microsoft/TypeScript/issues/63228
```
StyleShit: There is a [PR that fixes this](https://github.com/microsoft/TypeScript/pull/63239) already
Waiting for it to be released so we can remove this

packages/parser/tests/lib/parser.test.ts
jakebailey: I wouldn't bother, we will keep all the declarations (though I am confused by `lib` here, lib=es5 is actually still supported).

StyleShit: Not sure that I understand what you mean

It's about the change from lib to es6?

jakebailey: Er, well the default lib is inferred from target, and that now defaults to `LatestStandard`, which is ES2025.

StyleShit: I'm sorry, idk if it's the language barrier here (I'm not a native speaker), or it's just me being silly, but I still don't get what you mean 😅

packages/eslint-plugin/package.json
kirkwaiblinger: We've normally set it as `<6.1.0`, since each minor version may require us to make changes.

packages/integration-tests/fixtures/eslint-v8/tsconfig.json
kirkwaiblinger: FYI ts-api-utils v2.5.0 is out now and contains this fix (https://github.com/JoshuaKGoldberg/ts-api-utils/pull/873), so you should be good to update the ts-api-utils version.

As for the old eslint v8 fixtures... might as well keep these those as they are, I think. Doesn't hurt for them to be explicit.

packages/eslint-plugin/package.json
StyleShit: Done

packages/integration-tests/fixtures/eslint-v8/tsconfig.json
StyleShit: Done, thanks!

patches/typescript-6.patch
stevelandeydescript: Looks like 6.0.2 went out earlier today. (Thanks for your work on this, by the way!)

packages/parser/tests/lib/parser.test.ts
StyleShit: reverted

bradzacher (dismissed):
This is all looking good to me -- thanks for doing this for us!

packages/parser/src/parser.ts
```diff
@@ -50,8 +50,9 @@ function getLib(compilerOptions: ts.CompilerOptions): Lib[] {
       .filter((lib): lib is Lib => !!lib);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-deprecated -- Deprecated in TS 6 but we support TS < 6
   const target = compilerOptions.target ?? ScriptTarget.ES5;
```
kirkwaiblinger: Should we also change the default here if the detected TS version is >= 6?

(Similar to https://github.com/JoshuaKGoldberg/ts-api-utils/blob/f9f5042a7541a94d0569cb18a51645c5522767b3/src/compilerOptions.ts#L148)

StyleShit: Ah, I wondered the same
I did change it initially, but got somewhat confused by https://github.com/typescript-eslint/typescript-eslint/pull/12124#discussion_r2925976461 and reverted

kirkwaiblinger: I'm not super familiar with this area of the code but it looks like this is used to populate globals for scope analysis, therefore, I'm guessing there'd be an observable difference in lint runs that attempt to reference any globals that are defined by default with TS 6 (ES2015) but not with ES 5 (not sure off the top of my head whether what would be in this set).

Seems to me that putting 
```ts
const target = compilerOptions.target ?? (tsVersionAtLeast(6) ? ScriptTarget.ES2015 : ScriptTarget.ES5);
```
would be the right thing to do. That said, it's probably something only encountered as a pretty niche edge case, though.

jakebailey: The default target in TS 6 is acutally "the current year" so ES2025.

jakebailey: Well, so actually, you probably want `ScriptTarget.LatestStandard`, if 6+.

StyleShit: Done (b91ab84e9b9674f2b3bb048f67ae7af4c83e6321)
I don't really like the `new Date()` thing I did there... Open for suggestions

Thanks @jakebailey for the clarification

bradzacher:
Let's get this in so it's in for the next release.
We can always tweak things in tests later if we want.

bradzacher:
thanks so much for getting this ready for us!

fabb:
has been released in [v8.58.0](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v8.58.0)

PR #12136, opened 2026-03-17
chore: add missing build dependency in website-eslint

StyleShit:
~This is a tiny change in the series of #11204
Removed the build process of eslint-plugin-internal, and as a side-effect converted the eslint config file to TS~

Also found a missing build dependency in website-eslint that breaks tasks like `typecheck` (and `website:build`) after a `pnpm clean` run

~It might not save much time, but I guess that bunch of small, quick wins will add up eventually? 🤷~

### DIFF
--- a/packages/website-eslint/package.json
+++ b/packages/website-eslint/package.json
@@ -55,6 +55,7 @@
       "build": {
         "command": "tsx build.mts",
         "dependsOn": [
+          "^build",
           "types:copy-ast-spec"
         ],
         "cache": true,


eslint.config.mts
kirkwaiblinger, line 1: Drive-by comment: I'm personally not a fan of the way eslint uses jiti for the eslint config, and I'd prefer we keep it as a an ordinary JS file until we can use node's native TS loading (I think this requires node version >= 20, and, specifying an experimental flag with eslint, see, eg https://github.com/eslint/eslint/issues/19985 🙁)

StyleShit: i agree
but we're using node 24 in CI (since yesterday basically), so we can load TS files natively without any flag 😄

bradzacher (commented):
could we give the same treatment to `rule-schema-to-typescript-types`?
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/rule-schema-to-typescript-types/package.json

also do you have a rough idea about how much this improves?

also given that type stripping is done by SWC in node -- if we were to, say, migrate our "Build" step to use SWC for the TS -> JS conversion -- wouldn't this change be moot, or rather even a net-negative as it would strip on-the-fly each lint run instead of using a "cached" JS build?

StyleShit:
> how much this improves?

This i _very_ neglibigle (~[2s](https://github.com/typescript-eslint/typescript-eslint/actions/runs/22831767823/job/66221385288#step:4:182)), it was mostly to remove a redundant task 

> wouldn't this change be moot, or rather even a net-negative

That's actually a great question...
TBH, I'm not familiar enough with how Node uses SWC, and whether the caching will be shared, so I don't have a good answer here
But that's genuinely a good question that makes me question this PR 😅 

> could we give the same treatment to rule-schema-to-typescript-types?

If we decide to go forward with this change, sure

eslint.config.mts
kirkwaiblinger: i think that's using jiti though?

StyleShit: huh... you're right!

so maybe that's not the best idea
(unless we're fine with using jiti or adding the flag)

JoshuaKGoldberg:
@StyleShit what is the conclusion here? Sounds like just fixing the missing website dependency is what we have consensus on?

StyleShit:
> @StyleShit what is the conclusion here? Sounds like just fixing the missing website dependency is what we have consensus on?

Yes, I've pushed this specific fix and updated the PR title/description

JoshuaKGoldberg (approved):
Thanks! 🙏

PR #12139, opened 2026-03-18
chore: typecheck using tsgo

StyleShit:
This PR adds typecheck using tsgo in the CI

~~Next one in the series of #11204~~

~~Replacing tsc with tsgo for typecheck~~

~~As always, benchmarks from my machine (Mac M4 Pro, 24GB), about 34% faster:~~

~~`$ hyperfine "CI=true pnpm typecheck" "CI=true pnpm typecheck-fast"`~~

~~`Benchmark 1: CI=true pnpm typecheck`~~
~~`  Time (mean ± σ):     25.511 s ±  0.442 s    [User: 27.868 s, System: 6.621 s]`~~
~~`  Range (min … max):   24.420 s … 26.009 s    10 runs`~~

~~`Benchmark 2: CI=true pnpm typecheck-fast`~~
~~`  Time (mean ± σ):     19.078 s ±  0.424 s    [User: 25.284 s, System: 5.371 s]`~~
~~`  Range (min … max):   18.591 s … 19.908 s    10 runs`~~

~~`Summary`~~
~~`  CI=true pnpm typecheck-fast ran`~~
~~`    1.34 ± 0.04 times faster than CI=true pnpm typecheck`~~

~~I ensured the typecheck tasks didn't have nx cache or tsbuildinfo cache in both benchmarks, but _did_ have build cache (as we have in CI when running typecheck)~~

~~It's kinda hard to understand how much has been improved in CI due to all the caching. Every commit I push has different timings~~


.vscode/settings.json
```diff
@@ -21,4 +21,6 @@
   "typescript.preferences.quoteStyle": "single",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "nxConsole.generateAiAgentRules": true,
+
```
JoshuaKGoldberg: Nitty nit: no need

```suggestion
```

packages/eslint-plugin/src/rules/naming-convention.ts
```diff
@@ -716,7 +716,8 @@ export default createRule<Options, MessageIds>({
         return [
           selector,
           (node: Parameters<typeof handler>[0]): void => {
-            handler(node, validator);
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+            handler(node as never, validator);
```
JoshuaKGoldberg, line 720: [Non-Actionable] ... I hate this rule so much 😂 

```plaintext
src/rules/naming-convention.ts:719:21 - error TS2345: Argument of type 'ArrayExpression | ArrayPattern | ArrowFunctionExpression | AssignmentExpression | AssignmentPattern | ... 163 more ... | VariableDeclarator' is not assignable to parameter of type 'never'.
  Type 'AccessorPropertyComputedName' is not assignable to type 'never'.

719             handler(node, validator);
                            ~~~~
```

I can't tell if this is a good change on tsgo's part or not. It's always irked me that the generic types / function overloads don't flow through to make this work.

JoshuaKGoldberg (approved):
LGTM, thanks! ✨

tools/scripts/generate-sponsors.mts
```diff
@@ -11,6 +11,7 @@ const excludedNames = new Set([
 
 const filteredTerms = ['casino', 'deepnude', 'tiktok'];
 
+// @ts-expect-error - https://github.com/microsoft/typescript-go/issues/3140
```
kirkwaiblinger: (optional) let's just switch these to a named function to avoid this altogether?

kirkwaiblinger (commented):
I like the idea of using and previewing TS 7 in general, but I think it has a bit of a drawback for this project... Since typescript-eslint uses the TS API, linting on the typescript-eslint repo itself will depend on TS version will be tied to TS 6, which isn't identical to TS 7, and therefore we'll potentially run into the classic "typechecking with a different version of TS than you see in your editor" problem. IMO that's a bigger problem for us, being the linter ourselves, than for any old repo. 

Let's be sure we at least have a task for typechecking with TS 6 to fall back on?

tools/scripts/generate-sponsors.mts
StyleShit: well, since https://github.com/microsoft/typescript-go/issues/3140 has been fixed already, I guess we could remove it and have zero diff here

StyleShit:
> "typechecking with a different version of TS than you see in your editor"

I totally agree, but isn't part of TS 6's existence is to have a "transitional version" between TS 5.9 and TS 7?
As in "TS 7 should be identical to TS 6, except for being faster"

At least this is what I inferred from both [TS 6](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#preparing-for-typescript-7.0) & [TS 7](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0-beta/#updates-since-5.x-and-new-behaviors-from-6.0) announcements:

> TypeScript 7.0 is made to be compatible with TypeScript 6.0’s type-checking and command-line behavior. Any TypeScript code that compiles cleanly with TypeScript 6.0 (with the stableTypeOrdering flag on, and without the ignoreDeprecations flag set) should compile identically in TypeScript 7.0.

And if that's the case, maybe we'll actually be the ones to help improving tsgo's parity with TS 6 (🤓)

(Well, there is technically a small difference, as we saw [here](https://github.com/typescript-eslint/typescript-eslint/pull/12139#discussion_r3253521288), which I'm not sure whether is bug or intentional, but maybe @jakebailey would be able to shed some light here)

> ... typechecking with TS 6 to fall back on

If we'll go that path, I'm not sure this PR is relevant, since the whole point is to improve the CI performance (#11204), and adding another layer of typechecking will just make things slower

packages/eslint-plugin/src/rules/naming-convention.ts
jakebailey: I don't know what broke here, but I would probably suggest reporting it upstream if it doesn't repro with TS 6.0 + `--stableTypeOrdering`?

tools/scripts/generate-sponsors.mts
kirkwaiblinger: Ah, even better!

JoshuaKGoldberg (changes requested):
> As in "TS 7 should be identical to TS 6, except for being faster"

From a user's perspective, yes. From the API perspective, not so because of the different languages & runtimes.

Kirk brings up a good point. I agree that including TS 6 is necessary given how we're a big part of the standard usage of TS 6 APIs.

I do think there's value in including TS 7 type-checking:

* It'd be good to ensure we _also_ type-check nicely with TS 7, since there are some differences
* It'll get us faster CI results for specifically type-checking

StyleShit:
I understand the concern but I'm not sure that I follow

What does typechecking of our internal code has to do with the TS program we're initializing for type-checked rules in runtime?
That programs would still be TS 6, and we'll still use the TS 6 APIs. Using tsgo doesn't change that

Unless you're talking about dogfooding and ensuring the codebase is being type-linted properly in both TS 6 _and_ TS 7?
In that case, I guess having both TS 6 & 7 typechecks could make sense.
But if I understand correctly, having different typecheck results in TS 7 basically means a bug, no? Isn't it the same as typechecking with all supported TS versions in case one of them introduced a bug?

Am I missing something here?

Note: I might not have the full story of how type-checked rules work under the hood, if that's related

JoshuaKGoldberg:
> Unless you're talking about dogfooding and ensuring the codebase is being type-linted properly in both TS 6 and TS 7?
> In that case, I guess having both TS 6 & 7 typechecks could make sense.

Yes, exactly, that.

> But if I understand correctly, having different typecheck results in TS 7 basically means a bug, no? Isn't it the same as typechecking with all supported TS versions in case one of them introduced a bug?

Not necessarily. It's possible that our code relies on a specific behavior in TS 6 that's changed in TS 7 to newly report a type error on our code. That could be from a bug in TS 6, or a bug in TS 7, or bugs in both, or just some random nuance that happens to be treated differently in the new code. So really (0 or 1) bug in each of (TS 6 and TS 7).

At the end of the day, we just need to ensure that we type-check properly for our consumers in both TS versions.

tools/scripts/generate-sponsors.mts
StyleShit: Removed

StyleShit:
OK, gotcha, makes sense
Added tsgo as an additional check

packages/eslint-plugin/src/rules/naming-convention.ts
StyleShit: It does reproduce with `--stableTypeOrdering`, so I guess this is a TS < 7 bug? 🤷

.vscode/settings.json
StyleShit: Removed

packages/eslint-plugin/src/rules/naming-convention.ts
JoshuaKGoldberg: Do you have bandwidth to report that bug @StyleShit? It'd be nice to have a link in this comment as an explainer.

JoshuaKGoldberg (dismissed):
LGTM! I think we can merge without waiting for an upstream issue or comment, but it would be nice.

packages/eslint-plugin/src/rules/naming-convention.ts
StyleShit: By "bug" I mean "a known wrong behavior that existed before TS 7 and was fixed in TS 7 and in TS 6 with stableTypeOrdering"

So no issue is required I guess

StyleShit:
Sorry for the last pushes. I thought something was wrong with my PR, but it turns out it was due to #12369 and #12368

JoshuaKGoldberg:
Is this ready to go now?

StyleShit:
If CI passes, yes ⚡️

PR #12144, opened 2026-03-19
fix(eslint-plugin): [no-unnecessary-condition] fix false positive with template literal types

rami0617:
## PR Checklist

- [v] Addresses an existing open issue: fixes []
- [v] That issue was marked as [accepting prs](https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aopen+is%3Aissue+label%3A%22accepting+prs%22)
- [v] Steps in [Contributing](https://typescript-eslint.io/contributing) were taken

## Overview
The current rule produces a false positive when comparing a template literal type with a string literal, as it fails to account for potential pattern matches. To resolve this, I have updated the logic to relax the check when a template literal type is involved in the comparison.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-condition.ts
@@ -63,6 +63,13 @@ function toStaticValue(
   if (type.flags === ts.TypeFlags.Null) {
     return { value: null };
   }
+
+  // Template literal types like `a${string}` represent multiple possible values,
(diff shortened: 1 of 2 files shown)


rami0617:
> LGTM
> 
> But I _think_ ths attached issue is wrong?

My apologies. I completely mixed up this issue with a different one regarding no-restricted-imports. I realized my changes for template literal types are unrelated to the current issue scope. Closing this PR to avoid confusion. Sorry for the noise!

PR #12147, opened 2026-03-20
fix(eslint-plugin): [no-unnecessary-condition] use assignability checks in checkTypePredicates

### LINKED ISSUE #11798: Enhancement: [no-unnecessary-condition] use assignability checks in `checkTypePredicates`

### Description

Currently the rule only flags an asserted argument if the type of the variable is _equal_ to the asserted type. I propose to change the logic so that the rule also flags if it's a subtype. 

### Fail

```TypeScript
declare function isStringOrNumber(x: unknown): x is string | number;

declare const s: string;

if (isStringOrNumber(s)) {
}
```

### Pass

```TypeScript
declare function isStringOrNumber(x: unknown): x is string | number;

declare const s: string | number | bigint;

if (isStringOrNumber(s)) {
}
```

### Additional Info

This is split out from https://github.com/typescript-eslint/typescript-eslint/issues/11716#issuecomment-3439853329.

I've begun working on this and there is a bit of nuance beyond simply using `checker.isAssignableTo()` and calling it a day... Examples like the following are troublesome:

```ts
interface Wider {
  a: string;
}

interface Narrower {
  a: string;
  b?: number;
}

declare function isNarrower(x: unknown): x is Narrower;

declare const n: Narrower;
declare const w: Wider;

// These types are mutually assignable:
n satisfies Wider;
w satisfies Narrower;

// Nominally, `isNarrower(w)` seems to be unnecessary, since `w` is 
// already assignable to `Narrower`.
//
// But, using the `isNarrower()` narrowing has a real impact: it allows
// the optional methods in `Narrower` to be invoked on `w`.
if (isNarrower(w)) {
  w.b?.toFixed(); 
}
```

"is assignable to" and "is a subtype of" are not quite same thing in TS, so avoiding false positives on these may require some thought.

cc @ronami

zaewc:
## PR Checklist

- [X] Addresses an existing open issue: fixes #11798

## Overview

  `checkTypePredicates` previously only flagged a type guard call as unnecessary when the argument's   
  type was the exact same TS type object as the predicate's asserted type. This meant cases
  where the argument is a strict subtype of the asserted type were silently passed over.               
                                                            
  e.g, `isStringOrNumber(s)` where `s: string` was not flagged, even though `string` is always 
  assignable to `string | number`, making the check redundant.
                                                                                                       
  This pr extends the check to also report when the argument type is a strict subtype of the asserted  
  type — i.e., assignable one-way but not mutually assignable. Mutual assignability is explicitly
  excluded to avoid false positives with structurally compatible types that still produce meaningful   
  narrowing.


bradzacher (changes requested):
looking good so far

```ts
interface Wider {
  a: string;
}
interface Narrower {
  a: string;
  b?: number;
}
declare function isWider(x: unknown): x is Wider;
declare const n: Narrower;
if (isWider(n)) {
}
```

Could you add this as a test case please.
It _should_ error IIUC

zaewc:
@bradzacher 

The rule wasn't catching this because it compared types with `===`, so structurally compatible but distinct interfaces were missed. Fixed by switching to `checker.isTypeAssignableTo()`. This also uncovered that string literal -> string type guard cases (e.g. `isString('falafel')`) were incorrectly allowed, moved those to `invalid` as well.

bradzacher (approved):
thanks!

JoshuaKGoldberg (approved):
🆒 !

ArnaudBarre:
I think the last test case should be allowed, it's useful to have valid access to the `b` property:

https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgOrACbWQbwFDLJwBcyAzmFKAOYDceAvnqJLIigHJxRQD2A7tnyES5SjXqEARgH5SIAK4BbKdHpMsCADbcUMBSARhgvEMmBkuPAdAAUAD1IGA1iAEgAlKXvmyyK3yCUPSaOlAoCKYUyPyk6FjBePwAdFLMMMi2FgE2ULb8Hh64BDGpjHhAA

bradzacher:
@ArnaudBarre you don't need a predicate at all in the last test case because `b` is an optional property. So you don't need a runtime assertion at all.

[Eg `const n: Narrower = w` is valid TS code when the "new" property is marked as optional.](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgOrACbWQbwFDLJwBcyAzmFKAOYDceAvnqJLIigHJxRQD2A7tnyES5SjXqEARgH5SIAK4BbKdHpMsCADbcUMBSARhgvEMmBkuPAdAAUAD1IGA1iAEgAlKXvmyyK3yCUPSaOlAoCKYUyPyk6FjBeJEg0SCkATZQyAC8MfQgAHRSeEA)

If the property is non-optional then the check is required and the rule does not report.

ArnaudBarre:
Oh yeah true, that also explain the edge case (even if I think it would be better implemented by actually checking the the narrower type has at least one required property)

The fact that you need to create a new variable in was not be obvious to me and could be added in the message

PR #12164, opened 2026-03-24
fix(eslint-plugin): avoid crashing no-unnecessary-type-arguments fixer

### LINKED ISSUE #12157: Bug: [no-unnecessary-type-arguments] Exception thrown: Cannot read properties of undefined (reading 'range')

### Playground Link

https://typescript-eslint.io/play/#ts=5.8.3&fileType=.tsx&code=MYGwhgzhAEBiD28A8AVAfNA3gWAFDWgAcBXAIxAEthph4A7CAFwCdjhH5mAKMALmjB0AngBpopfoNE1JwsQBNZQgJRZoAXzybcFOowCmzAGZhg%2B6ACEwzLHgIA3MCCUBuLXhD7G0I4mgBeaDp9AHc4RCQrZjQuAAYxeOhEzGhHZySNZRcgA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWiIDmsALaIWqDJETRoHaJHBgAviF1A&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
class Foo<T> {
    public constructor(a: any, b: any, c: any, d: any) {}
}
interface Bar {
    val: any;
}
let foo = new Foo<Bar>(0, 0, 0, { val: 0 });
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-unnecessary-type-arguments": "error",
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    // ...
  }
}
```

### Expected Result

No exception

### Actual Result

Exception thrown

### Additional Info

This seems to have been introduced in the (brand new) [v8.57.2](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v8.57.2). I do not see this with 8.57.1.

teee32:
## Problem

`@typescript-eslint/no-unnecessary-type-arguments` can throw while building fixes for constructor calls with a single explicit type argument and multiple runtime arguments.

## Reproduction / Observation

This reproduces on current `main` with code such as:

```ts
class Foo<T> {
  public constructor(a: any, b: any, c: any, d: any) {}
}
interface Bar {
  val: any;
}
const foo = new Foo<Bar>(0, 0, 0, { val: 0 });
```

The rule reports `canBeInferred`, then crashes while normalizing fixes with:

```text
TypeError: Cannot read properties of undefined (reading 'range')
```

## Root cause

`checkTSArgsAndParameters()` stores the last type-argument index in a local `i`, but the `parent.arguments.forEach((argument, i) => ...)` callback shadows that variable.

The fixer closure then reads the argument index instead of the type-argument index when computing the removal range. For calls with more runtime arguments than type arguments, that can access a non-existent `typeArguments.params[...]` entry.

## Change

- keep the type-argument index in a dedicated `typeArgumentIndex`
- use a separate `argumentIndex` for iterating call/new arguments
- add a regression test covering a constructor call that previously crashed

## Why this fix

This keeps the existing rule semantics and fixer behavior intact. The change only removes the accidental index shadowing that made the fixer read the wrong array slot.

## Risk

Low. The change is localized to fixer range calculation in one rule and is covered by an explicit regression test.

## Validation

- `pnpm vitest run packages/eslint-plugin/tests/rules/no-unnecessary-type-arguments.test.ts --config vitest.config.mts`
- `pnpm eslint packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts packages/eslint-plugin/tests/rules/no-unnecessary-type-arguments.test.ts`

Fixes #12157.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
@@ -81,9 +81,9 @@ export default createRule<[], MessageIds>({
       tsNode: ParameterCapableTSNode,
     ): void {
       // Just check the last one. Must specify previous type parameters if the last one is specified.
-      const i = typeArguments.params.length - 1;
-      const typeArgument = typeArguments.params[i];
-      const typeParameter = typeParameters.at(i);
+      const typeArgumentIndex = typeArguments.params.length - 1;
+      const typeArgument = typeArguments.params[typeArgumentIndex];
+      const typeParameter = typeParameters.at(typeArgumentIndex);
 
       if (!typeParameter) {
         return;
@@ -101,8 +101,8 @@ export default createRule<[], MessageIds>({
           tsNode as ts.CallExpression | ts.NewExpression,
         );
 
-        parent.arguments.forEach((argument, i) => {
-          const parameter = sig?.parameters.at(i);
+        parent.arguments.forEach((argument, argumentIndex) => {
+          const parameter = sig?.parameters.at(argumentIndex);
 
           if (
             !parameter?.valueDeclaration ||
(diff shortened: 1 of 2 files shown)


kirkwaiblinger:
https://typescript-eslint.io/contributing/ai-policy/

PR #12182, opened 2026-03-31
fix(eslint-plugin): [no-shadow] correct rule to match ESLint v10 handling

### LINKED ISSUE #12070: Bug: [no-shadow] Difference in handling from ESLint v10 no-shadow rule

### Playground Link

https://typescript-eslint.io/play/#ts=5.9.2&fileType=.tsx&code=JYWwDg9gTgLgBAbziApiCcC%2BcBmUIhwDkUKAhgMYxEDcAUHRRAHYDO8AYhBAEJlQBhApGYpm8ALzI0EABQ4ArsyrAWcLr35DwLMTFkBKRHThxSMBVGZwAPABNgANwB8GmwHoHL%2BpgM0gA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1uQAsBDACYcA7uiiJo0DtEjgwAXxDygA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

### Repro Code

```TypeScript
import { memo } from 'react';

const FooBarComponent = memo(function FooBarComponent() {
  return <div>Foo</div>;
});
```

### ESLint Config

```javascript
module.exports = {
  rules: {
    "@typescript-eslint/no-shadow": ["error"],
  },
};
```

### tsconfig

```jsonc
{
  "compilerOptions": {
    "strictNullChecks": true
  }
}
```

### Expected Result

I expect that the typescript-eslint version of no-shadow would match the behavior of the core ESLint rule it's extending here. 

### Actual Result

The behavior is different in ESLint vs TypeScript-ESLint now, as of ESLint v10. (the current tseslint behavior matches v9, to be fair). And this is not a difference which is caused by type awareness or an improvement over the ESLint rule necessarily.

### Additional Info

This was discovered in https://github.com/oxc-project/oxc/issues/19461

See [the ESLint playground where a violation is raised](https://eslint.org/play/#eyJ0ZXh0IjoiaW1wb3J0IHsgbWVtbyB9IGZyb20gJ3JlYWN0JztcblxuY29uc3QgRm9vQmFyQ29tcG9uZW50ID0gbWVtbyhmdW5jdGlvbiBGb29CYXJDb21wb25lbnQoKSB7XG4gIHJldHVybiA8ZGl2PkZvbzwvZGl2Pjtcbn0pOyIsIm9wdGlvbnMiOnsicnVsZXMiOnsibm8tc2hhZG93IjpbImVycm9yIl19LCJsYW5ndWFnZU9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hRmVhdHVyZXMiOnsianN4Ijp0cnVlfSwic291cmNlVHlwZSI6Im1vZHVsZSJ9LCJlY21hVmVyc2lvbiI6MjAyNiwicGFyc2VyIjoiQHR5cGVzY3JpcHQtZXNsaW50L3BhcnNlciJ9fX0=) for comparison with ESLint.

It looks like ESLint changed the behavior of this rule for v10 here: https://github.com/eslint/eslint/commit/43677de07ebd6e14bfac40a46ad749ba783c45f2

nevette-bailey:
## PR Checklist

- [x] Addresses an existing open issue: fixes #12070 

## Overview

The `no-shadow` rule reports when a variable in an inner scope "shadows" (reuses the same name and hides) a variable in an outer scope. There is a deliberate exception when named function/class expressions are assigned directly to a variable of the same name that is allowed, common, and a harmless self-reference:

```typescript
var a = function a() {};
```

In this case, the function's inner name only exists inside of the function itself, so reporting it as a shadow would be noisy and unhelpful.

### The Bug

The `isOnInitializer` helper previously used a range-containment check to decide if this exception should be applied, by asking "is the inner variable located _somewhere inside_ the declaration of the outer variable?" This definition was too loose because it suppressed shadow reports for named expressions _anywhere_ inside an initializer, not _just_ direct initializers, such as this case that should be reported because the inner definition is a genuinely separate binding that shadows the outer one:

```typescript
const FooBarComponent = memo(function FooBarComponent() {});
```

ESLint v10 started to correctly report these cases.

### The Fix

Instead of asking "is the inner name _somewhere inside_ the initializer?", this PR rewrites to ask "is the inner expression _literally_ the initializer?" It does this by:

1. Rewriting `isOnInitializer` to walk from the outer variable to its actual initializer node
2. Checking whether the inner function/class expression **is** that node through an identity check and not just range-containment

The `unwrapExpression` helper handles these still-valid cases where the function expression can transparently become `a` at runtime:

```typescript
var a = foo || function a() {};
var a = foo ? function a() {} : bar;
```

The previous code also had a check for class expressions that verified the node type but not the definition type, even though a definition bundles both. The fix adds a missing `DefinitionType.ClassName` check for class expressions so that both `innerDef.node.type` and `innerDef.type` are now checked.

Fixes #12070.

💖

### DIFF
--- a/packages/eslint-plugin/src/rules/no-shadow.ts
+++ b/packages/eslint-plugin/src/rules/no-shadow.ts
@@ -462,35 +462,84 @@ export default createRule<Options, MessageIds>({
     }
 
     /**
-     * Checks if a variable is inside the initializer of scopeVar.
+     * Finds the uppermost expression node that can evaluate to the given one,
+     * unwrapping through LogicalExpression and non-test ConditionalExpression branches.
+     * @param node The node to unwrap.
+     * @returns The topmost unwrapped node.
+     */
+    function unwrapExpression(node: TSESTree.Node): TSESTree.Node {
+      const { parent } = node;
+      if (
+        parent?.type === AST_NODE_TYPES.LogicalExpression ||
+        (parent?.type === AST_NODE_TYPES.ConditionalExpression &&
+          parent.test !== node)
+      ) {
+        return unwrapExpression(parent);
+      }
+      return node;
+    }
+
+    /**
+     * Checks if a variable is the name of a function or class expression that is
+     * directly assigned (or transparently through `||`/`?:`) as the initializer
+     * of scopeVar.
      *
-     * To avoid reporting at declarations such as `var a = function a() {};`.
-     * But it should report `var a = function(a) {};` or `var a = function() { function a() {} };`.
+     * Allows `var a = function a() {}` but reports `var a = wrap(function a() {})`.
      * @param variable The variable to check.
      * @param scopeVar The scope variable to look for.
-     * @returns Whether or not the variable is inside initializer of scopeVar.
+     * @returns Whether or not the variable is the direct initializer name of scopeVar.
      */
     function isOnInitializer(
       variable: TSESLint.Scope.Variable,
       scopeVar: TSESLint.Scope.Variable,
     ): boolean {
-      const outerScope = scopeVar.scope;
       const outerDef = scopeVar.defs.at(0);
-      const outer = outerDef?.parent?.range;
-      const innerScope = variable.scope;
       const innerDef = variable.defs.at(0);
-      const inner = innerDef?.name.range;
 
-      return !!(
-        outer &&
-        inner &&
-        outer[0] < inner[0] &&
-        inner[1] < outer[1] &&
-        ((innerDef.type === DefinitionType.FunctionName &&
-          innerDef.node.type === AST_NODE_TYPES.FunctionExpression) ||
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (approved):
Super, thanks for this! 🙌

jraoult:
@JoshuaKGoldberg @nevette-bailey Hi, assuming we liked the previous behaviour (`const FooBarComponent = memo(function FooBarComponent() {});` considered valid), can we restore it? Is there an option for that?

bradzacher:
https://typescript-eslint.io/contributing/issues#commenting

Do not comment on closed PRs.

PR #12192, opened 2026-04-01
docs(eslint-plugin): clarify no-redeclare type-value collision not covered by TypeScript

### LINKED ISSUE #11461: Docs: Not all reported problems of `no-redeclare` are covered by TypeScript

### Suggested Changes

Adjust the "Danger" banner of the `no-redeclare` lint rule that currently states:

> The code problem checked by this ESLint rule is automatically checked by the TypeScript compiler. Thus, it is not recommended to turn on this rule in new TypeScript projects. You only need to enable this rule if you prefer the ESLint error messages over the TypeScript compiler error messages.

That's incorrect, because this rule does report naming collisions between a const and a type, while TypeScript doesn't. This rule uncovered an accidental issue of an incorrect name in our code we wouldn't have discovered with TypeScript. See reproduction link below.

I'm not sure what a better message would be or if that means the banner needs to removed from the docs page.

### Affected URL(s)

https://typescript-eslint.io/rules/no-redeclare/

### Additional Info

https://typescript-eslint.io/play/#ts=5.9.2&fileType=.tsx&code=JYWwDg9gTgLgBALzgMyhEcBECIBNMDcAUEQKYAekscAxhAHYDO8AYhBAMo0AWpIAhnAC8iAHQQARgCtSNGAAoA3nDBowpWAE8AXGOZRg9AObyAlHAC%2Bp4mUrR4MTerhtOPPoJEJRh5BoA8juoQyC7sXLwCAHzEQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQHYHsBaaRAE0QGN4BDE9KfAM0cgBpxtIABAFwE8ADigrQAlgJ6EU8Ubh4B6AsTKUadDJETRo%2BaJA4BfEAaA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

KillerBean:
## PR Checklist

- [x] Addresses an existing open issue: Closes #11461

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-redeclare.mdx
+++ b/packages/eslint-plugin/docs/rules/no-redeclare.mdx
@@ -15,8 +15,8 @@ import Admonition from '@theme/Admonition';
   <p>
     Some cases checked by this rule are also checked by the TypeScript compiler
     (variable redeclarations, function overloads, and declaration merging).
-    However, this rule additionally reports collisions between value declarations
-    (e.g., <code>const Foo</code>) and type declarations (e.g.,{' '}
+    However, this rule additionally reports collisions between value
+    declarations (e.g., <code>const Foo</code>) and type declarations (e.g.,{' '}
     <code>type Foo</code> or <code>interface Foo</code>), which TypeScript does{' '}
     <strong>not</strong> check.
   </p>
--- a/packages/eslint-plugin/docs/rules/no-redeclare.mdx
+++ b/packages/eslint-plugin/docs/rules/no-redeclare.mdx
@@ -14,13 +14,7 @@ import TypeScriptOverlap from '@site/src/components/TypeScriptOverlap';
 <TypeScriptOverlap />
 
 It adds support for TypeScript function overloads and declaration merging.
-
-:::note
-The overlap with TypeScript applies only to true variable redeclarations.
-This rule also reports naming collisions between values (e.g., `const Foo`) and types
-(e.g., `type Foo` or `interface Foo`), which TypeScript does **not** check.
-This is intentional — such collisions are often accidental and can lead to confusing code.
-:::
+Unlike TypeScript, this rule also reports collisions between value and type declarations with the same name (e.g., `const Foo` and `type Foo`).
 
 ## Options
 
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (changes requested):
Change request: please rephrase the banner as suggested in the issue's discussion. 🙂

packages/eslint-plugin/docs/rules/no-redeclare.mdx
```diff
@@ -9,11 +9,22 @@ import TabItem from '@theme/TabItem';
...
 > See **https://typescript-eslint.io/rules/no-redeclare** for documentation.
 
-import TypeScriptOverlap from '@site/src/components/TypeScriptOverlap';
-
-<TypeScriptOverlap />
-
-It adds support for TypeScript function overloads, and declaration merging.
+import Admonition from '@theme/Admonition';
+
+<Admonition type="danger">
```
Josh-Cena: Use Markdown.

Also I wonder if we can make the `TypeScriptOverlap` component take additional content like what you are adding here, so we don't deviate from the template too much.

KillerBean: Thanks for the feedback! I've updated the file to use Markdown instead of JSX.

KillerBean: Regarding extending TypeScriptOverlap to accept additional content — I think that's a great idea, but it would affect other rules that use the component. To keep this PR focused, would you prefer I open a separate PR for that change?

Josh-Cena:
That is not what I asked for. I'm asking for `<TypeScriptOverlap context={...} />` or something similar.

KillerBean:
Hi @Josh-Cena ! I'd like to pick this up in a new PR since the requested changes touch both the component and the doc file.                                                                                                                            
Before I proceed, a quick question: should the context prop be typed as React.ReactNode or string? The clarification text includes inline code (e.g., `const Foo`, `type Foo`), so ReactNode would let the caller pass JSX directly in MDX — but wanted to confirm that's the intended usage before modifying the component.

kirkwaiblinger:
@KillerBean feel free to continue the work within this PR, there's no reason to create a new one.

JoshuaKGoldberg (approved):
Thanks!

PR #12194, opened 2026-04-02
docs: clarify consistent-type-imports guidance for verbatimModuleSyntax

### LINKED ISSUE #11681: Docs: [consistent-type-imports] Revise recommended usage

### Suggested Changes

The `consistent-type-imports` page specifically recommends that if you use `verbatimModuleSyntax`, you should turn off this rule:

> Because there are some differences, using both this rule and `verbatimModuleSyntax` at the same time can lead to conflicting errors. As such we recommend that you only ever use one _or_ the other -- never both.

But this rule does something very important that `verbatimModuleSyntax` does not: it makes sure you use `import type` for imports used only as types.

> `consistent-type-imports` is concerned with exactly one thing - ensuring your value imports are used as values, warning you when your value imports aren't used as values. https://github.com/typescript-eslint/typescript-eslint/issues/6379#issuecomment-1405902005

So in this situation:

```ts
import { GraphQLSchema } from "graphql";

export class Foo {
  private readonly schema: GraphQLSchema;

  constructor(schema: GraphQLSchema) {
    this.schema = schema;
  }

  // …
}
```

If you enable `verbatimModuleSyntax`, oops, now you have a side-effect import of `"graphql"`—and all of the weird problems that brings (https://github.com/graphql/graphql-js/issues/2801).

For this reason, I think it would be good to update the recommendations for this rule, since it can specifically help guard against unwanted side effects once `verbatimModuleSyntax` is enabled.

### Affected URL(s)

https://typescript-eslint.io/rules/consistent-type-imports

### Additional Info

_No response_

kyungilcho:
## PR Checklist

- [x] Addresses an existing open issue: fixes #11681

## Overview

Clarifies that `verbatimModuleSyntax` is not a direct replacement for `consistent-type-imports`.

The docs now explain the distinct responsibilities of the compiler option and the rule, update the comparison section, and align the FAQ wording so it no longer recommends disabling the rule outright when `verbatimModuleSyntax` is enabled.

💖

### DIFF
--- a/packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
+++ b/packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
@@ -113,23 +113,18 @@ Otherwise you can explicitly tell our tooling to analyze your code as if the com
 ## Comparison with `importsNotUsedAsValues` / `verbatimModuleSyntax`
 
 [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) was introduced in TypeScript v5.0 (as a replacement for `importsNotUsedAsValues`).
-It is not a direct replacement for this rule.
-`consistent-type-imports` is an ESLint rule that enforces how you _write_ imports.
-`verbatimModuleSyntax` is a TypeScript compiler option that controls how those imports are preserved in emitted JavaScript.
-
-There are a few important behavior differences:
+This rule and `verbatimModuleSyntax` overlap in some situations, but there are a few behavior differences:
 | Situation | `consistent-type-imports` (ESLint) | `verbatimModuleSyntax` (TypeScript) |
 | -------------------------------------------------------------- | --------------------------------------------------------- | ----------------------------------------------------------- |
-| `import { Foo } from 'foo'; type T = Foo;` | Reports and can auto-fix to `import type` | Allowed, and the import is preserved in emitted JavaScript |
+| Unused imports | Ignored (consider using [`@typescript-eslint/no-unused-vars`](/rules/no-unused-vars)) | Type error |
 | Usage with `emitDecoratorMetadata` & `experimentalDecorators` | Ignores files that contain decorators | Reports on files that contain decorators |
 | Failures detected | Does not fail `tsc` build; can be auto-fixed with `--fix` | Fails `tsc` build; cannot be auto-fixed on the command-line |
-| `import { type T } from 'T';` | Allowed with `fixStyle: 'inline-type-imports'`; use `fixStyle: 'separate-type-imports'` to prefer `import type` | Emits `import {} from 'T'` |
+| `import { type T } from 'T';` | TypeScript will emit nothing (it "elides" the import) | TypeScript emits `import {} from 'T'` |
 
-If your goal is to make type-only imports explicit in source code, keep this rule enabled even when using `verbatimModuleSyntax`.
-If your goal is only predictable emit semantics, `verbatimModuleSyntax` may be sufficient on its own.
-
-Projects that use both often also want [`no-import-type-side-effects`](./no-import-type-side-effects.mdx), especially when using inline `type` specifiers.
-If the overlap between ESLint and TypeScript diagnostics is noisy in your project, you can choose one of them as the source of truth.
+Because there are some differences, using both this rule and `verbatimModuleSyntax` at the same time can lead to duplicate reports on the same import.
+For example, TypeScript may report that a type must be imported using `import type`, and this rule may also report that the import should use your configured style.
+If you want ESLint autofixes and style enforcement for explicit `import type`, keep this rule enabled.
+If TypeScript's diagnostics are sufficient for your project, you can disable this rule to avoid duplicate reports.
 
 ## When Not To Use It
 
--- a/packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
+++ b/packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
@@ -113,17 +113,23 @@ Otherwise you can explicitly tell our tooling to analyze your code as if the com
 ## Comparison with `importsNotUsedAsValues` / `verbatimModuleSyntax`
 
 [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) was introduced in TypeScript v5.0 (as a replacement for `importsNotUsedAsValues`).
-This rule and `verbatimModuleSyntax` _mostly_ behave in the same way.
-There are a few behavior differences:
+It is not a direct replacement for this rule.
+`consistent-type-imports` is an ESLint rule that enforces how you _write_ imports.
+`verbatimModuleSyntax` is a TypeScript compiler option that controls how those imports are preserved in emitted JavaScript.
+
+There are a few important behavior differences:
 | Situation | `consistent-type-imports` (ESLint) | `verbatimModuleSyntax` (TypeScript) |
 | -------------------------------------------------------------- | --------------------------------------------------------- | ----------------------------------------------------------- |
-| Unused imports | Ignored (consider using [`@typescript-eslint/no-unused-vars`](/rules/no-unused-vars)) | Type error |
-| Usage with `emitDecoratorMetadata` & `experimentalDecorations` | Ignores files that contain decorators | Reports on files that contain decorators |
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
```diff
@@ -113,17 +113,23 @@ Otherwise you can explicitly tell our tooling to analyze your code as if the com
...
-This rule and `verbatimModuleSyntax` _mostly_ behave in the same way.
-There are a few behavior differences:
+It is not a direct replacement for this rule.
+`consistent-type-imports` is an ESLint rule that enforces how you _write_ imports.
+`verbatimModuleSyntax` is a TypeScript compiler option that controls how those imports are preserved in emitted JavaScript.
+
+There are a few important behavior differences:
 | Situation | `consistent-type-imports` (ESLint) | `verbatimModuleSyntax` (TypeScript) |
 | -------------------------------------------------------------- | --------------------------------------------------------- | ----------------------------------------------------------- |
-| Unused imports | Ignored (consider using [`@typescript-eslint/no-unused-vars`](/rules/no-unused-vars)) | Type error |
```
JoshuaKGoldberg, line 120: Why remove the _Unused imports_ situation? That's a good one to note, I'd think.

packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
```diff
@@ -113,17 +113,23 @@ Otherwise you can explicitly tell our tooling to analyze your code as if the com
 ## Comparison with `importsNotUsedAsValues` / `verbatimModuleSyntax`
 
 [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) was introduced in TypeScript v5.0 (as a replacement for `importsNotUsedAsValues`).
-This rule and `verbatimModuleSyntax` _mostly_ behave in the same way.
-There are a few behavior differences:
+It is not a direct replacement for this rule.
+`consistent-type-imports` is an ESLint rule that enforces how you _write_ imports.
+`verbatimModuleSyntax` is a TypeScript compiler option that controls how those imports are preserved in emitted JavaScript.
```
JoshuaKGoldberg: This is not correct. Both change how you write imports. With `verbatimModuleSyntax`, you can in fact get type errors that enforce the way you write imports.

```ts
src/options/readPackageData.ts:4:10 - error TS1484: 'PartialPackageData' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

4 import { PartialPackageData } from "../types.js";
```

packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
```diff
@@ -113,17 +113,23 @@ Otherwise you can explicitly tell our tooling to analyze your code as if the com
...
-| `import { type T } from 'T';` | TypeScript will emit nothing (it "elides" the import) | TypeScript emits `import {} from 'T'` |
+| `import { type T } from 'T';` | Allowed with `fixStyle: 'inline-type-imports'`; use `fixStyle: 'separate-type-imports'` to prefer `import type` | Emits `import {} from 'T'` |
+
+If your goal is to make type-only imports explicit in source code, keep this rule enabled even when using `verbatimModuleSyntax`.
+If your goal is only predictable emit semantics, `verbatimModuleSyntax` may be sufficient on its own.
 
-Because there are some differences, using both this rule and `verbatimModuleSyntax` at the same time can lead to conflicting errors.
-As such we recommend that you only ever use one _or_ the other -- never both.
+Projects that use both often also want [`no-import-type-side-effects`](./no-import-type-side-effects.mdx), especially when using inline `type` specifiers.
+If the overlap between ESLint and TypeScript diagnostics is noisy in your project, you can choose one of them as the source of truth.
```
JoshuaKGoldberg: This sentence is pretty vague. Both the whole thing in general and especially:

> you can choose one of them as the source of truth

* Can you address the root cause of why the noise would happen?
  * Many readers read these docs to learn these things and wouldn't have the practical experience to know it already.
* What does "choose one of them" mean?

JoshuaKGoldberg (changes requested):
Thanks for getting started on this! I think you went down a little deeper of a rabbit hole in a few areas than was necessary. I'd suggest rereading the issue & TypeScript docs to regain context, try out the referenced lint rules & compiler options a bit, rework this PR, and then re-request review once you feel confident it's ready. Cheers!

kyungilcho:
Thanks, that helped. I'd gone too far in trying to redefine the relationship between the rule and `verbatimModuleSyntax`.

I trimmed the docs back to stay closer to the original section, restored the unused-imports example, and rewrote the recommendation to focus on the practical tradeoff: enabling both can lead to duplicate reports on the same import, but the rule can still be useful if you want ESLint autofixes and style enforcement for explicit `import type`.

I also tightened the FAQ wording to match. Pushed in `34d167252`.

packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
kyungilcho: Good call. I restored that row.

packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
kyungilcho: You were right. That wording drew too sharp a line between the rule and `verbatimModuleSyntax`. I removed that framing and narrowed the section back down to the concrete differences that matter here.

packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
kyungilcho: Makes sense. I rewrote this to spell out the actual source of the noise: both tools can report on the same import, with TypeScript enforcing a type-only import and the rule also enforcing the configured import style. I also replaced “source of truth” with the more concrete choice between keeping ESLint autofixes/style enforcement or relying on TypeScript’s diagnostics alone.

kyungilcho:
@JoshuaKGoldberg I've addressed the review feedback, replied on the threads, and all checks are green now. When you have a chance, could you take another look?

packages/eslint-plugin/docs/rules/consistent-type-imports.mdx
```diff
@@ -113,17 +113,18 @@ Otherwise you can explicitly tell our tooling to analyze your code as if the com
...
 | Unused imports | Ignored (consider using [`@typescript-eslint/no-unused-vars`](/rules/no-unused-vars)) | Type error |
-| Usage with `emitDecoratorMetadata` & `experimentalDecorations` | Ignores files that contain decorators | Reports on files that contain decorators |
+| Usage with `emitDecoratorMetadata` & `experimentalDecorators` | Ignores files that contain decorators | Reports on files that contain decorators |
 | Failures detected | Does not fail `tsc` build; can be auto-fixed with `--fix` | Fails `tsc` build; cannot be auto-fixed on the command-line |
 | `import { type T } from 'T';` | TypeScript will emit nothing (it "elides" the import) | TypeScript emits `import {} from 'T'` |
 
-Because there are some differences, using both this rule and `verbatimModuleSyntax` at the same time can lead to conflicting errors.
-As such we recommend that you only ever use one _or_ the other -- never both.
+Because there are some differences, using both this rule and `verbatimModuleSyntax` at the same time can lead to duplicate reports on the same import.
```
kirkwaiblinger: The new version is a non sequitur as written: the differences between them don't cause _duplicate_ reports (it's the similarities that do). The original is more accurate.

perhaps we can succinctly incorporate both like so?

```suggestion
Because they have slightly different goals, using both this rule and `verbatimModuleSyntax` at the same time can lead to conflicting and/or duplicate reports on the same import.
```

kirkwaiblinger (commented):
One comment on some awkward wording but otherwise I think this LGTM 👍

kirkwaiblinger:
@kyungilcho Do you have a few minutes to address the final pending feedback on this? It's very close to the finish line and I'd love to get it merged 🙂

JoshuaKGoldberg:
👋 @kyungilcho did you mean to re-request review on this?

kyungilcho:
@JoshuaKGoldberg Yes, exactly. Sorry for the slow reply - I just saw your comment. I meant to re-request review after pushing the final wording tweak from kirk's feedback in `ab5b620a3`. If you have a moment, I'd appreciate another look.

JoshuaKGoldberg (approved):
LGTM, thanks! ✨

kirkwaiblinger (approved):

PR #12206, opened 2026-04-08
feat(eslint-plugin): check inferred types (`no-unnecessary-type-arguments`)

### LINKED ISSUE #4978: Enhancement: [no-unnecessary-type-arguments] Detect inferred type

ej-shafran:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4978

## Overview

Take two! This time, adding tests for all sorts of issues that popped up when this was merged the first time, and keeping the new logic behind an opt-in option so that issues are less pressing.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
+++ b/packages/eslint-plugin/src/rules/no-unnecessary-type-arguments.ts
@@ -22,9 +22,15 @@ type ParameterCapableTSNode =
   | ts.TypeQueryNode
   | ts.TypeReferenceNode;
 
-export type MessageIds = 'isDefaultTypeArgument';
+export type MessageIds = 'isDefaultTypeArgument' | 'canBeInferred';
 
-export default createRule<[], MessageIds>({
+export type Options = [
+  {
+    checkForInferrableTypes?: boolean;
+  },
+];
+
+export default createRule<Options, MessageIds>({
   name: 'no-unnecessary-type-arguments',
   meta: {
     type: 'suggestion',
@@ -37,10 +43,24 @@ export default createRule<[], MessageIds>({
     messages: {
       isDefaultTypeArgument:
         'This is the default value for this type parameter, so it can be omitted.',
+      canBeInferred:
+        'This value can be trivially inferred for this type parameter, so it can be omitted.',
     },
-    schema: [],
+    defaultOptions: [{}],
+    schema: [
+      {
+        type: 'object',
+        additionalProperties: false,
+        properties: {
+          checkForInferrableTypes: {
+            type: 'boolean',
+            description:
+              'Whether to check for type arguments that could be inferred.',
+          },
+        },
+      },
+    ],
   },
-  defaultOptions: [],
   create(context) {
     const services = getParserServices(context);
     const checker = services.program.getTypeChecker();
@@ -80,6 +100,7 @@ export default createRule<[], MessageIds>({
     function checkTSArgsAndParameters(
       typeArguments: TSESTree.TSTypeParameterInstantiation,
       typeParameters: readonly ts.TypeParameterDeclaration[],
+      tsNode: ParameterCapableTSNode,
     ): void {
       // Just check the last one. Must specify previous type parameters if the last one is specified.
       const typeArgumentIndex = typeArguments.params.length - 1;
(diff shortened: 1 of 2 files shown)


ej-shafran:
Hello again! Using this comment to point out some things that might still be missing/problematic, or that I need assistance with. Until these are resolved, I'm keeping this PR as a draft.

## Failing Test Case

- [ ] Resolved

Currently, a test case is failing:

```ts
type TUnionType = { type: 'idle' | 'busy' } | { type: 'error', message: string };

declare const defaultVal: TUnionType | undefined;

declare const func: <T>(val: T) => T;

const val1 = func<TUnionType>(defaultVal ?? { type: 'idle' });
```

This is because passing the `TUnionType` type argument essentially coerces the `{ type: "idle" }` literal type be of type `{ type: "idle" | "busy" }`, whereas not passing this type argument means the literal is of type `{ type: string }`.

What do we want to do here? Is this somewhere where an error is to be expected, and the code should explicitly say `{ type: "idle" as const }` or `{ type: "idle" } as TUnionType`? If not, how can we tell that this is somewhere where the type argument is necessary?

Either way - it seems that the suggestion/code fix being to remove the type altogether is causing issues. It's not necessarily true that we can just remove the unnecessary type argument and things will work out just the same. Should I remove the automatic fix, and leave just suggestions?

## Fix Prevents a Different Test Case

- [ ] Resolved

In this PR, I added code that confirms that the "offending" type argument can actually be removed (i.e. it either has a default type argument or it's the very first type argument, thus allowing it to be unspecified). This prevents one of the issues that popped up from occurring. However, it makes (re)adding this invalid test case a problem:

```ts
{
  code: `
function g<T, U>(x: T, y: U) {}
g<number, number>(10, 10);
      `,
  options: [{ checkForInferrableTypes: true }],
  errors: [
    {
      messageId: 'canBeInferred',
    },
  ],
  output: [
    `
function g<T, U>(x: T, y: U) {}
g<number>(10, 10);
      `,
    `
function g<T, U>(x: T, y: U) {}
g(10, 10);
      `,
  ],
}
```

This kind of makes sense: the first `output` here is invalid TypeScript code, since you can't pass the `T` type argument and not pass `U`.

Should this rule detect cases where _all_ the type arguments are unnecessary? Should we detect cases where non-removable type arguments are still unnecessary, and simply not offer a fix or a suggestion?

## Notes

- In this PR, I tried _not_ to use assignability to check if types are equivalent, but instead to use an extracted version of `areTypesEquivalent` based on the code that already exists in the rule right now. It seems to be working fine, but might have edge cases I haven't thought about...

StyleShit:
Hey @ej-shafran! Just checking in, is this still something you have the time/energy to work on?

ej-shafran:
Hey! Sorry for not replying, I was on my honeymoon :))
I'm available to continue working on this, if that's still relevant.

StyleShit:
Sure, feel free to reopen and continue the work when you're available

And congrats! 😄

ej-shafran:
Thanks!

Could someone help me decide what to do with the two test cases I listed above? I could arbitrarily choose a direction but I'd prefer to have some insight ahead of time if possible.

StyleShit:
> What do we want to do here? Is this somewhere where an error is to be expected... how can we tell that this is somewhere where the type argument is necessary?

IMO it shouldn't be flagged because `T` is not the same. Passing a type argument narrows the type. Citing Brad: 
"it is acceptable to _narrow_ the type ..."
"it is **unacceptable** to use the exact same type."

> Should this rule detect cases where all the type arguments are unnecessary? 

I think so, yes

> Should we detect cases where non-removable type arguments are still unnecessary, and simply not offer a fix or a suggestion?

If it's not removable, I _think_ we shouldn't flag it?

And for the specific case you mentioned, we can maybe detect such cases and remove all the type arguments at once? Not sure how it'll complicate things, though. If that's too complicated, I guess we can ignore it for now and have a follow-up PR for this

Pinging @typescript-eslint/triage-team to hear their thoughts

